PR #622, opened 2015-05-29
Implement Scala2 traits

odersky:
Added new phases that let Dotty classes implement Scala 2.x compiled traits.

The treatment of lazy vals is probably not yet correct. @DarkDimius can you take a look.

If @retronym or @lrytz could take a look to review this would be helpful as well.


odersky:
Rebased to #623

odersky:
Rebased to latest make-tests-pass to be able to pass junit tests.

odersky:
Added one more commit to make tests pass and rebased to master.

odersky:
partest seems to hang for all PRs. The log says it's succesful but the task is not terminated.

smarter:
@adriaanm: any idea why /synch is necessary in all PRs, but only for partest?

lrytz:
i had a quick pass over the change, couldn't see any obvious issues. how will this be tested?

adriaanm:
I checked scabot's log but didn't see any irregularities. I'm on vacation this week, but will have a look next week. I would not recommend blindly trusting the result of /synch, because it seems it's marking things green that shouldn't be. I've never seen it do that before... I'll disable that behavior at least.

retronym:
Is this expected to work?

```
% tail sandbox/{scala-trait,dotty-subclass}.scala
==> sandbox/scala-trait.scala <==
trait T {
  def d = 42
  val v = ""
  object O
}

==> sandbox/dotty-subclass.scala <==
class Sub extends T

object Main {
    val sub = new Sub
    println(sub.d)
    println(sub.v)
    println(sub.O)
  }
}

% scalac sandbox/scala-trait.scala && ./bin/dotc sandbox/dotty-subclass.scala && scala Main
java.lang.NoSuchMethodError: T.$init$()V
	at Sub.<init>(dotty-subclass.scala)
```

The last tree from `-Xprint:all`

```
  class Sub extends Object with T {
    def <init>(): Sub = {
      super()
      super[T]()
      ()
    }
    private val v$$local: String
    <accessor> def v(): String = this.v$$local
    private val O$$local: T.this.T$O$
    <accessor> module def O(): T.this.T$O$ = this.O$$local
    <accessor> def T$_setter_$v_=(x$0: String): Unit = ()
    <accessor> def T$_setter_$O_=(x$0: T.this.T$O$): Unit = ()
  }
```

retronym:
Looks like `LinkScala2ImplClasses` is called for `Sub` before `Mixin::superCallOpt`. I'm not up to speed with dotty's mini-phase / `SymTransformer` architecture to understand why this happens.

Moving it to a new phase just before `LambdaLift` advances the error to:

```
java.lang.AbstractMethodError: Sub.d()I
	at Main$.main(dotty-subclass.scala:5)
	at Main.main(dotty-subclass.scala)
```

smarter:
> I would not recommend blindly trusting the result of /synch, because it seems it's marking things green that shouldn't be

Where are you seeing that?

odersky:
@retronym Oops that was not quite ready for prime time yet. I got distracted with the build problems. It's in somewhat better shape with the recent commits which address the problems you have encountered. There's a more comprehensive test in `scala2traits` (or rather a pre-test, it would be good if someone turned this into a real test). 

@darkdimius Lazy vals do not work yet under this scheme. I have left some code in the test files regarding to uncomment once they are implemented.

odersky:
Still something wrong with isCurrent. I'm investigating.

adriaanm:
> I would not recommend blindly trusting the result of /synch, because it seems it's marking things green that shouldn't be
>> Where are you seeing that?
Those WARNING statuses are green by default: https://github.com/scala/scabot/blob/master/server/src/main/scala/Actors.scala#L361 (they are stupidly added when the connection fails.. my bad)

adriaanm:
It looks like it's an issue with Jenkins. At the end of the partest job, it says:
```
Notifying endpoint 'HTTP:http://scala-ci.typesafe.com:8888/jenkins'
ERROR: Failed to notify endpoint 'HTTP:http://scala-ci.typesafe.com:8888/jenkins'
```

This means Scabot likely did not see the status update (it's listening at that URL for Jenkins to inform us about the job ending). I suspect it could be because the ansicolor plugin does some fancy formatting with the buildlog, which is part of the notification. Maybe it can't be serialized.

retronym:
@adriaanm I presume that dotty could turn off direct archiving of partest logs as a workaround?

adriaanm:
Yep, that would work around it, but I've deployed a patched notification plugin that doesn't include logs/artifacts (not needed anyway).

odersky:
This PR has strirred up quite a lot of trouble. Let's see whether the dust settles down now.

odersky:
Yay! And we did not even need a /synch! Thanks @adriaanm for this.

odersky:
@smarter There's plenty of precendent regarding equals != ==:

scala> 1L.equals(1)
res5: Boolean = false

scala> 1L == 1
res6: Boolean = true

smarter:
Yeah, but if I have a `val x` which I know is not null, then having `x == x` and `x.equals(x)` behave differently is pretty surprising. In any case breaking the semantics of value classes here does not buy us anything. We should explore allowing the user to override `equals` and `hashCode` in value classes but this will only work correctly if we have extension methods in universal traits (which is something I'm looking at).

odersky:
@smarter The advantage of Dotty's scheme is that it treats value classes and case classes in the same way. All the bad scenarios you describe hold equally for case classes.

smarter:
For value classes we rewrite `new Foo(3).==(new Foo(3))` to `3.==(4)` in `VCInline` because we assume that `equals` cannot be overriden. For case classes we don't make this assumption and don't do this rewriting, so `equals` and `==` stay consistent.

odersky:
And for hashcode? That's the Scala bug we were talking about. But I think
the resolution should still be that universal traits cannot override equals
and hashCode.



On Tue, Jun 2, 2015 at 3:32 PM, Guillaume Martres <notifications@github.com>
wrote:

> For value classes we rewrite new Foo(3).==(new Foo(3)) to 3.==(4) in
> VCInline because we assume that equals cannot be overriden. For case
> classes we don't make this assumption and don't do this rewriting, so
> equals and == stay consistent.
>
> —
> <https://github.com/lampepfl/dotty/pull/622#issuecomment-107953905>.
>



-- 
Martin Odersky
EPFL

smarter:
Yes, disallowing equals/hashCode in universal traits might be the only sane solution, it breaks the scala standard library but we've already given up on that anyway.

odersky:
@smarter Agreed. Can you open an issue? Thanks! -

DarkDimius:
> But I think the resolution should still be that universal traits cannot override equals and hashCode.

Why? Those methods could be handled by the same rewriting scheme, that will supersede current special rule for hashcode and equals, making whole scheme more regular. And what about toString?

smarter:
See #629.

adriaanm:
> Yay!

:tada: thanks -- happy it's working!

odersky:
Any more review remarks? Would be good to get this in, as it fixes quite a lot of problems and unblocks the bootstrapping project.

DarkDimius:
I propose we merge this one, and continue with run tests.

PR #621, opened 2015-05-29
Stop running scala/scala tests instead of our junit tests.

DarkDimius:

Issue #580, opened 2015-05-16
Regression: Lifted inner method tries to use this


smarter:
This is exactly the same issue than #545, it was fixed in https://github.com/lampepfl/dotty/commit/b0856feefb90d1c0647bb3725ed9797fc76e0603 but in the same pull request https://github.com/lampepfl/dotty/commit/32a70042fc8a0338e093dba0424c98d25fb3a754 caused a regression.
In the example from #545, the enclosing class is the module class `Test$` and the top class is also the module class `Test$`, so the test `encClass.isProperlyContainedIn(topClass)` fails.

Assigned to @DarkDimius .

odersky:
The fix causes the build to fail. ClassPath does not compile anymore. This is what I have been seeing all along. It shows the value of a test infrastructure we can rely on. Without it we are flying blind.

PR #623, opened 2015-05-29
Make tests pass

odersky:
Subsumes #621, adding fixes for the test failures. #580 is open again. @review by @DarkDimius @smarter


odersky:
I believe the tests hang. Running locally shows that lots of tests fail, all because of the revert of 78aa682. 
To summarize:

If 78aa682 is enabled junit tests fail because we cannot compile ClassPath. 
If the commit is disabled lots of run tests fail with verify errors. For instance:

Users/odersky/workspace/dotty/tests/run> dotc arybufgrow.scala -d /classes 
/Users/odersky/workspace/dotty/tests/run> java Test
Exception in thread "main" java.lang.VerifyError: Bad local variable type
Exception Details:
  Location:
    Test$.$anonfun$$init$$1(I)Lscala/collection/mutable/ArrayBuffer; @0: aload_0
  Reason:
    Type integer (current frame, locals[0]) is not assignable to reference type
  Current Frame:
    bci: @0
    flags: { }
    locals: { integer }
    stack: { }
  Bytecode:
    0x0000000: 2ab6 005e 126f b600 73b0               

The other failures are similar.

I am at the end of what I can do. @darkdimius you need to take this over. It's really important to get the build back to work quickly!

DarkDimius:
@odersky, taking this over.

DarkDimius:
this failure is indeed manifestation of reverted commit:
```
private static scala.collection.mutable.ArrayBuffer $anonfun$$init$$1(int);
    Code:
       0: aload_0
       1: invokevirtual #94                 // Method buf:()Lscala/collection/mutable/ArrayBuffer;
```

This is static method, that tries to access `this`.

odersky:
Curiously, the test of #580 was not affected by reverting the fix. It still works, but lots of other run tests fail. So I am bringing it back. @DarkDimius over to you now.

odersky:
@DarkDimius You might want to cherry-pick db11fc579f0bcfc9eaac94756538ddb0bad75
It did not help when I tested it, but maybe it just needs tweaking.

smarter:
@DarkDimius : Can you get this to rebuild? It seems that Jenkins restarted and now it's stuck saying "validate-partest — [85] Building. " even though the job was aborted.

smarter:
Forced rebuild by changing the commit message.

smarter:
Partest now finish running! It fails because of `tests/run/t6260-delambdafy.scala` for exactly the reason I explained in: https://github.com/smarter/dotty/commit/4d23983013c2520d331a0dfdbb1067c00bee37d2#commitcomment-11339871 It tries to check an implementation detail that can change. I'm going to push another commit to disable checking the methods of the lambda.

DarkDimius:
@smarter, we should not _disable_ checking of lamdas. Tests exist to ensure that the way how lambdas are generated does not change without our notice.
If the behavior changes, and the change is expected, the check file should be updated.

smarter:
OK, I just updated the check file, but this might be annoying if we change the details of lambda generation often.

smarter:
After fixing `ElimStaticThis`, all tests now pass! :sparkles: :sparkles:
@DarkDimius: For some reason github says that `validate-partest` is still `Building.`, but if you click on https://scala-ci.typesafe.com/job/dotty-master-validate-partest/88/ you'll still that the build finished successfully.

src/dotty/tools/dotc/transform/ElimStaticThis.scala
```diff
@@ -0,0 +1,43 @@
...
+    }
+    else tree
+
+  override def transformIdent(tree: tpd.Ident)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    val meth = ctx.owner.enclosingMethod
+    // We cannot use meth.enclosingClass because it skips other static classes,
+    // so instead we require this phase to run after Flatten and use meth.owner
+    if (meth.is(JavaStatic) && meth.owner.is(ModuleClass)) {
+      tree.tpe match {
+        case TermRef(thiz: ThisType, _) if (thiz.underlying.typeSymbol == meth.owner) =>
```
odersky: Instead of `.underlying.termSymbol` it's simpler to use just `.cls`.

odersky:
LGTM. Glad to have this working in master.

DarkDimius:
@odersky, some of the changes that have been merged in this PR should be reverted.
 3d240ad actually broke the transformation of inner objects, and it needs to be reverted if we want to move labdas to static methods in companion class.

PR #625, opened 2015-05-29
Trees: Add toString for AST nodes which are not case classes

smarter:
This makes debugging trees easier

Review by @odersky .

### DIFF
--- a/src/dotty/tools/dotc/ast/Trees.scala
+++ b/src/dotty/tools/dotc/ast/Trees.scala
@@ -346,7 +346,9 @@ object Trees {
   }
 
   class BackquotedIdent[-T >: Untyped] private[ast] (name: Name)
-    extends Ident[T](name)
+    extends Ident[T](name) {
+    override def toString = "BackquotedIdent($name)"
+  }
 
   /** qualifier.name */
   case class Select[-T >: Untyped] private[ast] (qualifier: Tree[T], name: Name)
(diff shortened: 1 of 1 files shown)


adriaanm:
https://github.com/scala/scabot/blob/master/README.md#commands

adriaanm:
The only time I'm aware of you need to trigger a synch is when Jenkins was restarted while scabot was fetching the job status. It normally checks all PRs but it keeps a cache so it can get out of synch

odersky:
LGTM

src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -346,7 +346,9 @@ object Trees {
   }
 
   class BackquotedIdent[-T >: Untyped] private[ast] (name: Name)
-    extends Ident[T](name)
+    extends Ident[T](name) {
+    override def toString = "BackquotedIdent($name)"
```
namin: missing the s prefix for string interpolation

src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -355,7 +357,9 @@ object Trees {
   }
 
   class SelectWithSig[-T >: Untyped] private[ast] (qualifier: Tree[T], name: Name, val sig: Signature)
-    extends Select[T](qualifier, name)
+    extends Select[T](qualifier, name) {
+    override def toString = "SelectWithSig($qualifier, $name, $sig)"
```
namin: here too missing the s for string interpolation

src/dotty/tools/dotc/ast/Trees.scala
smarter: woops, thanks for catching that.

Issue #592, opened 2015-05-20
Typing of the method ignores knowledge that could be collected from type bounds on method targs


DarkDimius:
```
trait A {
  type T1
  type T2
  def foo[B <: T1 >: T2] = ... // is T2 <:< T1 in method?
}
``` 

As in order to call method `foo` one needs to verify type bounds, it seems safe for me to assume `T2 <:< T1` in the body of the method

This is needed for some compiler rewritings to be type-safe. Eg, specialization of a 
```
class Foo[+A] {
  def bop[@specialized B >: A]: Foo[B] = this
}
```

will yield
```
  def bop_sp[@specialized B >: A & Int <: Int & Any]: Foo[B] = this
```
Without typer knowing that `Foo$A` is subtype of `Int` in this example, specialization needs to insert casts.
Unfortunately, for the scheme to be complete, specialization needs to insert casts in every place where we have a source of expected type: arguments of methods, vals, defs. Otherwise the tree does not typecheck.

DarkDimius:
ping @odersky @adriaanm @smarter @samuelgruetter @namin 

Handling this case seems to account for 90% of problems with specialization that we have seen so far.

samuelgruetter:
So in the body of `def bop_sp[@specialized B >: A & Int <: Int & Any]: Foo[B]`, you want to have the knowledge that `A <: Int`, and you think that

```
A & Int <: Int & Any   ==>   A <: Int
```

The problem is that the above implication does not hold, because `A & Int <: Int & Any` could have been derived from `Int <: Int & Any`, which could have been derived from `Int <: Int ∧ Int <: Any`.

Note: This is (at least slightly) related to https://github.com/lampepfl/dotty/issues/525, where we see that dotty checks that the lower bound of a type parameter always is a subtype of the upper bound.

DarkDimius:
>So in the body of def bop_sp[@specialized B >: A & Int <: Int & Any]: Foo[B], you want to have the knowledge that A <: Int, and you think that
```
A & Int <: Int & Any   ==>   A <: Int
```

@samuelgruetter , I had an error in my original text. Please see the update.

```
def bop_sp[@specialized B >: A  <: Int & Any]: Foo[B]
```

samuelgruetter:
Do you want to the return type to be `Foo[B]` or `Foo[Int]`?

Return type `Foo[B]` works in scalac:

```
class Foo[+A] {
  def bop_sp[B >: A <: Int]: Foo[B] = this
}
[...]

PR #630, opened 2015-06-03
Implement method type specialisation

AlexSikia:
This implements method type specialization.

@DarkDimius Please review (and of course let me know if I should do anything differently with the pull request!)
Thanks


bench/src/test/test.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
DarkDimius: This shouldn't be part of PR.

DarkDimius:
Alex, a lot of your commits do reformatting of already existing code for not good reason(I guess your IDE has different formatting options), could you please update your commits to not do this, as this affects git history.


Also, you have committed files that are specific to your IDE(I guess that's Intellij) that shouldn't be part of PR.

Neither should there be anything in out/ folder.

DarkDimius:
Junit failed in GenBCode, as output directory does not exist..
`out/.keep` should be kept.

DarkDimius:
Great to see all the tests pass :+1: 

PS: this PR only contains method specialization, it does not handle classes or fields.

src/dotty/tools/dotc/transform/TypeSpecializer.scala
```diff
@@ -0,0 +1,281 @@
...
+  private val newSymbolMap: mutable.HashMap[Symbol, mutable.HashMap[List[Type], Symbols.Symbol]] = mutable.HashMap.empty
+
+  def allowedToSpecialize(sym: Symbol, numOfTypes: Int)(implicit ctx: Context): Boolean = {
+    numOfTypes <= maxTparamsToSpecialize &&
+      numOfTypes > 0 &&
+      sym.name != nme.asInstanceOf_ &&
+      sym.name != nme.isInstanceOf_ &&
+      !(sym is Flags.JavaDefined) &&
+      !sym.isConstructor &&
+      !sym.name.toString.contains("Function2")
```
DarkDimius: why?

VladUreche:
How did you guys fix the transitivity issue?

DarkDimius:
As a last transformation step, in all places where expected type propagation starts, [casts are introduced](https://github.com/AlexSikia/dotty/commit/bdfb1c1233f80eca7bce755ad2ea4bb100d1087d#diff-a979e1f98bfa8b4c0d7cba7a8f64f8f6R193).

This isn't a solution, this is a hacky-workaround. I do not see a good solution besides #592.

src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -326,6 +326,7 @@ class Definitions {
   lazy val TransientAnnot                = ctx.requiredClass("scala.transient")
   lazy val NativeAnnot                   = ctx.requiredClass("scala.native")
   lazy val ScalaStrictFPAnnot            = ctx.requiredClass("scala.annotation.strictfp")
+  lazy val specializedAnnot              = ctx.requiredClass("scala.specialized")
```
DarkDimius: capital S

tests/pos/specialization/bounds_specialization.scala
```diff
@@ -0,0 +1,21 @@
+object bounds_specialization {
+  /*class Foo[@specialized K] {
```
DarkDimius: why is this test commented?

tests/pos/specialization/nothing_specialization.scala
```diff
@@ -0,0 +1,8 @@
+object nothing_specialization {
+  def ret_nothing[@specialized(Char) T] = {
+    //val a: List[T] = List[Nothing]()
```
DarkDimius: why is it commented?

src/dotty/tools/dotc/transform/PreSpecializer.scala
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * This phase retrieves all `@specialized` anotations before they are thrown away,
+ * and stores them for the `TypeSpecializer` phase.
+ */
+class PreSpecializer extends MiniPhaseTransform {
+
+  override def phaseName: String = "prespecialize"
+
+  private final def primitiveCompanionToPrimitive(companion: Type)(implicit ctx: Context) = {
+    if (companion.asInstanceOf[TermRef].name.toString == "AnyRef") { // Handles `@specialized(AnyRef)` cases
```
DarkDimius: why based on names, not on symbols?

DarkDimius:
```scala
trait Spec {
 def plus[@specialized T](a: T, b:T)(ev: Numeric[T]): T = plus(b, a)(ev)
}
```

Generates specialized versions, but does not dispatch to them in recursive call.

VladUreche:
@DarkDimius @AlexSikia doesn't casting introduce boxes?

DarkDimius:
@VladUreche the problems we've been seeing here mostly have to do with type arguments of enclosing scope, that are bounded by specialized types defined in inner scope. This happens infrequently in real code, and when it happens, it's mostly in higher kinded types, ie casts will be erased.

Examples can be found here: https://github.com/AlexSikia/dotty/blob/add/Type-specialisation/tests/pos/specialization/this_specialization.scala

But yes, it could potentially introduce boxing, and this will could some calls from being dispatched to specialized versions.

src/dotty/tools/dotc/transform/TypeSpecializer.scala
```diff
@@ -0,0 +1,284 @@
...
+        }).flatten
+      }
+      else {
+        generateSpecializedSymbols(instantiations.reverse, names.reverse, poly, decl)
+      }
+    }
+    def generateSpecializedSymbols(instantiations: List[Type], names: List[String], poly: PolyType, decl: Symbol)
+                                  (implicit ctx: Context): List[Symbol] = {
+      val newSym =
+        ctx.newSymbol(decl.owner, (decl.name + "$mc" + names.mkString + "$sp").toTermName,
```
smarter: If I understand the name mangling of specialization correctly, this should be `(decl.name + "$m" + names.mkString + "c$sp")`.
 `m` means method and `c` means class, for example if there's a specialized Int method in a specialized Int class, the signature will contain `$mIcI$sp`.

src/dotty/tools/dotc/transform/TypeSpecializer.scala
```diff
@@ -0,0 +1,284 @@
...
+    Map(defn.ByteType -> "B",
+      defn.BooleanType -> "Z",
+      defn.ShortType -> "S",
+      defn.IntType -> "I",
+      defn.LongType -> "J",
+      defn.FloatType -> "F",
+      defn.DoubleType -> "D",
+      defn.CharType -> "C",
+      defn.UnitType -> "V",
+      defn.AnyRefType -> "L")
```
retronym: This seems inefficient, the calling code `specialisedTypeToSuffix.apply(tp)` created the entire map and then throws it away.

DarkDimius: This function is simply not needed. There's already a `specializedFor` in NameOps: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/NameOps.scala#L239

DarkDimius: Thought this method currently does not handle method type arguments. This functionality should be added.

AlexSikia: Thanks for pointing this out! I'll take care of changing this when I reach Geneva later today.

src/dotty/tools/dotc/transform/TypeSpecializer.scala
AlexSikia: Ah, good point; thanks for finding that.

DarkDimius:
needs to be rebased over current master.

DarkDimius:
@AlexSikia, This PR has been around for quite some time. Let's do a push and get this in?

odersky:
I think it would be better to keep this separate until we have tackled the larger problem of class specialization.

smarter:
Commit https://github.com/AlexSikia/dotty/commit/07e9ae12d574fa0bbb07012cd78611973e2afeb7 should be removed from this pull request, currently it's just an accidental revert of https://github.com/lampepfl/dotty/commit/10ea7125998b0246e899706f407cc57719f37086 .

odersky:
Superseded by new work

PR #633, opened 2015-06-07
Fix/refs to inner objects

odersky:
Several formerly disabled tests are brought back. One required fixes of references to inner objects. Review by @DarkDimius


src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -326,23 +326,24 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
-          }
-        prefix.select(tp)
+      case pre: SingletonType => correctForPath(singleton(pre)).select(tp)
       case pre => SelectFromTypeTree(TypeTree(pre), tp)
     } // no checks necessary
 
   def ref(sym: Symbol)(implicit ctx: Context): Tree =
     ref(NamedType(sym.owner.thisType, sym.name, sym.denot))
 
+  private def correctForPath(t: Tree)(implicit ctx: Context) = t match {
```
DarkDimius: What does `For` in the name stand for? 
As this method is used only in `ref` and is private anyway, maybe it's better to move its definition to `ref` and give it a proper name, eg `followOuterLinks`?

DarkDimius:
Otherwise LGTM

odersky:
Since I am in train bringing back disabled tests, here's one more. AFAIK that's all the tests we had to disable in the past weeks, so we should be good on that side for now.

DarkDimius:
LGTM

PR #635, opened 2015-06-08
Some fixes around mixin and memoization

odersky:
(1) Fix implementations of private trait setters
(2) Copy annotations from trait methods to their implementations
(3) Copy selected annotations from getters to fields

PR #636, opened 2015-06-08
Add @volatile to scala2 lazyvals

odersky:
Add @volatile to inherited lazyvals. Right now this dies in backed. Based on #635. @DarkDimius can you take this over?


DarkDimius:
@odersky please review last commit.

PR #637, opened 2015-06-09
Fix #629 - disallow equals and hashCode in universal traits

### LINKED ISSUE #629: Disallow equals and hashCode in universal traits
`equals` and `hashCode` should always be synthesised in value classes so that `new Meter(3) == new Meter(4)` can be rewritten as `3 == 4`, so allowing `equals` and `hashCode` in universal traits doesn't make much sense.

odersky:
Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/transform/SymUtils.scala
+++ b/src/dotty/tools/dotc/transform/SymUtils.scala
@@ -48,6 +48,9 @@ class SymUtils(val self: Symbol) extends AnyVal {
   def isAnyOverride(implicit ctx: Context) = self.is(Override) || self.is(AbsOverride)
     // careful: AbsOverride is a term only flag. combining with Override would catch only terms.
 
+  def isUniversalTrait(implicit ctx: Context) =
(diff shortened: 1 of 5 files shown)


DarkDimius:
@odersky, I'd like to discuss this matter a bit more, can this wait until Friday?

densh:
I'd like to see support for custom equals and hashCode for scala-offheap use case. 

For example when one defines a simple data class:

```scala
@data class Point(x: Int, y: Int)
```

It desugars into value class:

```scala
class Point(addr: Addr) extends AnyVal { ... }
```

Without ability to define custom equals/hashCode we are bound to always have by-reference equality based on underlying field. What we would like to have instead is ability to have equality and hashCode based on values `x` and `y` fields that is only possible if one can redefine those for value classes and hopefully universal traits too.

odersky:
It seems we do not want to go ahead with this, and rather make hashCode and equals work efficiently when inherited from universal traits.

PR #641, opened 2015-06-15
Evaluate annotations in Namer

odersky:
Annotations were evaluated in Typer, which meant that they could be used only after type checking was
complete (or compilation order dependencies would be introduced). This commit makes annotations be
installed as part of completion in Namer. However, annotations are now lazy so as to avoid cyclic
references. This is completely analogous to the scheme in unpickler and ClassfileParser.

### DIFF
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -15,6 +15,7 @@ import annotation.tailrec
 import ErrorReporting._
 import tpd.ListOfTreeDecorator
 import config.Printers._
+import Annotations._
 import language.implicitConversions
 
 trait NamerContextOps { this: Context =>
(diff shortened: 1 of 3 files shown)


odersky:
Review by @xeno-by

odersky:
@xeno-by I would like to do it even earlier, during enter, but that currently does not pass the test suite.

xeno-by:
I will try take a look tomorrow. Would that work for you?

odersky:
@xeno-by Sure.

xeno-by:
So, if I'm getting this right, this pull requests makes typechecking annotation types eager upon completion of the underlying member, but annotation arguments still remain lazy?

odersky:
I tried to follow @olhotak's suggestion but that caused errors, unfortunately. A more conservative commit will follow.

odersky:
@olhotak Anything further, or can we merge this?

olhotak:
LGTM

PR #661, opened 2015-06-15
Fix #651 Generalize criterion for isOpAssign

### LINKED ISSUE #651: += is not a member of Set[String]
```
[info] # partest /Users/dark/workspace/dotty/tests/partest-generated/run/colltest1.scala
[info] % dotc colltest1.scala
[info]
[info] compiling /Users/dark/workspace/dotty/tests/partest-generated/run/colltest1.scala
[info] options: -pagewidth 160 -Yno-deep-subtypes -Yno-double-bindings -Ycheck:tailrec,resolveSuper,mixin,restoreScopes -d ./tests/partest-generated/run/colltest1-run.obj
[info] warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$GenSetFactory$$CC; assuming they are invariant
[info] warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$SetFactory$$CC; assuming they are invariant
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/colltest1.scala:148: error: += is not a member of scala.collection.Set[String]
[info]     s += ("D", "E", "F")
```

odersky:
Methods like + can have multiple parameters. In that case
+= also takes multiple parameters.

Review by @DarkDimius


tests/run/colltest1.scala
```diff
@@ -139,7 +139,8 @@ object Test extends dotty.runtime.LegacyApp {
 
   def setTest(empty: => Set[String]): Unit = {
     var s = empty + "A" + "B" + "C"
-    s += ("D", "E", "F")
```
DarkDimius: Is this modification intentional? it makes thus test not cover `+=` operation

odersky: No, that was an oversight.

PR #662, opened 2015-06-15
Add/resident compiler

odersky:
This adds a resident compiler command. For usage information, see doc comment in Resident.scala.
We'll later refine this to become a general compile server.

### DIFF
--- a/src/dotty/tools/dotc/Bench.scala
+++ b/src/dotty/tools/dotc/Bench.scala
@@ -8,19 +8,7 @@ package dotc
 import core.Contexts.Context
 import reporting.Reporter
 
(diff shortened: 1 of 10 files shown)


odersky:
I'd like to get this in before it starts to rot. Any objections?

odersky:
Rebased to current master

DarkDimius:
While being rebased this PR removed most of changes that were made in #691. I'll make a PR to bring them back

xeno-by:
What's the main benefit of having this feature? Is it cutting down the startup time of the compiler or there's more? Also, how does this compare with sbt-server?

PR #664, opened 2015-06-15
Enable tests that pass, move macro tests to disabled.

DarkDimius:
based on #636, all commits but e2a9573 should just move files around.

Total: 40 tests enabled, 1600+ tests disabled. Macros had quite good test base.


DarkDimius:
@odersky, please review.

odersky:
@DarkDimius Try with a parameterized `def foo`.

DarkDimius:
AFAIK ` ref(mixin.implClass).select(getter.name).suchThat(_ is Lazy)` isn't going to work: there's no particular tree that could be constructed by `ref(mixin.implClass).select(getter.name)`. What would be it's type? we do not have an `OverloadedType` that scalac has.

I've pushed a commit that does the same on denotations.

DarkDimius:
Unfortunately, Lazy flag isn't always set for the getter method inside implementation flag.

odersky:
LGTM

PR #666, opened 2015-06-18
Fix stack overflow when testing for shadowing

odersky:
Shadowing tests could go into an infinite recursion when
the found shadowing member itself needs an implicit that
is resolved and then shadowed again by the same member.

A test case is neg/arrayclone-new.scala. This caused
a SO before, now gives two errors.

review by @adriaanm


odersky:
Same problem as other #665: 

sbt.ResolveException: download failed: org.scalastyle#scalastyle-sbt-plugin;0.6.0!scalastyle-sbt-plugin.jar

adriaanm:
LGTM -- ideally, commits should be squashed.

odersky:
rebased to master

PR #668, opened 2015-06-19
Fix checking whether types are instantiable.

odersky:
The logic for checking aginst the self type was wrong, as demonstrated
by pos/checkInstantiable.scala. Review by @retronym.


src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -49,8 +49,11 @@ object Checking {
         if (cls.is(AbstractOrTrait))
           ctx.error(d"$cls is abstract; cannot be instantiated", pos)
         if (!cls.is(Module)) {
-          val selfType = tp.givenSelfType.asSeenFrom(tref.prefix, cls.owner)
-          if (selfType.exists && !(tp <:< selfType))
+          // Create a synthetic singleton type instance, and check whether
+          // it conforms to the self type of the class as seen from that instance.
+          val stp = SkolemType(tp)
```
retronym: Am I right in assuming that this is similar to `tp.narrow` in scalac?

retronym: In those terms, this LGTM, although I must admit I wouldn't have come up with this formulation myself.

```
scala> class LS[T] { self: LS[T] => }
defined class LS

scala> val cls = symbolOf[LS[_]]
cls: $r.intp.global.TypeSymbol = class LS

scala> val tp = typeOf[LS[Any]]
tp: $r.intp.global.Type = LS[Any]

scala> val stp = tp.narrow
stp: $r.intp.global.Type = <refinement>.type

scala> val selfType = cls.selfType.asSeenFrom(stp, cls)
selfType: $r.intp.global.Type = LS[Any]

scala> stp <:< selfType
res7: Boolean = true
```

vs

```
scala> class LS[T] { self: LS[T with String] => }
defined class LS

scala> val tp = typeOf[LS[Any]]
tp: $r.intp.global.Type = LS[Any]

scala> val cls = symbolOf[LS[_]]
cls: $r.intp.global.TypeSymbol = class LS

scala> val selfType = cls.selfType.asSeenFrom(stp, cls)
selfType: $r.intp.global.Type = LS[T with String]

scala> stp <:< selfType
res8: Boolean = false
```

odersky:
@retronym SkolemType is indeed a cheaper version of narrow. It took some iterations and dead ends to come up with the final formulation of checkInstantiable.

PR #672, opened 2015-06-19
Compile dotty.*, not only dotty.tools.

DarkDimius:
On my machine succeeds first compilation, but fails second, with >200 erorrs of such kind:

```
./src/dotty/./tools/dotc/core/Flags.scala:565: error: type mismatch:
 found   : (dotty.tools.dotc.core.Flags.FlagSet, dotty.tools.dotc.core.Flags.FlagSet)
 required: scala.collection.Seq[dotty.tools.dotc.core.Flags.FlagSet] @Repeated
  final val JavaInterface = allOf(JavaDefined, Trait)
```

@odersky, could you take it from here?

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -154,7 +154,7 @@ class tests extends CompilerTest {
 
   @Test def run_all = runFiles(runDir)
 
-  @Test def dotty = compileDir(dottyDir, "tools", "-deep" :: "-Ycheck-reentrant" :: allowDeepSubtypes ++ twice) // note the -deep argument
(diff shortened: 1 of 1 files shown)


DarkDimius:
@odersky, this pr updates bundled compiler fork to https://oss.sonatype.org/content/repositories/releases/me/d-d/scala-compiler/2.11.5-20150619-173733-3bcd390afa/scala-compiler-2.11.5-20150619-173733-3bcd390afa.jar

This is a binary incompatible update.

smarter:
@DarkDimius : have you pushed your changes to the compiler fork somewhere? https://github.com/DarkDimius/scala/commits/sharing-backend doesn't seem to have been updated.

PR #673, opened 2015-06-19
Refactor viewExists

odersky:
Factor out isValueSubClass into separate method. Will probably come in handly
elsewhere, and makes the code easier to understand.

Review by @darkdimius

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -601,6 +601,19 @@ object Types {
       ctx.typeComparer.isSameType(this, that)
     }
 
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/core/Types.scala
```diff
@@ -601,6 +601,19 @@ object Types {
       ctx.typeComparer.isSameType(this, that)
     }
 
+    /** Is this type a primitive value type which can be widened to the primitive value type `to`? */
```
DarkDimius: `to` -> `that`

DarkDimius:
Otherwise LGTM

PR #674, opened 2015-06-19
Fix/#655 eta expansion

odersky:
Review by @lrytz or @retronym. Based on #673, so only the second commit is new.


src/dotty/tools/dotc/core/Types.scala
```diff
@@ -601,6 +601,19 @@ object Types {
       ctx.typeComparer.isSameType(this, that)
     }
 
+    /** Is this type a primitive value type which can be widened to the primitive value type `to`? */
```
lrytz: typo `to` -> `that`

lrytz:
lgtm

retronym:
In the test case, what was the too-narrow type that needed widening? Sharing that sort of detail makes review easier (and comes in handy when looking back at the history of this code.)

retronym:
I just traced this:

```
methType=TermRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Test$)),foo)
methType.show=(str: => String)(i: Int)Test.Result(Test.foo)
methType.widen.show=(str: => String)(i: Int)Test.Result
```

After looking at the surrounding code, I would suggest that this widening should be done by the caller to `liftArgs`. This would be consistent with the way that `adaptInterpolated`/`adapt` passes the widened type as the argument for the `methType` parameter of `etaExpand`.

odersky:
It was a TermRef. Stupid mistake, easy to make.


On Fri, Jun 19, 2015 at 10:46 PM, Jason Zaugg <notifications@github.com>
wrote:

> In the test case, what was the too-narrow type that needed widening?
> Sharing that sort of detail makes review easier (and comes in handy when
> looking back at the history of this code.)
>
> —
> <https://github.com/lampepfl/dotty/pull/674#issuecomment-113638462>.
>



-- 
Martin Odersky
EPFL

retronym:
I see. In scalac we don't have to widen, because tree's have a `symbol` member to bind references to defintions, whereas dotc represents this with a `NamedType` instead.

odersky:
@retronym I don't think there's a fixed convention whether we do widen in the caller or callee. The way it is in EtaExpansion is DRYer. If we move to caller, we have to duplicate the call.

retronym:
In that case, the paramaeter name of `methType` is misleading.

odersky:
@retronym Good point.

PR #677, opened 2015-06-21
Closes #579 Implement mini-phase for classOf[T]

### LINKED ISSUE #579: classOf/getClass are not implemented
- `classOf[Int]` is erased to `classOf()` which is wrong. In `scalac` it is not erased but is replaced by `java.lang.Integer.TYPE` in `cleanup`.
- `2.getClass()` causes a crash in the backend, in `scalac` it is erased to `runtime.this.ScalaRunTime.anyValClass(scala.Int.box(2), (ClassTag.Int(): scala.reflect.ClassTag))`

alexander-myltsev:
Closes https://github.com/lampepfl/dotty/issues/579


DarkDimius:
Hi, Nice to see your first contribution!

src/dotty/tools/dotc/Compiler.scala
```diff
@@ -47,6 +47,7 @@ class Compiler {
            new ExtensionMethods,
            new ExpandSAMs,
            new TailRec),
+      List(new ClassOf),
```
DarkDimius: Looking at the implementation, it could be safely squashed with previous group.

src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -260,7 +260,7 @@ class Definitions {
 
   lazy val BoxedUnitClass = ctx.requiredClass("scala.runtime.BoxedUnit")
 
-    lazy val BoxedUnit_UNIT = BoxedUnitClass.linkedClass.requiredValue("UNIT")
```
DarkDimius: spacing was intentional, that's a derivative of `BoxedUnitClass`

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,44 @@
...
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+class ClassOf extends MiniPhaseTransform {
+  import tpd._
+
+  override def phaseName: String = "classOf"
+
+  override def transformTypeApply(tree: TypeApply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    val predefModule = ctx.definitions.ScalaPredefModule
+    val classOfMethod = ctx.requiredMethod(predefModule.moduleClass.asClass, "classOf".toTermName)
```
DarkDimius: it's better to pre-cache this value in `prepareForUnit`.

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    if (tree.symbol eq classOfMethod) {
+      val tp = tree.args.head.tpe
+      val defn = ctx.definitions
+      val claz = tp.classSymbol
+
+      if (ValueClasses.isDerivedValueClass(claz)) {
+        Literal(Constant(ref(claz.companionModule).tpe))
```
DarkDimius: this is a wrong class you are giving. It should be the class of the value class itself, not it's companion object.

DarkDimius: btw: it looks like this line isn't covered by tests. Can you please add one?

alexander-myltsev:
Glad to contribute :)

Regarding the test https://github.com/lampepfl/dotty/blob/master/tests/pending/run/classof.scala. I think it does not cover the case when an object of a class that is generated in `ClassOf` phase has same value of `toString` as `classof.check` expects, but not a class that is expected. For example, I might return some object of a class from `Dotty` that has `boolean` output of `toString` but is not `java.lang.Boolean.TYPE` under the hood.

Another question is that piece of code (https://github.com/lampepfl/dotty/blob/0d32755206e0f477c49e86b529e66e1154d7027f/src/dotty/tools/dotc/core/Definitions.scala#L274-282):

```scala
  lazy val BoxedBooleanModule = ctx.requiredModule("java.lang.Boolean")
  lazy val BoxedByteModule    = ctx.requiredModule("java.lang.Byte")
  lazy val BoxedShortModule   = ctx.requiredModule("java.lang.Short")
  lazy val BoxedCharModule    = ctx.requiredModule("java.lang.Character")
  lazy val BoxedIntModule     = ctx.requiredModule("java.lang.Integer")
  lazy val BoxedLongModule    = ctx.requiredModule("java.lang.Long")
  lazy val BoxedFloatModule   = ctx.requiredModule("java.lang.Float")
  lazy val BoxedDoubleModule  = ctx.requiredModule("java.lang.Double")
  lazy val BoxedVoidModule    = ctx.requiredModule("java.lang.Void")
```

Is there a better way to do it? I tried to use `lazy val BoxedBooleanModule = BoxedBooleanClass.linkedClass` but things do not work.

Ooops.. All tests should pass, right? I broke tests?

src/dotty/tools/dotc/core/Definitions.scala
alexander-myltsev: OK.

src/dotty/tools/dotc/transform/ClassOf.scala
alexander-myltsev: Sure.

src/dotty/tools/dotc/Compiler.scala
alexander-myltsev:
```scala
      List(new RefChecks,
           new ElimRepeated,
           new NormalizeFlags,
           new ExtensionMethods,
           new ExpandSAMs,
           new TailRec,
           new ClassOf),
```
Right?

src/dotty/tools/dotc/transform/ClassOf.scala
alexander-myltsev: OK.

DarkDimius:
> Ooops.. All tests should pass, right? I broke tests? 

The broken test is the one where typer inferred Class[Int] and this type was needed for typechecking, and you've replaced it by Class[Integer]. This does is the same before erasure, but isn't after erasure. The solution is to add `ensureConforms(tree.tpe)` in all branches handling primitives.

alexander-myltsev:
Should cases as follows be covered?:
```scala
    println(classOf[Any])  // current: class java.lang.Object
    println(classOf[AnyRef])  // current: class java.lang.Object
    println(classOf[Null])  // current: class scala.runtime.Null$
    println(classOf[Nothing])  // current: class scala.runtime.Nothing$
```

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,49 @@
...
+          case defn.BooleanClass => TYPE(defn.BoxedBooleanModule)
+          case defn.ByteClass    => TYPE(defn.BoxedByteModule)
+          case defn.ShortClass   => TYPE(defn.BoxedShortModule)
+          case defn.CharClass    => TYPE(defn.BoxedCharModule)
+          case defn.IntClass     => TYPE(defn.BoxedIntModule)
+          case defn.LongClass    => TYPE(defn.BoxedLongModule)
+          case defn.FloatClass   => TYPE(defn.BoxedFloatModule)
+          case defn.DoubleClass  => TYPE(defn.BoxedDoubleModule)
+          case defn.UnitClass    => TYPE(defn.BoxedVoidModule)
+          case _                 => Literal(Constant(TypeErasure.erasure(tp)))
```
smarter: If you write `Literal(Constant(TypeErasure.erasure(tp, semiEraseVCs = false)))` here, you should be able to remove the special case for value classes above.

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,49 @@
...
+class ClassOf extends MiniPhaseTransform {
+  import tpd._
+
+  override def phaseName: String = "classOf"
+
+  private var classOfMethod: TermSymbol = _
+
+  override def prepareForUnit(tree: tpd.Tree)(implicit ctx: Context): TreeTransform = {
+    val predefModule = ctx.definitions.ScalaPredefModule
+    classOfMethod = ctx.requiredMethod(predefModule.moduleClass.asClass, "classOf".toTermName)
```
smarter: Use `nme.classOf` instead of "classOf".toTermName, you might need to `import StdNames._`.

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,49 @@
...
+package transform
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.core.Constants.Constant
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Decorators.StringDecorator
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.TypeErasure
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo, TreeTransform}
+
```
smarter: This is not really important, but in dotty we usually don't write fully qualified import names, instead we usually write `import core._` then `import Contexts._, Symbols._`, etc.

src/dotty/tools/dotc/transform/ClassOf.scala
```diff
@@ -0,0 +1,49 @@
...
+class ClassOf extends MiniPhaseTransform {
```
smarter: There should be a comment documenting this phase before this line. See other phases like `VCInline` or `Getters` for examples of comments that document phases.

src/dotty/tools/dotc/transform/ClassOf.scala
DarkDimius: Do we actually currently have a single place that uses `semiEraseVCs = false`? 

If we do not, I'd prefer to leave as is, as this code is simple and easy to comprehend. Instead I'd better remove `semiEraseVCs` flag from erasure, as it makes it more complicated.

alexander-myltsev: :+1:

smarter: Yes, we do: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/Erasure.scala#L274

I don't think that `semiEraseVCs` is more confusing than a special case.

tests/run/classof.scala
```diff
@@ -1,8 +1,11 @@
+class ValueClass(val i: Integer) extends AnyVal
 class SomeClass
 
 object Test {
   def main(args: Array[String]): Unit = {
     val cls: Predef.Class[SomeClass] = classOf[SomeClass]
+    val valCls: Predef.Class[ValueClass] = classOf[ValueClass]
+
```
DarkDimius: It would be good to add a test, that tests the error that was triggered by previous build.
That could be:

```scala
val iCls: Class[Int] = classOf[Int]
```

src/dotty/tools/dotc/transform/ClassOf.scala
smarter: (It would probably help if `semiEraseVCs` was documented in `TypeErasure#erasure`, currently it's only documented for the class `TypeErasure`: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/TypeErasure.scala#L260-L262)

DarkDimius: `ErasedValueType` is internal affairs of erasure. It get's eliminated &etc, it is very, very special. This type shouldn't be eliminated\changed by other phases. And In backend I'll need to generate a bytecode representation of constant of this type. Backend does not know what `Constant(Literal(ErasedValueType))` is. And shouldn't.

I'd better have `ErasedValueType` contained and not leak into other phases without a good reason. Loosely coupled systems are better ones.

alexander-myltsev: This doesn't work. The output is as follows:
```scala
println(classOf[Unit]) // class scala.runtime.BoxedUnit instead of void
println(classOf[Boolean]) // class java.lang.Boolean instead of boolean
println(classOf[Byte]) // class java.lang.Byte instead of byte
println(classOf[Short]) // class java.lang.Short instead of short
println(classOf[Char]) // class java.lang.Character instead of char
println(classOf[Int]) // class java.lang.Integer instead of int
println(classOf[Long]) // class java.lang.Long instead of long
println(classOf[Float]) // class java.lang.Float instead of float
println(classOf[Double]) // class java.lang.Double instead of double
```

smarter: So what you're saying is that `semiEraseVCs = false` should be the default, I guess we could do that yeah, it just involves adding `semiEraseVCs = true` in a lot of places in `Erasure` and `TypeErasure`, because most of the time when you erase a type you want to semi-erase it.

smarter: @alexander-myltsev : I didn't say that you should remove the special cases for primitive classes, but the special case for value classes: `if (ValueClasses.isDerivedValueClass(claz)) { ...}`

DarkDimius: @smarter I am not speaking about erasure. I am speaking about this phase. This phase should for a `classOf[Meter]` generate a tree that is either `Literal(Constant(ClassInfo(Meter))` or `Literal(Constant(TypeRef(Meter))`, but not `Literal(Constant(ErasedValueType(Meter))`. The last one is not expected to reach backend and is not guaranteed to work.

smarter: @DarkDimius: I agree. But this phase uses `TypeErasure.erasure` where the default is `semiEraseVCs = true`, and what I was saying is that I agree that it makes sense for the default to be `semiEraseVCs = false` so that any phase can use `TypeErasure.erasure` without worrying about semi-erasure.

alexander-myltsev:
@smarter thanks for meaningful remarks 

All should be fixed now

smarter:
@DarkDimius : The backend contains code to deal with `classOf`: https://github.com/lampepfl/dotty/blob/d2c96d02fccef3a82b88ee1ff31253b6ef17f900/src/dotty/tools/backend/jvm/DottyBackendInterface.scala#L247-L248 Is it dead code?

DarkDimius:
For dotty it is, for scalac it isn't.

smarter:
But this code is in `DottyBackendInterface`, can this interface be used by scalac too?

DarkDimius:
That's the code that handles annotations. Annotations do contain trees, but they are special in many ways and their handling is very ad-hoc.

PR #678, opened 2015-06-22
Avoid junk produced by Constraint#replace.

### LINKED ISSUE #670: Orphan PolyParam when typing overloaded array apply
dotc ../pending/pos/t8230a.scala 
error while pickling type T
error when pickling type T'
error when pickling tree Arr.apply[T']
error when pickling tree Arr.apply[T']("2")
error when pickling tree I.arrToTrav[String'](Arr.apply[T']("2"))
error when pickling tree Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
error when pickling tree () extends Object() { this: Test.Okay.type => 
  Arr.apply[String']("1")
  import I.{arrToTrav, longArrToTrav}
  Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
}
error when pickling tree final module class Okay$() extends Object() { this: Test.Okay.type => 
  Arr.apply[String']("1")
  import I.{arrToTrav, longArrToTrav}
  Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
}
error when pickling tree () extends Object() { this: Test.type => 
  def foo(t: scala.collection.Traversable[Any]): Unit = {
    ()
  }
  final lazy module val Okay: Test.Okay$ = new Test.Okay$()
  final module class Okay$() extends Object() { this: Test.Okay.type => 
    Arr.apply[String']("1")
    import I.{arrToTrav, longArrToTrav}
    Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
  }
  final lazy module val Fail: Test.Fail$ = new Test.Fail$()
  final module class Fail$() extends Object() { this: Test.Fail.type => 
    import I.arrToTrav
    Test.foo(I.arrToTrav[String'](Arr.apply[T']("3")))
  }
}
error when pickling tree final module class Test$() extends Object() { this: Test.type => 
  def foo(t: scala.collection.Traversable[Any]): Unit = {
    ()
  }
  final lazy module val Okay: Test.Okay$ = new Test.Okay$()
  final module class Okay$() extends Object() { this: Test.Okay.type => 
    Arr.apply[String']("1")
    import I.{arrToTrav, longArrToTrav}
    Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
  }
  final lazy module val Fail: Test.Fail$ = new Test.Fail$()
  final module class Fail$() extends Object() { this: Test.Fail.type => 
    import I.arrToTrav
    Test.foo(I.arrToTrav[String'](Arr.apply[T']("3")))
  }
}
error when pickling tree package <empty> {
    def foo(t: scala.collection.Traversable[Any]): Unit = {
      ()
    }
    final lazy module val Okay: Test.Okay$ = new Test.Okay$()
    final module class Okay$() extends Object() { this: Test.Okay.type => 
      Arr.apply[String']("1")
      import I.{arrToTrav, longArrToTrav}
      Test.foo(I.arrToTrav[String'](Arr.apply[T']("2")))
    }
    final lazy module val Fail: Test.Fail$ = new Test.Fail$()
    final module class Fail$() extends Object() { this: Test.Fail.type => 
      import I.arrToTrav
      Test.foo(I.arrToTrav[String'](Arr.apply[T']("3")))
    }
  }
}
exception occurred while compiling ../pending/pos/t8230a.scala
Exception in thread "main" java.lang.AssertionError: assertion failed: orphan poly parameter: PolyParam(T)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleNewType$1(TreePickler.scala:253)
	at dotty.tools.dotc.core.tasty.TreePickler.dotty$tools$dotc$core$tasty$TreePickler$$pickleType$1(TreePickler.scala:132)
	at dotty.tools.dotc.core.tasty.TreePickler.dotty$tools$dotc$core$tasty$TreePickler$$pickleTpt$1(TreePickler.scala:288)

odersky:
There were two instances where a constraint undergoing a replace would still refer
to poly params that are no longer bound after the replace.

1. In an ordering the replaced parameter was not removed from the bounds of the others.
2. When a parameter refers to the replaced parameter in a type, (not a TypeBounds), the
replaced parameter was not replaced.

We now have checking in place that in globally committable typer states, TypeVars are not instantiated
to PolyParams and (configurable) that constraints of such typer states are always closed.

Fixes #670. Review by @retronym

### DIFF
--- a/src/dotty/tools/dotc/core/OrderingConstraint.scala
+++ b/src/dotty/tools/dotc/core/OrderingConstraint.scala
@@ -445,7 +445,7 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
         val bndss1 = bndss.map(_.filterConserve(_.binder ne pt))
         if (bndss.corresponds(bndss1)(_ eq _)) bndss else bndss1
       }
-      po.remove(pt).mapValues(removeFromBoundss)
+      po.remove(pt).mapValuesNow(removeFromBoundss)
     }
     newConstraint(boundsMap.remove(pt), removeFromOrdering(lowerMap), removeFromOrdering(upperMap))
   }
--- a/src/dotty/tools/dotc/core/OrderingConstraint.scala
+++ b/src/dotty/tools/dotc/core/OrderingConstraint.scala
@@ -442,8 +442,8 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
   def remove(pt: PolyType)(implicit ctx: Context): This = {
     def removeFromOrdering(po: ParamOrdering) = {
       def removeFromBoundss(key: PolyType, bndss: Array[List[PolyParam]]): Array[List[PolyParam]] = {
-        val bndss1 = bndss.map(bnds => bnds.filterConserve(_.binder ne pt))
-        if ((0 until bndss.length).forall(i => bndss(i) eq bndss1(i))) bndss else bndss1
+        val bndss1 = bndss.map(_.filterConserve(_.binder ne pt))
+        if (bndss.corresponds(bndss1)(_ eq _)) bndss else bndss1
       }
       po.remove(pt).mapValues(removeFromBoundss)
     }
--- a/src/dotty/tools/dotc/core/OrderingConstraint.scala
+++ b/src/dotty/tools/dotc/core/OrderingConstraint.scala
@@ -399,7 +399,7 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
       def removeParam(ps: List[PolyParam]) =
         ps.filterNot(p => p.binder.eq(poly) && p.paramNum == idx)
 
-      def replaceParam(tp: Type, atPoly: PolyType, atIdx: Int) = tp match {
+      def replaceParam(tp: Type, atPoly: PolyType, atIdx: Int): Type = tp match {
         case bounds @ TypeBounds(lo, hi) =>
 
           def recombine(andor: AndOrType, op: (Type, Boolean) => Type, isUpper: Boolean): Type = {
@@ -424,7 +424,8 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
           }
 
           bounds.derivedTypeBounds(replaceIn(lo, isUpper = false), replaceIn(hi, isUpper = true))
-        case _ => tp
+        case _ =>
+          tp.substParam(param, replacement)
       }
 
       var current =
@@ -438,8 +439,16 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
     }
   }
 
-  def remove(pt: PolyType)(implicit ctx: Context): This =
-    newConstraint(boundsMap.remove(pt), lowerMap.remove(pt), upperMap.remove(pt))
+  def remove(pt: PolyType)(implicit ctx: Context): This = {
+    def removeFromOrdering(po: ParamOrdering) = {
+      def removeFromBoundss(key: PolyType, bndss: Array[List[PolyParam]]): Array[List[PolyParam]] = {
+        val bndss1 = bndss.map(bnds => bnds.filterConserve(_.binder ne pt))
+        if ((0 until bndss.length).forall(i => bndss(i) eq bndss1(i))) bndss else bndss1
+      }
+      po.remove(pt).mapValues(removeFromBoundss)
+    }
+    newConstraint(boundsMap.remove(pt), removeFromOrdering(lowerMap), removeFromOrdering(upperMap))
+  }
 
   def isRemovable(pt: PolyType, removedParam: Int = -1): Boolean = {
     val entries = boundsMap(pt)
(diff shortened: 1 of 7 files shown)


src/dotty/tools/dotc/core/OrderingConstraint.scala
```diff
@@ -438,8 +439,16 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
     }
   }
 
-  def remove(pt: PolyType)(implicit ctx: Context): This =
-    newConstraint(boundsMap.remove(pt), lowerMap.remove(pt), upperMap.remove(pt))
+  def remove(pt: PolyType)(implicit ctx: Context): This = {
+    def removeFromOrdering(po: ParamOrdering) = {
+      def removeFromBoundss(key: PolyType, bndss: Array[List[PolyParam]]): Array[List[PolyParam]] = {
+        val bndss1 = bndss.map(bnds => bnds.filterConserve(_.binder ne pt))
+        if ((0 until bndss.length).forall(i => bndss(i) eq bndss1(i))) bndss else bndss1
```
retronym: `bndss.indices.forall`

src/dotty/tools/dotc/core/OrderingConstraint.scala
```diff
@@ -438,8 +439,16 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
...
+      }
+      po.remove(pt).mapValues(removeFromBoundss)
```
retronym: I had to check the sources to find out, but `SimpleMap#mapValues` is eager, as opposed to `s.c.Map#mapValues`. That looks like what we want here.

I could imagine this subtle difference could get confusing when maintaining dotty. Might I suggest a different method name, like `mapValuesNow`? This might be the first call to this method, so a slightly longer name won't cause widespread ugliness.

src/dotty/tools/dotc/core/OrderingConstraint.scala
retronym: Maybe:

```scala
var removed = false
val bndss1 = bndss.map(bnds => bnds.filterConserve(x => (x.binder ne pt) || { removed = true; false }))
if (removed) bndss1 else bndss
```

src/dotty/tools/dotc/core/OrderingConstraint.scala
```diff
@@ -491,6 +500,19 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
...
+    def isFreePolyParam(tp: Type) = tp match {
+      case PolyParam(binder, _) => !contains(binder)
+      case _ => false
+    }
+    def checkClosedType(tp: Type, where: String) =
+      if (tp != null)
+        assert(!tp.existsPart(isFreePolyParam), i"unclosed constraint: $this refers to $tp in $where")
+    boundsMap.foreachBinding((_, tps) => tps.foreach(checkClosedType(_, "bounds")))
+    lowerMap.foreachBinding((_, paramss) => paramss.foreach(_.foreach(checkClosedType(_, "lower"))))
+    upperMap.foreachBinding((_, paramss) => paramss.foreach(_.foreach(checkClosedType(_, "upper"))))
```
retronym: We use `s.c.r.internal.util.Collections#mforeach` and friends to cleanup code dealing with nested lists.

tests/pos/t8230a.scala
```diff
@@ -15,12 +15,12 @@ object Test {
...
-  }
+   import I.{ arrToTrav, longArrToTrav }
+   val x = foo(Arr("2"))
+ }
 
   object Fail {
-    import I.arrToTrav
-    foo(Arr("3")) // found String, expected Long
+//    import I.arrToTrav
+//    foo(Arr("3")) // found String, expected Long
```
retronym: Does this require further fixes before it compiles? If so, leave a breadcrumb by splitting this part out into a file in `pending/pos`.

retronym:
The commit comment has a few lines longer than the git convention, makes it hard to read in the GitHub UI.

src/dotty/tools/dotc/core/OrderingConstraint.scala
odersky: I settled for:

        val bndss1 = bndss.map(_.filterConserve(_.binder ne pt))
        if (bndss.corresponds(bndss1)(_ eq _)) bndss else bndss1

src/dotty/tools/dotc/core/OrderingConstraint.scala
odersky: That's a good point. In fact, I believe mapValues is the misnomer here. It should be `mapValuesView` or something like it. But gsince we can't change it now, renaming the SimpleMap op seems like a good idea.

src/dotty/tools/dotc/core/OrderingConstraint.scala
odersky: Does that also work for mixed Array/List code like here?

tests/pos/t8230a.scala
odersky: No in fact, that code compiles. It was left commented out as an oversight. Thanks for noticing!

src/dotty/tools/dotc/core/OrderingConstraint.scala
retronym: Nope, it is specialized to `List[List[_]]`.

PR #680, opened 2015-06-22
TypeErasure#erasure: do not semi-erase types by default

smarter:
This way we can simplify `ClassOf`.
Review by @DarkDimius .


smarter:
Alternatively, we could make semi-erasure a Typer mode I think, but I'm not sure that this would make things clearer.

odersky:
I think it would be better to have a method semiErasure alongside erasure instead of spreading the boolean parameter everywhere.

odersky:
LGTM. I'll merge and then try to get rid of SemiEraseVCs altogether.

PR #691, opened 2015-06-24
Partest output redirection over context.reporter

vsalvis:
For concurrent partest, output redirection by redirecting sysout is not suited. In scalac, the output stream of the compiler could be set in the Global. For dotty, this is done through the context and required an additional argument when creating the initial context. However, the compiler also needs to print through `ctx.println` instead of regular `println`.


DarkDimius:
LGTM, thanks!

PR #693, opened 2015-06-25
Fix Patmat and FunctionalInterfaces, enable 61 tests that succeed.

DarkDimius:
Several fixes to pattern matcher and FunctionalInterfaces, as well as fixes to backend. @odersky please review.


odersky:
According to the gitter discussion, it would make sense to do a PR against 2.12 that drops the name sorting. When is the best time to do this?

DarkDimius:
> FATAL: Failed to fetch from https://github.com/lampepfl/dotty.git

/rebuild

DarkDimius:
Passes all the tests on my local machine. Should be ready for review. 
@odersky please have a look.

adriaanm:
/rebuild -- worker ran out of disk space

adriaanm:
/rebuild (we set the timeout to 0, which jenkins interpreted as 3... ugh!)

src/dotty/tools/dotc/core/Names.scala
```diff
@@ -347,4 +347,27 @@ object Names {
       StringBuilder.newBuilder.mapResult(s => from.fromChars(s.toCharArray, 0, s.length))
     def apply(): Builder[Char, Name] = termNameBuilder
   }
+
+  implicit val NameOrdering: Ordering[Name] = new Ordering[Name] {
+    def compare(x: Name, y: Name): Int = {
+      if (x.isTermName && y.isTypeName) 1
+      else if (x.isTypeName && y.isTermName) -1
+      else if (x.start == y.start && x.length == y.length) 0
```
odersky: I think you can just use 

    if (x eq y)

here because names are hash-consed.

src/dotty/tools/dotc/core/Names.scala
```diff
@@ -347,4 +347,27 @@ object Names {
...
+      else {
+        val until = Math.min(x.length, y.length)
```
odersky: x.length min y.length

is more idiomatic.

src/dotty/tools/dotc/core/Names.scala
```diff
@@ -347,4 +347,27 @@ object Names {
...
+        var i = 0
+
+        while (i < until && x(i) == y(i)) i = i + 1
+
+        if (i < until) {
+          if (x(i) < y(i)) -1
+          else /*(x(i) > y(i))*/ 1
+        } else {
+          if (x.length < y.length) 1
```
odersky: So shorter names are bigger than longer ones. Is there a reason for this? Why not use lexicographic ordering?

src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -503,6 +503,8 @@ object Symbols {
     def showKind(implicit ctx: Context): String = ctx.kindString(this)
     def showName(implicit ctx: Context): String = ctx.nameString(this)
     def showFullName(implicit ctx: Context): String = ctx.fullNameString(this)
+
+    override def hashCode(): PhaseId = id
```
odersky: `PhaseId` is the wrong type.

src/dotty/tools/dotc/core/Names.scala
DarkDimius:
```
if (x eq y)
```
will not be true for `TypeName('A')` and `TermName('A')`

src/dotty/tools/dotc/transform/FunctionalInterfaces.scala
```diff
@@ -63,7 +63,12 @@ class FunctionalInterfaces extends MiniPhaseTransform {
         val m = tree.meth.tpe.widen.asInstanceOf[MethodType]
 
         if (shouldSpecialize(m)) {
-          val interfaceName = (functionName ++ m.paramTypes.length.toString).specializedFor(m.resultType, m.paramTypes)
+          val functionSymbol = tree.tpe.widenDealias.classSymbol
+          val names = ctx.atPhase(ctx.erasurePhase.prev) {
```
odersky: Why `erasurePhase.prev`? Note that at `erasurePhase` we are still *before* erasure. At `erasurePhase.prev` we are one phase ahead (i.e. at the beginning of `resolveSuper`. Is that really where we want to be? If yes, we need at least a comment.

odersky:
Otherwise LGTM

src/dotty/tools/dotc/core/Names.scala
DarkDimius: I believe at some moment we had a meeting where we agreed to have a style guideline to minimize usage of infix notation. Didn't we?

src/dotty/tools/dotc/core/Names.scala
odersky: But you have already tested for that case

DarkDimius: Ahh, agreed than. It wasn't the case initially.

src/dotty/tools/dotc/core/Names.scala
odersky: The queston here was whether we want to write

     x.length.min(y.length)

or

    x.length min y.length

not go all the way back to `Math.min` :-)

I think `min` is a well-established operator, so infix is OK.

DarkDimius: Between all those tree variants I prefer syntax that `Math.min` gives, because it's more close to the notation that I am used used to in math classes.

 Additionally, both `x.length.min(y.length)` and `x.length min y.length`, make first argument look like a receiver, making it special. 

Though I'm fine to change if you insist.

VladUreche:
> SpecializeNames: Duplicate scalac behaviour, sort tparams

@DarkDimius, why do you want to sort tparam names? Miniboxing got away without doing it and it never caused any problems.

DarkDimius:
@VladUreche this is code that uses Scala2 specialized functions, that sort their type params:
https://github.com/scala/scala/blob/2.11.x/src/compiler/scala/tools/nsc/transform/SpecializeTypes.scala#L333

DarkDimius:
@odersky comments addressed.

PR #694, opened 2015-06-25
Fix/dependent methods

odersky:
Some fixes related to implicit resolution which make some more tests go through. In particular, typelevel Peano numbers. Review by @adriaanm maybe?


odersky:
Other depmeth tests in pos/pending still give errors. Needs more investigation.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1279,10 +1279,37 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+     *  prototype is, say, Succ[Succ[Zero]], we can infer that it's argument type is Succ[Zero].
+     *  But if the prototype is N? >: Succ[Succ[Zero]] <: Succ[Succ[Zero]], the same
+     *  decomposition does not work - we'd get a N?#M where M is the element type name of Succ
+     *  instead.
+     */
+    def followAlias(tp: Type)(implicit ctx: Context): Type = {
+      val constraint = ctx.typerState.constraint
+      def inst(tp: Type): Type = tp match {
+        case TypeBounds(lo, hi) =>
+          if ((lo eq hi) || (hi <:< lo)(ctx.fresh.setExploreTyperState)) inst(lo) else NoType
```
smarter: Could you add a comment explaining why `NoType` instead of `tp` makes sense here?

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1279,10 +1279,37 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+        case tp: PolyParam =>
+          var tvar1 = constraint.typeVarOfParam(tp)
```
smarter: Should be a `val`.

odersky:
@smarter 

    mt.paramTypes.map(_ => WildcardType)

is even shorter (and I think clearer than either solution). I wrote the const to save a closure.

retronym:
Sadly, Function.const(x) still needs eta expansion and a new closure as it is a method with two param lists, rather than a method with one param list that returns a Function.

odersky:
@retronym Ah, that got probably done when we introduced specialization. So, no need to use Function.const.

Issue #687, opened 2015-06-24
f"" string interpolator does not work because it is implemented by a scala.reflect macro


DarkDimius:
```
[info] % dotc t8266-octal-interp.scala
[info] % java \
[info]   -classpath \
[info]   /Users/dark/workspace/dotty/tests/partest-generated/run/t8266-octal-interp-run.obj:/Users/dark/workspace/dotty/target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar:/Users/dark/.ivy2/cache/me.d-d/scala-compiler/jars/scala-compiler-2.11.5-20150619-173733-3bcd390afa.jar:/Users/dark/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.11.5.jar:/Users/dark/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.5.jar \
[info]   -Dfile.encoding=UTF-8 \
[info]   -Djava.library.path=/Users/dark/workspace/dotty/tests/partest-generated/run \
[info]   -Dpartest.output=/Users/dark/workspace/dotty/tests/partest-generated/run/t8266-octal-interp-run.obj \
[info]   -Dpartest.lib=/Users/dark/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.5.jar \
[info]   -Dpartest.reflect=/Users/dark/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.11.5.jar \
[info]   -Dpartest.comp=/Users/dark/workspace/dotty/target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar \
[info]   -Dpartest.cwd=/Users/dark/workspace/dotty/tests/partest-generated/run \
[info]   -Dpartest.test-path=/Users/dark/workspace/dotty/tests/partest-generated/run/t8266-octal-interp.scala \
[info]   -Dpartest.testname=t8266-octal-interp \
[info]   -Djavacmd=java \
[info]   -Djavaccmd=javac \
[info]   -Duser.language=en \
[info]   -Duser.country=US \
[info]   scala.tools.nsc.MainGenericRunner \
[info]   -usejavacp \
[info]   Test \
[info]   jvm > t8266-octal-interp-run.log
[info] java.lang.NoSuchMethodError: scala.StringContext.f(Lscala/collection/Seq;)Ljava/lang/String;
[info]  at X.f(t8266-octal-interp.scala:9)
[...]

PR #698, opened 2015-06-26
Introduce dotty error to see if tested

odersky:
Just want to make usre GenBCode is actually compiled in the tests, so I
introduce syntax error for dotty but not for scalac.

This is a test only. No review necessary.


odersky:
Good to see this is actually compiled. So the Cyclic reference errors I was seeing locally in #687 are probably due to compilation order.

Issue #685, opened 2015-06-24
t920.scala: java.lang.NoClassDefFoundError:


DarkDimius:
```
[info] % dotc t920.scala
[info] % java \
[info]   -classpath \
[info]   /Users/dark/workspace/dotty/tests/partest-generated/run/t920-run.obj:/Users/dark/workspace/dotty/target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar:/Users/dark/.ivy2/cache/me.d-d/scala-compiler/jars/scala-compiler-2.11.5-20150619-173733-3bcd390afa.jar:/Users/dark/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.11.5.jar:/Users/dark/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.5.jar \
[info]   -Dfile.encoding=UTF-8 \
[info]   -Djava.library.path=/Users/dark/workspace/dotty/tests/partest-generated/run \
[info]   -Dpartest.output=/Users/dark/workspace/dotty/tests/partest-generated/run/t920-run.obj \
[info]   -Dpartest.lib=/Users/dark/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.5.jar \
[info]   -Dpartest.reflect=/Users/dark/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.11.5.jar \
[info]   -Dpartest.comp=/Users/dark/workspace/dotty/target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar \
[info]   -Dpartest.cwd=/Users/dark/workspace/dotty/tests/partest-generated/run \
[info]   -Dpartest.test-path=/Users/dark/workspace/dotty/tests/partest-generated/run/t920.scala \
[info]   -Dpartest.testname=t920 \
[info]   -Djavacmd=java \
[info]   -Djavaccmd=javac \
[info]   -Duser.language=en \
[info]   -Duser.country=US \
[info]   scala.tools.nsc.MainGenericRunner \
[info]   -usejavacp \
[info]   Test \
[info]   jvm > t920-run.log
[info] java.lang.NoClassDefFoundError: Test$B (wrong name: Test$b)
[info] 	at java.lang.ClassLoader.defineClass1(Native Method)
[info] 	at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
[...]

PR #697, opened 2015-06-26
Fix erasure of this types and refactor value class erasure

odersky:
Thistypes erased to the underlying class. This is wrong. When seen as part of some other type,
a ThisType has to erase to the erasure of the underlying type (i.e. the erasure if the selftype
of the class). unittest-collections.scala failed with a MethodNotFound error because the erasure
was computed incorrectly.

On the other hand, a tree with a ThisType type, should keep the type, analogous to a
tree with a TermRef type.

Also, refactored value class erasure to avoid spreading semiEraseVCs.

Fixes #682 

Rview by @DarkDimius @smarter

PR #699, opened 2015-06-26
Fix trait constructors

odersky:
Trait constructors added a this to the constructor of a trait, which is useless
because the constructor's type is unit. By contrast, it's good to rename the DefDef
to the new name.

Discovered while trying to isolate #685. This causes a -Ycheck:traitConstructor error for all traits.
Review by @DarkDimius


smarter:
> This causes a -Ycheck:traitConstructor error for all traits.

Can we add `traitConstructors` to `defaultOptions` in `tests.scala` or would that break something ?

odersky:
Note: The error we are seeing here is the same I have been chasing for a while and one that did not show up in the tests for several releases (including commits to main) before.But they did show up in local runs.  It is fixed in #697. 

So the right order should be to get #697 in and then try again.

odersky:
Rebased on top of pending #697

odersky:
> Can we add traitConstructors to defaultOptions in tests.scala or would that break something ?

@smarter Even without traitConstructors, we get "bad owners" in several tests. Not sure why. Would be good to investigate.

PR #706, opened 2015-07-01
Generate less code for lazy vals

DarkDimius:
Makes modules which are entirely synthetic be eager.
This fix additionally makes inner objects of other objects remain static classes, instead of becoming lazy vals.

### DIFF
--- a/src/dotty/tools/dotc/transform/LazyVals.scala
+++ b/src/dotty/tools/dotc/transform/LazyVals.scala
@@ -46,18 +46,38 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer {
     * before this phase starts processing same tree */
   override def runsAfter = Set(classOf[Mixin])
 
-    override def transformDefDef(tree: DefDef)(implicit ctx: Context, info: TransformerInfo): Tree = {
-      if (!(tree.symbol is Flags.Lazy) || tree.symbol.owner.is(Flags.Trait)) tree
-      else {
-          val isField = tree.symbol.owner.isClass
+  override def transformDefDef(tree: tpd.DefDef)(implicit ctx: Context, info: TransformerInfo): tpd.Tree =
+   transformLazyVal(tree)
 
-          if (isField) {
-            if (tree.symbol.isVolatile || tree.symbol.is(Flags.Module)) transformMemberDefVolatile(tree)
-            else transformMemberDefNonVolatile(tree)
-          }
-          else transformLocalDef(tree)
+
+  override def transformValDef(tree: tpd.ValDef)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    transformLazyVal(tree)
+  }
+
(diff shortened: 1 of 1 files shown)


DarkDimius:
@odersky please review.

src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -46,18 +46,38 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer {
...
+  def transformLazyVal(tree: ValOrDefDef)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    val sym = tree.symbol
+    if (!(sym is Flags.Lazy) || sym.owner.is(Flags.Trait) || (sym.isStatic && sym.is(Flags.Module))) tree
+    else {
+
+      val isField = sym.owner.isClass
+
+      if (isField) {
+        if (sym.isVolatile ||
+          (sym.is(Flags.Module) && !sym.is(Flags.Synthetic))) // companion class is synthesized. Should be threadsafe to
```
odersky: Did not understand comment.

src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -46,18 +46,38 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer {
...
+        // make inner lazy vals thread safe
+          transformMemberDefVolatile(tree)
+        else if (sym.is(Flags.Module)) { // synthetic module
+          val holderSymbol = ctx.newSymbol(sym.owner, sym.asTerm.name ++ nme.LAZY_LOCAL,
```
odersky: I think this would flow better if this block was refactored into a method `transformSyntheticCompanion`.

DarkDimius: 1. Agreed;
2. It this code makes an eager val, not a non-threadsafe lazy val;
3. 
 >I thought we would make plain objects non-volatile, demand @volatile to be given explicitly?

 I assumed we did not reach agreement on this, the last argument was that this makes top level objects behave differently from inner ones, as outer ones are `@volatile` by default, but inner ones are not.

4. Are we sure that we'll never have a phase or a plugin that wants to introduce members to companion objects after `lazyvals`? Though this is currently true, It could be violated in future. Additionally, I'd prefer to have general rule that applies to the whole pipeline that every class has a companion.

Given those arguments, I believe it is better to drop empty objects late in the pipeline, maybe as late as GenBcode.

odersky: 2. OK understood. A comment would be good.
3. OK, let's keep them sharable between threads, by assuming @volatile implicitly
4. I'm OK with either strategy. It's just that later it might be harder to detect structures generated
    for empty objects. But with the current simplifications, maybe that's a non-issue.

DarkDimius:
@odersky review comments should be addressed now.

odersky:
LGTM

PR #708, opened 2015-07-01
Check that dotty is reentrant

odersky:
Not quite done yet. But the basic checking is in place and a lot of false positives are eliminated. One true possible datarace involving Symbols.nextId was eliminated. partest afterwards runs OK in parallel on my machine.


DarkDimius:
Succeeds running 8 threads on my machine in 4 min 15 sec, sequential time was 7 min 20 sec.

src/dotty/tools/dotc/transform/CtxLazy.scala
```diff
@@ -0,0 +1,20 @@
...
+ *  This should be used whenever the evaluation of a lazy expression
+ *  depends on some context, but the value can be re-used afterwards
+ *  with a different context.
+ *
+ *  A typical use case is a lazy val in a phase object which exists once per root context where
+ *  the expression intiializing the lazy val depends only on the root context, but not any changes afterwards.
+ */
+class CtxLazy[T](expr: Context => T) {
+  private var myValue: T = _
+  private var forced = false
```
smarter: `forced` is never set to true.

odersky: Oops.

DarkDimius:
@odersky, ready for review?

odersky:
Yes, please review.

On Fri, Jul 3, 2015 at 11:31 AM, Dmitry Petrashko <notifications@github.com>
wrote:

> @odersky <https://github.com/odersky>, ready for review?
>
> —
> <https://github.com/lampepfl/dotty/pull/708#issuecomment-118293287>.
>



-- 
Martin Odersky
EPFL

src/dotty/tools/dotc/core/Names.scala
```diff
@@ -200,44 +203,21 @@ object Names {
   private final val fillFactor = 0.7
 
   /** Memory to store all names sequentially. */
+  @sharable // because it's only mutated in synchronized block of termName
```
DarkDimius: Keeping track and maintaining such annotation could be a burden. 
Instead of annotating explicitly, could we infer that private fields that are modified in `synchronized(this)` blocks are sharable?

src/dotty/tools/dotc/transform/CheckReentrant.scala
```diff
@@ -0,0 +1,99 @@
...
+ *   - Handle arrays: Array elements are currently ignored because they are often used
+ *     in an immutable way anyway. To do better, it would be helpful to have a type
+ *     for immutable array.
+ */
+class CheckReentrant extends MiniPhaseTransform { thisTransformer =>
+  import ast.tpd._
+
+  override def phaseName = "checkReentrant"
+
+  def transformInfo(tp: Type, sym: Symbol)(implicit ctx: Context): Type = tp
```
DarkDimius: this class does not implement `InfoTransformer`, this code seems dead.

src/dotty/tools/dotc/Resident.scala
```diff
@@ -23,7 +23,7 @@ import scala.annotation.tailrec
  */
 object Resident extends Driver {
 
-  object residentCompiler extends Compiler
+  @unshared object residentCompiler extends Compiler
```
DarkDimius: Why is resident compiler an object?
Couldn't you have several resident compilers in the same process?

I'd better have resident compiler be a class, which is instantiated by some other object that has a `main` method.
This would also remove the need for `@unshared` annotation. I believe if something is indeed unshared - it could be a class.

DarkDimius:
Otherwise LGTM.

src/dotty/tools/dotc/core/Names.scala
odersky: Maybe at some point. But right now that's probably too much of a detour. Let's see first whether @sharable thing becomes a feature of general utility. Even now it serves a purpose: If we want to figure out data races, it's good to scrutinize members with @sharable annotations. For the others we can run -Ycheck:reentrant to rule them out (well, almost, modulo to the shortcomings mentioned in the comment).

src/dotty/tools/dotc/transform/CheckReentrant.scala
smarter: `ClassOf` does not implement `InfoTransformer` either and yet it seems to work, I don't know why: https://github.com/lampepfl/dotty/blob/6a36dd83e78fe2d6269626480577f8166ab7e37e/src/dotty/tools/dotc/transform/ClassOf.scala#L19

odersky:
Rebased to master.

Issue #705, opened 2015-07-01
Inner value class kill compiler


DarkDimius:
This code shouldn't be allowed at all, currently it fails y-check after `extmethods`. If y-check is disabled, it will instead fail in erasure.
```
class Foo {
 class B(val a: Int) extends AnyVal
}
```
Ycheck failure:
```
checking B.scala after phase TreeTransform:{refchecks, elimRepeated, normalizeFlags, extmethods, expandSAMs, tailrec, classOf}

Exception in thread "main" java.lang.AssertionError: assertion failed: error while typing B$.this, method hashCode is not contained in object B$
	at scala.Predef$.assert(Predef.scala:165)
	at dotty.tools.dotc.transform.TreeChecker$Checker.typedThis(TreeChecker.scala:238)
	at dotty.tools.dotc.typer.Typer.typedUnnamed$1(Typer.scala:1038)
	at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:1077)
	at dotty.tools.dotc.typer.ReTyper.typedUnadapted(ReTyper.scala:91)
	at dotty.tools.dotc.transform.TreeChecker$Checker.typedUnadapted(TreeChecker.scala:182)
	at dotty.tools.dotc.typer.Typer$$anonfun$typed$2.apply(Typer.scala:1087)
	at dotty.tools.dotc.typer.Typer$$anonfun$typed$2.apply(Typer.scala:1085)
	at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:148)
	at dotty.tools.dotc.core.Contexts$Context.traceIndented(Contexts.scala:52)
	at dotty.tools.dotc.typer.Typer.typed(Typer.scala:1085)
	at dotty.tools.dotc.typer.ReTyper.typedSelect(ReTyper.scala:35)
	at dotty.tools.dotc.transform.TreeChecker$Checker.typedSelect(TreeChecker.scala:232)
	at dotty.tools.dotc.typer.Typer.typedNamed$1(Typer.scala:1017)
	at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:1075)
	at dotty.tools.dotc.typer.ReTyper.typedUnadapted(ReTyper.scala:91)
	at dotty.tools.dotc.transform.TreeChecker$Checker.typedUnadapted(TreeChecker.scala:182)
	at dotty.tools.dotc.typer.Typer$$anonfun$typed$2.apply(Typer.scala:1087)
[...]

PR #710, opened 2015-07-02
Value class related checks

odersky:
Most of the restrictions on value classes were not enforced. This PR addresses the majority of them. It also enforces restrictions on inheriting from final and sealed classes, which was also missing so far.

review by @smarter


odersky:
Fixes #705

smarter:
LGTM

PR #712, opened 2015-07-04
Tests that objects nested in value class methods work

smarter:
This confirms that SI-6359 does not apply to Dotty.

@DarkDimius : please review


DarkDimius:
LGTM

PR #714, opened 2015-07-04
Move the inlining of value class methods before Erasure

smarter:
@DarkDimius : please review


smarter:
This fails because dealing with 0 parameter lists and > 1 parameter lists is trickier than I thought, will fix tomorrow.

smarter:
OK to review now!

src/dotty/tools/dotc/Compiler.scala
```diff
@@ -51,7 +51,8 @@ class Compiler {
       List(new PatternMatcher,
            new ExplicitOuter,
            new Splitter),
-      List(new SeqLiterals,
+      List(new VCInlineMethods),
```
DarkDimius: Does it actually need a separate group?

src/dotty/tools/dotc/Compiler.scala
```diff
@@ -51,7 +51,8 @@ class Compiler {
...
+           List(new SeqLiterals,
```
DarkDimius: Identation.

src/dotty/tools/dotc/transform/VCInlineEqAndField.scala
```diff
@@ -7,20 +7,18 @@ import Contexts._, Trees._, StdNames._, Symbols._
 import DenotTransformers._, TreeTransforms._, Phases.Phase
 import ExtensionMethods._, TreeExtractors._, ValueClasses._
 
-/** This phase inlines calls to methods and fields of value classes.
+/** This phase inlines calls to == and to the field of a value class
```
DarkDimius: I would prefer to treat all methods equally, not special-casing `==`, that could also be specialized & etc.

src/dotty/tools/dotc/transform/VCInlineEqAndField.scala
```diff
@@ -7,20 +7,18 @@ import Contexts._, Trees._, StdNames._, Symbols._
...
+ *  we avoid unnecessary allocations:
  *     new V(u1) == new V(u2) => u1 == u2
  *    (new V(u)).underlying() => u
  */
-class VCInline extends MiniPhaseTransform with IdentityDenotTransformer {
+class VCInlineEqAndField extends MiniPhaseTransform with IdentityDenotTransformer {
   import tpd._
 
-  override def phaseName: String = "vcInline"
+  override def phaseName: String = "vcInlineEqAndField"
```
DarkDimius: if `==` is moved to a different phase, I'd rename this phase to `vcBoxUnbox`.

src/dotty/tools/dotc/transform/VCInlineMethods.scala
```diff
@@ -0,0 +1,117 @@
...
+          }
+        } else {
+          ref(extensionMeth)
+            .appliedToTypeTrees(mtArgs)
+            .appliedTo(qual)
+            .appliedToArgss(mArgss)
+        }
+    }
+
+  override def transformSelect(tree: Select)(implicit ctx: Context, info: TransformerInfo): Tree =
```
DarkDimius: handling of all 3 different nodes is the same, I'd extract it to a method to remove code duplication.

tests/pos/valueclasses/paramlists.scala
```diff
@@ -0,0 +1,25 @@
...
+object Test {
+  def test: Unit = {
+    val m1 = new Meter(1)
+    m1.zero
+    m1.zero2
+    m1.one
+    m1.one2
+    m1.one3(10)
+    m1.two(11)(12)
+    m1.two2(11)(12)
```
DarkDimius: I'd like to add several tests here:
```
import m1._
zero
zero2
....
```
You are pattern-matching over specific form of tree, we used to have plain `Apply(Ident(one), Nil)` before fixes to `prefixIsElidable`. I suppose we shouldn't get naked Idents back, but it would be good to have a test.

src/dotty/tools/dotc/Compiler.scala
smarter: No it doesn't, will fix

src/dotty/tools/dotc/transform/VCInlineEqAndField.scala
smarter: I was planning to have `vcInlineMethods`, `vcInlineEq`(new phase which will be in a later phase) and `vcInlineField`. I don't think that `vcBoxUnbox` is a very good name, this phase does not do any boxing or unboxing, that's what Erasure does.

src/dotty/tools/dotc/transform/VCInlineMethods.scala
smarter: You're right, I'll do that.

smarter:
Comments addressed.

src/dotty/tools/dotc/transform/VCInlineEqAndField.scala
DarkDimius: The comment still stands.
`==` is in no way special, and should be treated as other methods. It does not need any special treatment at all.
Currently, this PR will not allow to define a well-behaving
```scala
def ==[@specialized T](t: T): Boolean
```

>I don't think that vcBoxUnbox is a very good name, this phase does not do any boxing or unboxing, that's what Erasure does.

I do not agree with you. What this phase does is a rewriting
`new Meter(a).underlying` to `a`.
Note that `new Meter(a)` is `box` and `_.underlying` is `unbox`. So what this phase does is peephole-optimization of `box-unbox` pattern. It is not `field inlining`(`vcInlineField`), which is a separate notion defined in papers and has nothing to do with value classes.

smarter: >  == is in no way special, and should be treated as other methods. It does not need any special treatment at all.

`==` handling before this PR is special and the name of the phase reflects that, we may change how we deal with `==` but that will be in a separate PR. Note that the special handling of `==` is only for the `==` defined in `Any`, if you define your own `==` then it will behave like a normal method.

> So what this phase does is peephole-optimization of box-unbox pattern. It is not field inlining(vcInlineField), which is a separate notion defined in papers and has nothing to do with value classes.

OK, maybe field inlining has another meaning, but I'd still like it if the name of the phase described what the phase does, what about `VCInlineGetter` or `VCElideBoxUnbox`?

DarkDimius: `VCElideBoxUnbox` sounds good to me.
Otherwise LGTM.

DarkDimius:
LGTM

PR #716, opened 2015-07-07
Fix problems with higher-kinded implicits

### LINKED ISSUE #553: Implicit search fails for new Random.shuffle(List(1,2))
This compiles with scalac but not with dotty:
```scala
import scala.util.Random

object Test {
  def test = {
    val rand = new Random
    rand.shuffle(List(1,2)) // fails
  }
}
```
```scala
try/t6948b.scala:6: error: no implicit argument of type scala.collection.generic.CanBuildFrom[
  (CC?
     >: ([+HK$0] => scala.collection.immutable.List[HK$0])[HK$0 =+ (T? >: Int)]
       <: 
    ([HK$0] =>  <: scala.collection.TraversableOnce[HK$0])
  )[(T? >: Int)]#Apply
, (T? >: Int), 
  (CC?
     >: ([+HK$0] => scala.collection.immutable.List[HK$0])[HK$0 =+ (T? >: Int)]
       <: 
    ([HK$0] =>  <: scala.collection.TraversableOnce[HK$0])
  )[(T? >: Int)]#Apply
] found for parameter bf of method shuffle in class Random
    rand.shuffle(List(1,2))
                           ^
one error found
```
But explicitly doing an implicit search is enough to make it work:
```scala
import scala.util.Random
import scala.collection.generic._
import scala.collection.immutable._

object Test {
  implicit val z: CanBuildFrom[List[Int], Int, List[Int]] =
    implicitly[CanBuildFrom[List[Int], Int, List[Int]]] // succeeds

  def test = {
    val rand = new Random
    rand.shuffle(List(1,2)) // now succeeds
  }
}
```
```

### PR DESCRIPTION
**Pull request by odersky**
Closes #553 and #681. Review by @smarter

### CODE REVIEW

**Comment by odersky** [comment]
@smarter:
 
    ([T] => List[T])[Int] # Apply   <:   [T] => Traversable[T]

or, expaned

    Lambda$P { type Apply = List[$HK0] } { $HK0 = Int } # Apply 
    <:
    Lambda$P { type Apply = Traversable[$HK0] }

File: src/dotty/tools/dotc/core/Types.scala [hunk 1]
```diff
@@ -869,7 +869,13 @@ object Types {
         case SkolemType(tp) =>
           tp.lookupRefined(name)
         case pre: WildcardType =>
-          WildcardType
+          pre.optBounds match {
+            case TypeBounds(lo, hi) if name.isTypeName =>
+              // for term names, forming a Termref with just the name risks losing overloaded instance
+              WildcardType(TypeBounds(NamedType(lo, name), NamedType(hi, name)))
```
smarter: The comment on https://github.com/dotty-staging/dotty/blob/fix/shuffle/src/dotty/tools/dotc/core/Types.scala#L1423-L1424 is now wrong since the wildcard is now bounded, not unbounded. it'd be nice to update it and maybe also document the handling of wildcard in `lookupRefined` itself.

smarter: So, does this mean that we consider `(_ >: Foo <: Bar)#T` to be equivalent to `_ >: Foo#T <: Bar#T` ? If so, isn't this wrong in general? For example, if `T` is contravariant we may have `Foo <: Bar` but `Foo#T >: Bar#T` so the bounds `_ >: Foo#T <: Bar#T` would not be satisfiable.

smarter: I think that having a few test cases would be helpful.

odersky: @smarter Good point about the change in lookupRefined. I have backed out of it, and replaced it with a different solution.

Test cases: We have Random.shuffle, I think that's already a good one. If you have more, please add them.

smarter:
Otherwise LGTM,  though I wonder if it's possible that we may add something to `theMap.discarded` in `wildApprox`  which should not be part of the implicit scope (which is determined by `Types#namedPartsWith`).

smarter:
I also suggest adding the test case https://raw.githubusercontent.com/retronym/scala/e439651befb85e6c3b131db6bf346513ecedc783/test/pending/pos/t6948.scala which fails in Scala 2.11 (see https://issues.scala-lang.org/browse/SI-6948?focusedCommentId=61848&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-61848) but works fine in Dotty (yay!)

retronym:
@smarter I would suggest to add make that test case independent of the collections library before using it. It will be less fragile and easier to reason about in isolation.

Why does it work in dotty? Here's an approach to quickly and directly see how higher kinded inference works:

```
scala> def foo[M[_], A](m: M[A]): M[A] = ???
warning: there was one feature warning; re-run with -feature for details
foo: [M[_], A](m: M[A])M[A]

scala> def bar = foo(null: T[Int] with U[String])
bar: T[Int]

scala> def bar = foo(null: U[String] with T[Int])
bar: U[String]

scala> def bar = foo(null: U[String] with T[Int] {})
bar: U[String]

scala> type Alias = U[String]
defined type alias Alias

scala> def bar = foo(null: Alias)
bar: U[String]
```

The logic in play is here: https://github.com/scala/scala/blob/v2.11.6/src/reflect/scala/reflect/internal/Types.scala#L3067-L3087

odersky:
I verified that the "shuffle" test no longer fails. So this seems to be fixed by other means now.

smarter:
The shuffle test and many others were fixed by https://github.com/lampepfl/dotty/pull/799 .The bug that https://github.com/dotty-staging/dotty/commit/ec2d7d675cdec3bc4b9a1e276971430f00cd2800 fixes probably still exists but we don't see it anymore because #799 changed inference such  that we no longer end up with wildcard prefixes for implicit higher-kinded types. I could try to construct a test case for it, however I'm aware of several issues with implicit scope caching that I'd like to fix first since they might overlap with this.

odersky:
Should we leave the PR open then?


On Tue, Feb 16, 2016 at 4:43 PM, Guillaume Martres <notifications@github.com
> wrote:

> The shuffle test and many others were fixed by #799
> <https://github.com/lampepfl/dotty/pull/799> .The bug that dotty-staging@
> ec2d7d6
> <https://github.com/dotty-staging/dotty/commit/ec2d7d675cdec3bc4b9a1e276971430f00cd2800>
> fixes probably still exists but we don't see it anymore because #799
> <https://github.com/lampepfl/dotty/pull/799> changed inference such that
> we no longer end up with wildcard prefixes for implicit higher-kinded
> types. I could try to construct a test case for it, however I'm aware of
> several issues with im plicit scope caching that I'd like to fix first
> since they might overlap with this.
>
> —
> <https://github.com/lampepfl/dotty/pull/716#issuecomment-184737726>.
>



-- 
Martin Odersky
EPFL

smarter:
Sure.

smarter:
>  The bug that dotty-staging@ec2d7d6 fixes probably still exists but we don't see it anymore because #799 changed inference such that we no longer end up with wildcard prefixes for implicit higher-kinded types. I could try to construct a test case for it.

The only test cases I can come up are of this form:
```scala
class X {
  type Elem
}
object X {
  implicit def foo: X#Elem = ???
}

object Test {
  def get[T <: X](implicit ev: T#Elem) = ev
  get
}
```
Which does not compile anyway because of the restrictions on type projections that https://github.com/lampepfl/dotty/pull/1051 added, so we can probably close this for now, but keep this in mind if we ever loosen these restrictions.

PR #717, opened 2015-07-08
Add VCEqMethods to overload `==` and `!=` in value classes

smarter:
This replaces the peephole optimization in VCElideAllocations (which we only keep to be able to optimize equality for value classes compiled by Scala 2). This is useful because it means that `v1 == v2` and `v1.equals(v2)` will always agree without preventing the user from overloading `equals`.

Note that I'm not sure if we should get this in now, the semantics are rather tricky and they might lead to some unexpected behaviors, it might be better to decide how we want to deal with equality in Dotty first (for example, we generate calls to `==` in various places, some of them use `Any#==` and some of them do overload resolution on `==`, this should be spec'ed).

@DarkDimius : Please review


odersky:
I would agree to wait with this for now.

smarter:
Rebased.

DarkDimius:
Note that out of 277 added lines, 90 lines are tests and ~90 lines are documentation, so the actual code change here is pretty small.

DarkDimius:
LGTM 

@odersky, what's you current opinion on merging this? I believe this change simplifies the rules and additionally lifts a restriction that would help me in linker.

felixmulder:
Needs rebase

PR #720, opened 2015-07-10
Neg tests check files for // error markers

vsalvis:
Neg tests run by JUnit now check that for each compiler error, there's a // error marker on the corresponding line in the source file. For errors that don't have a source position, add // nopos-error to any of the files (for an example of this, see [instantiateAbstract](https://github.com/vsalvis/dotty/blob/6d311edadc076e6a9d2860d5571c2ff261445164/tests/neg/instantiateAbstract.scala)).

Also FYI I had one case where not all errors get displayed by the compiler, maybe if there are more than two errors for a given line? In this file there are three errors displayed, but four errors total, and the test gets all four:
```
[info] Test dotc.tests.neg_autoTupling started
./tests/pos/autoTuplingTest.scala:3: error: too many arguments for method apply: (x: (A? >: Int(1)))Some[(A? >: Int(1))]
  val x = Some(1, 2)                                  // error when running with -language:noAutoTupling
                  ^
./tests/pos/autoTuplingTest.scala:6: error: wrong number of argument patterns for Some; expected: (Nothing)
    case Some(a, b) => a + b                          // error // error when running with -language:noAutoTupling
             ^
./tests/pos/autoTuplingTest.scala:6: error: not found: b
    case Some(a, b) => a + b                          // error // error when running with -language:noAutoTupling
                           ^
four errors found

=== found: List(./tests/pos/autoTuplingTest.scala:3)
=== found: List(./tests/pos/autoTuplingTest.scala:6, ./tests/pos/autoTuplingTest.scala:6, ./tests/pos/autoTuplingTest.scala:6)
=== all errors:
class dotty.tools.dotc.reporting.Reporter$Error at ./tests/pos/autoTuplingTest.scala:6: type mismatch:
 found   : Nothing(a)
 required: ?{ + : ? }
 both method Double2double in object Predef$ and method unaugmentString in object Predef$ convert from Nothing(a) to ?{ + : FunProto(b):? }
class dotty.tools.dotc.reporting.Reporter$Error at ./tests/pos/autoTuplingTest.scala:6: not found: b
class dotty.tools.dotc.reporting.Reporter$Error at ./tests/pos/autoTuplingTest.scala:6: wrong number of argument patterns for Some; expected: (Nothing)
class dotty.tools.dotc.reporting.Reporter$Error at ./tests/pos/autoTuplingTest.scala:3: too many arguments for method apply: (x: (A? >: Int(1)))Some[(A? >: Int(1))]
```

Do we plan on running neg tests on partest? In that case I have to add the functionality there as well, for now it's just the JUnit tests, while partest only checks the xerrors as before.

### DIFF
--- a/src/dotty/tools/dotc/reporting/Reporter.scala
+++ b/src/dotty/tools/dotc/reporting/Reporter.scala
@@ -210,6 +210,8 @@ abstract class Reporter {
   var warningCount = 0
   def hasErrors = errorCount > 0
   def hasWarnings = warningCount > 0
+  private var errors: List[Error] = Nil
+  def allErrors = errors
 
   val unreportedWarnings = new mutable.HashMap[String, Int] {
     override def default(key: String) = 0
@@ -220,7 +222,9 @@ abstract class Reporter {
     d match {
       case d: ConditionalWarning if !d.enablingOption.value => unreportedWarnings(d.enablingOption.name) += 1
       case d: Warning => warningCount += 1
-      case d: Error => errorCount += 1
+      case d: Error =>
+        errors = d :: errors
+        errorCount += 1
       case d: Info => // nothing to do here
       // match error if d is something else
     }
--- a/test/test/CompilerTest.scala
+++ b/test/test/CompilerTest.scala
@@ -190,71 +190,128 @@ abstract class CompilerTest extends DottyTest {
     val nerrors = reporter.errorCount
     assert(nerrors == xerrors, s"Wrong # of errors. Expected: $xerrors, found: $nerrors")
 
-    // is neg test, check errors occur on right line
+    // NEG TEST
     if (xerrors > 0) {
       val errorLines = reporter.allErrors.map(_.pos)
       // reporter didn't record as many errors as its errorCount says
       assert(errorLines.length == nerrors, s"Not enough errors recorded.")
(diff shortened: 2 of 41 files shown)


vsalvis:
Thank you very much for your comments. I restructured the code to start by reading the files, and then check the two error categories independently. Hopefully it's clearer now.

src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -210,6 +210,8 @@ abstract class Reporter {
   var warningCount = 0
   def hasErrors = errorCount > 0
   def hasWarnings = warningCount > 0
+  private var errors: List[Error] = Nil
+  def allErrors = errors
```
odersky: I am not super-happy with recording all errors here. That should be the job of individual reporters, not of the base class `Reporter`. I believe it would be better to install a particular reporter that stores the errors in the test framework.

DarkDimius:
@vsalvis, do you want to give it a shot and try to make a reporter that records errors? Otherwise I'd merge this to make new tests use this machinery.

vsalvis:
I'm in a dance camp, so I don't think it's going to happen this week. But I
can do it next week, or you can merge already and I'll add it by August 16
when I'm back from Serbia/Croatia.

VladimirNik:
Rebased version: https://github.com/lampepfl/dotty/pull/1106

PR #721, opened 2015-07-10
Don't add lower bounds of abstract types to the implicit scope

smarter:
As the [spec](http://www.scala-lang.org/files/archive/spec/2.11/07-implicits.html#implicit-parameters) says:
>The parts of a type T are [...] if T is an abstract type, the parts of
its upper bound

Review by @odersky .

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -3118,6 +3118,8 @@ object Types {
             apply(foldOver(maybeAdd(x, tp), tp), tp.underlying)
           case tp: TypeRef =>
             foldOver(maybeAdd(x, tp), tp)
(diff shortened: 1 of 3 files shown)


odersky:
It times out after three minutes... Could you build and test locally?

smarter:
All tests pass locally.

smarter:
@adriaanm : any idea why the builds keep getting aborted on Jenkins?

odersky:
LGTM

adriaanm:
We set the timeout to 0 on Jenkins to disable time outs. It turns out 0 == 3 modulo Jenkins-logic... The other issues you may have seen was due to running out of disk space, because we are now running the community build on the new CI. This should also be fixed.

PR #722, opened 2015-07-13
SyntheticMethods: override productPrefix for case classes

smarter:
The productPrefix of a case class should be the name of the class itself
to match Scala 2.

Review by @DarkDimius, hopefully this should fix various pending tests which print case classes.


DarkDimius:
LGTM

PR #724, opened 2015-07-13
Enable Ycheck after labelDef. Fixes #701

### LINKED ISSUE #701: -Ycheck failures when checking after last phase
We get "bad owner" violations when checking after the last group with -Ycheck:labelDefs.

DarkDimius:
LabelDefs reorders labels. As a result of reordering label-def defined inside other label-def could be lifted outside.

Those commits make TreeChecker aware of this.


DarkDimius:
@odersky please review

DarkDimius:
On my machine, time that it takes for the full junit testing suite to pass went from 4 min to 4 min 40 secs.
I would actually prefer to enable `-Ycheck:all`, at least in the Jenkins testing suite, but frontend currently does not pass Ycheck #725.

odersky:
I think 40 secs is sufficiently bad not to do it. Compromise proposal:
Let's test some files with -Ycheck:labeldefs and others with
-Ycheck:restoreScopes. Maybe pos with labelchecks and run with
restoreScopes (or vice versa)?


On Thu, Jul 16, 2015 at 10:50 AM, Dmitry Petrashko <notifications@github.com
> wrote:

> On my machine, time that it takes for the full junit testing suite to pass
> went from 4 min to 4 min 40 secs.
> I would actually prefer to enable -Ycheck:all, at least in the Jenkins
> testing suite, but frontend currently does not pass Ycheck #725
> <https://github.com/lampepfl/dotty/issues/725>.
>
> —
> <https://github.com/lampepfl/dotty/pull/724#issuecomment-121882278>.
>



-- 
Martin Odersky
EPFL

DarkDimius:
During the meeting we agreed to leave -Ycheck:labeldefs disabled in test suite that is ran locally, but enable `-Ycheck:all` in Jenkins CI.

odersky:
@darkdimius What needs to be done scriptwise to achieve this?

DarkDimius:
@odersky, scrips are updated. Ycheck:all is not enabled due to #725

PR #726, opened 2015-07-14
Fix #713. Put tasty attribute on mirror class of modules.

### LINKED ISSUE #713: The TASTY attribute for an object Foo should be stored in Foo.class, not Foo$.class
When we call `Foo.x` we may end up only loading the classfile `Foo.class` which contains a static forwarder `x`.

DarkDimius:


DarkDimius:
@smarter please review

smarter:
LGTM.

PR #728, opened 2015-07-14
Fix StackOverflow in Names.rehash

DarkDimius:
Testing suite of specialization is creating huge amount of names,
that triggered a StackOverflow in non-tail-recursive method.
Not rehash is tail-recursive.


DarkDimius:
@odersky please review

PR #729, opened 2015-07-15
WORK IN PROGRESS: New representation for arrays of value classes

smarter:
For a value class V whose underlying type is U, instead of representing
an array of V as V[] on the JVM, we use our own class VCXArray where X
is "U" if U is a primitive type and is "Object" otherwise.

This avoids boxing when creating arrays but we still need to box and
unbox when using such an array in a generic position, this also breaks
reflection and makes it pretty hard to use an array of a value class
from Java or Scala 2.

### DIFF
--- a/src/dotty/DottyPredef.scala
+++ b/src/dotty/DottyPredef.scala
@@ -1,7 +1,9 @@
 package dotty
 
+import scala.collection.mutable.WrappedArray
 import scala.reflect.ClassTag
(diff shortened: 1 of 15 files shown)


VladUreche:
Cool! Looking forward to this patch making it into dotty! Congrats @smarter and @DarkDimius!

smarter:
(Superceded by https://github.com/lampepfl/dotty/pull/1228)

PR #730, opened 2015-07-16
tpd#New: Make constructor TermRefs non-ambiguous

smarter:
Before this change, the prefix of a constructor TermRef was the prefix
of the class, this means that it was possible for two constructors to
have the same prefix, the same name (`<init>`) and the same signature.
This could cause double binding errors.

Review by @odersky


odersky:
Otherwise LGTM. Nice to see the simplification this entails.

smarter:
Comment addressed.

PR #731, opened 2015-07-22
Change HK types scheme

odersky:
I recently found out that the original scheme for hk types in dotty
that I proposed does not work. Here is a counter example:

    trait Iterable[IA, C <: Iterable]
    trait Seq[SA, C <: Seq] extends Iterable[A, C]
    val s: Seq[Int, Seq] = ???
    val i: Iterable[Int, Iterable]

Now, we said that the type C in Iterable has two type parameters, the
IA and the C. So if C <: Iterable, then

    C[T, D] would translate to C { type IA = D, type C = D }.

Analogously, if C <: Seq, then

    C[T, D] would translate to C { type SA = D, type C = D }

The problem is when we see a C defined to be <: Seq as a subtype of Iterable.
The two versions have the same number parameters, but the fields are different!
We'd be translating then C[T, D] to C { type IA = D, type C = D } which makes no
sense, it would just give us a double binding of IA. It turns out this
cannot be worked around in general; asSeenFrom interacts in fatal ways
with the scheme.

To fix this, we now always use the (formerly backup) scheme of type lambdas, paired
by a new notion of "eta-reduction" to keep types small. 

Review by @adriaanm

### DIFF
--- a/src/dotty/tools/dotc/core/Substituters.scala
+++ b/src/dotty/tools/dotc/core/Substituters.scala
@@ -277,4 +277,31 @@ trait Substituters { this: Context =>
   final class SubstParamsMap(from: BindingType, to: List[Type]) extends DeepTypeMap {
     def apply(tp: Type) = substParams(tp, from, to, this)
   }
+
+  /** A map for "cycle safe substitutions" which do not force the denotation
+   *  of a TypeRef unless the name matches up with one of the substituted symbols.
+   */
+  final class SafeSubstMap(from: List[Symbol], to: List[Type]) extends TypeMap {
+    def apply(tp: Type): Type = tp match {
+      case tp: NamedType =>
+        try {
+          var sym: Symbol = null
+          var fs = from
+          var ts = to
+          while (fs.nonEmpty) {
+            if (fs.head.name == tp.name) {
+              if (sym == null) sym = tp.symbol
+              if (fs.head eq sym) return ts.head
+            }
+            fs = fs.tail
+            ts = ts.tail
+          }
+          tp.newLikeThis(apply(tp.prefix))
+        }
+        catch {
+          case ex: CyclicReference => tp.derivedSelect(apply(tp.prefix))
+        }
+      case _ => mapOver(tp)
+    }
+  }
 }
--- a/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -44,7 +44,7 @@ class TypeApplications(val self: Type) extends AnyVal {
   /** The type parameters of this type are:
    *  For a ClassInfo type, the type parameters of its class.
    *  For a typeref referring to a class, the type parameters of the class.
-   *  For a typeref referring to an alias or abstract type, the type parameters of
+   *  For a typeref referring to a Lambda class, the type parameters of
    *    its right hand side or upper bound.
    *  For a refinement type, the type parameters of its parent, unless the refinement
    *  re-binds the type parameter with a type-alias.
@@ -58,7 +58,10 @@ class TypeApplications(val self: Type) extends AnyVal {
       case tp: TypeRef =>
         val tsym = tp.typeSymbol
         if (tsym.isClass) tsym.typeParams
-        else tp.underlying.typeParams
+        else {
+          val lam = LambdaClass(forcing = false)
+          if (lam.exists) lam.typeParams else Nil//tp.underlying.typeParams
+        }
       case tp: RefinedType =>
         val tparams = tp.parent.typeParams
         tp.refinedInfo match {
--- a/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -489,14 +489,22 @@ class TypeApplications(val self: Type) extends AnyVal {
    *  `this.Arg$i`.
    *
    *  TypeBounds are lambda abstracting by lambda abstracting their upper bound.
+   *
+   *  @param cycleParanoid   If `true` don't force denotation of a TypeRef unless
+   *                         its name matches one of `boundSyms`. Needed to avoid cycles
+   *                         involving F-boundes hk-types when reading Scala2 collection classes
+   *                         with new hk-scheme.
    */
-  def LambdaAbstract(boundSyms: List[Symbol])(implicit ctx: Context): Type = {
+  def LambdaAbstract(boundSyms: List[Symbol], cycleParanoid: Boolean = false)(implicit ctx: Context): Type = {
     def expand(tp: Type) = {
       val lambda = defn.lambdaTrait(boundSyms.map(_.variance))
       val substitutedRHS = (rt: RefinedType) => {
         val argRefs = boundSyms.indices.toList.map(i =>
           RefinedThis(rt).select(tpnme.lambdaArgName(i)))
-        tp.subst(boundSyms, argRefs).bounds.withVariance(1)
+        val substituted =
+          if (cycleParanoid) new ctx.SafeSubstMap(boundSyms, argRefs).apply(tp)
+          else tp.subst(boundSyms, argRefs)
+        substituted.bounds.withVariance(1)
       }
       val res = RefinedType(lambda.typeRef, tpnme.Apply, substitutedRHS)
       //println(i"lambda abstract $self wrt $boundSyms%, % --> $res")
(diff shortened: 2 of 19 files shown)


src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -410,91 +441,48 @@ class TypeApplications(val self: Type) extends AnyVal {
...
    *
-   *  where XYZ reflets that variances of the bound symbols and
-   *  `subst` is a substitution that replaces every bound symbol sym_i by
-   *  `this.Arg$i`.
+   *  where
+   *   - XYZ reflects the variances of the bound symbols,
+   *   - `bounds` consists of type declarations `type hk$i >: toHK(L) <: toHK(U),
+   *     one for each type parameter in `T` with non-trivial bounds L,U.
+   *   - `toHK` is a substitution that replaces every bound symbol sym_i by
+   *     `this.Arg$i`.
```
smarter: Should be `this.hk$i`.

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -410,91 +441,48 @@ class TypeApplications(val self: Type) extends AnyVal {
...
       }
-      val res = RefinedType(lambda.typeRef, tpnme.Apply, substitutedRHS)
-      //println(i"lambda abstract $self wrt $boundSyms%, % --> $res")
-      res
+      val lambdaWithBounds =
+        RefinedType.make(lambda.typeRef, boundNames.toList, boundss.toList.map(toHK))
+      RefinedType(lambdaWithBounds, tpnme.hkApply, toHK(tp))
     }
     self match {
       case self @ TypeBounds(lo, hi) =>
```
smarter: This is not a new change but I'm wondering why on the next line we do `expand(TypeBounds.upper(hi))` and not `expand(hi)`.

src/dotty/tools/dotc/core/Substituters.scala
```diff
@@ -277,4 +277,31 @@ trait Substituters { this: Context =>
...
+              if (sym == null) sym = tp.symbol
+              if (fs.head eq sym) return ts.head
+            }
+            fs = fs.tail
+            ts = ts.tail
+          }
+          tp.newLikeThis(apply(tp.prefix))
+        }
+        catch {
+          case ex: CyclicReference => tp.derivedSelect(apply(tp.prefix))
```
smarter: Could you add a comment explaining why we need to deal with `CyclicReference` exceptions this way?

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -44,7 +45,7 @@ class TypeApplications(val self: Type) extends AnyVal {
   /** The type parameters of this type are:
    *  For a ClassInfo type, the type parameters of its class.
    *  For a typeref referring to a class, the type parameters of the class.
-   *  For a typeref referring to an alias or abstract type, the type parameters of
+   *  For a typeref referring to a Lambda class, the type parameters of
```
smarter: Shouldn't this be `For a typeref referring to an alias type or a Lambda class, ...` ?

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -185,18 +202,31 @@ class TypeApplications(val self: Type) extends AnyVal {
...
     }
   }
 
   /** Simplify a fully instantiated type of the form `LambdaX{... type Apply = T } # Apply` to `T`.
    */
   def simplifyApply(implicit ctx: Context): Type = self match {
-    case self @ TypeRef(prefix, tpnme.Apply) if prefix.isInstantiatedLambda =>
-      prefix.member(tpnme.Apply).info match {
+    case self @ TypeRef(prefix, tpnme.hkApply) if prefix.isInstantiatedLambda =>
+      prefix.member(tpnme.hkApply).info match {
```
smarter: Why not `self.refinedInfo match {` ?

src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -466,21 +466,19 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
       false
   }
 
-  /** If `projection` is of the form T # Apply where `T` is an instance of a Lambda class,
-   *  and `other` is not a type lambda projection, then convert `other` to a type lambda `U`, and
+  /** If `projection` is a hk projection T#$apply
+   *  and `other` is not a hk projection, then convert `other` to a hk projection `U`, and
    *  continue with `T <:< U` if `inOrder` is true and `U <:< T` otherwise.
    */
   def compareHK(projection: NamedType, other: Type, inOrder: Boolean) =
```
smarter: Since you're changing `compareHK`, you may want to adapt and apply https://github.com/dotty-staging/dotty/commit/eac33f06cbc37327ded86dd51cae735f89f15519 too.

tests/pos/partialApplications.scala
```diff
@@ -1,13 +1,46 @@
...
+  val rs = Map[Int, Float]()
+
+  val ss: RMap[Float, Int] = rs
+
+}
+
+object Test2 {
+  type Histogram = Map[_, Int]
+
+  type StringlyHistogram = Histogram[_ >: String] // error
```
smarter: If these are errors, shouldn't they be in `tests/neg`, not `tests/pos`?

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -822,14 +822,54 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+          //
+          //    Lambda$I { type hk$0; type hk$1 = Int; type $apply = Map[$hk1, $hk0] } # $apply
+          //
+          // Let's call the last type T. You would expect that
+          //
+          //    Map[Int, String] <: RMap[_, Int]
+          //
+          // But that's not the case given the standard subtyping rules. In fact, the rhs reduces to
+          //
+          //    Map[Int, T # $hk0]
```
smarter: Since `RMap[_, Int]` is not fully applied I don't think it's reduced, instead it looks like we end up in `compareHK` where we eta-expand `Map[Int, String]` and then we check:
```scala
Lambda$II { type hk$0 = Int; type hk$1 = String; type $apply = Map[$hk0, $hk1] } <:<
Lambda$II { type hk$0; type hk$1 = Int; type $apply = Map[$hk1, $hk0] }
```
Which only fails because `hk$0` and `hk$1` are reversed in the two lambdas, it seems that if we simply changed the eta-expansion of `Map[Int, String]` in this particular case to be `Lambda$II { type hk$1 = Int; type hk$0 = String; type $apply = Map[$hk1, $hk0] }`, then the test would succeed, I don't know if this could be done in general in a satisfactory way (maybe with a method like `etaExpandAsSeenFrom` ?).

src/dotty/tools/dotc/core/TypeComparer.scala
odersky: No, that one should now be redundant.

tests/pos/partialApplications.scala
odersky: They are not errors anymore. Good catch.

src/dotty/tools/dotc/typer/Typer.scala
odersky: Tha's a little bit like what's done in the last commits of #802.

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: Yes, I fixed the comments in #802

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: Not sure about this one.

src/dotty/tools/dotc/core/Substituters.scala
odersky: I should have noted the reason before. There was an CyclicReference error raised but I no longer remember under what circumstances.

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: "referring" is taken to  go through aliases in the comment.

odersky:
Closed in favor of #802

src/dotty/tools/dotc/typer/Typer.scala
smarter: But wildcard arguments to higher-kinded types are still disallowed,  I think it would be possible to allow them with a better implementation of `compareHK`, but I'm not sure.

Issue #733, opened 2015-07-23
Implement getClass method


alexander-myltsev:
Note the case:

```scala
scala> ().getClass
res0: Class[Unit] = void
```

PR #734, opened 2015-07-23
Implement getClass

alexander-myltsev:
Closes https://github.com/lampepfl/dotty/issues/733

### DIFF
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -62,6 +62,7 @@ class Compiler {
            new AugmentScala2Traits,
            new ResolveSuper),
       List(new Erasure),
+      List(new GetClass), // getClass transformation should be applied to specialized methods
       List(new ElimErasedValueType,
            new VCElideAllocations,
            new Mixin,
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -62,7 +62,6 @@ class Compiler {
            new AugmentScala2Traits,
            new ResolveSuper),
       List(new Erasure),
-      List(new GetClass), // getClass transformation should be applied to specialized methods
       List(new ElimErasedValueType,
            new VCElideAllocations,
            new Mixin,
@@ -71,7 +70,8 @@ class Compiler {
            new LinkScala2ImplClasses,
            new CapturedVars, // capturedVars has a transformUnit: no phases should introduce local mutable vars here
            new Constructors,
-           new FunctionalInterfaces),
+           new FunctionalInterfaces,
+           new GetClass), // getClass transformation should be applied to specialized methods
       List(new LambdaLift,   // in this mini-phase block scopes are incorrect. No phases that rely on scopes should be here
            new ElimStaticThis,
            new Flatten,
--- a/src/dotty/tools/dotc/transform/GetClass.scala
+++ b/src/dotty/tools/dotc/transform/GetClass.scala
@@ -1,0 +1,49 @@
+package dotty.tools.dotc
+package transform
+
+import ast.tpd
+import core.Contexts.Context
+import core.StdNames.nme
+import core.Symbols.TermSymbol
+import core.Phases.Phase
+import TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/** Rewrite `getClass` calls as follow:
+  *
+  *  For every instance of primitive class C whose boxed class is called B:
+  *    instanceC.getClass    -> B.TYPE
+  *  For every instance of non-primitive class D:
+  *    instanceD.getClass    -> instanceD.getClass
+  */
+class GetClass extends MiniPhaseTransform {
+  import tpd._
+
+  override def phaseName: String = "getClass"
+
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Erasure])
+
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    import ast.Trees._
+
+    tree match {
+      case Apply(Select(qual, nme.getClass_), Nil) =>
+        val defn = ctx.definitions
+        val claz = qual.tpe.classSymbol
+
+        def TYPE(module: TermSymbol) = ref(module).select(nme.TYPE_).ensureConforms(tree.tpe)
+        claz match {
+          case defn.BooleanClass => TYPE(defn.BoxedBooleanModule)
+          case defn.ByteClass    => TYPE(defn.BoxedByteModule)
+          case defn.ShortClass   => TYPE(defn.BoxedShortModule)
+          case defn.CharClass    => TYPE(defn.BoxedCharModule)
+          case defn.IntClass     => TYPE(defn.BoxedIntModule)
+          case defn.LongClass    => TYPE(defn.BoxedLongModule)
+          case defn.FloatClass   => TYPE(defn.BoxedFloatModule)
+          case defn.DoubleClass  => TYPE(defn.BoxedDoubleModule)
+          case defn.UnitClass    => TYPE(defn.BoxedVoidModule)
+          case _                 => tree
+        }
+      case _ => tree
+    }
+  }
+}
--- a/tests/run/getclass.scala
+++ b/tests/run/getclass.scala
@@ -1,0 +1,41 @@
+class ValueClass(val i: Integer) extends AnyVal
+class SomeClass
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    val cls: Predef.Class[_] = new SomeClass().getClass
+    val valCls: Predef.Class[_] = new ValueClass(1).getClass
+    val iCls: Class[Int] = 1.getClass
+    val f1: Function2[Int, Int, Unit] = (a: Int, b: Int) => println(a + b)
+    val f2: Function1[Int, Boolean] = (a: Int) => a % 2 == 0
+
+    println("Value types:")
+    println(().getClass)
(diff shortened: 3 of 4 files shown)


alexander-myltsev:
@DarkDimius @smarter, please check if a test covers expected behaviour.

DarkDimius:
This PR does not currently have an implementation, only a test that fails due to non-implemented #733. Did you forget to push something?

alexander-myltsev:
It does not have yet. Is the test correct?

DarkDimius:
All names but name of the anonymous lambdas (`Test$$anonfun$1`) are correct.
For lambdas: I guess you would get something like `Test$$lambda$1/XXXXX`, where XXXX is a number of origin that I am unaware of.

tests/run/getclass.scala
```diff
@@ -0,0 +1,39 @@
...
+    println(true.getClass)
+    println(1.asInstanceOf[Byte].getClass)
+    println(1.asInstanceOf[Short].getClass)
+    println('a'.getClass)
+    println(1.getClass)
+    println(1L.getClass)
+    println(1f.getClass)
+    println(1d.getClass)
+
+    println("Class types")
```
DarkDimius: `:\n`

alexander-myltsev:
There are compile-time warnings: 
```
dotty/out$ ../bin/dotc ../tests/run/getclass.scala
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$GenMapFactory$$CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$MapFactory$$CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$ImmutableMapFactory$$CC; assuming they are invariant
three warnings found
```

And runtime errors as follows:
```
dotty/out$ java -cp .:/home/vagrant/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.5.jar:/home/vagrant/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.11.5.jar:/home/vagrant/dotty/target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar Test
Exception in thread "main" java.lang.BootstrapMethodError: java.lang.NoClassDefFoundError: scala/compat/java8/JProcedure2
        at Test$.main(getclass.scala:9)
        at Test.main(getclass.scala)
Caused by: java.lang.NoClassDefFoundError: scala/compat/java8/JProcedure2
        ... 2 more
Caused by: java.lang.ClassNotFoundException: scala.compat.java8.JProcedure2
        at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        ... 2 more
```

What should be fixed?

DarkDimius:
>There are compile-time warnings: 

Those are known warnings. Ignore them.

>NoClassDefFoundError: scala/compat/java8/JProcedure2

That is strange. That is a class defined in `dotty/src/scala/compat/java8/JFunction2.java`, and it should(and on my machine is) part of `dotty_2.11-0.1-SNAPSHOT.jar`

alexander-myltsev:
Fixed. My CP was wrong.

There is another problem. Seems like `Array(List("1"))` would have type of `Object` after erasure. So, `Array(List("1")).getClass` would be `class java.lang.Object` instead of `class [Lscala.collection.immutable.List;`.

DarkDimius:
> Seems like Array(List("1")) would have type of Object after erasure

AFAIK it should not. Though you should not handle arrays at all: classes that extend AnyRef have a runtime `.getClass` that should be used.

src/dotty/tools/dotc/transform/GetClass.scala
```diff
@@ -0,0 +1,45 @@
...
+import core.StdNames.nme
+import core.Symbols.TermSymbol
+import core.TypeErasure
+import TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/**
+ * FILL ME
+ */
+class GetClass extends MiniPhaseTransform {
+  import tpd.{Apply, Tree, Literal, ref, TreeOps}
```
smarter: Just `import tpd._`

src/dotty/tools/dotc/transform/GetClass.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  override def phaseName: String = "getClass"
+
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    import ast.Trees.{Apply, Select}
```
smarter: Just `import ast.Trees._` at the top of the file.

src/dotty/tools/dotc/transform/GetClass.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    tree match {
+      case Apply(Select(typ, nm), List()) if nm.toString == "getClass" =>
```
smarter: I would rename `typ` to `qual` and `nm` to `name`.
Also, instead of `nm.toString == "getClass"` you should do `nm == nme.getClass`

smarter:
There is code in `InterceptedMethods` to deal with `getClass`: https://github.com/lampepfl/dotty/blob/cf17b73cc02daf514f6d5499fd7a5562fc93349e/src/dotty/tools/dotc/transform/InterceptedMethods.scala#L116-L125
It'd be good to at least avoid duplication (for example by moving `primitiveGetClassMethods` inside `Definitions.scala` and using it in both `GetClass` and `InterceptedMethods`) but it might make sense to simply move this code from `InterceptedMethods` to `GetClass`, so that all the code that deals with `getClass` is inside `GetClass`. @DarkDimius : what do you think?

DarkDimius:
> but it might make sense to simply move this code from InterceptedMethods to GetClass.

This code is left as it tries to cover non-trivial bug. Though I've checked now and the way how this was fixed in scalac does not work in dotty: `5.asInstanceOf[Int with AnyRef].getClass` is still erased to `5.getClass`. 

I believe that we need a different fix. The way I would fix it in dotty is to rewrite 
```
5.asInstanceOf[Int with AnyRef].getClass
```

to 
```
(5.asInstanceOf[Int with AnyRef]: Object).getClass
```

This should do the trick.

src/dotty/tools/dotc/transform/GetClass.scala
```diff
@@ -0,0 +1,42 @@
...
+          case defn.BooleanClass => TYPE(defn.BoxedBooleanModule)
+          case defn.ByteClass    => TYPE(defn.BoxedByteModule)
+          case defn.ShortClass   => TYPE(defn.BoxedShortModule)
+          case defn.CharClass    => TYPE(defn.BoxedCharModule)
+          case defn.IntClass     => TYPE(defn.BoxedIntModule)
+          case defn.LongClass    => TYPE(defn.BoxedLongModule)
+          case defn.FloatClass   => TYPE(defn.BoxedFloatModule)
+          case defn.DoubleClass  => TYPE(defn.BoxedDoubleModule)
+          case defn.UnitClass    => TYPE(defn.BoxedVoidModule)
+          case _                 => qual.selectWithSig(defn.Any_getClass).appliedToNone
```
DarkDimius: why not simply `tree`?

src/dotty/tools/dotc/transform/GetClass.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  override def phaseName: String = "getClass"
+
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    import ast.Trees._
+
+    tree match {
+      case Apply(Select(qual, nme.getClass_), Nil) =>
+        val defn = ctx.definitions
+        val claz = qual.tpe.classSymbol
```
DarkDimius: some types don't have symbols before erasure. I'd add 
```scala
def runsAfter = Set(classOf[Erasure])
```

src/dotty/tools/dotc/transform/GetClass.scala
alexander-myltsev: Yes, that works.

src/dotty/tools/dotc/transform/GetClass.scala
alexander-myltsev: I thought that order in `Compiler` is enough for phases ordering. Why do we explicitly need `runsAfter`?

DarkDimius: >I thought that order in Compiler is enough for phases ordering

It is enough.

>Why do we explicitly need runsAfter? 

We have a checks, before building compiler pipeline that ensures that those `runsAfter` are followed. This is very handy if someone wants to play around phase ordering, as it makes dependencies explicit.

src/dotty/tools/dotc/Compiler.scala
```diff
@@ -62,6 +62,7 @@ class Compiler {
            new AugmentScala2Traits,
            new ResolveSuper),
       List(new Erasure),
+      List(new GetClass), // getClass transformation should be applied to specialized methods
```
DarkDimius: Did you try to put it in next mini-phase block?

DarkDimius:
Otherwise LGTM

alexander-myltsev:
@DarkDimius , well, worked out. Should I squash both commits?

DarkDimius:
LGTM

DarkDimius:
@alexander-myltsev, thanks for your contribution!

Issue #739, opened 2015-07-24
Type parameters should be instantiated before doing an implicit search


smarter:
This is a separate issue from #738 but I think that they will need to be fixed together to get something that makes sense.
Consider the following code:
```scala
class Foo

object Test {
  def foo[T](x: T)(implicit ev: T): T = ???

  def test: Unit = {
    implicit val evidence: Foo = new Foo
    foo(new Foo)
  }
}
```

It currently fails with:
```scala
try/constrimpl.scala:9: error: ambiguous implicits: both getter StringCanBuildFrom in object Predef$ and getter NothingClassTag in object DottyPredef$ match type (T? >: Foo) of parameter ev of method foo in object Test$
    foo(new Foo)
                ^
try/constrimpl.scala:10: error: ambiguous implicits: both getter StringCanBuildFrom in object Predef$ and getter NothingClassTag in object DottyPredef$ match type (T? >: Foo) of parameter ev of method foo in object Test$
  }
   ^
two errors found
```

When we do the implicit search, `T` is not instantiated yet, the only thing we know is that its lower bound is `Foo` as the error message says: `... match type (T? >: Foo) of parameter ev of method foo ...`. I don't know how `scalac` deals with this exactly but it seems like it simply instantiate the type parameter before doing the implicit search, we should consider doing the same thing.

Note that this is also needed to properly fix #553, if we don't instantiate `CC` to `List`, then `CC` is only lower-bounded by `List`, and [lower bounds of abstract types are not part of the implicit scope](https://github.com/lampepfl/dotty/commit/0efa171e8ccca0d49fc6d800fd21e29f7b7336fd), this means that the implicit `List.canBuildFrom` will not be in the scope.

odersky:
scalac does not generally instantiate parameters before an implicit search. If it did, the whole concept of CanBuildFrom would not work, to give an example. But parameter instantiation is different between the two systems, and it seems this example shows where dotty's way is problematic. Maybe instantiate when there are ambiguity errors otherwise?

retronym:
FYI, here's how to get a bit of logging from scalac about type param inference.

```
% cat sandbox/test.scala
class Foo[A, B]

object Test {
  def foo[T, U](x: T)(implicit ev: Foo[T, U]): U = ???

  def test: Unit = {
    implicit val evidence1: Foo[Int, String] = ???
    implicit val evidence2: Foo[String, Int] = ???
    foo(1)
  }
}

[...]

PR #749, opened 2015-08-05
Typer#escapingRefs: don't let the types of lower bounds escape

smarter:
In 0efa171e8ccca0d49fc6d800fd21e29f7b7336fd I changed the definition of `NamedPartsAccumulator` to exclude lower bounds as this is required for the implicit search, but `NamedPartsAccumulator` is also used by `Typer#escapingRefs` so in the following code:
```scala
class Foo[T]
val z = {
  class C

  ??? : Foo[_ >: C]
}
```
the type of `z` was inferred to be `Foo[_ >: C]` instead of `Foo`.

To avoid this, `NamedPartsAccumulator` will only exclude lower bounds if the parameter `excludeLowerBounds` is explicitely set to `true`.

No test because there is no way to detect that a type has escaped, this might be something that could be added to `TreeChecker`.

<hr>

This PR also removes dead code from `Typer#escapingRefs`.

<hr>

Review by @odersky .

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -257,9 +257,14 @@ object Types {
 
     /** The parts of this type which are type or term refs and which
      *  satisfy predicate `p`.
+     *
+     *  @param p                   The predicate to satisfy
+     *  @param excludeLowerBounds  If set to true, the lower bounds of abstract
+     *                             types will be ignored.
      */
-    def namedPartsWith(p: NamedType => Boolean)(implicit ctx: Context): collection.Set[NamedType] =
-      new NamedPartsAccumulator(p).apply(mutable.LinkedHashSet(), this)
+    def namedPartsWith(p: NamedType => Boolean, excludeLowerBounds: Boolean = false)
+      (implicit ctx: Context): collection.Set[NamedType] =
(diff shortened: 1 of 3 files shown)


odersky:
In fact, given the observed failure of implicits, I would revert completely. Implicits should be searched in lower as well as upper bounds. So NamedParts can be reverted to use both bounds.

smarter:
@odersky : That's one way to do it, but I think this is unnecessary, the failures we get for implicits are due to issues in the inference of type parameters: https://github.com/lampepfl/dotty/issues/739

For example, in https://github.com/lampepfl/dotty/issues/553 we have the type `CC? >: ([+HK$0] => scala.collection.immutable.List[HK$0])[HK$0 =+ (T? >: Int)] <: ([HK$0] =>  <: scala.collection.TraversableOnce[HK$0])` and implicit search fail, but if https://github.com/lampepfl/dotty/issues/739 was fixed then `CC` would be instantiated to `List` like it is in Scala2 and the implicit search would succeed.

I have a prototype that fixes the inference of type parameters, but it uncovered some issues with `Typer#ensureNoLocalRefs` which I need to fix first, including the issue fixed by this pull request.

smarter:
@odersky : could you take another look at this? I just rebased it because it turns out I need this for the sbt phase.

odersky:
LGTM

PR #750, opened 2015-08-05
NamedPartsAccumulator: also include bounds of uninstantiated type variables

smarter:
This PR depends on https://github.com/lampepfl/dotty/pull/749

<hr>

In the following code:

``` scala
class Foo[T](x: T)
val z = {
  class X
  new Foo(new X)
}
```

When we call `Typer#ensureNoLocalRefs` on the block, its type is:

``` scala
Foo[T] where T <: X
```

Before this commit, `Typer#escapingRefs` returned the empty set for this
type since `T` is not instantiated yet, so the type was kept as is, and `X` escaped:

``` scala
val z: Foo[X] = {
  class X
  new Foo(new X)
}
```

We avoid this by inspecting the bounds of uninstantiated type variables
in `NamedPartsAccumulator` which is used by `Typer#escapingRefs`.

<hr>

@odersky : Please review. I'm not sure if this is the proper way to fix this, maybe the instantiation of type parameters could try to avoid types which are not in scope?

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -3135,6 +3135,18 @@ object Types {
             apply(x, tp.underlying)
           case tp: PolyParam =>
             apply(x, tp.underlying)
+          case tp: TypeVar =>
+            val inst = tp.instanceOpt
+            if (inst.exists) apply(x, inst)
+            else {
+              ctx.typerState.ephemeral = true
+              val bounds =
+                if (ctx.typerState.constraint.contains(tp))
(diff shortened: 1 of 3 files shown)


odersky:
I have to think about it. Maybe we should instantiate those type variables that have escaping refs as bounds.

smarter:
I just realized that with the instantiation scheme I proposed in #739 and prototyped in https://github.com/smarter/dotty/commits/fix/tp-instantiation-scheme, this issue does not occur anymore because the type parameters of a method are always instantiated early enough.

Issue #738, opened 2015-07-24
Type parameters are instantiated too early in some cases


smarter:
When a type parameter is used in the second parameter list of a method, the inference is not precise enough, unless it also part of the return type:
```scala
object Test {
  def one[T](ev: T): Nothing = ???
  def two1[T](ev: T)(): Nothing = ???
  def two2a[T]()(ev: T): Nothing = ???
  def two2b[T]()(ev: T): T = ???
  def two3a[T](ev: T)(ev2: T): Nothing = ???
  def two3b[T](ev: T)(ev2: T): T = ???

  def test = {
    one(42) // Correct: Infer T=Int
    two1(42)() // Correct: Infer T=Int
    two2a()(42) // Wrong: Infer T=Any
    two2b()(42) // Correct: Infer T=Int
    two3a(42)(42) // Wrong: Infer T=Any
    two3b(42)(42) // Correct: Infer T=Int
  }
}
```

Here's how `one(42)` is typed:
```
[log frontend]         ==> typing one(42)?
[log frontend]           ==> typing one?
typed ident one in method test
[log frontend]             ==> adapting Test.one of type ([T](ev: T)Nothing)(Test.one) to FunProto(42):??
[log frontend]             <== adapting Test.one of type ([T](ev: T)Nothing)(Test.one) to FunProto(42):? = Test.one[(T?)]
[log frontend]           <== typing Test.one = Test.one[(T?)]
[log frontend]           ==> adapting 42 of type Int(42) to (T?)?
adding constraint T >: Int(42) to
Constraint(
 uninstVars = (T?);
 constrained types = [T](ev: T)Nothing
 bounds = 
     T
 ordering = 
)
added constraint T >: Int(42) to
Constraint(
 uninstVars = (T? >: Int(42));
 constrained types = [T](ev: T)Nothing
 bounds = 
     T >: Int(42)
 ordering = 
) = true
[log frontend]           <== adapting 42 of type Int(42) to (T? >: Int(42)) = 42
[log frontend]           ==> adapting Test.one[(T? >: Int(42))](42) of type Nothing to ??
interpolate undet vars in Nothing, pos = [263..266..270], mode = Mode(ImplicitsEnabled,InferringReturnType), undets = ArrayBuffer((T? >: Int(42))@[263..266])
qualifying undet vars: ArrayBuffer(TypeVar(PolyParam(T)) / (T? >: Int(42))), constraint: Constraint(
 uninstVars = (T? >: Int(42));
[...]

Issue #718, opened 2015-07-09
type parameter inference behaves unexpectedly when a type parameter is used in a contravariant position


smarter:
```scala
class Foo[-T]

object Test {
  def foo[A](x: Foo[A]): Int = 0
  foo(new Foo[Int])
}
```
`scalac` infers `A` to be `Int`:
```scala
fsc -Xprint:typer try/brokeninf.scala
[[syntax trees at end of                     typer]] // brokeninf.scala
package <empty> {
  class Foo[-T] extends scala.AnyRef {
    def <init>(): Foo[T] = {
      Foo.super.<init>();
      ()
    }
  };
  object Test extends scala.AnyRef {
    def <init>(): Test.type = {
      Test.super.<init>();
      ()
    };
    def foo[A](x: Foo[A]): Int = 0;
    Test.this.foo[Int](new Foo[Int]())
  }
}
```
But dotty infers `A` to be `Nothing`:
```scala
./bin/dotc -Xprint:frontend try/brokeninf.scala
result of try/brokeninf.scala after frontend:
package <empty> {
  class Foo[T]() extends Object() { 
    type Foo$$T
    private[this] type T =- Foo$$T
  }
  final lazy module val Test: Test$ = new Test$()
  final module class Test$() extends Object() { this: Test.type => 
    def foo[A](x: Foo[A]): Int = 0
    Test.foo[Nothing](new Foo[Int][Int]())
  }
}
```

This is probably not what we want, especially for function types.

odersky:
Arguably, dotty is right here. In this case, the type parameter `A` is minimized. And `A = Nothing` fullfils the constraint

    Foo[A] >: Foo[Int]

so it is picked. `scalac` has a  somewhat weird fear of `Nothing` :-) do if the result would be `Nothing` it switches bounds. But that leads to other problems. For example, the fact that untyped dotty trees are `Tree[Null]` instead of `Tree[Nothing]` is because of scalac's weird behavior towards `Nothing`.

smarter:
But [the spec](http://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html#local-type-inference) says:
> A maximal type Ti will be chosen if the type parameter ai appears contravariantly in the type T of the expression. A minimal type Ti will be chosen in all other situations, i.e. if the variable appears covariantly, non-variantly or not at all in the type T. We call such a substitution an optimal solution of the given constraint system for the type T.

And the problem is not specific to `Nothing` as far as I can see, for example:
[...]

PR #752, opened 2015-08-09
Change handling of tvars in implicits

### LINKED ISSUE #739: Type parameters should be instantiated before doing an implicit search

odersky:
Fixes #739 by adding the following rule:

Before typing an implicit parameter list of a method m, instantiate all type parameters of m that occur in the type of some preceding value parameter of m.

Review by @smarter, @retronym

### DIFF
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -48,6 +48,7 @@ class Compiler {
            new ExtensionMethods,
            new ExpandSAMs,
            new TailRec,
+           new LiftTry,
            new ClassOf),
       List(new PatternMatcher,
            new ExplicitOuter,
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -68,6 +68,7 @@ class Compiler {
            new LazyVals,
            new Memoize,
            new LinkScala2ImplClasses,
+           new NonLocalReturns,
            new CapturedVars, // capturedVars has a transformUnit: no phases should introduce local mutable vars here
            new Constructors,
            new FunctionalInterfaces,
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -161,6 +161,10 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   def Bind(sym: TermSymbol, body: Tree)(implicit ctx: Context): Bind =
     ta.assignType(untpd.Bind(sym.name, body), sym)
 
+  /** A pattern corrsponding to `sym: tpe` */ 
+  def BindTyped(sym: TermSymbol, tpe: Type)(implicit ctx: Context): Bind =
+    Bind(sym, Typed(Underscore(tpe), TypeTree(tpe)))
+
   def Alternative(trees: List[Tree])(implicit ctx: Context): Alternative =
     ta.assignType(untpd.Alternative(trees), trees)
 
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -13,6 +13,7 @@ import config.Printers._
 import typer.Mode
 import collection.mutable
 import typer.ErrorReporting._
+import transform.Erasure
 
 import scala.annotation.tailrec
 
@@ -161,7 +162,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   def Bind(sym: TermSymbol, body: Tree)(implicit ctx: Context): Bind =
     ta.assignType(untpd.Bind(sym.name, body), sym)
 
-  /** A pattern corrsponding to `sym: tpe` */ 
+  /** A pattern corresponding to `sym: tpe` */
   def BindTyped(sym: TermSymbol, tpe: Type)(implicit ctx: Context): Bind =
     Bind(sym, Typed(Underscore(tpe), TypeTree(tpe)))
 
@@ -737,9 +738,14 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       tree.select(defn.Any_asInstanceOf).appliedToType(tp)
     }
 
-    /** `tree.asInstanceOf[tp]` unless tree's type already conforms to `tp` */
+    /** `tree.asInstanceOf[tp]` (or its box/unbox/cast equivalent when after
+     *  erasure and value and non-value types are mixed),
+     *  unless tree's type already conforms to `tp`.
+     */
     def ensureConforms(tp: Type)(implicit ctx: Context): Tree =
-      if (tree.tpe <:< tp) tree else asInstance(tp)
+      if (tree.tpe <:< tp) tree
+      else if (!ctx.erasedTypes) asInstance(tp)
+      else Erasure.Boxing.adaptToType(tree, tp)
 
     /** If inititializer tree is `_', the default value of its type,
      *  otherwise the tree itself.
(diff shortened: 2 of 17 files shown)


odersky:
Only the last commit is new. Everything else is merged by now.

smarter:
This new scheme seems OK to me but it does not address the issues we have regarding when a type parameter should be minimized or maximized, specifically https://github.com/lampepfl/dotty/issues/738 and https://github.com/lampepfl/dotty/issues/718. It'd be great to see at least https://github.com/lampepfl/dotty/issues/738 addressed to get a better idea of the semantics we're trying to achieve. (Note that both of these issues should be fixed in my prototype)

smarter:
I'm still confused by the rules regarding type parameter maximization, for example:
```scala
import scala.reflect._

class Contra[-T]

object Test {
  def getParam[T](c: Contra[T])(implicit ct: ClassTag[T]): Unit = {
    println(ct)
  }

    getParam(new Contra[Int])
  }
}
```
This will print  `Int` when compiled with scalac, but it will print `Nothing` when compiled with dotty, is that what we want?

odersky:
On Mon, Aug 10, 2015 at 4:44 PM, Guillaume Martres <notifications@github.com
> wrote:

> I'm still confused by the rules regarding type parameter maximization, for
> example:
>
> import scala.reflect._
> class Contra[-T]
> object Test {
>   def getParam[T](c: Contra[T])(implicit ct: ClassTag[T]): Unit = {
>     println(ct)
>   }
>
>   def main(args: Array[String]): Unit = {
>     getParam(new Contra[Int])
>   }
> }
>
> This will print Int when compiled with scalac, but it will print Nothing
> when compiled with dotty, is that what we want?
>

Yes, I think so. Processing the first parameter clause gives T <: Int. Then
when we instantiate T we do so to the lower bound. By what reasoning would
we infer Int?




—
> <https://github.com/lampepfl/dotty/pull/752#issuecomment-129648005>.
>



-- 
Martin Odersky
EPFL

smarter:
> Yes, I think so. Processing the first parameter clause gives T <: Int. Then
when we instantiate T we do so to the lower bound. By what reasoning would
we infer Int?

Because in the trees that the user will write (like `getParam(new Contra[Int])`), `T` only appears contravariantly, so we only add constraints to its upper boud, and it seems intuitive to instantiate `T` to the bound where we added constraints.

If we don't do type parameter maximization here, could you provide an example where we do type parameter maximization and it's useful (the program would not compile otherwise or would produce a different result) ?

retronym:
The reasoning in scalac:

```
scala> :paste -raw
// Entering paste mode (ctrl-D to finish)

import scala.reflect._

class Contra[-T]

object Test {
  def getParam[T](c: Contra[T])(implicit ct: ClassTag[T]): Unit = {
    println(ct)
  }
  def getParam1[T](c: Contra[T]) = ()
}


scala> val getParam1 = typeOf[Test.type].member(TermName("getParam1"))
getParam1: $r.intp.global.Symbol = method getParam1

scala> val tparams = getParam1.info.typeParams
tparams: List[$r.intp.global.Symbol] = List(type T)

scala> val tvars = tparams map (sym => TypeVar(sym))
tvars: List[$r.intp.global.TypeVar] = List(?T)

scala> val withTVars = getParam1.info.resultType.instantiateTypeParams(tparams, tvars)
withTVars: $r.intp.global.Type = (c: Contra[?T])Unit

scala> typeOf[Contra[Int]] <:< withTVars.params.head.info
res19: Boolean = true

scala> val formal1 = getParam1.info.params.head.info
formal1: $r.intp.global.Type = Contra[T]

scala> tvars.head.constr
res20: $r.intp.global.TypeConstraint =  <: Int

scala> varianceInType(getParam1.info.params.head.info)(tparams.head)
res27: scala.reflect.internal.Variance = contravariant

scala> varianceInType(formal1)(tparams.head)
res38: scala.reflect.internal.Variance = contravariant

scala> formal1.typeConstructor.typeParams.head.variance
res39: scala.reflect.internal.Variance = contravariant

scala> varianceInType(formal1.typeArgs.head)(tparams.head).flip
res40: scala.reflect.internal.Variance = contravariant

scala> solve(tvars, tparams, varianceInType(formal1)(tparams.head) :: Nil, upper = false, reflect.internal.Depth.AnyDepth)
res46: Boolean = true

scala> tvars.head.inst
res47: $r.intp.global.Type = Int
```

odersky:
Thanks for the analysis, Jason. It seems to point to a fundamental
difference between scalac's and dotty's type inference. I have to mull this
over.

 - Martin

On Mon, Aug 10, 2015 at 6:16 PM, Jason Zaugg <notifications@github.com>
wrote:

> The reasoning in scalac:
>
> scala> :paste -raw
> // Entering paste mode (ctrl-D to finish)
>
> import scala.reflect._
>
> class Contra[-T]
>
> object Test {
>   def getParam[T](c: Contra[T])(implicit ct: ClassTag[T]): Unit = {
>     println(ct)
>   }
>   def getParam1[T](c: Contra[T]) = ()
> }
>
>
> scala> val getParam1 = typeOf[Test.type].member(TermName("getParam1"))
> getParam1: $r.intp.global.Symbol = method getParam1
>
> scala> val tparams = getParam1.info.typeParams
> tparams: List[$r.intp.global.Symbol] = List(type T)
>
> scala> val tvars = tparams map (sym => TypeVar(sym))
> tvars: List[$r.intp.global.TypeVar] = List(?T)
>
> scala> val withTVars = getParam1.info.resultType.instantiateTypeParams(tparams, tvars)
> withTVars: $r.intp.global.Type = (c: Contra[?T])Unit
>
> scala> typeOf[Contra[Int]] <:< withTVars.params.head.info
> res19: Boolean = true
>
> scala> val formal1 = getParam1.info.params.head.info
> formal1: $r.intp.global.Type = Contra[T]
>
> scala> tvars.head.constr
> res20: $r.intp.global.TypeConstraint =  <: Int
>
> scala> varianceInType(getParam1.info.params.head.info)(tparams.head)
> res27: scala.reflect.internal.Variance = contravariant
>
> scala> varianceInType(formal1)(tparams.head)
> res38: scala.reflect.internal.Variance = contravariant
>
> scala> formal1.typeConstructor.typeParams.head.variance
> res39: scala.reflect.internal.Variance = contravariant
>
> scala> varianceInType(formal1.typeArgs.head)(tparams.head).flip
> res40: scala.reflect.internal.Variance = contravariant
>
> scala> solve(tvars, tparams, varianceInType(formal1)(tparams.head) :: Nil, upper = false, reflect.internal.Depth.AnyDepth)
> res46: Boolean = true
>
> scala> tvars.head.inst
> res47: $r.intp.global.Type = Int
>
> —
> <https://github.com/lampepfl/dotty/pull/752#issuecomment-129662448>.
>



-- 
Martin Odersky
EPFL

odersky:
Part of this is now in #799. I'll revive the rest in a separate PR.

PR #753, opened 2015-08-13
Workaround interaction between LiftTry and CapturedVars

DarkDimius:
This workaround is inefficient, but allows test to pass. 
The only complete way that I can think of currently is to run LiftTry and CapturedVars until they reach a fix point, to make sure that CapturedVars lifts everything that LiftTry lifted to a method and that LiftTry will account for all assigns that work on captured variables, as they are performed on non-empty stack.

@odersky please review.

### DIFF
--- a/src/dotty/tools/dotc/transform/LiftTry.scala
+++ b/src/dotty/tools/dotc/transform/LiftTry.scala
@@ -42,8 +42,9 @@ class LiftTry extends MiniPhase with IdentityDenotTransformer { thisTransform =>
       else liftingTransform
 
     override def prepareForAssign(tree: Assign)(implicit ctx: Context) =
-      if (tree.lhs.symbol.maybeOwner == ctx.owner.enclosingMethod) this
-      else liftingTransform
+      /* this is inefficient in many cases,
+         but after CapturedVars is completed this assign could become one with non-empty stack */
+     liftingTransform
 
     override def prepareForReturn(tree: Return)(implicit ctx: Context) =
       if (!isNonLocalReturn(tree)) this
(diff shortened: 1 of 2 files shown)


odersky:
@DarkDimius OK, now I see the problem. This is a really difficult interaction between two phases which in the general case requires a fixpoint iteration, as you suggest. It strikes me that an easier way out is to make captured var stores not push values on the stack before evaluating the expression. I.e. 

  intRef.elem = expr

would be translated to something like:

  <expr>
  aload intRef
  swap
  istore elem

instead of

  aload intRef
  expr
  istore elem

@retronym Maybe scalac does this also? What do people think, should we do this?

odersky:
@DarkDimius If this problem is a blocker for the bootstrap I am LGTM with the workaround to be merged now.

DarkDimius:
@odersky could you elaborate why the proposed translation is better?
It also uses at least a single entry of stack, making it have the very same problem: if any exception is thrown, stack entry will be pop-ed.

odersky:
@darkdimius  The new proposed translation evaluated expr entirely before anything is pushed. (It did not show previously in my comment because I wrote "expr" inside "<" ">" tags and that got swallowed by the markup it seems.

DarkDimius:
@odersky I understand your idea now. The same result can be achieved inside CapturedVars by introducing a temporary variable:
```scala
s = expr
```

instead of becoming
```scala
s.elem = expr
```
will become

```scala
val tmp$0 = exp
s.elem = tmp$0
```

This will enforce that CapturedVars does not introduce try-s on non-empty stack.
I've tried this idea, it works on this example. But I stumbled upon some problems with `changeOwnerAfter` interaction between Memoize and CapturedVars. I will have a better look next week.

odersky:
@darkdimius Great idea about the temportary variable. I implemented it in #758.

PR #754, opened 2015-08-13
Fixes to emission of static initializers and forwarders in backend.

### LINKED ISSUE #751: Bootstrap failure: wrong static initializer of inner class
If I compile everything in src/dotty and run the resulting program without arguments I get:

> dotc
Exception in thread "main" java.lang.NoSuchMethodError: dotty.tools.dotc.ast.Trees$Instance$TreeMap$: method <init>()V not found
	at dotty.tools.dotc.ast.Trees$Instance$TreeMap$.<clinit>(Trees.scala)
	at dotty.tools.dotc.ast.Trees$Instance.<init>(Trees.scala:1087)
	at dotty.tools.dotc.ast.untpd$.<init>(untpd.scala:13)
	at dotty.tools.dotc.ast.untpd$.<clinit>(untpd.scala)
	at dotty.tools.dotc.core.Contexts$InitialContext.<init>(Contexts.scala:467)
	at dotty.tools.dotc.core.Contexts$ContextBase.<init>(Contexts.scala:493)

Investigation with javap shows that TreeMap has a static initializer which tries to create a TreeMap object with an empty constructor. But TreeMap is an inner class, so the constructor takes an outer pointer.

> Compiled from "Trees.scala"
public final class dotty.tools.dotc.ast.Trees$Instance$TreeMap$ {
  public static final dotty.tools.dotc.ast.Trees$Instance$TreeMap$ MODULE$;

>  public static {};
    Code:
       0: new           #2                  // class dotty/tools/dotc/ast/Trees$Instance$TreeMap$
       3: invokespecial #23                 // Method "<init>":()V
       6: return

>  public dotty.tools.dotc.ast.Trees$Instance$TreeMap$(dotty.tools.dotc.ast.Trees$Instance);
    Code:

The whole static initializer seems to be a random artefact, since TreeMap does not have a companion object. 

Edit: No, it does have a synthetic companion object (which contains default arguments). But since it's an inner class of the outer class Instance it should not have a static initializer that creates an instance, it seems to me.

DarkDimius:
Fixes #751

### DIFF
--- a/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -632,7 +632,16 @@ class DottyBackendInterface()(implicit ctx: Context) extends BackendInterface{
      * True for module classes of modules that are top-level or owned only by objects. Module classes
      * for such objects will get a MODULE$ flag and a corresponding static initializer.
      */
-    def isStaticModuleClass: Boolean = sym.isStatic && (sym is Flags.Module)
+    def isStaticModuleClass: Boolean =
+      (sym is Flags.Module) && {
+        // scalac uses atPickling here
+        // this would not work if modules are created after pickling
+        // for example by specialization
+        val original = toDenot(sym).initial
+        val validity = original.validFor
+        val shiftedContext = ctx.withPhase(validity.phaseId)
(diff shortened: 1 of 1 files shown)


odersky:
LGTM

PR #755, opened 2015-08-13
GetClass: get classSymbol reliably

DarkDimius:
TermRefs do not have a classSymbol.

@AlexSikia thanks for reporting


DarkDimius:
> sbt.ResolveException: unresolved dependency: org.scala-sbt#completion;0.13.5: not found

/rebuild

PR #757, opened 2015-08-14
Typo fix in trait Lambda$II definition

senia-psm:


odersky:
LGTM. Thanks for the contribution!

PR #758, opened 2015-08-15
Fix lift try and captured vars interaction

odersky:
Replaces #753. Review by @DarkDimius.


odersky:
Note: This PR uncovered (and fixes) a deeper problem about miniphase assembly which caused a
CapturedVars treetransform to come with an idx of 0, so the system would think tarnsformFollowing needs to run LazyVals on it, so we got NotDefinedHere errors.

odersky:
@darkdimius Can you take a look at what fails here? The PR tests OK locally, but it seems we run out of memory on jenkins.

odersky:
It was a spurious failure, after all.

DarkDimius:
Otherwise LGTM

odersky:
@darkdimius Last two commits ready for review.

DarkDimius:
LGTM

PR #760, opened 2015-08-17
Fixes to scala2 Mixin

DarkDimius:
Those fixes allow to compile files inheriting from non-trivial Scala2 traits. Needed to compile [Names.scala](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/Names.scala#L38-L41)


DarkDimius:
@odersky please review.

odersky:
Otherwise LGTM

odersky:
@DarkDimius Makes sense, thanks for the explanation. Let's go with your version then.

PR #764, opened 2015-08-21
Mixin: class defined abstract members take precedence over interface defined.

DarkDimius:
http://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.3
If there is a class that defines method, it will take precedence over all interface definitions. 
Even if class defined method is abstract but interface defined is not.

@odersky please review. Blocks bootstrap.


odersky:
LGTM

PR #766, opened 2015-08-21
ElimRepeated should also work on Expr trees.

DarkDimius:
Needed for bootstrap.
@odersky please review.

### DIFF
--- a/src/dotty/tools/dotc/transform/ElimRepeated.scala
+++ b/src/dotty/tools/dotc/transform/ElimRepeated.scala
@@ -43,10 +43,8 @@ class ElimRepeated extends MiniPhaseTransform with InfoTransformer with Annotati
           paramTypes.init :+ last
         } else paramTypes
       tp.derivedMethodType(paramNames, paramTypes1, resultType1)
-    case tp: ExprType =>
-      val resultType = elimRepeated(tp.resultType)
-      if (resultType eq tp.resultType) tp
-      else ExprType(resultType)
+    case tp: ExprType if tp.resultType.isRepeatedParam =>
+       ExprType(tp.resultType.underlyingIfRepeated(isJava = false))
     case tp: PolyType =>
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/transform/ElimRepeated.scala
```diff
@@ -43,6 +43,8 @@ class ElimRepeated extends MiniPhaseTransform with InfoTransformer with Annotati
           paramTypes.init :+ last
         } else paramTypes
       tp.derivedMethodType(paramNames, paramTypes1, resultType1)
+    case tp: ExprType if tp.resultType.isRepeatedParam =>
+       ExprType(tp.resultType.underlyingIfRepeated(isJava = false))
```
odersky: I don't understand what the change achieves. elimRepeated removes repeated method parameters. It does not touch other occurrences of repeated types. In fact the expectation is that these were not generated in the first place. If elimRepeated should touch ExprTypes then it also should handle naked repeated types which are not wrapped in ExprTypes. But it does not do this.

DarkDimius: https://github.com/scala/scala/blob/2.11.x/src/library/scala/StringContext.scala#L54

Accessor `parts` gets type `ExprType(Repeated[String])`. Where would you propose to handle such types? They should not survive until erasure, as erasure will erase them to `<repeated>` which is wrong.

odersky: I think `parts` should be of type `Expr(Seq[String])` instead.

DarkDimius: >Question: Did you compile StringContext with dotty, or import it with the unpickler? 

It was unpickled from Scala2.
Similar code does not typecheck with dotty: https://github.com/lampepfl/dotty/pull/768

odersky:
Replaced by #770

PR #767, opened 2015-08-21
Fix #756 super accessors

### LINKED ISSUE #756: dotc may generate duplicate super accessors, resulting in ClassFormatError
here's the input code:
```
~/dotty % cat S.scala
trait T { def foo: Int = 3 }
trait T1 extends T { override def foo = super.foo }
trait T2 extends T { override def foo = super.foo }
object C extends T2 with T1 {
  def main(args: Array[String]) = ()
}
```

with Scala 2.11.7, no error:

```
~/dotty % /usr/local/scala-2.11.7/bin/scalac S.scala
~/dotty % java -classpath . C
```

but with Dotty:

```
~/dotty % rm *.class
~/dotty % bin/dotc S.scala
~/dotty % java -classpath . C
Exception in thread "main" java.lang.ClassFormatError: Duplicate method name&signature in class file C$
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
...
```

the problem is that the generated code has two identical methods named `super$foo`:

```
~/dotty % cfr-decompiler C\$.class
...
public final class C$
implements T,
T2,
T1 {
...
    @Override
    public /* synthetic */ int super$foo() {
        return super.foo();
    }
    @Override
    public /* synthetic */ int super$foo() {
        return super.foo();
    }
...
}
```

odersky:
Fix super accessor handling, specifically the problems detected in #756. Review by @DarkDimius, @SethTisue

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -96,6 +96,7 @@ class tests extends CompilerTest {
 
   @Test def new_all = compileFiles(newDir, twice)
 
+  @Test def neg_abstractOverride() = compileFile(negDir, "abstract-override", xerrors = 2)
   @Test def neg_blockescapes() = compileFile(negDir, "blockescapesNeg", xerrors = 1)
   @Test def neg_typedapply() = compileFile(negDir, "typedapply", xerrors = 4)
   @Test def neg_typedidents() = compileFile(negDir, "typedIdents", xerrors = 2)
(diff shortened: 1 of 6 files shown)


test/dotc/tests.scala
```diff
@@ -96,6 +96,7 @@ class tests extends CompilerTest {
 
   @Test def new_all = compileFiles(newDir, twice)
 
+  @Test def neg_abstractOverride() = compileFile(negDir, "abstract-override", xerrors = 2)
```
DarkDimius: `abstract-override.scala` is absent in this PR.

DarkDimius:
LGTM

PR #768, opened 2015-08-21
Test that repeated exprs work.

DarkDimius:
This commit is separate from #786, as it is going to fail.
```scala
i768.scala:1: error: type mismatch:
 found   : scala.collection.Seq[String] @Repeated(a)
 required: String
case class A(a: String*){}
           ^
one error found
```
@odersky could you please take this over?


odersky:
Replaced by #770

PR #771, opened 2015-08-23
Fix final vals

odersky:
An implementation of SIP 23 and fixes to constant propagation and code generation for constants. SIP 23 was added because it makes testing easier. Review by @DarkDimius


DarkDimius:
LGTM

PR #773, opened 2015-08-24
Memoize: duplicate scala2 behaviour: don't create fields for final vals.

DarkDimius:
This affect initialisation order and we rely on artifacts of initialisation order in Dotty.


DarkDimius:
@odersky your proposed code will perform transformation when `stats` are not idempotent. Is this intended?

DarkDimius:
It is not yet ready for review. It fails the supposed test.

odersky:
@darkdimius: No, notice the isPureExpr guard. That will do the right test.

odersky:
Rebased to master and added fixes to make this work.

odersky:
Also, applied my own earlier suggested patches.

PR #774, opened 2015-08-24
Constructors: fixes to maintain needed private fields.

### LINKED ISSUE #772: Constructors removes private fields used only in inner classes
For example `ctx` field here: 
https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/typer/VarianceChecker.scala#L21

I'm preparing a fix.

DarkDimius:
fixes #772

### DIFF
--- a/src/dotty/tools/dotc/transform/Constructors.scala
+++ b/src/dotty/tools/dotc/transform/Constructors.scala
@@ -41,7 +41,6 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
   // 3. It is accessed on an object other than `this`
   // 4. It is a mutable parameter accessor
   // 5. It is has a wildcard initializer `_`
-
   private var retainedPrivateVals = mutable.Set[Symbol]()
   private var seenPrivateVals = mutable.Set[Symbol]()
   private var insideConstructor = false
@@ -49,11 +48,8 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
   private def markUsedPrivateSymbols(tree: RefTree)(implicit ctx: Context): Unit = {
 
     val sym = tree.symbol
-    def retain = {
-      if (sym.toString.contains("initialValues"))
-        println("hooooo")
+    def retain =
       retainedPrivateVals.add(sym)
-    }
 
     if (mightBeDropped(sym) && sym.owner.isClass) {
       val owner = sym.owner.asClass
@@ -134,8 +130,7 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
 
   override def transformTemplate(tree: Template)(implicit ctx: Context, info: TransformerInfo): Tree = {
     val cls = ctx.owner.asClass
-    if (cls.toString.contains("VarianceChecker"))
-      println("hoho")
+
     val constr @ DefDef(nme.CONSTRUCTOR, Nil, vparams :: Nil, _, EmptyTree) = tree.constr
 
     // Produce aligned accessors and constructor parameters. We have to adjust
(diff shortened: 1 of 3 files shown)


DarkDimius:
> [error] (*:update) sbt.ResolveException: unresolved dependency: org.scala-sbt#relation;0.13.5: not found

rebuild

DarkDimius:
@odersky please review

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,61 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+        tree match {
+          case Ident(_) | Select(This(_), _) =>
+            def inConstructor = {
+              val method = ctx.owner.enclosingMethod
+              method.isPrimaryConstructor && ctx.owner.enclosingClass == owner
+            }
+            if (inConstructor && (sym.is(ParamAccessor) || seenPrivateVals.contains(sym))) {
+              // used inside constructor, accessed on this,
+              // could use constructor argument instead, no need to retain field
+              println("hoha")
```
smarter: hoha! You probably don't want to keep this :)

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,61 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+  // Collect all private parameter accessors and value definitions that need
+  // to be retained. There are several reasons why a parameter accessor or
+  // definition might need to be retained:
+  // 1. It is accessed after the constructor has finished
+  // 2. It is accessed before it is defined
+  // 3. It is accessed on an object other than `this`
+  // 4. It is a mutable parameter accessor
+  // 5. It is has a wildcard initializer `_`
+  private var retainedPrivateVals = mutable.Set[Symbol]()
+  private var seenPrivateVals = mutable.Set[Symbol]()
```
smarter: `retainedPrivateVals` and `seenPrivateVals` do not need to be `var`.

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,58 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+  private def markUsedPrivateSymbols(tree: RefTree)(implicit ctx: Context): Unit = {
+
+    val sym = tree.symbol
+    def retain =
+      retainedPrivateVals.add(sym)
+
+    if (mightBeDropped(sym) && sym.owner.isClass) {
+      val owner = sym.owner.asClass
+
+        tree match {
```
smarter: The return value of this match is not used anywhere.

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,58 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+          case Ident(_) | Select(This(_), _) =>
+            def inConstructor = {
+              val method = ctx.owner.enclosingMethod
+              method.isPrimaryConstructor && ctx.owner.enclosingClass == owner
+            }
+            if (inConstructor && (sym.is(ParamAccessor) || seenPrivateVals.contains(sym))) {
```
smarter: This block is empty.

src/dotty/tools/dotc/transform/Constructors.scala
DarkDimius: this match returns Unit. How could it be used?

src/dotty/tools/dotc/transform/Constructors.scala
DarkDimius: This is intended this block contains comment on why nothing happens here

src/dotty/tools/dotc/transform/Constructors.scala
smarter: Ah ok, I missed that `retain` had side-effects, I suggest replacing `def retain` by `def retain()` and using `retain()` to follow the scala style.

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,58 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+  private val retainedPrivateVals = mutable.Set[Symbol]()
+  private val seenPrivateVals = mutable.Set[Symbol]()
+
+  private def markUsedPrivateSymbols(tree: RefTree)(implicit ctx: Context): Unit = {
+
+    val sym = tree.symbol
+    def retain =
+      retainedPrivateVals.add(sym)
+
+    if (mightBeDropped(sym) && sym.owner.isClass) {
```
odersky: I'd reverse the order of the two conditions in the &&. mightBeDropped makes sense only for class members.

src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -33,6 +33,58 @@ class Constructors extends MiniPhaseTransform with SymTransformer { thisTransfor
...
+  }
+
+  override def transformSelect(tree: tpd.Select)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    markUsedPrivateSymbols(tree)
+    tree
+  }
+
+  override def transformValDef(tree: tpd.ValDef)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    if (mightBeDropped(tree.symbol))
+      (if (isWildcardStarArg(tree.rhs)) retainedPrivateVals else seenPrivateVals) += tree.symbol
```
odersky: Why the `isWildcardStarArg` test?

odersky:
Otherwise LGTM

PR #775, opened 2015-08-24
LazyVals: fix volatile lazy vals initialization

DarkDimius:
Used to incorrectly set status field.


DarkDimius:
@odersky please review.

PR #790, opened 2015-09-14
Memoize: Support creating valid trait setters

DarkDimius:
used to create trait setters that were assigning to non-existing fields.


DarkDimius:
@odersky please review.

odersky:
Otherwise LGTM

DarkDimius:
>You could shorten these two lines to
`name.drop(start).fieldName`

This would create one more intermediate name. I would prefer to keep it in current state as it is being called from `typedAssign`

odersky:
@DarkDimius OK, it's not a big issue anyway.

PR #794, opened 2015-09-16
Better diagnostic and printing for expandProvate assert

odersky:
To test whether it happens on jenkins (which it should)


DarkDimius:
Merged in as part of #795

PR #796, opened 2015-09-17
Fix #789 problems in erasure

### LINKED ISSUE #789: Problems in erasure
./out/dotty/tools/dotc/core/Phases.class

```java
    default public Object atPhase(Phase phase, Function1 op) {
        return this.atPhase((Phase)BoxesRunTime.boxToInteger((int)phase.id()), op);
    }
```

is obviously going to fail with java.lang.Integer cannot be cast to dotty.tools.dotc.core.Phases$Phase.

Unfortunatelly, running dotty with `isDerivedValueClass` set to false does not work, so this blocks bootstrap.

odersky:
Review by @DarkDimius


DarkDimius:
LGTM

PR #799, opened 2015-09-18
Change inference

odersky:
Change inference scheme to look at how type variables are constrained
instead of just taking into account the variance in which they occur in some
type. Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/core/ConstraintHandling.scala
+++ b/src/dotty/tools/dotc/core/ConstraintHandling.scala
@@ -104,7 +104,7 @@ trait ConstraintHandling {
     up.forall(addOneBound(_, lo, isUpper = false))
   }
 
-  protected final def isSubTypeWhenFrozen(tp1: Type, tp2: Type): Boolean = {
+  final def isSubTypeWhenFrozen(tp1: Type, tp2: Type): Boolean = {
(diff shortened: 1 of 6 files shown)


odersky:
Can we get this reviewed please? it's been lingering for a long time.

smarter:
Yes, I'm on it

smarter:
Overall I think this PR is a good improvement on the current situation but there are still things which don't work correctly, for example:
```scala
object Test {
  def one[T](x: T)(implicit ev: T): Nothing = ???

  def test = {
    implicit val ii: Int = 42

    one(10)
    // error: ambiguous implicits: both getter StringCanBuildFrom in object Predef$
    // and getter NothingClassTag in object DottyPredef$ match type Any of
    // parameter ev of method one in object Test$
  }
}
```
I'm working on fixing this(it should only require us to be less eager to instantiate in `interpolateUndetVars`) and other issues. I'm not sure if the new design of "interpolate in the direction that was constrained" is the correct thing to do, but we can always revert that in the future if it ends up being a problem.

odersky:
OK, should we get this in then and do further improvements in a separate PR?

smarter:
Yes, sounds good to me.

PR #787, opened 2015-09-11
Fix #779 recursive hk types

odersky:
Closes #779. Review by @DarkDimius

Issue #779, opened 2015-09-01
Compiling collections: subtyping failures for Factories


DarkDimius:
while compiling `scala.collection.immutable.Map`

```scala

$ dotc Map.scala
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$GenMapFactory$$CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$MapFactory$$CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type scala$collection$generic$ImmutableMapFactory$$CC; assuming they are invariant
Map.scala:79: error: Type argument [HK$0, +HK$1] => scala.collection.immutable.Map[HK$0, HK$1] does not conform to upper bound [HK$0, +HK$1] =>
   <:
    scala.collection.immutable.Map[HK$0, HK$1] &
      scala.collection.immutable.MapLike[HK$0, HK$1,
        LazyRef([HK$0, +HK$1] => scala.collection.immutable.Map[HK$0, HK$1])[
          HK$0
        , HK$1]
      ]
object Map extends ImmutableMapFactory[Map] {
                                       ^
three warnings found
one error found
```

PR #802, opened 2015-09-21
Change hk take 2

odersky:
Rebased #731 to master and added some changes to make all tests pass, including the standard collection tests in #787. In retrospect, it would have been good to add this ealier because the fixes
of #779 in #787 were cruder than what is in this PR. In particular, we no longer need to intercept cyclic reference errors.

Review by @DarkDimius or @smarter

### DIFF
--- a/src/dotty/tools/dotc/config/Config.scala
+++ b/src/dotty/tools/dotc/config/Config.scala
@@ -71,6 +71,11 @@ object Config {
   /** If this flag is set, take the fast path when comparing same-named type-aliases and types */
   final val fastPathForRefinedSubtype = true
 
+  /** If this flag is set, $apply projections are checked that they apply to a
+   *  higher-kinded type.
+   */
+  final val checkProjections = false
+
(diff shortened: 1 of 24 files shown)


odersky:
Seems the tests are not running?

SethTisue:
I'll look into why Scabot isn't kicking in here.

SethTisue:
it was because the PR includes more than 30 commits, which exceeded the GitHub API's default page size. I fixed Scabot to request a page size of 100 commits (the maximum allowed), and manually started a new Jenkins run: https://scala-ci.typesafe.com/view/dotty/job/dotty-master-validate-main/437/

smarter:
So what happens when we have a PR with 101 commits? :)

SethTisue:
> So what happens when we have a PR with 101 commits

I dare you!

DarkDimius:
Otherwise LGTM, though I do not fully understand all the consequences of this change.
Glad to have `Map.scala` pass.

PR #805, opened 2015-09-25
Change : freeze bounds

odersky:
There was a big hole in the type system so far in that lubs and glbs (and some other ops as well) used isSubType instead of isSubTypeWhenFrozen. Consequently, taking the lub of, say Set[A] and Set[B]
would have unified A and B! Keeps the constraints nice and tidy but is wrong. Fixing this uncovered several other problems which needed to be fixed in turn.

Review by @DarkDimius


odersky:
Note that this PR is based on #802.

odersky:
Rebased to master

odersky:
If there are no more reviewing comments I am going to merge this later today.

odersky:
@darkdimius I added some explanations to the And/Or rewritings. I did see some code fail which got fixed by the rewriting, but currently all tests pass with or without the rewriting. So it would still be good to get a test which pinpoints the difference at some point.

odersky:
@smarter 

> Hmm actually I don't think that will work, we can replace P by Q in P#T only if T is a member of Q, so we have to set variance = 0 to be conservative.

I did not understand that remark. Can you elaborate?

odersky:
I dropped the strawman from this PR, in order to resubmit this as a separate PR.

odersky:
I now understood @smarter's argument why 0a48b2a is unsafe.

odersky:
Going to merge now, I think we got all the reviews we could expect.

PR #813, opened 2015-09-30
Dotty deviation: cannot compile GenTraversableFactory

DarkDimius:
```
[info] Test dotc.tests.pos_GenTraversableFactory started
./tests/pos/GenTraversableFactory.scala:57: error: genericBuilder is not a member of GenTraversableFactory.this.Coll
    def apply(from: Coll) = from.genericBuilder[A]
                                 ^
./tests/pos/GenTraversableFactory.scala:222: error: < is not a member of T(step)
    while (if (step < zero) end < i else i < end) {
               ^
./tests/pos/GenTraversableFactory.scala:222: error: < is not a member of T(end)
    while (if (step < zero) end < i else i < end) {
                            ^
./tests/pos/GenTraversableFactory.scala:222: error: < is not a member of T
    while (if (step < zero) end < i else i < end) {
                                         ^
5 errors found
```

We also seem to have a minor bug somewhere: only 4 errors where reported, but total is 5 errors.

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -94,6 +94,7 @@ class tests extends CompilerTest {
   @Test def pos_anonClassSubtyping = compileFile(posDir, "anonClassSubtyping", twice)
   @Test def pos_extmethods = compileFile(posDir, "extmethods", twice)
   @Test def pos_companions = compileFile(posDir, "companions", twice)
+  @Test def pos_GenTraversableFactory = compileFile(posDir, "GenTraversableFactory", twice)
 
   @Test def pos_all = compileFiles(posDir) // twice omitted to make tests run faster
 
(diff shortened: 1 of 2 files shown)


DarkDimius:
@odersky could you please have a look?

odersky:
When I run this on #805 I get the first error like this:

GenTraversableFactory.scala:52: error: illegal wildcard type argument; does not correspond to type parameter of a class
  class GenericCanBuildFrom[A] extends CanBuildFrom[CC[_], A, CC[A]] {
                                                       ^

That explains it. We cannot do existentials over hk types. The code will have to be rewritten to compile under dotty.

DarkDimius:
@odersky, could you try compiling the file from this branch, not the file in standard library.
This file was modified to compile with dotty

odersky:
GenTraversableFactory is not legal code because it uses a hk type as the type of a parameter. This is now flagged by #821.

Issue #741, opened 2015-07-27
Flaw in type parameter encoding?


smarter:
The following:
```scala
class Foo[A]

class Outer {
  class Bar[B] extends Foo[B]
}

object Test {
  def test = {
    val foo: Foo[Int] = {
      val o = new Outer
      new o.Bar[Int]
     }
  }
}
```
Fails with:
```scala
error: type mismatch:
 found   : Foo{A = Outer$Bar$$B; Outer$Bar$$B = Int}
 required: Foo[Int]
    val foo: Foo[Int] = {
                        ^
```

The issue is that in `TypeAssigner#avoid` we replace:
```scala
o.Bar { Outer$bar$$B = Int }
```
by (this is not the best solution, see #711, but that's a separate issue):
```scala
Foo { Foo$$A = Outer$bar$$B } { Outer$Bar$$B = Int }
```
But at this point we cannot infer that `Foo$$A = Int` because these refinements are not prefixed by `RefinedThis` so they're independent. It seems that we should either:
- change the encoding of type parameters
- or add a simple and robust method to go from `o.Bar { Outer$Bar$$B = Int }` to its "real" parent type which is `Foo { Foo$$A = Int }`.

@odersky : what do you think?

Issue #711, opened 2015-07-04
The approximation of path-dependent classes in TypeAssigner#avoid is too imprecise


smarter:
```scala
class Outer {
  class Inner
}
object Test {
  def test = {
    // works
    val a: Outer#Inner = {
      val o = new Outer
      new o.Inner: Outer#Inner
    }


    // error: type mismatch:
    // found   : Object
    // required: Outer#Inner
    //    val b: Outer#Inner = {
    //                         ^
    val b: Outer#Inner = {
      val o = new Outer
      new o.Inner
    }
  }
}
```
The approximation is done in `TypeAssigner#avoid`:
```scala
        case tp: TypeRef if (forbidden contains tp.symbol) || toAvoid(tp.prefix) =>
           //...
```
Instead of approximating the class `o.Inner` by its parent class `Object`, it should approximate it by its supertype`Outer#Inner`.

PR #817, opened 2015-10-02
Fix various issues in `TypeAssigner#avoid`

smarter:
This fixes some of the issues with escaping refs, notably #711 and #741.

It does not include fixes for the issues with uninstantiated type parameters since we're not sure what the best solution for that is yet, see #750.

Review by @odersky


odersky:
Otherwise LGTM

Issue #818, opened 2015-10-05
Wanted: Strawman proposals for new collections architecture


odersky:
A redesign of the standard library is on the roadmap for one of the next Scala versions (could be as early as 2.13). Some requirements for the redesign should be:
1. Normal usage of collections should stay the same. That is, the standard operations
   on collections should still be available under the same names. Some more exotic and advanced
   scenarios such as `breakOut` can be dropped if alternative ways to achieve the same functionality
   exist.
2. User-defined implementations of collections should port to the new library as far as is reasonable. We should allow some breakage here, if it is necessary to achieve other goals.
3. We should strive for simplicity in APIs and implementations.
4. We should strive to better separate interfaces from implementation and avoid the fragile base
   class problem, where too much gets inherited automatically.
5. We should try to simplify the inheritance graphs, in particular those of often-used collections
   such as lists.
6. We should improve the integration of strict and lazy collections by means of a better architecture for views. Views should avoid accidental forcing, and should omit transforms from collections that require such forcing. However, forcing is still needed to support aggregations. 
7. We should try to integrate Java8 streams and/or have our own high-performance implementations for
   parallel collections.
8. We should generally be at least on par with current collections in what concerns efficiency. In particular, we should still allow specializations of collection operations for particular implementations. These optimizations should still work if the static collection type is abstract. E.g. an optimized implementation of `++` on `ArrayBuffer`s should be called even if the static types of the operands of `++` are `Iterable`s. 
9. The design should be friendly to optimizations such as inlining and, possibly, more advanced
   whole program optimizations.

To gain experience it would be good to experiment with some prototypes that illustrate possible designs and can point to strengths and weaknesses. A full collections design is too heavy to fill this role. It takes too long to write and understand. As an alternative, I propose to implement just a subset of the required functionality. The subset should implementable in about 500 LOC or less and at the same time should be as representative as possible of the whole collections. The API should be the same for all strawman proposals. 

After some experimentation, I came up with the following proposal for the API to be implemented by strawman proposals. There's still some modest room to add things if people feel it is important.
## Base Traits

```
Iterator
Iterable
Seq
```

For now we leave out maps and sets (which does not say they are not important). We gloss over the mutabile/immutable distinction. All collections are in the same package (this is only for the strawman, to keep things simple, not for the full collections, of course).

`Iterable` and `Seq` are considered collection types, but `Iterator` is not.
## Collection Implementations

```
List                    
ListBuffer   
ArrayBuffer          
[...]

PR #819, opened 2015-10-05
Collection strawman2

DarkDimius:
To support discussion in #818. 

It's not going to pass tests before #805 is merged.

It has 4 changes that are applied to Martin’s version,
and I believe they could be discussed independently:

https://github.com/dotty-staging/dotty/commit/48afa750b803eca44eb295c8889c2c3612c29529
Gets rid of view class. View is a simple type alias to () => Iterator[A].
This points the major difference between Iterators and views: iterators are consumed by 
any usage, while views are reusable.

https://github.com/dotty-staging/dotty/commit/52816883cd6a86a59cd92efdc482cf441aecd8aa
Makes .view method be a decorator. This is done in order to decouple views from collection higherarhy
and make implementation of custom collections easier. It would also allow to modularise collections library to a higher degree. I believe that the very same should be done to .par.

https://github.com/dotty-staging/dotty/commit/b77ab7c7a9f42df8996a3e3d814abc0406b2cca6
Makes Iterators have only next and hasNext methods. The motivation could be found in this code sample:

```
def intersectionIsEmpty[A](it1: Iterator[A], it2: Iterator[A]) =
       !(it1 exists (it2 contains _))
```

Which was taken from old bug in Dotty codebase https://github.com/dotty-staging/dotty/commit/03ec379926f1f900f09c50dd038fba86feae70f6

I believe that this code simply should not compile: both contains and exists have no side effects on 
mutable and immutable collections, but it has a silent side-effect on Iterators. It does not follow principle
of least surprise, I would actually say that this is one of most surprising methods for new-comments that 
felt the joy of ”you can use the same methods on parallel, mutable and immutable collections” 
when they discover that there is ”but be very careful if you use those methods on iterators”.

Ok, so what the change does: it moves all consuming methods from Iterator[A], to a decorator over View[A], that is, to () => Iterator[A]. This follows the patch that makes difference in use-cases of iterator[A] and View[A] more obvious, that I started in previous commit.

https://github.com/dotty-staging/dotty/commit/5244c5cf1f3d2ff2917a09e9760f9ad887ca4ce0
Extracts a LengthType from all the signatures for documentation purposes & makes it be Long for big-data needs.
This change has a problem with compatibility with ScalaJs, as they simulate Longs.

### DIFF
--- a/src/strawman/collections/CollectionStrawMan2.scala
+++ b/src/strawman/collections/CollectionStrawMan2.scala
@@ -40,6 +40,8 @@ object CollectionStrawMan1 {
     def length: Int
   }
 
+  type View[A] = () => Iterator[A]
+
   /* ------------ Operations ----------------------------------- */
 
   /** Operations returning types unrelated to current collection */
@@ -51,7 +53,7 @@ object CollectionStrawMan1 {
     def indexWhere(p: A => Boolean): Int = iterator.indexWhere(p)
     def isEmpty: Boolean = !iterator.hasNext
     def head: A = iterator.next
-    def view: View[A] = new View(iterator)
+    def view: View[A] = iterator _
     def collectAs[C[X] <: Iterable[X]](fi: FromIterator[C]): C[A] = fi.fromIterator(iterator)
   }
 
@@ -210,9 +212,6 @@ object CollectionStrawMan1 {
   }
 
   /** Concrete collection type: View */
-  class View[+A](it: => Iterator[A]) extends CanIterate[A] {
-    def iterator = it
-  }
 
   implicit class ViewOps[A](val v: View[A]) extends AnyVal with Ops[A] {
     def iterator = v.iterator
@@ -222,13 +221,13 @@ object CollectionStrawMan1 {
   implicit class ViewMonoTransforms[A](val v: View[A])
   extends AnyVal with MonoTransforms[A, View[A]] {
     protected def iter = v.iterator
-    protected def fromIter(it: => Iterator[A]): View[A] = new View(it)
+    protected def fromIter(it: => Iterator[A]): View[A] = it
   }
 
   implicit class ViewPolyTransforms[A](val v: View[A])
   extends AnyVal with PolyTransforms[A, View] {
     protected def iter = v.iterator
-    protected def fromIter[B](it: => Iterator[B]) = new View(it)
+    protected def fromIter[B](it: => Iterator[B]) = it
   }
 
   /** Concrete collection type: String */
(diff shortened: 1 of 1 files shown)


DarkDimius:
Major change in  48afa75: there is no class view. It's a type alias, and from this type alias there are decorators that add convenient operations for views.

DarkDimius:
As of  5281688 views are not part of collection operations. I believe that `view` method is fundamentally different from `foldRight\head\isEmpty` and should belong to a different implicit conversion.

DarkDimius:
As of b77ab7c, iterator class does not have operations.
Instead, operations are on views.

The motivating example is 

```scala
def average(xs: Iterator[Int]) = xs.sum/xs.size
```

I believe it simply should not compile. User should instead write 

```scala
def average(xs: View[Int]) = xs.sum/xs.size
```

Note that as of  48afa75 this `View[T]` is an alias to `() => Iterable[T]`, making this code actually do what it is expected to.

DarkDimius:
Major change in 48afa75: there is no class view. It's a type alias, and from this type alias there are decorators that add convenient operations for views.
Note that all existing ViewOps, ViewMonoTransforms and ViewPolyTransforms continue to work. This is a source-compatible change :-D

DarkDimius:
As of 5281688 views are not part of collection operations. I believe that view method is fundamentally different from foldRight\head\isEmpty and should belong to a different implicit conversion.

DarkDimius:
As of b77ab7c, iterator class does not have operations.
Instead, operations are on views.

The motivating example is

def average(xs: Iterator[Int]) = xs.sum/xs.size

I believe it simply should not compile. User should instead write

def average(xs: View[Int]) = xs.sum/xs.size

Note that as of 48afa75 this View[T] is an alias to () => Iterable[T], making this code actually do what it is expected to.

DarkDimius:
As an argument that Iterator should not have operations that consume it I'll remind about a bug in dotty that was in code for long time and was tricky to find:
https://github.com/dotty-staging/dotty/commit/03ec379926f1f900f09c50dd038fba86feae70f6

DarkDimius:
As of 5244c5c, there is a type alias `type LengthType = Long` which is used for all indexes. This means that `size` is Long.

DarkDimius:
As of  3c6ed1f23e2da86a4cb870bf91931b9bfc4925ab LengthType is a value class, that defines common arithmetic operations, and has implicit convertions to Long and from Int.
The underlying type could be different on JVM and JavaScript platform.

Given provided methods and implicit conversions, those code samples work:

```scala
val s: List = ...
s(1) // 1 is converted to LengthType through implicit convertion
s.drop(s.length - 1) // arithmetic operations on LengthType
assert(s.length < 100) // comparisons
```

odersky:
I don't think we need to keep this in the PR queue any longer.

PR #821, opened 2015-10-07
Check that some types are not higher-kinded.

odersky:
Invalidates #813. Review by @darkdimius.


odersky:
ping @darkdimius

DarkDimius:
LGTM

PR #823, opened 2015-10-09
Always fully define the types of lifted expressions

smarter:
Fixes #822 

Review by @odersky . I don't know if fixing this just in `EtaExpression#lift` is the correct thing to do, we could still accidentally create symbols with non-fully defined types. Should we instead call `fullyDefinedType(info)` inside `Symbols#newSymbol` ?


smarter:
@odersky : I tried adding a config option to check that `SymDenotations` are always fully defined (in a similar way than `Config.checkNoSkolemsInInfo`) but I run into cyclic reference errors, do you see a way to avoid that?

PR #824, opened 2015-10-13
[Bug] OuterClass.this seems typed incorrectly

DarkDimius:
It seems like`OuterClass.this` gets typed incorrectly:

```scala
// val self = Builder.this

Builder.scala:120: error: clear is not a member of Any(self)
      def clear() = self.clear()
                         ^
Builder.scala:121: error: ++ is not a member of Any(self)
      override def ++=(xs: TraversableOnce[Elem]): this.type = { self ++= xs; this }
                                                                 ^
Builder.scala:122: error: sizeHint is not a member of Any(self)
      override def sizeHint(size: Int) = self.sizeHint(size)
                                              ^
Builder.scala:123: error: sizeHintBounded is not a member of Any(self)
      override def sizeHintBounded(size: Int, boundColl: TraversableLike[_, _]) = self.sizeHintBounded(size, boundColl)
                                                                                       ^
Builder.scala:124: error: result is not a member of Any(self)
      def result: NewTo = f(self.result())
                                 ^
```

### DIFF
--- a/tests/pos/Builder.scala
+++ b/tests/pos/Builder.scala
@@ -1,0 +1,127 @@
+/*                     __                                               *\
+**     ________ ___   / /  ___     Scala API                            **
+**    / __/ __// _ | / /  / _ |    (c) 2003-2013, LAMP/EPFL             **
+**  __\ \/ /__/ __ |/ /__/ __ |    http://scala-lang.org/               **
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+
+
+package scala
+package collection
+package mutable
+
+import generic._
+
+/** The base trait of all builders.
+ *  A builder lets one construct a collection incrementally, by adding
+ *  elements to the builder with `+=` and then converting to the required
+ *  collection type with `result`.
+ *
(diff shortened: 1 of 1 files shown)


smarter:
The issue is that the we always infer the return type of an overridee to be the return type of the corresponding overriden member:
```scala
trait Hello {
  def hello(): AnyRef
}

class Foo extends Hello {
  def hello() = "" // scalac infers String, dotty infers AnyRef
}
```
scalac matches what the spec says: http://www.scala-lang.org/files/archive/spec/2.11/04-basic-declarations-and-definitions.html#method-return-type-inference
> A class member definition m that overrides some other function m′ in a base class of C may leave out the return type, even if it is recursive. In this case, the return type R′ of the overridden function m′, seen as a member of C, is taken as the return type of m for each recursive invocation of m. That way, a type R for the right-hand side of m can be determined, which is then taken as the return type of m. Note that R may be different from R′, as long as R conforms to R′.

I'll see if I can get dotty to match this behaviour.

odersky:
@smarter No, don't do that! The Dotty behavior is fully intentional and addresses a shortcoming of scalac.

 - Martin

PR #832, opened 2015-10-20
Fix #815.

### LINKED ISSUE #815: :: is not a class type
```scala
import scala.collection.immutable.::
object A {
     new ::(args, Nil)
  }
}
```

```
A.scala:4: error: :: is not a class type
     new ::(args, Nil)
         ^
one error found
```

odersky:
Need to handle hk-aliases specially in typedNew. Review by @DarkDimius


DarkDimius:
LGTM

PR #835, opened 2015-10-21
Add more tests

odersky:
Tests that have accumulated over work on previous commits. Collected here so they do not get lost. Review by @smarter.


smarter:
` tests/pending/neg/i827.scala` is empty, otherwise LGTM.

PR #837, opened 2015-10-21
Quick&dirty bootstrap

DarkDimius:
dotc script now has `-bootstrapped` option that will use dotty-compiled-by-dotty.

I has auto-detection of changes in dotty itself and will recompile itself using `dotc.build` on any change.


DarkDimius:
In order  to check that JVM uses classes compiled by dotty itself one could use command 
```dotc -bootstrapped -J-verbose:class```

Bootstrapped should be loaded from `dotty.jar`, while scalac compiled classes would load from `dotty_2.11-0.1-SNAPSHOT.jar`

odersky:
LGTM

PR #843, opened 2015-10-22
Fix #830: Compiler hangs on implicit search with singleton &/|

### LINKED ISSUE #830: Infinite compile time with type inference union-typed implicit
The following appears to send the compiler into an infinite loop, but only when the return type of `x` is used to infer the type parameter of the call to `y`:

```
object C {
  trait X[T]
  val x: 1 & 2 | 2 & 3 = y
}
```

My guess is that the interaction between the type inference of `y`'s parameter `T` and implicit search is underconstrained. The type expression (`1 & 2 | 2 & 3`) must contain at least one repeated type (here, `2`).

odersky:
In fact we get a deep subtype recursion when compileing i830.scala. The problem
goes away once we make use of the fact that the intersection of two singleton
types which are not subtypes of each other is empty. Review by @smarter


odersky:
@smarter How does it look now?

smarter:
The change LGTM but note that the following still causes an infinite loop:
```scala
object C {
  val a: Int = 1
  val b: Int = 2
  val c: Int = 2

  trait X[T]
}
```

PR #846, opened 2015-10-22
Fix #827

### LINKED ISSUE #827: Compiler crash extending inner class with same name from self-type
This code causes a compiler crash due to stack overflow:

```
trait A { trait Inner }
trait B { self: A =>
  trait Inner extends self.Inner
}
```

Given that mixing in two traits, each with inner traits of the same name is illegal in Dotty (not Scala), this isn't especially useful code, so maybe the error message should reflect that.

odersky:
Fixes two instances where we got infinite recursion in member searches. Review by @smarter


smarter:
LGTM.

PR #848, opened 2015-10-22
Run CI-partest with bootstrapped compiler

DarkDimius:
I'm doing dirty classpath tricks to make partest prefer dotty-compiled-dotty over scalac-compiled-dotty.


odersky:
Is this now superseded by #853?

DarkDimius:
Yes, it is.

PR #860, opened 2015-10-24
Change allow ex in hk

odersky:
Fixed problems with existentials in hk types. Review by @DarkDimius.


odersky:
Update: This now also solves a different problem: How to deal with Scala 2 inner classes which shadow each other (this is disallowed in dotty, but legal and in fact quite common in Scala 2). With the fix, GenTraversableFactory compiles as it is in stdlib.

odersky:
ping @DarkDimius

DarkDimius:
@odersky, I'm still reviewing the effect of this pr on typing the stdlib and linker. It changes the way how some types in stdlib are inferred and I'm making sure that those changes do not effect correctness.

odersky:
Can this go in now? I need this in order to make further progress on the stdlib.

DarkDimius:
LGTM

PR #862, opened 2015-10-24
Avoid completing classes during enter in some cases

### LINKED ISSUE #850: Failure to find class defined in an other compilation unit
Put the following in `A.scala`:
```scala
import B.hello

class A(private[this] var foo: Int)
```
And the following in `B.scala`:
```scala
object B {
  def hello: Int = 1
}
```
Then this fails:
```scala
% dotc A.scala B.scala
A.scala:1: error: not found: B
import B.hello
       ^
one error found
```
This does not happen if we replace the `var` by a `val`, or `private[this]` by `private` or nothing

smarter:
The `ctx.owner is Trait` check in `Desugar#valDef` forced the completion
of some classes during enter. Since `Trait` is always set upon symbol
creation the completion of the class is not needed, so we now avoid it.

Fixes #850

Review by @odersky .

### DIFF
--- a/src/dotty/tools/dotc/core/Flags.scala
+++ b/src/dotty/tools/dotc/core/Flags.scala
@@ -435,15 +435,20 @@ object Flags {
   /** Flags representing access rights */
   final val AccessFlags = Private | Protected | Local
 
-  /** Flags guaranteed to be set upon symbol creation */
-  final val FromStartFlags =
+  /** Common flags guaranteed to be set upon symbol creation */
+  final val FromStartCommonFlags =
     AccessFlags | Module | Package | Deferred | Final | MethodOrHKCommon | Param | ParamAccessor | Scala2ExistentialCommon |
     InSuperCall | Touched | JavaStatic | CovariantOrOuter | ContravariantOrLabel | ExpandedName | AccessorOrSealed |
     CaseAccessorOrTypeArgument | Fresh | Frozen | Erroneous | ImplicitCommon | Permanent |
     SelfNameOrImplClass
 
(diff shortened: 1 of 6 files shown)


smarter:
OK I give up, tests are randomly failing: https://scala-ci.typesafe.com/job/dotty-master-validate-partest/520/console https://scala-ci.typesafe.com/job/dotty-master-validate-partest/521/console https://scala-ci.typesafe.com/job/dotty-master-validate-partest/517/console
@DarkDimius : the first one fails with `AbstractMethodError` in the generated code

odersky:
That's one way to do it. But I think Lazy should also be a "from start" flag. I do not see a situation where it would be set only on completion. In that case, we can stick with the original scheme and make both Lazy and Trait from start flags. #860 contains a commit which does this.

smarter:
> But I think Lazy should also be a "from start" flag. I do not see a situation where it would be set only on completion.

Isn't that what happens in `ModuleCompleter`? https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/SymDenotations.scala#L1792
During `complete` we call `denot.setFlag(from.flags.toTermFlags & RetainedModuleValFlags)`, and `RetainedModuleValFlags` contains `Lazy`, so it's possible to set `Lazy` during completion.

odersky:
> Isn't that what happens in ModuleCompleter?

That does not follow. If lazy is a from start flag, it means that the module symbol gets it from the start. You'd have to prove that there is a case where (1) the module symbol does not have Lazy set, and (2) the module class symbol has Trait set. Neither of these can happen, I think.

odersky:
Superseded by #860

PR #865, opened 2015-10-25
Fix hole in subtyping of modules

odersky:
We did not handle correctly the case exemplified by `range.scala`: A module this type, which
is compared with a getter to the same module. Seen in the wild in scala.math.BigDecimal.

Review by @smarter


smarter:
LGTM.

PR #869, opened 2015-10-26
Fix #866

### LINKED ISSUE #866: Typer crash with unknown import and implicit parameter of unknown type
The following is sufficient to crash the typer:

```
import nonexistent._
object Foo {
  def bar(implicit x: NonExistent) = ???
  val baz = bar
}
```

An error is correctly emitted for the missing import, but not for the missing implicit type.

odersky:
Fixes #866 and also the problem that total counts of errors were sometimes more than the
number of error messages reported.


odersky:
Review by @smarter

smarter:
Otherwise LGTM.

PR #882, opened 2015-10-29
Set position on annotations parsed from Java

smarter:
Fix a crash in Typer#assertPositioned

Review by @odersky .


odersky:
LGTM

PR #885, opened 2015-10-29
Don't simplify |/& by taking lubs/glbs for explicitly declared types.

odersky:
A lub might return a different type (i.e. singleton types are widenened).
We should not do this if the type is given explicitly. 

This is a more conservative version of #844. Review by @smarter. 
@smarter can you push your tests on top of this PR? Thanks!

### DIFF
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -808,6 +808,15 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     def tpes: List[Type] = xs map (_.tpe)
   }
 
+  /** Simplify type of tree */
+  def simplifyType(tree: Tree)(implicit ctx: Context) = {
+    def isExplicitType(tree: Tree) = tree match {
+      case TypeTree(original) => !original.isEmpty
+      case _ => tree.isType
+    }
+    tree.overwriteType(new ctx.SimplifyMap(isExplicitType(tree))(tree.tpe))
+  }
+
   // convert a numeric with a toXXX method
   def primitiveConversion(tree: Tree, numericCls: Symbol)(implicit ctx: Context): Tree = {
     val mname      = ("to" + numericCls.name).toTermName
--- a/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/src/dotty/tools/dotc/core/TypeOps.scala
@@ -243,15 +243,17 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
       tp.derivedRefinedType(simplify(tp.parent, theMap), tp.refinedName, simplify(tp.refinedInfo, theMap))
     case tp: TypeAlias =>
       tp.derivedTypeAlias(simplify(tp.alias, theMap))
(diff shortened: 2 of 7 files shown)


smarter:
- With this patch, we still assign the type of `OrTypeTree` using `|`(https://github.com/dotty-staging/dotty/blob/change-simplify-types/src/dotty/tools/dotc/typer/TypeAssigner.scala#L369) which calls `lub`, so we're still losing singletons.
- Even if this is fixed, this patch does not help in situations like this:
```scala
val foo: a.type | b.type = a
val bar: a.type | b.type = foo
```
The type of the tree corresponding to `foo` in the right-hand side of `bar` is going to be adapted, and since the type is not written explicitly it will be simplified to `Int`, so we'll get a type mismatch.
- Since we call `Type#simplified` when instantiating tvars (https://github.com/dotty-staging/dotty/blob/change-simplify-types/src/dotty/tools/dotc/core/Types.scala#L2475), we can probably run into a situation where we instantiate a type variable in a way that result in a type mismatch, but I'm too jetlagged to think up an example right now.

odersky:
I think all this is correct. But what can we do? Disallow unions of
singletons in types?


On Fri, Oct 30, 2015 at 10:05 AM, Guillaume Martres <

>
>    - With this patch, we still assign the type of OrTypeTree using |(
>    https://github.com/dotty-staging/dotty/blob/change-simplify-types/src/dotty/tools/dotc/typer/TypeAssigner.scala#L369)
>    which calls lub, so we're still losing singletons.
>    - Even if this is fixed, this patch does not help in situations like
>    this:
>
> val foo: a.type | b.type = aval bar: a.type | b.type = foo
>
> The type of the tree corresponding to foo in the right-hand side of bar
> is going to be adapted, and since the type is not written explicitly it
> will be simplified to Int, so we'll get a type mismatch.
>
>    - Since we call Type#simplified when instantiating tvars (
>    https://github.com/dotty-staging/dotty/blob/change-simplify-types/src/dotty/tools/dotc/core/Types.scala#L2475),
>    we can probably run into a situation where we instantiate a type variable
>    in a way that result in a type mismatch, but I'm too jetlagged to think up
>    an example right now.
>
> —
> <https://github.com/lampepfl/dotty/pull/885#issuecomment-152465399>.
>



-- 
Martin Odersky
EPFL

smarter:
Well, #844 seems to work fine, do you see any cases where it might do something wrong?

odersky:
The danger with #844 is that it produces too large types and will make
inference very slow or even cause it to blow up. Hard to test for it.

On Fri, Oct 30, 2015 at 10:15 AM, Guillaume Martres <

> Well, #844 <https://github.com/lampepfl/dotty/pull/844> seems to work
> fine, do you see any cases where it might do something wrong?
>
> —
> <https://github.com/lampepfl/dotty/pull/885#issuecomment-152467164>.
>



-- 
Martin Odersky
EPFL

odersky:
I now believe we are on the wrong track altogether. Messing with the meaning of lubs gets us into trouble. If we want lubs to widen singleton types we have to disallow union types consisting of singleton types. Everything else would be too operational.

PR #886, opened 2015-10-30
Fix #884 - misdiagnosed ambiguous definition.

### LINKED ISSUE #868: Local package objects conflict with imported package objects
A wildcard import from a package with a package object defined makes it impossible to define a package object locally.

Both this:
```
package foo
import scala.collection._
package object bar
```
and
```
package foo.bar
import scala.collection._
object `package`
```
fail with the error:
```
pkg.scala:3: error: reference to package is ambiguous;
it is both imported by import collection._
and defined subsequently in package bar
object `package`
                ^
one error found
```

### LINKED ISSUE #884: Conflicts when defining package object and other package object is imported
This currently fails:
```scala
import scala.reflect._

object `package` {
}
```
```scala
error: reference to package is ambiguous;
it is both imported by import reflect._
and defined subsequently in package <empty>
object `package` {
                 ^
```

odersky:
Universal equality strikes again. 
Caused a bug in isDefinedInCurrentUnit.

Also fixes #868. Review by @smarter.


smarter:
/rebuild (this should be called /russianroulettebuild really)

smarter:
LGTM.

PR #887, opened 2015-10-30
Allow pattern matching anonymous functions of arity > 1

### LINKED ISSUE #873: "case" function literal incorrectly desugared
```scala
object Test {
  def call(k: (Int, Int) => Unit): Unit = ???
  def test = call({ case (x, y) => ()})
}
```
This fails with:
```scala
funlit.scala:9: error: wrong number of parameters, expected: 2
  def test = call({ case (x, y) => ()})
                  ^
result of funlit.scala after frontend:
package <empty> {
    def call(k: Function2[Int, Int, Unit]): Unit = ???
    def test: Unit = 
      Test.call({
        def $anonfun(x$1: <error>): Unit = 
          x$1 match {
            case Tuple2.unapply[Nothing', Nothing'](x @ _, y @ _) => 
              ()
          }
        closure($anonfun)
      })
  }
}
two errors found
```
Because `{ case (x, y) => ()}` is expanded to `x$1 => x$1 match { case (x, y) => ()}`, instead it should be expanded to `(x$1, x$2) => (x$1, x$2) match { case (x, y) => ()}` (like in Scala 2) because this compiles fine in Dotty:
```scala
object Test {
  def call(k: (Int, Int) => Unit): Unit = ???
  def test = call((a, b) => (a, b) match { case (x, y) => ()})
}
```

odersky:
This is specified in Sec. 8.5 of the SLS. Fixes #873. Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -556,14 +556,20 @@ object desugar {
       DefDef(nme.ANON_FUN, Nil, params :: Nil, tpt, body).withMods(synthetic),
       Closure(Nil, Ident(nme.ANON_FUN), EmptyTree))
 
-  /** Expand partial function
+  /** If `nparams` == 1, expand partial function
+   *
    *       { cases }
    *  ==>
-   *       x$0 => x$0 match { cases }
+   *       x$1 => x$1 match { cases }
+   *
+   *  If `nparams` != 1, expand instead to
+   *
+   *       (x$0, ..., x${n-1}) => (x$0, ..., x${n-1}) match { cases }
    */
-  def makeCaseLambda(cases: List[CaseDef])(implicit ctx: Context) = {
-    val param = makeSyntheticParameter()
-    Function(param :: Nil, Match(Ident(param.name), cases))
+  def makeCaseLambda(cases: List[CaseDef], nparams: Int = 1)(implicit ctx: Context) = {
+    val params = (1 to nparams).toList.map(makeSyntheticParameter(_))
+    val selector = makeTuple(params.map(p => Ident(p.name)))
+    Function(params, Match(selector, cases))
   }
 
   /** Add annotation with class `cls` to tree:
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -562,12 +562,12 @@ object desugar {
    *  ==>
    *       x$1 => x$1 match { cases }
    *
-   *  If `nparams` > 1, expand instead to
+   *  If `nparams` != 1, expand instead to
    *
    *       (x$1, ..., x$n) => (x$0, ..., x${n-1}) match { cases }
    */
   def makeCaseLambda(cases: List[CaseDef], nparams: Int = 1)(implicit ctx: Context) = {
-    val params = (1 to (nparams min 1)).toList.map(makeSyntheticParameter(_))
+    val params = (1 to nparams).toList.map(makeSyntheticParameter(_))
     val selector = makeTuple(params.map(p => Ident(p.name)))
     Function(params, Match(selector, cases))
   }
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -562,12 +562,12 @@ object desugar {
    *  ==>
    *       x$1 => x$1 match { cases }
    *
-   *  If `nparams` != 1, expand instead to
+   *  If `nparams` > 1, expand instead to
    *
-   *       (x$0, ..., x${n-1}) => (x$0, ..., x${n-1}) match { cases }
+   *       (x$1, ..., x$n) => (x$0, ..., x${n-1}) match { cases }
    */
   def makeCaseLambda(cases: List[CaseDef], nparams: Int = 1)(implicit ctx: Context) = {
-    val params = (1 to nparams).toList.map(makeSyntheticParameter(_))
+    val params = (1 to (nparams min 1)).toList.map(makeSyntheticParameter(_))
     val selector = makeTuple(params.map(p => Ident(p.name)))
     Function(params, Match(selector, cases))
   }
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -556,14 +556,20 @@ object desugar {
...
+  /** If `nparams` == 1, expand partial function
+   *
    *       { cases }
    *  ==>
-   *       x$0 => x$0 match { cases }
+   *       x$1 => x$1 match { cases }
+   *
+   *  If `nparams` != 1, expand instead to
+   *
+   *       (x$0, ..., x${n-1}) => (x$0, ..., x${n-1}) match { cases }
```
smarter: Minor nit: it should be `(x$1, ..., x$n)`, not `(x$0, ...`

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -556,14 +556,20 @@ object desugar {
...
 
-  /** Expand partial function
+  /** If `nparams` == 1, expand partial function
+   *
    *       { cases }
    *  ==>
-   *       x$0 => x$0 match { cases }
+   *       x$1 => x$1 match { cases }
+   *
+   *  If `nparams` > 1, expand instead to
```
retronym: scalac also allows pattern matching anonymous functions with an expected type of `Function0`:

```
% cat sandbox/test.scala
class Test {
  ({case _ if "".isEmpty => 0} : (() => Any)).apply()
}

%  qscalac -Xprint:typer sandbox/test.scala
[[syntax trees at end of                     typer]] // test.scala
package <empty> {
  class Test extends scala.AnyRef {
    def <init>(): Test = {
      Test.super.<init>();
      ()
    };
    ((() => () match {
  case _ if "".isEmpty() => 0
}): () => Any).apply()
  }
}
```

retronym: That might fall out of this implementation too, but the doc needs an update.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -629,7 +629,11 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   def typedMatch(tree: untpd.Match, pt: Type)(implicit ctx: Context) = track("typedMatch") {
     tree.selector match {
       case EmptyTree =>
-        typed(desugar.makeCaseLambda(tree.cases) withPos tree.pos, pt)
+        val arity = pt match {
+          case defn.FunctionType(args, _) => args.length
+          case _ => 1
+        }
```
retronym: Recent versions of scalac (2.10+?) are more lenient, the expected type of a patmat anon function need only have a `FunctionN` with fully defined argument types as base type.

```
class Test {
  trait X extends Function1[Int, String]
  implicit def f2x(f: Function1[Int, String]): X = ???
  ({case _ if "".isEmpty => ""} : X) // allowed, implicit view used to adapt

  ({case _ if "".isEmpty => 0} : X) // expected String, found Int
}
```

This change was motivated by spores.

odersky:
@retronym 

> need only have a FunctionN with fully defined argument types as base type.

This looks fishy. Does that not conflict with the concept of a SAM-type? What if I have

  trait Sammy extends (Int => Int) {
     def apply(x: Int, y: Int): Int
     def apply(x: Int) = apply(x, x)
  }

Does Sammy have one or two parameters?

odersky:
@retronym I added support for SAM functions when adapting the arity of case lambdas. With that your test compiles as expected.

smarter:
LGTM but I'd like @retronym to give his LGTM  too :).

retronym:
> This looks fishy. Does that not conflict with the concept of a SAM-type?

I actually argued against that change. The comment thread of https://github.com/scala/scala/pull/2650 is worth a read to discuss the interactions with SAMs and the quality of error reporting.

tests/pos/i873.scala
```diff
@@ -0,0 +1,10 @@
+object Test {
+  def call(k: (Int, Int) => Unit): Unit = ???
+  def test = call({ case (x, y) => ()})
+
+  trait X extends Function1[Int, String]
+  implicit def f2x(f: Function1[Int, String]): X = ???
+  ({case _ if "".isEmpty => ""} : X) // allowed, implicit view used to adapt
```
retronym: I guess this comment is not applicable to dotty. It might be worth leaving a comment that contrasts the approaches, though.

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -556,14 +556,20 @@ object desugar {
...
+   *  If `nparams` != 1, expand instead to
+   *
+   *       (x$1, ..., x$n) => (x$0, ..., x${n-1}) match { cases }
    */
-  def makeCaseLambda(cases: List[CaseDef])(implicit ctx: Context) = {
-    val param = makeSyntheticParameter()
-    Function(param :: Nil, Match(Ident(param.name), cases))
+  def makeCaseLambda(cases: List[CaseDef], nparams: Int = 1)(implicit ctx: Context) = {
+    val params = (1 to nparams).toList.map(makeSyntheticParameter(_))
+    val selector = makeTuple(params.map(p => Ident(p.name)))
```
retronym: Probably need to report an error if `n > MaxTupleArity`, which might result from a SAM method type.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -502,6 +502,16 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     assignType(cpy.If(tree)(cond1, thenp2, elsep2), thenp2, elsep2)
   }
 
+  private def decomposeProtoFunction(pt: Type, defaultArity: Int)(implicit ctx: Context): (List[Type], Type) = pt match {
+    case _ if defn.isFunctionType(pt) =>
+      (pt.dealias.argInfos.init, pt.dealias.argInfos.last)
+    case SAMType(meth) =>
+      val mt @ MethodType(_, paramTypes) = meth.info
+      (paramTypes, mt.resultType)
```
retronym: Do you need to do anything with dependent method types here?

I just looked and I see that the `SAMType` extractor excludes them.

A few more tests to show what is and isn't supported would be welcome.

src/dotty/tools/dotc/ast/Desugar.scala
odersky: @retronym We don't intend to keep MaxTupleArity for much longer anyway.

odersky:
I note the remarks but don't have the time to implement them right now. Does anyone else want to step up? If not I'll merge by tomorrow.

PR #889, opened 2015-10-30
Fix #877

### LINKED ISSUE #877: Nested defs with context bounds are desugared to nested defs with implicit parameters that have the same name
```scala
class First[A]
class Second[A]

class Foo {
  def foo[A: First] = {
    def bar[B: Second] = {
      val fst: First[A] = implicitly[First[A]]
      val snd: Second[B] = implicitly[Second[B]]
    }
  }
}
```
is desugared to:
```scala
package <empty> {
  class First[A]() extends Object() { 
    type First$$A
    private[this] type A = First$$A
  }
  class Second[A]() extends Object() { 
    type Second$$A
    private[this] type A = Second$$A
  }
  class Foo() extends Object() { 
    def foo[A](implicit evidence$0: First[A]): Unit = {
      def bar[B](implicit evidence$0: Second[B]): Unit = {
        val fst: First[A] = implicitly[First[A]](evidence$0)
        val snd: Second[B] = implicitly[Second[B]](evidence$0)
        ()
      }
      ()
    }
  }
}
```
Which fails to compile because `evidence$0` in `bar` shadows `evidence$0` in `foo`, scalac avoids this by using different names for the desugared implicit parameters(in fact, it seems to never reuse the same name in a given compilation unit, even in different classes):
```scala
    def foo[A](implicit evidence$1: First[A]): Unit = {
      def bar[B](implicit evidence$2: Second[B]): Unit = {
        val fst: First[A] = scala.this.Predef.implicitly[First[A]](evidence$1);
        val snd: Second[B] = scala.this.Predef.implicitly[Second[B]](evidence$2);
        ()
      };
      ()
    }
```

odersky:
Review by @smarter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 # sbt specific
 dist/*
 target/
-out/
 lib_managed/
 src_managed/
 project/boot/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # sbt specific
 dist/*
 target/
+out/
 lib_managed/
 src_managed/
 project/boot/
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -9,11 +9,10 @@ import Scopes._
 import typer.{FrontEnd, Typer, Mode, ImportInfo, RefChecks}
 import reporting.{Reporter, ConsoleReporter}
 import Phases.Phase
-import dotty.tools.dotc.transform._
-import dotty.tools.dotc.transform.TreeTransforms.{TreeTransform, TreeTransformer}
-import dotty.tools.dotc.core.DenotTransformers.DenotTransformer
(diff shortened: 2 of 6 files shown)


odersky:
I have no idea why the latest commit made the build pass. Normally, a freshNameCreator per ContextBase (which was the situation before the commit) should be fully isolated between different parallel compiles. But parallel compiles were failing randomly, with random symptoms.

smarter:
okay, change LGTM but I don't know how the parallel compiles work at all so can't help here.

odersky:
@DarkDimius Can you confirm that out/ should be added to .gitignore?

DarkDimius:
@odersky: https://github.com/lampepfl/dotty/pull/896 adds `out` contents to `.gitignore` but maintains `.keep` file.

odersky:
Reverted the gitignore commit, kept the rest.

PR #890, opened 2015-10-30
Fix #879

### LINKED ISSUE #879: Error when typechecking Java class where the superclass constructor has arguments
```java
class Foo {
  Foo(int i) {
  }
}


class Bar extends Foo {
  Bar() {
    super(10);
  }
}
```
```scala
cons.java:7: error: missing argument for parameter i of constructor Foo: (i: Int)Foo
class Bar extends Foo {
                  ^
```
(minimized from https://github.com/molecule-labs/molecule/blob/master/molecule-core/src/main/java/molecule/jsr166y/ForkJoinTask.java#L397)

CC @olhotak

odersky:
Don't insert a constructor call when typechecking Java classes. Review by @olhotak or @smarter.


smarter:
LGTM

PR #892, opened 2015-10-30
Avoid printing expanded names when displaying type parameters.

odersky:
Differences can be observed when checking neg/variances.scala.
Review and any necessary polishing by @smarter.


smarter:
LGTM

PR #893, opened 2015-10-31
Demonstrate race in partest.

odersky:
Can be used to test that a future fix is effective. This PR should not be merged into master.

### DIFF
--- a/src/dotty/tools/dotc/core/Contexts.scala
+++ b/src/dotty/tools/dotc/core/Contexts.scala
@@ -615,11 +615,22 @@ object Contexts {
 
   object Context {
 
+    // DO NOT MERGE INTO MASTER; this version is only intended
+    // to demonstrate that partest re-uses the same context base
+    // over multiple threads, which causes all sorts of races.
+    // Each thread in partest needs to start with its own context base.
+    @dotty.tools.sharable
+    private var thread: Thread = null
+
     /** Implicit conversion that injects all printer operations into a context */
     implicit def toPrinter(ctx: Context): Printer = ctx.printer
 
     /** implicit conversion that injects all ContextBase members into a context */
-    implicit def toBase(ctx: Context): ContextBase = ctx.base
+    implicit def toBase(ctx: Context): ContextBase = {
+      if (thread == null) thread = Thread.currentThread()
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -615,11 +615,22 @@ object Contexts {
...
+    private var thread: Thread = null
+
     /** Implicit conversion that injects all printer operations into a context */
     implicit def toPrinter(ctx: Context): Printer = ctx.printer
 
     /** implicit conversion that injects all ContextBase members into a context */
-    implicit def toBase(ctx: Context): ContextBase = ctx.base
+    implicit def toBase(ctx: Context): ContextBase = {
+      if (thread == null) thread = Thread.currentThread()
+      else assert(thread == Thread.currentThread(), "illegal multithreaded access to ContextBase")
```
DarkDimius: I'm not shure what is this test intended to demonstrate.
Currently it fails if more than 1 thread ever calls this method even if they use different `ContextBases`

DarkDimius: It currently died inside this code:

https://github.com/lampepfl/dotty/blob/master/test/dotty/partest/DPDirectCompiler.scala#L17-L19
```scala

val base = new dotty.tools.dotc.core.Contexts.ContextBase
      import base.settings._
      val ctx = base.initialCtx.fresh // inside fresh
```

```
java.lang.AssertionError: assertion failed: illegal multithreaded access to ContextBase
[error] 	at scala.Predef$.assert(Predef.scala:165)
[error] 	at dotty.tools.dotc.core.Contexts$Context$.toBase(Contexts.scala:631)
[error] 	at dotty.tools.dotc.core.Contexts$Context.setCreationTrace(Contexts.scala:244)
[error] 	at dotty.tools.dotc.core.Contexts$Context.init(Contexts.scala:380)
[error] 	at dotty.tools.dotc.core.Contexts$Context.fresh(Contexts.scala:385)
[error] 	at dotty.partest.DPDirectCompiler.compile(DPDirectCompiler.scala:19)
[error] 	at scala.tools.partest.nest.Runner.attemptCompile(Runner.scala:421)
```

odersky:
False alarm. I have pushed a more refined PR, which partest passes, at least locally.

PR #894, opened 2015-10-31
Check that access to context base is singlethreaded.

odersky:
ContextBase is not intended to be threadsafe, We now test that
indeed it is not shared by compileUnits calls operating on
different threads. Review by @DarkDimius.


DarkDimius:
> out/.keep

this file shouldn't be deleted. It will break tests in CI and new clones of the repo.

odersky:
But sbt deletes .keep. Or it flips it around all the time! Can you push a .gitignore which makes that problem go away?

odersky:
Force pushed a version that keeps out/.keep.

DarkDimius:
LGTM

PR #900, opened 2015-11-01
Deal gracefully with missing return types of abstract menthods.

### LINKED ISSUE #871: Confusing error message for abstract method without return type
The following is legal in `scalac` and creates a method with return type `Unit`:
```scala
trait Message {
  def first(x: Int)
}
```
In Dotty, it's illegal, I agree with that, but the error message should tell you that you're missing a return type, instead you get:
```scala
confusing.scala:3: error: '=' expected but '}' found.
 }
 ^
one error found
```
which is hard to diagnose.

odersky:
1) Better error message: "missing return type" instead of `=' expected.
2) Allow them under language:Scala2

Fixes #871. Review by @smarter


smarter:
LGTM.

PR #902, opened 2015-11-01
Fix numeric implicit args

odersky:
Fix problems encountered when trying to compile math.BigInt and BigDecimal. Review by @smarter or @retronym

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -606,7 +606,7 @@ object Types {
     }
 
     /** Is this type a primitive value type which can be widened to the primitive value type `that`? */
-    def isValueSubType(that: Type)(implicit ctx: Context) = widenExpr match {
+    def isValueSubType(that: Type)(implicit ctx: Context) = widen match {
       case self: TypeRef if defn.ScalaValueClasses contains self.symbol =>
         that.widenExpr match {
           case that: TypeRef if defn.ScalaValueClasses contains that.symbol =>
@@ -617,6 +617,9 @@ object Types {
       case _ =>
         false
     }
+
+    def relaxed_<:<(that: Type)(implicit ctx: Context) =
+      (this <:< that) || (this isValueSubType that)
 
     /** Is this type a legal type for a member that overrides another
      *  member of type `that`? This is the same as `<:<`, except that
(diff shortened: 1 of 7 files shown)


smarter:
So, this PR allows the following code to compile:
```scala
class Foo
object Test {
  implicit def fooToShort(x: Foo): Short = 1
  val x: Int = new Foo
}
```
This was not allowed in Scala so this is a language change, should we write an SIP for it? It seems like a useful feature to me, but it complicates the rules for implicit resolution and the implementation (if I'm not mistaken, the `numericValueTieBreak` function wouldn't be needed if we didn't allow this) and it makes the numeric types even more "magic". LGTM, but it'd be great to hear from @retronym or someone else with experience dealing with numeric types.

odersky:
@smarter Interesting example. Scala 2 has the concept of "weak conformance" which handles some of the things this PR addresses. We made a decision not to go with weak conformance because it is complicated. But without this PR, the standard library does not compile (for the reasons given in the commits). 

It seems to me the Dotty proposal is more local and less sweeping than weak conformance. It only applies to implicit search. In essence, it means we allow some implicit conversions (namely those between numeric types) to be composed with other conversions in implicit search. So this is an exception to the rule that generally implicit search is not transitive. And we need to clean up the fallout of possible ambiguities by a tie-break which essentially says the least distance covered by those additional numeric conversions the better.

odersky:
No additional arguments, so I am merging this for now. We should revisit the issue of weak conformance in the future.

propensive:
Would a small set of implicits like this one help?
```
implicit def shortToInt[T](t: T)(implicit fn: T => Short): Int = fn(t)
```
This one permits @smarter's example to compile on Scala 2.

odersky:
No because we run into the transitivity issue.

On Mon, Nov 9, 2015 at 10:07 AM, Jon Pretty <notifications@github.com>
wrote:

> Would a small set of implicits like this one help?
>
> implicit def shortToInt[T](t: T)(implicit fn: T => Short): Int = fn(t)
>
> This one permits @smarter <https://github.com/smarter>'s example to
> compile on Scala 2.
>
> —
> <https://github.com/lampepfl/dotty/pull/902#issuecomment-155001593>.
>



-- 
Martin Odersky
EPFL

PR #904, opened 2015-11-02
Fix nested pkg member

odersky:
One more fix to compile stdlib. Review by @smarter.


odersky:
Let's try this on a different branch.

PR #908, opened 2015-11-02
Parentheses around a wildcard should not produce a lambda

### LINKED ISSUE #903: Placeholder syntax for anonymous functions chooses too narrow a scope for `_: T`
This program compiles with scalac:

```scala
object Test {
  def contains(s: String, i: Int) = true
  def test1 = {
    val f = contains("", (_: Int)) // dotc:
    f.apply(0)
    //     sandbox/eta.scala:4: error: type mismatch:
    //  found   : Int => Int
    //  required: Int
    //     val f = contains("", (_: Int))
    //                          ^
    // sandbox/eta.scala:5: error: apply is not a member of Boolean(f)
    //     f.apply(0)
    //      ^
  }

  def test2 = {
    val f = "".contains("", (_: Int)) // dotc:
    f.apply(0)
    // sandbox/eta.scala:18: error: apply is not a member of Boolean(f)
    //     f.apply(0)
    //       ^
  }
}

```

But with dotty, if tails with the compile errors inlined in comments above. Here's the post-typer AST:

```scala
package <empty> {
    def contains(s: String, i: Int): Boolean = true
    def test1: <error> = {
      val f: Boolean =
        Test.contains("", {
          def $anonfun(_$1: Int): Int = _$1: Int
          closure($anonfun)
        })
      f.apply(0)
    }
    def test2: <error> = {
      val f: Boolean =
        augmentString("").contains[Any'](
          Tuple2.apply[String', (Int => Int)']("", {
            def $anonfun(_$2: Int): Int = _$2: Int
            closure($anonfun)
          })
        )
      f.apply(0)
    }
  }
}
```

Tested with edfb6e7

odersky:
`(_)` and `(_: T)` should not be converted to functions

    x => x
    (x: T) => x

Fixes #903. Review by @retronym?

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -322,6 +322,7 @@ object Parsers {
       case Ident(name1) => placeholderParams.nonEmpty && name1 == placeholderParams.head.name
       case Typed(t1, _) => isWildcard(t1)
       case Annotated(t1, _) => isWildcard(t1)
+      case Parens(t1) => isWildcard(t1)
       case _ => false
     }
 
--- a/tests/pos/i903.scala
+++ b/tests/pos/i903.scala
@@ -1,0 +1,24 @@
+object Test {
+  def contains(s: String, i: Int) = true
+  def test1 = {
+    val f = contains("", (_: Int))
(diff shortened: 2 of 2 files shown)


retronym:
Should also test the `_` case (no type), but otherwise LGTM

PR #917, opened 2015-11-02
Compile big parts of stdlib

DarkDimius:
Introduces a test that compiles whitelisted parts of standard library. 
Test reads the `test/dotc/scala-collections.whitelist` file, allowing more-or-less convenient syntax for it: 
  - `#` can be used for comments;
  - indentation can be used;
  - empty lines can be used.

This implementation assumes this workflow: 
  - If some part of scala standard library compiles with dotty we add it to  `test/dotc/scala-collections.whitelist `
  - If scala standard library needs to be slightly modified (such as https://github.com/DarkDimius/scala/commit/b90c79beff80ae7477187c48ba5dcc0692fe6188) pull requests need to be done against this branch: https://github.com/darkdimius/scala/tree/dotty-library

The workflow is based on idea that we want to iterate fast and not wait for scala/scala reviews.

@odersky please review.


DarkDimius:
Junit tests succeeded. The changes in this PR only affect junit tests.

Now let's play Russian roulette. 
/rebuild

odersky:
LGTM

PR #921, opened 2015-11-04
Add stable hash codes to PolyTypes and PolyParams.

DarkDimius:
review by @odersky.

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -33,6 +33,7 @@ import annotation.tailrec
 import Flags.FlagSet
 import typer.Mode
 import language.implicitConversions
+import scala.util.hashing.{ MurmurHash3 => hashing }
 
 object Types {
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2358,7 +2374,7 @@ object Types {
     // no customized hashCode/equals needed because cycle is broken in PolyType
     override def toString = s"PolyParam(${binder.paramNames(paramNum)})"
 
-    override def computeHash = doHash(paramNum, binder)
+    override def computeHash = finishHash(hashing.mix(hashing.mix(hashSeed, paramNum), binder.identityHash), 1)
```
odersky: For consistency, I'd either use one of the existing doHash methods or create a new one that fits. Hashmin.mix is not called directly in Types, and we should keep it that way.

odersky:
Otherwise LTGM

Issue #881, opened 2015-10-28
links between companions are missing after pickling/unpickling


smarter:
`A_1.scala`:
```scala
class A
object A {
  implicit val theA: A = new A
}
```
`B_2.scala`:
```scala
class B {
  def getA(implicit a: A): A = a
  def test = {
    getA
  }
}
```
If you compile `A_1.scala`, then try to compile `B_2.scala`, it will fail because the implicit was not found, this happens because when we call `companionModule` on `A` in `collectCompanions`, we get back `NoSymbol`, because `COMPANION_MODULE_METHOD` is missing from `A`.
@DarkDimius : Should we pickle `COMPANION_MODULE_METHOD` and `COMPANION_CLASS_METHOD`, or should we recreate them after unpickling?

DarkDimius:
> Should we pickle COMPANION_MODULE_METHOD and COMPANION_CLASS_METHOD, or should we recreate them after unpickling?

If they would be pickled, they shouldn't be upicked into trees.
I would say it's better to recreate them during unpickling, as the information available there should be enough.

odersky:
I also think it's better to recreate.

odersky:
I don't see @VladimirNik show up in the gitter

PR #922, opened 2015-11-04
Links between companions after unpickling

VladimirNik:
Fix for the problem with missing links between companions during unpickling:
https://github.com/lampepfl/dotty/issues/881

Review by @DarkDimius @odersky


src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -588,9 +588,19 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
           sym.info = readType()
           ValDef(sym.asTerm, readRhs(localCtx))
         case TYPEDEF | TYPEPARAM =>
-          if (sym.isClass)
+          def addCompanionMethod(target: Symbol, owner: Symbol) = {
```
DarkDimius: This method is equivalent to `registerCompanionPair` in scala2Unpickler. @VladimirNik, would you consider removing code duplication?

DarkDimius:
Otherwise LGTM.

VladimirNik:
@DarkDimius these methods are not equivalent. It seems that `registerCompanionPair` in scala2Unpickler duplicates companion methods:
```scala
module.unforcedDecls: Scopes(val FLAG_SERIALIZABLE, val FLAG_MARKERS, val FLAG_BRIDGES, method metafactory, method altMetafactory, method <clinit>, method companion$class, method companion$class, method companion$class)
claz.unforcedDecls: Scopes(method <init>, method companion$module, method companion$module, method companion$module)
```
Maybe it's better to add fixes for this issue in the separate pull request?

VladimirNik:
This pull request should be rebased on https://github.com/lampepfl/dotty/pull/956/files

PR #923, opened 2015-11-04
Reorder unpickled type params if necessary

odersky:
When compiling Iterator.scala it was observed that
the type parameters of BufferedCanBuildFrom appeared
in the wrong order. This fix corrects that, making
sure that type parameters appear in the decls scope
in the same order as they are given in the explicitly
unpickled type parameter list.

Review by @VladimirNik (to change reviewer jobs a bit)


VladimirNik:
LGTM

PR #926, opened 2015-11-05
Handle Scalac variance unsoundness with regards to constructors.

odersky:
See SI-9549 for a summary. Review by @retronym

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -500,7 +500,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     projection.name == tpnme.hkApply &&
     !other.isHKApply &&
     other.testLifted(projection.prefix.LambdaClass(forcing = true).typeParams,
-      if (inOrder) isSubType(projection.prefix, _) else isSubType(_, projection.prefix),
+        { xx => println(i"test lifted with $xx")
+          if (inOrder) isSubType(projection.prefix, xx) else isSubType(xx, projection.prefix)
+        },
       if (inOrder) Nil else classBounds(projection.prefix))
 
   /** The class symbols bounding the type of the `Apply` member of `tp` */
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -500,9 +500,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     projection.name == tpnme.hkApply &&
     !other.isHKApply &&
     other.testLifted(projection.prefix.LambdaClass(forcing = true).typeParams,
-        { xx => println(i"test lifted with $xx")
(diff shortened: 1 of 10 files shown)


src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -500,7 +500,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     projection.name == tpnme.hkApply &&
     !other.isHKApply &&
     other.testLifted(projection.prefix.LambdaClass(forcing = true).typeParams,
-      if (inOrder) isSubType(projection.prefix, _) else isSubType(_, projection.prefix),
+        { xx => println(i"test lifted with $xx")
```
DarkDimius: `println` left in code.

retronym:
LGTM.

I checked out this branch to test that:

```scala
class C[+A] { class Inner { def this(a: A) = this() } } // only allowed under the migration
```

I do see an seemingly spurious warning about procedure syntax, is that a known issue?

```
⚡ ./bin/dotc -migration -language:Scala2 sandbox/test.scala
sandbox/test.scala:1: warning: Procedure syntax no longer supported; `=' should be inserted here
class C[+A] { class Inner { def this(a: A) = this() } }
                                           ^
sandbox/test.scala:1: warning: According to new variance rules, this is no longer accepted; need to annotate with @uncheckedVariance:
covariant type A occurs in contravariant position in type A of value a
class C[+A] { class Inner { def this(a: A) = this() } }
                                       ^
two warnings found
```

This version is allowed:

```
class C[+A] { def this(a: A) = this() }
```

During my testing, I also noticed another difference wrt scalac:

```scala
class C[A] { class A } // sandbox/test.scala:1: error: A is already defined as type A
```

odersky:
The procedure syntax warning seems to be a bug. I'll follow up on this.


On Fri, Nov 6, 2015 at 10:07 PM, Jason Zaugg <notifications@github.com>
wrote:

> LGTM.
>
> I checked out this branch to test that:
>
> class C[+A] { class Inner { def this(a: A) = this() } } // only allowed under the migration
>
> I do see an seemingly spurious warning about procedure syntax, is that a
> known issue?
>
> ⚡ ./bin/dotc -migration -language:Scala2 sandbox/test.scala
> sandbox/test.scala:1: warning: Procedure syntax no longer supported; `=' should be inserted here
> class C[+A] { class Inner { def this(a: A) = this() } }
>                                            ^
> sandbox/test.scala:1: warning: According to new variance rules, this is no longer accepted; need to annotate with @uncheckedVariance:
> covariant type A occurs in contravariant position in type A of value a
> class C[+A] { class Inner { def this(a: A) = this() } }
>                                        ^
> two warnings found
>
> This version is allowed:
>
> class C[+A] { def this(a: A) = this() }
>
> During my testing, I also noticed another difference wrt scalac:
>
> class C[A] { class A } // sandbox/test.scala:1: error: A is already defined as type A
>
> —
> <https://github.com/lampepfl/dotty/pull/926#issuecomment-154540889>.
>



-- 
Martin Odersky
EPFL

PR #927, opened 2015-11-05
Drop checking that lower bound is a subtype of upper bound.

odersky:
Review by @smarter.


smarter:
Surprisingly, this opens a soudness hole, we must be missing some use-site checks:
```scala
object Test {
  def g[B >: String <: Int](x: B): Int = x
    g("foo")
  }
}
```
```
[...]
```

PR #944, opened 2015-11-09
Tailrec: fix compilation of IndexedSeqOptimized

### LINKED ISSUE #912: IndexedSeqOptimized: tailrec bugs
```
/scala-scala/src/library/scala/collection/IndexedSeqOptimized.scala:58: error: Cannot rewrite recursive call: it changes type of 'this' on a polymorphic recursive call
    if (start == end) z
    ^
./scala-scala/src/library/scala/collection/IndexedSeqOptimized.scala:57: error: TailRec optimisation not applicable, method not tail recursive
  private def foldl[B](start: Int, end: Int, z: B, op: (B, A) => B): B =
              ^
./scala-scala/src/library/scala/collection/IndexedSeqOptimized.scala:63: error: Cannot rewrite recursive call: it changes type of 'this' on a polymorphic recursive call
    if (start == end) z
    ^
./scala-scala/src/library/scala/collection/IndexedSeqOptimized.scala:62: error: TailRec optimisation not applicable, method not tail recursive
  private def foldr[B](start: Int, end: Int, z: B, op: (A, B) => B): B =
              ^
```

DarkDimius:
Fixes #912

### DIFF
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -52,6 +52,7 @@ class Compiler {
       List(new PatternMatcher,
            new ExplicitOuter,
            new ExplicitSelf,
+           new CrossCastAnd,
            new Splitter),
       List(new VCInlineMethods,
            new SeqLiterals,
--- a/src/dotty/tools/dotc/transform/CrossCastAnd.scala
+++ b/src/dotty/tools/dotc/transform/CrossCastAnd.scala
@@ -20,29 +20,11 @@ class CrossCastAnd extends MiniPhaseTransform { thisTransform =>
   override def phaseName: String = "crossCast"
 
   override def transformSelect(tree: tpd.Select)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
-    if (tree.symbol.is(Flags.Private)) {
-      // all private member selections have a symbol
-      tree.qualifier.tpe.widen match {
-        case t @ AndType(l, _) =>
 
-        // find a component of and type that owns the symbol
-        def findType(tp: Type): Type = {
-            tp match {
-              case AndType(l, r) =>
-                findType(l).orElse(findType(r))
-              case t =>
-                if (t.decl(tree.symbol.name).suchThat(_ == tree.symbol).exists)
-                  t
-                else NoType
-            }
-          }
-
-          val tp = findType(t)
-          if (l eq tp) tree
-          else tree.qualifier.asInstance(AndType(tp, tree.qualifier.tpe)).select(tree.symbol)
-        case _ => tree
-      }
-    }
-      else tree
+    lazy val qtype = tree.qualifier.tpe.widen
+    val sym = tree.symbol
+    if (sym.is(Flags.Private) && qtype.typeSymbol != sym.owner)
+      cpy.Select(tree)(tree.qualifier.asInstance(AndType(qtype.baseTypeWithArgs(sym.owner), tree.qualifier.tpe)), tree.name)
+    else tree
   }
 }
--- a/src/dotty/tools/dotc/transform/CrossCastAnd.scala
+++ b/src/dotty/tools/dotc/transform/CrossCastAnd.scala
@@ -1,0 +1,48 @@
+package dotty.tools.dotc.transform
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Flags
+import dotty.tools.dotc.core.Types.{NoType, Type, AndType}
+import dotty.tools.dotc.transform.TreeTransforms._
+import tpd._
+
+import scala.collection.mutable.ListBuffer
+
+
+/**
+ * This transform makes sure that all private member selections from
+ * AndTypes are performed from the first component of AndType.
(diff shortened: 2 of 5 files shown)


DarkDimius:
Seems to have uncovered a bug in erasure:

A private method `TypeOps.deskolemize`in is being tailrec-optimized https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/TypeOps.scala#L135

A new `$this` symbol is created of type `dotty.tools.dotc.core.Contexts.Context & dotty.tools.dotc.core.TypeOps` which is full self type of TypeOps.
Note that 3-argument `deskolemize` is private in TypeOps and there is a 1-argument `deskolemize` which is public

Erasure tries to re-type `$this.deskolemize(arg1, arg2, arg3)`, erases `$this` to `Context`, takes a public overload with 1 argument, and dies with `head of empty list`.

DarkDimius:
I was able to minimize the issue. @odersky could you please take it from here?

odersky:
Sure, I'll take a look.

On Mon, Nov 9, 2015 at 7:04 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> I was able to minimize the issue. @odersky <https://github.com/odersky>
> could you please take it from here?
>
> —
> <https://github.com/lampepfl/dotty/pull/944#issuecomment-155141881>.
>



-- 
Martin Odersky
EPFL

odersky:
It looks like a problem with tailrec, not with erasure, as my last commit demonstrate. It's not legal to given the $this parameter a type which is a subtype of the class out of which it is lifted. On the other hand, you can't given it the class type itself, because then you could not access members available only from the thistype. Maybe tailrec should not move things to the companion module?

DarkDimius:
> Maybe tailrec should not move methods to the companion module?

It does not move methods to companion modules AFAIK:

```scala
result of ErasureAnd.scala after TreeTransform:{refchecks, elimRepeated, normalizeFlags, extmethods, expandSAMs, tailrec, liftTry, classOf}:
package <empty> {
  <trait> trait A() extends Object { self: B =>
    @annotation.tailrec() private def foo(arg1: Int, arg2: Int): Int = {
      def tailLabel2($this: B & A)(arg1: Int, arg2: Int): Int = {
        def bar: Int = $this.foo(arg1, arg2)
        tailLabel2($this)(arg1, arg2)
      }
      tailLabel2(this)(arg1, arg2)
    }
    def foo(arg: Int): Int = arg
  }
  final lazy module val A: A$ = new A$()
  final module class A$() extends Object() { this: A.type =>}
  class B() extends Object() with A {}
  final lazy module val B: B$ = new B$()
  final module class B$() extends Object() { this: B.type =>}
}
```

odersky:
@darkdimius You are right inso far as it does not lift them. But lifting is not the essential part. The essential part is the $this. We can't access a private `foo` from a $this of type `B & A`. It should be an
access error, but accesses are not checked with -Ycheck, so the whole thing goes undetected
until erasure.

DarkDimius:
@odersky 
>We can't access a private foo from a $this of type B & A

Do you mean that inside A one could access private member `foo` of A:
```
class A{
  private def foo = 1
  public def calsFoo(other: A) = other.foo
}
```

but accessing a private member through an AndType should not be allowed: 
```
class A{
  private def foo = 1
  public def calsFoo(other: A & Any) = other.foo // error?
}
```

DarkDimius:
note that 
```scala
class A{
  private def foo = 1
  def callsFoo1(other: A & B): Int = other.foo // error?
  def callsFoo2(other: B & A): Int = other.foo // error?
}
class B {
  def foo(i: Int) = i
}
```
compiles fine, but the resulting bytecode is malformed: 

```
  public int callsFoo1(A);
    Code:
       0: aload_1
       1: invokespecial #17                 // Method foo:()I
       4: ireturn

  public int callsFoo2(B);
    Code:
       0: aload_1
       1: getfield      #25                 // Field B.foo:I
       4: invokestatic  #31                 // Method scala/runtime/BoxesRunTime.unboxToInt:(Ljava/lang/Object;)I
       7: ireturn
```

odersky:
Maybe we need to upcast in situations like this?

DarkDimius:
@odersky, please review a new phase that addresses the issue. It inserts casts to types that are equivalent for unerased typesystem, but are treated differently by erasure.

DarkDimius:
> Why use AndType here? Would not qtype.baseTypeWithArgs(sym.owner) by itself be sufficient?

Because it's not, in `PhasesBase` dotty have infered type arguments for call that are very precise, and `qtype.baseTypeWithArgs(sym.owner)` is not precise enough to make it it Ycheck, and even `qtype` is not enough, as it is already widened.

odersky:
OK, I see. LGTM then.

On Tue, Nov 10, 2015 at 2:11 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> Why use AndType here? Would not qtype.baseTypeWithArgs(sym.owner) by
> itself be sufficient?
>
> Because it's not, in PhasesBase dotty have infered type arguments for
> call that are very precise, and qtype.baseTypeWithArgs(sym.owner) is not
> precise enough to make it it Ycheck, and even qtype is not enough, as it
> is already widened.
>
> —
> <https://github.com/lampepfl/dotty/pull/944#issuecomment-155415526>.
>



-- 
Martin Odersky
EPFL

PR #946, opened 2015-11-10
Make hash codes on type more predictable.

DarkDimius:
May help finding the partest issue.


DarkDimius:
@odersky please review.

odersky:
I think that will just give us worse hashcode without changing much of the underlying problem. To get reproducible builds we need to make every map that is traversed a linked map. That's a better way to get there.

DarkDimius:
> I think that will just give us worse hashcode

Why would they be worse?

PR #949, opened 2015-11-11
Comment three entries that do not compile yet.

odersky:
I am not sure how thse passed Jenkins. But it's clear we cannot compile
Range, because it contains an implicit without result type, so the
whitelist cannot have passed the test. Review by @DarkDimius

### DIFF
--- a/test/dotc/scala-collections.whitelist
+++ b/test/dotc/scala-collections.whitelist
@@ -181,7 +181,7 @@
 ./scala-scala/src/library/scala/collection/SortedMap.scala
 ./scala-scala/src/library/scala/collection/SortedMapLike.scala
 ./scala-scala/src/library/scala/collection/SortedSet.scala
-./scala-scala/src/library/scala/collection/SortedSetLike.scala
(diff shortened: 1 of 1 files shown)


DarkDimius:
@odersky, did you pull https://github.com/DarkDimius/scala/commits/dotty-library ?

odersky:
I didn't, sorry!


On Wed, Nov 11, 2015 at 7:02 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> @odersky <https://github.com/odersky>, did you pull
> https://github.com/DarkDimius/scala/commits/dotty-library ?
>
> —
> <https://github.com/lampepfl/dotty/pull/949#issuecomment-155862728>.
>



-- 
Martin Odersky
EPFL

DarkDimius:
As a detail, that could help in resolving such issues in future,
I've updated jenkins script a week ago to print a short git history before running tests.

https://scala-ci.typesafe.com/job/dotty-master-validate-junit/660/consoleFull
You can search for "git log" there and see its output for clone of dotty as well as clone of scala.

Issue #925, opened 2015-11-05
Typer: failure to type usages of `tupple.zipped` with Predef compiled by Dotty


DarkDimius:
```
$ dotc scala-scala/src/library/scala/Predef.scala  ./src/dotty/./tools/dotc/ast/Desugar.scala


warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
./src/dotty/./tools/dotc/ast/Desugar.scala:771: error: no implicit argument of type runtime.Tuple2Zipped.Ops#T1 =>
  scala.collection.TraversableLike[Nothing', Nothing'] found for parameter w1 of method zipped in class Ops
          val pdefs = (defpats, rhss).zipped map (makePatDef(Modifiers(), _, _))
                                            ^
```

DarkDimius:
@odersky could you please have a look at this issue?

odersky:
It's fixed in #950 and, I assume, in master.

DarkDimius:
@odersky I can reproduce this on 661da00

DarkDimius:
```
dark@tsf-452-wpa-2-017 ~/workspace/dotty ((661da00...)) $ dotc scala-scala/src/library/scala/Predef.scala  ./src/dotty/./tools/dotc/ast/Desugar.scala
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
./src/dotty/./tools/dotc/ast/Desugar.scala:771: error: no implicit argument of type runtime.Tuple2Zipped.Ops#T1 =>
  scala.collection.TraversableLike[Nothing', Nothing'] found for parameter w1 of method zipped in class Ops
          val pdefs = (defpats, rhss).zipped map (makePatDef(Modifiers(), _, _))
                                            ^
./src/dotty/./tools/dotc/ast/Desugar.scala:771: error: map is not a member of (implicit w1: runtime.Tuple2Zipped.Ops#T1 =>
[...]

PR #951, opened 2015-11-11
Reproduce #925

DarkDimius:


odersky:
We figured it out. A missing type parameter in Predef. Dotty should have detected this, however.

PR #953, opened 2015-11-13
Fix huge amount of collisions in Scopes.

DarkDimius:
Using & to compute hash codes means that some of the buckets
will never be used.

Eg, lets say that array hash size 9.
IE 1001 in binary.
If we take `(1001 - 1) & name.hashcode`,
only 0 and 8 buckets will be used.


DarkDimius:
Ok. Playing with it more I see that there is an implicit assumption that table.size is power of 2. In that case the original code is good.

PR #954, opened 2015-11-13
Check that symbol accesses are single threaded

odersky:


odersky:
I don't think we need this anymore.

PR #957, opened 2015-11-16
Fix #938

### LINKED ISSUE #938: TraversableLike: error: local definition of $anon escapes
```
$ dotc ./scala-scala/src/library/scala/collection/TraversableLike.scala
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
warning: encountered F-bounded higher-kinded type parameters for type CC; assuming they are invariant
./scala-scala/src/library/scala/collection/TraversableLike.scala:666: error: local definition of $anon escapes as part of expression's type scala.collection.TraversableView{A =+ A; Coll =+ Repr; underlying: Repr}
  def view = new TraversableView[A, Repr] {
                 ^
four warnings found
one error found
```

odersky:
Review by @smarter. And, maybe take it over if there are still things to do.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -488,7 +488,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         case _ => tp2 isRef ObjectClass
       }
       compareJavaArray
-    case tp1: ExprType if ctx.phase.id >= ctx.gettersPhase.id =>
+    case tp1: ExprType if ctx.phase.id > ctx.gettersPhase.id =>
       // getters might have converted T to => T, need to compensate.
       isSubType(tp1.widenExpr, tp2)
     case _ =>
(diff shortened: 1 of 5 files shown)


src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -488,6 +488,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         case _ => tp2 isRef ObjectClass
       }
       compareJavaArray
+    case tp1: ExprType if ctx.phase.id >= ctx.gettersPhase.id =>
```
smarter: Should this be `ctx.phase.id > ctx.gettersPhase.id` like in https://github.com/dotty-staging/dotty/blob/fix-%23938/src/dotty/tools/dotc/core/TypeComparer.scala#L455 ?

odersky: Yes, good catch.

smarter:
LGTM.

odersky:
Rebased to master

PR #958, opened 2015-11-16
Links between companions after unpickling are added

VladimirNik:
Fix for the problem with missing links between companions during unpickling https://github.com/lampepfl/dotty/issues/881:
1. Companion methods generation added to TreeUnpickler;
2. def registerCompanionMethod is shared;
Old pull request https://github.com/lampepfl/dotty/pull/922

Review by @DarkDimius @odersky


DarkDimius:
LGTM

PR #965, opened 2015-11-19
Fixed != comparison

reactormonk:
By linter:
comparing values of types Char and String using `!=' will always yield
true


DarkDimius:
LGTM. Thanks for contribution!

PR #967, opened 2015-11-20
New tests for tasty

VladimirNik:
New tests for tasty are added - mainly checking of dotc parts that currently can be pickled/unpickled (compare the string representation of trees before and after pickler).


VladimirNik:
Review by @DarkDimius

PR #969, opened 2015-11-22
Swap order of arguments in annotations

odersky:
The fact that the annotation comes first is weird, because when I write
an annotated type it's <type> @<annotation>. Also, annotated types
are like RefinedTypes in that they derive from a parent type. And in
RefinedTypes the parent comes first.

So swapping the arguments improves consistency. Review by @VladimirNik.
@VladimirNik can you take this PR over to do the same change in the TASTY format?


VladimirNik:
I changed rule for Annotated pickling to:
```ANNOTATED      Length underlying_Type fullAnnotation_Term```

@odersky Could you please modify it in TASTY Reference Manual? I haven't got write access to it.

VladimirNik:
LGTM

PR #973, opened 2015-11-23
Allow refinement of types without a matching type in the parent

smarter:
This partially revert 61cb51acaedbe603add8c4af9e390a27f8b33f09 which
disallowed both refinement of types and of methods, this allows us to
compile scala.collection.immutable.Set, as well as a lot of Scala code
in the wild that uses type lambdas.

Review by @odersky


odersky:
Let's postpone a decision in this. My current tendency is not to go there.

Issue #977, opened 2015-11-25
Duplication of companion methods after using FromTasty


VladimirNik:
Example:
```scala
class X
```
Run:
```
dotc -tasty -Ycheck:all X
```
Problem:
```
Exception in thread "main" class dotty.tools.dotc.reporting.Reporter$Error at ?: method companion$module is already defined as method companion$module: => X$
```

PR #978, opened 2015-11-25
Duplication of companion methods after using FromTasty fixed

VladimirNik:
Fix for: https://github.com/lampepfl/dotty/issues/977

The problem was due to `scalacLinkedClass` method invocation during unpickling. Companion methods should be generated only if symbol returned by `scalacLinkedClass` is among unpickled symbols.

Review by @odersky, @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -41,11 +41,17 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
   }
 
   private val symAtAddr  = new mutable.HashMap[Addr, Symbol]
+  private val unpickledSyms = new mutable.HashSet[Symbol]
   private val treeAtAddr = new mutable.HashMap[Addr, Tree]
   private val typeAtAddr = new mutable.HashMap[Addr, Type] // currently populated only for types that are known to be SHAREd.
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -590,7 +590,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
         case TYPEDEF | TYPEPARAM =>
           if (sym.isClass) {
             val companion = sym.scalacLinkedClass
-            if (companion != NoSymbol) {
+            if (companion != NoSymbol && symAtAddr.valuesIterator.contains(companion)) {
```
DarkDimius: why not `symAtAddr.contains(companion)`?

VladimirNik: @DarkDimius because symAtAddr is a map, key is an address.

DarkDimius: `valuesIterator` takes linear time to check if element is contained.
It means that this change makes reading tasty quadratic in number of `TypeDefs`. 

if `companion != NoSymbol`, where does `companion` originate from if not from Tasty?

DarkDimius:
LGTM

PR #979, opened 2015-11-26
TypeComparer: delay looking up members of AndTypes

smarter:
In ParFactory.scala we have checks that look like:
```scala
  (Foo { type Bar = X }) & (Foo { type Bar = X }) <:< (Foo { type Bar = X })
```
where `Foo` is a recursive type.

Before this commit, we would first try to check this by looking up `Bar`
in the `AndType` on the left, which means looking it up in both branches
and then merging the result, but the merge requires more subtyping
checks, which in turn require looking up a member inside an `AndType`,
seemingly ad infinitum.

We now avoid this by checking if either branch of the `AndType` on the left
is a subtype of the `RefinedType` on the right before looking up a member
in the `AndType` itself.

Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -331,17 +331,23 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         }
       comparePolyParam
     case tp2: RefinedType =>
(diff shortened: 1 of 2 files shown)


odersky:
LGTM. I think the logic can be slightly simplified since AndTypes never need an eta lift. I'll take care of that in the higher-kinded PR.

Issue #970, opened 2015-11-23
Cyclic reference when using List and compiling Seq* classes at the same time


DarkDimius:
```
$ cat B.scala
object B{
  def main(args: Array[String]): Unit = {
    val s = List(1,2,3)
    ()
  }
}

$ dotc ./scala-scala/src/library/scala/collection/immutable/Seq.scala ./scala-scala/src/library/scala/package.scala ./scala-scala/src/library/scala/collection/GenSeqLike.scala ./scala-scala/src/library/scala/collection/SeqLike.scala ./scala-scala/src/library/scala/collection/generic/GenSeqFactory.scala B.scala
./scala-scala/src/library/scala/collection/SeqLike.scala:62: error: cyclic reference involving trait ParSeq
trait SeqLike[+A, +Repr] extends Any with IterableLike[A, Repr] with GenSeqLike[A, Repr] with Parallelizable[A, ParSeq[A]] { self =>
                                                                                                                ^
```

If `B.scala` is removed from compilation list, cyclic reference error is not triggered anymore.

smarter:
Can you still reproduce this? Seems to work fine here now.

DarkDimius:
I can reproduce it on 4463f5e705480c7da188695abdb0ccacf452103c locally

DarkDimius:
@smarter could you please try with different order of arguments: 

```
dotc B.scala ./scala-scala/src/library/scala/collection/immutable/Seq.scala ./scala-scala/src/library/scala/package.scala ./scala-scala/src/library/scala/collection/GenSeqLike.scala ./scala-scala/src/library/scala/collection/SeqLike.scala ./scala-scala/src/library/scala/collection/generic/GenSeqFactory.scala
```

smarter:
Yes, I can reproduce that way.

smarter:
OK, I think I got it but I'll need some help with the design: https://github.com/lampepfl/dotty/pull/981

odersky:
This is fixed by now.

PR #981, opened 2015-11-27
[PROOF-OF-CONCEPT/DO NOT MERGE] Avoid forcing type aliases to get their typeParams

smarter:
This is a strawman proposal for fixing #970, before this commit, the
following failed with a cyclic reference error:
```bash
dotc B.scala\
./scala-scala/src/library/scala/collection/immutable/Seq.scala\
./scala-scala/src/library/scala/package.scala\
./scala-scala/src/library/scala/collection/GenSeqLike.scala\
./scala-scala/src/library/scala/collection/SeqLike.scala\
./scala-scala/src/library/scala/collection/generic/GenSeqFactory.scala
```

where B.scala is defined as:
```scala
object B{
    val s = List(1,2,3)
    ()
  }
}
```

Here's what I think is happening:
1. Unpickling a Scala2 TypeRef requires us to perform type application.
   (`Scala2Unpickler#readType`)
2. To apply a TypeAlias to some arguments, we first need to determine its
   type parameters.
   (`TypeApplications#appliedTo` and `TypeApplications#typeParams`)
3. To do this, we look at the type parameters of the underlying type of
   the TypeAlias, this requires completing the TypeAlias.
4. If the TypeAlias is defined in a source tree and not unpickled, this
   forces us to typecheck its right-hand side.
   (`Namer#Completer#typeSig`)
5. In turns, this forces various classes to be completed, which might
   themselves refer indirectly to type aliases, forcing even more
   stuff.

This commit is a hacky way to avoid 3. by only completing the type parameters
of a type alias instead of completing the whole type alias when we don't
need it. Let me know if you think the basic idea is sound or not and if
you can think of a nicer way to implement it!

CC @odersky

### DIFF
--- a/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -16,6 +16,8 @@ import util.Positions.Position
 import config.Printers._
 import collection.mutable
 import java.util.NoSuchElementException
+import ast.untpd.TypeDef
+import typer.Namer
 
 object TypeApplications {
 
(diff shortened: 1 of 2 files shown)


odersky:
Superseded by #1027

Issue #982, opened 2015-11-27
Incorrect unpickling of path-dependent types


VladimirNik:
In some cases after unpickling path-dependent types become type projections. Example:
```scala
trait Z {
  type Q
  def test: Q
}
class X(val x: Z)
class Y(x: Z) extends X(x) {
  x.test
}
```
x.test before pickling:
```
<<Y.this.x:=> Z>.test:=> Y.this.x.Q>
```
x.test after unpickling:
```
<<Y.this.x:=> Z>.test:=> Z#Q>
```

PR #983, opened 2015-11-27
Add pickling/unpickling of stable modifier

VladimirNik:
Pickling/unpickling of stable modifier is added.

Fix for https://github.com/lampepfl/dotty/issues/982

Review by @odersky , @DarkDimius


DarkDimius:
LGTM

PR #984, opened 2015-11-29
Fix inefficieny in the presence of aliasing.

odersky:
The change to do compareAlias early caused a dramatic slowdown of compilation

compileStdLib went from 45 sec to 230 sec. The problem were many redundant tests
when every member of an alias chain was compared to every other.

The new scheme follows alias chains to their end before doing anything else. Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -286,7 +286,7 @@ class TypeApplications(val self: Type) extends AnyVal {
    */
   final def baseArgInfos(base: Symbol)(implicit ctx: Context): List[Type] =
     if (self derivesFrom base)
-      base.typeParams map (param => self.member(param.name).info.argInfo(param))
+      base.typeParams map (param => self.member(param.name).info.argInfo)
     else
       Nil
 
@@ -311,7 +311,7 @@ class TypeApplications(val self: Type) extends AnyVal {
   /** The first type argument of the base type instance wrt `base` of this type */
   final def firstBaseArgInfo(base: Symbol)(implicit ctx: Context): Type = base.typeParams match {
     case param :: _ if self derivesFrom base =>
-      self.member(param.name).info.argInfo(param)
+      self.member(param.name).info.argInfo
     case _ =>
       NoType
   }
@@ -371,7 +371,7 @@ class TypeApplications(val self: Type) extends AnyVal {
    *  Existential types in arguments are returned as TypeBounds instances.
    *  @param interpolate   See argInfo
    */
-  final def argInfos(interpolate: Boolean)(implicit ctx: Context): List[Type] = {
+  final def argInfos(implicit ctx: Context): List[Type] = {
     var tparams: List[TypeSymbol] = null
     def recur(tp: Type, refineCount: Int): mutable.ListBuffer[Type] = tp.stripTypeVar match {
       case tp @ RefinedType(tycon, name) =>
@@ -381,7 +381,7 @@ class TypeApplications(val self: Type) extends AnyVal {
           if (tparams == null) tparams = tycon.typeParams
           if (buf.size < tparams.length) {
             val tparam = tparams(buf.size)
-            if (name == tparam.name) buf += tp.refinedInfo.argInfo(tparam, interpolate)
+            if (name == tparam.name) buf += tp.refinedInfo.argInfo
             else null
           } else null
         }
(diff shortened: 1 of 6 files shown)


src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -140,35 +140,35 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
-            compareHK(tp1, tp2, inOrder = true) ||
-            compareHK(tp2, tp1, inOrder = false) ||
-            thirdTryNamed(tp1, tp2)
-          case _ =>
-            compareHK(tp2, tp1, inOrder = false) ||
-            compareAlias(NoType) ||
-            secondTry(tp1, tp2)
+        implicit val ctx: Context = this.ctx
+        tp2.info match {
+          case info2: TypeAlias => firstTry(tp1, info2.alias)
```
smarter: It seems like you could avoid a bunch of recursive calls to `firstTry` by using a utility function like:
```scala
def dealiasTypeRef(tp: TypeRef): Type =
  if (tp.symbol.isClass) tp
  else tp.info match {
   case TypeAlias(alias: TypeRef) => dealiasTypeRef(alias)
   case TypeAlias(alias) => alias
   case _ => tp
  }
```

odersky:
Yes, but would this be faster? I noticed we had some deteriorations in
compiler performance lately. compileStdLib was the poster child. But even
after the fix, compile dotty went up from 66 to 81 seconds. I guess we
should keep a watchful eye on this and do some measurements everytime we
touch core code.



On Sun, Nov 29, 2015 at 5:47 PM, Guillaume Martres <notifications@github.com
> wrote:

> /rebuild
>
> —
> <https://github.com/lampepfl/dotty/pull/984#issuecomment-160431489>.
>



-- 
Martin Odersky
EPFL

smarter:
It should be as fast, but less recursive calls mean less risk of deep subtype errors

smarter:
> stderr: error: could not lock config file .git/config: No space left on device

@SethTisue @adriaanm : Help :frowning:

odersky:
Note that the two calls in question are tailcalls anyway.

smarter:
> Note that the two calls in question are tailcalls anyway.

Yes, but `firstTry` cannot be tailcall-optimized so they still consume stack space, if you try to annotate it with `@tailrec` you'll see:
```scala
TypeComparer.scala:235: could not optimize @tailrec annotated method firstTry: it contains a recursive call not in tail position
       secondTry(tp1, tp2)
                ^
```

odersky:
@smarter Yes, you are right. It's not a recognized as a tailcall because it
comes from a nested method.

One thing we could try is simply call firstTry with dealiased types. That
way we could avoid the
distinctions in firstTry itself. Do you want to try that and see how it
affects performance?

On Sun, Nov 29, 2015 at 6:13 PM, Guillaume Martres <notifications@github.com
> wrote:

> Note that the two calls in question are tailcalls anyway.
>
> Yes, but firstTry cannot be tailcall-optimized so they still consume
> stack space, if you try to annotate it with @tailrec you'll see:
>
> TypeComparer.scala:235: could not optimize @tailrec annotated method firstTry: it contains a recursive call not in tail position
>        secondTry(tp1, tp2)
>                 ^
>
> —
> <https://github.com/lampepfl/dotty/pull/984#issuecomment-160432660>.
>



-- 
Martin Odersky
EPFL

smarter:
I just tried it and it seems to be about as fast as your patch, I'll make a PR

smarter:
Oh, I see you just changed `compareNamed` to take arguments, so I don't need to make a PR anymore.

smarter:
@odersky : I made a commit to avoid the recursive calls to `compareNamed`, feel free to add it to this PR: https://github.com/smarter/dotty/commits/optimize-subtyping

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -417,29 +414,11 @@ class TypeApplications(val self: Type) extends AnyVal {
 
   /** If this is the image of a type argument to type parameter `tparam`,
```
smarter: This still refers to `tparam` even though `tparam` is not a parameter of this method anymore. I think the comment could be simplified to just ` If this is the image of a type argument for a type parameter, recover the type argument, otherwise NoType.`, although I'd appreciate it if you could explain the meaning of `image` in the documentation.

smarter:
Otherwise LGTM.

odersky:
Could you do the change and commit please? It's currently hard for me to
make the context switch. Thanks!

On Mon, Nov 30, 2015 at 3:47 PM, Guillaume Martres <notifications@github.com
> wrote:

> Otherwise LGTM.
>
> —
> <https://github.com/lampepfl/dotty/pull/984#issuecomment-160650717>.
>



-- 
Martin Odersky
EPFL

smarter:
Sure!

SethTisue:
I will go fix the disk space problem now.

smarter:
@SethTisue : Hurray, thanks :)

PR #985, opened 2015-12-01
Add test for dotc.core pickling/unpickling

VladimirNik:
Tests for doc.core pickling/unpickling:
dotc.core
dotc.core.tasty
dotc.core.classfile
dotc.core.unpickleScala2

core/Types.scala is not presented in tests because Types contains transformed OR type after unpickling.

Review by @DarkDimius

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -230,6 +230,19 @@ class tests extends CompilerTest {
   @Test def tasty_dotc_printing = compileDir(dotcDir, "printing", testPickling)
   //@Test def tasty_dotc_reporting = compileDir(dotcDir, "reporting", testPickling)
   @Test def tasty_dotc_util = compileDir(dotcDir, "util", testPickling)
+  @Test def tasty_core = compileList("tasty_core", List(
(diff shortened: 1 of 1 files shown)


test/dotc/tests.scala
```diff
@@ -230,6 +230,40 @@ class tests extends CompilerTest {
   @Test def tasty_dotc_printing = compileDir(dotcDir, "printing", testPickling)
   //@Test def tasty_dotc_reporting = compileDir(dotcDir, "reporting", testPickling)
   @Test def tasty_dotc_util = compileDir(dotcDir, "util", testPickling)
+  @Test def tasty_core_annotations = compileFile(coreDir, "Annotations", testPickling)
```
DarkDimius: instead of having a test per file, would you consider using `compileList` to test them all together?

DarkDimius:
Otherwise LGTM

test/dotc/tests.scala
VladimirNik: @DarkDimius fixed

DarkDimius:
LGTM

PR #989, opened 2015-12-02
Add initial CompilerCallback implementation for IntelliJ

smarter:
Review by @DarkDimius


DarkDimius:
LGTM, but it would be good to hear from IntelliJ that this is good for them.
cc @alefas

niktrop:
Yes, I can use it. Thanks.

PR #994, opened 2015-12-06
Change typeapply

odersky:
Changes to arrive at a more tractable treatment of higher-kinded types.


src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -173,101 +252,216 @@ class TypeApplications(val self: Type) extends AnyVal {
...
-        }
-        val tparam = tparams.head
-        val tp1 = RefinedType(tp, tparam.name, arg.toBounds(tparam))
-        matchParams(tp1, tparams.tail, args1)
-      case nil => tp
-    }
+  /** Lambda abstract `self` with given type parameters. Examples:
+   *
+   *      type T[X] = U        becomes    type T = [X] -> U
+   *      type T[X] >: L <: U  becomes    type T >: L <: ([X] -> _ <: U)
```
smarter: Does that mean that we cannot refer to `X` in `L` ?

odersky: Yes

odersky:
Rebased to master

src/dotty/tools/dotc/core/TypeApplications.scala
smarter: OK, then this should be explicitly disallowed somewhere, currently the following:
```scala
class Test {
  type Foo[X] >: List[X]
}
```
Triggers an assertion: `unresolved symbols: type X`
On the other hand, couldn't we rewrite `type T[X] >: L <: U`  as `type T = [X] -> (_ >: L <: U)` or something else to allow this?

src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -502,19 +508,89 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
+   *   (1) If not `inOrder` then perform the next steps until they all succeed
+   *       for each base type of other which
+   *        - derives from a class bound of `projection`,
+   *        - has the same number of type parameters than `projection`
+   *        - has type parameter variances which conform to those of `projection`.
+   *       If `inOrder` then perform the same steps on the original `other` type.
+   *
+   *   (2) Try to eta expand the constructor of `other`.
+   *
+   *   (3a) In mode `TypeVarsMissConetxt` replace the projection's hk constructor parameter
```
smarter: Typo: `TypevarsMissContext`

src/dotty/tools/dotc/core/tasty/TreePickler.scala
```diff
@@ -181,17 +185,16 @@ class TreePickler(pickler: TastyPickler) {
...
-          // be reconstituted when unpickling.
-          pickleType(tpe.prefix)
-        else if (isLocallyDefined(tpe.symbol)) {
-          writeByte(if (tpe.isType) TYPEREFsymbol else TERMREFsymbol)
-          pickleSymRef(tpe.symbol); pickleType(tpe.prefix)
-        }
-        else {
-          writeByte(if (tpe.isType) TYPEREF else TERMREF)
-          pickleName(tpe.name); pickleType(tpe.prefix)
+        tpe match {
```
smarter: Unnecessary pattern matching.

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -840,26 +841,23 @@ class Namer { typer: Typer =>
 
   /** Eta expand all class types C appearing as arguments to a higher-kinded
    *  type parameter to type lambdas, e.g. [HK0] => C[HK0]. This is necessary
-   *  because in `typedAppliedTypeTree` we might ahve missed some eta expansions
+   *  because in `typedAppliedTypeTree` we might have missed some eta expansions
    *  of arguments in F-bounds, because the recursive type was initialized with
    *  TypeBounds.empty.
    */
+  // ### Check whether this is still needed!
```
smarter: So is this still needed? :)

smarter:
The following:
```scala
object Test {
  def foo[M[X] <: List[X]](x: M[Int]) = ???
}
```
gets pretty-printed after frontend as:
```scala
package <empty> {
    def foo[M <: [] -> <: scala.collection.immutable.List[X0]](x: M[Int]): Nothing = ???
  }
}
```
The list of type lambda arguments is `[]` but should be `[X0]`

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: That means bounds can be lambdas (and lambdas are complex arrangements of refinements and references). So what was a TypeBounds type now becomes something very complicated. I am afraid to got there. If we had explicit type lambdas as a separate type it would take it under consideration. But then we need also explicit AppliedType types.

odersky:
@smarter The print problem should be fixed with 1c77b03.

odersky:
All comments should be addressed now.

smarter:
I'm looking into why `t3152.scala` stopped compiling, I'd like to fix this before merging this PR.

smarter:
`adaptIfHK` is wrong, it bypasses variance checks and allows this to compile:
```scala
class HKCov[M[+_]] {
  def upcast(m: M[Int]): M[Any] = m
}
class Foo[T]
object Test {
  val x: HKCov[Foo] = new HKCov[Foo]
  val fi: Foo[Int] = new Foo[Int]
  val fa: Foo[Any] = x.upcast(fi)
}
```
The type parameter of `Foo` is invariant so it shouldn't be possible to pass `Foo` to `HKCov`, however `adaptIfHK` rewrites `[X] => Foo[X]` into `[+X] => Foo[X]`, this opens a soudness hole.

I've tried simply removing `adaptIfHK` and nothing seems to break, and even better: `tests/pending/pos/t3152.scala` now successfully compiles!

Do you have a test case where `adaptIfHK` is actually needed?

odersky:
The comment to adaptIfHK gives some clue. It certainly was needed at some point for compileStdLib. To make sure, it would be good to just compile with some of the mutable classes mentioned in the comment. I.e. starting with GenTraversable and ListBuffer and then add a few.

odersky:
Indeed we need to test whether the variances correspond before doing the adaptation. I.e. if the context requires nonvariant but the argument is covariant OK, in the other direction not.

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -173,101 +260,237 @@ class TypeApplications(val self: Type) extends AnyVal {
...
+   *
+   *      GenericCompanion[[X] -> ListBuffer[X]] <: GenericCompanion[[+X] -> GenTraversable[X]]
+   *
+   *   Without adaptation, this would be false, and hence an overriding error would
+   *   result. But with adaptation, the rhs argument will be adapted to
+   *
+   *     [X] -> GenTraversable[X]
+   *
+   *   which makes the subtype test succeed. The crucial point here is that, since
+   *   GenericCompanion only expects a non-variant CC, the fact that GenTraversable
```
smarter: I think you should use a different example than `GenericCompanion` here, because `GenericCompanion` does expect a covariant CC: https://github.com/DarkDimius/scala/blob/dotty-library/src/library/scala/collection/generic/GenericCompanion.scala#L26

smarter: Oops, nevermind, the CC parameter is covariant but the X parameter of the CC parameter is invariant.

PR #999, opened 2015-12-11
Change idempotent as seen from

odersky:
Review by @smarter? Only the last commit is new.


smarter:
Otherwise LGTM but I'm wondering if `asSeenFrom` being idempotent avoids some bugs/helps with something, or if you just thought it was a useful thing to do in general?

odersky:
Rebased on master

PR #1001, opened 2015-12-15
Change distribute or

odersky:
More changes to alias tests and rewriting of lubs.

Removes known unsoundness and makes previously seen deep subtypes go away. Review by @smarter.


smarter:
This looks better, but we are still making the assumption that type aliases and type bounds are invariant.There are two cases where we can have covariant aliases:

- In type parameters, but because they're private it doesn't open a soundness hole I think.
- In higher-kinded types with `#Apply` and here we have a soundness issue. To demonstrate this I started from a minimized `Iter2.scala` and got:
```scala
val c: Iterable[A] with FromIterator[C] = ???
val d: Iterable[A] with FromIterator[D] = ???
val cget = c.get()
var dget = d.get()
dget = cget
```
where the inferred type of `cget` is `(([+X0] -> Test.Iterable[X0]) & C)[Int]` and the inferred type of `dget` is `(([+X0] -> Test.Iterable[X0]) & D)[Int]`, so they should not be compatible and yet I can assign `dget = cget`, because we determined that `(([+X0] -> Test.Iterable[X0]) & C)#Apply <:< (([+X0] -> Test.Iterable[X0]) & D)#Apply` because we followed the alias on the right. The full code is at: https://gist.github.com/anonymous/ff0b523503b988ee161d

Unfortunately, if we avoid following aliases when variance != 0, we end up with the same deep subtyping issues we had before.

Note that the soundness bug is also present in master, because we assume that we can follow the lower bound of a `TypeBounds` (https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/TypeComparer.scala#L328), but that's incorrect when the `TypeBounds` is a `TypeAlias` with `variance = 1`

smarter:
I think we can fix this by:

1. Always distributing `#Apply`, this should be safe because it does not make sense to intersect a lambda with something that is not a lambda
2. Disallowing `A#Apply <:< B#Apply if A <:< B`

Then we can safely follow type aliases.

odersky:
Can you complete the example to give an actual classcast exception? Would
be good to have this as a test case for future versions of the type system.


On Tue, Dec 15, 2015 at 9:05 PM, Guillaume Martres <notifications@github.com
> wrote:

> I think we can fix this by:
>
>    1. Always distributing #Apply inside intersections and unions, this
>    should be safe because it does not make sense to intersect a lambda with
>    something that is not a lambda
>    2. Disallowing A#Apply <:< B#Apply if A <:< B
>
> Then we can safely follow type aliases.
>
> —
> <https://github.com/lampepfl/dotty/pull/1001#issuecomment-164879270>.
>



-- 
Martin Odersky
EPFL

smarter:
Sure: https://gist.github.com/anonymous/bca53534131e14ae01cc

smarter:
So the reason we end up with these types is that at some point we call `findMember(C, ...)` on `Test.Iterable[A] & Test.FromIterator[C]`, so we get `[X] => Iterable[X]` from the lhs and `C` from the rhs and we merge them with `&`, I think that if we teach `glb` (and `lub`) to deal with higher-kinded types, we're good.

odersky:
Rebased to master. Review by @smarter.

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1533,19 +1533,49 @@ object Types {
...
           // `C { type hk$0 = T0; ...; type hk$n = Tn } # $Apply`
           // where C is a class. In that case we eta expand `C`.
           derivedSelect(prefix.EtaExpandCore(this.prefix.typeConstructor.typeParams))
-        }
+        else if (Config.splitProjections)
+          prefix match {
+            case prefix: AndType =>
+              def isMissing(tp: Type) = tp match {
+                case tp: TypeRef => !tp.info.exists
+                case _ => false
```
smarter: Isn't `isMissing` too restrictive? If we have `(S & T)#A`, then `S#A` might be an alias in `S`, so `derived1` may not return a `TypeRef`.

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -173,8 +173,45 @@ object TypeApplications {
...
+   *
+   *     [v1 X1 B1, ..., vn Xn Bn] -> op(tp1[X1, ..., Xn], tp2[X1, ..., Xn])
+   *
+   *  where
+   *
+   *   - variances `vi` are the variances of corresponding type parameters for `tp1`
+   *     or `tp2`, or are 0 of the latter disagree.
+   *   - bounds `Bi` are the intersection of the corresponding type parameter bounds
+   *     of `tp1` and `tp2`.
+   */
```
smarter: Maybe add `@pre tparams1.length == tparams2.length` or add an assert.

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -173,8 +173,45 @@ object TypeApplications {
...
+      (tparams1, tparams2).zipped.map { (tparam1, tparam2) =>
+        val b1: RefinedType => TypeBounds =
+          tp1.memberInfo(tparam1).bounds.internalizeFrom(tparams1)
+        val b2: RefinedType => TypeBounds =
+          tp2.memberInfo(tparam2).bounds.internalizeFrom(tparams2)
+        (rt: RefinedType) => b1(rt) & b2(rt)
+      }
+    val app1: RefinedType => Type = rt => tp1.appliedTo(argRefs(rt, tparams1.length))
+    val app2: RefinedType => Type = rt => tp2.appliedTo(argRefs(rt, tparams2.length))
+    val body = (rt: RefinedType) => op(app1(rt), app2(rt))
```
smarter: Very minor point, but why not use the same style for `app1`,`app2` and `body`? Either they should all be written `val x: RefinedType => Type = rt => ...` or they should all be written `val x = (rt: RefinedType) => ...`

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -282,7 +282,7 @@ trait TypeAssigner {
       else if (!mix.isEmpty) findMixinSuper(cls.info)
       else if (inConstrCall || ctx.erasedTypes) cls.info.firstParent
       else {
-        val ps = cls.info.parents
+        val ps = cls.classInfo.instantiatedParents
```
smarter: Was this change motivated by some test that didn't pass?

smarter:
Otherwise LGTM, I can't break it anymore :).

src/dotty/tools/dotc/typer/TypeAssigner.scala
odersky: Yes, it was when compiling the dotty compiler itself.

src/dotty/tools/dotc/core/Types.scala
odersky: Don't understand. `isMissing` is supposed to detect when a branch of an |/& does not have a member with the given name. If it is an alias then it does have a member, so it won't be missing anyway.

smarter: Nevermind, I misunderstood how this function worked.

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: Yes, it does look much nicer when harmonized.

PR #1002, opened 2015-12-16
Add a phase to Dotty that hides Dotty bottom types from JVM.

DarkDimius:
It is very convenient to have bottom types in type system. Unfortunately
    JVM does not have bottom types and we need to insert casts that would make
    bytecode pass verification.

### DIFF
--- a/src/dotty/tools/backend/jvm/BottomTypes.scala
+++ b/src/dotty/tools/backend/jvm/BottomTypes.scala
@@ -44,7 +44,7 @@ class BottomTypes extends MiniPhaseTransform {
 
 
   def adaptBottom(treeOfBottomType: tpd.Tree, expectedType: Type)(implicit ctx: Context) = {
-    if (Erasure.Boxing.isNonJVMBottomType(treeOfBottomType.tpe) && (treeOfBottomType.tpe ne expectedType))
+    if (defn.isBottomType(treeOfBottomType.tpe) && (treeOfBottomType.tpe ne expectedType))
       Erasure.Boxing.adaptToType(treeOfBottomType, expectedType)
     else treeOfBottomType
   }
@@ -68,7 +68,7 @@ class BottomTypes extends MiniPhaseTransform {
   override def transformApply(tree: tpd.Apply)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
     val fun = tree.fun
     val newArgs: List[tpd.Tree] = tree.args.zip(fun.tpe.dealias.firstParamTypes).map(x => adaptBottom(x._1, x._2))
-    val changeNeeded = tree.args == newArgs // cpy.Apply does not check if elements are the same,
+    val changeNeeded = tree.args != newArgs // cpy.Apply does not check if elements are the same,
                                             // it only does `eq` on lists as whole
     if (changeNeeded) cpy.Apply(tree)(fun = fun, args = newArgs)
     else tree
--- a/src/dotty/tools/backend/jvm/BottomTypes.scala
+++ b/src/dotty/tools/backend/jvm/BottomTypes.scala
@@ -1,0 +1,81 @@
+package dotty.tools.backend.jvm
+
+import dotty.tools.dotc.ast.Trees.Thicket
+import dotty.tools.dotc.ast.{Trees, tpd}
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Types
+import dotty.tools.dotc.transform.TreeTransforms.{TransformerInfo, TreeTransform, MiniPhase, MiniPhaseTransform}
+import dotty.tools.dotc
+import dotty.tools.dotc.backend.jvm.DottyPrimitives
+import dotty.tools.dotc.core.Flags.FlagSet
+import dotty.tools.dotc.transform.Erasure
+import dotty.tools.dotc.transform.SymUtils._
+import java.io.{File => JFile}
+
+import scala.collection.generic.Clearable
+import scala.collection.mutable
+import scala.collection.mutable.{ListBuffer, ArrayBuffer}
+import scala.reflect.ClassTag
+import scala.reflect.internal.util.WeakHashSet
+import scala.reflect.io.{Directory, PlainDirectory, AbstractFile}
+import scala.tools.asm.{ClassVisitor, FieldVisitor, MethodVisitor}
+import scala.tools.nsc.backend.jvm.{BCodeHelpers, BackendInterface}
+import dotty.tools.dotc.core._
+import Periods._
+import SymDenotations._
+import Contexts._
+import Types._
+import Symbols._
+import Denotations._
+import Phases._
+import java.lang.AssertionError
+import dotty.tools.dotc.util.Positions.Position
+import Decorators._
+import tpd._
+import Flags._
+import StdNames.nme
+
+/**
+  *  Ensures that tree does not contain type subsumptions where subsumed type is bottom type
+  *  of our typesystem, but not the bottom type of JVM typesystem.
+  */
+class BottomTypes extends MiniPhaseTransform {
+  def phaseName: String = "bottomTypes"
+
+
+  def adaptBottom(treeOfBottomType: tpd.Tree, expectedType: Type)(implicit ctx: Context) = {
+    if (Erasure.Boxing.isNonJVMBottomType(treeOfBottomType.tpe) && (treeOfBottomType.tpe ne expectedType))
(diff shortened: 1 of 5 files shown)


DarkDimius:
@odersky please review.

src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -263,6 +263,12 @@ object Erasure extends TypeTestsCasts{
           else
             cast(tree, pt)
       }
+
+
+    /** Is `tpe` a type which is a bottom type for Dotty, but not a bottom type for JVM */
```
smarter: There is no bottom type on the JVM so I don't really understand the naming of this function, how is it different from `Definitions#isBottomType`?

src/dotty/tools/backend/jvm/BottomTypes.scala
```diff
@@ -0,0 +1,81 @@
...
+
+
+  override def transformTyped(tree: tpd.Typed)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    cpy.Typed(tree)(adaptBottom(tree.expr, tree.tpt.tpe), tree.tpt)
+  }
+
+  override def transformApply(tree: tpd.Apply)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    val fun = tree.fun
+    val newArgs: List[tpd.Tree] = tree.args.zip(fun.tpe.dealias.firstParamTypes).map(x => adaptBottom(x._1, x._2))
+    val changeNeeded = tree.args == newArgs // cpy.Apply does not check if elements are the same,
```
smarter: I think this should be `tree.args != newArgs` or I'm misunderstanding what you're trying to do (if I'm right then you should add a test case for this)

src/dotty/tools/backend/jvm/BottomTypes.scala
```diff
@@ -0,0 +1,81 @@
...
+                                            // it only does `eq` on lists as whole
+    if (changeNeeded) cpy.Apply(tree)(fun = fun, args = newArgs)
```
smarter: Why `fun = fun`?

tests/pos/i828.scala
```diff
@@ -0,0 +1,3 @@
+object X {
+  val x: Int = null.asInstanceOf[Nothing]
```
smarter: You should have test cases for the various nodes modified by `BottomTypes` and test cases for `Null` too.

src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -250,7 +250,7 @@ object Erasure extends TypeTestsCasts{
         case MethodType(Nil, _) if tree.isTerm =>
           adaptToType(tree.appliedToNone, pt)
         case tpw =>
-          if (pt.isInstanceOf[ProtoType] || tree.tpe <:< pt)
+          if (pt.isInstanceOf[ProtoType] || ((!pt.isValueType || !isNonJVMBottomType(tree.tpe)) && (tree.tpe <:< pt)))
```
smarter: The documentation of this method should be updated to explain what it does to bottom types too.

smarter:
`Nothing` is supposed to have no value, so I don't know if allowing `null.asInstanceOf[Nothing]` to run without any error is a good thing or not, maybe it doesn't matter.

src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -263,6 +263,12 @@ object Erasure extends TypeTestsCasts{
           else
             cast(tree, pt)
       }
+
+
+    /** Is `tpe` a type which is a bottom type for Dotty, but not a bottom type for JVM */
+    def isNonJVMBottomType(tpe: Type)(implicit ctx: Context): Boolean = {
+      tpe.derivesFrom(ctx.definitions.NothingClass) || tpe.derivesFrom(ctx.definitions.NullClass)
+    }
   }
```
odersky: Indeed, why not use Definitions#isBottomType?

DarkDimius: Definitions#isBottomType(tp) is defined to be `tp.symbol \in {NothingClass, NullClass}` 
It does not work for types that do not return a usable symbol, such as And\Or types.

DarkDimius: Should I update Definitions#isBottomType to account for `Null | Nothing` and `Int & Nothing`?

odersky: No, I think it's OK to leave the two separate. Maybe rename the second to
`derivesFromBottomType`
to make clear that refinements and & / | count as well?


On Fri, Dec 18, 2015 at 6:32 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> In src/dotty/tools/dotc/transform/Erasure.scala
> <https://github.com/lampepfl/dotty/pull/1002#discussion_r48048942>:
>
> > @@ -263,6 +263,12 @@ object Erasure extends TypeTestsCasts{
> >            else
> >              cast(tree, pt)
> >        }
> > +
> > +
> > +    /** Is `tpe` a type which is a bottom type for Dotty, but not a bottom type for JVM */
> > +    def isNonJVMBottomType(tpe: Type)(implicit ctx: Context): Boolean = {
> > +      tpe.derivesFrom(ctx.definitions.NothingClass) || tpe.derivesFrom(ctx.definitions.NullClass)
> > +    }
> >    }
>
> Should I update Definitions#isBottomType to account for Null | Nothing
> and Int & Nothing?
>
> —
> <https://github.com/lampepfl/dotty/pull/1002/files#r48048942>.
>



-- 
Martin Odersky
EPFL

src/dotty/tools/backend/jvm/BottomTypes.scala
DarkDimius: this method does not have default arguments.

src/dotty/tools/dotc/transform/Erasure.scala
DarkDimius: Will do.

DarkDimius:
@smarter 
>Nothing is supposed to have no value, so I don't know if allowing null.asInstanceOf[Nothing] to run without any error is a good thing or not, maybe it doesn't matter.

I do not have a strong opinion on this either. So I've implemented `be compatible with scalac` mode.

```scala
class A {
 val s: Nothing  = null.asInstanceOf[Nothing]
}
```
emmits:
```java
public class A {
    private final Nothing. s = null;

    public Nothing. s() {
        return this.s;
    }
}
```

PS: the `.` in `Nothing.` is not a coincidence. It is a late transformation done by GenBCode.

smarter:
Being compatible with scalac seems OK here, but note that the following fails at runtime in `scalac` but works with your patch:
```scala
object X {
    val x: Int = null.asInstanceOf[Nothing]
  }
}
``

**Comment by smarter** [comment]
And now that I think of it, this might be a good think because in the following:
```scala
object X {
    val x: Int = null.asInstanceOf
  }
}
```
The type parameter of `asInstanceOf` will be instantiated to `Nothing`, so it crashes at runtime in `scalac` but works with dotty.

src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -263,6 +263,12 @@ object Erasure extends TypeTestsCasts{
           else
             cast(tree, pt)
       }
+
+
+    /** Is `tpe` a type which is a bottom type for Dotty, but not a bottom type for JVM */
+    def isNonJVMBottomType(tpe: Type)(implicit ctx: Context): Boolean = {
+      tpe.derivesFrom(ctx.definitions.NothingClass) || tpe.derivesFrom(ctx.definitions.NullClass)
+    }
```
odersky: I now did a similar change in definitions (f780a371a3716c6917a510ed72a5927704634bb3). 
So it's best if you cherry-pick the commit and use defn.isBottomType.

DarkDimius:
I've addressed most comments except the `(null.asInstanceOf[Nothing]): Int)` one.
I do not have a clear understanding of how this works in scalac and why is it different from the other case.
If you someone is interested to get to the bottom of this, It could be done in a different PR. This one fixes a compiler crash so I'd like to get this in.

sjrd:
FTR, in Scala.js we recently decided to go against what scalac/JVM does for `null.asInstanceOf[Nothing]`. Now it always throws immediately (as an undefined behavior). See the PR https://github.com/scala-js/scala-js/pull/2126 fixing issue https://github.com/scala-js/scala-js/issues/2125

odersky:
Let's either rebase and try again or close.

PR #1004, opened 2015-12-20
Subtle changes to TypeComarer needed for Linker.

DarkDimius:
@Odersky those are the changes that are result of our discussion last Friday.

Only 9714565 is new here, it allows NamedTypes to static modules to be subtypes
of ThisTypes of static modules. This was already allowed for TermRefs.
The commit comment has a motivation example from stdlib.

@Odersky please review.

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -1695,8 +1695,9 @@ object Types {
     override def computeHash = unsupported("computeHash")
   }
 
-  final class TermRefWithFixedSym(prefix: Type, name: TermName, val fixedSym: TermSymbol) extends TermRef(prefix, name) with WithFixedSym
-  final class TypeRefWithFixedSym(prefix: Type, name: TypeName, val fixedSym: TypeSymbol) extends TypeRef(prefix, name) with WithFixedSym
+  // Those classes are non final as Linker extends them.
+  class TermRefWithFixedSym(prefix: Type, name: TermName, val fixedSym: TermSymbol) extends TermRef(prefix, name) with WithFixedSym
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1696,7 +1696,20 @@ object Types {
...
   final class TermRefWithFixedSym(prefix: Type, name: TermName, val fixedSym: TermSymbol) extends TermRef(prefix, name) with WithFixedSym
-  final class TypeRefWithFixedSym(prefix: Type, name: TypeName, val fixedSym: TypeSymbol) extends TypeRef(prefix, name) with WithFixedSym
+  class TypeRefWithFixedSym(prefix: Type, name: TypeName, val fixedSym: TypeSymbol) extends TypeRef(prefix, name) with WithFixedSym
+
+  /* used by linker */
+  final class TypeRefWithFixedSymAndInfo(prefix: Type, name: TypeName, fixedSym: TypeSymbol, val underl: Type) extends TypeRefWithFixedSym(prefix, name, fixedSym) {
+    override def derivedSelect(prefix: Type)(implicit ctx: Context): Type =
+      if (prefix eq this.prefix) this
+      else {
+        ???
```
odersky: ???

DarkDimius: In linker this didn't yet happen: the prefix that this type starts with is very precise type and it should not loose precision by getting less precise prefix.

DarkDimius: If this happens, I would need to start looking into changes to derivedSelect from https://github.com/lampepfl/dotty/pull/1001

DarkDimius:
@odersky please have a look and merge if LGT you.

odersky:
LGTM.

retronym:
For the record, I was working in this space in scalac recently: https://github.com/scala/scala/compare/2.12.x...retronym:ticket/9542-combo

PR #1007, opened 2015-12-20
Fix typo in command run by Jenkins

smarter:
The effect of this typo was that Ycheck:tailrec was never run on Jenkins.

Review by @DarkDimius .


DarkDimius:
LGTM

PR #1008, opened 2015-12-22
Fix off by 2 error for symbol positions

odersky:
Review by @smarter


odersky:
The neg test was removed because it was no longer illegal. It was not tested for a long time, just stayed in the neg directory. I found out by accident when I tried some compiles to diagnose the position error.

smarter:
LGTM.

PR #1010, opened 2015-12-23
Fix #1009: Do not forget to skolemize some types

### LINKED ISSUE #1009: New asSeenFrom does not skolemize some types that should be skolemized
#999 changed the way `asSeenFrom` works and it introduced a bug:
```scala
class Contra[-T]

class Foo {
  val foo: Contra[this.type] = new Contra[this.type]
}
object Test {
  def test: Unit = {
    val e1 = new Foo
    val f1: Contra[Foo] = e1.foo // error, as expected
    var e2 = new Foo
    val f2: Contra[Foo] = e2.foo // should be an error but now succeeds
  }
}
```
The problem is that we [removed this line](https://github.com/lampepfl/dotty/pull/999/files#diff-13bc6129f640e14c18ed0706b980ff1eL47):
```scala
val m = if (isLegalPrefix(pre)) null else new AsSeenFromMap(pre, cls)
```
but `asSeenFrom#toPrefix` still [assumes that the prefix is always legal when the map is null](https://github.com/lampepfl/dotty/pull/999/files#diff-13bc6129f640e14c18ed0706b980ff1eR67).

smarter:
Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/src/dotty/tools/dotc/core/TypeOps.scala
@@ -102,7 +102,7 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
             asSeenFrom(tp.parent, pre, cls, theMap),
             tp.refinedName,
             asSeenFrom(tp.refinedInfo, pre, cls, theMap))
-        case tp: TypeAlias if theMap == null => // if theMap exists, need to do the variance calculation
+        case tp: TypeAlias if tp.variance == 1 => // if variance != 1, need to do the variance calculation
           tp.derivedTypeAlias(asSeenFrom(tp.alias, pre, cls, theMap))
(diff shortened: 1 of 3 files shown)


smarter:
PR updated to reflect https://github.com/dotty-staging/dotty/commit/a7b4c14990dc8ec789cbb4d735ff9fb6bde1d16b#commitcomment-15152817

smarter:
> [error] java.lang.OutOfMemoryError: GC overhead limit exceeded

@DarkDimius: should we raise the amount of memory we use? Or should we investigate why we're using so much memory?

smarter:
Updated the PR with a small optimization: do not allocate an `AsSeenFromMap` if `TypeAlias#variance == 1`.

PR #1012, opened 2015-12-30
Fix check bounds

odersky:
I noticed that we do not check all bounds of type parameters. Fixing this revealed another
problem when compiling stdlib which this commit also fixes. However, bounds checking
in PostTyper now causes deep subtypes, so stdlib has to be compiled without -Yno-dee-subtypes.
Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -180,6 +180,15 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
         case tree: DefDef =>
           transformAnnots(tree)
           superAcc.wrapDefDef(tree)(super.transform(tree).asInstanceOf[DefDef])
+        case tree: TypeDef =>
+          transformAnnots(tree)
+          val sym = tree.symbol
+          val tree1 =
+            if (sym.isClass) tree
+            else {
+              cpy.TypeDef(tree)(rhs = TypeTree(tree.symbol.info))
+            }
+          super.transform(tree1)
         case tree: MemberDef =>
           transformAnnots(tree)
           super.transform(tree)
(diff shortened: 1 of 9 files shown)


smarter:
I haven't looked into it much yet, but I think the deep subtype error is worth investigating, it happens just by running:
```scala
./bin/dotc -Yno-deep-subtypes ./scala-scala/src/library/scala/collection/generic/ParMapFactory.scala
```
however, if I enable the `subtyping` `Printer` in `Printers.scala` and run:
```scala
./bin/dotc -Ylog:all -Yno-deep-subtypes ./scala-scala/src/library/scala/collection/generic/ParMapFactory.scala
```
Then the error disappears! So clearly it's possible to typecheck this without deep subtyping

odersky:
An explanation could be that LazyRefs forcing creates additional subtypes.
Printing forces ahead of time, so there are fewer subtype tests.


On Wed, Dec 30, 2015 at 5:51 PM, Guillaume Martres <notifications@github.com
> wrote:

> I haven't looked into it much yet, but I think the deep subtype error is
> worth investigating, it happens just by running:
>
> ./bin/dotc -Yno-deep-subtypes ./scala-scala/src/library/scala/collection/generic/ParMapFactory.scala
>
> however, if I enable the subtyping Printer in Printers.scala and run:
>
> ./bin/dotc -Ylog:all -Yno-deep-subtypes ./scala-scala/src/library/scala/collection/generic/ParMapFactory.scala
>
> Then the error disappears! So clearly it's possible to typecheck this
> without deep subtyping
>
> —
> <https://github.com/lampepfl/dotty/pull/1012#issuecomment-168032835>.
>



-- 
Martin Odersky
EPFL

smarter:
I think I got it, see #1014.

odersky:
Superseded by #1014

PR #1013, opened 2015-12-30
Add a compiler.properties resource file like in Scala 2

smarter:
Review by @DarkDimius .


DarkDimius:
@smarter would you kindly update dotc.build test to carry over the generated file to bootstrapped dotty?

Otherwise LGTM

smarter:
> @smarter would you kindly update dotc.build test to carry over the generated file to bootstrapped dotty?

This is slightly tricky because the generated file is in `target/scala-2.11/resource_managed`, but if we compile dotty with dotty it'll be in `target/scala-0.1/resource_managed`, since this resource is only useful for using dotty from sbt, and since the jar generated by `test/dotc/build.scala` is incomplete and cannot be used from sbt anyway, I'd rather not hack on this for now.

PR #1014, opened 2015-12-30
Avoid infinite subtyping checks when intersecting denotations

smarter:
This PR is based on #1012

Review by @odersky, I don't actually understand why `safeAnd` is safer than `&` but it seems to work well in practice, could you explain why it works so that I can add a comment?

### DIFF
--- a/src/dotty/tools/dotc/typer/Checking.scala
+++ b/src/dotty/tools/dotc/typer/Checking.scala
@@ -48,6 +48,18 @@ object Checking {
   def checkBounds(args: List[tpd.Tree], poly: PolyType)(implicit ctx: Context): Unit =
     checkBounds(args, poly.paramBounds, _.substParams(poly, _))
 
+  /** Check all AppliedTypeTree nodes in this tree for legal bounds */
+  val boundsChecker = new TreeTraverser {
+    def traverse(tree: Tree)(implicit ctx: Context) = tree match {
+      case AppliedTypeTree(tycon, args) =>
+        val tparams = tycon.tpe.typeSymbol.typeParams
+        val bounds = tparams.map(tparam =>
+          tparam.info.asSeenFrom(tycon.tpe.normalizedPrefix, tparam.owner.owner).bounds)
+        checkBounds(args, bounds, _.substDealias(tparams, _))
+      case _ => traverseChildren(tree)
+    }
+  }
+
   /** Check that `tp` refers to a nonAbstract class
    *  and that the instance conforms to the self type of the created class.
    */
--- a/src/dotty/tools/dotc/typer/Checking.scala
+++ b/src/dotty/tools/dotc/typer/Checking.scala
@@ -50,13 +50,16 @@ object Checking {
 
   /** Check all AppliedTypeTree nodes in this tree for legal bounds */
   val boundsChecker = new TreeTraverser {
-    def traverse(tree: Tree)(implicit ctx: Context) = tree match {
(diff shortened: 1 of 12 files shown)


odersky:
@smarter This kind of recursion, which is problematic, is only allowed for F-bounded upper types. So we always have a typebound.

src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -48,6 +48,18 @@ object Checking {
...
     checkBounds(args, poly.paramBounds, _.substParams(poly, _))
 
+  /** Check all AppliedTypeTree nodes in this tree for legal bounds */
+  val boundsChecker = new TreeTraverser {
+    def traverse(tree: Tree)(implicit ctx: Context) = tree match {
+      case AppliedTypeTree(tycon, args) =>
+        val tparams = tycon.tpe.typeSymbol.typeParams
+        val bounds = tparams.map(tparam =>
+          tparam.info.asSeenFrom(tycon.tpe.normalizedPrefix, tparam.owner.owner).bounds)
+        checkBounds(args, bounds, _.substDealias(tparams, _))
```
smarter: I think you need to call `traverseChildren(tree)` here too, otherwise the following compiles without errors:
```scala
class OnlyInt[T <: Int]

object Test {
  type T = List[OnlyInt[String]]
}
```

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -871,6 +873,19 @@ class Namer { typer: Typer =>
       sym.info = NoCompleter
       sym.info = checkNonCyclic(sym, unsafeInfo, reportErrors = true)
     }
+
+    // Here we pay the price for the cavalier setting info to TypeBounds.empty above.
```
smarter: The TypeBounds.empty thing is tricky and I don't really understand why it's done and how it works, I think it'd be great to explain it in more details, for example there's a comment above that says:
```scala
// The scheme critically relies on an implementation detail of isRef, which
// inspects a TypeRef's info, instead of simply dealiasing alias types.
```
But why would this scheme not work if isRef was implemented differently? (I don't even see how you could implement it differently, how can you know if an alias TypeRef refers to a particular symbol without looking at its info?)

smarter:
Otherwise LGTM, I rebased the PR, addressed your style comment, and added a description to `safe_&`: https://github.com/lampepfl/dotty/pull/1014/files#diff-ca3386647f5386b6f70e99f1a9e188cdR710

smarter:
Rebased, note that there are two unadressed comments that github is hiding:
- https://github.com/lampepfl/dotty/pull/1014#discussion_r49282553
- https://github.com/lampepfl/dotty/pull/1014#discussion_r49282619

odersky:
@smarter Is this good to merge now?

smarter:
@odersky : please review https://github.com/lampepfl/dotty/commit/bb59d81df80bb08a20e7b0547bb6a1e0c154c1fd and then we can merge. (we'll document the empty TypeBounds trick another time).

smarter:
Oops, pushed the wrong commit, please review https://github.com/lampepfl/dotty/commit/8141e6239ec3bb860b3013f51572994102047c09 instead.

smarter:
Merging, the junit failure is due to https://github.com/scala/scala-jenkins-infra/pull/158 but you can check that the tests actually succeeded so it's safe.

PR #1015, opened 2016-01-04
bin/dotc: fix "-d" argument

smarter:
It's already used to specify an output folder, don't try to interpret it
as a shortcut for -debug

Review by @DarkDimius


DarkDimius:
LGTM

PR #1016, opened 2016-01-04
Build.scala: only enable features we use

smarter:
-language:_ does not work when compiling dotty with dotty because it
implies -language:keepUnions which prevents dotty from typechecking

Review by @DarkDimius


DarkDimius:
LGTM

Issue #1017, opened 2016-01-04
If a method foo overrides a method foo(), then it cannot be called as foo()


smarter:
This is valid in Scala 2 but not in Dotty:

``` scala
class A {
  def foo(): Int = 1
}
class B extends A {
  override def foo: Int = 2
}

object Test {
  def main(args: Array[String]): Unit = {
    val b = new B
    println(b.foo())
  }
}
```

@odersky : Do you think this should be legal in Dotty? Otherwise, should we at least support it in Scala 2 mode ?

sjrd:
If the override itself is legal, then so should the call `b.foo()`.

Overriding a 0-arg method with a getter-style def is mainly meant for Java interop, as I see it. People want to be able to `override def toString`, for example, not `override def toString()`. It is for the same reason that `x.toString` itself is valid, IMO. In pure Scala such things should never have been needed.

odersky:
It would be hard to allow this, even in Scala 2 mode. def foo() and def foo have the same signatures
so we keep the overriding one as the only truth. I think it's OK to leave it as is. The reasoning would be: If I intentionally override `def foo()` with `def foo` then it's because I intend `foo` to be called without
`()`. So this intention should be honored by the callers.

odersky:
I am not against supporting it in Scala2 mode, if we can do so easily. But right now it does not look like we can.

sjrd:
The combination of 1) allowing the `override def foo` but 2) only allowing calling `x.foo` in that case, is, IMO, very dangerous. It will break a lot of code *in dependencies of libraries down the line*.

Let's say I have a codebase in library A, where the coding style is to remove the `()` for all non-side-effecting "getters", including those from Java, and including in overridings. The new interpretation would not cause harm to library itself. It would still compile, its tests would pass, and it would get published.

Now, a project B which used to depend on library A is updated. In B's codebase, the coding style is to be very strict about keeping the `()` at all times if they were defined as such (e.g., from Java), event at call site. Now B's code referring to methods from A breaks! And A never had any idea this would break its users.

I think this is a bad situation.

IMO we should either 1) keep Scala 2's behavior (as much as I dislike it) or 2) *disallow* overriding a method with `()` with a method without `()`. Note that the second proposition is *also* subject to the "A is fine but B is broken" problem, but much less often so: it requires B to actually extend classes from A (not just use them). The decision in the middle is asking for trouble.

odersky:
[...]

PR #1018, opened 2016-01-04
VCArrayPrototype: replace override "def clone" by "def clone()"

smarter:
This prevented Dotty from bootstrapping when the file arguments were
given in a certain order because of #1017. Regardless of what we do
regarding #1017, it makes sense to use "def clone()" here.

Review by @DarkDimius


DarkDimius:
LGTM

Issue #78, opened 2014-03-17
typer hang using `type` keyword on a class type param


retronym:
```scala
class B[type X] { def foo(x: X) = ??? }
```

```
"main" prio=5 tid=7fe537800800 nid=0x104197000 runnable [104191000]
   java.lang.Thread.State: RUNNABLE
	at dotty.tools.dotc.core.Types$Type.termSymbol(Types.scala:251)
	at dotty.tools.dotc.typer.Checking$class.checkValue(Checking.scala:42)
	at dotty.tools.dotc.typer.Typer.checkValue(Typer.scala:51)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedIdent$1.apply(Typer.scala:263)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedIdent$1.apply(Typer.scala:72)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.typer.Typer.typedIdent(Typer.scala:72)
```

The cycle:
```
(this, this.underlying(ctx), this.underlying(ctx).underlying(ctx), this.underlying(ctx).underlying(ctx).underlying(ctx))

_1 = {dotty.tools.dotc.core.Types$CachedTypeBounds@2600}"TypeAlias(TypeRef(ThisType(class B),X))"
_2 = {dotty.tools.dotc.core.Types$CachedTypeRef@2642}"TypeRef(ThisType(class B),X)"
_3 = {dotty.tools.dotc.core.Types$CachedTypeBounds@2600}"TypeAlias(TypeRef(ThisType(class B),X))"
_4 = {dotty.tools.dotc.core.Types$CachedTypeRef@2642}"TypeRef(ThisType(class B),X)"
```

felixmulder:
Fixed in https://github.com/lampepfl/dotty/pull/1072

PR #1019, opened 2016-01-12
Add named params - first iteration

odersky:
This PR adds named class parameters `[type A]` and named arguments to
applied types `T[A = U]`. 

This is not quite complete yet. We are still missing named arguments in method applications and patterns. Also, we need more tests in particular in what concerns type inference.


DarkDimius:
> [info] !! 222 - pos/printing_v1                           [compilation failed]

That's strange, as the very same file succeeded being compiled by unit-tests.

odersky:
We get OutOfMemoryErrors (GC overhead limit exceeded). Is there something one can do to fix this?

odersky:
/rebuild -- trying to find out whether this is random or not.

odersky:
rebased to master

odersky:
We see:

Native memory allocation (mmap) failed to map 2463629312 bytes for committing reserved memory.

odersky:
Tests pass locally.

smarter:
Try rebasing on top of master to get rid of the errors related to t7880.scala

smarter:
Note that no junit build will succeed until https://github.com/scala/scala-jenkins-infra/pull/158 is merged and deployed, sorry about that.

odersky:
Get this in now? It's been lingering for a long time.

odersky:
Rebased to master. This is ready to go in now. I know it is a big PR but it is an important one. If there are no reviews I'll merge.

smarter:
I'll do another pass on this today

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -759,9 +759,38 @@ object Parsers {
...
+          else Nil
+        first :: rest
+      }
+      if (namedOK && in.token == IDENTIFIER)
+        argType() match {
+          case Ident(name) if in.token == EQUALS =>
+            in.nextToken()
+            otherArgs(NamedArg(name, argType()), namedTypeArg)
+          case firstArg =>
+            if (in.token == EQUALS) println(s"??? $firstArg")
```
smarter: Don't use println in the compiler

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -97,17 +97,37 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
...
+      if (tree.isType) TypeTree(tree.tpe).withPos(tree.pos)
+      else literalize(tree.tpe)
+    tree match {
+      case tree: TypeTree =>
+        tree
+      case AppliedTypeTree(tycon, args) =>
+        // If `args` is a list of named argiments, return corresponding type parameters,
+        // or abstract types otherwise return type parameters unchanged
+        def matchNamed(tparams: List[TypeSymbol], args: List[Tree]): List[Symbol] = args match {
+          case (_: NamedArg) :: _ =>
```
smarter: This pattern match would be simpler if it was done on `args.head` instead of `args`

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -600,12 +603,19 @@ trait Applications extends Compatibility { self: Typer =>
...
+
   def typedTypeApply(tree: untpd.TypeApply, pt: Type)(implicit ctx: Context): Tree = track("typedTypeApply") {
-    var typedArgs = tree.args mapconserve (typedType(_))
+    val isNamed = hasNamedArg(tree.args)
+    var typedArgs = if (isNamed) typedNamedArgs(tree.args) else tree.args.mapconserve(typedType(_))
     val typedFn = typedExpr(tree.fun, PolyProto(typedArgs.tpes, pt))
     typedFn.tpe.widen match {
       case pt: PolyType =>
-        if (typedArgs.length <= pt.paramBounds.length)
+        if (typedArgs.length <= pt.paramBounds.length && !isNamed)
```
smarter: Why are named arguments special here? Don't we need to eta-expand them too?

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -302,9 +307,44 @@ trait TypeAssigner {
...
+              case PolyParam(`pt`, idx) => normArgs(idx)
+              case _ => mapOver(t)
+            }
+          }
+          val resultType1 = transform(pt.resultType)
+          if (gapBuf.isEmpty) resultType1
+          else {
+            val gaps = gapBuf.toList
+            pt.derivedPolyType(
+              gaps.map(paramNames.filterNot(argMap.contains)),
```
smarter: Why call `gaps.map` since `gaps` is always equal to `List(0, 1, 2, ...)` ?

src/dotty/tools/dotc/transform/PostTyper.scala
smarter: Alternatively, you could use `hasNamedArg` here.

src/dotty/tools/dotc/typer/TypeAssigner.scala
odersky: Yes, that logic was screwed up.

src/dotty/tools/dotc/typer/Applications.scala
odersky: Good point. I did not want to deal with interactions of named params and old-style hk types at first, but in the end we'll have to.

odersky:
Reviewer comments so far have been addressed. Are there others coming?

odersky:
Also fixes #78

odersky:
Ready to merge?

odersky:
Subsumed by #1072

PR #1020, opened 2016-01-12
Fix caching bug: don't assume that tvars instantiation cannot be retracted

smarter:
When TypeVar#inst is empty but an instantiation exists in the typer
state, we should set ephemeral to true, because this instantiation will
be retracted if we throw away the current typer state.

This makes hkrange.scala pass, it compiled before but the type parameter
of `f` was inferred to be `Nothing` because of this bug, and this failed
Ycheck.

For anyone who wonders how caching bugs manifest themselves, here's what
happened in details in hkrange.scala:

1. In an ExploreTyperState we set `CC` to be `IndexedSeq` in the
   constraint set
2. In that same typer state the TypeRef `CC[Int]` (it's a TypeRef
   because `CC` is a type lambda) gets the denotation `IndexedSeq[Int]`,
   which is correct, but the denotation is cached since `ephemeral` is
   false, which is wrong.
3. Later, we retract the ExplorerTyperState, so `CC` is uninstantiated
   again and unconstrained.
4. Then we do the subtyping check `CC[Int] <:< IndexedSeq[Int]`, because
   the denotation of `CC[Int]` was cached, this returns true, but `CC` stays
   unconstrained.
5. This means that when we instantiate `CC`, we get `Nothing`

After this fix, the TypeRef denotation is no longer cached, so when we
do `CC[Int] <:< IndexedSeq[Int]`, `CC` gets constrained as expected.

Review by @odersky


smarter:
@DarkDimius : note that I'm getting test failures (https://scala-ci.typesafe.com/job/dotty-master-validate-main/910/console) with the non-bootstrapped partest

DarkDimius:
Good to know that failures are independent from bootstrap.

odersky:
needs to be rebased to master

smarter:
Rebased.

odersky:
LGTM

PR #1021, opened 2016-01-12
Improve migration message for Scala 2 procedures

smarter:
It's not correct to just add `=` you also need to specify the return
type to be `Unit` otherwise things may not work as expected, this is
especially important for a `main` method.

Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1732,7 +1732,7 @@ object Parsers {
      */
     def defDefOrDcl(mods: Modifiers): Tree = atPos(tokenRange) {
       def scala2ProcedureSyntax =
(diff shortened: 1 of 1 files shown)


odersky:
Not sure about this. `: Unit =` is needlessly heavy, in particular for local procedures. So advising it unconditionally is a bit of an overkill.

sjrd:
I don't think there are many `Unit`-returning local procedures in typical codebases. Except possibly tail-recursive `Unit`-returning procedures, but those *need* the `: Unit`.

It's fine if users want to remove the `: Unit` part themselves when they think it's too heavy. But the compiler recommendation should at least always recommend something *correct*. And adding only `=` will sometimes (and more often than not IMO) be wrong.

odersky:
OK, you have convinced me. LGTM.

PR #1025, opened 2016-01-14
Fix #974

### LINKED ISSUE #974: Higher-kinded F-bounded types are not recognized
This causes a stack overflow:
```scala
class Foo[A]
class Bar[CC[X] <: Foo[CC[X]]]
```
The reason why is explained in https://github.com/lampepfl/dotty/issues/943#issuecomment-159323775

odersky:
Review by @smarter


odersky:
Will split from add-named-params

PR #1026, opened 2016-01-14
Don't force NamedType denotations in containsRefinedThis

### LINKED ISSUE #974: Higher-kinded F-bounded types are not recognized

odersky:
containsRefinedThis inspects symbols and infos of named types in order
to avoid needless traversals. As i974 shows, this can lead to infinite
recursions. The fix is not to force a NamedType denotation, assume the
worst and traverse the prefix if a NamedType is not yet populated
with a denotation.

Fixes #974 and makes MutableSortedSetFactory in stdlib compile.

Review by @smarter

### DIFF
--- a/tests/neg/i974.scala
+++ b/tests/neg/i974.scala
@@ -1,0 +1,8 @@
+trait Foo[T <: Bar[T]#Elem] // error: illegal cyclic reference
+trait Bar[T] {
+  type Elem = T
+}
+trait Foo2[T <: Bar2[T]#Elem] // error: illegal cyclic reference
+trait Bar2[T] {
+  type Elem = T
+}
--- a/scripts/common
+++ b/scripts/common
@@ -16,4 +16,4 @@ update() {
 
 export LC_ALL=en_US.UTF-8
 
-sbtArgs="-Ddotty.jenkins.build=yes -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13 -J-Xmx3G"
+sbtArgs="-Ddotty.jenkins.build=yes -J-Xmx1500 -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13"
--- a/scripts/common
+++ b/scripts/common
@@ -16,4 +16,4 @@ update() {
 
 export LC_ALL=en_US.UTF-8
 
-sbtArgs="-Ddotty.jenkins.build=yes -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13"
+sbtArgs="-Ddotty.jenkins.build=yes -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13 -J-Xmx3G"
--- a/scripts/common
+++ b/scripts/common
@@ -16,4 +16,4 @@ update() {
 
 export LC_ALL=en_US.UTF-8
 
-sbtArgs="-Ddotty.jenkins.build=yes -J-Xmx1500 -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13"
+sbtArgs="-Ddotty.jenkins.build=yes -J-Xmx1500m -Dfile.encoding=UTF-8 -ivy $baseDir/ivy2 -Dsbt.global.base=$HOME/.sbt/0.13 -sbt-dir $HOME/.sbt/0.13"
--- a/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -695,11 +695,13 @@ class TypeApplications(val self: Type) extends AnyVal {
       case RefinedThis(tp) =>
         tp eq target
       case tp: NamedType =>
-        if (tp.symbol.isClass) !tp.symbol.isStatic && recur(tp.prefix)
-        else tp.info match {
-          case TypeAlias(alias) => recur(alias)
-          case _ => recur(tp.prefix)
-        }
+        if (tp.denotationIsCurrent)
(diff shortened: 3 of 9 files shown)


odersky:
It was a GC overhead limit exceeded again. Is there a way to increase the heap or alternatively reduce the level of parallelism?

odersky:
Out of heap space this time. We need to change the build script to take more heap. Can someone do this please?

smarter:
I think this PR only addresses the symptoms and not the root of the problem, which is that we do not check for cycles in the prefixes of TypeRefs which are RefinementTypes. For example, the following causes an infinite loop instead of printing `illegal cyclic reference: upper bound Bar[T]#Elem of type T refers back to the type itself`:
```scala
trait Foo[T <: Bar[T]#Elem]
trait Bar[T] {
  type Elem = T
}
```

odersky:
The two problems are unrelated. #974 caused the StackOverflow before the
cycle check was even run.
@smarter, can you make this new failing test into a separate issue?

On Thu, Jan 14, 2016 at 5:05 PM, Guillaume Martres <notifications@github.com
> wrote:

> I think this PR only addresses the symptoms and not the root of the
> problem, which is that we do not check for cycles in the prefixes of
> TypeRefs which are RefinementTypes. For example, the following causes an
> infinite loop instead of printing illegal cyclic reference: upper bound
> Bar[T]#Elem of type T refers back to the type itself:
>
> trait Foo[T <: Bar[T]#Elem]trait Bar[T] {
>   type Elem = T
> }
>
> —
> <https://github.com/lampepfl/dotty/pull/1026#issuecomment-171683787>.
>



-- 
Martin Odersky
EPFL

smarter:
Are you sure? If I apply the following patch on top of master: https://github.com/dotty-staging/dotty/commit/7d9eb3f55908b3d905f51e83dd90ff1b83543145 then ` tests/pos/i974.scala` succeeds (however, other tests like `./tests/pos/tcpoly_ticket2096.scala` now fail)

odersky:
@smarter Maybe you are right. I fixed checkNonCyclic and your example compiles OK (i.e. with errors) now.

odersky:
The heap problem has gone away. We are back to the previous 

pos/printing: compilation failed

without any further diagnostic. It's super urgent that we fix this, no progress is possible without it.

smarter:
@odersky : Maybe try rebasing on master where the partest jenkins job does not use bootstrapped dotty anymore.

DarkDimius:
Ok. I was able to get to the archives of workspace. 
But the log does no say much. eg, for 
https://scala-ci.typesafe.com/job/dotty-master-validate-partest/906/console

where printing-pos failed, the compiler output only contains:
```
compiling /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/Disambiguation.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/PlainPrinter.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/Printer.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/Printers.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/RefinedPrinter.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/Showable.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/Texts.scala /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/printing/package.scala
options: -Xprint-types -Ytest-pickler -Ystop-after:pickler -pagewidth 160 -Yno-deep-subtypes -Yno-double-bindings -d ./tests/partest-generated/pos/printing-pos.obj -Ycheck:tailrec,resolveSuper,mixin,restoreScopes,labelDef
```
Not informative.

DarkDimius:
To get the archive of workspace:
https://scala-ci.typesafe.com/job/dotty-master-validate-partest/906/ than click the `Build Artifacts` link.
It should have a `*.clog` file for every failed test:

```
dark@reco ~/Downloads/archive 2 $ find . -iname "*.clog"
./tests/partest-generated/pos/printing-pos.clog
```

smarter:
> @smarter Maybe you are right. I fixed checkNonCyclic and your example compiles OK (i.e. with errors) now.

So should we revert the original fix since it's not actually needed?

tests/neg/i974.scala
```diff
@@ -0,0 +1,8 @@
+trait Foo[T <: Bar[T]#Elem] // error: illegal cyclic reference
+trait Bar[T] {
+  type Elem = T
+}
+trait Foo2[T <: Bar2[T]#Elem] // error: illegal cyclic reference
```
smarter: Did you intend to have the same test twice, with just the names changed? (`Foo -> Foo2`, `Bar -> Bar2`)

odersky: Yes, because 1 expected error also succeeds if the file is not found.

odersky:
@smarter I think the checkNonCyclic fix is worth doing anyway.

smarter:
> @smarter I think the checkNonCyclic fix is worth doing anyway.

You mean the [containsRefinedThis](https://github.com/dotty-staging/dotty/commit/9a3dbf9a6fb8e2c09ed89507e57ef48363e21dad), right? Anyway, LGTM once the tests pass.

smarter:
> Ok. I was able to get to the archives of workspace
But the log does no say much

@DarkDimius : looking at https://github.com/lampepfl/dotty/blob/master/test/dotty/partest/DPDirectCompiler.scala#L29 I see that it's trying to redirect the output by using a custom Context, but using a custom Context is currently broken, so I don' t think that can work: https://github.com/lampepfl/dotty/blob/da661c8a509efaa1c0ff926b0907c32cd37708a5/src/dotty/tools/dotc/Driver.scala#L37-L39
It also won't include the output of calls to println, etc. It might be simpler to just always redirect stdout and stderr using https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#setOut-java.io.PrintStream- and https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#setErr-java.io.PrintStream-

smarter:
>  Anyway, LGTM once the tests pass

Actually, there's a missing change:remove the comment at https://github.com/lampepfl/dotty/pull/1026/files#diff-2fe35c59082b97d7128120593f9ec2e7R172 about `#Apply` which is no longer accurate.

odersky:
Superseded by #1028

PR #1027, opened 2016-01-15
Change early typeparams

odersky:


odersky:
Fixes #970 and a couple other problems that were discovered after applying the fix. Review by @smarter.

DarkDimius:
>  validate-partest-bootstrapped — [4] FAILURE. Took 9 s. Say /rebuild on PR to retry *spurious* failure. 

Need to rebase over current master.

smarter:
The rebase ate my comment so to recap: https://github.com/dotty-staging/dotty/commit/5f1e298747cd8a6136cad04f6fa3eba83b63c0ba is not needed on master, it's only needed because https://github.com/dotty-staging/dotty/commit/beb6fce0ac7cd1c7adc93318915c6c0ae48e4503 is applied so I wonder if there's a bug in https://github.com/dotty-staging/dotty/commit/beb6fce0ac7cd1c7adc93318915c6c0ae48e4503 and I'd be in favor of reverting it since it's not needed anymore to avoid cyclic references (cf. https://github.com/lampepfl/dotty/pull/1026#issuecomment-171726922).

PR #1028, opened 2016-01-15
Change early typeparams2

odersky:


odersky:
Fixes #970 and a couple other problems that were discovered after applying the fix. Review by @smarter.

Supersedes #1027

src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -240,15 +241,21 @@ class TypeApplications(val self: Type) extends AnyVal {
...
-          // abstract type. We can't access the bounds of the symbol yet because that
-          // would cause a cause a cyclic reference. So we return `Nil` instead
-          // and try to make up for it later. The acrobatics in Scala2Unpicker#readType
-          // for reading a TypeRef show what's neeed.
-          Nil
-        else tsym.info.typeParams
+        else tsym.infoOrCompleter match {
+          case completer: TypeParamsCompleter =>
+            val tparams = completer.completerTypeParams(tsym)
+            if (tsym.isClass) tparams
```
smarter: I think this will always be false since we check `if (tsym.isClass)` above and go in another branch.

smarter:
> Btw "Beta reduce #Apply applications over classes" is still needed. IndexedSeq won't compile standalone without it.

Indeed, there's a real issue here, I think the problem is that the parent `IndexedSeqFactory[IndexedSeq]` of  `object IndexedSeq` is desugared into:
 ```scala
IndexedSeqFactory {
  type scala$collection$generic$IndexedSeqFactory$$CC = scala.collection.immutable.IndexedSeq
}
```
Which is wrong because `CC` is higher-kinded, it should be:
```scala
IndexedSeqFactory {
  type scala$collection$generic$IndexedSeqFactory$$CC = [X0] -> scala.collection.immutable.IndexedSeq[X0]
}
```

Instead of fixing this a posteriori by doing a special kind beta-reduction, it seems cleaner to try to eta-expand everything correctly in the first place, I have a branch on top of this branch which reverts the beta-reduction patch and instead changes `typedAppliedTypeTree` to force tparams so that arguments can be eta-expanded if necessary, this branch passes all tests. Let me know if you think this makes sense: https://github.com/dotty-staging/dotty/commits/change-early-typeparams2-force

odersky:
Rebased to master

src/dotty/tools/dotc/core/TypeApplications.scala
odersky: That is fixed in a subsequent commit.

odersky:
@smarter I think the -force branch is much better as a solution if it indeed passes all the tests. Can you make it a PR?

odersky:
[info] total memory = 910mb

So clearly increasing with -J-X had no effect. @DarkDimius please advise how to fix this. Also: is junit, bootstrapped partest and partest all run on the same JVM? That would increase the memory pressure of course.

smarter:
> @smarter I think the -force branch is much better as a solution if it indeed passes all the tests. Can you make it a PR?

OK, will do

> So clearly increasing with -J-X had no effect. @DarkDimius please advise how to fix this.

I have an attempt at https://github.com/lampepfl/dotty/pull/1030

src/dotty/tools/dotc/core/TypeApplications.scala
smarter: In which subsequent commit? I'm commenting on the whole diff here and I still see:
```scala
if (tsym.isClass) ...
else tsym.infoOrCompleter match {
  case ... =>
  val tparams = ...
  if (tsym.isClass) ...
```

odersky:
Superseded by #1031

PR #1029, opened 2016-01-15
partest: Print stack traces of swallowed exceptions

smarter:


smarter:
Review by @DarkDimius or @odersky. The tests failed which is perfect since that means you can check that my patch actually works :)
Note: It seems that all the failures we saw recently where due to OutOfMemoryException, the first time I saw an OutOfMemoryException was with https://github.com/lampepfl/dotty/pull/1010 so either this PR or a slightly older one must have increased the amount of memory we use.

odersky:
I fear there's something else. My latest pull requests have the GC overhead at 2G, but they still fail. What I see with recent commits is

 - either some memory size problem (GC overhead limit exceeded, out of Java heap space)
 - or a silent compilation failed.

The first happens for various files, the second always happens for pos/printing. I am not sure the second is also a heap problem because it persists no matter what heap size I set.

smarter:
@odersky : this PR should at least avoid silent compilation errors, you could try rebasing on top of it

PR #1030, opened 2016-01-16
Increase the maximum heap size on Jenkins

smarter:
Review by @odersky or @DarkDimius

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -6,7 +6,9 @@ import scala.reflect.io.Path
 
 object DottyBuild extends Build {
 
-  val travisMemLimit = List("-Xmx1g", "-Xss2m")
+  // Currently, this cannot be increased without hitting the maximum amount of memory
+  // available on the Jenkins VMs
+  val travisMemLimit = List("-Xmx1100m")
 
   val JENKINS_BUILD = "dotty.jenkins.build"
 


smarter:
```
[info] OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000006ded80000, 2463629312, 0) failed; error='Cannot allocate memory' (errno=12)
[info] #
[info] # There is insufficient memory for the Java Runtime Environment to continue.
[info] # Native memory allocation (mmap) failed to map 2463629312 bytes for committing reserved memory.
[info] # An error report file with more information is saved as:
[info] # /home/jenkins/workspace/dotty-master-validate-partest/hs_err_pid24921.log
```
I think this means that we're trying to allocate more memory than available on the VM.

@SethTisue, @adriaanm : How much memory do the VMs have? Is this something that could be increased?

In the meantime, I'll try reducing `-Xmx` a bit and see if I can find a middle ground.

DarkDimius:
@smarter 

Quoting @adriaanm 
> behemoths are c4.2xlarge, the others c4.xlarge

c4.2xlarge = 15 Gb Ram,
c4.xlarge = 7.5 Gb ram

smarter:
/rebuild -- let's see if this is stable now

smarter:
@DarkDimius : can we reduce the parallelism so that we use less memory?

smarter:
OK, I've done four runs with the maximum heap size increased from 1 GB to 1.1 GB for Jenkins and they all succeeded, so I propose we merge this now and worry about a longer-term solution later. I didn't change the memory settings used when running the tests locally because decreasing the heap size there would probably slow down the tests.

odersky:
Can we give it 1.5G? That would give us some margin for more complicated
tests.

 - Martin

On Sat, Jan 16, 2016 at 9:15 PM, Guillaume Martres <notifications@github.com
> wrote:

> OK, I've done four runs with the maximum heap size increased from 1 GB to
> 1.1 GB for Jenkins and they all succeeded, so I propose we merge this now
> and worry about a longer-term solution later. I didn't change the memory
> settings used when running the tests locally because decreasing the heap
> size there would probably slow down the tests.
>
> —
> <https://github.com/lampepfl/dotty/pull/1030#issuecomment-172252345>.
>



-- 
Martin Odersky
EPFL

smarter:
> Can we give it 1.5G? That would give us some margin for more complicated tests.

Unfortunately not, I tried that at first but got the failure described in https://github.com/lampepfl/dotty/pull/1030#issuecomment-172213236 because apparently we're using up all the RAM available on the VM instance, 1.1G is the max I could get away with, even 1.25G is too much. So this is really a short-term fix.

smarter:
Note that if the VM instances had a swap partition then at least we wouldn't get a crash when using too much RAM, it would just be very slow, but an even better fix would be to figure out how we're using the 15GB of RAM of the VMs (I have no idea how many JVMs we run in parallel and what controls this)

DarkDimius:
> I have no idea how many JVMs we run in parallel and what controls this

We run 1 JVM per core. 
Those Amazon EC2 VMs have 1.875 GB per core.

smarter:
> We run 1 JVM per core. 

You mean, `partest` starts 1 JVM per-core? But you also have to take into account the JVM used by sbt itself right?

smarter:
If N=Number of cores, could we try running partest with N-1 JVMs instead of N JVMs?

DarkDimius:
c4.2xlarge:
8 Cores,
15 GB Ram.

Even if we run 9 JVMs, that' 1.6GB per VM.

smarter:
You also have to take into account that some ram is used by the OS itself, and the total memory usage of a JVM is bigger than its heap size, but I don't know how much, so we might be close to the limit.

smarter:
(One thing that might be worth investigating: in the error I showed above the amount of memory that the JVM tried to "commit" was 2463629312 bytes = 2.35 GB, this is much bigger than the maximum heap size of 1.5G I specified, so something weird may be going on here, we might have more details once https://github.com/scala/scala-jenkins-infra/pull/157 is merged and we can take a look at the log file).

odersky:
Wait: Partest runs N threads in one VM. Otherwise we would not have seen
data races. Each thread runs a full compiler, and a full compiler needs a
couple 100 mb to compile a significant chunk of sources. So it seems to me
that if max heap if fixed we could still reduce memory pressure by reducing
N. What is N at the moment?


On Sun, Jan 17, 2016 at 12:11 AM, Guillaume Martres <

> (One thing that might be worth investigating: in the error I showed above
> the amount of memory that the JVM tried to "commit" was 2463629312 bytes =
> 2.35 GB, this is much bigger than the maximum heap size of 1.5G I
> specified, so something weird may be going on here, we might have more
> details once scala/scala-jenkins-infra#157
> <https://github.com/scala/scala-jenkins-infra/pull/157> is merged and we
> can take a look at the log file).
>
> —
> <https://github.com/lampepfl/dotty/pull/1030#issuecomment-172267809>.
>



-- 
Martin Odersky
EPFL

smarter:
> What is N at the moment?

It's the number of cores on the machine, here's an attempt at reducing this: https://github.com/lampepfl/dotty/pull/1032

SethTisue:
> @SethTisue, @adriaanm : How much memory do the VMs have? Is this something that could be increased?

Adriaan has some numbers at https://github.com/scala/scala-jenkins-infra/blob/master/doc/design.md, not entirely sure they're current. If you want to poke around yourself, we should probably get you access to the Jenkins nodes so you can ssh in and investigate on your own, try experiments from the command line, etc.  see https://github.com/scala/scala-jenkins-infra/blob/master/doc/client-setup.md#hosts-and-ssh-config

smarter:
@SethTisue : That could be useful in the future yes, I guess you'll need me to give you an ssh key, if so you can use: http://guillaume.martres.me/id_rsa.pub

SethTisue:
@adriaanm could you help Guillaume with this? I realized this is part of this process I don't know about, the docs just say "Send Adriaan your public key... He will use it to encrypt your credentials"

adriaanm:
Here's the section where you documented how to do this: https://github.com/scala/scala-jenkins-infra/blob/master/doc/client-setup.md#get-your-public-key-added :-)

SethTisue:
You can lead a horse to documentation, but you can't make him drink. Even if you are the horse.

SethTisue:
Hmm, I merged https://github.com/scala/scala-jenkins-infra/pull/159 and ran chef-client on jenkins-master, but I didn't see anything about the change in the resulting output, and I don't see Guillaume's key in `~ec2-user/.ssh/authorized_keys`. Is there something else we need to do to make the change kick in?

adriaanm:
`default['authorized_keys']['jenkins']` is for the `jenkins` user

SethTisue:
hmm. do I need to run chef-client on the workers?

adriaanm:
yes, but they should run that on a schedule

SethTisue:
ok. @smarter, if this isn't already working for you, let us know.

PR #1039, opened 2016-01-19
Fix #998

### LINKED ISSUE #998: Type inference too imprecise when approximating union of abstract types
In this example, `elem` is inferred to be of type `Any` instead of type `X` as in Scala 2:
```scala
object Test {
  def foo[A <: X, B <: X, X](left: A, right: B): Unit = {
    val elem = if (false) left else right
  }
}
```
The issue is in `TypeOps#approximateUnion`, we approximate a union of types using the intersection of the base classes of the types, but that's not correct when the types are bounded by abstract types.

odersky:
Needed a fix in approximateUnion. Review by @smarter


smarter:
LGTM, even though it feels like their should be a simpler way to do this.

PR #1040, opened 2016-01-20
Fix #1036

### LINKED ISSUE #1036: Issue with Type inference?
The following code snippet does not compile with Dotty but does with Scalac.
```scala
object Test {
  def second[A, B <: A](foo: A, bar: B): B = bar
  def expectString(s: String) = s

  def test = {
    second(Set.empty[String], Set.empty) map expectString
  }
}
```
It fails with the following error:
```scala
./tests/allan/Experiment.scala:7: error: type mismatch:
 found   : Any(s)
 required: String
    second(Set.empty[String], Set.empty) map expectString
                                             ^
```
It also fails to compile if `second` is defined as follow:
```scala
def second[A, B >: A](foo: A, bar: B): B = bar
```
with the following error:
```scala
./tests/allan/Experiment.scala:7: error: type mismatch:
 found   : (scala.collection.immutable.Set[String] | scala.collection.immutable.Set[Any]) @UnsafeNonvariant#A(s)
 required: String
    second(Set.empty[String], Set.empty) map expectString
                                             ^
```
Note that it does compile with `second` applied on `List` instead of `Set`:
```scala
second(List.empty[String], List.empty) map expectString
```
Not sure if it is a duplicate of #998

### LINKED ISSUE #1037: Stack overflow when compiling the same code in too many runs
Reproduce with:
```shell
sbt "run-main dotty.tools.dotc.Bench #runs 10 tests/pos/HelloWorld.scala"
```
Here's the full stack trace (only available if the JVM is started with `-XX:MaxJavaStackTraceDepth=-1` otherwise it's truncated): https://gist.github.com/anonymous/712e94d35fb69faf62fe

odersky:
Achieved by tweaking from where we get the parameter types of an eta-expansion. Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -1460,7 +1460,12 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
           }
       case wtp: MethodType if !pt.isInstanceOf[SingletonType] =>
         val arity =
-          if (defn.isFunctionType(pt)) defn.functionArity(pt)
+          if (defn.isFunctionType(pt))
+            if (!isFullyDefined(pt, ForceDegree.none) && isFullyDefined(wtp, ForceDegree.none))
+              // if method type is fully defined, but expected type is not,
+              // prioritize method parameter types as parameter types of the eta-expanded closure
+              0
+            else defn.functionArity(pt)
           else if (pt eq AnyFunctionProto) wtp.paramTypes.length
           else -1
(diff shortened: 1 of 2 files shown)


odersky:
Merged the three commits and changed commit message to point to correct issue.

odersky:
The failing test is a benchmark that tests that scala maps are not more than 10x slower than Java maps. This seems to have failed for this run. Is that our concern for dotty?

smarter:
LGTM

PR #1052, opened 2016-02-04
Improve and document the Driver#process API, fix partest logging

smarter:
Review by @DarkDimius or @odersky , this was very boring to do but very much needed.

Note that this means that when running the partest job, the compiler output will not be available just by scrolling through the giant log, for tests where compilation fails it will be at the bottom, for other tests (including tests which fail at runtime and not at compilation time), it will only be available in the *.clog files which can be retrieved by clicking on "Build Artifacts" on the Jenkins page for a job. If you're displeased with this, we could log it in the giant log too.

### DIFF
--- a/src/dotty/tools/dotc/Driver.scala
+++ b/src/dotty/tools/dotc/Driver.scala
@@ -40,19 +40,57 @@ abstract class Driver extends DotClass {
     (fileNames, ctx)
   }
 
+
+  /** Principal entry point to the compiler.
+   *  Creates a new compiler instance and run it with arguments `args`.
+   *
+   *  The optional arguments of this method all have `null` as their default
+   *  value, this makes it easier to call this method by reflection or from Java.
+   *
+   *  @param args       Arguments to pass to the compiler.
+   *  @param reporter   Used to log errors, warnings, and info messages.
+   *                    The default reporter is used if this is `null`.
+   *  @param callback   Used to execute custom code during the compilation
+   *                    process. No callbacks will be executed if this is `null`.
+   *  @return           The `Reporter` used. Use `Reporter#hasErrors` to check
+   *                    if compilation succeeded.
+   */
+  final def process(args: Array[String], reporter: Reporter = null,
+    callback: CompilerCallback = null): Reporter = {
+    val ctx = initCtx.fresh
+    if (reporter != null)
+      ctx.setReporter(reporter)
+    if (callback != null)
+      ctx.setCompilerCallback(callback)
+    process(args, ctx)
+  }
+
+  /** Entry point to the compiler with no optional arguments.
+   *
+   *  This overload is provided for compatibility reasons: the
+   *  `RawCompiler` of sbt expects this method to exist and calls
+   *  it using reflection. Keeping it means that we can change
+   *  the other overloads without worrying about breaking compatibility
+   *  with sbt.
+   */
+  final def process(args: Array[String]): Reporter =
+    process(args, null, null)
+
+  /** Entry point to the compiler using a custom `Context`.
+   *
+   *  In most cases, you do not need a custom `Context` and should
+   *  instead use one of the other overloads of `process`. However,
+   *  the other overloads cannot be overriden, instead you
+   *  should override this one which they call internally.
+   *
+   *  @param args       Arguments to pass to the compiler.
+   *  @param rootCtx    The root Context to use.
+   *  @return           The `Reporter` used. Use `Reporter#hasErrors` to check
+   *                    if compilation succeeded.
+   */
   def process(args: Array[String], rootCtx: Context): Reporter = {
     val (fileNames, ctx) = setup(args, rootCtx)
     doCompile(newCompiler(), fileNames)(ctx)
-  }
-
-  def process(args: Array[String], callback: CompilerCallback): Reporter = {
-    process(args, initCtx.setCompilerCallback(callback))
-  }
-
-  // We overload `process` instead of using a default argument so that we
-  // can easily call this method using reflection from `RawCompiler` in sbt.
-  def process(args: Array[String]): Reporter = {
-    process(args, initCtx)
   }
 
   def main(args: Array[String]): Unit = {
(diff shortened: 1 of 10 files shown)


src/dotty/tools/dotc/Driver.scala
```diff
@@ -33,28 +33,64 @@ abstract class Driver extends DotClass {
...
 
-  // We overload `process` instead of using a default argument so that we
-  // can easily call this method using reflection from `RawCompiler` in sbt.
-  def process(args: Array[String]): Reporter = {
-    process(args, initCtx)
+  /** Entry point to the compiler with no optional arguments.
+   *
+   *  This overload is provided for compatibility reasons: the
+   *  `RawCompiler` of sbt expect this method to exist and calls
+   *  it using reflection. Keeping it means that we can change
```
odersky: expects

src/dotty/tools/dotc/Driver.scala
```diff
@@ -33,28 +33,64 @@ abstract class Driver extends DotClass {
...
+   *  the other overloads without worrying about breaking compatibility
+   *  with sbt.
+   */
+  final def process(args: Array[String]): Reporter =
+    process(args, null, null)
+
+  /** Entry point to the compiler using a custom `Context`.
+   *
+   *  In most cases, you do not need a custom `Context` and should
+   *  instead use one of the other overload of `process`. However,
```
odersky: overloads

src/dotty/tools/dotc/Driver.scala
```diff
@@ -33,28 +33,64 @@ abstract class Driver extends DotClass {
...
+   *
+   *  @param args       Arguments to pass to the compiler.
+   *  @param reporter   Used to log errors, warnings, and info messages.
+   *                    The default reporter is used if this is `null`.
+   *  @param callback   Used to execute custom code during the compilation
+   *                    process. No callbacks will be executed if this is `null`.
+   *  @return           The `Reporter` used. Use `Reporter#hasErrors` to check
+   *                    if compilation succeeded.
+   */
+  final def process(args: Array[String], reporter: Reporter = null,
```
odersky: Do we need the default arguments here? Or do the overloaded variants of `process` cover all the use cases
already?

src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -426,7 +420,9 @@ object Contexts {
   abstract class FreshContext extends Context {
     def setPeriod(period: Period): this.type = { this.period = period; this }
     def setMode(mode: Mode): this.type = { this.mode = mode; this }
+    def setCompilerCallback(callback: CompilerCallback): this.type = { this.compilerCallback = callback; this }
     def setTyperState(typerState: TyperState): this.type = { this.typerState = typerState; this }
+    def setReporter(reporter: Reporter): this.type = setTyperState(typerState.withReporter(reporter))
```
odersky: That's a useful helper method. I think teher are several occurrences of `setTyperState(ctx.typerState.withReporter(...))` which would profit from that method. Best to do it now.

odersky:
Otherwise LGTM

src/dotty/tools/dotc/Driver.scala
smarter: I think so, yes, I don't want to add even more overloads like:
```scala
def process(args: Array[String], reporter: Reporter)
def process(args: Array[String], callback: CompilerCallback)
```

odersky: OK, fair enough.

On Thu, Feb 4, 2016 at 7:09 PM, Guillaume Martres <notifications@github.com>
wrote:

> In src/dotty/tools/dotc/Driver.scala
> <https://github.com/lampepfl/dotty/pull/1052#discussion_r51912246>:
>
> > -    process(args, initCtx.setCompilerCallback(callback))
> > +  /** Principal entry point to the compiler.
> > +   *  Creates a new compiler instance and run it with arguments `args`.
> > +   *
> > +   *  The optional arguments of this method all have `null` as their default
> > +   *  value, this makes it easier to call this method by reflection or from Java.
> > +   *
> > +   *  @param args       Arguments to pass to the compiler.
> > +   *  @param reporter   Used to log errors, warnings, and info messages.
> > +   *                    The default reporter is used if this is `null`.
> > +   *  @param callback   Used to execute custom code during the compilation
> > +   *                    process. No callbacks will be executed if this is `null`.
> > +   *  @return           The `Reporter` used. Use `Reporter#hasErrors` to check
> > +   *                    if compilation succeeded.
> > +   */
> > +  final def process(args: Array[String], reporter: Reporter = null,
>
> I think so, yes, I don't want to add even more overloads like:
>
> def process(args: Array[String], reporter: Reporter)def process(args: Array[String], callback: CompilerCallback)
>
> —
> <https://github.com/lampepfl/dotty/pull/1052/files#r51912246>.
>



-- 
Martin Odersky
EPFL

smarter:
@adriaanm, @SethTisue : any idea why Jenkins happily reports "success" when a job explodes with `OutOfMemoryException` ? https://scala-ci.typesafe.com/job/dotty-master-validate-junit/1000/console

smarter:
Nevermind, this seems to be sbt's fault, not Jenkins'.

smarter:
OK, tests now pass as expected, the missing ingredient was that `CompilerTest` reused the ctx field from `DottyTest` for every test, leading to weird errors, to prevent this `CompilerTest` no longer extends `DottyTest`. Ideally, someone should cleanup `DottyTest` and all the non-junit tests, see https://github.com/smarter/dotty/commit/73fad44a11dafa507e1d358f1bc222b04319a594 for details.

DarkDimius:
>  If you're displeased with this, we could log it in the giant log too.

I'd better be able to see the failure without needing to download and extract archive.

smarter:
> I would prefer to be able to see the failure without needing to download and extract archive.

If the compilation actually failed, the compilation logs will be displayed at the end of the partest run.

PR #1053, opened 2016-02-04
Fix #1045

### LINKED ISSUE #1045: Typing deviation in getOrElse
```
import scala.collection._
object T {
  val newSymbolMap: mutable.HashMap[String, mutable.HashMap[Int, Double]] = mutable.HashMap.empty
  val map = newSymbolMap.getOrElse("a", mutable.HashMap.empty)
  map.put(1, 0.0)
  newSymbolMap.put("a", map)
}
```
Compiles fine with scalac. Fails in dotty:
```
reproduce.scala:5: error: type mismatch:
 found   : Int(1)
 required: T.map.A
  map.put(1, 0.0)
          ^
reproduce.scala:5: error: type mismatch:
 found   : Double(0.0)
 required: T.map.B
  map.put(1, 0.0)
             ^
reproduce.scala:6: error: type mismatch:
 found   : (scala.collection.mutable.HashMap[Int, Double] |
  scala.collection.mutable.HashMap[Nothing, Nothing]
)(T.map)
 required: collection.mutable.HashMap[Int, Double]
  newSymbolMap.put("a", map)
                        ^
three warnings found
three errors found
```

This code is part of #630 and it used to compile fine.

odersky:
Thanks for putting me on the right track here @smarter. approximateUnion was indeed to blame.
In addition to distributing `|' inside refinements I also had to work on unifying type arguments.

Review by @smarter


src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -266,31 +266,72 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
...
+      }
+
+      tp1 match {
+        case tp1: RefinedType =>
+          tp2 match {
+            case tp2: RefinedType if tp1.refinedName == tp2.refinedName =>
+              return tp1.derivedRefinedType(
+                approximateUnion(OrType(tp1.parent, tp2.parent)),
+                tp1.refinedName,
+                homogenizedUnion(tp1.refinedInfo, tp2.refinedInfo).substRefinedThis(tp2, RefinedThis(tp1)))
```
smarter: Why not `homogenizedUnion(tp1.refinedInfo, tp2.refinedInfo.substRefinedThis(tp2, RefinedThis(tp1)))` ? This should require less traversals and is closer to what we were doing before https://github.com/lampepfl/dotty/commit/ee76fda79d446a2d6db51cb4af032a8e92936013

smarter:
Have you considered adding a parameter to `TypeComparer#distributeOr` instead to allow approximations? This would reduce the amount of traversals we would have to do, I'm not sure if the result would be more or less easy to understand than this.

odersky:
@smarter After having thought about it, I believe it's better to leave approximateUnion and distributeOr separate.

src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -266,31 +266,72 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
...
-          case nx => nx
-        }
-        tp.tp1 match {
-          case tp1: TypeProxy if !isClassRef(tp1) =>
-            approximateUnion(next(tp1) | tp.tp2)
-          case _ =>
-            tp.tp2 match {
-              case tp2: TypeProxy if !isClassRef(tp2) =>
-                approximateUnion(tp.tp1 | next(tp2))
+    def approximateOr(tp1: Type, tp2: Type)(implicit ctx: Context): Type = {
```
smarter: The implicit Context argument is unnecessary: TypeOps self-type is already a Context.

smarter:
Otherwise LGTM

src/dotty/tools/dotc/core/TypeOps.scala
odersky: Indeed. That was a leftover of a previous version where something got added to the context.

PR #1054, opened 2016-02-04
Changes to wildApprox

### LINKED ISSUE #1044: Dotty deviation in either implicit search or java wildcards
```
???.getClass.getMethods.head.getParameterTypes.mkString(",")
```

compiles fine with scalac. Does not in Dotty:

```
error: type mismatch:
 found   : Array[Class[_]]
 required: ?{ mkString: ? }
 both method wrapRefArray in class LowPriorityImplicits and method genericArrayOps in object Predef$ convert from Array[Class[_]] to ?{ mkString: FunProto(","):? }
    val baz_int_param = ???.getClass.getMethods.head.getParameterTypes.mkString(",")
```

This is part of test cases of #630, so it used to compile fine back then.

odersky:
See comment in new overloaded variant of wildApprox for an explanation.
Fixes #1044.

Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -379,6 +379,19 @@ object ProtoTypes {
     }
   }
 
+  /** Like wildApprox, but assume that all parameters of `disregarded` are not
+   *  bound in current constraint. This is necessary because an instance of
+   *  the type of a polymorphic function might already be part of the current
+   *  constraint when we try to check another occurrence of the same polymorphic
+   *  function for eligibility. An example is in i1044.scala where one expression
+   *  requires two implicit calls of the same refArrayOps method. When approximating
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -379,6 +379,19 @@ object ProtoTypes {
     }
   }
 
+  /** Like wildApprox, but assume that all parameters of `disregarded` are not
```
smarter: It's a bit weird to say "Like wildApprox" when this method is called wildApprox too, maybe use another name, or say, "Like the other overload of wildApprox"?

smarter:
I have an alternative fix which I think is a cleaner solution: https://github.com/lampepfl/dotty/pull/1056

smarter:
Superceded by #1064

PR #1055, opened 2016-02-04
Rewrite a.type & B to Nothing

### LINKED ISSUE #864: Compiler hangs on implicit search with union of intersection of singleton types
This is a more general case than https://github.com/lampepfl/dotty/issues/830 which was fixed in https://github.com/lampepfl/dotty/pull/843
```scala
object C {
  val a: Int = 1
  val b: Int = 2
  val c: Int = 2

  trait X[T]
}
```

odersky:
provided !(a.type <: B). Fixes #864, which turned out
to be a long sequence of comparisons between intersections
of singleton types.

Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -1047,7 +1047,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     else hkCombine(tp1, tp2, tparams1, tparams2, op)
   }
 
-  /** Try to distribute `&` inside type, detect and handle conflicts */
+  /** Try to distribute `&` inside type, detect and handle conflicts
+   *  pre: !(tp1 <: tp2) && !(tp2 <:< tp2) -- these cases were handled before
+   */
   private def distributeAnd(tp1: Type, tp2: Type): Type = tp1 match {
     // opportunistically merge same-named refinements
     // this does not change anything semantically (i.e. merging or not merging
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -1047,7 +1047,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     else hkCombine(tp1, tp2, tparams1, tparams2, op)
   }
 
-  /** Try to distribute `&` inside type, detect and handle conflicts */
+  /** Try to distribute `&` inside type, detect and handle conflicts
+   *  pre: !(tp1 <: tp2) && !(tp2 <:< tp2) -- these cases were handled before
```
smarter: `!(tp2 <:< tp2)` did you mean `!(tp2 <:< tp1)` ?

smarter: (Also we usually write `@pre`)

DarkDimius:
Should we also add such rule: 
If tp1 is class and tp2 is class and `!((tp1 <:< tp2) || (tp2 <:< tp1))` then `tp1& tp2 == Nothing`

odersky:
@darkdimius That would only work for unparameterized classes, and right now their glb would be Null. But yes, we could try that.

smarter:
This seems to be equivalent to what you tried before in https://github.com/lampepfl/dotty/pull/843 and it fails in the same way, afterwards you changed your patch to only apply to `ConstantType` and added a comment that says:
```
// Make use of the fact that the intersection of two constant types
// types which are not subtypes of each other is known to be empty.
// Note: The same does not apply to singleton types in general.
// E.g. we could have a pattern match against `x.type & y.type`
// which might succeed if `x` and `y` happen to be the same ref
// at run time. It would not work to replace that with `Nothing`.
// However, maybe we can still apply the replacement to
// types which are not explicitly written.
```

odersky:
@smarter Thanks for reminding me. Yes this is the same as #843 and it does not work. 
I have a new PR that fixes the problem in a different way: #1058.

PR #1056, opened 2016-02-05
Alternative fix for #1044: Properly constrain polymorphic implicit views

smarter:
Review by @odersky, this is an alternative to #1054
<hr>
When `wildApprox` encounters a PolyParam it approximates it by its
bounds in the current constraint set, this only makes sense if we have
added the PolyType corresponding to this PolyParam to the constrain set
using `ProtoTypes#constrained`, otherwise, we might get the bounds of
another use of this `PolyType`. This is exactly what happened in
i1044.scala: to compile, the implicit view `refArrayOps` needs to be
used twice with two different type parameters.

The fix is to always constrain polymorphic implicit views before
calling `wildApprox` on their result type.

This fix was inspired by the approach of #1054.


smarter:
I now think that we don't need to call `constrained` on `PolyTypes` in `discardForView` because `TypevarsMissContext` is set anyway, closed in favor of #1064.

PR #1057, opened 2016-02-05
Hide stack traces behind -Ydebug

smarter:
They're not very useful for end users and some tests like
tests/neg/selfreq.scala always print these exceptions which makes it
harder to read the test logs,

Also use Thread.dumpStack() instead of creating an Exception and calling
printStackTrace() on it.

Review by @odersky  or @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -151,6 +151,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val Yheartbeat = BooleanSetting("-Yheartbeat", "show heartbeat stack trace of compiler operations.")
   val Yprintpos = BooleanSetting("-Yprintpos", "show tree positions.")
   val YnoDeepSubtypes = BooleanSetting("-Yno-deep-subtypes", "throw an exception on deep subtyping call stacks.")
+  val YplainPrinter = BooleanSetting("-Yplain-printer", "Pretty-print using a plain printer.")
   val YprintSyms = BooleanSetting("-Yprint-syms", "when printing trees print info in symbols instead of corresponding info in trees.")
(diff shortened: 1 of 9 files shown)


smarter:
> Stack traces are useful documentation when something goes wrong. Users should paste them into their bugreports. Turning them off by default makes it harder to troubeshoot. 

Look what happens when a user tries to compile `tests/neg/selfreq.scala`:
https://gist.github.com/anonymous/0f49ed5beb84d51f097a
200 lines of mostly useless output, the user will probably think that there is a bug in the compiler, even though this is just debug information.

> I'm against disabling stacktraces in partest. They provide valuable information in case of failure.

Note that I'm not disabling stacktraces when the compiler fails with an exception, I'm only disabling stack traces that are explicitly printed in random places in the compiler with `printStackTrace()` or `dumpStack()`. In master currently we have 15 cases where this happen, 4 of these cases are hidden behind `if (ctx.debug)`, why these 4? Don't you think that we should be consistent?

odersky:
OK, I understand the reasoning that we want to hide these stacktraces by default because they do not signify real compiler crashes. 

Can you try to do the whole test suite under -Ydebug? If that passes, we are good. If not, we need to either debug -Ydebug first, or else create a more specific option (-Ydebug:stacktraces?), or else
use a printer in Printers.scala for it. In fact it strikes me that we could generalize the `cyclicErrors` printer to all stack traces.

smarter:
> Can you try to do the whole test suite under -Ydebug? If that passes, we are good

This now works after https://github.com/dotty-staging/dotty/commit/a6719ec4b68c8919c34537045cd98e7a2222deea and https://github.com/lampepfl/dotty/commit/ee67cd9581babc337604c3555c1e9733a7d6f9c4

DarkDimius:
@smarter could you maintain https://github.com/dotty-staging/dotty/commit/a6719ec4b68c8919c34537045cd98e7a2222deea under a `-Ytrace-context-creation`? It proved very useful while debugging typer&tree transforms.

smarter:
@DarkDimius : done!

PR #1058, opened 2016-02-05
Prune constraints that could turn into cycles

### LINKED ISSUE #864: Compiler hangs on implicit search with union of intersection of singleton types

odersky:
Fixes #864. Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/config/Config.scala
+++ b/src/dotty/tools/dotc/config/Config.scala
@@ -15,11 +15,12 @@ object Config {
    */
   final val checkConstraintsNonCyclic = false
 
-  /** Like `checkConstraintsNonCyclic`, but all constrained parameters
-   *  are tested for direct or indirect dependencies, each time a
-   *  constraint is added in TypeComparer.
+  /** Make sure none of the bounds in an OrderingConstraint contains
+   *  another constrained parameter at its toplevel (i.e. as an operand
+   *  of a combination of &'s and |'s.). The check is performed each time
+   *  a new bound is added to the constraint.
    */
-  final val checkConstraintsNonCyclicTrans = false
+  final val checkConstraintsSeparated = false
 
   /** Check that each constraint resulting from a subtype test
    *  is satisfiable.
--- a/src/dotty/tools/dotc/config/Config.scala
+++ b/src/dotty/tools/dotc/config/Config.scala
@@ -15,10 +15,10 @@ object Config {
    */
   final val checkConstraintsNonCyclic = false
 
-  /** Make sure none of the bounds in an OrderingConstraint contains
-   *  another constrained parameter at its toplevel (i.e. as an operand
-   *  of a combination of &'s and |'s.). The check is performed each time
-   *  a new bound is added to the constraint.
+  /** Make sure none of the bounds of a parameter in an OrderingConstraint
+   *  contains this parameter at its toplevel (i.e. as an operand of a
+   *  combination of &'s and |'s.). The check is performed each time a new bound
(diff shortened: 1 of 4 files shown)


odersky:
The issue uncovered a much deeper problem with constraint solving, which we solve by deliberatley widening the constraint.

smarter:
Interesting approach, do you think that you could create a `neg` test that demonstrate valid code that does not compile with this patch?
I see that you added an optional cycle detection check in `addOneBound`, couldn't we leave `addConstraint` as it was before, and instead always check for cycles in `addOneBound` and heal them? Would that be too costly?

odersky:
@smarter Logically, `prune` belongs where it is, because that's where we split a bound into a bound without type params and additional parameter/parameter relationships established by `addLess`. The cycle check in `addConstraint` can still fail, because TypeVars might be instantiated late.

odersky:
@smarter Also note that the cycle check does not catch cycles of length > 1.

src/dotty/tools/dotc/core/ConstraintHandling.scala
```diff
@@ -222,11 +235,67 @@ trait ConstraintHandling {
...
+        case bound: AndOrType =>
+          val p1 = prune(bound.tp1)
+          val p2 = prune(bound.tp2)
+          if (p1.exists && p2.exists) bound.derivedAndOrType(p1, p2)
+          else NoType
+        case bound: TypeVar if constraint contains bound.origin =>
+          prune(bound.underlying)
+        case bound: PolyParam if constraint contains bound =>
+          if (!addParamBound(bound)) NoType
+          else if (fromBelow) defn.NothingType
```
smarter: What about narrowing a bit less by using something like `prune(constraint.nonParamBounds(bound).hi)` instead of `defn.NothingType`?

odersky: That would not add anything. By recording the isLess we get the bounds anyway.

smarter:
Should we add a debug log message when we narrow bounds? Otherwise it might be easy to overlook when trying to understand why some code does not compile.

odersky:
I don't think we need a log message. The cases where this will fail will be really rare, and should be easily caught already by turning on the constr printer. Without constr, I think t will be impossible to diagnose.

src/dotty/tools/dotc/config/Config.scala
```diff
@@ -15,11 +15,12 @@ object Config {
    */
   final val checkConstraintsNonCyclic = false
 
-  /** Like `checkConstraintsNonCyclic`, but all constrained parameters
-   *  are tested for direct or indirect dependencies, each time a
-   *  constraint is added in TypeComparer.
+  /** Make sure none of the bounds in an OrderingConstraint contains
+   *  another constrained parameter at its toplevel (i.e. as an operand
```
smarter: You say "another constrained parameter" but the check does  `b eq param`, so shouldn't you say something like "Make sure none of the bounds of a parameter in an OrderingConstraint contains this parameter at its toplevel" ?

odersky: yes, indeed. Can you make that change please? Thanks!

smarter: Done, also fixed the documentation of `addConstraint`.

PR #1060, opened 2016-02-05
Force the initialization of throwMethod in Definitions

VladimirNik:
This change will fix the problem with unpickling of throw expression from TASTY representation.

### DIFF
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -786,7 +786,7 @@ class Definitions {
     OpsPackageClass)
 
     /** Lists core methods that don't have underlying bytecode, but are synthesized on-the-fly in every reflection universe */
-    lazy val syntheticCoreMethods = AnyMethods ++ ObjectMethods ++ List(String_+)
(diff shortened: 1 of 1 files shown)


DarkDimius:
LGTM after tests pass

VladimirNik:
Review by @odersky , @DarkDimius

src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -785,8 +785,8 @@ class Definitions {
     EmptyPackageVal,
     OpsPackageClass)
 
-    /** Lists core methods that don't have underlying bytecode, but are synthesized on-the-fly in every reflection universe */
-    lazy val syntheticCoreMethods = AnyMethods ++ ObjectMethods ++ List(String_+)
+  /** Lists core methods that don't have underlying bytecode, but are synthesized on-the-fly in every reflection universe */
+  lazy val syntheticCoreMethods = AnyMethods ++ ObjectMethods ++ List(String_+, throwMethod)
```
DarkDimius: indentation here was intentional: member definitions are indented with 2 more spaces compared to class definitions.

PR #1061, opened 2016-02-05
Fix #854: Optimize matches on primitive constants as switches.

### LINKED ISSUE #854: Implement @switch
Current implementation of pattern matcher does not optimize switch blocks on primitive constants.

sjrd:
This does not yet unable the checks that `@switch` verifies that the compiler was indeed able to perform the optimization.

This implementation also does not support guards. A match with guards will never be optimized as a switch.

### DIFF
--- a/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -303,8 +303,139 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
     def optimizeCases(prevBinder: Symbol, cases: List[List[TreeMaker]], pt: Type): (List[List[TreeMaker]], List[Tree])
     def analyzeCases(prevBinder: Symbol, cases: List[List[TreeMaker]], pt: Type, suppression: Suppression): Unit = {}
 
-    def emitSwitch(scrut: Tree, scrutSym: Symbol, cases: List[List[TreeMaker]], pt: Type, matchFailGenOverride: Option[Symbol => Tree], unchecked: Boolean): Option[Tree] =
-      None // todo
+    def emitSwitch(scrut: Tree, scrutSym: Symbol, cases: List[List[TreeMaker]], pt: Type, matchFailGenOverride: Option[Symbol => Tree], unchecked: Boolean): Option[Tree] = {
+      // TODO Deal with guards?
+
+      def isSwitchableType(tpe: Type): Boolean = {
+        (tpe isRef defn.IntClass) ||
+        (tpe isRef defn.ByteClass) ||
+        (tpe isRef defn.ShortClass) ||
+        (tpe isRef defn.CharClass)
+      }
+
+      object IntEqualityTestTreeMaker {
+        def unapply(treeMaker: EqualityTestTreeMaker): Option[Int] = treeMaker match {
+          case EqualityTestTreeMaker(`scrutSym`, _, Literal(const), _) =>
+            if (const.isIntRange) Some(const.intValue)
+            else None
+          case _ =>
+            None
+        }
+      }
+
+      def isSwitchCase(treeMakers: List[TreeMaker]): Boolean = treeMakers match {
+        // case 5 =>
+        case List(IntEqualityTestTreeMaker(_), _: BodyTreeMaker) =>
+          true
+
+        // case 5 | 6 =>
+        case List(AlternativesTreeMaker(`scrutSym`, alts, _), _: BodyTreeMaker) =>
+          alts.forall {
+            case List(IntEqualityTestTreeMaker(_)) => true
+            case _ => false
+          }
+
+        // case _ =>
+        case List(_: BodyTreeMaker) =>
+          true
+
+        /* case x @ pat =>
+         * This includes:
+         *   case x =>
+         *   case x @ 5 =>
+         *   case x @ (5 | 6) =>
+         */
+        case (_: SubstOnlyTreeMaker) :: rest =>
+          isSwitchCase(rest)
+
+        case _ =>
+          false
+      }
+
+      /* (Nil, body) means that `body` is the default case
+       * It's a bit hacky but it simplifies manipulations.
+       */
+      def extractSwitchCase(treeMakers: List[TreeMaker]): (List[Int], BodyTreeMaker) = treeMakers match {
+        // case 5 =>
+        case List(IntEqualityTestTreeMaker(intValue), body: BodyTreeMaker) =>
+          (List(intValue), body)
+
+        // case 5 | 6 =>
+        case List(AlternativesTreeMaker(_, alts, _), body: BodyTreeMaker) =>
+          val intValues = alts.map {
+            case List(IntEqualityTestTreeMaker(intValue)) => intValue
+          }
+          (intValues, body)
+
+        // case _ =>
+        case List(body: BodyTreeMaker) =>
+          (Nil, body)
+
+        // case x @ pat =>
+        case (_: SubstOnlyTreeMaker) :: rest =>
+          /* Rebindings have been propagated, so the eventual body in `rest`
+           * contains all the necessary information. The substitution can be
+           * dropped at this point.
+           */
+          extractSwitchCase(rest)
+      }
+
+      def doOverlap(a: List[Int], b: List[Int]): Boolean =
+        a.exists(b.contains _)
+
+      def makeSwitch(valuesToCases: List[(List[Int], BodyTreeMaker)]): Tree = {
+        def genBody(body: BodyTreeMaker): Tree = {
+          val valDefs = body.rebindings.emitValDefs
+          if (valDefs.isEmpty) body.body
+          else Block(valDefs, body.body)
+        }
+
+        val intScrut =
+          if (pt isRef defn.IntClass) ref(scrutSym)
+          else Select(ref(scrutSym), nme.toInt)
+
+        val (normalCases, defaultCaseAndRest) = valuesToCases.span(_._1.nonEmpty)
+
+        val newCases = for {
+          (values, body) <- normalCases
+        } yield {
+          val literals = values.map(v => Literal(Constant(v)))
+          val pat =
+            if (literals.size == 1) literals.head
+            else Alternative(literals)
+          CaseDef(pat, EmptyTree, genBody(body))
+        }
+
+        val catchAllDef = {
+          if (defaultCaseAndRest.isEmpty) {
+            matchFailGenOverride.fold[Tree](
+                Throw(New(defn.MatchErrorType, List(ref(scrutSym)))))(
+                _(scrutSym))
+          } else {
+            /* After the default case, assuming the IR even allows anything,
+             * things are unreachable anyway and can be removed.
+             */
+            genBody(defaultCaseAndRest.head._2)
+          }
+        }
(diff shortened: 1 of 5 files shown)


sjrd:
This is Work-in-Progress, but close enough to reality for a request for comments.

There is one failing test on my machine, t3833.scala, which looks like this:
```scala
package hello

object world {
  def mkArray[T <: A](atype: Int) :T#AType = {
    (atype match {
      case 1 =>
        new Array[Int](10)
        // Decompiled code: return (Object[])new int[10];
      case 2 =>
        new Array[Float](10)
    }).asInstanceOf[T#AType]
  }

    println(mkArray[I](1))
    //java.lang.ClassCastException: [I cannot be cast to [Ljava.lang.Object;
  }
}

trait A {
  type AType <: AnyRef
}
trait I extends A {
  type AType = Array[Int]
}
trait F extends A {
  type AType = Array[Float]
}
```
The tree emitted after `patternMatcher` looks good:
```scala
[error]     def mkArray[T <: hello.A](atype: Int): T#AType = {
[error]       (atype match {
[error]         case 1 => 
[error]           dotty.runtime.Arrays.newIntArray(10)
[error]         case 2 => 
[error]           dotty.runtime.Arrays.newFloatArray(10)
[error]       }).asInstanceOf[T#AType]
[error]     }
```
however, it causes a crash in GenBCode:
```
[error] java.lang.AssertionError: assertion failed: Expected primitive types I - Ljava/lang/Object;
[error]         at scala.Predef$.assert(Predef.scala:165)
[error]         at scala.tools.nsc.backend.jvm.BTypes$BType$class.conformsTo(BTypes.scala:157)
[error]         at scala.tools.nsc.backend.jvm.BTypes$ArrayBType.conformsTo(BTypes.scala:847)
[error]         at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.adapt(BCodeBodyBuilder.scala:896)
[error]         at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:436)
[error]         at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder$$anonfun$genMatch$3.apply(BCodeBodyBuilder.scala:844)
[error]         at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder$$anonfun$gen
...
```
I fear there might be a problem in Erasure; maybe it does not recognize `Match` nodes, and the necessity to "push" the expected adapted type `Object` inside the branches? I haven't investigate further yet.

Does anyone have an idea, off the top of your head?

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -303,8 +303,134 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
+          (values, body) <- normalCases
+        } yield {
+          val literals = values.map(v => Literal(Constant(v)))
+          val pat =
+            if (literals.size == 1) literals.head
+            else Alternative(literals)
+          CaseDef(pat, EmptyTree, genBody(body))
+        }
+
+        val wildcard = ctx.typeAssigner.assignType(untpd.Ident(nme.WILDCARD), defn.IntType)
```
sjrd: This is most probably a code smell. I haven't found a better way to generate a typed `Ident(WILDCARD)`, which is apparently what the back-end wants. Should there be a better way?

DarkDimius: `tpd.Underscore(IntType)`

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -303,8 +303,134 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
+      }
+
+      def doOverlap(a: List[Int], b: List[Int]): Boolean =
+        a.exists(b.contains _)
+
+      def makeSwitch(valuesToCases: List[(List[Int], BodyTreeMaker)]): Tree = {
+        def genBody(body: BodyTreeMaker): Tree = {
+          val valDefs = body.rebindings.emitValDefs
+          if (valDefs.isEmpty) body.body
+          else Block(valDefs, body.body)
```
sjrd: Is this the right way to avoid useless `Block`s over single expressions? I did not find any shortcut. Or do you not bother, in general, and just always create a `Block` in similar cases?

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -303,8 +303,134 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
+
+        // case _ =>
+        case List(_: BodyTreeMaker) =>
+          true
+
+        // case x =>
+        case List(_: SubstOnlyTreeMaker, _: BodyTreeMaker) =>
+          true
+
+        case _ =>
```
DarkDimius:
```
(1: Int) match {case x @ 1 => x}
```

src/dotty/tools/dotc/transform/PatternMatcher.scala
sjrd: Oh indeed! Thanks.

src/dotty/tools/dotc/transform/PatternMatcher.scala
DarkDimius: At the moment we do not bother in most cases.

src/dotty/tools/dotc/transform/PatternMatcher.scala
sjrd: It is not, indeed. I have actually deliberately omitted it, at the moment. It looked useless, as even as a human it would be obvious that `x` should be 1. I would have implemented it anyway, but for the fact the sequence of `treeMakers` was `SubstOnly`, `EqualityTest`, `Body`; and it seemed "dangerous" to move the `SubstOnly` after the `EqualityTest`.

I can reconsider this.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -303,8 +303,134 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
-    def emitSwitch(scrut: Tree, scrutSym: Symbol, cases: List[List[TreeMaker]], pt: Type, matchFailGenOverride: Option[Symbol => Tree], unchecked: Boolean): Option[Tree] =
-      None // todo
+    def emitSwitch(scrut: Tree, scrutSym: Symbol, cases: List[List[TreeMaker]], pt: Type, matchFailGenOverride: Option[Symbol => Tree], unchecked: Boolean): Option[Tree] = {
+      // TODO Deal with guards?
+
+      def isSwitchableType(tpe: Type): Boolean = {
+        (tpe isRef defn.IntClass) ||
+        (tpe isRef defn.ByteClass) ||
+        (tpe isRef defn.ShortClass) ||
+        (tpe isRef defn.CharClass)
```
DarkDimius:
```scala
|| (ValueClasses.isDerivedValueClass(tpe.denot) && 
     isSwitchableType(ValueClasses.underlyingOfValueClass(tpe.denot)))

```

sjrd: Although I agree it would be nice, this won't happen in practice, because you cannot express a *constant* value of a user-defined value class in the `case`s.

DarkDimius: After a bit of discussion: let's leave this possibility for future as it's not clear what is a constant for a value class.

src/dotty/tools/dotc/transform/PatternMatcher.scala
sjrd: OK I'll remove the test. I's a bit ugly because most cases don't have any `ValDef`, but let's follow the conventions.

DarkDimius: Feel free to leave it as is.
What I meant is that currently different parts of dotty already create a lot of Blocks with empty stats and there's no assumption that `stats.nonEmpty`.

sjrd: Oh OK.

src/dotty/tools/dotc/transform/PatternMatcher.scala
sjrd: Done. It now also support `case x @ (5 | 6) =>`, which can actually be useful.

sjrd:
Further progress on this is blocked by #1065.

sjrd:
Rebased on top of master, with the fix for #1065.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -303,8 +303,139 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
+       */
+      def extractSwitchCase(treeMakers: List[TreeMaker]): (List[Int], BodyTreeMaker) = treeMakers match {
+        // case 5 =>
+        case List(IntEqualityTestTreeMaker(intValue), body: BodyTreeMaker) =>
+          (List(intValue), body)
+
+        // case 5 | 6 =>
+        case List(AlternativesTreeMaker(_, alts, _), body: BodyTreeMaker) =>
+          val intValues = alts.map {
+            case List(IntEqualityTestTreeMaker(intValue)) => intValue
```
sjrd: This line causes a bootstrapping test (I think) to fail with
```
[error] java.lang.AssertionError: assertion failed: error at /home/jenkins/workspace/dotty-master-validate-partest/tests/partest-generated/pos/transform/PatternMatcher.scala:340
[error] type mismatch:
[error]  found   : $this.EqualityTestTreeMaker
[error]  required: TreeMakers.this.EqualityTestTreeMaker
[error] tree = TypeApply(Select(Ident(p38),asInstanceOf),List(TypeTree[TypeRef(TermRef(NoPrefix,$this),EqualityTestTreeMaker)]))
[error] 	at scala.Predef$.assert(Predef.scala:165)
[error] 	at dotty.tools.dotc.transform.TreeChecker$Checker.adapt(TreeChecker.scala:333)
[error] 	at dotty.tools.dotc.typer.ProtoTypes$FunProto.typedArg(ProtoTypes.scala:205)
```
It seems to be a Ycheck error. Apparently it cannot relate the synthetic `$this` thing (a capture?) to the enclosing `TreeMakers.this`.

Not sure what to make of this, atm.

smarter: AFAIK only two transformations introduce `$this`: `ExtensionMethods` and `TailRec`, and `extractSwitchCases` looks tail-recursive. So my guess is that this method makes pattern-matching and tail-recursiveness interact in an interesting way we haven't tested before (probably because `IntEqualityTestTreeMaker` is a local class). It'd be great if you could try to reduce this to a simple test case.

sjrd: I'll to minimize it, yes.

sjrd:
Rebased on top of #1091 (therefore depends on #1091 to be merged first)

sjrd:
Looks like this works, now :)

DarkDimius:
LGTM

PR #1066, opened 2016-02-08
Fix #997

### LINKED ISSUE #997: Private types should not be exposed by non-private members
We should not allow a private type to appear in the signature of a non-private member. 
The private type is not accessible outside the scope of the member. In a "by-name" based
system for resolving types this can lead to crashes. 

Example: `collection.mutable.IndexedSeqView` crashes when the accessibility of the `This` type
is restricted from `protected[this]` to `private[this]`. The crash happens in mixin when we try to refer
to `This` computing the signature of `filter` in `AbstractTransformedX`, which is an inner subclass
of `IndexedSeqView`. It can't work because `This` is not visible as an inherited member of `AbstractTransformedX`. To prevent crashes like these we have to be stricter in the frontend.

odersky:
Ensure that private references do not leak. Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/typer/Checking.scala
+++ b/src/dotty/tools/dotc/typer/Checking.scala
@@ -342,7 +342,7 @@ object Checking {
             else errors1
           } else errors1
         case tp: ClassInfo =>
-          (apply(errors, tp.prefix) /: tp.typeRef.parentsWithArgs)(apply)
+          (apply(errors, tp.prefix) /: tp.parentsWithArgs)(apply)
         case _ =>
           foldOver(errors, tp)
       }
--- a/src/dotty/tools/dotc/typer/Checking.scala
+++ b/src/dotty/tools/dotc/typer/Checking.scala
@@ -312,6 +312,46 @@ object Checking {
     checkNoConflict(Private, Protected)
     checkNoConflict(Abstract, Override)
   }
+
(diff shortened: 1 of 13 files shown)


src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -296,6 +296,46 @@ object Checking {
...
+        if (sym.is(Private)) sym.owner
+        else if (sym.privateWithin.exists) sym.privateWithin
+        else if (sym.is(Package)) sym
+        else accessBoundary(sym.owner)
+      def apply(errors: Errors, tp: Type): Errors = tp match {
+        case tp: NamedType =>
+          val errors1 =
+            if (tp.symbol.is(Private) &&
+               !accessBoundary(sym).isContainedIn(tp.symbol.owner)) {
+              (d"non-private $sym refers to private ${tp.symbol}\n in its type signature ${sym.info}", tree.pos) :: errors
```
smarter: Why not just call `ctx.errorOrMigrationWarning` here?

odersky: Because we might end up not issuing the error if a parent is a TypeAlias.

odersky:
Ready to merge?

DarkDimius:
LGTM

odersky:
Rebased to master

Issue #684, opened 2015-06-24
tuple-zipped.scala: support for case-syntax of lambda arguments


DarkDimius:
```
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/tuple-zipped.scala:27: error: wrong number of parameters, expected: 2
[info]       val sum1 = (cc1, cc2).zipped map { case (x, y) => x + y } sum
[info]                                        ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/tuple-zipped.scala:27: error: type mismatch:
[info]  found   : Nothing(x)
[info]  required: ?{ + : ? }
[info] Note that implicit conversions cannot be applied because they are ambiguous;
[info]  both method Character2char in object Predef$ and method Byte2byte in object Predef$ convert from Nothing(x) to ?{ + : FunProto(<y:Nothing>):Nothing' }
[info]       val sum1 = (cc1, cc2).zipped map { case (x, y) => x + y } sum
[info]                                                         ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/tuple-zipped.scala:27: error: ambiguous implicits: both object FloatIsFractional in object Numeric$ and object ByteIsIntegral in object Numeric$ match type scala.math.Numeric[Nothing'] of parameter num of method sum in trait TraversableOnce
[info]       val sum1 = (cc1, cc2).zipped map { case (x, y) => x + y } sum
[info]                                                                    ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/tuple-zipped.scala:34: error: wrong number of parameters, expected: 3
[info]       val sum1 = (cc1, cc2, cc3).zipped map { case (x, y, z) => x + y + z } sum
[info]                                             ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/tuple-zipped.scala:34: error: type mismatch:
[info]  found   : Nothing(x)
[info]  required: ?{ + : ? }
[info] Note that implicit conversions cannot be applied because they are ambiguous;
[info]  both method Character2char in object Predef$ and method Byte2byte in object Predef$ convert from Nothing(x) to ?{ + : FunProto(<y:Nothing>):?{ + : FunProto(<z:Nothing>):Nothing' } }
[info]       val sum1 = (cc1, cc2, cc3).zipped map { case (x, y, z) => x + y + z } sum
[info]                                                                 ^
[...]

Issue #681, opened 2015-06-24
run/vector1.scala fails to find implicit


DarkDimius:
```
[info] compiling /Users/dark/workspace/dotty/tests/partest-generated/run/vector1.scala
[info] options: -pagewidth 160 -Yno-deep-subtypes -Yno-double-bindings -Ycheck:tailrec,resolveSuper,mixin,restoreScopes -d ./tests/partest-generated/run/vector1-run.obj
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/vector1.scala:128: error: no implicit argument of type scala.collection.generic.CanBuildFrom[
[info]   (CC? >: ([HK$0] => scala.collection.mutable.WrappedArray[HK$0])[HK$0 = Int] <: ([HK$0] =>  <: scala.collection.TraversableOnce[HK$0]))[(T? = Int)]#Apply
[info] , (T? = Int),
[info]   (CC? >: ([HK$0] => scala.collection.mutable.WrappedArray[HK$0])[HK$0 = Int] <: ([HK$0] =>  <: scala.collection.TraversableOnce[HK$0]))[(T? = Int)]#Apply
[info] ] found for parameter bf of method shuffle in class Random
[info]     val pos = scala.util.Random.shuffle(scala.collection.mutable.WrappedArray.make[Int](Array.tabulate[Int](N)(i => i)))
[info]                                                                                                                         ^
```

smarter:
Probably duplicate of https://github.com/lampepfl/dotty/issues/553

smarter:
Fixed by #799, leaving this open until we move the test out of `pending`.

odersky:
@smarter I think it's preferable to move the test out of pending immediately if that can close an issue. If we do not do this, issues stay open and it takes unnecessary effort to come back to the context.

Issue #654, opened 2015-06-15
Convertions can't compare BigInt to numerics


DarkDimius:
```
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:213: error: type mismatch:
[info]  found   : Byte(-128)
[info]  required: scala.math.BigInt
[info]     val isByte = bi >= Byte.MinValue && bi <= Byte.MaxValue
[info]                             ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:213: error: type mismatch:
[info]  found   : Byte(127)
[info]  required: scala.math.BigInt
[info]     val isByte = bi >= Byte.MinValue && bi <= Byte.MaxValue
[info]                                                    ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:214: error: type mismatch:
[info]  found   : Short(-32768)
[info]  required: scala.math.BigInt
[info]     val isShort = bi >= Short.MinValue && bi <= Short.MaxValue
[info]                               ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:214: error: type mismatch:
[info]  found   : Short(32767)
[info]  required: scala.math.BigInt
[info]     val isShort = bi >= Short.MinValue && bi <= Short.MaxValue
[info]                                                       ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:215: error: type mismatch:
[info]  found   : Char('\00')
[info]  required: scala.math.BigInt
[info]     val isChar = bi >= Char.MinValue && bi <= Char.MaxValue
[info]                             ^
[info] /Users/dark/workspace/dotty/tests/partest-generated/run/is-valid-num.scala:215: error: type mismatch:
[info]  found   : Char('￿')
[info]  required: scala.math.BigInt
[info]     val isChar = bi >= Char.MinValue && bi <= Char.MaxValue
```

odersky:
This is a consequence of the fact that we dropped weak conformance in Dotty. BigInt has an implicit conversion from Int to BigInt, but none from Byte, Short, Char. And we cannot chain implicit conversions.

The proper solution should be to give BigInt the missing conversions.
I am adding them to DottyPredef for the moment.

smarter:
Fixed by #902, leaving open until we move this test out of pending.

PR #1071, opened 2016-02-08
Enable tests for old fixed issues

### LINKED ISSUE #530: Import of symbolic identifiers does not work
Example:

    object P {
      def !#@ : Nothing = ???
    }

    object Test {
      import P.!#@
      def f = !#@
    }

gives: 

    importSymbolic.scala:10: error: not found: !#@

      def f = !#@
              ^
    one error found

The problem is that import selectors are untyped trees, and therefore do not undergo decoding of symbolic characters. The problem will be fixed once we move decoding into a late phase.

smarter:
Fix #530, #654, #681, #684, #685 

Review by @odersky


odersky:
LGTM

PR #1073, opened 2016-02-10
Handle implicits with default parameters.

### LINKED ISSUE #576: default values for implicit parameters are ignored
This works fine in `scalac` but does not compile with dotty:
```scala
class A

object Impl {
  def foo()(implicit x: A = null): Int = 2
  def test: Int = {
    foo()
  }
}
```
```scala
[error] try/impldef.scala:6: error: no implicit argument of type A found for parameter x of method foo in object Impl$
[error]     foo()
[error]          ^
```

odersky:
Fixes #576.

run/pending/Meter now compiles; crashes at runtime.

Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -1435,9 +1435,14 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         wtp.paramTypes.foreach(instantiateSelected(_, tvarsToInstantiate))
         val constr = ctx.typerState.constraint
         def addImplicitArgs = {
-          def implicitArgError(msg: => String): Tree = {
-            ctx.error(msg, tree.pos.endPos)
+          val errors = new mutable.ListBuffer[() => String]
+          def implicitArgError(msg: => String) = {
+            errors += (() => msg)
             EmptyTree
+          }
+          def issueErrors() = {
+            for (err <- errors) ctx.error(err(), tree.pos.endPos)
+            tree
           }
           val args = (wtp.paramNames, wtp.paramTypes).zipped map { (pname, formal) =>
             def where = d"parameter $pname of $methodStr"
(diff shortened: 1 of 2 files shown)


smarter:
The first commit in this PR was a merge commit which is confusing, I've fixed and push-forced

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1435,29 +1435,39 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+          def argsOrError(pnames: List[TermName], ptypes: List[Type], acc: mutable.ListBuffer[Tree])
+            : AnyRef /* List[Tree] | () => String */ = (pnames, ptypes) match {
+            case (pname :: pnames1, ptype :: ptypes1) =>
+              def where = d"parameter $pname of $methodStr"
+              inferImplicit(ptype, EmptyTree, tree.pos.endPos) match {
+                case SearchSuccess(arg, _, _) =>
+                  acc += arg
+                  argsOrError(pnames1, ptypes1, acc)
+                case ambi: AmbiguousImplicits =>
+                  () => s"ambiguous implicits: ${ambi.explanation} of $where"
```
smarter: Why not simply return a String?

odersky: Because we do not want to incur the cost of constructing it when the error does not get reported. Error strings are always lazy.

smarter: OK, intuitively I'd think that the cost of creating a closure would be higher than the cost of constructing a String

smarter:
This doesn't work for implicit parameter lists where only some of the parameters have default values:
```scala
class A

object Impl {
  def foo()(implicit ev: Int, x: A = null): Int = 2
  def test: Int = {
    implicit val ii: Int = 1
    foo()
  }
}
```
stack trace: https://gist.github.com/smarter/40c990590e42be4b28f5

src/dotty/tools/dotc/typer/Typer.scala
odersky: Certainly not if the string is interpolated.

odersky:
Now also handles mixed lists where some implicits have defaults and others don't.

odersky:
I am going to merge because I have to touch this code again for the classtags issue and I don't want to be slowed down with merging the two fixes afterwards.

PR #1075, opened 2016-02-11
Drop empty companion objects

odersky:
Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
+++ b/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
@@ -26,10 +26,12 @@ class DropEmptyCompanions extends MiniPhase { thisTransform =>
 
     /** Is `tree` an empty companion object? */
     private def isEmptyCompanion(tree: Tree)(implicit ctx: Context) = tree match {
-      case TypeDef(_, impl: Template) =>
-        tree.symbol.is(Module) &&
+      case TypeDef(_, impl: Template) if
+        tree.symbol.is(SyntheticModule) &&
         tree.symbol.companionClass.exists &&
-        impl.body.forall(_.symbol.isPrimaryConstructor)
+        impl.body.forall(_.symbol.isPrimaryConstructor) =>
+        //println(i"removing ${tree.symbol}")
+        true
       case _ =>
         false
     }
--- a/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
+++ b/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
@@ -12,55 +12,55 @@ import ast.Trees._
 import collection.mutable
 import Decorators._
 import NameOps._
-import TreeTransforms.{TreeTransform, MiniPhase}
+import TreeTransforms.MiniPhaseTransform
 import dotty.tools.dotc.transform.TreeTransforms.TransformerInfo
 
-/** Remove companion objects that are empty */
-class DropEmptyCompanions extends MiniPhase { thisTransform =>
+/** Remove companion objects that are empty
+ *  Lots of constraints here:
+ *  1. It's impractical to place DropEmptyCompanions before lambda lift because dropped
+ *     modules can be anywhere and have hard to trace references.
+ *  2. DropEmptyCompanions cannot be interleaved with LambdaLift or Flatten because
+ *     they put things in liftedDefs sets which cause them to surface later. So
+ *     removed modules resurface.
+ *  3. DropEmptyCompanions has to be before RestoreScopes.
+ *  The solution to the constraints is to put DropEmptyCompanions between Flatten
+ *  and RestoreScopes and to only start working once we are back on PackageDef
+ *  level, so we know that all objects moved by LambdaLift and Flatten have arrived
+ *  at their destination.
+ */
+class DropEmptyCompanions extends MiniPhaseTransform { thisTransform =>
   import ast.tpd._
   override def phaseName = "dropEmpty"
-  val treeTransform = new Transform(Set())
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Flatten])
 
-  class Transform(dropped: Set[Symbol]) extends TreeTransform {
-    def phase = thisTransform
+  override def transformPackageDef(pdef: PackageDef)(implicit ctx: Context, info: TransformerInfo) = {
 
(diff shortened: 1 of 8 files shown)


src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -367,7 +367,7 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer {
             appendOffsetDefs += (companion.moduleClass -> new OffsetInfo(List(offsetTree), ord))
         }
 
-        val containerName = ctx.freshName(x.name ++ StdNames.nme.LAZY_LOCAL).toTermName
```
odersky: @DarkDimius This is just DRY in order not to do the same low-level name plumbing over and over again.

odersky:
@DarkDimius This was unexpectedly messy. It makes me wonder whether having synthesized companion objects is worth it. Could we not just use static members everywhere? We'd shed a ton
of complexity that way.

DarkDimius:
> Could we not just use static members everywhere?

I'm not sure I understood what you're proposing. Are you proposing to deprecate value classes and revert to previous implementation of lazy vals?

odersky:
No, I am proposing neither of that.

odersky:
I mean: Instead putting things in companion objects, make them static members of their class.

DarkDimius:
@odersky, it seems to me that instances of synthetic objects are never 
passed around.

I guess we can compile members of synthetic objects into static methods on 
class itself, by using https://github.com/scala/scala.github.com/pull/491. 
If all members of synthetic object are marked static we're free not to emit 
the $ class.

odersky:
Yes, but why do we put synthetic members into companion objects in the
first place? Creating synthetic companion objects for all classes is a
minor pain and getting rid of them again is a huge pain. Why do it?



On Sat, Feb 13, 2016 at 3:01 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> @odersky, it seems to me that instances of synthetic objects are never
> passed around.
>
> I guess we can compile members of synthetic objects into static methods on
> class itself, by using https://github.com/scala/scala.github.com/pull/491.
>
> If all members of synthetic object are marked static we're free not to
> emit
> the $ class.
>
>
> On 13 February 2016 11:02:35 odersky <notifications@github.com> wrote:
>
> > I mean: Instead putting things in companion objects, make them static
> > members of their class.
> >
> > ---
> > Reply to this email directly or view it on GitHub:
> > https://github.com/lampepfl/dotty/pull/1075#issuecomment-183634363
>
> —
> <https://github.com/lampepfl/dotty/pull/1075#issuecomment-183672265>.
>



-- 
Martin Odersky
EPFL

DarkDimius:
>Creating synthetic companion objects for all classes is a
>minor pain and getting rid of them again is a huge pain.

I believe it can be implemented with less pain using functionality from 
https://github.com/scala/scala.github.com/pull/491



>why do we put synthetic members into companion objects in the
first place?

Because it allows to ensure correct scoping using Ycheck, because those 
methods can be called without an instance of the class. etc: those methods 
do not behave as-if they were members of the class.


I'm convinced that's its fundamentally better to handle them entirely in 
backend instead of special-casing them in ycheck, tpd, type-assigner, 
lambdalift, flatten and increasing overall complexity. Even more given that 
we can reuse implementation needed for 
https://github.com/scala/scala.github.com/pull/491

I would have agreed if we had full control of code that we make static. 
This is true for lazy Vals, but false for value classes. Given this, I'd 
prefer to have Ycheck fully working, as bodies of methods of value classes 
that are lifted to static  can contain arbitrarily complex code.

I have a simpler proposal, though less principled than @static: we can 
allow phases to register predicates that apply to classes and say if 
companion module is needed. Such predicates will be registered by lazy vals 
and value classes phases.

odersky:
OK, I understand the arguments. Alternatives are not necessarily better. Let's get this in, it is quite workable as is.

DarkDimius:
Agreed. LGTM

src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
```diff
@@ -0,0 +1,98 @@
...
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Flatten])
+
+  override def transformPackageDef(pdef: PackageDef)(implicit ctx: Context, info: TransformerInfo) = {
+
+    /** Is `tree` an empty companion object? */
+    def isEmptyCompanion(tree: Tree) = tree match {
+      case TypeDef(_, impl: Template) if tree.symbol.is(SyntheticModule) &&
+        tree.symbol.companionClass.exists &&
+        impl.body.forall(_.symbol.isPrimaryConstructor) =>
+        println(i"removing ${tree.symbol}")
```
DarkDimius: println left in code. 
I'll make a pr to remove it.

odersky: It's already removed in another PR of mine. So it should go away by itself
soon.

On Thu, Feb 18, 2016 at 11:23 AM, Dmitry Petrashko <notifications@github.com
> wrote:

> In src/dotty/tools/dotc/transform/DropEmptyCompanions.scala
> <https://github.com/lampepfl/dotty/pull/1075#discussion_r53295488>:
>
> > + *  level, so we know that all objects moved by LambdaLift and Flatten have arrived
> > + *  at their destination.
> > + */
> > +class DropEmptyCompanions extends MiniPhaseTransform { thisTransform =>
> > +  import ast.tpd._
> > +  override def phaseName = "dropEmpty"
> > +  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Flatten])
> > +
> > +  override def transformPackageDef(pdef: PackageDef)(implicit ctx: Context, info: TransformerInfo) = {
> > +
> > +    /** Is `tree` an empty companion object? */
> > +    def isEmptyCompanion(tree: Tree) = tree match {
> > +      case TypeDef(_, impl: Template) if tree.symbol.is(SyntheticModule) &&
> > +        tree.symbol.companionClass.exists &&
> > +        impl.body.forall(_.symbol.isPrimaryConstructor) =>
> > +        println(i"removing ${tree.symbol}")
>
> println left in code.
> I'll make a pr to remove it.
>
> —
> <https://github.com/lampepfl/dotty/pull/1075/files#r53295488>.
>



-- 
Martin Odersky
EPFL

PR #1076, opened 2016-02-11
partest: run tests with -Xms64M -Xmx1024M instead of JVM defaults

smarter:
This is the same settings that scalac uses, this should help use avoid
using too much memory (the default on 64 bits Linux seems to be
-Xms248M -Xmx3938M).

Review by @DarkDimius


DarkDimius:
LGTM

Issue #1077, opened 2016-02-12
We should issue warning on class names that differ only in case


odersky:
Given this program

```
class A
class a

object Test {
  def main(args: Array[String]): Unit = {
    new a
    new A
  }
}
```

scalac issues this warning:

```
pos.scala:2: warning: Class a differs only in case from A. Such classes will overwrite one another on case-insensitive filesystems.
class a
      ^
```

I think dotc should do the same. Also affects `pending/run/t920.scala` which has a similar problem involving class/object on MacOS and Windows.

smarter:
GenBCode has commented-out code to handle this: https://github.com/lampepfl/dotty/blob/3004af37018d2f64c08f33500becf939141ec2e8/compiler/src/dotty/tools/backend/jvm/GenBCode.scala#L190-L202

nicolasstucki:
Fixed by #3897

PR #1082, opened 2016-02-14
Dotty REPL initial version

odersky:
An initial version of a REPL for Dotty. Basic functionality works.

Class `dotty.tools.dotc.repl.Main` is the main entry point. To start the REPL, before we have a dedicated command, use

     java dotty.tools.dotc.repl.Main

(when started under scala, it does not find an Execution context. Hopefully somebody else is better than me at figuring this out).

A comment below explains how the REPL code was derived, and contains a list of TODOs with suggestions to improve it. I am very much hoping for the contributions of others to move this forward. 

Review by @hubertp or @DarkDimius or @smarter or anybody wishing to contribute.


odersky:
Explanations, copied from comment on `repl.Main.scala`:

This REPL was adapted from an old (2008-ish) version of the Scala
REPL. The original version from which the adaptation was done is found in:

     https://github.com/odersky/legacy-svn-scala/tree/spoon

The reason this version was picked instead of a more current one is that
the older version is much smaller, therefore easier to port. It is also
considerably less intertwined with nsc than later versions.

There are a number of TODOs:

 - re-enable jline support (urgent, easy, see TODO in InteractiveReader.scala)
 - figure out why we can launch REPL only with `java`, not with `scala`.
 - make a doti command (urgent, easy)
 - create or port REPL tests (urgent, intermediate)
 - make interpreter run a pseudo line on startup to pre-load compiler (somewhat urgent, easy)
 - copy improvements of current Scala REPL wrt to this version
   (somewhat urgent, intermediate)
 - re-enable bindSettings (not urgent, easy, see TODO in InterpreterLoop.scala)
 - make string generation more functional (not urgent, easy)
 - better handling of ^C (not urgent, intermediate)
 - syntax highlighting (not urgent, intermediate)
 - integrate with presentation compiler for command completion (not urgent, hard)

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -0,0 +1,831 @@
...
+          else Interpreter.Error
+        }
+    }
+  }
+
+  /** A counter used for numbering objects created by <code>bind()</code>. */
+  private var binderNum = 0
+
+  override def bind(name: String, boundType: String, value: Any)(implicit ctx: Context): Interpreter.Result = {
+    val binderName = "binder" + binderNum
```
DarkDimius: maybe `binder$` to ensure absence of name clashes?

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -0,0 +1,831 @@
...
+    compileString(
+      "object " + binderName +
+        "{ var value: " + boundType + " = _; " +
+        " def set(x: Any) = value=x.asInstanceOf[" + boundType + "]; }")
+
+    val binderObject =
+      Class.forName(binderName, true, classLoader)
+    val setterMethod =
+      binderObject.getDeclaredMethods.find(_.getName == "set").get
+    var argsHolder: Array[Any] = null // this roundabout approach is to try and make sure the value is boxed
```
DarkDimius: I'm not sure what's happening is this line and next 3 lines.
Why not simply pass `(value: Object)`

odersky: I don't know either. The whole thing is disabled now. I am not really sure what `bindSettings` is intended to accomplish. It's on the todo list to figure this out.

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -0,0 +1,831 @@
...
+  import CompilingInterpreter._
+
+  /** directory to save .class files to */
+  val virtualDirectory =
+    if (ictx.settings.d.isDefault(ictx)) new VirtualDirectory("(memory)", None)
+    else new PlainDirectory(new Directory(new java.io.File(ictx.settings.d.value(ictx)))) // for now, to help debugging
+
+  /** A GenBCode phase that uses `virtualDirectory` for its output */
+  private class REPLGenBCode extends GenBCode {
+    override def outputDir(implicit ctx: Context) = virtualDirectory
```
smarter: Instead of using a custom phase, couldn't we just make sure the `Context` instance we use in the compiler is set up so that `settings.d.value` is set to `virtualDirectory` ?

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -0,0 +1,831 @@
...
+      !compileCtx.reporter.hasErrors && {
+        this.typeOf = findTypes(compileCtx)
+        val resultCtx = compileSources(
+          List(new SourceFile("<console>", resultObjectSourceCode.toCharArray)))
+        !resultCtx.reporter.hasErrors
+      }
+    }
+
+    /** Dig the types of all bound variables out of the compiler run.
+     *  TODO: Change the interface so that we typecheck, and then transform
```
smarter: I don't really understand this comment, what interface? What transformations?

smarter:
I'd be interested in hearing from @retronym or others familiar with the scalac REPL (@som-snytt ?) about pitfalls we could fall into and mistakes we are likely to repeat. (Suggestions on cool things we could do are also welcome!)

som-snytt:
I subscribed to this PR this morning because I'm interested in contributing code. I'll make an effort to make time for it.

VladimirNik:
@odersky To launch repl with scala I used:
```shell
scala -J-Xbootclasspath/a:\
[path]/.ivy2/cache/me.d-d/scala-compiler/jars/scala-compiler-2.11.5-[version-info].jar:\
[path-to-dotty]/dotty/bin/../target/scala-2.11/dotty_2.11-0.1-SNAPSHOT.jar \
dotty.tools.dotc.repl.Main
```
Fork of scala compiler from **me.d-d** and **dotty** are added to bootclasspath.

retronym:
> I'd be interested in hearing from @retronym or others familiar with the scalac REPL

Adding tab completion support to the REPL will be a good way to start experimenting with ways to make the typechecker IDE friendly, without the waiting for integration into Eclipse. This will be handy to validate that the parser and typechecker are fault tolerant and don't discard trees that correspond to the textual sources.

See https://github.com/scala/scala/pull/4766 for an example of such a problem in Scalac, and https://github.com/scala/scala/pull/4725 for the change to use the presentation compiler to drive tab completion.

Try to avoid splicing user-written code into the synthetic wrappers as text. Ideally you want to be able to typecheck the expression as entered after programattically setting up a typechecker context that includes the imports of previously defined symbols and with previously entered imports. Once that is typechecked, it could be spliced into the synthetic wrapper via AST manipulation. This avoids awkward problems of quoting / escaping, and could probably avoid the need to have the deeply nested `$iw` modules polluting the owner chain of definitions.

tests/run/t920.scala
```diff
@@ -1,20 +0,0 @@
-object Test {
```
smarter: Why was this test deleted?

odersky: t920 fails on MacOS. See #1077. I'll push a version in pending.

On Mon, Feb 15, 2016 at 3:01 PM, Guillaume Martres <notifications@github.com
> wrote:

> In tests/run/t920.scala
> <https://github.com/lampepfl/dotty/pull/1082#discussion_r52901262>:
>
> > @@ -1,20 +0,0 @@
> > -object Test {
>
> Why was this test deleted?
>
> —
> <https://github.com/lampepfl/dotty/pull/1082/files#r52901262>.
>



-- 
Martin Odersky
EPFL

smarter: > I'll push a version in pending.

What about just renaming `b` to something else?

odersky: Then I don't know what it would test?


On Mon, Feb 15, 2016 at 5:40 PM, Guillaume Martres <notifications@github.com
> wrote:

> In tests/run/t920.scala
> <https://github.com/lampepfl/dotty/pull/1082#discussion_r52920045>:
>
> > @@ -1,20 +0,0 @@
> > -object Test {
>
> I'll push a version in pending.
>
> What about just renaming b to something else?
>
> —
> <https://github.com/lampepfl/dotty/pull/1082/files#r52920045>.
>



-- 
Martin Odersky
EPFL

smarter: Not sure either, but I don't think that this test was added to scalac to test for that, cf https://issues.scala-lang.org/browse/SI-920 and https://github.com/scala/scala/commit/485a79aa79142c2740a64f8e7a83036067c7ee9c#diff-2a17630cbd90645d1f9a67af6e848c30
If it doesn't test anything useful then we can just drop it.

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
odersky: No, because virtualDirectory does not have a name, so it cannot be used in the -d option.

odersky:
@som-snytt Contributions are very welcome!
@retronym I agree. Widening the interface to the compiler can streamline things a lot and can provide useful new functionality.

odersky:
After discussion at dotty meeting we decided to merge

PR #1092, opened 2016-02-16
Add checks for value classes

VladimirNik:
Added new checks for value classes:
errors in case of incorrect parameter;
improvement to error message if value class is defined as a local class;

Review by @DarkDimius


DarkDimius:
LGTM

PR #1094, opened 2016-02-16
Change nested typer creation to allow creation of derived typers

j2rodrig:
When the symbol completer runs, it may create a new Typer instance. But if the current typer is a derived typer the completer won't create an instance of the derived type. This change adds a method "newLikeThis" to the Typer class that can be overridden by derived typers.

### DIFF
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -508,7 +508,7 @@ class Namer { typer: Typer =>
         if (sym is Module) moduleValSig(sym)
         else valOrDefDefSig(original, sym, Nil, Nil, identity)(localContext(sym).setNewScope)
       case original: DefDef =>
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -508,7 +508,7 @@ class Namer { typer: Typer =>
         if (sym is Module) moduleValSig(sym)
         else valOrDefDefSig(original, sym, Nil, Nil, identity)(localContext(sym).setNewScope)
       case original: DefDef =>
-        val typer1 = new Typer
```
DarkDimius: @odersky, I'm curious why getting a signature of a DefDef requires creating a new Typer\Namer, while ValDefs and TypeDefs are fine using existing one?

odersky:
LGTM

PR #1100, opened 2016-02-17
Refactor Diagnostic

odersky:
Break it out from Reporter and eliminate all dependencies
to Context. This is done so that Diagnostics can be part
of a public and minimal compiler API.

Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/reporting/Reporter.scala
+++ b/src/dotty/tools/dotc/reporting/Reporter.scala
@@ -12,44 +12,9 @@ import config.Printers
 import java.lang.System.currentTimeMillis
 import typer.ErrorReporting.DiagnosticString
 import typer.Mode
+import Diagnostic.{ERROR, WARNING, INFO}
 
 object Reporter {
-
-  private val ERROR = 2
-  private val WARNING = 1
-  private val INFO = 0
-
-  class Diagnostic(msgFn: => String, val pos: SourcePosition, val level: Int) extends Exception {
-    import DiagnosticString._
-
-    private var myMsg: String = null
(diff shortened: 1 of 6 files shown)


src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -12,44 +12,9 @@ import config.Printers
 import java.lang.System.currentTimeMillis
 import typer.ErrorReporting.DiagnosticString
```
smarter: I think the weird test errors are because `DiagnosticString` does not exist anymore.

odersky: Good analysis. Can you take it over and integrate the PR request directly
into your work?

On Wed, Feb 17, 2016 at 6:53 PM, Guillaume Martres <notifications@github.com
> wrote:

> In src/dotty/tools/dotc/reporting/Reporter.scala
> <https://github.com/lampepfl/dotty/pull/1100#discussion_r53202952>:
>
> > @@ -12,44 +12,9 @@ import config.Printers
> >  import java.lang.System.currentTimeMillis
> >  import typer.ErrorReporting.DiagnosticString
>
> I think the weird test errors are because DiagnosticString does not exist
> anymore.
>
> —
> <https://github.com/lampepfl/dotty/pull/1100/files#r53202952>.
>



-- 
Martin Odersky
EPFL

smarter: Sure.

smarter:
Closing, will be integrated in my PR

PR #1102, opened 2016-02-18
Keep named arguments during eta expansion

### LINKED ISSUE #1096: EtaExpansion.liftArgs removes NamedArgs
This could make correlating trees with positions harder.
We have an agreement that NamedArgs should survive until FirstTransform.

The place where it's eliminated:
https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/typer/EtaExpansion.scala#L53

odersky:
Fixes #1096. Review by @xeno-by

### DIFF
--- a/src/dotty/tools/dotc/typer/EtaExpansion.scala
+++ b/src/dotty/tools/dotc/typer/EtaExpansion.scala
@@ -48,13 +48,11 @@ object EtaExpansion {
   }
 
   /** Lift a function argument, stripping any NamedArg wrapper */
-  def liftArg(defs: mutable.ListBuffer[Tree], arg: Tree, prefix: String = "")(implicit ctx: Context): Tree = {
-    val arg1 = arg match {
-      case NamedArg(_, arg1) => arg1
-      case arg => arg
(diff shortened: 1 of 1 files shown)


xeno-by:
Does this `NamedArg` carry a reference to the corresponding method parameter?

DarkDimius:
the `name` field indicates what was the name of method parameter.

src/dotty/tools/dotc/typer/EtaExpansion.scala
```diff
@@ -48,13 +48,11 @@ object EtaExpansion {
   }
 
   /** Lift a function argument, stripping any NamedArg wrapper */
```
smarter: I think the documentation needs to be updated, since we don't strip the NamedArg anymore.

xeno-by:
@DarkDimius Is there a stronger connection, e.g. via a symbol?

DarkDimius:
No. Symbols of arguments of methods are only accessible inside the body of the method in Dotty.

src/dotty/tools/dotc/typer/EtaExpansion.scala
odersky: We do strip it, but then rewrap it around something else. I meant to do the change, but then decided it's too much busywork for too minor an improvement.

PR #1104, opened 2016-02-18
Improvements & bugs that were discovered while implementing specialization.

DarkDimius:
The big one is the change in mixin.
It reduces the size of dotty.jar by 4%.

### DIFF
--- a/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala.disabled
+++ b/src/dotty/tools/dotc/transform/DropEmptyCompanions.scala.disabled
@@ -30,7 +30,7 @@ import dotty.tools.dotc.transform.TreeTransforms.TransformerInfo
  */
 class DropEmptyCompanions extends MiniPhaseTransform { thisTransform =>
   import ast.tpd._
-  override def phaseName = "dropEmpty"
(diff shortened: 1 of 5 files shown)


smarter:
https://github.com/dotty-staging/dotty/commit/89386d05cff42f0f16c4e301749a2dc5ed8c329c and https://github.com/dotty-staging/dotty/commit/13792086f90f9f231c3774e5681556a8c41352b3 should have tests.

DarkDimius:
@smarter https://github.com/dotty-staging/dotty/commit/89386d05cff42f0f16c4e301749a2dc5ed8c329c cannot be tested as this is the usage of internal helper-API that previously was supported and now requires using more low-level api.

Speaking about https://github.com/dotty-staging/dotty/commit/13792086f90f9f231c3774e5681556a8c41352b3 - I'm not sure if there can be a test that does not need specialization at the moment. 

I believe that those changes generally improve robustness of compiler and are worth getting in.

DarkDimius:
@odersky please review. 74d5820 in particular needs careful review.

odersky:
Otherwise LGTM

odersky:
@darkdimius Do you want to do the final changes to get this in now?

PR #1106, opened 2016-02-18
Neg tests check files for // error markers (rebased and updated)

VladimirNik:
Rebased and updated version of https://github.com/lampepfl/dotty/pull/720

Differences between this and previous pull request (https://github.com/lampepfl/dotty/pull/720):
1) All neg tests are updated in accordance with new checks;
2) Starting from  https://github.com/lampepfl/dotty/commit/73fad44a11dafa507e1d358f1bc222b04319a594 CompilerTest doesn't extend DottyTest and we cannot use DottyTest#ctx in CompilerTest. Because of that:
```scala
val allFiles = CompilerCommand.distill(allArgs)(ctx).arguments
```
is changed to:
```scala
val allFiles = (allArgs filter {
  arg => !arg.startsWith("-") && (arg.endsWith(".scala") || arg.endsWith(".java"))
}).toList
```

Review by @odersky @DarkDimius

### DIFF
--- a/test/test/CompilerTest.scala
+++ b/test/test/CompilerTest.scala
@@ -3,9 +3,12 @@ package test
 import dotty.partest.DPConfig
 import dotty.tools.dotc.{Main, Bench, Driver}
 import dotty.tools.dotc.reporting.Reporter
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.config.CompilerCommand
 import scala.collection.mutable.ListBuffer
-import scala.reflect.io.{ Path, Directory, File => SFile }
+import scala.reflect.io.{ Path, Directory, File => SFile, AbstractFile }
 import scala.tools.partest.nest.{ FileManager, NestUI }
+import scala.annotation.tailrec
 import java.io.{ RandomAccessFile, File => JFile }
 
 import org.junit.Test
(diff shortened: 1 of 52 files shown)


DarkDimius:
>is changed to:

AFAIK all special scalac options start from `-`, so I'd add filter that filenames cannot start from `-`

VladimirNik:
@DarkDimius '-' filtering updated

tests/pos/t0273.scala
```diff
@@ -2,6 +2,6 @@ class A
 
 object Test {
 def a = () => ()
-def a[T] = (p:A) => ()
+def a[T] = (p:A) => ()              // error
```
DarkDimius: is this a pos test or a neg test. 
If some option is needed to make test fail, please specify it.

tests/pos/t1279a.scala
```diff
@@ -4,12 +4,12 @@ abstract class M {
 
   type T
   final type selfType = M {type T <: self.T}
-  type actualSelfType >: self.type <: selfType // this no longer compiles because self.type is not a subtype of selfType
+  type actualSelfType >: self.type <: selfType // error: this no longer compiles because self.type is not a subtype of selfType
```
DarkDimius: Same as https://github.com/lampepfl/dotty/pull/1106/files#r53447850

test/test/CompilerTest.scala
```diff
@@ -178,13 +181,140 @@ abstract class CompilerTest {
...
+      val errorLines = reporter.allErrors.map(_.pos)
+      // reporter didn't record as many errors as its errorCount says
+      assert(errorLines.length == nerrors, s"Not enough errors recorded.")
+
+      val allFiles = (allArgs filter {
+          arg => !arg.startsWith("-") && (arg.endsWith(".scala") || arg.endsWith(".java"))
+      }).toList
+      val expectedErrorsPerFile = allFiles.map(getErrors(_))
+
+      // Some compiler errors have an associated source position. Each error
```
DarkDimius: I propose we change all compiler errors to require a source position.
Than neg tests will not need to take `xerrors` as arguments and keep it updated both in file and in test configuration.
Additionally, this would allow to have a neg-dir instead of test-per-neg-test, where every file specifies how many errors it should have with `// error` comments.

VladimirNik: @DarkDimius If we decide to do it I propose to open separate pull request for it. This pull request already has long history.

tests/pos/t0273.scala
VladimirNik: This test is no longer neg test. Comment is removed

tests/pos/t1279a.scala
VladimirNik: This test is also no longer in neg tests. Comment is removed.

DarkDimius:
LGTM

PR #1112, opened 2016-02-19
Value classes: add support for private[this] parameter

VladimirNik:
Allow private[this] parameter in value classes:
```scala
class A(x: Int) extends AnyVal
class B(private[this] val x: Int) extends AnyVal
```
Review by @odersky @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/transform/Getters.scala
+++ b/src/dotty/tools/dotc/transform/Getters.scala
@@ -12,6 +12,7 @@ import Constants._
 import TreeTransforms._
 import Flags._
 import Decorators._
+import ValueClasses._
 
 /** Performs the following rewritings for fields of a class:
  *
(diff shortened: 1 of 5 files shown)


odersky:
I don't think it will work that way. You'll need to create a public accessor for the private field and have all unboxing operations go through the accessor instead of the field.

DarkDimius:
I believe that the easiest implementation would be to drop `Private | PrivateLocal` flags from fields inside value classes just after pickling.

odersky:
That won't work because you might get accidental overrides. If you drop the
flags you need to mangle the name at the same time. But in any case a
non-private parameter is represented as a private field
and a non-private getter. So you might just add the getter and leave the
field in place.


On Fri, Feb 19, 2016 at 5:40 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> I believe that the easiest implementation would be to drop Private |
> PrivateLocal flags from fields inside value classes just after pickling.
>
> —
> <https://github.com/lampepfl/dotty/pull/1112#issuecomment-186292691>.
>



-- 
Martin Odersky
EPFL

VladimirNik:
Update: initial problem with passing tests was due to neg_valueClasses test (I forgot to update the number of errors in tests.scala).

odersky:
LGTM. That was surprisingly simple. Nice!

Issue #1114, opened 2016-02-20
Manual dotty bootstrap fails


odersky:
I just did this:

    /Users/odersky/workspace/dotty/src/dotty/tools/dotc> time java -Xms1g -Xmx3g dotty.tools.dotc.Main -d /classes *.scala */*.scala */*/*.scala

And then did it again (with /classes at the head of the classpath)
I got at first:

    Exception in thread "main" java.lang.NoClassDefFoundError: Ldotty/tools/dotc/core/Phases$PhasesBase$TerminalPhase$;
	at java.lang.Class.getDeclaredFields0(Native Method)
	at java.lang.Class.privateGetDeclaredFields(Class.java:2575)
	at java.lang.Class.getDeclaredField(Class.java:2060)
	at dotty.runtime.LazyVals$.getOffset(LazyVals.scala:71)

This pointed to a problem that somewhere a reference to a removed empty companion object survived. I disabled the dropEmptyCompanion phase and then that part worked. (I instrumented the "getOffset" method and got this.

    /Users/odersky/workspace/dotty/src/dotty/tools/dotc> time java -Xms1g -Xmx3g dotty.tools.dotc.Main -d /classes *.scala */*.scala */*/*.scala
    getting offset of class dotty.tools.dotc.core.Contexts$ContextBase . bitmap$0
    getting offset of class dotty.tools.dotc.config.PathResolver . bitmap$0
    getting offset of class dotty.tools.dotc.core.Contexts$Context . bitmap$0
    getting offset of class dotty.tools.dotc.core.unpickleScala2.Scala2Unpickler . bitmap$0
    getting offset of class dotty.tools.dotc.core.StdNames$ScalaNames . bitmap$0
    getting offset of class dotty.tools.dotc.core.classfile.ClassfileParser . bitmap$0
    getting offset of class dotty.tools.dotc.transform.PatternMatcher$Translator . bitmap$0
    getting offset of class dotty.tools.dotc.transform.PatternMatcher$Translator . bitmap$0
    getting offset of class dotty.tools.dotc.core.Definitions . bitmap$0
    getting offset of class dotty.tools.dotc.core.StdNames$ScalaTermNames . bitmap$0
    getting offset of class dotty.tools.dotc.typer.VarianceChecker . bitmap$0

So, this is fine. But after some more time compiling with the bootstrapped compiler, I got this:

[...]

PR #1115, opened 2016-02-20
Ensure that partest runs bootstrapped Dotty.

DarkDimius:
This pr is going to fail. Putting it here so that we can collaboratively fix regressions that stop us from running partest bootstrapped again.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -177,7 +177,7 @@ object DottyBuild extends Build {
           path = file.getAbsolutePath
         } yield "-Xbootclasspath/p:" + path
         // dotty itself needs to be in the bootclasspath
-        val fullpath = ("-Xbootclasspath/a:" + bin) :: path.toList
+        val fullpath = ("-Xbootclasspath/p:" + "dotty.jar") :: ("-Xbootclasspath/a:" + bin) :: path.toList
         // System.err.println("BOOTPATH: " + fullpath)
 
         val travis_build = // propagate if this is a travis build
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -171,16 +171,12 @@ object desugar {
 
     def defaultGetters(vparamss: List[List[ValDef]], n: Int): List[DefDef] = vparamss match {
       case (vparam :: vparams) :: vparamss1 =>
-        def resultTpt = vparam.tpt match {
-          case ByNameTypeTree(result) => result
-          case tpt => tpt
-        }
         def defaultGetter: DefDef =
           DefDef(
             name = meth.name.defaultGetterName(n),
             tparams = meth.tparams.map(tparam => dropContextBound(toDefParam(tparam))),
             vparamss = takeUpTo(normalizedVparamss, n),
-            tpt = resultTpt,
+            tpt = TypeTree(),
             rhs = vparam.rhs
           ).withMods(Modifiers(mods.flags & AccessFlags, mods.privateWithin))
         val rest = defaultGetters(vparams :: vparamss1, n + 1)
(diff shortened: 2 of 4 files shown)


DarkDimius:
One issue was pointed out by @odersky in #1114 that refernces to needed companions get dropped. I believe that this could be a bug due to fact that `DropEmptyCompanions` is in the miniphase block where all the scopes\members&info are wrong and lying.

DarkDimius:
The other issue is #1116

odersky:
@DarkDimius: No, unfortunately the problem is different and harder. dropEmptyCompanions does not look in other compilation units when it decides to drop a companion object. Concretely: We drop an innner companion object in Phases, but the surviving reference is in Contexts. I have to idea how to do this in the face of separate compilation. 

It seems it is time to go back to the idea of not generating synthetic companion objects eagerly in the first place. We can do one of the following things:

 - generate companion objects on demand only, when a phase requires them, or 
 - not put generated stuff in companion objects, make it static instead.

DarkDimius:
> but the surviving reference is in Contexts

why would Contexts need a reference to Phases$PhasesBase$TerminalPhase$, if Phases$PhasesBase didn't need it so that it was able to eliminate it? I'd like to have a look in more detail to understand what's happening here.

I do agree that making all synthetic lazy-vals and value-class needed methods\fields is a better way to go, while generating companions on-demand is easier.

odersky:
It's the initialization of the lazy val in the implementing class, I think.

On Sat, Feb 20, 2016 at 2:44 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> but the surviving reference is in Contexts
>
> why would Contexts need a reference to Phases$PhasesBase$TerminalPhase$,
> if Phases$PhasesBase didn't need it so that it was able to eliminate it?
> I'd like to have a look in more detail to understand what's happening here.
>
> I do agree that making all synthetic lazy-vals and value-class needed
> methods\fields is a better way to go, while generating companions on-demand
> is easier.
>
> —
> <https://github.com/lampepfl/dotty/pull/1115#issuecomment-186600602>.
>



-- 
Martin Odersky
EPFL

odersky:
I propose to revert the workaround commit and rebase.

odersky:
Looks like the partest project needs to be recompiled. It still defines default methods with the long names.

DarkDimius:
The CI does not mainain any state, recompiling would not help.
This branch is rebased over https://github.com/lampepfl/dotty/pull/1119
Partest at the moment is compiled by scalac(as it's part of tests) and Dotty is compiled by dotty.
The failure seems to be a different manifestation of https://github.com/lampepfl/dotty/issues/1116

odersky:
It is a different error, also having to do with default getters. It seems we really did not have very good test coverage for this, otherwise we should have found the errors before.

odersky:
Previous commit setting type explicitly was a red herring. Without it, the test `defaultGetters` does not compile. The fix has to lies elsewhere.

odersky:
Repeated GC overhead limit exceeded failures.

smarter:
There's also some kind of pattern-matching bug: https://gist.github.com/smarter/1d04f869870df03d0976

smarter:
https://github.com/dotty-staging/dotty/commit/00638e96d6f4de6c1cd730c52ba1b3281de14b6b should be removed or reverted from this PR now that we know we're using the correct classes.

DarkDimius:
> There's also some kind of pattern-matching bug: https://gist.github.com/smarter/1d04f869870df03d0976

I'm not sure if this is a bug in patmat. The exception correctly indicates a missing match for NoType$

smarter:
Indeed, the corresponding line is https://github.com/dotty-staging/dotty/blob/7bf01004020106ec1639df61268afdf841694e1e/src/dotty/tools/dotc/core/SymDenotations.scala#L782 so it seems like the bootstrapped compiler is able to produce a ClassInfo for a module without a self-type, even though we should always have one.

odersky:
Once #1289 is merged, this should be rebased and tried again. My attempts at manual bootstrap work fine.

DarkDimius:
@odersky. I've rebased it as you've asked. It fails on 3 tests, including the `defaultGetters` added by you in this PR. Could you take it from here?

odersky:
I am seeing an AOM in validate-partest.

retronym:
Tracking the infrastructure problem (I saw it too) as https://github.com/scala/scala-jenkins-infra/issues/181

PR #1117, opened 2016-02-20
Do not create companions that will be dropped later.

DarkDimius:
Fix blocker bug reported in #1114
I dislike this fix as now phase needs to know in advance
if it will ever need a companion for the class.
On the bright side, this change makes it clear
which phases need companions


DarkDimius:
@odersky please review.

DarkDimius:
Single test failed: TASTY pickling with [error] java.lang.OutOfMemoryError: GC overhead limit exceeded

Fix passes all the test on my machine and is ready for review.

src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -46,6 +47,12 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer {
     * before this phase starts processing same tree */
   override def runsAfter = Set(classOf[Mixin])
 
+  def isCompanionNeeded(cls: ClassSymbol)(implicit ctx: Context): Boolean = {
+    println("testing " + cls)
```
smarter: I assume you'll drop the println before merging?

DarkDimius: Thanks for noticing.

smarter:
I think DropEmptyCompanions.scala should be deleted too, otherwise it's easy to think that it's still in use.

odersky:
@smarter There are actually some other phases as well which are still in transfrom even though they are not used. We could get rid of them all maybe, or rename to .scala.disabled (?)

DarkDimius:
rebased.

odersky:
LGTM

PR #1119, opened 2016-02-20
Default getters accessibility fix

### LINKED ISSUE #1116: Dotty and Scalac generate different names for default arguments
```
[info] [Loaded dotty.partest.DPDirectCompiler from file:/Users/dark/workspace/dotty/target/scala-2.11/test-classes/]
[info] [Loaded dotty.tools.dotc.reporting.ConsoleReporter from dotty.jar]
...
[error] java.lang.NoSuchMethodError: dotty.tools.dotc.reporting.ConsoleReporter$.$lessinit$greater$default$1()Ljava/io/BufferedReader;
[error] 	at dotty.partest.DPDirectCompiler.compile(DPDirectCompiler.scala:19)
```

The actual method created by Dotty has fullname&sig: 
`dotty$tools$dotc$reporting$ConsoleReporter$$$$lessinit$greater$default$1()Ljava/io/BufferedReader;`
while scalac looks for:
`$lessinit$greater$default$1()Ljava/io/BufferedReader;`

odersky:
Default getters should have same accessibility as the method they belong to.
Previously, it was the accessibility of the parameter, which makes no sense.
Fixes #1116.

Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -186,7 +186,7 @@ object desugar {
             vparamss = takeUpTo(normalizedVparamss, n),
             tpt = TypeTree(),
             rhs = vparam.rhs
-          ).withMods(vparam.mods & AccessFlags)
+          ).withMods(mods & AccessFlags)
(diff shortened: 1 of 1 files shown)


DarkDimius:
LGTM

Issue #1110, opened 2016-02-19
Implement support for negDir


DarkDimius:
Neg tests at the moment take expected number of errors as an argument. Since https://github.com/lampepfl/dotty/pull/1106 we can optionally mark lines where errors are expected.

I propose we change all negtests to require marked lines.
This would allow to have a neg-dir instead of test-per-neg-test. This in turn would allow to simply drop a file there and will reduce number of merge conflicts.

odersky:
+1

PR #1122, opened 2016-02-22
Neg tests: remove xerror parameter from neg tests (#1110)

VladimirNik:
Neg tests at the moment take expected number of errors as an argument. Since #1106 we can optionally mark lines where errors are expected.

With this change we don't need to pass the expected number of errors as an argument to neg tests anymore. In addition most of the neg files don't require separate test (except for files that should use specific arguments for compilation or directories that should be compiled inside `neg` dir):

```scala
@Test def neg_all = compileFiles(negDir, verbose = true, compileSubDirs = false)
```

New subfolders in neg dir:

- `compExceptions` - contains tests that result in exception during the compilation (currently only structural.scala);
- `customArgs` - contains neg tests that should be compiled with specific arguments;

Several files were in neg dir but were not added to `tests.scala` as tests. Now they became a part of neg_all test:
```shell
tests/neg/Main.scala
tests/neg/i827.scala
tests/neg/implicitDefs.scala
tests/neg/t1292.scala
tests/neg/t7278.scala
```

Several files that were in neg dir have no compilation errors. Now they are moved to pos tests:
```shell
tests/neg/patternUnsoundness.scala
tests/neg/t1164.scala
```

Related issue: #1110
Review by @DarkDimius


test/test/CompilerTest.scala
```diff
@@ -140,54 +153,69 @@ abstract class CompilerTest {
...
+  private def compileArgs(args: Array[String], expectedErrorsPerFile: List[ErrorsInFile], negTest: Boolean)
       (implicit defaultOptions: List[String]): Unit = {
     val allArgs = args ++ defaultOptions
     val processor = if (allArgs.exists(_.startsWith("#"))) Bench else Main
     val reporter = processor.process(allArgs)
 
     val nerrors = reporter.errorCount
+    val xerrors = (expectedErrorsPerFile map {_.totalErrors}).sum
+    if (negTest)
+      assert((negTest && xerrors > 0), s"No errors found in neg test")
```
DarkDimius: this if is strange

test/test/CompilerTest.scala
```diff
@@ -140,54 +153,69 @@ abstract class CompilerTest {
...
-  private def compileArgs(args: Array[String], xerrors: Int = 0)
+  private def expectedErrors(filePaths: List[String], negTest: Boolean): List[ErrorsInFile] = {
+    if (negTest) {
+      filePaths.map(getErrors(_))
+    } else Nil
+  }
+
+  private def expectedErrors(filePath: String, negTest: Boolean): List[ErrorsInFile] = expectedErrors(List(filePath), negTest)
+
+  private def compileArgs(args: Array[String], expectedErrorsPerFile: List[ErrorsInFile], negTest: Boolean)
```
DarkDimius: isn't it always true that `negTest == expectedErrorsPerFile.nonEmpty`?
If yes, I propose to remove negTest from argument list and give `expectedErrorsPerFile` default value of `Nil`.

test/dotc/tests.scala
```diff
@@ -108,88 +108,22 @@ class tests extends CompilerTest {
...
-  @Test def neg_autoTupling2 = compileFile(negDir, "autoTuplingTest", xerrors = 3)
-  @Test def neg_companions = compileFile(negDir, "companions", xerrors = 1)
-  @Test def namedParams = compileFile(negDir, "named-params", xerrors = 14)
-  @Test def neg_over = compileFile(negDir, "over", xerrors = 3)
-  @Test def neg_overrides = compileFile(negDir, "overrides", xerrors = 14)
-  @Test def neg_overrideClass = compileFile(negDir, "overrideClass", List("-language:Scala2"), xerrors = 1)
-  @Test def neg_i39 = compileFile(negDir, "i39", xerrors = 2)
-  @Test def neg_i50_volatile = compileFile(negDir, "i50-volatile", xerrors = 3)
-  @Test def neg_zoo = compileFile(negDir, "zoo", xerrors = 12)
+  @Test def neg_all = compileFiles(negDir, verbose = true, negTest = true, compileSubDirs = false)
```
DarkDimius: Do we need indication that directory\file is a neg-test anymore? 
We can discover this automatically by looking if file contains `// error`

odersky: That might be a bit too fragile. Also, I think it is good to have a specified directory where we put the neg tests.

test/dotc/tests.scala
```diff
@@ -108,88 +108,22 @@ class tests extends CompilerTest {
...
+  @Test def neg_typedIdents() = compileDir(negDir, "typedIdents", negTest = true)
+
+  val negCustomArgs = negDir + "customArgs/"
+  @Test def neg_typers() = compileFile(negCustomArgs, "typers", negTest = true)(allowDoubleBindings)
+  @Test def neg_overrideClass = compileFile(negCustomArgs, "overrideClass", List("-language:Scala2"), negTest = true)
+  @Test def neg_autoTupling = compileFile(posDir, "autoTuplingTest", args = "-language:noAutoTupling" :: Nil, negTest = true)
```
odersky: I think neg tests should be distinguished by the directory they are in. We should avoid a separate boolean flag negTest. In particular, a neg test should not be in a pos dir (as is the case for autoTuplingTest). I realize it is being done in order not to copy the test file, but I think in this case copying (or linking) is the lesser evil.

VladimirNik:
Update: now we don't need negTest flag anymore. Neg tests are binded to neg dir.

test/test/CompilerTest.scala
```diff
@@ -79,59 +79,69 @@ abstract class CompilerTest {
...
 
   /** Compiles the code files in the given directory together. If args starts
     * with "-deep", all files in subdirectories (and so on) are included. */
-  def compileDir(prefix: String, dirName: String, args: List[String] = Nil, xerrors: Int = 0, runTest: Boolean = false)
+  def compileDir(prefix: String, dirName: String, args: List[String] = Nil, runTest: Boolean = false)
       (implicit defaultOptions: List[String]): Unit = {
-    if (!generatePartestFiles || !partestableDir(prefix, dirName, args ++ defaultOptions, xerrors)) {
-      if (runTest)
-        log(s"WARNING: run tests can only be run by partest, JUnit just verifies compilation: $prefix$dirName")
+    val computeFilePathsAndExpErrors = { () =>
```
odersky: I see no reason why this should be a `val`. Make it a `def`?

odersky:
Otherwise LGTM

test/test/CompilerTest.scala
VladimirNik: Fixed in last commit

PR #1123, opened 2016-02-22
ConsoleReporter: handling of non-sensical messages is now reusable 

smarter:
Review by @odersky , this PR is based on https://github.com/lampepfl/dotty/pull/1100 with the fix mentioned in https://github.com/lampepfl/dotty/pull/1100#discussion_r53202952
<hr>
This is now handled by a separate trait HideNonSensicalMessages that can
be mixed in, similar to UniqueMessagePositions. This way we'll be able
to reuse this functionality for other kind of Reporters.

This also means that we don't need `doReport` to return a Boolean
anymore, so we change it to return a Unit as it did before
065a002


odersky:
LGTM

PR #1124, opened 2016-02-24
Fix FatalError's constructor to forward `msg` to super.

sjrd:
Previously, `getMessage()` always returned `null`, causing NPEs when trying to report a `FatalError`.

### DIFF
--- a/src/dotty/tools/dotc/Driver.scala
+++ b/src/dotty/tools/dotc/Driver.scala
@@ -1,5 +1,6 @@
 package dotty.tools.dotc
 
+import dotty.tools.FatalError
 import config.CompilerCommand
 import core.Contexts.{Context, ContextBase}
 import util.DotClass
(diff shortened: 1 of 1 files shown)


sjrd:
Review by @smarter?

smarter:
This change LGTM, though it's weird that we have both a `FatalError` defined in `Driver.scala` and a `FatalError` forwarded from `scala.reflect.internal` defined in [src/dotty/tools/package.scala](https://github.com/lampepfl/dotty/blob/36ce4bcb47252b488c728378278c9d3b10f83e78/src/dotty/tools/package.scala#L5-L6)

sjrd:
Hum indeed.

Should we try and remove one or the other?

smarter:
Yup, I'll let you decide which one :)

sjrd:
Updated.

odersky:
The change re FatalError looks good to me.

PR #1126, opened 2016-02-26
Initial infrastructure and hello world for the Scala.js back-end.

sjrd:
The Scala.js back-end can be enabled with the `-scalajs` command-line option. Currently, it adds one phase to the pipeline, which emits .sjsir files from trees.

A sandbox project `sjsSandbox`, in `sandbox/scalajs/`, can be used to easily test Scala.js compilation. One can run the `main()` method of the `hello.world` object with

    > sjsSandbox/run

The back-end only contains the bare mimimum to compile the hello world application in the sandbox. Anything else will blow up (for example, primitive method calls). It is a work-in-progress.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -4,6 +4,9 @@ import java.io.{ RandomAccessFile, File }
 import java.nio.channels.FileLock
 import scala.reflect.io.Path
 
+import org.scalajs.sbtplugin.ScalaJSPlugin
+import org.scalajs.sbtplugin.ScalaJSPlugin.autoImport._
+
 object DottyBuild extends Build {
 
   val jenkinsMemLimit = List("-Xmx1300m")
@@ -33,6 +36,10 @@ object DottyBuild extends Build {
     )
   }
 
+  /** Enforce 2.11.5. Do not let it be upgraded by dependencies. */
+  private val overrideScalaVersionSetting =
+    ivyScala := ivyScala.value.map(_.copy(overrideScalaVersion = true))
+
   lazy val `dotty-interfaces` = project.in(file("interfaces")).
     settings(
       // Do not append Scala versions to the generated artifacts
@@ -44,6 +51,8 @@ object DottyBuild extends Build {
   lazy val dotty = project.in(file(".")).
     dependsOn(`dotty-interfaces`).
     settings(
+      overrideScalaVersionSetting,
+
       // set sources to src/, tests to test/ and resources to resources/
       scalaSource in Compile := baseDirectory.value / "src",
       javaSource in Compile := baseDirectory.value / "src",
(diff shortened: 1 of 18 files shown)


sjrd:
I am not sure whether you would consider this for inclusion in the repo, at this point. But I figured it was in a good enough state to suggest it.

In any case, I would very much like comments on the general aspect. Please tell me if I'm doing anything stupid at this point.

It might be interesting to compare this with the scalac implementation at https://github.com/scala-js/scala-js/tree/master/compiler/src/main/scala/org/scalajs/core/compiler
In particular, the "giant" `JSCodeGen` is mostly parts of [`GenJSCode`](https://github.com/scala-js/scala-js/blob/master/compiler/src/main/scala/org/scalajs/core/compiler/GenJSCode.scala).

I guess all of @odersky @DarkDimius @smarter might be interested in having a look at this, if not all the details.
Also /cc @nicolasstucki and @gzm0.

DarkDimius:
>In any case, I would very much like comments on the general aspect. Please tell me if I'm doing anything stupid at this point.

I followed most of the code. Ideologically it looks good to me. I'd be curious to see how much different would be handling of `Try` and code generated by patmat compared to scalac. Please tell me if you'll need help understanding some details(especially in LabelDefs).

sjrd:
Rebased. I saved the earlier comments at https://github.com/sjrd/dotty/commit/57c1b1c7fdee3129d44be8b5c3e0c4c7d7bdc373
Note that, if you comment on the Files tab *of the PR* (instead of in the commits), the comments survive rebases on GitHub.

sjrd:
@DarkDimius I added a commit moving `JSDefinitions` to a custom `SJSPlatform` as you suggested. This required to have a valid `ctx` with properly initialized `settings` when creating the `ContextBase.platform`, which required to defer its initialization.

I am not sure if the new situation is desirable. But in any case the required changes highlight that, in master, `platform` is a useless concept, since it is not configurable.

sjrd:
Are OOMs in partest CI common? Or are the failures in fact not spurious?

smarter:
I'm getting them too so it's probably not your fault, we're just steadily using more and more memory, you can try increasing https://github.com/lampepfl/dotty/blob/master/project/Build.scala#L11 and see if it helps.

sjrd:
It appears to have helped, indeed.

So, from my side, I'm happy with this PR the way it is. I guess it deserves a real review, now.

sjrd:
@DarkDimius Following our conversation from this afternoon, I have added a commit addressing your concern about the bootstrap'ability of scalajs-ir. Instead of depending on the compiled artifacts, the build now depends on the sources and compiles them as part of dotty. In bootstrap tests, they are therefore part of the compilation of dotty by itself.

sjrd:
As for *testability*, something we could do would be to simply add a CI command `sbt sjsSandbox/run`. This triggers the Dotty+Scala.js compilation, the linking (including verying the IR--similar to how the JVM verifies bytecode), and the execution with Rhino of the generated .js code.

DarkDimius:
@sjrd, for testability, I leave this for you to decide. Please tell me if you wish to add it to CI.
LGTM

sjrd:
Thanks!

> Please tell me if you wish to add it to CI.

I think it would be worth it, yes.

smarter:
How long do the tests take?

sjrd:
> How long do the tests take?

`sjsSandbox/run`? Less than 3 seconds on my machine.

smarter:
OK then, add it :).

PR #1127, opened 2016-02-28
Upgrade sbt to 0.13.11 and scalastyle-sbt-plugin to 0.8.0

smarter:
Upgrading sbt will allow us to experiment with bootstrapping
using sbt since 0.13.11 is the first release that supports building
with dotty (see https://github.com/smarter/dotty-bridge).

Upgrading scalastyle-sbt-plugin allows us to remove the workaround for
https://github.com/scalastyle/scalastyle/issues/156

Review by @DarkDimius

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.9
+sbt.version=0.13.11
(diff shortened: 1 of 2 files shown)


DarkDimius:
LGTM

PR #1128, opened 2016-02-29
Make Types#membersBasedOnFlags safer and faster

smarter:
By filtering excluded flags earlier, we avoid forcing some denotations.

Review by @DarkDimius


DarkDimius:
compileStdLib seems to take forever now.

DarkDimius:
Seems to be a non-stability of CI.

PR #1129, opened 2016-02-29
Rename AllFlags to AnyFlags

smarter:
`membersBasedOnFlags(requiredFlags = AnyFlags, excludedFlags = ...)` is
easier to understand than
`membersBasedOnFlags(requiredFlags = AllFlags, excludedFlags = ...)`

Review by @DarkDimius


DarkDimius:
This commit does not seem to actually affect `requiredFlags` of `membersBasedOnFlags`.
In the place that it does change, `flagMask` seem to be actually `AllFlags` as you mask all flags.

PR #1133, opened 2016-03-01
Fixes for lambda lift

odersky:
Lambda lift previously ignored the problem of local traits. The transformation would add fields
for any free variables of such traits, ignoring the problem that these fields cannot be transformed anymore because LambdaLift runs after addGetters and Mixin. 

The problem is fixed by letting local traits not cache fields.


adriaanm:
Thanks! This helped enormously to  develop a fix for what was blocking M4: https://github.com/scala/scala/compare/2.12.x...adriaanm:traits-fields-lambda?expand=1 (WIP, but included test case is passing)

odersky:
@adriaanm I discovered more problems which are all reflected in run/llift.scala. I had to touch quite a lot of places in LambdaLift to make these work. Fortunately, they make the logic less quirky, because
we do not jump around constructors anymore, (which we did a lot before).

src/dotty/tools/dotc/transform/ExplicitOuter.scala
```diff
@@ -308,7 +313,7 @@ object ExplicitOuter {
     /** The path of outer accessors that references `toCls.this` starting from
```
smarter: Comment should be updated to say "starting from the node \`start\`" I think.

adriaanm:
Cool, looks much nicer! I'll port these over hopefully tomorrow.

odersky:
Now everything is done. @adriaanm do you want to review?

adriaanm:
Sure, happy to provide feedback while I'm backporting this. It'll be a few more days, though. I'm working on this on and off today, but traveling back to SF tomorrow.

src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -148,27 +197,29 @@ class LambdaLift extends MiniPhase with IdentityDenotTransformer { thisTransform
...
-            ctx.debuglog(i"$sym is free in $enclosure")
+          else markFree(sym, enclosure.enclosure)
+        narrowLiftedOwner(enclosure, intermediate orElse sym.enclosingClass)
+        if (!intermediate.isRealClass || enclosure.isConstructor) {
+          // Constructors and methods nested inside traits get the free variables
+          // of the enclosing trait or class.
+          // Conversely, local traits do not get free variables.
+          if (!enclosure.is(Trait)) {
+            val ss = symSet(free, enclosure)
+            if (!ss(sym)) {
```
adriaanm: The lines `val ss = ....; if (...) ; ss += sym` could be simplified to `if (symSet(free, enclosure) add sym)`.

adriaanm:
(I'll start with some easy suggestions, but I hope to provide some more high-level feedback once I've digested it all.)

src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -85,15 +121,26 @@ class LambdaLift extends MiniPhase with IdentityDenotTransformer { thisTransform
     private def symSet(f: LinkedHashMap[Symbol, SymSet], sym: Symbol): SymSet =
       f.getOrElseUpdate(sym, newSymSet)
 
-    def proxies(sym: Symbol): List[Symbol] = {
-      val pm: Map[Symbol, Symbol] = proxyMap.getOrElse(sym, Map.empty) // Dotty deviation: Type annotation needed. TODO: figure out why
-      free.getOrElse(sym, Nil).toList.map(pm)
+    def freeVars(sym: Symbol): List[Symbol] = free.getOrElse(sym, Nil).toList
```
adriaanm: Could we come up with a more descriptive name than `free`? `storedCapturesOf`?

src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -195,17 +245,16 @@ class LambdaLift extends MiniPhase with IdentityDenotTransformer { thisTransform
...
+            if (isLocal(sym)) {
               if (sym is Label)
                 assert(enclosure == sym.enclosure,
                   i"attempt to refer to label $sym from nested $enclosure")
               else if (sym is Method) markCalled(sym, enclosure)
               else if (sym.isTerm) markFree(sym, enclosure)
-            } else if (sym.maybeOwner.isClass)
-              narrowTo(sym.owner.asClass)
+            }
+            if (sym.maybeOwner.isClass) narrowTo(sym.owner.asClass)
```
adriaanm: What's the reasoning behind the call to `narrowTo` here? Before `narrowTo` was only called for non-local symbols, now it's called for all of them.

adriaanm: I'm asking because it would be kind of nice to add `if sym.isTerm && isLocal(sym)` as a guard to the Ident case (and `sym.isMethod && isLocal(sym)`) and perhaps treat the narrowing separately.

src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -216,8 +265,18 @@ class LambdaLift extends MiniPhase with IdentityDenotTransformer { thisTransform
                 // On the other hand, all other methods will be indirectly owned by their
                 // top-level class. This avoids possible deadlocks when a static method
                 // has to access its enclosing object from the outside.
-            else if (sym.isPrimaryConstructor && sym.owner.owner.isTerm)
-              symSet(called, sym) += sym.owner
+            else if (sym.isConstructor) {
```
adriaanm: collapse with nested if

src/dotty/tools/dotc/transform/LambdaLift.scala
odersky: Not sure that's better.

src/dotty/tools/dotc/transform/LambdaLift.scala
odersky: If the owner of a symbol is a class, then previously that meant the symbol has to be non-local. Now, it could be local, but then it's either owned by a trait or is a constructor. In both cases we
do not want to move the symbol out of the enclosing class.

src/dotty/tools/dotc/transform/LambdaLift.scala
odersky: That's done in a later commit.

PR #1139, opened 2016-03-03
Change isCompanionNeeded

odersky:
Fix an inefficiency when compiling stdlib. The fix brings compilation time down from 185s to 44s.
But I seem to remember we used to compile stdlib in ~25s. We should investigate what else
could have slowed it down.

Review by @darkdimius.


DarkDimius:
LGTM

PR #1141, opened 2016-03-03
Do not choose implicits that are ambiguous references

smarter:
Before this commit, implicit_ambiguous_ref.scala failed to compile
because the implict search returned `x`, we now always try to typecheck
a reference to an identifier before returning it from the implicit search.


DarkDimius:
both partests failed. Does not look like false failure to me:
```
[error] Test dotc.tests.compileMixed failed: java.lang.AssertionError: assertion failed: Wrong # of errors. Expected: 0, found: 29, took 1.174 sec
```

smarter:
Oh, I missed those, the failures I saw were due to download failures.

smarter:
Closing since this was the wrong approach and #1142 needs to be take care of first.

PR #1142, opened 2016-03-03
Do the implicit search shadowing check in the correct context

smarter:
This commit fixes a very sneaky bug, the following code:
```scala
lazy val shadowing =
  typed(untpd.Ident(ref.name) withPos pos.toSynthetic, funProto)
       (nestedContext.addMode(Mode.ImplicitShadowing).setExploreTyperState)
```
is parsed by scalac as:
```scala
lazy val shadowing =
  typed(untpd.Ident(ref.name) withPos pos.toSynthetic, funProto);
(nestedContext.addMode(Mode.ImplicitShadowing).setExploreTyperState);
```
So we don't actually use the nested context in `typed`, instead we end
up implicitly using `ctx`!

Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/typer/Implicits.scala
+++ b/src/dotty/tools/dotc/typer/Implicits.scala
@@ -491,8 +491,8 @@ trait Implicits { self: Typer =>
             pt)
         val generated1 = adapt(generated, pt)
         lazy val shadowing =
-          typed(untpd.Ident(ref.name) withPos pos.toSynthetic, funProto)
-               (nestedContext.addMode(Mode.ImplicitShadowing).setExploreTyperState)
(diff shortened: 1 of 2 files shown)


smarter:
This sort of bug is the reason why GCC 6 added a warning `-Wmisleading-indentation`, I wonder if we could do the same in scala.

smarter:
The tests now fail with `assertion failed: inconsistent: some typevars were added to uncommittable constraint`.

@odersky : What's the harm of adding typevars to an uncommittable constraint set? Could we bypass this check?

smarter:
Changing the assertion check was enough to make all tests pass, so I think this is good to go now.

odersky:
LGTM

PR #1147, opened 2016-03-05
Fix #1146: Fix bug of ExpandSAMs with non-apply SAM methods.

### LINKED ISSUE #1146: AbstractMethodError with non-JVM SAMs with a non-apply method
The following crashes at run-time with an `AbstractMethodError`:
```scala
abstract class NonJVMSam {
  def foo(x: Int): Int
}

object Test {
    val f: NonJVMSam = x => x + 1
    println(f.foo(3))
  }
}
```
That's because `ExpandSAMs` always generates a forwarder for a method named `apply`, irrespective of the name of the method in the actual SAM type.

sjrd:
When expanding a SAM, ExpandSAMs always used the name `apply` for the generated forwarder, instead of the name of the method which is abstract in the SAM type. This commit fixes this issue.


sjrd:
Review by @odersky, as you have written most of `ExpandSAMs` in the first place.

sjrd:
> For future PR's, it's best to push them to staging first and then make the PR from there. That lets everyone else push further commits on a PR with minimum hassle.

Ah OK.

> But the way it is is also fine, so I leave it up to you whether you want to change it.

Done.

PR #1151, opened 2016-03-07
Add support for raw docstrings in ASTs

felixmulder:
Hello, as some of you might know (@heathermiller, @odersky) - I'm coming to EPFL in a couple of weeks to help out on **dottydoc** :tada:

As such, I wanted to get a head start on the implementation. I've perused the code and started an implementation for docstring parsing.

In this PR the scanner has been modified to save the docstrings according to the following heuristic:

* In the scanner, parse all comments
* ~~If the comment is a docstring, save it in a `closestDocString: Option[Comment]` var, replacing the old value~~
* If the comment is a docstring, save it in a `closestDocString: mutable.Queue[Comment]` val
* If `keepComments` is set to true also save it to `revComments` (not sure if this is necessary, but didn't want to ruin someone else's possibly ongoing implementation)
* In the parser, put the saved docstring into the `MemberDef` tree if applicable

I'm not sure if this is the optimal implementation, but it seems to eliminate the need for positional checking and instead follows "if the docstring is the closest docstring to an entity, it is the entity's docstring." The `closestDocString` var would in this approach, however, have to be emptied at the end of parsing bodies (not done yet).

There are two questions I'm looking to have answered:

1. Am I on the right track, or is there a better way to go about this?
2. I want to add tests that check the resulting AST using some example code - where should I place these? Do we have similar tests somewhere I can have a look at?

**Sidenote:** in a separate branch I've also started the implementation of a dottydoc tool which uses only the `FrontEnd` phase to get an AST. As such, finding the correct solution to adding comments to the AST will be a requirement for moving forward with said branch.


DarkDimius:
```
./src/dotty/tools/dotc/typer/Typer.scala:551: error: ')' expected but '}' found.
  }

```

does not seem to compile.

felixmulder:
Thanks @DarkDimius, it should compile - even so, this PR does not touch the typer.

smarter:
> 2. I want to add tests that check the resulting AST using some example code - where should I place these? Do we have similar tests somewhere I can have a look at?

Look in https://github.com/lampepfl/dotty/tree/master/test/test, you might need to extend [`DottyTest`](https://github.com/lampepfl/dotty/blob/master/test/test/DottyTest.scala), note that most of these tests haven't been maintained in a long time and may be broken in subtle ways, feel free to clean up/refactor things if it helps you.

felixmulder:
@smarter thanks! I'll have a look later today.

felixmulder:
Traits without a body caused the structure of only keeping one docstring to fail (since the parser would gobble them up before finding out that the trait had no body). So the docstrings are now saved in a queue. Maybe that's what the `revComments` list was for.

felixmulder:
Good afternoon, I now feel that this PR is nearing a state where it should be evaluated on a first review basis. When first opened, it did not meet the quality criteria and for this I apologize (@DarkDimius).

The tests now show what this implementation can handle, please see: https://github.com/lampepfl/dotty/pull/1151/files#diff-a8ce7031a634f23d40bb1c04c8aeaf38R57

These are the tests performed, and should help you with an overview:

* NoComment
* SingleClassInPackage
* MultipleOpenedOnSingleClassInPackage
* MultipleClassesInPackage
* SingleCaseClassWithoutPackage
* SingleTraitWihoutPackage
* MultipleTraitsWithoutPackage
* MultipleMixedEntitiesWithPackage
* NestedClass
* NestedClassThenOuter
* Objects
* ObjectsNestedClass
* PackageObject
* MultipleDocStringsBeforeEntity
* MultipleDocStringsBeforeAndAfter
* ValuesWithDocString
* VarsWithDocString
* DefsWithDocString
* TypesWithDocString

The current heuristic for associating comments is simply selecting the closest docstring (preceding the entity) from a `List[List[Comment]]` (where the outer list represents the block scope).

Entering and leaving a block scope will, push respectively pop the docstring comments var.

Speaking with Heather, I've been told to ping @odersky (since he is responsible for this part of the compiler).

I hope I'm not stepping on any more toes. Sincerely,
Felix

DarkDimius:
> it did not meet the quality criteria and for this I apologize (@DarkDimius).

It's not it wasn't meeting quality criteria. It's that it was failing a specific test so I've tried to tell you what file to try to compile locally to reproduce the issue.

I'll have a look at the PR, though getting feedback from @odersky would be great.

src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -317,6 +317,10 @@ object Trees {
     private[ast] def rawMods: Modifiers[T] =
       if (myMods == null) genericEmptyModifiers else myMods
 
+    private[this] var myComment: Option[String] = None
```
DarkDimius: hmmm, adding new fields in `trees` is likely to increase memory footprint.
AS most trees do not have comments,@felixmulder, would you consider putting comments into attachments of the tree?

src/dotty/tools/dotc/parsing/Scanners.scala
```diff
@@ -182,6 +182,38 @@ object Scanners {
      */
     var revComments: List[Comment] = Nil
 
+    /** The currently closest docstring, replaced every time a new docstring is
+     *  encountered
+     */
+    var closestDocString: List[List[Comment]] = List(List())
```
DarkDimius: could you please elaborate on why it this a `List[List[Comment]]`? the type seems no to correspond to name\comment.

test/test/DottyDocTests.scala
```diff
@@ -0,0 +1,506 @@
...
+  def source: String
+
+  /** A test to run against the resulting code */
+  def assertion: PartialFunction[Tree[Untyped], Unit]
+
+  private def defaultAssertion: PartialFunction[Tree[Untyped], Unit] = {
+    case x => assert(false, "Couldn't match resulting AST to expected AST in: " + x.show)
+  }
+
+  private def compiler(assertion: PartialFunction[Tree[Untyped], Unit]) = new Compiler {
```
DarkDimius: DottyTest already has a similar method `compilerWithChecker(checkAfterPhase: String, source:String)(assertion:(tpd.Tree, Context) => Unit)`
It would be better to reuse it instead of re-implementing.

src/dotty/tools/dotc/parsing/Scanners.scala
felixmulder: So let's say we have this scenario:

```scala
/** Docstring 1*/
object O {
  /** Docstring 2 */
  class Inner

  /** Cheeky docstring */
}

class Outer
```

The class `Outer` should not have a docstring. So when entering / exiting blocks - I push and pop the docstring environment.

But sure - a better name for it would be something like  ~ `docStringsForBlock`

src/dotty/tools/dotc/ast/Trees.scala
felixmulder: Of course - didn't know about attachments. Will look into it :+1:

test/test/DottyDocTests.scala
felixmulder: `DottyTest` uses a typed tree for their method while the scanner simply adds the comments to the untyped tree - I could parameterize the function? Or should I be adding the docstrings to the typed tree as well?

test/test/DottyDocTests.scala
```diff
@@ -0,0 +1,506 @@
...
+    MultipleDocStringsBeforeEntity,
+    MultipleDocStringsBeforeAndAfter,
+
+    ValuesWithDocString,
+    VarsWithDocString,
+    DefsWithDocString,
+    TypesWithDocString
+  )
+
+  def main(args: Array[String]): Unit = {
```
DarkDimius: Would you kindly make this into `JUnit` test?
This would allow to remove a lot of boilerplate in this object and the companion.

DarkDimius: IE, you can make a method-per-every DottyDocTest, that only calls 

`checkCompile("frontend", source)(assertion)`

test/test/DottyDocTests.scala
```diff
@@ -0,0 +1,506 @@
...
+ *
+ * Instances also need to be placed in the `DottyDocTests::tests` sequence to
+ * be included in the run
+ */
+trait DottyDocTest extends DottyTest { self =>
+  /** Source code in string format */
+  def source: String
+
+  /** A test to run against the resulting code */
+  def assertion: PartialFunction[Tree[Untyped], Unit]
```
DarkDimius: For any non-trivial check\error reporting, you would need an appropriate context.
Have a look how it's done in `DottyTest` it has a lot of infrastructure that you can use.

test/test/DottyDocTests.scala
felixmulder: Certainly =)

src/dotty/tools/dotc/parsing/Scanners.scala
DarkDimius: you seem to operate with this list as if it was a Stack. Maybe make it be a stack?

felixmulder: That was my original thought, but I wasn't sure how much of the standard library stuff we could use - but I'll change it to a stack on your recommendation.

DarkDimius: Sorry, I mean call it a stack, eg `docsPerBlockStack`, I didn't mean to propose to use a deprecated collection :-).

test/test/DottyDocTests.scala
DarkDimius: The `assertion` in `compilerWithChecker` is also provided a `Context`. You can use it to extract `ctx.compilationUnit.untpdTree` in a similar way how you do now.

felixmulder:
@DarkDimius the two latest commits contain your suggested changes :+1: thanks for taking a look!

DarkDimius:
Otherwise LGTM. Good job!

src/dotty/tools/dotc/parsing/Scanners.scala
```diff
@@ -182,6 +182,38 @@ object Scanners {
      */
     var revComments: List[Comment] = Nil
```
felixmulder: Thanks for the LGTM @DarkDimius - regarding this `revComments` variable - should I delete it? I can not see any purpose it might serve now...

felixmulder:
Rebased and fixupped the latest commit to include the change "back" to `setComment` according to your previous suggestion @DarkDimius.

Let me know if this needs to be squished in order to be merged. Thanks!

DarkDimius:
@felixmulder is this still a WIP branch, or is it ready for review?

felixmulder:
@DarkDimius - ready for review!

odersky:
LGTM, Thanks!

Issue #1137, opened 2016-03-01
Separate compilation with value class fails due to missing companion object


smarter:
`A_1.scala`:
```scala
class A(val self: Double) extends AnyVal
```
`B_2.scala`:
```scala
object B {
  def foo: A = new A(1)
}
```
```shell
% ./bin/dotc A_1.scala
% ./bin/dotc B_2.scala
exception occurred while compiling B.scala
Exception in thread "main" java.lang.AssertionError: NoDenotation.owner
        at dotty.tools.dotc.core.SymDenotations$NoDenotation.owner(SymDenotations.scala:1799)
        at dotty.tools.dotc.ast.tpd$.ref(tpd.scala:346)
        at dotty.tools.dotc.transform.Erasure$Boxing$.cast(Erasure.scala:230)
        at dotty.tools.dotc.transform.Erasure$Boxing$$anonfun$unbox$2.apply(Erasure.scala:200)
        at dotty.tools.dotc.transform.Erasure$Boxing$$anonfun$unbox$2.apply(Erasure.scala:175)
...
```
This does not happen if I manually create a companion in `A_1.scala`:
```scala
class A(val self: Double) extends AnyVal
object A
```

odersky:
I guess this is due to the recent changes where we create companion objects only when we know they will be needed. @DarkDimius do you agree?

smarter:
@odersky : I tried to set `needsCompanion` to `true` in `FirstTransform` so that we always generate companions and the issue still manifest itself. @DarkDimius suggested that the issue is that `FirstTransform` only operates on trees, not types, and when we unpickle `A` we only transform its type, so we never add its missing companion.

DarkDimius:
@odersky that's indeed independent for the recent change.
I have suggested @VladimirNik to try to move creation of symbols for companion objects to info-transformer of FirstTransform.

odersky:
Question to clarify:

The problem arises specifically when we read a value class from another
dotty module via TASTY? It does not arise when unpickling Scala 2 value
classes?



On Thu, Mar 3, 2016 at 8:28 PM, Dmitry Petrashko <notifications@github.com>
wrote:

> @odersky <https://github.com/odersky> that's indeed independent for the
> recent change.
> I have suggested @VladimirNik <https://github.com/VladimirNik> to try to
> move creation of symbols for companion objects to info-transformer of
> FirstTransform.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #1153, opened 2016-03-07
Fix for separate compilation with value class issue (#1137)

VladimirNik:
Fix for #1137: Separate compilation with value class fails due to missing companion object
Review by @smarter @odersky


odersky:
LGTM

PR #1154, opened 2016-03-07
First steps towards rewriting from Scala2 in dotty

odersky:
This is an attempt to do the rewriting right in the dotty compiler. The logic is, if dotty can detect
Scala2 incompatibilities it should be able to apply fixes at the same time. That way there is no error-prone syncing between different tools.

This PR

 - adds rewrite logic
 - adds logic to navigate from typed to untyped trees
 - performs specific rewritings: 
    - rewrite procedure syntax, 
    - add @volatile for lazy vals
    - drop redundant trailing `_'.
    - mask variance errors with @uncheckedVariance

There is also a change the way annotations are propagated to derived symbols.

With this PR, stdlib can be patched and recompiled without -language:Scala2 (pending fix of
@volatile implicit issues).


src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -983,7 +986,18 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       case rhs @ Ident(nme.WILDCARD) => rhs withType tpt1.tpe
       case rhs => typedExpr(rhs, tpt1.tpe)
     }
-    assignType(cpy.ValDef(vdef)(name, tpt1, rhs1), sym)
+    val vdef1 = assignType(cpy.ValDef(vdef)(name, tpt1, rhs1), sym)
+    patchIfLazy(vdef1)
+    vdef1
+  }
+
+  /** Add a @volitile to lazy vals when rewriting from Scala2 */
```
xeno-by: Typo

odersky:
Any objections to merge this one?

smarter:
I'll review it today.

src/dotty/tools/dotc/typer/VarianceChecker.scala
```diff
@@ -108,11 +110,13 @@ class VarianceChecker()(implicit ctx: Context) {
...
 
   private object Traverser extends TreeTraverser {
-    def checkVariance(sym: Symbol) = Validator.validateDefinition(sym) match {
+    def checkVariance(sym: Symbol, pos: Position) = Validator.validateDefinition(sym) match {
       case Some(VarianceError(tvar, required)) =>
         def msg = i"${varianceString(tvar.flags)} $tvar occurs in ${varianceString(required)} position in type ${sym.info} of $sym"
-        if (ctx.scala2Mode && sym.owner.isConstructor)
-          ctx.migrationWarning(s"According to new variance rules, this is no longer accepted; need to annotate with @uncheckedVariance:\n$msg", sym.pos)
+        if (ctx.scala2Mode && sym.owner.isConstructor) {
+          ctx.migrationWarning(s"According to new variance rules, this is no longer accepted; need to annotate with @uncheckedVariance:\n$msg, pos = ${sym.pos}", sym.pos)
```
smarter: I don't think we need `pos = ${sym.pos}` in the migration warning

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -498,16 +497,16 @@ object desugar {
 
   /** If `pat` is a variable pattern,
    *
-   *    val/var p = e
+   *    val/var/lazy val p = e
    *
    *  Otherwise, in case there is exactly one variable x_1 in pattern
-   *   val/var p = e  ==>  val/var x_1 = (e: @unchecked) match (case p => (x_1))
+   *   val/var/lazy val p = e  ==>  val/var x_1 = (e: @unchecked) match (case p => (x_1))
```
smarter: `val/var x_1` should be `val/var/lazy val x_1` I think

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -498,16 +497,16 @@ object desugar {
...
    *   in case there are zero or more than one variables in pattern
-   *   val/var p = e  ==>  private synthetic val t$ = (e: @unchecked) match (case p => (x_1, ..., x_N))
-   *                   val/var x_1 = t$._1
+   *   val/var/lazy p = e  ==>  private synthetic [lazy] val t$ = (e: @unchecked) match (case p => (x_1, ..., x_N))
+   *                   val/var/def x_1 = t$._1
    *                   ...
-   *                  val/var x_N = t$._N
+   *                   val/var/def x_N = t$._N
    *  If the original pattern variable carries a type annotation, so does the corresponding
    *  ValDef.
```
smarter: `so does the corresponding ValDef or DefDef`

src/dotty/tools/dotc/ast/NavigateAST.scala
```diff
@@ -0,0 +1,83 @@
...
+    tree match {
+      case tree: MemberDef[_] =>
+        untypedPath(tree.envelope) match {
+          case path @ (last: DefTree[_]) :: _ => path
+          case path if !exactMatch => path
+          case _ => Nil
+        }
+      case _ =>
+        untypedPath(tree.pos) match {
+          case path @ last :: _ if last.pos == tree.pos || !exactMatch => path
```
smarter: Syntactic suggestion: write `case path @ (last :: _) if ...` to make visual parsing of this line easier.

src/dotty/tools/dotc/ast/Positioned.scala
```diff
@@ -58,17 +58,43 @@ abstract class Positioned extends DotClass with Product {
    *  and transitively visit their children.
    */
   private def setChildPositions(pos: Position): Unit = {
```
smarter: So what was wrong with the previous version of this method? And was it impossible to write the new version using recursive calls? I think it'd be helpful to add some comments explaining how this algorithm work and what are the invariants for all the mutable variables.

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1761,13 +1762,18 @@ object Parsers {
      *  DefSig ::= id [DefTypeParamClause] ParamClauses
      */
     def defDefOrDcl(mods: Modifiers): Tree = atPos(tokenRange) {
-      def scala2ProcedureSyntax =
-        testScala2Mode("Procedure syntax no longer supported; `: Unit =' should be inserted here")
+      def scala2ProcedureSyntax(resultTypeStr: String) = {
+        val toInsert = if (in.token == LBRACE) s"$resultTypeStr =" else ": Unit "
```
smarter: Shouldn't this be `else ": Unit"` to avoid inserting a trailing space?

src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -64,19 +66,17 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer with Nee
...
     else {
-
       val isField = sym.owner.isClass
-
       if (isField) {
         if (sym.isVolatile ||
-          (sym.is(Flags.Module) && !sym.is(Flags.Synthetic)))
-          // module class is user-defined.
-          // Should be threadsafe, to mimic safety guaranteed by global object
+           (sym.is(Flags.Module)/* || ctx.scala2Mode*/) && !sym.is(Flags.Synthetic)) // TODO assume @voliat
```
smarter: Typo: `@volatile`

test/test/CompilerTest.scala
```diff
@@ -89,7 +90,20 @@ abstract class CompilerTest {
     if (!generatePartestFiles || !partestableFile(prefix, fileName, extension, args ++ defaultOptions)) {
       if (runTest)
         log(s"WARNING: run tests can only be run by partest, JUnit just verifies compilation: $prefix$fileName$extension")
-      compileArgs((s"$filePath" :: args).toArray, expErrors)
+      if (args.contains("-rewrite")) {
```
smarter: Instead of implicitly doing additional checks when one of the arg is `-rewrite`, I think it'd be better to ask for these checks explicitly, either have a `compileFileCheckRewrite` method or an additional argument `checkRewrite: Boolean = false` to `compileFile`.

src/dotty/tools/dotc/rewrite/Rewrites.scala
```diff
@@ -0,0 +1,99 @@
...
+      out.close()
+    }
+  }
+
+  /** If -rewrite is set, record a patch that replaces the range
+   *  given by `pos` in `source` by `replacement`
+   */
+  def patch(source: SourceFile, pos: Position, replacement: String)(implicit ctx: Context): Unit =
+    ctx.settings.rewrite.value match {
+      case Some(rewrites: Rewrites) =>
```
smarter: Can be written as `case Some(rewrites) =>` since the type of `ctx.settings.rewrite.value` is `Option{Rewrites]`

src/dotty/tools/dotc/rewrite/Rewrites.scala
```diff
@@ -0,0 +1,99 @@
...
+        rewrites.patched.get(source) match {
```
smarter: Suggestion: rewrite this as:
```scala
rewrites.patched.getOrElseUpdated(source, new Patches(source)).addPatch(pos, replacement)
```

src/dotty/tools/dotc/rewrite/Rewrites.scala
```diff
@@ -0,0 +1,99 @@
...
+            patch(source, pos, replacement)
+        }
+      case _ =>
+    }
+
+  /** If -rewrite is set, apply all patches and overwrite patched source files.
+   */
+  def writeBack()(implicit ctx: Context) =
+    ctx.settings.rewrite.value match {
+      case Some(rewrites: Rewrites) =>
```
smarter: can be written `case Some(rewrites) =>` too

src/dotty/tools/dotc/rewrite/Rewrites.scala
```diff
@@ -0,0 +1,99 @@
...
+        for (source <- rewrites.patched.keys) {
```
smarter: why not:
```scala
for ((source, patches) <- rewrites.patched) {
  ctx.println(s"[patched file ${source.file.path}]")
  patches.writeBack()
}

File: src/dotty/tools/dotc/rewrite/Rewrites.scala [hunk 14]
```diff
@@ -0,0 +1,99 @@
...
+            assert(outNew == ds.length, s"$outNew != ${ds.length}")
+        }
+      }
+      loop(patches, 0, 0)
+      ds
+    }
+
+    def writeBack(): Unit = {
+      val out = source.file.output
+      val chars = apply(source.content)
```
smarter: Should probably be `apply(source.underlying.content)` to account for `ScriptSourceFile` although I'd be happier if we could get rid of the script source files suppot and let some external tool handle them.

src/dotty/tools/dotc/transform/LazyVals.scala
smarter: Could you expand on this TODO? Otherwise we're likely to forget what it's for

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -983,7 +986,18 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+  private def patchIfLazy(vdef: ValDef)(implicit ctx: Context): Unit = {
+    val sym = vdef.symbol
+    if (sym.is(Lazy, butNot = Deferred | Module | Synthetic) && !sym.isVolatile &&
+        ctx.scala2Mode && ctx.settings.rewrite.value.isDefined &&
+        !ctx.isAfterTyper)
+      patch(ctx.compilationUnit.source, Position(toUntyped(vdef).envelope.start), "@volatile ")
```
smarter: Minor point but If we always patch in `ctx.compilationUnit.source`, maybe we don't need to pass it as an argument to `patch`

smarter: What's the difference between `vdef.envelope.start` and `toUntyped(vdef).envelope.start`?

smarter:
Probably not worth fixing if it's complicated to deal with but note that:
```scala
object Foo {
  lazy val x, y = 1
}
```
gets rewritten to:
```scala
object Foo {
  @volatile @volatile lazy val x, y = 1
}
```

src/dotty/tools/dotc/rewrite/Rewrites.scala
odersky: Indeed. Good point.

PR #1155, opened 2016-03-07
Implement @static sip.

### LINKED ISSUE #1149: Locks for volatile lazy vals should be static
The following program fails in LambdaLift:

    trait Iterator {

      def span() = {
        val self: Int = 33
        class Leading {
          def finish(): Unit = println("finished")
        }
        val leading = new Leading

        class Trailing {
          @volatile lazy val it = leading.finish()
        }
        val trailing = new Trailing
        (leading, trailing)
      }
    }

We get (with the improvements to EpxlicitOuter in another PR):

Exception in thread "main" java.lang.AssertionError: assertion failed: failure to construct path from value while$/getter it/class Trailing/trait Iterator/package <empty>/package <root> to `this` of trait Iterator;
class Trailing in trait Iterator does not have an outer accessor
	at dotty.tools.dotc.transform.ExplicitOuter$OuterOps$.loop$1(ExplicitOuter.scala:323)
	at dotty.tools.dotc.transform.ExplicitOuter$OuterOps$.path$extension(ExplicitOuter.scala:329)
	at dotty.tools.dotc.transform.LambdaLift$LambdaLifter.memberRef(LambdaLift.scala:362)
	at dotty.tools.dotc.transform.LambdaLift$LambdaLifter.transformIdent(LambdaLift.scala:429)

The problem is that we need to establish a reference to the companion object of `Trailing` in order to
do the CAS. But that requires an explicit outer pointer and ExplicitOuter did not generate one. The best solution, I think is to make the locks static, also from a standpoint of speed.

Currently this is blocking another important change: We should always assume @volatile under Scala2 mode. But right now we cannot because compilestdlib does not compile due to the issue.

DarkDimius:
This pull request implements most of machinery needed for
https://github.com/scala/scala.github.com/pull/491

 Only 3-rd check is not implemented by this commit. I propose to get this in
faster to fix #1149

### DIFF
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -320,7 +320,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       case _ =>
         false
     }
-    try test || tp.symbol.is(JavaStatic)
+    try test || tp.symbol.is(JavaStatic) || tp.symbol.hasAnnotation(defn.ScalaStaticAnnot)
     catch { // See remark in SymDenotations#accessWithin
       case ex: NotDefinedHere => test(ctx.addMode(Mode.FutureDefsOK))
     }
@@ -337,6 +337,8 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     else if (prefixIsElidable(tp)) Ident(tp)
     else if (tp.symbol.is(Module) && ctx.owner.isContainedIn(tp.symbol.moduleClass))
       followOuterLinks(This(tp.symbol.moduleClass.asClass))
+    else if (tp.symbol hasAnnotation defn.ScalaStaticAnnot)
+         Ident(tp)
     else tp.prefix match {
       case pre: SingletonType => followOuterLinks(singleton(pre)).select(tp)
       case pre => SelectFromTypeTree(TypeTree(pre), tp)
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -338,7 +338,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     else if (tp.symbol.is(Module) && ctx.owner.isContainedIn(tp.symbol.moduleClass))
       followOuterLinks(This(tp.symbol.moduleClass.asClass))
     else if (tp.symbol hasAnnotation defn.ScalaStaticAnnot)
-         Ident(tp)
+      Ident(tp)
     else tp.prefix match {
       case pre: SingletonType => followOuterLinks(singleton(pre)).select(tp)
       case pre => SelectFromTypeTree(TypeTree(pre), tp)
--- a/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -285,10 +285,6 @@ object SymDenotations {
     final def addAnnotation(annot: Annotation): Unit =
       annotations = annot :: myAnnotations
 
-    /** Add given annotation to the annotations of this denotation */
-    final def addAnnotation(annot: ClassSymbol)(implicit ctx: Context): Unit =
-      addAnnotation(ConcreteAnnotation(tpd.ref(annot)))
-
     /** Remove annotation with given class from this denotation */
     final def removeAnnotation(cls: Symbol)(implicit ctx: Context): Unit =
       annotations = myAnnotations.filterNot(_ matches cls)
--- a/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -285,6 +285,10 @@ object SymDenotations {
     final def addAnnotation(annot: Annotation): Unit =
       annotations = annot :: myAnnotations
 
+    /** Add given annotation to the annotations of this denotation */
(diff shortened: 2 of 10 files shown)


src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -337,6 +337,8 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     else if (prefixIsElidable(tp)) Ident(tp)
     else if (tp.symbol.is(Module) && ctx.owner.isContainedIn(tp.symbol.moduleClass))
       followOuterLinks(This(tp.symbol.moduleClass.asClass))
+    else if (tp.symbol hasAnnotation defn.ScalaStaticAnnot)
+         Ident(tp)
```
sjrd: Indentation issue?

src/dotty/tools/dotc/transform/CheckStatic.scala
```diff
@@ -0,0 +1,83 @@
...
+        }
+      } else hadNonStaticField = hadNonStaticField || defn.isInstanceOf[ValDef]
+
+    }
+    tree
+  }
+
+  override def transformSelect(tree: tpd.Select)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    if (tree.symbol.hasAnnotation(defn.ScalaStaticAnnot))
+      ref(tree.symbol)
```
sjrd: This will silently drop any side effect of evaluating the qualifier.

If the qualifier is a simple reference to the enclosing object, that's fine (that's part of the semantic changes of `@static`), but it's not fine for `{ x += 1; SomeObject}.someStatic`

src/scala/annotation/static.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import scala.annotation.meta._
+
+/** https://github.com/scala/scala.github.com/pull/491 */
+
+@field
+@getter
+@beanGetter
+@beanSetter
+@param
```
sjrd: Why `@param`?

DarkDimius: For consistency. Now if _any_ kind of member is created from a `@static` one it has to still be `@static`

src/dotty/tools/dotc/transform/CheckStatic.scala
DarkDimius: fix incoming

src/scala/annotation/static.scala
odersky: But parameters cannot be @static?

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -285,6 +285,10 @@ object SymDenotations {
     final def addAnnotation(annot: Annotation): Unit =
       annotations = annot :: myAnnotations
 
+    /** Add given annotation to the annotations of this denotation */
+    final def addAnnotation(annot: ClassSymbol)(implicit ctx: Context): Unit =
+      addAnnotation(ConcreteAnnotation(tpd.ref(annot)))
+
```
odersky: I don't think you need that here. Annotation already contains a bunch of overloaded apply methods to create annotations; we should not add more overloadings in other places because then there are two places to look for the right variant instead of one. You can get waht you want already by writing

    addAnnotation(Annotation(cls, Nil))

If one objects to the `Nil`, one can add another variant of `apply` to `Annotation`.

DarkDimius:
```
symbol.addAnnotation(Annotation(defn.ScalaStaticAnnot)))
```
seems like too much repetition to me, but if you insist, I'll make the change.

src/dotty/tools/dotc/transform/CheckStatic.scala
```diff
@@ -0,0 +1,83 @@
...
+
+        val clashes = companion.asClass.membersNamed(defn.name)
+        if (clashes.exists) {
+          ctx.error("companion classes cannot define members with same name as @static member", defn.pos)
+        }
+
+        if (defn.symbol.is(Flags.Mutable) && companion.is(Flags.Trait)) {
+          ctx.error("Companions of traits cannot define mutable @static fields")
+        }
+      } else hadNonStaticField = hadNonStaticField || defn.isInstanceOf[ValDef]
```
odersky: Drop { ... } around single-statement blocks?

src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -91,6 +91,7 @@ class LazyVals extends MiniPhaseTransform with IdentityDenotTransformer with Nee
     appendOffsetDefs.get(cls) match {
       case None => template
       case Some(data) =>
+        data.defs.foreach(_.symbol.addAnnotation(defn.ScalaStaticAnnot))
```
odersky: To avoid the repetition, why not define in LazyVals:

    def staticAnnot = Annotation(defn.ScalaStaticAnnot, Nil)

src/scala/annotation/static.scala
DarkDimius: Not if written by user. But I would like to leave the door open for tools\optimizers\analyzers.

sjrd: I guess the question is: what *is* a static param, even if produced by a tool/optimizer/analyzer? I don't have any idea what that would look like in the bytecode.

DarkDimius: Simply a static field.

sjrd: Then it's covered by `@field`, isn't it?

odersky:
LGTM

Issue #1152, opened 2016-03-07
/*/* not recognized as a double comment opening


sjrd:
The following code:
```scala
package hello

object world extends App {
  println("hello dotty!")
  /*/* one
  */
  two
  */
  println("foo")
}
```
fails to compile with
```
[error] examples/hello.scala:7: error: not found: two
[error]   two
[error]   ^
[error] examples/hello.scala:8: error: not found: */
[error]   */
[error]   ^
[error] two errors found
```
So, apparently, the opening `/*/*` is considered as a single comment opening instead of two.

Inserting a space in `/* /*` works around the issue.

PR #1156, opened 2016-03-07
Allow successive opening comments.

odersky:
Fixes #1052. Review by @sjrd


sjrd:
The issue number in the commit message is not correct. It should be #1152 (instead of #1052).

Otherwise LGTM.

PR #1157, opened 2016-03-08
Doc on how to build on Eclipse

odersky:
This is what works for me. I have written it down so that I remember
the next time I have to update the project. If others have different
schemes, please comment and edit.

### DIFF
--- a/docs/how-to-build-on-eclipse.md
+++ b/docs/how-to-build-on-eclipse.md
@@ -1,0 +1,20 @@
+To set up the dotty project on Eclipse, or if the build has changed, perform the following
+steps:
+
+1. Run sbt eclipse
+2. In dotty, select properties (CTRL-I), select "java build path" and then "Libraries".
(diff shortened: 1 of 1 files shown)


sjrd:
1 + 2 is basically what I do.

Also, I have to explicitly change the Compliance levels of `dotty-interfaces` to use Java 8 compliant settings. Otherwise it complains on `default` methods.

smarter:
I can't comment on the content but I suggest putting that kind of information on the wiki, the [Getting Started](https://github.com/lampepfl/dotty/wiki/Getting-Started) page already contains information on the Scala IDE, maybe update that?

odersky:
@smarter The linked wiki page is for users of `dotc`. This one if for contributors. So I don't think they should be the same page. But I am OK with putting it on the wiki as a separate page if people prefer that.

odersky:
I opened a wiki page with the contensts

PR #1160, opened 2016-03-09
Add/collection strawman

odersky:
Two additiions:

 1. A new strawman centered around views rather than iterators
 2. An improvement to overloading resolution which lets us use the Scala 2.8 calling
     idioms for `map` on `String`. Previously, this required an explicit type annotation.

### DIFF
--- a/src/strawman/collections/CollectionStrawMan4.scala
+++ b/src/strawman/collections/CollectionStrawMan4.scala
@@ -1,0 +1,486 @@
+package strawman.collections
+
+import Predef.{augmentString => _, wrapString => _, _}
+import scala.reflect.ClassTag
+
+/** A strawman architecture for new collections. It contains some
+ *  example collection classes and methods with the intent to expose
+ *  some key issues. It would be good to compare this to other
+ *  implementations of the same functionality, to get an idea of the
+ *  strengths and weaknesses of different collection architectures.
+ *
+ *  For a test file, see tests/run/CollectionTests.scala.
+ */
+object CollectionStrawMan4 {
+
+  /* ------------ Base Traits -------------------------------- */
+
+  /** Iterator can be used only once */
+  trait IterableOnce[+A] {
+    def iterator: Iterator[A]
+  }
+
+  /** Base trait for generic collections */
+  trait Iterable[+A] extends IterableOnce[A] with FromIterable[Iterable] {
+    def iterator: Iterator[A]
+    def view: View[A] = View.fromIterator(iterator)
+    def knownLength: Int = -1
+  }
+
+  /** Base trait for instances that can construct a collection from an iterator */
+  trait FromIterable[+C[X] <: Iterable[X]] {
+    def fromIterable[B](v: Iterable[B]): C[B]
+  }
+
+  /** Base trait for companion objects of collections */
+  trait IterableFactory[+C[X] <: Iterable[X]] extends FromIterable[C] {
+    def empty[X]: C[X] = fromIterable(View.Empty)
+    def apply[A](xs: A*): C[A] = fromIterable(View.Elems(xs: _*))
+  }
+
+  /** Base trait for sequence collections */
+  trait Seq[+A] extends Iterable[A] with FromIterable[Seq] {
+    def apply(i: Int): A
+    def length: Int
+  }
+
+  trait Builder[-A, +To] {
+    def +=(x: A): this.type
+    def ++=(xs: IterableOnce[A]): Unit = xs.iterator.foreach(+=)
+    def result: To
+  }
+
+  /* ------------ Operations ----------------------------------- */
+
+  /** Operations returning types unrelated to current collection */
+  trait Ops[A] extends Any {
+    def iterator: Iterator[A]
+    def foreach(f: A => Unit): Unit = iterator.foreach(f)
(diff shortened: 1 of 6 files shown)


src/strawman/collections/CollectionStrawMan4.scala
```diff
@@ -0,0 +1,486 @@
...
+  trait View[+A] extends Iterable[A] with FromIterable[View] {
+    override def view = this
+    override def fromIterable[B](c: Iterable[B]) = c match {
+      case c: View[B] => c
+      case _ => View.fromIterator(c.iterator)
+    }
+  }
+
+  /** Iterator defined in terms of indexing a range */
+  trait RandomAccessView[+A] extends View[A] {
```
DarkDimius: AFAIK Rust initially had `RandomAccessIterator`, but then they decided to drop it. It was creating more problems than solving.

src/strawman/collections/CollectionStrawMan4.scala
```diff
@@ -0,0 +1,486 @@
...
+      (fromIterable(pn.left), fromIterable(pn.right))
+    }
+    def drop(n: Int): Repr = fromIterable(View.Drop(coll, n))
+    def to[C[X] <: Iterable[X]](fv: FromIterable[C]): C[A] = fv.fromIterable(coll)
+  }
+
+  trait PolyTransforms[A, C[X]] extends Any {
+    protected def coll: Iterable[A]
+    protected def fromIterable[B](it: Iterable[B]): C[B]
+    def map[B](f: A => B): C[B] = fromIterable(View.Map(coll, f))
```
lrytz: From the perspective of the optimizer this looks "rather" complex, optimistically speaking. Probably too hard to optimize (cc @sjrd)

Here's how things expand:

```
List(1,2,3).map(_ + 1)

// extension method
new IterablePolyTransforms(List(1,2,3)).map(_ + 1)

// inline map
val pt = new IterablePolyTransforms(List(1,2,3))
pt. fromIterable(View.Map(pt.coll, _ + 1))

// inline IterablePolyTransforms.fromIterable
val pt = new IterablePolyTransforms(List(1,2,3))
pt.coll.fromIterable(View.Map(pt.coll, _ + 1))

// inline List.fromIterable
val pt = new IterablePolyTransforms(List(1,2,3))
List.fromIterable(View.Map(pt.coll, _ + 1))

// inline List$.fromIterable
val pt = new IterablePolyTransforms(List(1,2,3))
View.Map(pt.coll, _ + 1) match {
  case View.Concat(xs, ys: Iterable[B]) => ...
  case View.Drop(xs: List[B], n) => ..
  case _ => fromIterator(c.iterator)
}

// inline fromIterator -- actually not possible because it's recursive in non-tail position
val pt = new IterablePolyTransforms(List(1,2,3))
def fromIt(it: Iterator) = if (it.hasNext) Cons(it.next, fromIterator(it)) else Nil
fromIt(View.Map(pt.coll, _ + 1).iterator)

// stack-allocate View.Map and inline .iterator
val pt = new IterablePolyTransforms(List(1,2,3))
def fromIt(it: Iterator) = if (it.hasNext) Cons(it.next, fromIterator(it)) else Nil
fromIt(pt.coll.iterator.map(_ + 1))

// inline Iterator.map
val pt = new IterablePolyTransforms(List(1,2,3))
def fromIt(it: Iterator) = if (it.hasNext) Cons(it.next, fromIterator(it)) else Nil
fromIt(new Iterator {
  val f = (x: Int) => x + 1
  val self = pt.coll.iterator
  def hasNext = self.hasNext
  def next() = f(self.next())
})
```

If `fromIt` was not recursive we could inline it and start inlining `hasNext` and `next`, and eventually end up with a monomorphic loop. I jumped over many details, I'm really not sure we could push the optimizer that far.

odersky: I am not sure the problem with recursive `fromIterator` will persist in the full system. Instead of using `Cons` we'd use a `ListBuffer` and append at the end. Then `fromIterator` would be tail recursive.

lrytz: sure, but that is not the hard part for the optimizer

lrytz:
Looked a bit more through the strawman

  - Could `Iterator` and `View` be unified?
  - What's the intention use for `IterableOnce`?

DarkDimius:
> Could Iterator and View be unified?

The only diffence between view and iterator lies in fact that views can be traversed several times.
In my variation of the original strawman, I was proposing to make this change:

```
class View[A](val v: ()=> Iterator[A]) extends AnyVal
```

Ie: view is just a producer of iterators

lrytz:
Ah, thanks - so the by-name here is crucial:

```
object View {
 def fromIterator[A](it: => Iterator[A]): View[A] = new View[A] { def iterator = it }
}
```

odersky:
Alternatively, demand that all parameter types are the same. That's what dotty does currently. It compiles the code in SI-9943 OK.

adriaanm:
Yep, this is the main aspect where Scala 2's implementation differs. It seemed more regular to me to just `lub` the argument types, but maybe we need to be more strict, based on this bug report.

adriaanm:
Interestingly, this turned out to be a different problem: we admitted sealed classes (like <:<) as SAM types. [Closing that hole](https://github.com/scala/scala/pull/5442) made the type checker happy.

Issue #1159, opened 2016-03-09
dotty-interfaces publishLocal fails


andreaTP:
This is due to a wrong configuration of javacOptions in ```compile:doc```.

More specifically:
```
javadoc: error - invalid flag: -Xlint:unchecked
javadoc: error - invalid flag: -Xlint:deprecation
```

andreaTP:
closed with #1161

PR #1161, opened 2016-03-09
Removed invalid javacOptions in (Compile, doc)

andreaTP:
Targeting #1159 

this is the smallest fix that let publish, probably you should consider to add the javac options only on relevant Tasks.


DarkDimius:
thanks @andreaTP 
LGTM (hand-checked the CLA http://www.lightbend.com/contribute/cla/scala/check/andreaTP)

PR #1163, opened 2016-03-09
Fix bug where ambiguous references were not reported

### LINKED ISSUE #1145: Dotty allows an import in a block to shadow a prior definition
```scala
object A {
   def x = 3

   def y = {
     import B._
     x
   }
}
object B {
  def x = 3
}
```
This fails in scalac with:
```
A.scala:6: error: reference to x is ambiguous;
it is both defined in object A and imported subsequently by
import B._
     x
     ^
one error found
```
but succeeds in dotty (`B.x` is picked).

odersky:
There was a mssing condition which meant Tyepr thought it was
at the outermost scope where but was mistaken.

Fixes #1145. Review by @smarter


smarter:
LGTM

PR #1164, opened 2016-03-10
Make sure lazy accessors in traits are not private.

### LINKED ISSUE #1140: Separate compilation goes wrong with private object or private lazy val in trait
`A_1.scala`:
```scala
trait A {
  val foo = 0 + x
  private lazy val x = 1
}
```
`B_2.scala`:
```scala
object B extends A {
    println(foo)
  }
}
```
```scala
% ./bin/dotc A_1.scala
% ./bin/dotc B_2.scala
% scala B
java.lang.NoSuchMethodError: A.A$$x()I
        at B$.<init>(B.scala:1)
        at B$.<clinit>(B.scala)
        at B.main(B.scala)
        at java.lang.reflect.Method.invoke(Method.java:497)
        at scala.reflect.internal.util.ScalaClassLoader$$anonfun$run$1.apply(ScalaClassLoader.scala:70)
        at scala.reflect.internal.util.ScalaClassLoader$class.asContext(ScalaClassLoader.scala:31)
        at scala.reflect.internal.util.ScalaClassLoader$URLClassLoader.asContext(ScalaClassLoader.scala:101)
        at scala.reflect.internal.util.ScalaClassLoader$class.run(ScalaClassLoader.scala:70)
        at scala.reflect.internal.util.ScalaClassLoader$URLClassLoader.run(ScalaClassLoader.scala:101)
        at scala.tools.nsc.CommonRunner$class.run(ObjectRunner.scala:22)
        at scala.tools.nsc.ObjectRunner$.run(ObjectRunner.scala:39)
        at scala.tools.nsc.CommonRunner$class.runAndCatch(ObjectRunner.scala:29)
        at scala.tools.nsc.ObjectRunner$.runAndCatch(ObjectRunner.scala:39)
        at scala.tools.nsc.MainGenericRunner.runTarget$1(MainGenericRunner.scala:65)
        at scala.tools.nsc.MainGenericRunner.run$1(MainGenericRunner.scala:87)
        at scala.tools.nsc.MainGenericRunner.process(MainGenericRunner.scala:98)
        at scala.tools.nsc.MainGenericRunner$.main(MainGenericRunner.scala:103)
        at scala.tools.nsc.MainGenericRunner.main(MainGenericRunner.scala)
```
Here's what `A.class` looks like when we compile `A_1.scala` and `B_2.scala` together:
```java
public interface A {
    }

    public int foo();

    default public int initial$foo() {
        return 0 + this.A$$x();
    }

    default public int A$$x() {
        return 1;
    }
}
```
and here's what it looks like when we just compile `A_1.scala`:
```java
public interface A {
    }

    public int foo();

    default public int initial$foo() {
        return 0 + this.x();
    }

    default private int x() {
        return 1;
    }
}
```
(Decompilation done using http://www.benf.org/other/cfr/)

odersky:
Fixes #1140. Review by @DarkDimius or @smarter.

### DIFF
--- a/src/dotty/tools/dotc/core/Flags.scala
+++ b/src/dotty/tools/dotc/core/Flags.scala
@@ -448,7 +448,7 @@ object Flags {
   final val FromStartFlags =
     AccessFlags | Module | Package | Deferred | Final | MethodOrHKCommon | Param | ParamAccessor | Scala2ExistentialCommon |
     InSuperCall | Touched | JavaStatic | CovariantOrOuter | ContravariantOrLabel | ExpandedName | AccessorOrSealed |
-    CaseAccessorOrBaseTypeArg | Fresh | Frozen | Erroneous | ImplicitCommon | Permanent |
+    CaseAccessorOrBaseTypeArg | Fresh | Frozen | Erroneous | ImplicitCommon | Permanent | Synthetic |
     LazyOrTrait | SuperAccessorOrScala2x | SelfNameOrImplClass
 
   assert(FromStartFlags.isTermFlags && FromStartFlags.isTypeFlags)
--- a/src/dotty/tools/dotc/transform/ExpandPrivate.scala
+++ b/src/dotty/tools/dotc/transform/ExpandPrivate.scala
@@ -57,8 +57,11 @@ class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { t
    *  static members of the companion class, we should tighten the condition below.
    */
   private def ensurePrivateAccessible(d: SymDenotation)(implicit ctx: Context) =
-    if (d.is(PrivateTerm) && d.owner != ctx.owner.enclosingClass)
+    if (d.is(PrivateTerm) && d.owner != ctx.owner.enclosingClass) {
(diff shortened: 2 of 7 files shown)


smarter:
So, why was separate compilation needed to reproduce the bug? Could we avoid special-casing non-separate compilation to make it easier to find bugs like this?

odersky:
There are two kinds of tests for making things non-private. One kind makes things not private if they are referenced from some context that demands it. The other makes them not private unconditionally. The failing case was overlooked in the second kind of tests, but still showed up in the first kind. That's why
the error only showed up under separate compilation.

smarter:
Could we make `ExpandPrivate#ensurePrivateAccessible` assert that the denotation we're making public comes from the current compilation unit? If it comes from a different compilation unit and `ensurePrivateAccessible` is needed, then separate compilation will always fail, won't it?

odersky:
> Could we make expandPrivate#ensurePrivateAccessible assert that the denotation we're making public comes from the current compilation unit? 

Good idea. I added commits to do this.

smarter:
Nice! LGTM.

PR #1165, opened 2016-03-10
Pickle/unpickle SuperAccessor names

### LINKED ISSUE #1144: Separate compilation fails with object extending a trait that refers to super
`AB_1.scala`:
```scala
trait A {
  def x = 3
}
trait B extends A {
  override def x = super.x * 2
}
```
`C_2.scala`:
```scala
object C extends B {
  def main(args: Array[String]) = assert(x == args(0).toInt, x)
}
```
```bash
% dotc AB_1.scala
% dotc C_2.scala
C_2.scala:1: error: object creation impossible, since def super$x: => Int is not defined 
object C extends B {
         ^
one error found
```

`A.class` and `B.class` are identical whether or not separate compilation is used:
```java
/*
 * Decompiled with CFR 0_114.
 */
public interface A {
    }

    default public int x() {
        return 3;
    }
}
/*
 * Decompiled with CFR 0_114.
 */
public interface B
extends A {
    @Override
    }

    public /* synthetic */ int B$$super$x();

    @Override
    default public int x() {
        return this.B$$super$x() * 2;
    }
}
```

odersky:
Needed to restore the SuperAccessor flag in separate compilation.
Fixes #1144.

### DIFF
--- a/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -383,7 +383,6 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
         case TastyName.SuperAccessor(_) => Flags.SuperAccessor
         case _ => EmptyFlags
       }
-      val superFlag = if (rawName.isInstanceOf)
       pickling.println(i"creating symbol $name at $start with flags $givenFlags")
       val flags = normalizeFlags(tag, givenFlags | nameFlags(rawName), name, isAbstractType, rhsIsEmpty)
       def adjustIfModule(completer: LazyType) =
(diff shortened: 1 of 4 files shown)


src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -378,9 +378,14 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
       val rhsIsEmpty = noRhs(end)
       if (!rhsIsEmpty) skipTree()
       val (givenFlags, annots, privateWithin) = readModifiers(end)
-      val expandedFlag = if (rawName.isInstanceOf[TastyName.Expanded]) ExpandedName else EmptyFlags
+      def nameFlags(tname: TastyName): FlagSet = tname match {
+        case TastyName.Expanded(_, original) => ExpandedName | nameFlags(tastyName(original))
+        case TastyName.SuperAccessor(_) => Flags.SuperAccessor
+        case _ => EmptyFlags
+      }
+      val superFlag = if (rawName.isInstanceOf)
```
smarter: This line isn't used

VladimirNik:
LGTM.

PR #1166, opened 2016-03-10
Fix typing of SeqLiterals

### LINKED ISSUE #1136: Empty JavaSeqLiteral (vararg) is incorrectly typed, resulting in runtime error
The following crashes at runtime, because the second argument to `getMethod` (an empty vararg) is incorrectly typed:
```scala
object Test {
    val x = classOf[String].getMethod("toString")
  }
}
```
The signature of `getMethod` is `Method getMethod(String name, Class<?>... parameterTypes)`, so the second argument after typing should be an empty array of type `Array[Class[_]]`, but instead we produce an empty array of type `Array[Nothing]`:
```scala
      val x: java.lang.reflect.Method = 
        classOf[String].getMethod("toString", []: Array[Nothing])
      ()
    }
  }
```
This is incorrect because `Array[Nothing]` is not a subtype of `Array[Class[_]]` and it blows up at runtime:
```java
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    Test$.main([Ljava/lang/String;)V @8: invokevirtual
  Reason:
    Type '[Lscala/runtime/Nothing$;' (current frame, stack[2]) is not assignable to
    '[Ljava/lang/Class;'
```

The issue is in [`TypeAssigner#assignType`](https://github.com/lampepfl/dotty/blob/0ae3ef2010b90bf06d76a768b0f0c5aa56c1180a/src/dotty/tools/dotc/typer/TypeAssigner.scala#L395-L397):
```scala
  def assignType(tree: untpd.SeqLiteral, elems: List[Tree])(implicit ctx: Context) = tree match {
    case tree: JavaSeqLiteral =>
      tree.withType(defn.ArrayOf(ctx.typeComparer.lub(elems.tpes).widen))
```
We set the array element type to the lub of the types of the `JavaSeqLiteral` elements, but when the `JavaSeqLiteral` is empty, the lub is `Nothing`.

We could fix the type by hand in [`TypedApply#makeVarArg`](https://github.com/lampepfl/dotty/blob/0ae3ef2010b90bf06d76a768b0f0c5aa56c1180a/src/dotty/tools/dotc/typer/Applications.scala#L445) where the expected type is known, but if we do this then `-Ycheck` will fail since retyping will produce `Array[Nothing]` again.

odersky:
Now it's more like what scalac is doing - unfortunately the previous, far simpler scheme fails in the presence of empty non-variant Java seq literals, as #1136 illustrates.

Fixes #1136.


DarkDimius:
@odersky, as current scheme fails only for empty seq literals, instead of going for the old scalac scheme, would you consider this:

```scala
case class SeqLiteral[-T >: Untyped] private[ast] (elems: List[Tree[T]])
class EmptySeqLiteral[-T >: Untyped] private[ast] (elems: List[Tree[T]], elemtpt: Tree[T]) extends SeqLiteral(elems)
```

Only empty SeqLiterals would need to retain the elem type and only TypeAssigner would needs to know of their existance.

odersky:
@darkdimius It turns out this does not help much, since the element type has to be propagated everywhere in any case.

DarkDimius:
@odersky, as empty sequences are only created by Typer, couldn't EmptySeqLiteral simply remember the prototype that they were created with?

odersky:
@darkdimius Adding the second parameter is precisely for remembering the prototype.

PR #1168, opened 2016-03-11
Tweaks to ExplicitOuter and TreeTypeMap

### LINKED ISSUE #1131: ExplicitOuter crash with local class defined in val in trait if the local class extends a path-dependent class
```scala
class Outer {
  class Inner
}

trait MustBeATrait {
  val o = new Outer
  val inner = new o.Inner {}
}
```
fails with `assertion failed: asTerm called on not-a-Term val <none>` somewhere in `ExplicitOuter`.
Stack trace: https://gist.github.com/smarter/992bb93f2d0f2705a6a6

odersky:
Fixes #1131.

The test shows interesting behavior where we have to copy local classes using a TreeTypeMap. Since
this happens after erasure (in phase Mixin), we need at the same time be prepared to issue explicit outer paths. There were two problems encountered with this, which are fixed in this PR.

### DIFF
--- a/src/dotty/tools/dotc/ast/TreeTypeMap.scala
+++ b/src/dotty/tools/dotc/ast/TreeTypeMap.scala
@@ -82,9 +82,11 @@ final class TreeTypeMap(
           constr = tmap.transformSub(constr),
           parents = parents mapconserve transform,
           self = tmap.transformSub(self),
-          body = impl.body mapconserve tmap.transform
+          body = impl.body mapconserve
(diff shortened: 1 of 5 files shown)


odersky:
The PR is based on top of #1154 because it needed the better explicit outer diagnostic that is part of #1154. Only the last three commits are new.

DarkDimius:
Otherwise, LGTM for last 3 commits.

odersky:
@DarkDimius I rebased the PR to master to better isolate its commits. Can you have a look at dad9dfc
to complete the review?

DarkDimius:
LGTM

PR #1171, opened 2016-03-12
Have checkNotPrivate skip over private aliases.

### LINKED ISSUE #1130: Crash when a class extends a trait with a member signature that contains a private type member
```scala
trait A {
  private type Foo = Int

  def foo: Foo = 1
}
class B extends A
```
fails to compile with
```scala
dotty.tools.dotc.core.Types$MissingType: cannot resolve reference to type B(B.this).Foo
the classfile defining the type might be missing from the classpath
````
Full stack trace: https://gist.github.com/smarter/6a7a6d840ce0f37b1d12

Note that even if we disallow using private types in signatures, we still have to deal with traits coming from Scala 2 where this is allowed, the following fails in exactly the same way:
```scala
abstract class Foo extends scala.collection.mutable.IndexedSeqView[Nothing, Nothing]
```

odersky:
CheckNotPrivate now avoids private aliases, so that types sat prepresentable in
transformation phases.

This does not solve the problem that private classes or abstract types might leak coming
from Scala 2, but there is not really a good cure for that. We can reject them outright or
allow them under language:Scala2 and hope for the best.

Fixes #1130 

Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -131,8 +131,13 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
       annot.derivedAnnotation(transformAnnot(annot.tree))
 
     private def transformMemberDef(tree: MemberDef)(implicit ctx: Context): Unit = {
-      tree.symbol.transformAnnotations(transformAnnot)
-      Checking.checkNoPrivateLeaks(tree)
+      val sym = tree.symbol
+      sym.transformAnnotations(transformAnnot)
+      if (!sym.is(SyntheticOrPrivate) && sym.owner.isClass) {
+        val info1 = Checking.checkNoPrivateLeaks(sym, tree.pos)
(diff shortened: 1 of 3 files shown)


smarter:
> so that types sat prepresentable in
transformation phases

I think there's a few typos in this description, what's "prepresentable"?

smarter:
> This does not solve the problem that private classes or abstract types might leak coming
from Scala 2, but there is not really a good cure for that.

So how come this work in Scala 2 with separate compilation? Is there something we should do differently in unpickling to see these private types?

DarkDimius:
> So how come this works in Scala 2 with separate compilation?

See related issues to https://issues.scala-lang.org/browse/SI-6794 I would say that it works even without separate compilation.

odersky:
> So how come this works in Scala 2 with separate compilation?

I'll have to write this up in detail at some point. It's a fundamental issue.

smarter:
LGTM.

> I'll have to write this up in detail at some point. It's a fundamental issue.

Sounds interesting!

PR #1173, opened 2016-03-14
make variable name consistent in ScalaSettings

liufengyun:
Options with syntax "-Y" should NOT be named with a variable beginning with `X`.

Actually following setting variables are not used in current `dotty`. If there's no plan to support them in the future, it makes more sense to just remove them.


DarkDimius:
LGTM.

>If there's no plan to support them in the future, it makes more sense to just remove them.

AFAIK, All of those options except `Yshowtrees` we are not planning to implement, at least in the near future. Feel free to remove them,

liufengyun:
Thanks @DarkDimius . Actually I see there are a lot of unimplemented options. I'll send a new pull request to clean up all unimplemented options (except those planned).

PR #1176, opened 2016-03-14
fix command line processing error

liufengyun:
Fix command line error for `-help` and `-no-bootcp`.

### DIFF
--- a/bin/dotc
+++ b/bin/dotc
@@ -245,7 +245,7 @@ main_class=$CompilerMain
 while [[ $# -gt 0 ]]; do
 case "$1" in
            --) shift; for arg; do addResidual "$arg"; done; set -- ;;
-     -h|-help) usage; exit 1 ;;
+     -h|-help) help=true && shift ;;
      -bootstrapped) bootstrapped=true && shift ;;
   -v|-verbose) verbose=true && shift ;;
     -debug) debug=true && shift ;;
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -40,7 +40,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val jvmargs = PrefixSetting("-J<flag>", "-J", "Pass <flag> directly to the runtime system.")
   val defines = PrefixSetting("-Dproperty=value", "-D", "Pass -Dproperty=value directly to the runtime system.")
   val toolcp = PathSetting("-toolcp", "Add to the runner classpath.", "")
-  val nobootcp = BooleanSetting("-nobootcp", "Do not use the boot classpath for the scala jars.")
```
DarkDimius: I do not think we should remain options inherited from scalac, unless they have different meaning.

liufengyun: OK, then I  change `./bin/dotc` to accept `-nobootcp` instead of `-no-bootcp`.

liufengyun:
I've updated the commit with only changes to `./bin/dotc`.

odersky:
@DarkDimius Can we merge this?

DarkDimius:
LGTM

PR #1177, opened 2016-03-14
WIP - remove unimplemented command line options

liufengyun:
As we're not going to support a lot of options, I just did a little clean up.

Currently I just removed all options that are not implemented. It's better to just comment out those that we plan to support in the future and create issues for them.

@DarkDimius could you please help review and list the options that are under plan to be implemented? Thanks.


src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -16,16 +16,11 @@ class ScalaSettings extends Settings.SettingGroup {
 
   /** Other settings.
    */
-  val dependencyfile = StringSetting("-dependencyfile", "file", "Set dependency tracking file.", ".scala_dependencies")
   val deprecation = BooleanSetting("-deprecation", "Emit warning and location for usages of deprecated APIs.")
   val migration = BooleanSetting("-migration", "Emit warning and location for migration issues from Scala 2.")
-  val encoding = StringSetting("-encoding", "encoding", "Specify character encoding used by source files.", Properties.sourceEncoding)
```
sjrd: Not even *allowing* `encoding` to be specified will break a lot of builds. It is commonly seen in sbt builds to explicitly specify `-encoding utf-8`.

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -43,72 +38,31 @@ class ScalaSettings extends Settings.SettingGroup {
   val nobootcp = BooleanSetting("-nobootcp", "Do not use the boot classpath for the scala jars.")
   val strict = BooleanSetting("-strict", "Use strict type rules, which means some formerly legal code does not typecheck anymore.")
 
-  val argfiles = BooleanSetting("@<file>", "A text file containing compiler arguments (options and source files)")
```
sjrd: This is important for Windows, which has a limit on the command-line length. Without this option it is impossible to compile the Scala standard library on Windows, for example, because there are too many files to list on the command line.

odersky: OK, so I guess rather than dropping the option we should implement it /
port it from scalac.


On Mon, Mar 14, 2016 at 4:35 PM, Sébastien Doeraene <

> In src/dotty/tools/dotc/config/ScalaSettings.scala
> <https://github.com/lampepfl/dotty/pull/1177#discussion_r56019985>:
>
> > @@ -43,72 +38,31 @@ class ScalaSettings extends Settings.SettingGroup {
> >    val nobootcp = BooleanSetting("-nobootcp", "Do not use the boot classpath for the scala jars.")
> >    val strict = BooleanSetting("-strict", "Use strict type rules, which means some formerly legal code does not typecheck anymore.")
> >
> > -  val argfiles = BooleanSetting("@<file>", "A text file containing compiler arguments (options and source files)")
>
> This is important for Windows, which has a limit on the command-line
> length. Without this option it is impossible to compile the Scala standard
> library on Windows, for example, because there are too many files to list
> on the command line.
>
> —
> <https://github.com/lampepfl/dotty/pull/1177/files#r56019985>.
>



-- 
Martin Odersky
EPFL

liufengyun: Thanks for your feedback. I created this PR just for more clarification about what's to be implemented, what's to be dropped. I'll mark planned options as TODO, create issues for them and start implementing or porting.

liufengyun:
Close this for the moment.

PR #1179, opened 2016-03-15
support `xs @ _*` and `_*` in Scala2 mode

### LINKED ISSUE #1059: Support @ _* syntax under -language:Scala2
Dotty requires `x : _*` syntax for repeated types, though scala required `xs @ _*`.
It would be good to add support for old syntax under compatibility flag and issue a warning.
Also see https://github.com/lampepfl/dotty/issues/1048

liufengyun:
Fix #1059 

The standard syntax in Dotty now is `xs : _*`. In Scala2 mode,
following code should be valid:

    list match {
      case List(_, _, _, _ @ _*)     =>   0
      case List(_, _, _*)            =>   1
      case List(_, _: _*)            =>   2
      case Nil                       =>   3
    }

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -259,11 +259,19 @@ object Parsers {
       } finally inFunReturnType = saved
     }
 
+    private val isScala2Mode =
+      ctx.settings.language.value.contains(nme.Scala2.toString)
+
+    def migrationWarningOrError(msg: String, offset: Int = in.offset) =
+      if (isScala2Mode)
+        ctx.migrationWarning(msg, source atPos Position(offset))
+      else
+        syntaxError(msg, offset)
+
     /** Cannot use ctx.featureEnabled because accessing the context would force too much */
     private def testScala2Mode(msg: String, pos: Position = Position(in.offset)) = {
-      val s2 = ctx.settings.language.value.contains(nme.Scala2.toString)
-      if (s2) ctx.migrationWarning(msg, source atPos pos)
-      s2
+      if (isScala2Mode) ctx.migrationWarning(msg, source atPos pos)
+      isScala2Mode
     }
 
 /* ---------- TREE CONSTRUCTION ------------------------------------------- */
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1282,6 +1287,12 @@ object Parsers {
 
     /* -------- PATTERNS ------------------------------------------- */
 
+    def scala2WildcardStarMessage1  =
```
odersky: I'd refactor a little bit differently.

1. Introduce a method 

        def migrationWarningOrError(msg: String, pos: Int) = ...

2. Drop the message defs here

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1308,7 +1319,28 @@ object Parsers {
      */
     val pattern2 = () => infixPattern() match {
       case p @ Ident(name) if isVarPattern(p) && in.token == AT =>
-        atPos(p.pos.start, in.skipToken()) { Bind(name, infixPattern()) }
+        val pos = in.skipToken()
+
+        // compatibility for Scala2 `x @ _*` syntax
+        infixPattern() match {
+          case pt @ Ident(tpnme.WILDCARD_STAR) =>
```
odersky: 3. Then this code can be written shorter like this:

        migrationWarningOrError("The syntax `x @ _*' is no longer supported; use `x : _*' instead", p.pos)

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1308,7 +1319,28 @@ object Parsers {
...
+            else
+              syntaxError(scala2WildcardStarMessage1, p.pos)
+
+            atPos(p.pos.start, pos) { Typed(p, pt) }
+          case p =>
+            atPos(p.pos.start, pos) { Bind(name, p) }
+        }
+      case p @ Ident(tpnme.WILDCARD_STAR) =>
+        // compatibility for Scala2 `_*` syntax
+        if (isScala2Mode)
```
odersky: (4) An equivalently here.

odersky:
Otherwise LGTM

liufengyun:
Now I've refined the PR.

odersky:
LGTM

PR #1180, opened 2016-03-17
dotc settings: String options after settings with colon shouldn't be ignored

niktrop:
Format of some of the compiler settings is incompatible with scalac. For example "-target:jvm-1.8" is a valid argument for scalac, but it should be "-target: jvm-1.8" (with a whitespace) for dotc. This commit make both ways to work.

### DIFF
--- a/src/dotty/tools/dotc/config/Settings.scala
+++ b/src/dotty/tools/dotc/config/Settings.scala
@@ -115,11 +115,13 @@ object Settings {
         case (ListTag, _) =>
           if (argRest.isEmpty) missingArg
           else update((argRest split ",").toList, args)
+        case (StringTag, _) if choices.nonEmpty =>
+          if (argRest.isEmpty) missingArg
+          else if (!choices.contains(argRest))
+            fail(s"$arg is not a valid choice for $name", args)
+          else update(argRest, args)
         case (StringTag, arg2 :: args2) =>
-          if (choices.nonEmpty && !(choices contains arg2))
-            fail(s"$arg2 is not a valid choice for $name", args2)
-          else
-            update(arg2, args2)
+          update(arg2, args2)
         case (IntTag, arg2 :: args2) =>
           try {
             val x = arg2.toInt


sjrd:
I suspect that expecting a space was a bug in the first place. I would suggest disallowing the variant with the space, but maybe wait for someone else to confirm.

odersky:
Yes, we should not accept a space here.

On Thu, Mar 17, 2016 at 10:47 AM, Sébastien Doeraene <

> I suspect that expecting a space was a bug in the first place. I would
> suggest disallowing the variant with the space, but maybe wait for someone
> else to confirm.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/lampepfl/dotty/pull/1180#issuecomment-197792284>
>



-- 
Martin Odersky
EPFL

niktrop:
I have added a check for non-empty argument without whitespace for such settings.

src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -109,17 +109,21 @@ object Settings {
...
           else update((argRest split ",").toList, args)
-        case (StringTag, arg2 :: args2) =>
-          if (choices.nonEmpty && !(choices contains arg2))
-            fail(s"$arg2 is not a valid choice for $name", args2)
-          else
-            update(arg2, args2)
+        case (StringTag, _) =>
+          if (argRest.isEmpty && args.isEmpty) missingArg
+          else {
+            val (arg, tail) = if (argRest.isEmpty) (args.head, args.tail) else (argRest, args)
```
odersky: We should simplify here and not accept a following argument. I suspect then the whole thing will be shorter (which is what we want).

src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -142,10 +146,16 @@ object Settings {
           missingArg
       }
 
-      if (prefix != "" && arg.startsWith(prefix))
-        doSet(arg drop prefix.length)
-      else if (prefix == "" && name == arg.takeWhile(_ != ':'))
-        doSet(arg.dropWhile(_ != ':').drop(1))
+      if (prefix != "")
```
odersky: I think we can leave the old logic unchanged here. StringSettings should behave like MultiString settings (those which have a ListTag), that's all.

niktrop:
It seems that only ChoiceSettings should behave like Multistring. Other StringSettings don't require colon. I force-pushed simplified version.

odersky:
LGTM, thanks!

PR #1182, opened 2016-03-18
Repl fixes and tests

odersky:
The REPL broke in a recent merge. Nobody noticed because we don't have tests. This PR fixes the breakage and several other bugs and adds tests for the REPL.

### DIFF
--- a/tests/repl/imports.check
+++ b/tests/repl/imports.check
@@ -1,0 +1,24 @@
+scala> import scala.collection.mutable
+import scala.collection.mutable
+scala> val buf = mutable.ListBuffer[Int]()
+buf: scala.collection.mutable.ListBuffer[Int] = ListBuffer()
+scala> object o {
+     |   val xs = List(1, 2, 3)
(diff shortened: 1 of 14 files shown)


odersky:
rebased to master

tests/repl/imports.check
```diff
@@ -0,0 +1,24 @@
...
+defined module o
+scala> import o._
+import o._
+scala> buf += xs
+<console>:11: error: type mismatch:
+ found   : scala.collection.immutable.List[Int](o.xs)
+ required: String
+buf += xs
+       ^
+<console>:11: error: type mismatch:
```
DarkDimius: Why are two errors printed here?
Scalac reports only one.

odersky: Not sure. I might depend on where the error is reported - same position in scalac, different ones in dotty. or some other artefact how += is compiled. But in any case `dotc` behaves like `doti`, so it's not a repl failure.

PR #1184, opened 2016-03-18
better positioning of `_*` not last argument error

liufengyun:
The error should be at the beginning of the underscore `_`,
not the next token after the star `*`, which is the default.

Before:

```
tests/neg/i1059.scala:5: error: `_*' can be used only for last argument
    case _ : _*                    =>   0
                                   ^
```

After:
```
tests/neg/i1059.scala:5: error: `_*' can be used only for last argument
    case _ : _*                    =>   0
             ^                       
```

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -227,6 +227,8 @@ object Parsers {
 
     /** Consume one token of the specified type, or
       * signal an error if it is not there.
+      *
+      * @return The offset at the start of the token to accept
       */
     def accept(token: Int): Int = {
       val offset = in.offset
@@ -234,7 +236,7 @@ object Parsers {
         syntaxErrorOrIncomplete(expectedMsg(token))
       }
       if (in.token == token) in.nextToken()
-      in.offset
+      offset
     }
 
     /** semi = nl {nl} | `;'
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -234,7 +234,7 @@ object Parsers {
         syntaxErrorOrIncomplete(expectedMsg(token))
       }
       if (in.token == token) in.nextToken()
-      in.offset
+      offset
```
liufengyun: This change is dubious, but it seems to be what's in the mind of the original author. Otherwise, there's no need to define the variable `offset`.

odersky: Yes, good catch. It should be `offset`. Can you add to the doc comment a clause which says "@return   The offset at the start of the token to accept"? Thanks!

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -88,7 +88,7 @@ object Parsers {
     def syntaxError(msg: String, offset: Int = in.offset): Unit =
       if (offset > lastErrorOffset) {
         syntaxError(msg, Position(offset))
-        lastErrorOffset = in.offset
+        lastErrorOffset = offset
```
odersky: No, I think here you need to leave in.offset. (also in JavaParsers). lastErrorOffset is the offset we skipped to (as its doc comment implies). That change explains the test failures we are seeing.

liufengyun: I've restored this change to make tests pass.

But it seems following line is also problematic, `DefDcl ::= DefSig `:' Type` may not have `=`.

https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/parsing/Parsers.scala#L1831

It seems `lastErrorOffset = in.offset` accidentally masked the error above. I'll investigate if there's indeed a problem here and make another PR if necessary.

liufengyun:
I've updated the commit addressing the concerns above. Now the tests pass.

odersky:
All LGTM now.

PR #1186, opened 2016-03-19
Improvements to cyclic checking, avoidance, named parameters

### LINKED ISSUE #1185: checkNonCyclic does not analyze prefixes of TypeRefs
We get a stack overflow when compiling this program, courtesy of Roland Kuhn:

    object Test {
      import language.higherKinds

      class NotUsed

      trait FO[+Out, +Mat] { self =>
        type Repr[+O] <: FO[O, Mat] {
          type Repr[+OO] = self.Repr[OO]
        }
        def map[T](f: Out => T): Repr[T] = ???
      }

      class Source[+O, +M] extends FO[O, M] {
        type Repr[+OO] <: Source[OO, M]
      }

      class Flow[-I, +O, +M] extends FO[O, M] {
        type Repr[+OO] <: Flow[I, OO, M]
      }

      implicit class x[O, M, F[o, m] <: FO[o, m]](val f: F[O, M]) extends AnyVal {
        def xx(i: Int): f.Repr[O] = f.map(identity)
      }

      type IntFlow[O, M] = Flow[Int, O, M]

      val s1 = new Source[Int, NotUsed].xx(12)
      val s2: Source[Int, NotUsed] = s1
      val f1 = x[Int, NotUsed, IntFlow](new Flow[Int, Int, NotUsed]).xx(12)
      val f2: Flow[Int, Int, NotUsed] = f1
    }

The problem is that the F-bound cycle goes through a hk type appliction (which takes the form of an #Apply node) and we do not traverse prefixes of TypeRefs. So the cycle is not detected and therefore fails to be mitigated with a LazyRef.

odersky:
Fixes #1185

### DIFF
--- a/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -1069,6 +1069,9 @@ object SymDenotations {
     /** The type parameters of a class symbol, Nil for all other symbols */
     def typeParams(implicit ctx: Context): List[TypeSymbol] = Nil
 
+    /** The named type parameters declared or inherited by this symbol */
+    def namedTypeParams(implicit ctx: Context): Set[TypeSymbol] = Set()
+
     /** The type This(cls), where cls is this class, NoPrefix for all other symbols */
     def thisType(implicit ctx: Context): Type = NoPrefix
 
@@ -1202,6 +1205,8 @@ object SymDenotations {
     /** TODO: Document why caches are supposedly safe to use */
     private[this] var myTypeParams: List[TypeSymbol] = _
 
+    private[this] var myNamedTypeParams: Set[TypeSymbol] = _
+
     /** The type parameters of this class */
     override final def typeParams(implicit ctx: Context): List[TypeSymbol] = {
       def computeTypeParams = {
@@ -1212,6 +1217,16 @@ object SymDenotations {
       }
       if (myTypeParams == null) myTypeParams = computeTypeParams
       myTypeParams
+    }
+
+    /** The named type parameters declared or inherited by this class */
+    override final def namedTypeParams(implicit ctx: Context): Set[TypeSymbol] = {
+      def computeNamedTypeParams: Set[TypeSymbol] =
+        if (ctx.erasedTypes || is(Module)) Set() // fast return for modules to avoid scanning package decls
+        else memberNames(abstractTypeNameFilter).map(name =>
+          info.member(name).symbol.asType).filter(_.is(TypeParam, butNot = ExpandedName)).toSet
+      if (myNamedTypeParams == null) myNamedTypeParams = computeNamedTypeParams
+      myNamedTypeParams
     }
 
     override protected[dotc] final def info_=(tp: Type) = {
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -1075,15 +1075,29 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     }
   }
 
-  /** `op(tp1, tp2)` unless `tp1` and `tp2` are type-constructors.
+  /** `op(tp1, tp2)` unless `tp1` and `tp2` are type-constructors with at least
+   *  some unnamed type parameters.
    *  In the latter case, combine `tp1` and `tp2` under a type lambda like this:
    *
    *    [X1, ..., Xn] -> op(tp1[X1, ..., Xn], tp2[X1, ..., Xn])
+   *
+   *  Note: There is a tension between named and positional parameters here, which
+   *  is impossible to resolve completely. Say you have
+   *
+   *    C[type T], D[type U]
+   *
+   *  Then do you expand `C & D` to `[T, U] -> C[T, U] & D[T, U]` or not? Under the named
+   *  type parameter interpretation, this would be wrong whereas under the traditional
+   *  higher-kinded interpretation this would be required. The problem arises from
+   *  allowing both interpretations. A possible remedy is to be somehow stricter
+   *  in where we allow which interpretation.
    */
   private def liftIfHK(tp1: Type, tp2: Type, op: (Type, Type) => Type) = {
     val tparams1 = tp1.typeParams
     val tparams2 = tp2.typeParams
-    if (tparams1.isEmpty || tparams2.isEmpty) op(tp1, tp2)
+    def onlyNamed(tparams: List[TypeSymbol]) = tparams.forall(!_.is(ExpandedName))
+    if (tparams1.isEmpty || tparams2.isEmpty ||
+        onlyNamed(tparams1) && onlyNamed(tparams2)) op(tp1, tp2)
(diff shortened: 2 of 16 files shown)


src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -156,16 +156,24 @@ object Checking {
         tp
     }
 
-    def apply(tp: Type) = tp match {
+    private def apply(tp: Type, cycleOK: Boolean, nestedCycleOK: Boolean): Type = {
```
smarter: Maybe call these parameters `overrideCycleOK` or `cycleOKOverride` to avoid confusion with the class fields

src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -182,19 +190,12 @@ object Checking {
             case _: RefinedType => true
             case _ => false
           }
-          // If prefix is interesting, check info of typeref recursively, marking the referred symbol
-          // with NoCompleter. This provokes a CyclicReference when the symbol
-          // is hit again. Without this precaution we could stackoverflow here.
```
smarter: How come all of this isn't necessary anymore?

tests/pos/flowops.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  implicit class x[O, M, F[o, m] <: FO[o, m]](val f: F[O, M]) extends AnyVal {
+    def xx(i: Int): f.Repr[O] = f.map(identity)
+  }
+
+  type IntFlow[O, M] = Flow[Int, O, M]
+
+  val s1 = new Source[Int, NotUsed].xx(12)
+  val s2: Source[Int, NotUsed] = s1
+  val f1 = x[Int, NotUsed, IntFlow](new Flow[Int, Int, NotUsed]).xx(12)
```
rkuhn: Is this explicit type ascription and conversion necessary?

src/dotty/tools/dotc/typer/Checking.scala
odersky: I picked the same names intentionally. I think there's also value in making clear they are the same.

src/dotty/tools/dotc/typer/Checking.scala
odersky: Not sure. We never run into stack overflows anymore. Maybe it's because we also do no longer the #Apply insertion that the comment still referred to.

tests/pos/flowops.scala
odersky: No, I just did it step by step. The next lines drop the conversion.

odersky:
This seems to be stuck in the validator. How can this be made unstuck?

smarter:
@SethTisue : any idea what we can do to unstuck the build?

smarter:
@odersky : you could always do `git commit --amend --no-edit` to change the hash of the last commit and then push -f

odersky:
@smarter OK, that has helped.

src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -1075,15 +1075,29 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
    *  In the latter case, combine `tp1` and `tp2` under a type lambda like this:
    *
    *    [X1, ..., Xn] -> op(tp1[X1, ..., Xn], tp2[X1, ..., Xn])
+   *
+   *  Note: There is a tension between named and positional parameters here, which
+   *  is impossible to resolve completely. Say you have
+   *
+   *    C[type T], D[type U]
+   *
+   *  Then do you expand `C & D` to `[T, U] -> C[T, U] & D[T, U]` or not? Under the named
```
smarter: Do you mean `[T, U] -> C[T] & D[U]` ?

smarter: Or actually what we would do for unnamed tparams is `[T] -> C[T] & D[T]` in such a case.

odersky:
Rebased to master

odersky:
Ready for review now. This has turned out to be somewhat a grab-bag of features but overall we got some good simplifications in addition to fixing type inference for named parameters.

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1214,6 +1219,16 @@ object SymDenotations {
       myTypeParams
     }
 
+    /** The named type parameters declared or inherited by this class */
+    override final def namedTypeParams(implicit ctx: Context): Set[TypeSymbol] = {
+      def computeNamedTypeParams: Set[TypeSymbol] =
+        if (ctx.erasedTypes || is(Module)) Set() // fast return for modules to avoid scanning package decls
+        else memberNames(abstractTypeNameFilter).map(name =>
```
smarter: Shouldn't we create a `NameFilter` instance that filters named type params directly instead?
```scala
  object namedTypeParamNameFilter extends NameFilter {
    def apply(pre: Type, name: Name)(implicit ctx: Context): Boolean =
      name.isTypeName && {
        val mbr = pre.member(name)
        mbr.symbol.is(Deferred) && mbr.symbol.is(TypeParam, butNot = ExpandedName) &&
        mbr.info.isInstanceOf[RealTypeBounds]
      }
  }
```

tests/pos/CollectionStrawMan3.scala
```diff
@@ -114,13 +114,13 @@ object CollectionStrawMan1 {
   /* --------- Concrete collection types ------------------------------- */
 
   /** Concrete collection type: List */
-  sealed trait List[+A] extends Seq[A] with FromIterator[List] {
+  sealed trait List[type +Elem] extends Seq[Elem] with FromIterator[List] {
```
smarter: Does that mean that `TypeComparer#compareAliasedRefined` is no longer needed?
It uses this as an example:
```
trait Lambda[type Elem]
trait Lst[T] extends Lambda[T]
```
and if I understand this commit correctly, this is no longer allowed

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -93,27 +94,18 @@ trait TypeAssigner {
             case _ =>
               mapOver(tp)
           }
+        case tp @ AppliedType(tycon, args) if toAvoid(tycon) =>
+          val base = apply(tycon)
+          val args = tp.baseArgInfos(base.typeSymbol)
+          if (base.typeParams.length == args.length) base.appliedTo(args) else base
```
smarter: `args` might contain something we want to avoid, this should be `if (base.typeParams.length == args.length) apply(base.appliedTo(args)) else base`.

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3213,6 +3232,55 @@ object Types {
     def apply(tp: Type) = tp
   }
 
+  abstract class ApproximatingTypeMap(implicit ctx: Context) extends TypeMap { thisMap =>
```
smarter: Interesting abstraction, do you think we could implement `TypeAssigner#avoid` using it?

PR #1187, opened 2016-03-21
Fix problem involving classtag based pattern matches.

### LINKED ISSUE #1174: Context bound type parameter in def with explicit return type does not compile
The following code does not compile:

```scala
import scala.reflect.ClassTag
import scala.util._

object Main {
  class A

  def constructAs[T <: A: ClassTag]: Try[T] = Try {
    new A()
  }.flatMap {
    case inst: T => Success(inst)
    case _ =>
      val tag = implicitly[ClassTag[T]]
      Failure(new ClassCastException(s"Failed to construct instance of class ${tag.runtimeClass.getName}"))
  }
}
```

Dotc gives the following error:

```
Main.scala:10: error: type mismatch:
 found   : Main.A(inst)
 required: Nothing'
    case inst: T => Success(inst)
                            ^
one error found
```

If I leave out the context bound, the code compiles fine. In Scala 2.11.8 it compiles too.

Weirdly enough, if I leave out the explicit return type, it also compiles fine.

odersky:
Rewriting did not go far enough, as evidenced by pos/i1174.scala

Fixes #1174. Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -842,7 +842,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     val proto2 = // the computed type of the `elemtpt` field
       if (!tree.elemtpt.isEmpty) WildcardType
       else if (isFullyDefined(proto1, ForceDegree.none)) proto1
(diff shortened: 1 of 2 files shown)


smarter:
LGTM

PR #1188, opened 2016-03-23
Fix #1167: Reduce the magic in Arrays.newRefArray. Implement multidimensional arrays

### LINKED ISSUE #1167: Removing the magic from dotty.runtime.Arrays
[`dotty.runtime.Arrays`](https://github.com/lampepfl/dotty/blob/master/src/dotty/runtime/Arrays.scala) has a bunch of magic methods, including the very magic `newRefArray`, which is one of the 3 methods that stay generic after erasure.

It is very easy to get rid of this complexity, by given an actual parameter of type `Class[T]` to `newRefArray`:
```scala
def newRefArray[T <: AnyRef](componentType: Class[T], length: Int): Array[T]
```
The back-end still needs to know about this magic method, and "assume" that `componentType` is a `Literal(Constant(tpe))` (i.e., crash if it isn't). **But then the method does not need to be generic anymore after erasure, which is a substantial win, IMO.**

We can remove even more magic by given *actual* bodies to all the `newXArray` method, leveraging `java.lang.reflect.Array.newInstance`:
```scala
def newRefArray[T <: AnyRef](componentType: Class[T], length: Int): Array[T] =
  java.lang.reflect.Array.newInstance(componentType, length).asInstanceOf[Array[T]]

def newByteArray(length: Int): Array[Byte] =
  java.lang.reflect.Array.newInstance(classOf[Byte], length).asInstanceOf[Array[Byte]]

...

def newUnitArray(length: Int): Array[Unit] =
  java.lang.reflect.Array.newInstance(classOf[BoxedUnit], length).asInstanceOf[Array[Unit]]
```
and now the back-end does not *need* to treat these methods specially. Their user-space implementation is perfectly correct.

Of course, on the JVM, that implementation is inefficient, so the JVM back-end would still specialize them. But other back-ends need not have the same problem. For example, Scala.js does not need to specialize those to get full speed, because `Array.newInstance` is intrisified by the optimizer when the `componentType` is a literal `classOf`.

sjrd:
Previously, the method `Arrays.newRefArray` was one of the only 3
methods that are kept generic after erasure. This commit removes
this magic, by making it take an actual `j.l.Class[T]` as
parameter.

Moreover, the methods `newXArray` all receive an actual body,
implemented on top of Java reflection, which means that a back-end
does not *have to* special-case those methods for correctness.
It might still be required for performance, though, depending on
the back-end.

The JVM back-end is made non-optimal in this commit, precisely
because it does not specialize that method anymore. Doing so
requires modifying the fork of scalac that we use, which should
be done separately.

The JS back-end is adapted simply by doing nothing at all on any
of the `newXArray` methods. It will normally call the user-space
implementations which use reflection. The Scala.js optimizer will
inline and intrinsify the reflective calls, producing optimal
code, at the end of the day.

### DIFF
--- a/src/dotty/runtime/Arrays.scala
+++ b/src/dotty/runtime/Arrays.scala
@@ -26,42 +26,6 @@ object Arrays {
 
   /** Create an array of a reference type T.
    */
-  def newRefArray[T](componentType: Class[T])(length: Int): Array[T] =
-    jlr.Array.newInstance(componentType, length).asInstanceOf[Array[T]]
-
-  /** Create a Byte[] array */
-  def newByteArray(length: Int): Array[Byte] =
-    jlr.Array.newInstance(classOf[Byte], length).asInstanceOf[Array[Byte]]
-
-  /** Create a Short[] array */
-  def newShortArray(length: Int): Array[Short] =
-    jlr.Array.newInstance(classOf[Short], length).asInstanceOf[Array[Short]]
-
-  /** Create a Char[] array */
-  def newCharArray(length: Int): Array[Char] =
-    jlr.Array.newInstance(classOf[Char], length).asInstanceOf[Array[Char]]
-
-  /** Create an Int[] array */
-  def newIntArray(length: Int): Array[Int] =
-    jlr.Array.newInstance(classOf[Int], length).asInstanceOf[Array[Int]]
-
-  /** Create a Long[] array */
-  def newLongArray(length: Int): Array[Long] =
-    jlr.Array.newInstance(classOf[Long], length).asInstanceOf[Array[Long]]
-
-  /** Create a Float[] array */
-  def newFloatArray(length: Int): Array[Float] =
-    jlr.Array.newInstance(classOf[Float], length).asInstanceOf[Array[Float]]
-
-  /** Create a Double[] array */
-  def newDoubleArray(length: Int): Array[Double] =
-    jlr.Array.newInstance(classOf[Double], length).asInstanceOf[Array[Double]]
-
-  /** Create a Boolean[] array */
-  def newBooleanArray(length: Int): Array[Boolean] =
-    jlr.Array.newInstance(classOf[Boolean], length).asInstanceOf[Array[Boolean]]
-
-  /** Create a scala.runtime.BoxedUnit[] array */
-  def newUnitArray(length: Int): Array[Unit] =
-    jlr.Array.newInstance(classOf[scala.runtime.BoxedUnit], length).asInstanceOf[Array[Unit]]
+  def newArray[Arr](componentType: Class[_], returnType: Class[Arr], dimensions: Array[Int]): Arr =
+    jlr.Array.newInstance(componentType, dimensions: _*).asInstanceOf[Arr]
 }
--- a/src/dotty/runtime/Arrays.scala
+++ b/src/dotty/runtime/Arrays.scala
@@ -2,6 +2,8 @@ package dotty.runtime
 
 import scala.reflect.ClassTag
 
+import java.lang.{reflect => jlr}
+
 /** All but the first two operations should be short-circuited and implemented specially by
  *  the backend.
  */
@@ -22,35 +24,44 @@ object Arrays {
     arr
   }
 
-  /** Create an array of type T. T must be of form Array[E], with
-   *  E being a reference type.
+  /** Create an array of a reference type T.
    */
-  def newRefArray[T](length: Int): T = ???
+  def newRefArray[T](componentType: Class[T])(length: Int): Array[T] =
+    jlr.Array.newInstance(componentType, length).asInstanceOf[Array[T]]
 
   /** Create a Byte[] array */
-  def newByteArray(length: Int): Array[Byte] = ???
+  def newByteArray(length: Int): Array[Byte] =
+    jlr.Array.newInstance(classOf[Byte], length).asInstanceOf[Array[Byte]]
 
   /** Create a Short[] array */
-  def newShortArray(length: Int): Array[Short] = ???
+  def newShortArray(length: Int): Array[Short] =
+    jlr.Array.newInstance(classOf[Short], length).asInstanceOf[Array[Short]]
 
   /** Create a Char[] array */
-  def newCharArray(length: Int): Array[Char] = ???
+  def newCharArray(length: Int): Array[Char] =
+    jlr.Array.newInstance(classOf[Char], length).asInstanceOf[Array[Char]]
 
   /** Create an Int[] array */
-  def newIntArray(length: Int): Array[Int] = ???
+  def newIntArray(length: Int): Array[Int] =
+    jlr.Array.newInstance(classOf[Int], length).asInstanceOf[Array[Int]]
 
   /** Create a Long[] array */
-  def newLongArray(length: Int): Array[Long] = ???
+  def newLongArray(length: Int): Array[Long] =
+    jlr.Array.newInstance(classOf[Long], length).asInstanceOf[Array[Long]]
 
   /** Create a Float[] array */
-  def newFloatArray(length: Int): Array[Float] = ???
+  def newFloatArray(length: Int): Array[Float] =
+    jlr.Array.newInstance(classOf[Float], length).asInstanceOf[Array[Float]]
 
   /** Create a Double[] array */
-  def newDoubleArray(length: Int): Array[Double] = ???
+  def newDoubleArray(length: Int): Array[Double] =
+    jlr.Array.newInstance(classOf[Double], length).asInstanceOf[Array[Double]]
 
   /** Create a Boolean[] array */
-  def newBooleanArray(length: Int): Array[Boolean] = ???
+  def newBooleanArray(length: Int): Array[Boolean] =
+    jlr.Array.newInstance(classOf[Boolean], length).asInstanceOf[Array[Boolean]]
 
   /** Create a scala.runtime.BoxedUnit[] array */
-  def newUnitArray(length: Int): Array[Unit] = ???
+  def newUnitArray(length: Int): Array[Unit] =
+    jlr.Array.newInstance(classOf[scala.runtime.BoxedUnit], length).asInstanceOf[Array[Unit]]
 }
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -371,9 +371,12 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       newArr("Generic").appliedToTypeTrees(typeArg :: Nil)
     else if (elemClass.isPrimitiveValueClass)
       newArr(elemClass.name.toString)
-    else
-      newArr("Ref").appliedToTypeTrees(
-        TypeTree(defn.ArrayOf(elemType)).withPos(typeArg.pos) :: Nil)
+    else {
+      val typeApplied = newArr("Ref").appliedToTypeTrees(typeArg :: Nil)
+      val classOfArg =
+        ref(defn.Predef_classOf).appliedToTypeTrees(typeArg :: Nil)
+      Apply(typeApplied, classOfArg :: Nil).withPos(typeArg.pos)
+    }
   }
 
   // ------ Creating typed equivalents of trees that exist only in untyped form -------
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -838,7 +838,10 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       case tpnme.Float => TYPE(defn.BoxedFloatModule)
       case tpnme.Double => TYPE(defn.BoxedDoubleModule)
       case tpnme.Unit => TYPE(defn.BoxedUnitModule)
-      case _ => Literal(Constant(TypeErasure.erasure(tp)))
+      case _ =>
+        if(ctx.erasedTypes || !tp.derivesFrom(defn.ArrayClass))
+          Literal(Constant(TypeErasure.erasure(tp)))
+        else Literal(Constant(tp))
     }
   }
 
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -362,21 +362,21 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
    *  kind for the given element type in `typeArg`. No type arguments or
    *  `length` arguments are given.
    */
-  def newArray(typeArg: Tree, pos: Position)(implicit ctx: Context): Tree = {
-    val elemType = typeArg.tpe
-    val elemClass = elemType.classSymbol
-    def newArr(kind: String) =
-      ref(defn.DottyArraysModule).select(s"new${kind}Array".toTermName).withPos(pos)
-    if (TypeErasure.isUnboundedGeneric(elemType))
-      newArr("Generic").appliedToTypeTrees(typeArg :: Nil)
-    else if (elemClass.isPrimitiveValueClass)
-      newArr(elemClass.name.toString)
-    else {
-      val typeApplied = newArr("Ref").appliedToTypeTrees(typeArg :: Nil)
-      val classOfArg =
-        ref(defn.Predef_classOf).appliedToTypeTrees(typeArg :: Nil)
-      Apply(typeApplied, classOfArg :: Nil).withPos(typeArg.pos)
-    }
+  def newArray(elemTpe: Type, returnTpe: Type, pos: Position, dims: JavaSeqLiteral)(implicit ctx: Context): Tree = {
+    val elemClass = elemTpe.classSymbol
+    def newArr(symbol: TermSymbol) =
+      ref(defn.DottyArraysModule).select(symbol).withPos(pos)
+
+    if (!ctx.erasedTypes)
+      if (TypeErasure.isUnboundedGeneric(elemTpe)) {
+        //exists only before erasure
+        assert(dims.elems.tail.isEmpty)
+        newArr(defn.newGenericArrayMethod).appliedToType(elemTpe).appliedTo(dims.elems.head)
+      }
+      else
+        newArr(defn.newArrayMethod).appliedToTypeTrees(TypeTree(returnTpe) :: Nil).appliedToArgs(clsOf(elemTpe) :: clsOf(returnTpe) :: dims :: Nil).withPos(pos)
+    else  // after erasure
+      newArr(defn.newArrayMethod).appliedToArgs(clsOf(elemTpe) :: clsOf(returnTpe) :: dims :: Nil).withPos(pos)
   }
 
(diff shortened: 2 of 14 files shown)


retronym:
While you are in this area, it might be good to consider whether and how to extend array support to use `multinewarray` for `Array.ofDim`.

```java
public class Test {
	public static void main(String... args) {
		int[][] multi = new int[5][10];
	}
}
```
```
  public static void main(java.lang.String...);
    Code:
       0: iconst_5
       1: bipush        10
       3: multianewarray #2,  2             // class "[[I"
       7: astore_1
       8: return
```

vs

```
  public <T> java.lang.Object[] ofDim(int, int, scala.reflect.ClassTag<T>);
    Code:
       0: getstatic     #319                // Field scala/reflect/ClassTag$.MODULE$:Lscala/reflect/ClassTag$;
       3: getstatic     #101                // Field scala/runtime/ScalaRunTime$.MODULE$:Lscala/runtime/ScalaRunTime$;
       6: aload_3
       7: invokeinterface #322,  1          // InterfaceMethod scala/reflect/ClassTag.runtimeClass:()Ljava/lang/Class;
      12: invokevirtual #326                // Method scala/runtime/ScalaRunTime$.arrayClass:(Ljava/lang/Class;)Ljava/lang/Class;
      15: invokevirtual #329                // Method scala/reflect/ClassTag$.apply:(Ljava/lang/Class;)Lscala/reflect/ClassTag;
      18: iload_1
      19: invokeinterface #152,  2          // InterfaceMethod scala/reflect/ClassTag.newArray:(I)Ljava/lang/Object;
      24: checkcast     #330                // class "[Ljava/lang/Object;"
      27: astore        4
      29: getstatic     #335                // Field scala/runtime/RichInt$.MODULE$:Lscala/runtime/RichInt$;
      32: getstatic     #340                // Field scala/Predef$.MODULE$:Lscala/Predef$;
      35: iconst_0
      36: invokevirtual #344                // Method scala/Predef$.intWrapper:(I)I
      39: iload_1
      40: invokevirtual #348                // Method scala/runtime/RichInt$.until$extension0:(II)Lscala/collection/immutable/Range;
      43: iload_2
      44: aload_3
      45: aload         4
      47: invokedynamic #356,  0            // InvokeDynamic #10:apply$mcVI$sp:(ILscala/reflect/ClassTag;[Ljava/lang/Object;)Lscala/runtime/java8/JFunction1$mcVI$sp;
      52: invokevirtual #361                // Method scala/collection/immutable/Range.foreach$mVc$sp:(Lscala/Function1;)V
      55: aload         4
      57: areturn
```

sjrd:
@retronym IIUC, this requires non-trivial changes in the back-end. IIRC, @DarkDimius wants to keep the back-end as close as possible to the scalac one.

sjrd:
Review by @DarkDimius ?
Note: since I have no control over the scalac fork with the shared back-end, I cannot implement the last bit to recover efficiency, which is to specialize `newRefArray(classOf[T], len)` in the back-end.

DarkDimius:
I'm very hesitant to accept the original change at the moment.
It would require either very ugly hacks in DottyBackendInterace(creating ill-typed trees on the fly) or modifying shared backed for more dotty specific stuff. 

I have a feeling that we're getting a very small win: there are still generic methods after erasure. It does not give any win for any phases after erasure, except JS backend. But instead it would complicate shared backed a lot.

@retronym @sjrd GenBCode does support emitting `multianewarray` in both scalac and Dotty: https://github.com/DarkDimius/scala/blob/sharing-backend/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala#L633 . Though AFAIK this is dead code simply because neither dotty nor scalac has a tree that would reach this `if` (it expects a `new*Array` with more that 1 argument).

DarkDimius:
More precisely, it allows to remove 15 quite simple lines lines from JS backend while adding 70+ lines of hard-to-maintain-between-frontends code to shared backend.

src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -372,9 +372,12 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       newArr("Generic").appliedToTypeTrees(typeArg :: Nil)
     else if (elemClass.isPrimitiveValueClass)
       newArr(elemClass.name.toString)
-    else
-      newArr("Ref").appliedToTypeTrees(
-        TypeTree(defn.ArrayOf(elemType)).withPos(typeArg.pos) :: Nil)
+    else {
+      val typeApplied = newArr("Ref").appliedToTypeTrees(typeArg :: Nil)
+      val classOfArg =
+        ref(defn.Predef_classOf).appliedToTypeTrees(typeArg :: Nil)
```
DarkDimius: This code would only work before `classOf` phase, and would stop working even after patmat.

It should instead use `tpd.clsOf(typeArg)`

DarkDimius:
On the other side: this PR makes
`newRefArray` very special amoung  `newXXXArray` methods: it has different signature and needs to be handled specially.
I we decide to get this in, I would argue that we shouldn't have 2 kinds of array-construction methods.
We remove other, only keeping  `newRefArray` as now it can create primitive arrays as well as ref-arrays.

sjrd:
> 70+ lines of hard-to-maintain-between-frontends code to shared backend.

Hum ... I'm expecting this to be around 3 lines:
```scala
        case Apply(_, Constant(Literal(elemTpe: Type)) :: args) if isSyntheticArrayConstructor(app.symbol) =>
          generatedType = ArrayBType(toTypeKind(elemTpe))
          mkArrayConstructorCall(generatedType, app, args)
```
right before the line you were pointing at in your comment.

> it allows to remove 15 quite simple lines lines from JS backend

I don't care about those. I care about https://github.com/lampepfl/dotty/pull/1188/files#diff-7d76351125e3a0376bea4b9db34a05ccL471

src/dotty/tools/dotc/ast/tpd.scala
sjrd: `tpd.clsOf(typeArg)` does not work. I tried that first. It choked in pickling because the pickler cannot pickle a `JavaArrayType`.

sjrd:
> We remove other, only keeping newRefArray as now it can create primitive arrays as well as ref-arrays.

Makes sense.

src/dotty/tools/dotc/ast/tpd.scala
DarkDimius: well this code also does not work for 80% of pipeline(it broke my optimization phases)
This code would not pass Ycheck after erasure either.

sjrd: So ... `tpd.clsOf(typeArg)` is unreliable before pickling, and `classOf[typeArg]` is unreliable after `ClassOf`. Is there anything that works all the time to generate a literal `classOf[T]`? Should I fix `clsOf` so that it *is* reliable along the entire pipeline?

DarkDimius:
`Apply(_, Constant(Literal(elemTpe: Type)) :: args)`

would not work without extending SharedBackendInterface.
Currently you cannot test if something is a `Type`.  Type is just an abstract type alias. 
`elemTpe: Type` would be erazed to `elemTpe: Object` and would always succeed.

src/dotty/tools/dotc/ast/tpd.scala
DarkDimius: > Should I fix clsOf so that it is reliable along the entire pipeline?

I guess it would be the way to go. It should check if the type it is given derives from `scala.Array`, if it does, only don't yet erase its type before erasure.

sjrd:
OK, then it can be done as [in `genConstant`](https://github.com/DarkDimius/scala/blob/sharing-backend/src/compiler/scala/tools/nsc/backend/jvm/BCodeBodyBuilder.scala#L505):
```scala
        case Apply(_, Constant(const) :: args)
            if isSyntheticArrayConstructor(app.symbol) && const.tag == ClazzTag =>
          generatedType = ArrayBType(toTypeKind(const.typeValue))
          mkArrayConstructorCall(generatedType, app, args)
```

DarkDimius:
@sjrd, let me try to play with this branch before Dotty meeting and let's discuss there.
I would try keeping only 1 `newArray` method and seeing how would it be handled in backed between scalac and Dotty.

DarkDimius:
@retronym @sjrd GenBCode does support emitting multianewarray in both scalac and Dotty

https://github.com/scala/scala/pull/5060/files indicates that though code existed, it was never executed or tested.

DarkDimius:
ping @sjrd please have a look at changes that I'm proposing.
@odersky please review 2828ef1 and d5d92e7

DarkDimius:
@retronym thx for suggestion. Implemented: https://github.com/lampepfl/dotty/pull/1188/commits/b2215ed23311b2c99ea638f9d7fcad9737dba588

DarkDimius:
Waiting for scala-compiler fork to reach maven central.
Here is the link for those who need to update version by hand in the IDE: https://oss.sonatype.org/content/repositories/releases/me/d-d/scala-compiler/2.11.5-20160322-171045-e19b30b3cd/

Backwards INcompatible with previous version.

DarkDimius:
> sbt.ResolveException: unresolved dependency: org.scala-sbt#launcher-interface;1.0.0-M1: not found

/rebuild

src/dotty/runtime/Arrays.scala
```diff
@@ -26,42 +26,6 @@ object Arrays {
...
-
-  /** Create a Boolean[] array */
-  def newBooleanArray(length: Int): Array[Boolean] =
-    jlr.Array.newInstance(classOf[Boolean], length).asInstanceOf[Array[Boolean]]
-
-  /** Create a scala.runtime.BoxedUnit[] array */
-  def newUnitArray(length: Int): Array[Unit] =
-    jlr.Array.newInstance(classOf[scala.runtime.BoxedUnit], length).asInstanceOf[Array[Unit]]
+  def newArray[Arr](componentType: Class[_], returnType: Class[Arr], dimensions: Array[Int]): Arr =
+    jlr.Array.newInstance(componentType, dimensions: _*).asInstanceOf[Arr]
```
sjrd: Why do you need `returnType`, here?

Also, why no bound? Right now `componentType` is completely unchecked; it might not be in sync. What about the following signature?
```scala
  def newArray[T](componentType: Class[T], returnType: Class[Arrary[T]], dimensions: Array[Int]): Array[T] =
    jlr.Array.newInstance(componentType, dimensions: _*).asInstanceOf[Arrary[T]]
```

src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -363,21 +363,21 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
-      newArr("Generic").appliedToTypeTrees(typeArg :: Nil)
-    else if (elemClass.isPrimitiveValueClass)
-      newArr(elemClass.name.toString)
-    else {
-      val typeApplied = newArr("Ref").appliedToTypeTrees(typeArg :: Nil)
-      val classOfArg =
-        ref(defn.Predef_classOf).appliedToTypeTrees(typeArg :: Nil)
-      Apply(typeApplied, classOfArg :: Nil).withPos(typeArg.pos)
-    }
+  def newArray(elemTpe: Type, returnTpe: Type, pos: Position, dims: JavaSeqLiteral)(implicit ctx: Context): Tree = {
```
sjrd: Hum, the signature of this method changes, but I see no call site change in this commit. That's suspicious ;)

src/dotty/tools/dotc/transform/ArrayConstructors.scala
```diff
@@ -0,0 +1,59 @@
...
+        case TypeApply(tycon, targ :: Nil)  =>
+          rewrite(targ.tpe, tree.args)
+        case _ =>
+          ???
+      }
+    } else if ((tree.fun.symbol.maybeOwner eq defn.ArrayModule) && (tree.fun.symbol.name eq nme.ofDim) && !tree.tpe.isInstanceOf[MethodicType]) {
+
+      tree.fun match {
+        case Apply(TypeApply(t: Ident, targ), dims) =>
+          rewrite(targ.head.tpe, dims)
```
sjrd: What happens to the implicit `ClassTag` argument to `ofDim`? Isn't part of `dims`? In that case, `rewrite` would do something bad with it wouldn't it?

sjrd:
That's all I have at this point.

src/dotty/runtime/Arrays.scala
DarkDimius: > Why do you need returnType, here?

I need to recreate the returned type at backend. And this is the easiest way I've found to do this.

>Also, why no bound? Right now componentType is completely unchecked; it might not be in sync. What about the following signature?

It's NOT in sync before erasure. When you're creating a `Array[Int]`, where `Int` is a primitive, but you have to pass https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#TYPE, that is the `Class` for primitive Int, but is statically typed as `Class[java.lang.Integer]`

src/dotty/tools/dotc/transform/ArrayConstructors.scala
DarkDimius: implicit ClassTag argument to ofDim is the argument of `tree: Apply`, that has a form
`Apply(Apply(TypeApply(_, targs), dims, classTag)`. 
I'm deconstructing the inner Apply here.

src/dotty/runtime/Arrays.scala
sjrd: > that is the Class for primitive Int, but is statically typed as Class[java.lang.Integer].

Ouch! Why do we need this? Why can't we keep a `Literal(Constant(defn.IntType))`? That's how scalac does it, and I find it much more consistent.

src/dotty/tools/dotc/transform/ArrayConstructors.scala
sjrd: Ah, right. OK.

DarkDimius:
@odersky @sjrd tests passed. Ready for review.

src/dotty/runtime/Arrays.scala
```diff
@@ -22,35 +24,8 @@ object Arrays {
     arr
   }
 
-  /** Create an array of type T. T must be of form Array[E], with
-   *  E being a reference type.
+  /** Create an array of a reference type T.
```
sjrd: This comment is obsolete.

src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -363,18 +363,16 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
+    val elemClass = elemTpe.classSymbol
+    def newArr(symbol: TermSymbol) =
+      ref(defn.DottyArraysModule).select(symbol).withPos(pos)
+
+    if (!ctx.erasedTypes) {
+      assert(!TypeErasure.isUnboundedGeneric(elemTpe)) //needs to be done during typer. See Applications.convertNewGenericArray
+      newArr(defn.newArrayMethod).appliedToTypeTrees(TypeTree(returnTpe) :: Nil).appliedToArgs(clsOf(elemTpe) :: clsOf(returnTpe) :: dims :: Nil).withPos(pos)
+    } else  // after erasure
+      newArr(defn.newArrayMethod).appliedToArgs(clsOf(elemTpe) :: clsOf(returnTpe) :: dims :: Nil).withPos(pos)
```
sjrd: Both calls to `newArr` have the same argument `defn.newArrayMethod`. So `newArr` does need that argument and can directly use `defn.newArrayMethod` instead of `symbol`.

src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -751,7 +749,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
 
     /** If inititializer tree is `_', the default value of its type,
      *  otherwise the tree itself.
-     */
+     * */
```
sjrd: Spurious change.

sjrd:
That's all I have.

src/dotty/runtime/Arrays.scala
```diff
@@ -22,35 +24,8 @@ object Arrays {
...
-
-  /** Create a Double[] array */
-  def newDoubleArray(length: Int): Array[Double] = ???
-
-  /** Create a Boolean[] array */
-  def newBooleanArray(length: Int): Array[Boolean] = ???
-
-  /** Create a scala.runtime.BoxedUnit[] array */
-  def newUnitArray(length: Int): Array[Unit] = ???
+  def newArray[Arr](componentType: Class[_], returnType: Class[Arr], dimensions: Array[Int]): Arr =
```
odersky: Are we sure this does not lose performance? I am thinking of applications that create (say) billions of 3-element float vectors.

src/dotty/tools/dotc/transform/ArrayConstructors.scala
```diff
@@ -0,0 +1,58 @@
...
+  override def transformApply(tree: tpd.Apply)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    def rewrite(elemType: Type, dims: List[Tree]) =
+      tpd.newArray(elemType, tree.tpe, tree.pos, JavaSeqLiteral(dims, TypeTree(defn.IntClass.typeRef)))
+
+    if (tree.fun.symbol eq defn.ArrayConstructor) {
+      tree.fun match {
+        case TypeApply(tycon, targ :: Nil)  =>
+          rewrite(targ.tpe, tree.args)
+        case _ =>
+          ???
```
odersky: ??? left in.

If this case can never happen, use a pattern val instead.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -632,11 +632,20 @@ trait Applications extends Compatibility { self: Typer =>
...
+   *  It is performed during typer as creation of generic arrays needs a classTag.
+   *  we rely on implicit search to find one.
+   */
+  def convertNewGenericArray(tree: tpd.Tree)(implicit ctx: Context): tpd.Tree = tree match {
+    case Apply(TypeApply(tycon, targs@(targ :: Nil)), args) if tycon.symbol == defn.ArrayConstructor =>
       fullyDefinedType(tree.tpe, "array", tree.pos)
-      tpd.cpy.Apply(tree)(newArray(targ, tree.pos), args)
+
+      def newGenericArrayCall =
+        ref(defn.DottyArraysModule).select(defn.newGenericArrayMethod).withPos(tree.pos).appliedToTypeTrees(targs).appliedToArgs(args)
```
odersky: This looks more readable on multiple lines.

src/dotty/runtime/Arrays.scala
DarkDimius: That's the intention!
If `dimension.size ==1` backend will create a normal array
If `dimension.size > 1` backend will issue `multianewarray` instruction.

See example here: https://github.com/dotty-staging/dotty/blob/remove-newarray-magic/tests/run/MultiArr.scala

odersky:
What's the status of this PR?

DarkDimius:
I assumed it was merged long time ago. Will rebase&merge tomorrow.

PR #1189, opened 2016-03-24
TASTY tests for src/dotty

VladimirNik:
TASTY tests for src/dotty are added.

In comments to tests there are files for which we have different representation before pickling and after unpickling.

Review by @odersky

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -50,8 +50,11 @@ class tests extends CompilerTest {
   val sourceDir = "./src/"
   val dottyDir  = sourceDir + "dotty/"
   val toolsDir  = dottyDir + "tools/"
+  val backendDir = toolsDir + "backend/"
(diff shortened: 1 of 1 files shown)


odersky:
LGTM. Now we just have to find out whether the failures are real or artificial. Note that it's not guaranteed that trees with types print the same before and after pickling. That would be too restrictive and low-level. Already now we paper over differences with the "homogenize" method in PlainPrinter. So we should classify any remaining failure into one of

 1) Real bug, needs to be fixed.
 2) Semantically the same, can be addressed by refining "homogenize"
 3) Semantically the same, impractical to make print out the same, ignore the test.

DarkDimius:
@VladimirNik weren't  those tests supposed to fail?

VladimirNik:
@DarkDimius These tests are correct. On Monday I'll publish tests for files that are excluded from pickling in this pull request - I found simple examples for them.

PR #1197, opened 2016-03-31
Temporarily enable printing of original failure signature.

odersky:
Temporarily enable printing of original failure signature to debug failing builds.


DarkDimius:
LGTM.

PR #1199, opened 2016-04-03
Make erased value type take a TypeRef instead of a ClassSymbol

### LINKED ISSUE #1198: Stale symbol in ErasedValueType
ErasedValueType refers to a class symbol. This symbol can become stale if the ErasedValueType survives as part of a signature into the next run, but the symbol is redefined. To reproduce:

    cd src/dotty/tools/dotc
    scala -Xms1g -Xmx3g dotty.tools.dotc.Bench #runs 2 -d ~/tmp/classes *.scala */*.scala

This will give a stale symbol error on class Position in run 2.

odersky:
To address #1198, we need to avoid putting symbols in ErasedValueTypes.
Fixes #1198.

Review by @VladimirNik.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -69,17 +69,17 @@ object TypeErasure {
    *  @param   cls               The value class symbol
    *  @param   erasedUnderlying  The erased type of the single field of the value class
    */
-  abstract case class ErasedValueType(cls: ClassSymbol, erasedUnderlying: Type)
+  abstract case class ErasedValueType(tycon: TypeRef, erasedUnderlying: Type)
   extends CachedGroundType with ValueType {
-    override def computeHash = doHash(cls, erasedUnderlying)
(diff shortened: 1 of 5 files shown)


smarter:
Wait, how come an `ErasedValueType` ends up existing in multiple runs? They should only exist between `Erasure` and `ElimErasedValueType`

src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -69,17 +69,17 @@ object TypeErasure {
    *  @param   cls               The value class symbol
```
smarter: param doc needs to be updated

odersky:
@smarter 

> Wait, how come an ErasedValueType ends up existing in multiple runs? They should only exist between Erasure and ElimErasedValueType

I believe it's as a result of an InfoTransformer.

smarter:
Should we compile more tests like the `dotty` test with `twice` to catch this kind of bug?

VladimirNik:
LGTM

PR #1201, opened 2016-04-04
Check Java classes are not used as objects

### LINKED ISSUE #972: Ycheck that java classes are not used as objects
This passes Ycheck, bu it should not:

```
import java.lang.Math

val myMath: Math.type = Math
myMath.sqrt(3.0)
```

liufengyun:
Following code should not pass Ycheck:

    import java.lang.Math

    val myMath: Math.type = Math
    myMath.sqrt(3.0)
    id(myMath)

Fix #972. Review @DarkDimius .

### DIFF
--- a/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -181,6 +181,80 @@ class TreeChecker extends Phase with SymTransformer {
       if (tree.symbol.maybeOwner.isTerm)
         assert(nowDefinedSyms contains tree.symbol, i"undefined symbol ${tree.symbol}")
 
+    /** assert Java classes are not used as objects */
+    def assertIdentNotJavaClass(tree: Tree)(implicit ctx: Context): Unit = tree match {
+      case _ : untpd.Ident =>
+        assert(!tree.symbol.is(JavaModule), "Java class can't be used as value: " + tree)
+      case _ =>
+    }
+
+    /** check Java classes are not used as objects */
+    def checkIdentNotJavaClass(tree: Tree)(implicit ctx: Context): Unit = tree match {
+      // case tree: untpd.Ident =>
+      // case tree: untpd.Select =>
+      // case tree: untpd.SelectFromTypeTree =>
+      // case tree: untpd.Bind =>
+      case vd : ValDef =>
+        assertIdentNotJavaClass(vd.forceIfLazy)
+      case dd : DefDef =>
+        assertIdentNotJavaClass(dd.forceIfLazy)
+      // case tree: untpd.TypeDef =>
+      case Apply(fun, args) =>
+        assertIdentNotJavaClass(fun)
+        args.foreach(assertIdentNotJavaClass _)
+      // case tree: untpd.This =>
+      // case tree: untpd.Literal =>
+      // case tree: untpd.New =>
+      case Pair(left, right) =>
+        assertIdentNotJavaClass(left)
+        assertIdentNotJavaClass(right)
+      case Typed(expr, _) =>
+        assertIdentNotJavaClass(expr)
+      case NamedArg(_, arg) =>
+        assertIdentNotJavaClass(arg)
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -181,6 +181,106 @@ class TreeChecker extends Phase with SymTransformer {
...
+      case Match(selector, cases) =>
+        assertIdentNotJavaClass(selector)
+        cases.foreach(caseDef => assertIdentNotJavaClass(caseDef.body))
+      case Return(expr, _) =>
+        assertIdentNotJavaClass(expr)
+      case Try(expr, cases, finalizer) =>
+        assertIdentNotJavaClass(expr)
+        cases.foreach(caseDef => assertIdentNotJavaClass(caseDef.body))
+        assertIdentNotJavaClass(finalizer)
+      case untpd.Throw(expr) =>
```
DarkDimius: No need to handle untyped trees here.
They would never reach Ycheck.

src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -181,6 +181,106 @@ class TreeChecker extends Phase with SymTransformer {
...
 
+    /** assert Java classes are not used as objects */
+    def assertIdentNotJavaClass(tree: untpd.Tree)(implicit ctx: Context): Unit = tree match {
+      case _ : untpd.Ident =>
+        assert(!tree.symbol.is(JavaModule), "Java class can't be used as value: " + tree)
+      case _ =>
+    }
+
+    /** check Java classes are not used as objects */
+    def checkIdentNotJavaClass(tree: untpd.Tree)(implicit ctx: Context): Unit = tree match {
```
DarkDimius: Ycheck checks typed trees. No need to support all the `untpd` stuff.

src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -181,6 +181,106 @@ class TreeChecker extends Phase with SymTransformer {
...
+      // case tree: untpd.Ident =>
+      // case tree: untpd.Select =>
+      // case tree: untpd.SelectFromTypeTree =>
+      // case tree: untpd.Bind =>
+      case vd : ValDef =>
```
liufengyun: I keep the unchecked cases as comments to make it explicit what cases are ignored, in order to facilitate code maintenance. I'm not sure if it's a good style.

src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -181,6 +181,106 @@ class TreeChecker extends Phase with SymTransformer {
...
+        assertIdentNotJavaClass(body)
+      // case untpd.GenFrom(_, expr) =>
+      case untpd.GenAlias(_, expr) =>
+        assertIdentNotJavaClass(expr)
+      // case untpd.ContextBounds(_, _) =>
+      case _ =>
+    }
+
+    override def typed(tree: untpd.Tree, pt: Type = WildcardType)(implicit ctx: Context): tpd.Tree = {
+      checkIdentNotJavaClass(tree)
```
DarkDimius: I would go with

```
checkIdentNotJavaClass(super.typed(tree))
```

As this pattern is commonly used in typer.

liufengyun: Thanks, I'll follow the approach.

BTW, it's safe to assume all untyped ASTs are desugared after `FrontEnd`?

DarkDimius: > BTW, it's safe to assume all untyped ASTs are desugared after FrontEnd?

yes.

DarkDimius:
LGTM

liufengyun:
Thanks @DarkDimius .

When you have time, could you please add some description to the benchmark infrastructure task #1210 ? That would help a lot for me to get started. Thanks a lot.

PR #1202, opened 2016-04-04
[do not merge] Issues with TASTY - dotty pickling/unpickling

VladimirNik:
After checking (pickling/unpickling) src/dotty:
- There are 10 files that have different trees representations after unpickling (string representation).
- For each case I added sample file representing issue in addition to original source file from Dotty (files Fail1.scala - Fail10.scala)
- I checked each sample file with FromTasty: **6 files** can be compiled and checked with -Ycheck:all, **4 files** have issues during compilation/checking.

#### **Compilation issues**:

**1)** Issue with private field access (`val Unknown` in `object Test`) from `class Test` (Fail3.scala):
```scala
class Test() {
  import Test._
  val myStatus = Unknown
}
object Test {
  private val Unknown: Int = 0
}
```
Unpickling with FromTasty results in:
```
assertion failed: private getter Unknown in object Test$ in null 
accessed from constructor Test in class Test in Simple(Fail3.scala)
```
Similar code is in: dotty/tools/dotc/core/Types.scala
<br>
**2)** Issue with var parameter in class (Fail7.scala): 
```scala
class Fail7(var in: Int)
```
Unpickling with FromTasty results in:
```
error: class Test needs to be abstract, since 
<accessor> var in_=: (x$1: Int)Unit is not defined
(Note that an abstract var requires a setter in addition to the getter)
```

<br>
#### **-Ycheck:all issues**

**3)** Code example (Fail1.scala):
```scala
class Test[T] {
  def testMethod: Unit =
    new Foo(this)
}
class Foo[T] {
  def this(ct: Test[T]) = this()
}
```
Differences (in `def testMethod`). Such expression before pickling:
```
<<new Foo:Foo>:([T](ct: Test[T])Foo[T])>[T]
```
after unpickling becomes:
```
<<new Foo:Foo>:([T = Foo @UnsafeNonvariant#T](ct: Test[Foo @UnsafeNonvariant#T])Foo[T])>[T]
```
Unpickling with FromTasty using `-Ycheck:all` results in:
```
*** error while checking Simple(Fail1.scala) after phase frontend ***
java.lang.AssertionError: assertion failed: error at Simple(Fail1.scala):2
type mismatch:
 found   : Test[T(in Test)](Test.this)
 required: Test[Foo @UnsafeNonvariant#T(in Foo)]
tree = This(Test)
```
Similar code is in: dotty/runtime/vc/VCPrototype.scala
<br>
**4)** Code example (Fail5.scala):
```scala
class Fail5 {
  val someClass: Class[_] = ???
  val resultMethod = someClass.getMethod("result")
}
```
Differences in types (in `val resultMethod`). Such expression before pickling:
```
    val resultMethod: Method = 
        this:Fail5.someClass.getMethod:((x$0: String, x$1: Class[_]*)Method)>(
            <"result":String>,
            <<[ : java.lang.Class[_]]:scala.Array[java.lang.Class[_]]>: ...
```
after unpickling becomes:
```
    val resultMethod: Method = 
        this:Fail5.someClass.getMethod:((x$0: String, x$1: Class[_]*).Method)>(
            <"result":String("result")>,
            <<[ : java.lang.Class[_]]:scala.collection.Seq[java.lang.Class[_]]>: ...
```

The difference is that the type for x$1 arg changes from **scala.Array** to **scala.collection.Seq**

Unpickling with FromTasty using `-Ycheck:all` results in:
```
*** error while checking Simple(Fail5.scala) after phase frontend ***
java.lang.AssertionError: assertion failed: error at Simple(Fail5.scala):2
type mismatch:
 found   : scala.collection.Seq[Class[_]]
 required: Array[Class[_]]
```
Similar code is in: dotty/tools/dotc/repl/CompilingInterpreter.scala

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -47,6 +47,8 @@ class tests extends CompilerTest {
   val runDir        = testsDir + "run/"
   val newDir        = testsDir + "new/"
   val replDir       = testsDir + "repl/"
+  val tastyDir      = testsDir + "tasty/"
+  val tastyFailedDir = tastyDir + "failed/"
 
   val sourceDir = "./src/"
   val dottyDir  = sourceDir + "dotty/"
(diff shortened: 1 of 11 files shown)


DarkDimius:
In 4) Seems that `JavaSeqLiteral` was unpicked as normal `SeqLiteral`.

odersky:
@VladimirNik Do we need to keep this still around?

VladimirNik:
@odersky No, I think we can close it.

PR #1203, opened 2016-04-04
Two fixes to TermRefWithSignature.newLikeThis

DarkDimius:
Used to compute `asMemberOf(prefix)` twice and assume that symbol does not change.

@odersky please review


src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1743,7 +1743,7 @@ object Types {
       if (symbol.exists && !candidate.symbol.exists) { // recompute from previous symbol
         val ownSym = symbol
         val newd = asMemberOf(prefix)
-        candidate.withDenot(newd.suchThat(_ eq ownSym))
```
odersky: Yes, that what made no sense at all. Good catch.

odersky:
LGTM

PR #1204, opened 2016-04-05
simple integration with jline

### LINKED ISSUE #1084: Add jline support to REPL
We should integrate the REPL with jline to allow line editing.

liufengyun:
Fix #1084. Please review @odersky .

In this PR, I only added the basic functionality. Are there any specific features that are needed?

### DIFF
--- a/bin/dotc
+++ b/bin/dotc
@@ -38,7 +38,7 @@ unset verbose quiet cygwin toolcp colors saved_stty CDPATH
 
 CompilerMain=dotty.tools.dotc.Main
 FromTasty=dotty.tools.dotc.FromTasty
-ReplMain=test.DottyRepl
+ReplMain=dotty.tools.dotc.repl.Main
 
 
 
--- a/src/dotty/tools/dotc/repl/InteractiveReader.scala
+++ b/src/dotty/tools/dotc/repl/InteractiveReader.scala
@@ -8,24 +8,19 @@ trait InteractiveReader {
   val interactive: Boolean
 }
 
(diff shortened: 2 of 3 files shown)


odersky:
LGTM. For future commits, can you push to staging and make the pull request from there? We try to follow that schema to make it easier for several people to collaborate on a PR. Thanks!

odersky:
Oen suggestion: Can we make it accept more than one version of jline? AFAIK 2.12 works fine, so do we need to demand that people upgrade?

liufengyun:
OK, I'll ask @DarkDimius to add me to the staging repo.

Regarding the version, I'm not sure if `ivy` will bind to `2.12` when `2.14` is available. The syntax `[2.12,)` seems to achieve what you want, but it's a bit weird in the Java world.

I agree let's restore it to be `2.12` for the moment.

liufengyun:
Thanks @odersky .

som-snytt:
I have a repl branch which incorporates scala repl bits at https://github.com/som-snytt/dotty/blob/topic/dopler/src/dotty/tools/dotc/repl/jline/Reader.scala

A goal is to have well-defined extension points in the repl package, and a supplied repl with fewer bells and whistles. That was for purpose of adding features, but I'll also check for jline-neutrality.

The day job is doing its annual release so I haven't had the time I'd like, but I'll try to contribute it soon.

Should that be a normal PR against dotty repo?

smarter:
> Should that be a normal PR against dotty repo?

Sure, why not?

PR #1211, opened 2016-04-07
Fix Tasty errors

odersky:
Fixes one of the errors reported in #1202. If I can fix more errors, I'll try to push those fixes on this PR later. Review by @VladimirNik .


odersky:
Now the other errors reported in #1202 should also be fixed.

VladimirNik:
@odersky Thank you! LGTM

src/dotty/tools/dotc/FromTasty.scala
```diff
@@ -64,6 +64,9 @@ object FromTasty extends Driver {
   }
 
   class ReadTastyTreesFromClasses extends FrontEnd {
+
+    override def isTyper = false
```
DarkDimius: Unfortunately, this is not enough. 
A lot of code checks for `Phases.isAfterTyper` that is entirely independent from this method

src/dotty/tools/dotc/transform/Pickler.scala
```diff
@@ -68,7 +68,7 @@ class Pickler extends Phase {
   private def pickleSourcefile(pickler: TastyPickler, source: SourceFile): Unit = {
     val buf = new TastyBuffer(10)
     pickler.newSection("Sourcefile", buf)
-    buf.writeNat(pickler.nameBuffer.nameIndex(source.file.path).index)
+    buf.writeNat(pickler.nameBuffer.nameIndex(source.file.canonicalPath).index)
```
DarkDimius: This may leak private knowledge of folder structure.
Why is relative path not enough?

odersky: ExpandPrivate contains an assertion that compares two source files for equality. It only works when we canonicalize the paths. We could weaken the assertion and demand e.g. that the paths have a common non-trivial suffix. I'll make that change.

src/dotty/tools/dotc/FromTasty.scala
odersky: True. But it does not matter for anyway. The change did not affect the fixes. I'll add a comment to explain better what isTyper is.

DarkDimius:
I'm not sure if change proposed in 5aa59ac17e6f48719f8b52048e0b86570e9399e3 is good for tooling.

It makes it very complicated for an external tool to match TASTY files with source files. As the tool would be required to be able to read the TASTY section.

Though it also has bright sides, as it would simplify storing TASTY for classes originating from different source files.

odersky:
If necessary we could provide a method that returns the top level classes of a tasty file. Then we can read the source files from them. I think it is an advantage that A tasty file is not tied to one source.

VladimirNik:
I added commit with updates for TASTY tests (for dotty/src)

PR #1213, opened 2016-04-11
Fix issues with `enterBlock` for comment parsing

felixmulder:
There was a small error in my code where `enterBlock` would not add an empty list to the beginning of the `docsPerBlockStack` but instead add the elements from the empty list to the current list of lists.

review: @DarkDimius


DarkDimius:
LGTM

PR #1205, opened 2016-04-06
Ycheck that all methods have method type

liufengyun:
Fix #1046 . Please review @DarkDimius .

PR #1215, opened 2016-04-11
Fix #1214: PolyType return types may need wrapping with ExprType

### LINKED ISSUE #1214: Polymorphic methods without term params don't have MethodType
This issue is related to #1046 and PR #1205. It's expected that all methods have either `MethodicType` or `PolyType(MethodicType)`.

However, this is not true in current dotty. Given following simple program:

```Scala
class Test {
  def foo[T] = 5
}
```

The method `foo` takes a non-method type `[T]Int`. The full type signature for `foo` is as follows:

```
PolyType(
     List(T),
     List(
         TypeBounds(
             TypeRef(ThisType(TypeRef(NoPrefix,scala)),Nothing),
             TypeRef(ThisType(TypeRef(NoPrefix,scala)),Any)
          )
     ),
     TypeRef(ThisType(TypeRef(NoPrefix,scala)),Int)
)
```

Is this an intentional design or a bug?

DarkDimius:
Namer.methodType is used to get type of method given types of arguments,
    type parameters and return type.
    It has a special case if method does not take arguments that wraps return
    type with ExprType.
    Such wrapping needs to be also performed in case method has type
    parameters.

### DIFF
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -473,7 +473,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
         case tree: Select if qualifier.tpe eq tree.qualifier.tpe =>
           tree1.withTypeUnchecked(tree.tpe)
         case _ => tree.tpe match {
-          case tpe: NamedType => tree1.withType(tpe.derivedSelect(qualifier.tpe))
+          case tpe: NamedType => tree1.withType(tpe.derivedSelect(qualifier.tpe.widenIfUnstable))
           case _ => tree1.withTypeUnchecked(tree.tpe)
         }
       }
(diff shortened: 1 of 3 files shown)


DarkDimius:
Ping @liufengyun, please check if this fixes Y-check failure
@odersky please review.

liufengyun:
Just checked, this PR fixed the Y-check failure in #1205. Good job @DarkDimius !

odersky:
That's not how types are supposed to be modeled. The function 

    def f[T]: U

should have type `PolyType(<T>, <U>)`, not `PolyType(<T>, `ExprType<U>`)`. That's the assumption everywhere and we should not change it.

DarkDimius:
@odersky, would you mind explaining intension behind this design decision?
It seems natural to me to assume that return type of a `PolyType` is a `MethodicType`.
It would be also great of you could share your mind beind `def f: U` being `ExprType<U>` instand of simpy `<U>`.

PR #1218, opened 2016-04-14
Workaround for ScalaMeter incorrect report path

liufengyun:
The ScalaMeter issue is reported here:

https://github.com/scalameter/scalameter/pull/163/files

The issue exists both in v0.7 and v0.6. As dotty uses v0.6 now,
use this workaround until we upgrate to a new version of ScalaMeter.

With this PR, we no longer need to maintain two separate folders for reports.

Please review @DarkDimius .

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -239,7 +239,6 @@ object DottyBuild extends Build {
 
       libraryDependencies ++= Seq("com.storm-enroute" %% "scalameter" % "0.6" % Test,
         "com.novocode" % "junit-interface" % "0.11"),
-      testFrameworks += new TestFramework("org.scalameter.ScalaMeterFramework"),
 
       fork in Test := true,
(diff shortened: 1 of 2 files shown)


project/Build.scala
```diff
@@ -239,7 +239,6 @@ object DottyBuild extends Build {
 
       libraryDependencies ++= Seq("com.storm-enroute" %% "scalameter" % "0.6" % Test,
         "com.novocode" % "junit-interface" % "0.11"),
-      testFrameworks += new TestFramework("org.scalameter.ScalaMeterFramework"),
```
liufengyun: I removed this, because we never run `dotty-bench/test`, we use `dotty-bench/test:run`. The latter doesn't require the framework integration.

DarkDimius:
LGTM

PR #1219, opened 2016-04-15
Fix strawmans

odersky:
I worked on a new collection strawman proposal which uncovered quite a few bugs. More bugs were discovered when trying to dotty-compile this proposal and the earlier ones and then compile DottyTest against it. All discovered errors are fixed in this PR.

Review by @VladimirNik


VladimirNik:
LGTM

PR #1226, opened 2016-04-18
Multiple fixes to @static

DarkDimius:
This PR fixes multiple issues in `@static` and makes lazy vals use them instead of fields in companion classes.

### DIFF
--- a/src/dotty/tools/dotc/transform/MoveStatics.scala
+++ b/src/dotty/tools/dotc/transform/MoveStatics.scala
@@ -35,6 +35,25 @@ class MoveStatics extends MiniPhaseTransform with SymTransformer { thisTransform
       val (classes, others) = trees.partition(x => x.isInstanceOf[TypeDef] && x.symbol.isClass)
       val pairs = classes.groupBy(_.symbol.name.stripModuleClassSuffix).asInstanceOf[Map[Name, List[TypeDef]]]
 
+      def rebuild(orig: TypeDef, newBody: List[Tree]): Tree = {
+        if (orig eq null) return EmptyTree
+
+        val staticFields = newBody.filter(x => x.isInstanceOf[ValDef] && x.symbol.hasAnnotation(defn.ScalaStaticAnnot)).asInstanceOf[List[ValDef]]
+        val newBodyWithStaticConstr =
+          if (staticFields.nonEmpty) {
+            /* do NOT put Flags.JavaStatic here. It breaks .enclosingClass */
+            val staticCostructor = ctx.newSymbol(orig.symbol, Names.STATIC_CONSTRUCTOR, Flags.Synthetic  | Flags.Method, MethodType(Nil, defn.UnitType))
+            staticCostructor.addAnnotation(Annotation(defn.ScalaStaticAnnot))
+            staticCostructor.entered
+
+            val staticAssigns = staticFields.map(x => Assign(ref(x.symbol), x.rhs.changeOwner(x.symbol, staticCostructor)))
+            tpd.DefDef(staticCostructor, Block(staticAssigns, tpd.unitLiteral)) :: newBody
+          } else newBody
+
+        val oldTemplate = orig.rhs.asInstanceOf[Template]
+        cpy.TypeDef(orig)(rhs = cpy.Template(orig.rhs)(oldTemplate.constr, oldTemplate.parents, oldTemplate.self, newBodyWithStaticConstr))
+      }
+
       def move(module: TypeDef, companion: TypeDef): List[Tree] = {
         if (!module.symbol.is(Flags.Module)) move(companion, module)
         else {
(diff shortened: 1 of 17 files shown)


DarkDimius:
As a funny side effect of those changes:

```scala
import annotation.static 

class T
object T{
  @static val dummy: Unit = println("I am static initializer")
}
```
Compiles to code equivalent to:

```java
public class T {
    public static BoxedUnit dummy;

    public static {
        Predef..MODULE$.println((Object)"I am static initializer");
        dummy = BoxedUnit.UNIT;
    }
}
```

DarkDimius:
@lrytz, would you kindly take a look?

tests/neg/static-no-companion.scala
```diff
@@ -0,0 +1,4 @@
+import annotation.static
+object T {
+  @static val foo = 10 // error: needs companion class
+}
```
lrytz: don't you need a `.check` file for this test?

DarkDimius: In dotty, our neg-tests test for presence of errors on lines containing `// error: `. 
We don't check the error message.

lrytz: cool

DarkDimius:
This pr revealed a bug in lazy vals. Fixing it.

src/dotty/tools/dotc/transform/MoveStatics.scala
```diff
@@ -0,0 +1,64 @@
...
+      val flags = if (sym.is(Flags.Method)) sym.flags else sym.flags | Flags.Mutable
+      sym.copySymDenotation(owner = sym.owner.companionClass, initFlags = flags)
+    }
+    else sym
+  }
+
+  override def transformStats(trees: List[Tree])(implicit ctx: Context, info: TransformerInfo): List[Tree] = {
+    if (ctx.owner.is(Flags.Package)) {
+      val (classes, others) = trees.partition(x => x.isInstanceOf[TypeDef] && x.symbol.isClass)
+      val pairs = classes.groupBy(_.symbol.name.stripModuleClassSuffix).asInstanceOf[Map[Name, List[TypeDef]]]
```
lrytz: why do you need the cast?

DarkDimius: Static type is `Map[Name, List[Tree]]`.

lrytz: oh i see now -- since you don't need `others`, you can write `val classes = trees collect { case td: TypeDef if td.symbol.isClass => td }` and don't need the cast

src/dotty/tools/dotc/transform/MoveStatics.scala
```diff
@@ -0,0 +1,64 @@
...
+          Thicket(rebuild(companion, newCompanionBody), rebuild(module, newModuleBody))
+        }
+      }
+      val newPairs =
+        for ((name, classes) <- pairs)
+          yield
+            if (classes.tail.isEmpty) classes.head
+            else move(classes.head, classes.tail.head)
+      Trees.flatten(newPairs.toList)
+    } else trees
```
lrytz: maybe invert the condition and move this case up -- makes it easier to follow

src/dotty/tools/dotc/transform/MoveStatics.scala
```diff
@@ -0,0 +1,64 @@
...
+/** Move static methods from companion to the class itself */
+class MoveStatics extends MiniPhaseTransform with SymTransformer { thisTransformer =>
+
+  import tpd._
+  override def phaseName = "moveStatic"
+
+
+  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
+    if (sym.hasAnnotation(defn.ScalaStaticAnnot) && sym.owner.is(Flags.Module)) {
+      sym.owner.companionClass.asClass.enter(sym.symbol)
```
lrytz: do you need to remove / unlink the symbol from the module's scope?

DarkDimius: I'm kind of intentionally not doing so. 
To make sure it can be found in both places.

lrytz: ok, i don't know if that's a good idea (literally - maybe it is..)

DarkDimius: Note here we're already near the backend, so I'm preparing the tree specifically for GenBCode. 
Phases that go in this block are backend-specific and break some assumptions of compiler.
Eg `LabelDefs` reorders `<label> def`s in a _magical_ order that no one should touch.

lrytz: maybe it's related to exactly that discussion: this currently creates a static field in both `T` and `T$`

```
import annotation.static
class T
object T {
  @static val x = 99
}
```

```
public class T {
  public static I x
}
```

```
public final class T$ {
  public static I x
}
```

the one in the module class is not initialized.

lrytz: > magical

✨  :)

DarkDimius:
I have several more fixes pending. This pr at the moment has problems with initialization of lazy vals. 
This is why CI died...

DarkDimius:
@odersky please review.

odersky:
LGTM

PR #1228, opened 2016-04-19
[work in progress] Arrays of value classes

VladimirNik:
Represent an array of T where T is value class:
```scala
class T(val param: X) extends AnyVal
```
with the instance of VCX\*Array, where X\* is underlying type of T if X is primitive type and `Object` otherwise. In the current version of Dotty arrays of value classes are represented with T[] .

The version in this pull request passes dotty test suite - runtime and compile time tests (checked based on 8af61ab2df46456fb943ce614c4d22b22a297239 Dotty version).

The work on this version is in progress. Current roadmap:
1. ArrayWrapper, ArrayOps adaptation for new representation of arrays of value classes;
2. Changes required for https://github.com/lampepfl/dotty/pull/1188;
3. Support for arrays of value classes with underlying type that is value class;
4. Bug fixes.

Previous pull request: https://github.com/lampepfl/dotty/pull/729/commits/63243ab36063a59a9784d952f99c12b0c0060a4e  (part of the code is reused in this pull request and included in ee203b20305a77d906015ac34f32ecd2788b26b8)


felixmulder:
See: https://github.com/lampepfl/dotty/issues/1558

PR #1229, opened 2016-04-20
[do not merge] remove magic with benchmark tests

liufengyun:
- remove the magic with ScalaMeter test
- remove `pos` tests
- merge all dotc related tests to a single `dotty` test
- remove `Ycheck:all`

The rationale for merging dotc tests is that measurement units should be either big(whole project) or as small as a single file in order to shed light on which part of the compiler might be slower or faster.

We can add specific single file bench test as we go. Please review @DarkDimius , thanks.

### DIFF
--- a/bench/src/test/scala/Benchmarks.scala
+++ b/bench/src/test/scala/Benchmarks.scala
@@ -1,17 +1,14 @@
 package dotty.tools.benchmarks
 
-import java.lang.annotation.Annotation
-import java.lang.reflect.Method
 
-import org.junit.runner.Request
-import org.junit.runner.notification.RunNotifier
 import org.scalameter.Key.reports._
 import org.scalameter.PerformanceTest.OnlineRegressionReport
 import org.scalameter.api._
-import org.scalameter.{Context, History, persistence, currentContext}
+import org.scalameter.{Context, History, currentContext, persistence}
 import org.scalameter.reporting.RegressionReporter.Tester
+import test.CompilerTest
 
-import scala.collection.mutable.ListBuffer
+import scala.io.Source
 
 // decorator of persitor to expose info for debugging
 class DecoratorPersistor(p: Persistor) extends SerializationPersistor {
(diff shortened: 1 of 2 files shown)


liufengyun:
I've rebased the PR. The bench test runs without problem on my machine, but it generates following error on the benchmark testing machine:

```
java.lang.ClassCastException: scala.tools.nsc.backend.jvm.BTypes$ClassBType cannot be cast to scala.tools.nsc.backend.jvm.BTypes$ArrayBType
```

I guess it's related to some environment settings, I'll need to investigate more.

liufengyun:
Thanks @DarkDimius for your clue, now the cause of exception on benchmark machine is spotted and fixed in 23d14359. The cause is that bench project uses a different version of scala-compiler. Explicit specifying the version fixes the problem.

I've also rebased this PR over master. If the tests pass, it's ready to be merged and we'll start tracking performance with just two projects: `stdlib` and `dotty`. We can add more standard projects as well.

The benchmarks are at `macro-level`. We can add interesting single file `micro-level` tests as we go. And I'll play with flame graph later so that we'll have tools to diagnose micro-level benchmarks if there's significant regression.

liufengyun:
One tricky issue with using a static version of dotty for benchmarks is that the older version of dotty may fail to compile. E.g., the introduction of multiverse makes previous version of dotty code illegal.

If such things do happen, we'll have to update `dotty-src`, which breaks our objective to use a static version of code for benchmarks.

liufengyun:
Due to the fact that using a static version of dotty may not compile later, I've reverted back to always compile the latest version of dotty.

sbt incremental compilation will not interfere after this PR is merged, as I've separated bench tests from regression tests.

Review @DarkDimius .

DarkDimius:
LGTM!
Now we need to start adding small tests that benchmark specific features.
For start, i propose [patmatexhaust-huge.scala](https://github.com/dotty-staging/dotty/blob/96d939935dbbbb029e062baec48beec822611196/tests/patmat/patmatexhaust-huge.scala) for benchrmarking pattern matcher & exhaustivity checker.

liufengyun:
Thanks @DarkDimius . I'll make a `tests/bench/<feature>/` folder, and we can add feature-specific benchmarks there.

Issue #878, opened 2015-10-27
Do not maximize unconstrained tvars which appear only contravariantly in implicit parameter types


smarter:
Currently, the rules for instantiation are (from https://github.com/lampepfl/dotty/blob/c1facd5c40fa7208d7549e6727e6e7498d4d3f34/src/dotty/tools/dotc/typer/Inferencing.scala#L54):

``` scala
   *   1. T is minimized if the constraint over T is only from below (i.e.
   *      constrained lower bound != given lower bound and
   *      constrained upper bound == given upper bound).
   *   2. T is maximized if the constraint over T is only from above (i.e.
   *      constrained upper bound != given upper bound and
   *      constrained lower bound == given lower bound).
   *  If (1) and (2) do not apply:
   *   3. T is maximized if it appears only contravariantly in the given type.
   *   4. T is minimized in all other cases.
```

This is not enough to get some innocent-looking code to compile, consider:

``` scala
class X

class Message[-A]
class Seg[+A]
class IChan[A] {
  def add[B >: A](x: Seg[B])(implicit ev: Message[B]): IChan[B] = ???
}

class Test {
  def test: Unit = {
    implicit val mx: Message[X] = ???
    val fx: IChan[X] = ???
    val sx: Seg[X] = ???
    fx.add(sx)
  }
}
```

(minimized from https://github.com/molecule-labs/molecule/blob/master/molecule-core/src/main/scala/molecule/stream/ichan/immutable/SpanT.scala#L60)

The implicit parameter to `fx.add(sx)` is not found because `B` is instantiated to `Any` instead of `X`, this happens because `B` is not constrained by `sx` (so case 2 does not apply) and `B` only appear contravariantly in `Message[B]` (so case 3 applies). I think that when instantiating tvars before doing an implicit search, it makes sense to ignore case 3 and minimize them.

@odersky : What do you think?

odersky:
I think we should try to do this. But when doing this we should probably
also change the comparison function on implicits so that smaller always
wins independent of variance. I have a branch somewhere that does this;
have to dig it out.



On Tue, Oct 27, 2015 at 9:09 PM, Guillaume Martres <notifications@github.com
> wrote:

> Currently, the rules for instantiation are (from
> https://github.com/lampepfl/dotty/blob/c1facd5c40fa7208d7549e6727e6e7498d4d3f34/src/dotty/tools/dotc/typer/Inferencing.scala#L54
> ):
>
[...]

PR #1231, opened 2016-04-21
Fixes related to equality strawman

odersky:
A strawman proposal for multiversal equality and fixes to make it work under dotty. Review by @smarter.

### DIFF
--- a/tests/neg/EqualityStrawman1.scala
+++ b/tests/neg/EqualityStrawman1.scala
@@ -1,20 +1,23 @@
 package strawman.equality
+import annotation.implicitNotFound
 
 object EqualityStrawman1 {
 
   trait Eq[-T]
-  trait Impossible
+
+  @implicitNotFound("cannot compare value of type ${T} with a value outside its equality class")
+  trait Impossible[T]
 
   object Eq extends Eq[Any]
 
   trait Base {
     def === (other: Any): Boolean = this.equals(other)
-    def === (other: CondEquals)(implicit ce: Impossible): Boolean = ???
+    def === [T <: CondEquals](other: T)(implicit ce: Impossible[T]): Boolean = ???
   }
 
   trait CondEquals extends Base {
     def === [T >: this.type <: CondEquals](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
-    def === (other: Any)(implicit ce: Impossible): Boolean = ???
+    def === [T](other: T)(implicit ce: Impossible[T]): Boolean = ???
   }
 
   trait Equals[-T] extends CondEquals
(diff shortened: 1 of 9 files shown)


smarter:
LGTM, I verified that this fix the example in https://github.com/lampepfl/dotty/issues/878, I'll push an additional commit with this example as a test.
I'd like to see if I can make the logic in `IsFullyDefinedAccumulator` slightly easier to follow but that will have to wait for more urgent things to be done first.

tests/neg/EqualityStrawman1.scala
```diff
@@ -0,0 +1,79 @@
...
+    def === (other: Any): Boolean = this.equals(other)
+    def === [T <: CondEquals](other: T)(implicit ce: Impossible[T]): Boolean = ???
+  }
+
+  trait CondEquals extends Base {
+    def === [T >: this.type <: CondEquals](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
+    def === [T](other: T)(implicit ce: Impossible[T]): Boolean = ???
+  }
+
+  trait Equals[-T] extends CondEquals
```
smarter: Why have both `CondEquals` and `Equals[-T]`? Do they serve different purpose?

odersky: On Fri, Apr 22, 2016 at 5:46 PM, Guillaume Martres <notifications@github.com
> wrote:

> In tests/neg/EqualityStrawman1.scala
> <https://github.com/lampepfl/dotty/pull/1231#discussion_r60760108>:
>
> > +  @implicitNotFound("cannot compare value of type ${T} with a value outside its equality class")
> > +  trait Impossible[T]
> > +
> > +  object Eq extends Eq[Any]
> > +
> > +  trait Base {
> > +    def === (other: Any): Boolean = this.equals(other)
> > +    def === [T <: CondEquals](other: T)(implicit ce: Impossible[T]): Boolean = ???
> > +  }
> > +
> > +  trait CondEquals extends Base {
> > +    def === [T >: this.type <: CondEquals](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
> > +    def === [T](other: T)(implicit ce: Impossible[T]): Boolean = ???
> > +  }
> > +
> > +  trait Equals[-T] extends CondEquals
>
> Why have both CondEquals and Equals[-T]? Do they serve different purpose?
>

No. The next PR has only one class: EqClass for both.

 - Martin


> —
> You are receiving this because you authored the thread.
> <https://github.com/lampepfl/dotty/pull/1231/files/89f9091c2d10aa23c5d6259694df104fc14c60f7#r60760108>
>



-- 
Prof. Martin Odersky
LAMP/IC, EPFL

PR #1233, opened 2016-04-22
Syntax highlighting for REPL using ammonite as base instead of JLine

felixmulder:
One was implemented by hand and the other by using dotty's scanner. The one
built by hand is ~short, and behaves correctly.

The scanner one is unfortunately not ready for testing - there are too
many things that are workarounds for it to be a good solution as of now

The code added from Ammonite is licensed under MIT, not sure where to
put the license - but will add it once I know.


felixmulder:
Once #1238 is merged I believe this will compile under dotty.

odersky:
@felixmulder You can rebase and try again. #1238 is merged.

felixmulder:
@odersky: there was a test that compiled dotty with `-Ycheck-reentrant`. This failed because some immutable structures (with internal mutable state) where not marked `@sharable`. So now everything seems to be working fine and this is ready for review.

DarkDimius:
@odersky, it would be nice to get this in before ScalaDays.

odersky:
LGTM, thanks!

PR #1234, opened 2016-04-22
Add multiversal equals

odersky:
Based on #1231. Review by @smarter.

This makes == safe for classes that inherit from dotty.EqClass.

### DIFF
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -211,7 +211,7 @@ class Definitions {
 
   /** A trait with the following signature:
    *
-   *    trait EqClass {
+   *    trait EqClass[-U] {
    *      /** Comparison operations between values in the same equality class */
    *      final def == [T >: this.type <: EqClass[_](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
    *      final def != [T >: this.type <: EqClass[_](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -56,10 +56,19 @@ class Definitions {
   private def newSyntheticTypeParam(cls: ClassSymbol, scope: MutableScope, paramFlags: FlagSet, suffix: String = "T0") =
     newTypeParam(cls, suffix.toTypeName.expandedName(cls), ExpandedName | paramFlags, scope)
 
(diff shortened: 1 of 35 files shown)


tests/neg/equality.scala
```diff
@@ -0,0 +1,61 @@
+import dotty.{Eq,EqClass}
+
+object equality {
+
+  case class Str(str: String) extends EqClass[_]
```
smarter: Shouldn't this be `EqClass[Str]`? I'm not even sure what `extends EqClass[_]` means, that syntax is not allowed in scala 2.

odersky: If you extend `EqClass[_]` it means you have to provide instances of `Eq` yourself; the `eqEq` one won't work.

odersky:
Rebased to master. This is complete enough to be useful. Ready to be reviewed.

smarter:
It's interesting to compare this approach with the one in scalac: https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L1092-L1126 :
Besides the special handling of case classes and value classes, the main idea is to warn when comparing two types whose lub is `Object`, in dotty we usually don't trigger these warnings because `DotClass`, `Showable`, or `Hashable` is the lub. This points to an alternative approach: we could treat some classes like `Object`: warn if they're the lub of an equality check, these classes could be marked with an annotation.

src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -200,6 +209,38 @@ class Definitions {
...
         Object_finalize, Object_notify, Object_notifyAll, Object_wait, Object_waitL, Object_waitLI)
 
+  /** A trait with the following signature:
+   *
+   *    trait EqClass[-U] {
+   *      /** Comparison operations between values in the same equality class */
+   *      final def == [T >: this.type <: EqClass[_](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
+   *      final def != [T >: this.type <: EqClass[_](other: T)(implicit ce: Eq[T]): Boolean = this.equals(other)
+   *    }
+   *
```
liufengyun: A minor typo here: missing closing bracket for type parameters.

This seems to me a clever design that I feel a little difficult to appreciate fully the intricacies, e.g. what would happen: 

- if we change `trait EqClass[-U]` to `trait EqClass`, `U` is not explicitly used in the body?
- if we use `[T >: this.type <: EqClass[T]` instead `[T >: this.type <: EqClass[_]`?

odersky:
Superseded by #1247

PR #1237, opened 2016-04-25
arrayConstructors: do not rewrite ofDim for arrays of value classes

VladimirNik:
ofDim shouldn't be rewritten to newArray for arrays of value classes.

review by @DarkDimius


DarkDimius:
LGTM

PR #1238, opened 2016-04-27
Three fixes prompted by #1235

### LINKED ISSUE #1235: Match fails to compile under dotty, but not scalac
This fails to compile under dotty (but not scalac):

```scala
case class LazyList[T](headThunk: () => T, tailThunk: () => LazyList[T]){
  lazy val head = headThunk()
  lazy val tail = tailThunk()
}

object ~: {
  def unapply[T](x: LazyList[T]) = Some((x.head, x.tail))
}

object MinimizedMatchFail {
  val ll = LazyList(() => 1, () => LazyList(() => 2, () => ???))

  ll match {
    case lb ~: rest => println("success")
  }
}
```

The error given is: https://gist.github.com/felixmulder/785f75e1da5a584c5dd48abf96747649

ping: @odersky

odersky:
Fixes #1235.

Three fixes:

 - avoid forming a nonsensical type
 - sharpen criterion for dependent method types
 - better error messages for "not an extractor" problems.


smarter:
So, what's wrong with extractors with dependent method types? They seem to compile with scalac

odersky:
@smarter 

> So, what's wrong with extractors with dependent method types? They seem to compile with scalac

Indeed they do. I'll look into it.

odersky:
Review by @felixmulder ?

felixmulder:
@odersky: I rebased your branch over my repl branch, it can now compile both files that caused issues before. Regarding the implementation - the diffs look good to me - then again, I don't feel fully qualified to judge 😅

LGTM

smarter:
You reverted the "Better error diagnostics" commit but it looks like it would still be useful, but needs to be adapted to not say anything about dependent method types, otherwise LGTM.

odersky:
@smarter It's explained in the commit message. I think writing unapply's with more than one parameter is a very unlikely mistake, which does not deserve so many lines of special casing to handle.

Issue #1222, opened 2016-04-18
Failed to unpickle play-json Format (resulting in NoDenotation.owner error)


agboom:
I hope the title makes any sense. The following code fails to unpickle (using play-json 2.4.3):

```scala
import play.api.libs.json._

object NoDenotationOwnerError {
  case class Foo()

  trait FooFormats {
    implicit val fooFormats: Format[Foo] = new Format[Foo] {
      override def reads(json: JsValue): JsResult[Foo] = JsSuccess(json.as[Foo])

      override def writes(o: Foo): JsValue = Json.toJson(o)
    }
  }
}
```
The following stacktrace is printed: http://pastebin.com/B0adHAkS. The `val <none>` lines are a result of a println of the symbol when an `IDENTtree` is matched in the Scala2Unpickler.

(a full project with this code snippet and dotty configured in sbt can be found [here](https://github.com/agboom/dotty-test))

odersky:
This should be fixed now. @agboom, can you verify that it works for your use case?

agboom:
Yes, it also works for my usecase, both the example code and my project. Thanks!

PR #1241, opened 2016-04-29
Transform annotations only if defined in current run

odersky:
There's no point transforming annotations that come from
classfiles. It's inefficient to do so and it's also risky
because it means we'd have to make sense of Scala-2 generated trees.

This should avoid the error in #1222.

Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -46,7 +46,11 @@ object desugar {
      */
     override def ensureCompletions(implicit ctx: Context) =
       if (!(ctx.owner is Package))
-        if (ctx.owner is ModuleClass) ctx.owner.linkedClass.ensureCompleted()
+        if (ctx.owner.isClass) {
+          ctx.owner.ensureCompleted()
+          if (ctx.owner is ModuleClass)
+            ctx.owner.linkedClass.ensureCompleted()
+        }
         else ensureCompletions(ctx.outer)
 
     /** Return info of original symbol, where all references to siblings of the
--- a/src/dotty/tools/dotc/transform/TreeTransform.scala
+++ b/src/dotty/tools/dotc/transform/TreeTransform.scala
@@ -185,12 +185,19 @@ object TreeTransforms {
 
       ref match {
         case ref: SymDenotation =>
-          val annotTrees = ref.annotations.map(_.tree)
-          val annotTrees1 = annotTrees.mapConserve(annotationTransformer.macroTransform)
-          val annots1 = if (annotTrees eq annotTrees1) ref.annotations else annotTrees1.map(new ConcreteAnnotation(_))
+          val annots1 =
+            if (!ref.symbol.isDefinedInCurrentRun) ref.annotations // leave annotations read from class files alone
+            else {
+              val annotTrees = ref.annotations.map(_.tree)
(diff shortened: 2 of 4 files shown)


odersky:
@smarter can you review last commit please?

odersky:
The observed error was for compile-stdlib Junit test on my machine.

src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -185,12 +185,19 @@ object TreeTransforms {
 
       ref match {
         case ref: SymDenotation =>
-          val annotTrees = ref.annotations.map(_.tree)
-          val annotTrees1 = annotTrees.mapConserve(annotationTransformer.macroTransform)
-          val annots1 = if (annotTrees eq annotTrees1) ref.annotations else annotTrees1.map(new ConcreteAnnotation(_))
+          val annots1 =
+            if (!ref.symbol.isDefinedInCurrentRun) ref.annotations // leave annotations read from class files alone
```
DarkDimius: lgtm

DarkDimius:
@odersky, I'm not getting the change in 1c7c738. Would you mind providing either example or explanation?

odersky:
@DarkDimius It was something detected and diagnosed by @smarter. Maybe he can give an example. The test case he found was super-complicated though. The patch makes obvious sense; of course the class owner of some type parameters need to be completed. The only surprising thing is that this is not always the case. @smarter knows better what scenario leads to this.

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -46,7 +46,11 @@ object desugar {
      */
     override def ensureCompletions(implicit ctx: Context) =
```
smarter: The comment above could be updated to note that we also need to have the class itself completed.

smarter:
I observed this when desugaring repeated parameters (but this can happen in a bunch of place in `Desugar`):

1. In https://github.com/lampepfl/dotty/blob/96fcdd9da51e1febe9e320f774424b5ac3f8ff3d/src/dotty/tools/dotc/ast/Desugar.scala#L880-L885 we construct the untyped tree `AppliedTypeTree(ref(seqType), t))` where `ref(seqType)` is a spliced type tree whose type is a `TypeRef` to `Seq`, notice that this does not involve forcing the info of `Seq`.
2. When typing this `AppliedTypeTree`, we must first type the type constructor but there's nothing to do since it's a spliced type tree, the info of `Seq` is still unforced.
3. Then we need to type the argument of the `AppliedTypeTree`, this requires forcing the info of the corresponding type parameter (`Seq$$A`).
4. Forcing the info of `Seq$$A` fails in https://github.com/lampepfl/dotty/blob/96fcdd9da51e1febe9e320f774424b5ac3f8ff3d/src/dotty/tools/dotc/typer/Typer.scala#L854-L867 because the info of `Seq` is still unforced, so the `OriginalSymbol` attachment is missing.

So I see two ways to fix this: either make sure in `Desugar` that we never construct an `AppliedTypeTree` without forcing the info of the type constructor or changing the definition of `ensureCompletions` to ensure that, which is what https://github.com/lampepfl/dotty/commit/1c7c738a148390ffb2072fc97a81646031fb49b2 does and is very similar to what I did locally.

LGTM.

PR #1242, opened 2016-04-29
[experimental] curried update methods

acdenhartog:
This is a small experimental modification to enable once-curried update methods. This allows use of multiple parameters, varargs, and auto-tuple, on both sides of the equal sign. These things could be achieved with macros, but I think it is better to fix the built in sugar.

In theory this should not break any existing Scala code.

The `curriedUpdate` method is worth looking at because it shows how simply this issue can be dealt with. Merely un-tupling rhs the of equal sign enables multiple parameters without breaking actual tuples due to auto-tupling.

Anyway, point of this is: I dislike current update syntax and would like to see it improved. And I was curious how hard it could be to implement myself. Sorry if I broke any rules and/or wasted time.

Here is an additional snippet to demonstrate its effects:

```scala
object CurriedUpdate extends App {

  def rnd:Int = { math.random*99 }.toInt
  def prt(times:Int)(stuff: =>Any*): Unit ={
    var n = times
    while(n > 0) println{
      n -= 1
      stuff.mkString(" ... ")
    }
  }

  /** Briefly demonstrates some issues with current update syntax
    */
  object Zero { def update(a:(Double,Double),b:Int*) = prt(1)(a,b) }
  println("no auto-tuple on lhs, and no literal varargs on rhs")
  Zero((0.1,rnd)) = 2
  println("split varargs")
  Zero((0.2,rnd),1,2,3,4) = 5
  println("skipped lhs")
  Zero() = (0,3)
  println(":_* on the rhs compiles with scalac, but not with dotty")
  //Zero((0.4,rnd)) = List[Int](1,2,3):_* // <<--- does not compile

  /** Proposed solution is once-curried update method.
    * Benefits:
    * - intuitive method definition (lhs)(rhs)
    * - allows multiple parameters on rhs
    * - predictable varargs behaviour
    * - lhs auto-tupling as with any other method
    * - no need to break compatibility
    */

  object One { def update(a: =>String)(b: =>Int) = prt(3)(a,b) }
  println("pass by name works on both sides")
  One(rnd.toString) = rnd

  object Two { def update(a: =>Int*)(b: =>Int*) = prt(3)(a,b) }
  println("varargs works on both sides")
  Two(2,1,rnd) = (2,1,rnd,rnd,rnd)
  Two(2,2,rnd) = rnd
  Two(List(2,3,rnd):_*) = (rnd)
  //Two(2,4) = List(2,4,rnd):_* // <<--- does not compile

  object Three { def update(a:(String,String))(b:(Int,Int)) = prt(3)(a,b) }
  println("compiler always tries to untuple rhs, but auto-tupling restores it")
  Three(("3","1")) = (rnd,rnd)
  println("auto-tupling also works on lhs, unlike current scala")
  Three("3","2") = (rnd,rnd)

  object Four { def update(a:(Int,Int))(b: =>(Int,Int)*) = prt(3)(a,b) }
  Four(4,1) = ((1,2),(3,4),(5,6),(rnd,rnd))
  //Four(4,2) = (1,2) <<--- corner case does not compile because auto-tupling not enabled with varargs

}
```

Update methods are in separate objects to remove compile-time "merge" errors, whatever those are...
But in general curried update methods would require that lhs type signature is unique; further ravings about that: https://gist.github.com/acdenhartog/a47fad6ebe8d32ff0ff5#file-5-overloadingequality-scala

### DIFF
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -460,16 +460,38 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   }
 
   def typedAssign(tree: untpd.Assign, pt: Type)(implicit ctx: Context) = track("typedAssign") {
+    def curriedUpdate(canCopy:untpd.Tree,update:untpd.Tree,lhsArgs:List[untpd.Tree]) = {
+      val step1 = cpy.Apply(canCopy)(update,lhsArgs)
+      typed(tree.rhs match {
+        case untpd.Tuple(rhsArgs) => untpd.Apply(step1,rhsArgs)
+        case rhsNotATuple => untpd.Apply(step1,rhsNotATuple)
+      },pt)
+    }
+    /** method tries normal update syntax, if that causes a type error, switch to curried mode
+      * not an effcient way to do this...
+      * Uses uncurried method in some ambiguous cases; should maybe give an error/warning.
+      */
+    def compatibilityMode(canCopy:untpd.Tree,update:untpd.Tree,lhsArgs:List[untpd.Tree]) = {
+      val trying = ctx.fresh.setNewTyperState
+      val uncurriedUpdate = cpy.Apply(canCopy)(update, lhsArgs :+ tree.rhs)(trying)
+      val tried =
+        try Some(adapt(typedUnadapted(uncurriedUpdate, pt)(trying), pt, uncurriedUpdate)(trying))
+        catch { case _: TypeError => None }
+      if(trying.reporter.hasErrors)
+        curriedUpdate(canCopy,update,lhsArgs)
+      else {
+        trying.typerState.commit()
+        tried.get
+      }
+    }
     tree.lhs match {
-      case lhs @ Apply(fn, args) =>
-        typed(cpy.Apply(lhs)(untpd.Select(fn, nme.update), args :+ tree.rhs), pt)
-      case untpd.TypedSplice(Apply(MaybePoly(Select(fn, app), targs), args)) if app == nme.apply =>
+      case lhs : untpd.Apply => compatibilityMode(lhs,untpd.Select(lhs.fun, nme.update),lhs.args)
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -460,16 +460,38 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+        case rhsNotATuple => untpd.Apply(step1,rhsNotATuple)
+      },pt)
+    }
+    /** method tries normal update syntax, if that causes a type error, switch to curried mode
+      * not an effcient way to do this...
+      * Uses uncurried method in some ambiguous cases; should maybe give an error/warning.
+      */
+    def compatibilityMode(canCopy:untpd.Tree,update:untpd.Tree,lhsArgs:List[untpd.Tree]) = {
+      val trying = ctx.fresh.setNewTyperState
+      val uncurriedUpdate = cpy.Apply(canCopy)(update, lhsArgs :+ tree.rhs)(trying)
```
odersky: There's a tryEither method in the Typer that embodies this logic. It seems DRYer to use that.

odersky:
I think this needs a SIP. In particular we have to decide whether to allow both update conventions at the same time. I tend to allow only one of them, but then we'd have to talk about interop with old binaries.

acdenhartog:
Thanks for the comments.

I posted about this issue in some Scala related forums (google groups and reddit) but it got little response there. I did consider writing a SIP proposal then, but I can't do paperwork to save my life.

I do completely agree with the spirit of allowing only one convention.

At first glance an auto-rewrite seems easy, but currying methods could lead to ambiguities wherever 'equality has been overloaded'. And that splitting varargs behavior [can be exploited with macros](https://coderwall.com/p/_cyziw/varargs-update-support-in-scala) or just by using the same types on both sides.

acdenhartog:
I have given this some additional thought and it may be worth further extending `typedAssign` and the notion of un-tupling. I am going to close this pull request and make a new one when I have implemented something more comprehensive.

Issue #1240, opened 2016-04-28
Fun with overrides


DarkDimius:
Simple example. Fails in scalac\javac, succeeds in Dotty.

```scala
class A[T] {
  def foo(x: T): T
  def foo(x: Int): Int
  new A[Int]().foo(1) // what foo should be called? scalac/javac issue an error here. Dotty calls the lexically-first one
}
```

More complicated example:
```scala
abstract class A {
  type C[X]
  def foo[B](x: C[B]): C[B]       = {println("A.C"); x}
  def foo[B](x: List[B]): List[B] = {println("A.List"); x}
  def give[X]: C[X]
}

class B extends A {
  type C[X] = List[X]
  override def give[X] = Nil
  override def foo[B](x: C[B]): C[B] =  {println("B.C"); x} 
                                           // which method is overriden? 
                                           // should any bridges be generated?
  val a: A = this
  a.foo(a.give[Int]) // what method should be called here in runtime?
}
```

In this example neither scalac nor dotty generate bridges.
Ping @retronym @adriaanm

retronym:
```
scala> class C[A]{def foo(a: A) = 0; def foo(i: String) = 1}; class D extends C[String]
defined class C
defined class D

scala> new D().foo("")
<console>:14: error: ambiguous reference to overloaded definition,
both method foo in class C of type (i: String)Int
and  method foo in class C of type (a: String)Int
match argument types (String)
       new D().foo("")
               ^

scala> (new D(): C[String]).foo("")
<console>:14: error: ambiguous reference to overloaded definition,
both method foo in class C of type (i: String)Int
and  method foo in class C of type (a: String)Int
match argument types (String)
       (new D(): C[String]).foo("")
                            ^

scala> (new D(): C[_]).foo("")
res12: Int = 1
```

Compare with:

```java
public class Test {
	public static class C<A> {
		public int foo(A a) { return 0; }
		public int foo(String i) { return 1; }
	}
	public static class D extends C<String> {}
}
```

```
% javac sandbox/Test.java
sandbox/Test.java:6: error: methods foo(A) from C<String> and foo(String) from C<String> are inherited with the same signature
	public static class D extends C<String> {}
	              ^
  where A is a type-variable:
    A extends Object declared in class C
1 error
```

I'd argue we should perform the same check in `refchecks`.

We're in good company with this omission, JetBrains only fixed it in their Java support three months ago.

[...]

PR #1243, opened 2016-05-01
Fix overriding problems

odersky:
This fixes the missing "ambiguous overload" error of #1240. More work is needed to fix the multiple overriddenSymbols scenario.

Review by @retronym or @DarkDimius


src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -282,8 +282,15 @@ object Denotations {
...
             val sym1 = denot1.symbol
             val sym2 = denot2.symbol
-            val sym2Accessible = sym2.isAccessibleFrom(pre)
 
+            if (sym1.exists && sym2.exists &&
+                (sym1 ne sym2) && (sym1.owner eq sym2.owner) &&
+                !sym1.is(Bridge) && !sym2.is(Bridge))
+              // double definition of two methods with same signature in one class;
+              // don't merge them.
+              return NoDenotation
```
retronym: Can we get here with private `sym1` or `sym2`? That's another case that prevents overriding.

retronym: Also, (after looking at your next commit), can we get here with constructors? These also cannot be invoved in overriding.

odersky: In both cases, the existing resolution will pick the right symbol. So these should not be a problem.

odersky:
Going to merge by tomorrow unless there are further comments

odersky:
Added one important commit and rebased to master.

smarter:
> Interestingly, the failure only triggers when -Ycheck options are *disabled*

Any idea why?

DarkDimius:
@odersky I didn't yet have a chance  to see if this fix is sufficient. I'm fine with meging it now if it's needed in the mainline.

PR #1244, opened 2016-05-02
Add sbt incremental compilation support

smarter:
Review by @DarkDimius

***
To test this with sbt, see
https://github.com/lampepfl/dotty/wiki/Using-Dotty-with-sbt

The following flags are added:
 - `-Yforce-sbt-phases`: Run the phases used by sbt for incremental compilation
   (ExtractDependencies and ExtractAPI) even if the compiler is ran outside of
   sbt, for debugging.
 - `-Ydump-sbt-inc`: For every compiled foo.scala, output the sbt API
   representation and dependencies in foo.inc, implies -Yforce-sbt-phases.

This commit introduces two new phases which do not transform trees:
- `ExtractDependencies` which extracts the dependency information of the current
  compilation unit and sends it to sbt via callbacks
- `ExtractAPI` which creates a representation of the API of the current compilation
  unit and sends it to sbt via callbacks

Briefly, when a file changes sbt will recompile it, if its API has
changed (determined by what `ExtractAPI` sent) then sbt will determine
which reverse-dependencies (determined by what `ExtractDependencies`
sent) of the API have to be recompiled depending on what changed.
See http://www.scala-sbt.org/0.13/docs/Understanding-Recompilation.html for
more information on how sbt incremental compilation works.

This phase was originally based on
https://github.com/adriaanm/scala/tree/sbt-api-consolidate/src/compiler/scala/tools/sbt
which attempts to integrate the sbt phases into scalac (and is itself based
on https://github.com/sbt/sbt/tree/0.13/compile/interface/src/main/scala/xsbt),
but it has been heavily refactored and adapted to Dotty. The main
functional differences are:
- ExtractDependencies runs right after Frontend (so that we don't lose
  dependency informations because of the simplifications done by PostTyper),
  but ExtractAPI runs right after PostTyper (so that SuperAccessors are
  part of the API).
- `ExtractAPI` only extract types as they are defined and never "as seen
  from" some some specific prefix, see its documentation for more details.
- `ExtractDependenciesTraverser` and `ExtractUsedNames` have been fused into
  one tree traversal in `ExtractDependenciesCollector`.

TODO: Try to run these phases in parallel with the rest of the compiler
pipeline since they're independent (except for the sbt callbacks in `GenBCode`) ?

### DIFF
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -42,7 +42,9 @@ class Compiler {
   def phases: List[List[Phase]] =
     List(
       List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
+      List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
       List(new PostTyper),          // Additional checks and cleanups after type checking
+      List(new sbt.ExtractAPI),     // Sends a representation of the API of classes to sbt via callbacks
       List(new Pickler),            // Generate TASTY info
       List(new FirstTransform,      // Some transformations to put trees into a canonical form
            new CheckReentrant),     // Internal use only: Check that compiled program has no data races involving global vars
--- a/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -160,6 +160,8 @@ class ScalaSettings extends Settings.SettingGroup {
   val YtestPickler = BooleanSetting("-Ytest-pickler", "self-test for pickling functionality; should be used with -Ystop-after:pickler")
   val YcheckReentrant = BooleanSetting("-Ycheck-reentrant", "check that compiled program does not contain vars that can be accessed from a global root.")
   val YkeepComments = BooleanSetting("-Ykeep-comments", "Keep comments when scanning source files.")
+  val YforceSbtPhases = BooleanSetting("-Yforce-sbt-phases", "Run the phases used by sbt for incremental compilation (ExtractDependencies and ExtractAPI) even if the compiler is ran outside of sbt, for debugging.")
+  val YdumpSbtInc = BooleanSetting("-Ydump-sbt-inc", "For every compiled foo.scala, output the API representation and dependencies used for sbt incremental compilation in foo.inc, implies -Yforce-sbt-phases.")
   def stop = YstopAfter
 
   /** Area-specific debug output.
--- a/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -868,7 +868,7 @@ object SymDenotations {
      */
     final def topLevelClass(implicit ctx: Context): Symbol = {
       def topLevel(d: SymDenotation): Symbol = {
-        if ((d is PackageClass) || (d.owner is PackageClass)) d.symbol
+        if (d.isEffectiveRoot || (d is PackageClass) || (d.owner is PackageClass)) d.symbol
         else topLevel(d.owner)
       }
       val sym = topLevel(this)
--- a/src/dotty/tools/dotc/sbt/ExtractAPI.scala
+++ b/src/dotty/tools/dotc/sbt/ExtractAPI.scala
@@ -1,0 +1,479 @@
+package dotty.tools.dotc
+package sbt
+
+import ast.{Trees, tpd}
+import core._, core.Decorators._
+import Contexts._, Flags._, Phases._, Trees._, Types._, Symbols._
+import Names._, NameOps._, StdNames._
+
+import dotty.tools.io.Path
+import java.io.PrintWriter
+
+import scala.collection.mutable
+
+/** This phase sends a representation of the API of classes to sbt via callbacks.
+ *
+ *  This is used by sbt for incremental recompilation.
+ *
+ *  See the documentation of `ExtractAPICollector`, `ExtractDependencies`,
+ *  `ExtractDependenciesCollector` and
+ *  http://www.scala-sbt.org/0.13/docs/Understanding-Recompilation.html for more
+ *  information on incremental recompilation.
+ *
+ *  The following flags affect this phase:
+ *   -Yforce-sbt-phases
+ *   -Ydump-sbt-inc
+ *
+ *  @see ExtractDependencies
+ */
+class ExtractAPI extends Phase {
+  override def phaseName: String = "sbt-api"
+
+  // SuperAccessors need to be part of the API (see the scripted test
+  // `trait-super` for an example where this matters), this is only the case
+  // after `PostTyper` (unlike `ExtractDependencies`, the simplication to trees
+  // done by `PostTyper` do not affect this phase because it only cares about
+  // definitions, and `PostTyper` does not change definitions).
+  override def runsAfter = Set(classOf[transform.PostTyper])
+
+  override def run(implicit ctx: Context): Unit = {
+    val unit = ctx.compilationUnit
+    val dumpInc = ctx.settings.YdumpSbtInc.value
+    val forceRun = dumpInc || ctx.settings.YforceSbtPhases.value
+    if ((ctx.sbtCallback != null || forceRun) && !unit.isJava) {
+      val sourceFile = unit.source.file.file
+      val apiTraverser = new ExtractAPICollector
+      val source = apiTraverser.apiSource(unit.tpdTree)
+
+      if (dumpInc) {
+        // Append to existing file that should have been created by ExtractDependencies
+        val pw = new PrintWriter(Path(sourceFile).changeExtension("inc").toFile
+          .bufferedWriter(append = true), true)
+        try {
+          pw.println(DefaultShowAPI(source))
+        } finally pw.close()
+      }
+
+      if (ctx.sbtCallback != null)
+        ctx.sbtCallback.api(sourceFile, source)
+    }
+  }
+}
+
+/** Extracts full (including private members) API representation out of Symbols and Types.
+ *
+ *  The exact representation used for each type is not important: the only thing
+ *  that matters is that a binary-incompatible or source-incompatible change to
+ *  the API (for example, changing the signature of a method, or adding a parent
+ *  to a class) should result in a change to the API representation so that sbt
+ *  can recompile files that depend on this API.
+ *
+ *  Note that we only records types as they are defined and never "as seen from"
+ *  some other prefix because `Types#asSeenFrom` is a complex operation and
+ *  doing it for every inherited member would be slow, and because the number
+ *  of prefixes can be enormous in some cases:
+ *
+ *    class Outer {
+ *      type T <: S
+ *      type S
+ *      class A extends Outer { /*...*/ }
+ *      class B extends Outer { /*...*/ }
+ *      class C extends Outer { /*...*/ }
+ *      class D extends Outer { /*...*/ }
+ *      class E extends Outer { /*...*/ }
+ *    }
+ *
+ *  `S` might be refined in an arbitrary way inside `A` for example, this
+ *  affects the type of `T` as seen from `Outer#A`, so we could record that, but
+ *  the class `A` also contains itself as a member, so `Outer#A#A#A#...` is a
+ *  valid prefix for `T`. Even if we avoid loops, we still have a combinatorial
+ *  explosion of possible prefixes, like `Outer#A#B#C#D#E`.
+ *
+ *  It is much simpler to record `T` once where it is defined, but that means
+ *  that the API representation of `T` may not change even though `T` as seen
+ *  from some prefix has changed. This is why in `ExtractDependencies` we need
+ *  to traverse used types to not miss dependencies, see the documentation of
+ *  `ExtractDependencies#usedTypeTraverser`.
+ *
+ *  TODO: sbt does not store the full representation that we compute, instead it
+ *  hashes parts of it to reduce memory usage, then to see if something changed,
+ *  it compares the hashes instead of comparing the representations. We should
+ *  investigate whether we can just directly compute hashes in this phase
+ *  without going through an intermediate representation, see
+ *  http://www.scala-sbt.org/0.13/docs/Understanding-Recompilation.html#Hashing+an+API+representation
+ */
+private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder {
+  import tpd._
+  import xsbti.api
+
+  /** This cache is necessary for correctness, see the comment about inherited
+   *  members in `apiClassStructure`
+   */
+  private[this] val classLikeCache = new mutable.HashMap[ClassSymbol, api.ClassLike]
+  /** This cache is optional, it avoids recomputing representations */
+  private[this] val typeCache = new mutable.HashMap[Type, api.Type]
+
+  private[this] object Constants {
+    val emptyStringArray = Array[String]()
+    val local            = new api.ThisQualifier
+    val public           = new api.Public
+    val privateLocal     = new api.Private(local)
+    val protectedLocal   = new api.Protected(local)
+    val unqualified      = new api.Unqualified
+    val thisPath         = new api.This
+    val emptyType        = new api.EmptyType
+    val emptyModifiers   =
+      new api.Modifiers(false, false, false, false, false,false, false, false)
+  }
+
+  /** Some Dotty types do not have a corresponding type in xsbti.api.* that
+   *  represents them. Until this is fixed we can workaround this by using
+   *  special annotations that can never appear in the source code to
+   *  represent these types.
+   *
+   *  @param tp      An approximation of the type we're trying to represent
+   *  @param marker  A special annotation to differentiate our type
+   */
+  private def withMarker(tp: api.Type, marker: api.Annotation) =
+    new api.Annotated(tp, Array(marker))
+  private def marker(name: String) =
+    new api.Annotation(new api.Constant(Constants.emptyType, name), Array())
+  val orMarker = marker("Or")
+  val byNameMarker = marker("ByName")
+
+
+  /** Extract the API representation of a source file */
+  def apiSource(tree: Tree): api.SourceAPI = {
+    val classes = new mutable.ListBuffer[api.ClassLike]
+    def apiClasses(tree: Tree): Unit = tree match {
+      case PackageDef(_, stats) =>
+        stats.foreach(apiClasses)
+      case tree: TypeDef =>
+        classes += apiClass(tree.symbol.asClass)
+      case _ =>
+    }
+
+    apiClasses(tree)
+    forceThunks()
+    new api.SourceAPI(Array(), classes.toArray)
+  }
+
+  def apiClass(sym: ClassSymbol): api.ClassLike =
+    classLikeCache.getOrElseUpdate(sym, computeClass(sym))
+
+  private def computeClass(sym: ClassSymbol): api.ClassLike = {
+    import xsbti.api.{DefinitionType => dt}
+    val defType =
+      if (sym.is(Trait)) dt.Trait
+      else if (sym.is(ModuleClass)) {
+        if (sym.is(PackageClass)) dt.PackageModule
+        else dt.Module
+      } else dt.ClassDef
+
+    val selfType = apiType(sym.classInfo.givenSelfType)
+
+    val name = if (sym.is(ModuleClass)) sym.fullName.sourceModuleName else sym.fullName
+
+    val tparams = sym.typeParams.map(tparam => apiTypeParameter(
+      tparam.name.toString, tparam.variance,
+      tparam.info.bounds.lo, tparam.info.bounds.lo))
+
+    val structure = apiClassStructure(sym)
+
+    new api.ClassLike(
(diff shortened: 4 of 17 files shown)


smarter:
To run the scripted tests:
```shell 
git clone https://github.com/smarter/sbt.git -b inc-dotty
cd sbt
sbt "scripted source-dependencies/*"
```

interfaces/src/main/java/dotty/tools/dotc/interfaces/CompilerCallback.java
```diff
@@ -25,4 +27,12 @@
    *                        Example: ./src/library/scala/collection/Seq.scala
    */
   default void onSourceCompiled(SourceFile source) {};
+
+  default void sourceDependency(File dependsOn, File source, xsbti.DependencyContext context) {};
```
DarkDimius: I dont see what a method does given its name. Should it be `addSourceDependency`?

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1060,6 +1060,8 @@ object SymDenotations {
       else defn.RootClass
     }
 
+    def isPublic(implicit ctx: Context): Boolean = !this.is(Protected | Private) && !privateWithin.exists
```
DarkDimius: why do you need to check `privateWithin` if no `Protected | Private` is set?

DarkDimius: why do you need to check `privateWithin` if no `Protected | Private` is set?

smarter: Because we remove `Private` when a qualifier exists: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/parsing/Parsers.scala#L1422-L1427 `scalac` does the same thing. I guess this is because private qualified are not private in the bytecode, maybe we should do this in a miniphase later.

interfaces/src/main/java/dotty/tools/dotc/interfaces/CompilerCallback.java
smarter: I'll remove these extra callbacks, they're not needed now that I added `Context#sbtCallback` instead.

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -160,6 +160,8 @@ class ScalaSettings extends Settings.SettingGroup {
   val YtestPickler = BooleanSetting("-Ytest-pickler", "self-test for pickling functionality; should be used with -Ystop-after:pickler")
   val YcheckReentrant = BooleanSetting("-Ycheck-reentrant", "check that compiled program does not contain vars that can be accessed from a global root.")
   val YkeepComments = BooleanSetting("-Ykeep-comments", "Keep comments when scanning source files.")
+  val YforceSbtPhase = BooleanSetting("-Yforce-sbt-phase", "Run the sbt incremental compiler phase even if the compiler is ran outside of sbt, for debugging.")
+  val YdumpSbtApi = BooleanSetting("-Ydump-sbt-api", "For every compiled foo.scala, output the sbt API representation in foo.api, implies -Yforce-sbt-phase.")
```
DarkDimius: `dependsOn(YforceSbtPhase)`?

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1060,6 +1060,12 @@ object SymDenotations {
       else defn.RootClass
     }
 
+    /** Is this denotation public ? */
+    def isPublic(implicit ctx: Context): Boolean =
+      // private-qualified denotations do not have the Private flag set,
+      // see `ParserCommon#normalize`
+      !this.is(Protected | Private) && !privateWithin.exists
```
DarkDimius: Why would you need to check for `privateWithin.exists` if neither `Protected`  nor `Private` flag is set?

src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -569,7 +569,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile)(implicit ctx: Context
     def isConstructor: Boolean = toDenot(sym).isConstructor
     def isAnonymousFunction: Boolean = toDenot(sym).isAnonymousFunction
     def isMethod: Boolean = sym is Flags.Method
-    def isPublic: Boolean =  sym.flags.is(Flags.EmptyFlags, Flags.Private | Flags.Protected)
+    def isPublic: Boolean =  toDenot(sym).isPublic
```
DarkDimius: private [foo] methods should NOT be private on bytecode level.

src/dotty/tools/dotc/config/ScalaSettings.scala
smarter: Looks like this would force the user to write `-Yforce-sbt-phase -Ydump-sbt-api`, I can do the change if you want but I don't think it's very useful

src/dotty/tools/backend/jvm/DottyBackendInterface.scala
smarter: I agree, but note that `isPublic` in scalac is defined in the same way: https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/HasFlags.scala#L117
So I think the backend already takes this into account.

src/dotty/tools/dotc/core/SymDenotations.scala
smarter: See https://github.com/lampepfl/dotty/pull/1244#discussion_r62421206

src/dotty/tools/backend/jvm/DottyBackendInterface.scala
DarkDimius: The common bakend does not.
I'm not sure why this is needed in scalac. I'm not in favor of such change in Dotty backend.
Nor do I think that we should add this incorrect implementation of `isPrivate` method: If method is defined as `private [package] def foo = ???` it has to be compiled as public. 
If some specific phase has different notion of `isPublic` - such implementation can go into this phase.

src/dotty/tools/dotc/core/SymDenotations.scala
smarter: This link behaves weirdly so copy-pasting:


Because we remove Private when a qualifier exists: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/parsing/Parsers.scala#L1422-L1427 scalac does the same thing. I guess this is because private qualified are not private in the bytecode, maybe we should do this in a miniphase later.

DarkDimius: What you're saing that in your sbt-specific phase you have a different notion of public. 
For all other phases notion of `public` is simple: can a method\symbol\type be accessed from classes that are not subclasses of this class. Method that is `private[package]` is public from this definition. 

I'm against such change to either backend or SymDenotations.scala.

smarter: A method that is `private[package]` is public in the bytecode but it's not public in the Scala language. `public` means accessible from everywhere, `private[package]` is only accessible in some cases (if you're in the same package).
Anyone who writes their own phase for dotty and wants to check if a symbol is public in the Scala language will want `isPublic` to return false for something which is `private[package]`, the fact that it is public in the bytecode is an implementation detail.

Anyway, I don't want to fight on this, I'll just put `def isPublic` in sbt and leave the rest alone.

smarter: done.

smarter:
I didn't realize this before but there's a single test that does not pass now: [`trait-super`](https://github.com/smarter/dotty-bridge/tree/incremental/src/sbt-test/source-dependencies/trait-super) because it relies on the presence or absence of superaccesors in the API to know what should be recompiled, but in dotty superaccessors are created by `PostTyper` but the sbt phase is run before `PostTyper` so that it can see constant types and trees of types (and so that the [`constants`](https://github.com/smarter/dotty-bridge/tree/incremental/src/sbt-test/source-dependencies/constants) test pass), I don't know if there's a way to fix this without doing some phase-travelling. Anyway, this is something that can be fixed later.

smarter:
OK I just fixed [`trait-super`](https://github.com/smarter/dotty-bridge/tree/incremental/src/sbt-test/source-dependencies/trait-super) by splitting the sbt phase into two phases, `ExtractDependencies` before `PostTyper` and `ExtractAPI` after `PostTyper` (they were already completely independent before, but ran in the same phase one after the other), this works out because:
- `ExtractDependencies` needs to be before `PostTyper` to avoid loosing dependency informations (see [`constants`](https://github.com/smarter/dotty-bridge/tree/incremental/src/sbt-test/source-dependencies/constants) for an example where this matters).
- `ExtractAPI` needs to be after `PostTyper` because SuperAccessors need to be part of the API (see again [`trait-super`](https://github.com/smarter/dotty-bridge/tree/incremental/src/sbt-test/source-dependencies/trait-super) for an example where this matter). Unlike `ExtractDependencies`, the simplication to trees done by `PostTyper` do not affect this phase because it only cares about definitions, and `PostTyper` does not change definitions.

I've also cleaned up the changes needed to `dotty-bridge`, they now sit in https://github.com/smarter/dotty-bridge/pull/1 and will be merged once this PR has been merged.

DarkDimius:
Needs rebasing.

src/dotty/tools/dotc/Compiler.scala
```diff
@@ -41,56 +41,58 @@ class Compiler {
    */
   def phases: List[List[Phase]] =
     List(
-      List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
```
DarkDimius: I'd prefer not to loose git history here.
Indentation is not that important.

Additionally, it will introduce a lot of merge conflicts with linker.

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -160,6 +160,8 @@ class ScalaSettings extends Settings.SettingGroup {
   val YtestPickler = BooleanSetting("-Ytest-pickler", "self-test for pickling functionality; should be used with -Ystop-after:pickler")
   val YcheckReentrant = BooleanSetting("-Ycheck-reentrant", "check that compiled program does not contain vars that can be accessed from a global root.")
   val YkeepComments = BooleanSetting("-Ykeep-comments", "Keep comments when scanning source files.")
+  val YforceSbtPhases = BooleanSetting("-Yforce-sbt-phases", "Run the phases used by sbt for incremental compilation (ExtractDependencies and ExtractAPI) even if the compiler is ran outside of sbt, for debugging.")
+  val YdumpSbtInc = BooleanSetting("-Ydump-sbt-inc", "For every compiled foo.scala, output the sbt API representation and dependencies in foo.inc, implies -Yforce-sbt-phases.")
```
DarkDimius: it's not clear for me what does `Inc` stand for. The description does speak about it.

test/dotc/tests.scala
```diff
@@ -23,7 +23,7 @@ class tests extends CompilerTest {
   val defaultOutputDir = "./out/"
 
   implicit val defaultOptions = noCheckOptions ++ List(
-      "-Yno-deep-subtypes", "-Yno-double-bindings",
+      "-Yno-deep-subtypes", "-Yno-double-bindings", "-Yforce-sbt-phases",
```
DarkDimius: This is going to affect benchmarks and skew history.

src/dotty/tools/dotc/sbt/ExtractAPI.scala
```diff
@@ -0,0 +1,479 @@
...
+  // after `PostTyper` (unlike `ExtractDependencies`, the simplication to trees
+  // done by `PostTyper` do not affect this phase because it only cares about
+  // definitions, and `PostTyper` does not change definitions).
+  override def runsAfter = Set(classOf[transform.PostTyper])
+
+  override def run(implicit ctx: Context): Unit = {
+    val unit = ctx.compilationUnit
+    val dumpInc = ctx.settings.YdumpSbtInc.value
+    val forceRun = dumpInc || ctx.settings.YforceSbtPhases.value
+    if ((ctx.sbtCallback != null || forceRun) && !unit.isJava) {
```
DarkDimius: What if someone changes java source? Don't you still need to rebuild?

src/dotty/tools/dotc/sbt/ThunkHolder.scala
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /** Store the parameter `s` in a `Lazy` container, since `s` is not by-name, there
+   *  is nothing to force.
+   *
+   *  TODO: Get rid of this method. It is only needed because some xsbti.api classes
+   *  take lazy arguments when they could be strict, but this can be fixed in sbt,
+   *  see https://github.com/sbt/zinc/issues/114
+   */
+  def strictlzy[T <: AnyRef](t: T): api.Lazy[T] =
```
DarkDimius: this method has unreadable name and is part of public interface.
Suggestion:
```
private def lazify...
```

src/dotty/tools/dotc/sbt/ThunkHolder.scala
```diff
@@ -0,0 +1,59 @@
+package dotty.tools.dotc
+package sbt
+
+import scala.annotation.tailrec
+import scala.collection.mutable.ListBuffer
+import xsbti.api
+
+/** Create and hold thunks */
```
DarkDimius: It's not clear for me what are thunks.

src/dotty/tools/dotc/sbt/ShowAPI.scala
```diff
@@ -0,0 +1,156 @@
+// This file is copied straight from
```
DarkDimius: as we always have sbt on classpath, isn't it simpler to simply use the one provided by sbt instead of maintaining a copy of it?

src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -0,0 +1,265 @@
...
+      tp match {
+        case tp: NamedType =>
+          val sym = tp.symbol
+          if (!sym.is(Package)) {
+            addDependency(sym)
+            if (!sym.isClass)
+              traverse(tp.info)
+            traverse(tp.prefix)
+          }
+        case tp: ThisType =>
```
DarkDimius:
```
case tp: TypeProxy =>
  traverse(tp.underlying)
```
instead of all the cases?

src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -0,0 +1,265 @@
...
+ *  The following flags affect this phase:
+ *   -Yforce-sbt-phases
+ *   -Ydump-sbt-inc
+ *
+ *  @see ExtractAPI
+ */
+class ExtractDependencies extends Phase {
+  override def phaseName: String = "sbt-deps"
+
+  // This phase should be run directly after `Frontend`, if it is run after
```
DarkDimius: this predicate can be checked by inspecting `ctx.phases` in `def run`

src/dotty/tools/dotc/sbt/ThunkHolder.scala
DarkDimius: This trait is very small and used only by `ExtractAPICollector`. It's not clear what it does. I propose to merge it into `ExtractAPICollector`

DarkDimius:
I've done a quick review, as I'm not into details of `xsbti.api`.

I'd be curious to see how much does introduction of this phase affect compilation time of dotty&stdlib.

Otherwise LGTM

src/dotty/tools/dotc/Compiler.scala
smarter: OK, will revert formatting change (though note that `git blame -w` and `git diff -w` make this less important)

src/dotty/tools/dotc/config/ScalaSettings.scala
smarter: inc stands for incremental, I'll make that explicit in the description

test/dotc/tests.scala
smarter: Yup, no choice.

src/dotty/tools/dotc/sbt/ExtractAPI.scala
smarter: Java sources are handled differently by sbt, I don't actually know the details.

src/dotty/tools/dotc/sbt/ThunkHolder.scala
smarter: I'm confused by your suggestion, I can't have an overloaded method where the only difference is that the argument is `T` or `=> T`, they would conflict with each other.

`strictlzy` is only temporary anyway so its name is not super important, I could rename it to `strictLzy` which is easier to understand.

src/dotty/tools/dotc/sbt/ThunkHolder.scala
smarter: A thunk is an unevaluated value that can be evaluated once, it's a well known term (especially used in Haskell, but not only). I can make the description longer to make this clearer.

This trait was intentionally split from `ExtractAPICollector` and I'd like to keep it this way:
- It has private mutable state that does not need to be modified by `ExtractAPICollector`
- The implementation is somewhat tricky, but it can be understood on its own without worrying about how it's used in `ExtractAPICollector`
- Similarly, someone can understand how `ExtractAPICollector` works without having to know the implementation details of `ThunkHolder`

src/dotty/tools/dotc/sbt/ShowAPI.scala
smarter: When running dotty outside of sbt, we only have `"org.scala-sbt" % interface` (which only contains Java interfaces) on the classpath, `ShowAPI.scala` is in `"org.scala-sbt" % api` which will only be on the classpath when running dotty from sbt, but we want `-Ydump-sbt-inc` to work without running dotty from sbt to make debugging easier.

src/dotty/tools/dotc/sbt/ExtractDependencies.scala
smarter: No, that would be incorrect. A `TypeBounds` is a `TypeProxy` whose `underlying` is the upper bound, but as the example in the documentation of this object shows, we need to record both the upper and lower bounds, `TypeTraverser#traverseChildren` will do that for us.

smarter:
> I'd be curious to see how much does introduction of this phase affect compilation time of dotty&stdlib.

I spent a bunch of time optimizing it and currently it adds a few seconds to compilation, most of that time is just because of the extra tree traversal done by `ExtractDependenciesCollector` which I don't think we can avoid (I don't think `ExtractDependencies` can be fused with any other phase). It would be interesting to see if we could just run it in a separate thread.

src/dotty/tools/dotc/sbt/ThunkHolder.scala
DarkDimius: > I'm confused by your suggestion, I can't have an overloaded method where the only difference is that the argument is T or => T, they would conflict with each other.

Indeed. I was assuming that we can follow the code pattern in [Ensuring](https://github.com/scala/scala/blob/2.12.x/src/library/scala/Predef.scala#L315-L317). But without an argument this would indeed lead to ambiguous overload. 

>strictLzy

`strict2lzy` would be even easier to understand.

src/dotty/tools/dotc/sbt/ExtractDependencies.scala
DarkDimius: If the intended behaviour is special treatment of `TypeBounds` amoung all the  `TypeProxy` classes than let's make it explicit. At the moment it's implicitly hidden inside default case.
```scala

case tp: TypeBounds => 
  traverse(tp.hi); 
  traverse(tp.lo);
case tp: TypeProxy =>
  traverse(tp.underlying)
```

smarter: I also had to redefine `SuperType` and `RefinedType`, I'm not convinced that copy-pasting behavior is better than relying on the default case, but I'll do the change anyway

smarter: I'm reverting that change and going back to the previous version, it turns out that adding a case on all `TypeProxy` is dangerous because a `RefinedThis` is a `TypeProxy` and going to its underlying values results in infinite loops. I could add a special case for that but then we'll have even more special cases than the original version, even though the original version has some behavior which is implicit it's at least consistent with `NamedPartsAccumulator`

liufengyun:
We have a  big improvement in performance, I guess it's due to incremental compilation:

https://d-d.me/tnc/dotty/web/

smarter:
That's very weird, the incremental compilation cannot possibily be used by these tests, they're not set up to be run through sbt and the dotty-bridge

Blaisorblade:
> ExtractAPI runs right after PostTyper (so that SuperAccessors are part of the API).

@smarter Can you confirm this prevents issues like https://github.com/sbt/sbt/issues/466, as I'm thinking?

To summarize: nowadays adding super calls to a `trait` requires recompiling implementing classes/objects, but SBT does not manage, because `superaccessors` is ignored by SBT (in https://github.com/sbt/sbt/issues/466#issuecomment-5918336 I wondered whether that could be fixed). It seems your design prevents such problems?

PR #1245, opened 2016-05-03
eliminate self symbol in Template and ClassInfo

### LINKED ISSUE #1239: Eliminate self-symbols from trees & types
Following discussion with Martin, it seems that self-symbols are not needed after typer.
If this is true, it makes sense to create a Mini-Phase&InfoTransformer that would eliminate self symbols both from trees and from types.
It it's false, it would be great to know how are they used.

liufengyun:
Fix #1239 , review @DarkDimius . Thanks.

As this transform is simple, I didn't bother creating a new transform, but integrated it in `FirstTransform`.


DarkDimius:
LGTM, thanks!

PR #1246, opened 2016-05-04
Multiversal Equality

odersky:
The first 4 commits were taken from the previous PR to add mutliversal equals. Review by @smarter.


src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -439,7 +439,19 @@ trait Implicits { self: Typer =>
           result
         case result: AmbiguousImplicits =>
           val deepPt = pt.deepenProto
-          if (deepPt ne pt) inferImplicit(deepPt, argument, pos) else result
+          if (deepPt ne pt) inferImplicit(deepPt, argument, pos)
+          else if (ctx.scala2Mode && !ctx.mode.is(Mode.OldOverloadingResolution)) {
+            inferImplicit(pt, argument, pos)(ctx.addMode(Mode.OldOverloadingResolution)) match {
+              case altResult: SearchSuccess =>
+                ctx.migrationWarning(
+                  s"According to new implicit resolution rules, this will be ambiguous:\n ${result.explanation}",
```
odersky: There was one test that failed under new rules: run/t2030.scala. It was moved to pos-scala2.

src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -361,7 +361,7 @@ trait ImplicitRunInfo { self: RunInfo =>
         computeIScope(cacheResult = false)
       else implicitScopeCache get tp match {
         case Some(is) => is
-        case None => computeIScope(cacheResult = true)
+        case None => computeIScope(cacheResult = seen.isEmpty)
```
smarter: This greatly reduces the effectiveness of the cache since `seen` is almost never empty. I'm OK with merging this now because I have an old WIP branch where I try to fix this and other implicit issues that I will rebase and finish once I'm done with the incremental compilation stuff: https://github.com/dotty-staging/dotty/commits/fix/implicit-caching

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -915,13 +915,54 @@ trait Applications extends Compatibility { self: Typer =>
...
+     *  Intuitively `<:s` means subtyping `<:`, except that all top-level arguments
+     *  to contravariant parameters are compared as if they were covariant. E.g. given class
+     *
+     *     class Cmp[-X]
+     *
+     *  `Cmp[T] <:s Cmp[U]` if `T <: U`. On the other hand, nested occurrences
+     *  of parameters are not affected.
+     *  So `T <: U` would imply `List[Cmp[U]] <:s List[Cmp[T]]`, as usual.
+     *
+     *  This relation might seem strange, but it models closely what happens for methods.
```
smarter: When you say "for methods" you mean in the context of implicit resolution, right? It would be useful to make this explicit, maybe even mention what part of the spec this refers to.

src/scala/Eq.scala
```diff
@@ -0,0 +1,17 @@
+package scala
+
+/** A marker class indicating that values of kind `T` can be compared. */
+class Eq[-L, -R]
```
smarter: The comments for `Eq` and `eqAny` below need to be updated to mention `L` and `R` instead of `T`.

odersky:
This implements #1247

src/scala/Eq.scala
```diff
@@ -0,0 +1,23 @@
+package scala
+
+import annotation.implicitNotFound
+
+/** A marker trait indicating that values of kind `T` can be compared to values of type `U`. */
+@implicitNotFound("Values of types ${L} and ${R} cannot be compared with == or !=")
+trait Eq[-L, -R]
```
smarter: I think this trait should be `sealed` since the only instance should be the object `Eq`

odersky: Agreed.

src/scala/Eq.scala
```diff
@@ -0,0 +1,23 @@
+package scala
+
+import annotation.implicitNotFound
+
+/** A marker trait indicating that values of kind `T` can be compared to values of type `U`. */
```
smarter: > values of kind `T` can be compared to values of type `U`

I'd have written:
> values of type `L` can be compared to values of type `R`

smarter:
Can we change the signature of `Any#==` seen by the user to be:
```scala
  final def ==[T](that: T)(implicit _eq: Eq[this.type, T]): Boolean = this.equals(that)
```
?

smarter:
Actually we might not even need to make `==` polymorphic if my understanding of dependent method types is correct:
```scala
  final def ==(that: Any)(implicit _eq: Eq[this.type, that.type]): Boolean =
    this.equals(that)
```

odersky:
@smarter 

> Can we change the signature of Any#== seen by the user to be:
>  final def ==[T](that: T)(implicit _eq: Eq[this.type, T]): Boolean = this.equals(that)

This would be complicated. The problem is that Any#== calls are generated by the pattern matcher, but we cannot do an implicit search then. So if we do it (and I am sure why we should!) then it can only be done during Typer, afterwards we'd have to change the signature. Does not look very appealing to me.

odersky:
Ready to merge?

smarter:
I'll try to review this more deeply today or tomorrow

smarter:
> There was one test that failed under new rules: run/t2030.scala. It was moved to pos-scala2.

The two possible implicits are `scala.collection.immutable.SortedSet.newCanBuildFrom` and `scala.collection.immutable.TreeSet.newCanBuildFrom`, previously the first one was chosen, but now they're ambiguous. I was curious about what was going on here so I minimized it and explained it, this could be added as a testcase:
```scala
// Minimized from t2030.scala

class Both[-A, +B]

trait Factory[A] {
  implicit def make: Both[A, A] = new Both[A, A]
}

trait X
object X extends Factory[X] {
  override implicit def make: Both[X, X] = super.make
}

class Y extends X
object Y extends Factory[Y] {
  // override implicit def make: Both[Y, Y] = super.make
}

object Test {
  def get(implicit ev: Both[Y, X]) = ev

  // There are two possible implicits here: X.make and Y.make, neither are
  // subtype of each other, so who wins?
  // - Under the old scheme it's X.make because `isAsGood` sees that X.make is defined
  // in X whereas Y.make is defined in Factory
  // - Under the new scheme it's ambiguous because we replace contravariance by covariance
  // in top-level type parameters so Y.make is treated as a subtype of X.make
  // In both schemes we can get Y.make to win by uncommenting the override for make in Y
  // (Y wins against X because `isDerived` also considers the subtyping relationships
  // of companion classes)
  get
}
```

This means that we can fix the issue in the standard library by adding an override of `newCanBuildFrom` in `TreeSet`, as a bonus this means that `res1` in `t2030.scala` gets the inferred type `TreeSet[Int]` instead of `SortedSet[Int]`.

felixmulder:
@odersky: 3c18a2a has not been properly fixed, if you want this to be merged quickly - I'd revert it.

Also, I'm investigating a more thorough fix using a MiniPhase that will be more of a general fix.

odersky:
@felixmulder Why do you need a mini-phase? I am thinking of a simple fix in typer.

PR #1249, opened 2016-05-10
Evaluate annotations before completing tree of definitions

odersky:
Motive: That way we can identify annotation macros without special
name resolution rules.

This was surprisingly easy.

Review by @xeno-by

### DIFF
--- a/src/strawman/collections/CollectionStrawMan4.scala
+++ b/src/strawman/collections/CollectionStrawMan4.scala
@@ -12,6 +12,8 @@ import annotation.tailrec
  *  strengths and weaknesses of different collection architectures.
  *
  *  For a test file, see tests/run/CollectionTests.scala.
+ *
+ *  Strawman4 is like strawman1, but built over views instead of by-name iterators
  */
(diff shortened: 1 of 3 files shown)


src/strawman/collections/CollectionStrawMan4.scala
```diff
@@ -12,6 +12,8 @@ import annotation.tailrec
  *  strengths and weaknesses of different collection architectures.
  *
  *  For a test file, see tests/run/CollectionTests.scala.
+ *
+ *  Strawman4 is like strawman1, but built over views instead of by-name iterators
```
xeno-by: Is this commit here intentionally?

xeno-by:
I'll take a look on Mon/Tue 16/17 May after my arrival back to Europe. Looking forward to understanding better how Dotty works!

src/strawman/collections/CollectionStrawMan4.scala
odersky: Yes, I just threw it in to save time.

Issue #1208, opened 2016-04-06
Add a `dotty` runner script that is able to run dotty-compiled class with `main`


DarkDimius:

smarter:
Fixed by https://github.com/lampepfl/dotty/pull/1250

PR #1250, opened 2016-05-12
Add dotty runner script able to run dotty-compiled class with `main`

felixmulder:
Addresses #1208

Review: @DarkDimius / @smarter


bin/dotty
```diff
@@ -0,0 +1,25 @@
...
+DOTTY_ROOT="`( cd \"$DOTTY_ROOT\" && pwd )`/.."  # absolute
+
+# CLASS_PATH is derived from the DOTTY_ROOT
+CLASS_PATH="-J-Xbootclasspath/a:.:$DOTTY_ROOT/target/scala-2.11/classes/"
+
+function runMain {
+    scala $CLASS_PATH $@
+}
+
+if [ -z "$1" ]; then
```
smarter: I think no arguments should launch the REPL.

felixmulder: Good idea! Will update the PR

bin/dotty
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Try to autodetect real location of the script
+DOTTY_ROOT="`readlink \"$0\"`"             #  relative, symbolic links resolved
```
smarter: `DOTTY_ROOT="$(readlink "$0")"` (no need to escape `"` inside `$(...)`)

felixmulder: Changed it in the ´dotc´ script as well

DarkDimius:
I'd better not call this script `dotty`. `dotty` is part of graphviz library that is very common in unix environments.

felixmulder:
@DarkDimius, sure - if it's put on `$PATH` that'd be bad. Suggestions on the name? `dotr`?

DarkDimius:
I'm using the script `dottyr` on my machine.

felixmulder:
As discussed in the meeting, change the name of the script to `dotr`

DarkDimius:
needs rebasing, and a patch to silence `cd` in `dotr`.
Otherwise LGTM.

PR #1251, opened 2016-05-12
Fix stdout redirect for REPL's println

felixmulder:
When printing in the REPL via `println`, the output would end up on the
same line, since stdout had not been redirected. This commit remedies
that.

It also adds syntax highlighting to result types.

Review: @smarter

### DIFF
--- a/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -94,6 +94,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val sourceReader = StringSetting("-Xsource-reader", "classname", "Specify a custom method for reading source files.", "")
   val XnoValueClasses = BooleanSetting("-Xno-value-classes", "Do not use value classes. Helps debugging.")
   val XreplLineWidth = IntSetting("-Xrepl-line-width", "Maximial number of columns per line for REPL output", 390)
+  val XreplNoColor = BooleanSetting("-Xrepl-no-color", "Removes syntax highlighting in REPL")
   val XoldPatmat = BooleanSetting("-Xoldpatmat", "Use the pre-2.10 pattern matcher. Otherwise, the 'virtualizing' pattern matcher is used in 2.10.")
   val XnoPatmatAnalysis = BooleanSetting("-Xno-patmat-analysis", "Don't perform exhaustivity/unreachability analysis. Also, ignore @switch annotation.")
   val XfullLubs = BooleanSetting("-Xfull-lubs", "Retains pre 2.10 behavior of less aggressive truncation of least upper bounds.")
--- a/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -28,7 +28,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val help = BooleanSetting("-help", "Print a synopsis of standard options")
   val nowarn = BooleanSetting("-nowarn", "Generate no warnings.")
   val print = BooleanSetting("-print", "Print program with Scala-specific features removed.")
-  val color = ChoiceSetting("-color", "mode", "Colored output", List("always", "never", "auto"), "auto")
+  val color = ChoiceSetting("-color", "mode", "Colored output", List("always", "never"/*, "auto"*/), "always"/* "auto"*/)
   val target = ChoiceSetting("-target", "target", "Target platform for object files. All JVM 1.5 targets are deprecated.",
     List("jvm-1.5", "jvm-1.5-fjbg", "jvm-1.5-asm", "jvm-1.6", "jvm-1.7", "jvm-1.8", "msil"),
     "jvm-1.8")
--- a/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -28,6 +28,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val help = BooleanSetting("-help", "Print a synopsis of standard options")
   val nowarn = BooleanSetting("-nowarn", "Generate no warnings.")
   val print = BooleanSetting("-print", "Print program with Scala-specific features removed.")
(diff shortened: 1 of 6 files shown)


smarter:
This doesn't fix the followings which work fine with the scala REPL:
```scala
> Console.err.println("hi")
> System.out.println("hi")
> System.err.println("hi")
```

test/test/TestREPL.scala
```diff
@@ -38,7 +38,8 @@ class TestREPL(script: String) extends REPL {
     out.close()
     val printed = out.toString
     val transcript = printed.drop(printed.indexOf(config.prompt))
-    if (transcript.toString != script) {
+    val transcriptNoColors = transcript.toString.replaceAll("\u001B\\[[;\\d]*m", "")
```
smarter: What about having a config option to disable all colors instead?

smarter: the scala REPL has `-Dscala.color`: https://github.com/scala/scala/blob/99a82be91cbb85239f70508f6695c6b21fd3558c/src/repl/scala/tools/nsc/interpreter/ReplProps.scala#L20

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -392,27 +413,26 @@ class CompilingInterpreter(out: PrintWriter, ictx: Context) extends Compiler wit
...
-          (printList :+ new String(res), true)
+          (printList :+ res, true)
         }
-      } catch {
-        case NonFatal(ex) =>
-          def cause(ex: Throwable): Throwable =
-            if (ex.getCause eq null) ex else cause(ex.getCause)
-          val orig = cause(ex)
-          (stringFrom(str => orig.printStackTrace(str)) :: Nil, false)
+      } catch { case NonFatal(ex) =>
```
smarter: Don't put a case on the same line as `catch`

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -94,6 +94,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val sourceReader = StringSetting("-Xsource-reader", "classname", "Specify a custom method for reading source files.", "")
   val XnoValueClasses = BooleanSetting("-Xno-value-classes", "Do not use value classes. Helps debugging.")
   val XreplLineWidth = IntSetting("-Xrepl-line-width", "Maximial number of columns per line for REPL output", 390)
+  val XreplNoColor = BooleanSetting("-Xrepl-no-color", "Removes syntax highlighting in REPL")
```
felixmulder: @smarter: It felt like the precedence was to name the option `-Xrepl-no-color` as opposed to the `-Dscala.color` in scalac

(Other suggested changes in place)

smarter: The name doesn't really matter, but I think you should phrase the option positively (`-Xrepl-color`, or just `-color` since this isn't an "advanced" option) and you really want three possible values for color: `never`,`auto` and `always`, `auto` means that we try to guess if we're in a terminal (as opposed to redirected to a file) and only if that's the case do we enable colors, this is how color is specified in many cli tools (gcc `-fdiagnostics-color`: https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Language-Independent-Options.html, git, ...). But this isn't needed for this PR of course.

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -28,6 +28,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val help = BooleanSetting("-help", "Print a synopsis of standard options")
   val nowarn = BooleanSetting("-nowarn", "Generate no warnings.")
   val print = BooleanSetting("-print", "Print program with Scala-specific features removed.")
+  val color = ChoiceSetting("-color", "mode", "Colored output", List("always", "never", "auto"), "auto")
```
felixmulder: @smarter: done. For now it defaults to "auto" and the repl will be colored if the setting is set to "auto" or "always".

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -383,24 +387,54 @@ class CompilingInterpreter(out: PrintWriter, ictx: Context) extends Compiler wit
...
+      try {
+        Console.withOut(os)(Console.withErr(os)(op(os)))
+      } finally {
+        System.setOut(oldOut)
+        System.setErr(oldErr)
+      }
+    }
+
     /** load and run the code using reflection.
      *  @return  A pair consisting of the run's result as a string, and
```
smarter: That comment is no longer accurate, the first element of the pair is no longer a string

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -383,24 +387,54 @@ class CompilingInterpreter(out: PrintWriter, ictx: Context) extends Compiler wit
...
       val interpreterResultObject: Class[_] =
         Class.forName(resultObjectName, true, classLoader)
-      val resultValMethod: java.lang.reflect.Method =
+      val valMethodRes: java.lang.reflect.Method =
         interpreterResultObject.getMethod("result")
       try {
-        (resultValMethod.invoke(interpreterResultObject).toString, true)
+        withOutput(new ByteOutputStream) { ps =>
+          val rawRes    = valMethodRes.invoke(interpreterResultObject).toString
+          val res       =
```
smarter: I'm not a fan of vertical alignment in general but here it seems particularly useless, there's nothing to align on the right of that `=`

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -383,24 +387,54 @@ class CompilingInterpreter(out: PrintWriter, ictx: Context) extends Compiler wit
...
+            if (List("auto", "always").contains(ictx.settings.color.value))
```
smarter: Code duplication with `AmmoniteReader.scala`, sounds like you should have a `ictx.useColors` method or something.

DarkDimius:
"auto" does not do anything. It is simply lying to user that it's implemented, leading to bad user experience.

I propose to comment it out. 
Otherwise LGTM

Issue #1196, opened 2016-03-31
Implement infrastructure to check emitted bytecode


DarkDimius:
Both those pull request:
https://github.com/lampepfl/dotty/pull/1061
https://github.com/lampepfl/dotty/pull/1188

restructure code with expectation that backend would emit different bytecode instruction. 
Currently our CI lacks a way to check if it actually works.

felixmulder:
I'm currently exploring this using a similar approach as in scalac

PR #1256, opened 2016-05-17
Add bytecode checking infrastructure

felixmulder:
Addresses: #1196 

Planning to add checking of successful switches to this PR before merge.

### DIFF
--- a/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -307,10 +307,15 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
       // TODO Deal with guards?
 
       def isSwitchableType(tpe: Type): Boolean = {
-        (tpe isRef defn.IntClass) ||
-        (tpe isRef defn.ByteClass) ||
-        (tpe isRef defn.ShortClass) ||
-        (tpe isRef defn.CharClass)
+        val actualTpe = tpe match {
+          case t @ AnnotatedType(inner, _) => inner
+          case x => x
+        }
+
+        (actualTpe isRef defn.IntClass) ||
+        (actualTpe isRef defn.ByteClass) ||
+        (actualTpe isRef defn.ShortClass) ||
+        (actualTpe isRef defn.CharClass)
       }
 
       object IntEqualityTestTreeMaker {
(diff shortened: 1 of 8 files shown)


felixmulder:
When trying to verify switches, I discovered: https://github.com/lampepfl/dotty/issues/1258

felixmulder:
Discovered errors in creating multidimensional arrays: https://github.com/lampepfl/dotty/issues/1259

test/test/DottyBytecodeTests.scala
```diff
@@ -0,0 +1,109 @@
...
+      val hadCorrectInstr =
+        instructionsFromMethod(method)
+        .collect { case x @ NewArray(op, _, dims) if op == 197 && dims == 2 => x }
+        .length > 0
+
+      assert(hadCorrectInstr,
+             "Did not contain \"multianewarray\" instruction in:\n" +
+             instructionsFromMethod(method).mkString("\n"))
+    }
+  }
```
felixmulder: @sjrd: any other test you'd like here?

sjrd: Well you should *also* have `Array.ofDim[Int](2)`.
Also the more "primitive" form `new Array[Int](2)` should give the same.

Ideally you should also test with other element types. A typical selection is `Int`, `Unit` and `String`, because this set exploits the 3 main implementation strategies (primitive, special-case of `Unit`, and reference type).

felixmulder: Great, thanks. I'll have it in by tomorrow morning... Sleepytime

felixmulder:
This is ready for review: @DarkDimius, @smarter

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -307,10 +307,15 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
       // TODO Deal with guards?
 
       def isSwitchableType(tpe: Type): Boolean = {
-        (tpe isRef defn.IntClass) ||
-        (tpe isRef defn.ByteClass) ||
-        (tpe isRef defn.ShortClass) ||
-        (tpe isRef defn.CharClass)
+        val actualTpe = tpe match {
+          case t @ AnnotatedType(inner, _) => inner
```
smarter: What if the type has several annotations? Use `stripAnnots` instead (in fact, it might make sense to use `stripAnnots` in `isRef`, but you'd have to check if that's correct everywhere `isRef` is used)

felixmulder: Agreed, I ran the tests and looked at the call sites and we should be fine with putting it in `isRef`

DarkDimius:
LGTM

PR #1261, opened 2016-05-18
[Deprecated] exhaustivity & redundancy check for pattern matching

liufengyun:
This is a work in process, it basically works. The test examples are currently put under `tests/patmat/`.

Following things need to be done:

- [x] test framework for checking warnings (use check files for now)
- [x] handling of constants and backquote pattern in patterns
- [x] more testing & bug fix
  - [x] build test set from scala issues
  - [x] fix test set
- java enums (SI-8700, SI-2442)
  - [x] java enums from classfiles
  - [x] java enums from source files
- code cleanup:
  - [x] put all patmat-related code into the `transform.patmat` package
  - [x] move test examples from `examples/` to `tests/neg/patmat/`
- [x] disable checking for top-level matching in partial functions
- [x] suppress unreachable warning in for-comprehension
- [x] disable checking for try/catch
- [x] false warning about Nil
- [ ] auto tupling
- [x] special handling of Array
- [x] bounded type members ([SI-9657](https://issues.scala-lang.org/browse/SI-9657))

A summary is [below](https://github.com/lampepfl/dotty/pull/1261#issuecomment-221736473).


smarter:
Not super important, but if possible it's nice to put the tests in the same commit than the changes that make the tests pass.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+        Or(
+          ss1.zip(ss2).map((minus _).tupled).zip(0 to ss2.length - 1).map {
+            case (ri, i) => Kon(t1, ss1.updated(i, ri))
+          })
+  }
+}
+
+// scala implementation of space logic
+class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
+  def debug(s: String): Unit = {
```
smarter: There's `ctx.debuglog`

smarter: But you might want to add a `Printer` to https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/config/Printers.scala and use that instead, this way you can take advantage of `ctx.traceIndented` too.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+    debug(s"====signature of $t: $sign")
+    sign
+  }
+
+  def getPartitions(t: Type): List[Type] = t match {
+    case OrType(t1, t2) => List(t1, t2)
+    case _ =>
+      val children = t.classSymbol.children.map { c =>
+        if (c is Flags.ModuleClass)
+          c.info.asInstanceOf[ClassInfo].selfType
```
smarter: `c.asClass.classInfo`, even better if you could make `ClassDenotations#children` a set of `ClassSymbol` instead of `Symbol` then you could do `c.classInfo`

smarter:
For consistency, we tend to denotes type arguments with `tp: Type`, not `t: Type`

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  // does the type t have @unchecked annotation
+  def skipCheck(t: Type): Boolean = t match {
+    case AnnotatedType(t, annot) => ctx.definitions.UncheckedAnnot == annot.symbol
+    case _ => false
+  }
+
+  // widen a type and eliminate anonymous class
+  private def widen(t: Type): Type = t.widen match {
```
smarter: Looks like this method is adapted from `PatternMatcher#elimAnonymousClass`, instead of copying code it would be better to share it between this and the PatternMatcher, you could for example put `elimAnonymousClass` in `object PatternMatcher`

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+
+import core.Types._
+import core.Contexts._
+import core.Flags
+import ast.Trees._
+import ast.tpd
+import core.Decorators._
+
+
+/** space logic for checking exhaustivity and unreachability of
```
smarter: Please document in more details how the algorithm works.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+        case s::Nil => s
+        case Nil => Empty
+      }
+    case Typ(t) =>
+      if (canDecompose(t) && getPartitions(t).isEmpty) Empty
+      else space
+    case _ => space
+  }
+
+  // is a subspace of b, equivalent to `a - b == Empty`, but faster
```
smarter: Use `/** */` style comments to properly document methods.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+case class Kon(t: Type, params: List[Space]) extends Space
+case class Or(spaces: List[Space]) extends Space
+
+// abstract space logic
+trait SpaceLogic {
+  def isSubType(t1: Type, t2: Type): Boolean
+  def isEqualType(t1: Type, t2: Type): Boolean
+  def isCaseClass(t: Type): Boolean
+
+  // Can a type be decompsed? i.e. all values of the type can be covered
```
smarter: typo: decomposed

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+
+    debug(s"====signature of $t: $sign")
+    sign
+  }
+
+  def getPartitions(t: Type): List[Type] = t match {
+    case OrType(t1, t2) => List(t1, t2)
+    case _ =>
+      val children = t.classSymbol.children.map { c =>
+        if (c is Flags.ModuleClass)
```
smarter: I think it's nicer to import `Flags._` so that you can do `c.is(ModuleClass)` (also our current coding style discourage the use of infix notation, prefer `foo.is(bar)` to `foo is bar`)

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  // find the first redundant pattern and stop
+  def redundancy(_match: tpd.Match): Unit = {
+    val Match(sel, cases) = _match
+    val selTyp = widen(sel.tpe)
+
+    // starts from the second, the first can't be redundant
+    var i = 1
+    while(i < cases.length) {
```
smarter: Why not `cases.foreach` ?

src/dotty/tools/dotc/transform/Space.scala
liufengyun: I'll eventually remove the `debug` code when everything works. I tried `ctx.debuglog`, but it seems I can't make it print in the console even if I used the flag `-Ydebug -Ylog:patternmatcher`.

src/dotty/tools/dotc/transform/Space.scala
liufengyun: I avoid `cases.foreach` here because I need to break the loop in the middle. I'm not sure how to do that with `foreach`.

liufengyun:
@smarter thanks a lot for many good points, I'll address them in following commits.

src/dotty/tools/dotc/transform/Space.scala
smarter: You can try using `return`

smarter: But do you really need to break? What happens if you don't?

liufengyun: Basically I want to stop the redundancy check after the first redundant pattern is found -- that saves computation for subsequent patterns. Of course, we could always check redundancy for all patterns if we don't break.

smarter: If it's not a significant cost I would just do all the checks.

liufengyun: The cost depends on the program, but for normal programs the cost should be negligible. I'll do as you suggested. Thanks.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+      else
+        Or(
+          ss1.zip(ss2).map((minus _).tupled).zip(0 to ss2.length - 1).map {
+            case (ri, i) => Kon(t1, ss1.updated(i, ri))
+          })
+  }
+}
+
+// scala implementation of space logic
+class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
```
smarter: do `import tpd._` here so that you can use `Tree` instead of `tpd.Tree` below, a lot of phases do this.

smarter:
In https://gist.github.com/liufengyun/66bf0c1bc44cdad12e45dd6db689e9b3 you said:
> The space covered by a pattern p is defined as follows:
> ...
> if p is pat guard, space(pat)

I think this is not what we want, unless the guard evaluates to true in all cases, the space covered should just be `Empty`, you should always have a case without a guard if you want the pattern match to be considered exhaustive. This is not the current scalac behavior but it is the one that it will soon have: https://github.com/scala/scala/pull/4929

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+import core.Decorators._
+
+
+/** space logic for checking exhaustivity and unreachability of
+  * pattern matching.
+  */
+
+
+// space definition
+sealed trait Space
```
smarter: This is pretty cool: you're using the exhaustiveness algorithm to test its own exhaustivity! :collision:

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,316 @@
...
+object Empty extends Space
```
smarter: `case object`

liufengyun: Good catch! Does the compiler do something special to case objects?

smarter: Not much, they implement `Product0`, they have a readable `toString`, not sure what else.

liufengyun:
@smarter Guards can be very tricky, generally we can do thing with them.

However, some ad-hoc heuristics is possible, e.g. the case you mentioned in scalac. But the ad-hoc strategy is very limited and may cause problems:

``` Scala
def foo(l: List[int]): Boolean = l match {
  case x if x.length > 0 => true
  case None => false
}
```

smarter:
@liufengyun : Yes, you can do tricky things with them, but our response should be: don't do that :)
Your code can easily be rewritten as:
```scala
def foo(l: List[int]): Boolean = l match {
  case Nil => false
  case _ => true
}
```
People should either rewrite their code or add `@unchecked`

smarter:
I left some comments on https://gist.github.com/liufengyun/66bf0c1bc44cdad12e45dd6db689e9b3

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -14,35 +14,41 @@ import core.Decorators._
...
 case class Kon(t: Type, params: List[Space]) extends Space
 case class Or(spaces: List[Space]) extends Space
 
-// abstract space logic
+/** abstract space logic */
 trait SpaceLogic {
-  def isSubType(t1: Type, t2: Type): Boolean
-  def isEqualType(t1: Type, t2: Type): Boolean
-  def isCaseClass(t: Type): Boolean
+  /** is `tp1` a subtype of `tp2` */
```
smarter: Sorry for being pedantic, but please start comments with a capital letter :). And put `?` at the end if it's a question.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -14,35 +14,41 @@ import core.Decorators._
...
 
-  // return parameters types of case class
-  def getSignature(t: Type): List[Type]
+  /** is `tp` a case class */
+  def isCaseClass(tp: Type): Boolean
 
-  // get components of decomposable types
-  def getPartitions(t: Type): List[Type]
+  /** is the type `tp` decompsable? i.e. all values of the type can be covered
+    * by its decomposed types.
```
smarter: It's weird but the comment style in dotty is different from the normal style, you should write:
```scala
/** Line 1
 *  Line 2
 *  Line 3
 */
Not:
```scala
/** Line 1
  * Line 2
  * Line 3
  */
```
See `SymDenotations` for an example of mostly well-documented code.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -54,221 +60,217 @@ trait SpaceLogic {
...
+    case (Kon(tp1, ss1), Kon(tp2, ss2)) =>
+      if (!isEqualType(tp1, tp2)) Empty
       else if (ss1.zip(ss2).exists(p => simplify(intersect(p._1, p._2)) == Empty)) Empty
       else
-        Kon(t1, ss1.zip(ss2).map((intersect _).tupled))
+        Kon(tp1, ss1.zip(ss2).map((intersect _).tupled))
   }
 
-  //  the space of a not covered by b
+  /**  the space of a not covered by b */
```
smarter: Put only one space betwen `/**` and the start of the comment message.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -288,14 +290,12 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-  // find the first redundant pattern and stop
-  def redundancy(_match: tpd.Match): Unit = {
+  def redundancy(_match: Match): Unit = {
     val Match(sel, cases) = _match
     val selTyp = widen(sel.tpe)
 
     // starts from the second, the first can't be redundant
-    var i = 1
-    while(i < cases.length) {
+    (1 until cases.length).foreach { i =>
```
smarter: I think you can just do `cases.foreach { c => ... }`

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,313 @@
...
+        case s::Nil => s
+        case Nil => Empty
+      }
+    case Typ(tp) =>
+      if (canDecompose(tp) && getPartitions(tp).isEmpty) Empty
+      else space
+    case _ => space
+  }
+
+  /** is a subspace of b, equivalent to `a - b == Empty`, but faster */
```
smarter: Missing the word `a`: It should be ``Is `a` a subtype of `b` ? This is equivalent to ...`

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,313 @@
...
+/** scala implementation of space logic */
+class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
+  import tpd._
+
+  def debug(s: String): Unit = {
+    if (ctx.debug) println(s)
+  }
+
+  /** if roundUp, approximate extractors to its type, else approximate extractors to Empty */
+  def project(pat: Tree)(implicit ctx: Context, roundUp: Boolean = true): Space = pat match {
```
smarter: Why is `roundUp` an implicit parameter?

liufengyun:
@smarter I just responded to your comments on the algorithm.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,313 @@
...
+    case Ident(_) => Typ(pat.tpe)
+    case Select(_, _) => Typ(pat.tpe)
+    case Alternative(trees) => Or(trees.map(project _))
+    case Bind(_, pat) => project(pat)
+    case UnApply(_, _, pats) =>
+      if (pat.tpe.classSymbol.is(CaseClass))
+        Kon(pat.tpe, pats.map(project _))
+      else if (roundUp) Typ(pat.tpe)
+      else Empty
+    case Typed(pat@UnApply(_, _, _), _) => project(pat)
```
smarter: The style we use is to put spaces around `@`: `pat @ UnApply(_, _, _)`

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,313 @@
...
+
+  def getSignature(tp: Type): List[Type] = {
+    val ktor = tp.classSymbol.primaryConstructor.info
+
+    def typeParams(tp: Type): List[Type] = tp match {
+      case t: RefinedType => t.refinedInfo.asInstanceOf[TypeAlias].alias :: typeParams(t.parent)
+      case _ => Nil
+    }
+
+    val sign = (tp match {
```
smarter: Consider splitting this into two vals:
```scala
val something = tp match { ...
val sign = something.firstParamTypes. ...

File: src/dotty/tools/dotc/transform/Space.scala [hunk 20]
```diff
@@ -0,0 +1,313 @@
...
+
+  /** is the type `tp` decompsable? i.e. all values of the type can be covered
+    * by its decomposed types.
+    *
+    * abstract sealed class and OrType can be decomposed
+    */
+  def canDecompose(tp: Type): Boolean
+
+  /** return parameters types of the case class `tp` */
+  def getSignature(tp: Type): List[Type]
```
smarter: This is not Java, we don't put `get` in front of our method names for no reason :). I think `def signature(tp: Type)` would be fine.

src/dotty/tools/dotc/transform/Space.scala
liufengyun: That's fine, I'm very grateful to your detailed feedback, thanks a lot :)

src/dotty/tools/dotc/transform/Space.scala
liufengyun: I need an `i` here to take the first `i` cases to check redundancy.

smarter: Ah, I see, you could do `cases.zipWithIndex.foreach { (c, i) => ... }`. There might be a better way but I can't think of it.

src/dotty/tools/dotc/transform/Space.scala
liufengyun: It's made implicit so that we don't need to pass it around in recursive calls.

smarter: You don't need it to be implicit for that, a default value is enough:
```scala
def project(pat: Tree, roundUp: Boolean = true)(implicit ctx: Context)
```

liufengyun: It seems we do need to pass default parameters around, in case it's a user-supplied value.

smarter: Ah, I see what you mean. While it's true that your solution work, I would advise you against it: it doesn't really makes sense to have an implicit `Boolean` in scope, its meaning depends on where it's used and your pattern stops working as soon as you add another `Boolean` parameter to the method. Even though this requires more typing, I would advise you to pass the value of `roundUp` to every recursive call. This is what we do in dotty usualy, for example: https://github.com/lampepfl/dotty/blob/134ad7a6a172fec97dc438dd0aff3766a0f0944a/src/dotty/tools/dotc/typer/ProtoTypes.scala#L383

src/dotty/tools/dotc/transform/Space.scala
liufengyun: `zipWithIndex` actually is a little inconvenient here, as I want to start from the second element.

src/dotty/tools/dotc/transform/Space.scala
liufengyun: Fair enough, I've renamed `getSignature` and `getPartitions` in the latest commit. The boolean implicit is also addressed in the commit.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -54,221 +60,219 @@ trait SpaceLogic {
...
+  import tpd._
+
   def debug(s: String): Unit = {
     if (ctx.debug) println(s)
   }
 
-  // if roundUp, approximate extractors to its type
-  // else approximate extractors to Empty
-  def project(pat: tpd.Tree)(implicit ctx: Context, roundUp: Boolean = true): Space = pat match {
+  /** If roundUp, approximate extractors to its type, else approximate extractors to Empty */
```
smarter: ``If `roundUp` is true, approximate an extractor to its type, otherwise approximate it to `Empty` ``

smarter: Also the first line of this comment should explain what this method does, I guess it's something like ``Return the space that represents the pattern `pat` ``

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -54,221 +60,219 @@ trait SpaceLogic {
...
+      debug(s"child of ${tp.show}: ${children.map(_.show).mkString(", ")}")
 
       children
   }
 
-  // parent is specialized, specialize child as well
-  def specializeChild(p: Type, c: Type): Type = p match {
-    case p: RefinedType => p.wrapIfMember(specializeChild(p.parent, c))
-    case _ => c
+  /** Parent is specialized, specialize child as well */
```
smarter: What's the meaning of specialized here? This should be documented.

src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -90,9 +90,14 @@ object RefChecks {
         if (pclazz.is(Sealed) && pclazz.associatedFile != cls.associatedFile)
           ctx.error(d"cannot extend sealed $pclazz in different compilation unit", cls.pos)
         checkSelfConforms(parent, "illegal inheritance", "parent")
+
+        // Maintain children table for sealed class to facilitate exhaustivity
+        // and redundancy check of patttern matching
+        if (pclazz.is(Sealed)) pclazz.addChild(cls)
       }
```
liufengyun: @smarter FYI, this seems to be a good place to maintain the table.

smarter: @odersky: Do you think the list of children of a sealed class should be stored in the `ClassSymbol` of the class?

smarter:
The following gives a warning:
```scala
sealed trait Base
case class Foo() extends Base

object Test {
  def foo1b(x: Base) = x match {
    case Foo() =>
  }
}
```
```
It would fail on the following inputs: _: Base
  def foo1b(x: Base) = x match {
                       ^
```
The warning disappears If I replace `sealed trait Base` by `abstract sealed trait Base` but that makes no sense since all traits are abstract by definitions

smarter:
The algorithm does not seem to handle path-dependent types:
```scala
trait Module {
  sealed abstract class Tree
 
  case class LetL() extends Tree
}
 
class Patmat(val module: Module) {
  def patmat(tree: module.Tree) = tree match {
    case module.LetL() => ???
  }
}
```
```
warning: match may not be exhaustive.
It would fail on the following inputs: _: LetL
  def patmat(tree: module.Tree) = tree match {
                                  ^
```

smarter:
In:
```scala
sealed abstract class CL3Literal
case object IntLit extends CL3Literal
case object CharLit extends CL3Literal
case object BooleanLit extends CL3Literal
 
 
sealed abstract class Tree
case class LetL(value: CL3Literal) extends Tree
 
object Test {
  def transform(tree: Tree) : Any = tree match {
    case LetL(CharLit) =>
  }
}
```
The warning is:
```scala
try/t8430.scala:11: warning: match may not be exhaustive.
It would fail on the following inputs: LetL(IntLit, BooleanLit)
  def transform(tree: Tree) : Any = tree match {
                                    ^
```
Which is a bit confusing, it should be: `It would fail on the following inputs: LetL(IntLit), LetL(BooleanLit)`

liufengyun:
Thanks @smarter for reporting. I'll address them after I'm done with the annotations. How the children is stored in annotation is weird and inconvenient for usage:

```
    def makeChild(sym: Symbol)(implicit ctx: Context) =
      deferred(defn.ChildAnnot,
        implicit ctx => New(defn.ChildAnnotType.appliedTo(sym.owner.thisType.select(sym.name, sym)), Nil))
```

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -231,6 +231,15 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
                   ctx.compilationUnit.source.exists &&
                   sym != defn.SourceFileAnnot)
                 sym.addAnnotation(Annotation.makeSourceFile(ctx.compilationUnit.source.file.path))
+
+              for (parent <- sym.asClass.classInfo.classParents) {
```
smarter: The comment on top of `class PostTyper` document eveything that the phase does, you should add something like `(10) Adds Child annotations to all sealed classes`

liufengyun:
@smarter I've made a commit which addressed your feedback f9bc81a.

smarter:
Nice! Some comments on the warning messages, if I have:
```scala
trait Module {
  sealed trait Tree
 
  case class LetL() extends Tree
  case class LetP() extends Tree
}
 
class Patmat(val module: Module) {
  def patmat(tree: module.Tree) = tree match {
    case module.LetL() =>
  }
}
```
Then the warning is:
```scala
warning: match may not be exhaustive.
It would fail on the following inputs: _: LetP
  def patmat(tree: module.Tree) = tree match {
                                  ^
```
This is almost correct but it should be `_: module.LetP` instead of `_: LetP` which is not a valid type. Also it would be nice to display `module.LetP()` instead of `_: module.LetP`.

liufengyun:
Thanks @smarter for the feedback, I've addressed the problem in 5cdacbf, now  it shows `module.LetP()` instead of `_: LetP`.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,385 @@
...
+      ctx.definitions.UnqualifiedOwnerTypes.exists(_.symbol == sym) ||
+      sym == enclosingCls.typeSymbol
+
+    def refine(tp: Type): String = tp match {
+      case tp: RefinedType => refine(tp.parent)
+      case tp: ThisType => refine(tp.tref)
+      case tp: TypeRef if isOmittable(tp.symbol) => ""
+      case tp: TypeRef =>
+        val pre = refine(tp.prefix)
+        if (pre.isEmpty) tp.name.show.stripSuffix("$")
```
smarter: You can use `tp.name.sourceModuleName.show`

smarter:
Looks like your exhaustivity checker is already doing better than scalac in at least one case: it is not affected by https://issues.scala-lang.org/browse/SI-9289 :sparkles:

smarter:
Suggestion that does not need to be implemented for this PR: it would be nice if the redundancy analysis also told you with which case the current case is redundant, C compilers do this with switch cases:
```c
$ cat unreach.c
int main(int argc, char **argv) {
  switch (argc) {
    case 0: // A
      break;
    case 0: // B
      break;
  }
  return 0;
}
$ gcc unreach.c
unreach.c: In function ‘main’:
unreach.c:5:5: error: duplicate case value
     case 0: // B
     ^
unreach.c:3:5: error: previously used here
     case 0: // A
     ^
```

liufengyun:
@smarter that feature in `C` is possible to implement.

It's my pleasure to know that the new algorithm performs better in that test case. I'm almost done with the test framework for checking warnings with `.check` file.

Next I'm going to add all cases from Scala issues to the test set and implement the support for literal patterns and stable identifier patterns.

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -231,6 +231,15 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
...
+                val pclazz = parent.classSymbol
+                if (pclazz.is(Sealed)) {
+                  println(s"add $sym to $pclazz")
+                  pclazz.addAnnotation(Annotation.makeChild(sym))
+                }
+              }
+
```
liufengyun: Anonymous classes caused pickling-related error for `tests/run/t8611b.scala`:

https://gist.github.com/liufengyun/f4aca0853e06f96c9130a0a05bfa0d4f

If I ignore anonymous classes, then it works:

``` Scala
              if (!sym.isAnonymousClass) // ignore anonymous class
                for (parent <- sym.asClass.classInfo.classParents) {
                  val pclazz = parent.classSymbol
                  if (pclazz.is(Sealed)) pclazz.addAnnotation(Annotation.makeChild(sym))
                }
```

The problem is that ignore anonymous classes would make the algorithm silent in case it should report a warning:

``` Scala
sealed trait A
case class B() extends A
case class C() extends A

object M {
  val x = new A {}

  x match {
     case _: B => true
     case _: C => true
  }
}
```

Any ideas, @odersky @smarter ?

smarter: If a sealed trait has anonymous subclasses then it's impossible to pattern match on all subclasses. I think the simplest solution is to disallow anonymous subclasses of sealed traits.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -166,26 +166,34 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
       Empty
   }
 
-  def isSubType(tp1: Type, tp2: Type): Boolean = {
-    val res = tp1 <:< tp2
-    debug(s"$tp1 <: $tp2 ? $res")
-    res
+  /** Is `tp1` a subtype of `tp2`?
+   *
+   *  Ignore type parameters in comparison due to erasure, i.e., Some[Int] <: Some[T]
```
smarter: You might be able to use `TypevarsMissContext` instead (e.g: `tp1.<:<(tp2)(TypevarsMissContext)(ctx.fresh.addMode(Mode.TypevarsMissContext).setExploreTyperState)` instead), but I don't really understand why this is needed in your case

liufengyun: Thanks @smarter. The test case that drives me to make this change is [t6450](https://github.com/dotty-staging/dotty/blob/09e096832d5f6b0e8814bf02525d94d649635210/tests/pos/patmat/t6450.scala).

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,406 @@
...
+      else "_"
+    case Kon(tp, params) =>
+      if (ctx.definitions.isTupleType(tp))
+        "(" + params.map(p => show(p)).mkString(", ") + ")"
+      else if (tp.widen.classSymbol.showFullName == "scala.collection.immutable.::")
+        if (mergeList) params.map(p => show(p, mergeList)).mkString(", ")
+        else "List(" + params.map(p => show(p, true)).mkString(", ") + ")"
+      else
+        showType(tp) + "(" + params.map(p => show(p)).mkString(", ") + ")"
+    case Or(spaces) => "{ " + spaces.map(p => show(p)).mkString(", ") + " }"
```
smarter: The default implementation of `Set` does not have a stable ordering, if you don't want your messages to be non-deterministic you should sort the spaces before printing them.

src/dotty/tools/dotc/transform/Space.scala
```diff
@@ -0,0 +1,406 @@
...
+  /** Display spaces */
+  def show(s: Space, mergeList: Boolean = false): String = s match {
+    case Empty => ""
+    case Typ(tp, decomposed) =>
+      val sym = tp.widen.classSymbol
+
+      if (sym.is(ModuleClass))
+        showType(tp)
+      else if (ctx.definitions.isTupleType(tp))
+        "(" + signature(tp).map(_ => "_").mkString(", ") + ")"
```
smarter: You could do `.mkString("(", ", ", ")")`

src/dotty/tools/dotc/transform/Space.scala
liufengyun: Thanks for the tip, I already face the indeterministic test problem.

src/dotty/tools/dotc/transform/Space.scala
liufengyun: Thanks for the tip, much neater!

smarter:
FYI, here's how Java enums from sources got exhaustiveness warnings in Scala 2.12: https://github.com/scala/scala/pull/4898

tests/patmat/t9232.check
```diff
@@ -1,5 +1,5 @@
 ./tests/patmat/t9232.scala:13: warning: match may not be exhaustive.
-It would fail on the following input: Node1(_), Node2()
+It would fail on the following input: Node2()
```
smarter: That's incorrect: `Node1(_)` should also be included.

liufengyun: Yes, it's imprecise. Currently I take **extractor** to match everything. I'll need to refine the algorithm for a better modelling of **extractors**.

smarter: OK, in cases where you don't know statically what an extractor match I think it's best to act as if it matched nothing

liufengyun: The interpretation of extractor depends on context -- in exhaustivity check, we think they cover more in order to be sound -- it's possible in practice programmers just use extractor for extracting, thus matching everything.

In redundancy check, we need to think extractors in previous casedefs cover less(Empty), but in current casedef cover more.

liufengyun:
cc: @odersky @smarter @DarkDimius @adriaanm 

This is an invitation for critical review of the new exhaustivity & redundancy check implementation. Ref #1261 .

I'm afraid that further polishing would make the code complex enough, which will make code review difficult. So I think it's the ripe time to review and estimate the new algorithm, to see whether it's worthy to continue the work. If the algorithm is indeed a new choice, we can continue to polish the details.

## Basic Idea

Basic idea of the algorithm is very simple: we think types and patterns as spaces of values.

For the problem of exhaustivity check, its formulation in terms of space is as follows:

    Is the space Typ(T) a subspace of the union of space covered by all the patterns?

The problem of unreachable patterns can be formulated as follows:

    Is the space covered by a pattern a subspace of the space covered by previous patterns?

 Some spaces can be completely decomposed into subspaces, such as abstract sealed classes, or-types, java enumerations and booleans. Case classes form a kind of product spaces.

The core algorithm is an abstract arithmetic about intersection, subtraction and subspace relationship of spaces.

## Features

- Support Java enums both from source and classfile
- Support constructor patterns
- Support constant patterns
- Support stable identifier patterns
- Support Or-types
- Support nice printing of tuples
- Support nice printing of list
- Support nice printing of path-dependent types

## Summary

Pros:

- Better performance
- Simpler & easier to maintain
- Fix scalac open issues
- Nicer message for path-dependent types

Cons:

- Less nice message for patterns with constants
- Inability with extractors
- Inability with guards
- add more here...

## Strong Points

### 1. Better Performance

The general claim may not hold, but for following case it's much better:

- tests/patmat/patmatexhaust-huge.scala

### 2. Better Maintainability

- Core algorithm 230LOC, total 488LOC - 361LOC without comments
- Based on very simple ideas
- Seems easy to adapt and fix issues
- Separation of exhaustivity & redundancy check and pattern match compilation

### 3. Fix following scalac open issues

- tests/patmat/t9672.scala - [link](https://issues.scala-lang.org/browse/SI-9672)
- tests/patmat/t9630.scala - [link](https://issues.scala-lang.org/browse/SI-9630) (fixed in 2.12-M4)
- tests/patmat/t9573.scala - [link](https://issues.scala-lang.org/browse/SI-9573)
- tests/patmat/t9399.scala - [link](https://issues.scala-lang.org/browse/SI-9399) (fixed in 2.12-M4)
- tests/patmat/t9289.scala - [link](https://issues.scala-lang.org/browse/SI-8511)
- tests/patmat/t8511.scala - [link](https://issues.scala-lang.org/browse/SI-8178)
- tests/patmat/t8178.scala - [link](https://issues.scala-lang.org/browse/SI-8178)
- tests/patmat/t8068.scala - [link](https://issues.scala-lang.org/browse/SI-8068)
- tests/patmat/t7466.scala - [link](https://issues.scala-lang.org/browse/SI-7466)
- tests/patmat/t6450.scala - [link](https://issues.scala-lang.org/browse/SI-6450)

Yet to investigate [SI-9657](https://issues.scala-lang.org/browse/SI-9657).

### 4. Nicer message for path-dependent types

`module.LetR()` instead `LetR`

- tests/patmat/t9289.scala

## Weak Points

### less friendly message for patterns with constants

- scalac warning: `Some((x: Int forSome x not in 3))`
- dotty warning: `Some(_)`

``` Scala
def foo(x:Id[Option[Int]]) = x match {
  case Some(3) => "foo"
  case None => "bar"
}
```

It's possible to do better in dotty, but it would complicate the algorithm that I want to postpone after review.

### Inability with extractors

Extractors are either taken to be match everything or nothing depending on context. It seems possible to do better with extractors in following example, but it would greatly complicate the algorithm that I want to postpone after review:

``` Scala
sealed trait Node
case class NodeA(i: Int) extends Node
case class NodeB(b: Boolean) extends Node
case class NodeC(s: String) extends Node

object Node {
  def unapply(node: Node): Option[(Node, Node)] = ???
}

// currently scalac doesn't do anything with following
// it's possible to do better in our case
object Test {
  def foo(x: Node): Boolean = x match { // unexhaustive
    case Node(NodeA(_), NodeB(_)) => true
    case Node(NodeA(4), NodeB(false)) => true // unreachable code
  }
}
```

### Inability with guards

Guards are either taken to be false or true depending on context. Scalac does better by recognizing constant guards like `if 1 > 0`.

It's possible to do the same if we've a constant folding transform.

- tests/patmat/patmatexhaust.scala

## Other Issues

### Anonymous class

In dotty, an anonymous class can inherit a sealed trait. However, if we add anonymous class to the trait's children annotations, there would be a pickling exception.

Guillaume suggests it might be good to forbid creating anonymous classes from **sealed** base.

Related discussion: [link](https://github.com/lampepfl/dotty/pull/1261/files/a3a345834922f56225b4aa686ffd3d0fa3987c4d..e665334b723cd982b328153cb050874469423e93#r64054377).

### Inheritance of case class

Without extractors and guards, the core algorithm seems very natural -- except in the case of a class inheriting case class.

Suppose we have following definition:

``` Scala
case class A(x: Int | Boolean)
class B(...) extends A(v)
class C(...) extends A(v)
```

For simplicity, current implementation defines `B /\ A(Boolean)` to be `B`,  but `A(Boolean) /\ A` to `A(Boolean)`. This breaks commutativity of intersection, which is incorrect generally. But inheriting case classes is weird in practice(though quite common in dotty), I expect the code would be rarely executed.

A more appropriate handling seems to check the type `T` of `v` in `extends A(v)`, and define `B /\ A(Boolean)` to be `A(T) /\ A(Boolean)`. But this way of elevating a subtype to its super-type may result in a space that covers values of class C and class A, which is semantically problematic.

## Links

- Textual description of algorithm - [link](https://gist.github.com/liufengyun/66bf0c1bc44cdad12e45dd6db689e9b3)
- [Warnings for pattern matching](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.530.8651&rep=rep1&type=pdf), L. Maranget, 2007

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -470,7 +470,7 @@ trait TypeAssigner {
     tree.withType(sym.nonMemberTermRef)
 
   def assignType(tree: untpd.Annotated, annot: Tree, arg: Tree)(implicit ctx: Context) =
-    tree.withType(AnnotatedType(arg.tpe, Annotation(annot)))
+    tree.withType(AnnotatedType(arg.tpe.widen, Annotation(annot)))
```
liufengyun: I'm not sure about this change, but without it there will be several TreeChecker errors. Could you please advise @smarter ? Thanks.

DarkDimius: I'm not sure if this is a good change. It makes ` val a : 1 = (1 : @annotation)` not typecheck as type of `1` will get widened to Int.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,512 @@
...
+import core.Decorators._
+import core.TypeApplications._
+import core.Symbols._
+import core.StdNames._
+import core.NameOps._
+import core.Constants._
+
+/** Space logic for checking exhaustivity and unreachability of pattern matching.
+ *
+ *  The core idea of the algorithm is that patterns and types are value
```
odersky: Would be good to give some intuition what a Space is here

src/dotty/tools/dotc/typer/TypeAssigner.scala
liufengyun: Thanks for the comment, I'll try remove it and investigate the errors in TreeChecker.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+  def isCaseClass(tp: Type): Boolean
+
+  /** Is the type `tp` decomposable? i.e. all values of the type can be covered
+   *  by its decomposed types.
+   *
+   * Abstract sealed class, OrType, Boolean and Java enums can be decomposed.
+   */
+  def canDecompose(tp: Type): Boolean
+
+  /** Return parameters types of the case class `tp` */
```
DarkDimius: Document if those are type parameters or term parameters(constructor arguments)?

DarkDimius:
There seems to be not a single test-case where selector is non abstract class.
Similar to this:
```
scala> sealed class A{}; class B extends A; class C extends B
defined class A
defined class B
defined class C

scala> class D extends B
defined class D

scala> (null: A) match {case t: B =>; case t: C => D}
error: error while loading D, class file './D.class' is broken
(class java.lang.ArrayIndexOutOfBoundsException/320)

scala> (null: A) match {case t: B =>; case t: C => ;}
<console>:15: warning: unreachable code
       (null: A) match {case t: B =>; case t: C => ;}
                                                ^
<console>:15: warning: match may not be exhaustive.
It would fail on the following input: A()
       (null: A) match {case t: B =>; case t: C => ;}
            ^
```

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -810,6 +810,13 @@ object Types {
       case _ => this
     }
 
+    /** Eliminate anonymous classes */
+    final def elimAnonymousClass(implicit ctx: Context): Type = this match {
```
DarkDimius: `deAnonymize` seems more in sync with `deAlias`.
Additionally  `elim` at least in my head means eliminating deeply inside the type, ie if there's a reference to an anonymous class somewhere deep you'll also eliminate it.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -50,6 +51,13 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
   override def transformMatch(tree: Match)(implicit ctx: Context, info: TransformerInfo): Tree = {
     val translated = new Translator()(ctx).translator.translateMatch(tree)
 
+    // check exhaustivity and unreachability
+    val engine = new SpaceEngine
```
DarkDimius: PatternMatcher now has a 🌠SpaceEngine🌠 and can do interstellar travels. Looking forward to battle space invaders.
👾👾👾👾

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+    flatten(s).map(doShow(_, false)).distinct.mkString(", ")
+  }
+
+  def checkable(tree: Match): Boolean = {
+    def isCheckable(tp: Type): Boolean = tp match {
+      case AnnotatedType(tp, annot) =>
+        (ctx.definitions.UncheckedAnnot != annot.symbol) && isCheckable(tp)
+      case _ => true // actually everything is checkable unless @unchecked
+
```
DarkDimius: scalac only checks sealed classes. We should consider if we indeed want to change this.
I'll create an issue about this.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -59,10 +67,16 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
       ctx.newSymbol(ctx.owner, ctx.freshName("ex").toTermName, Flags.Synthetic | Flags.Case, defn.ThrowableType, coord = tree.pos)
     val sel = Ident(selector.termRef).withPos(tree.pos)
     val rethrow = tpd.CaseDef(EmptyTree, EmptyTree, Throw(ref(selector)))
+
+    val tryMatch =  tpd.Match(sel, tree.cases ::: rethrow :: Nil)
```
DarkDimius: this conflicts with PR by @nicolasstucki

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -231,6 +233,13 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
                   ctx.compilationUnit.source.exists &&
                   sym != defn.SourceFileAnnot)
                 sym.addAnnotation(Annotation.makeSourceFile(ctx.compilationUnit.source.file.path))
+
```
DarkDimius: If we're unlucky, this will conflict with #1343.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+        if (acc.isEmpty) flat.map(s => Kon(tp, Nil :+ s))
+        else for (Kon(tp, ss) <- acc; s <- flat) yield Kon(tp, ss :+ s)
+      }
+    case Or(spaces) =>
+      spaces.flatMap(flatten _)
+    case _ => List(space)
+  }
+
+  /** Is `a` a subspace of `b`? Equivalent to `a - b == Empty`, but faster */
+  def subspace(a: Space, b: Space): Boolean = (a, b) match {
```
DarkDimius: If it returns a Boolean, it's better to follow `is*` naming convention, eg isConstained\isSubspace.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    case (Var(_, tp1), Typ(tp2, _)) => isSubType(tp1, tp2)
+    case (Var(_, _), Or(ss)) => ss.exists(subspace(a, _))
+    case (Var(_, _), _) => false
+    case (_, Var(_, _)) => false
+  }
+
+  /** Intersection of two spaces  */
+  def intersect(a: Space, b: Space): Space = (a, b) match {
+    case (Empty, _) | (_, Empty) => Empty
+    case (_, Or(ss)) => Or(ss.map(intersect(a, _)))
```
DarkDimius: filterConserve(_ ne Empty)? Same in next one.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    case (Or(ss), _) => Or(ss.map(intersect(_, b)))
+    case (Typ(tp1, _), Typ(tp2, _)) =>
+      if (isSubType(tp1, tp2)) a
+      else if (isSubType(tp2, tp1)) b
+      else Empty
```
DarkDimius: Why is `A intersect B` empty?
If two traits are unrelated they still can have common inhabitants. See below:
```scala
trait A
trait B

sealed class Parent
class Foo extends Parent with A
class Bar extends Parent with B

(null: parent) match {
  case (_: A) | (_: B) =>
}
```

DarkDimius: Upd: scalac does NOT complain about non-exhastive check here. Maybe it gives up :-)

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+/** abstract space logic */
+trait SpaceLogic {
+  /** Is `tp1` a subtype of `tp2`? */
+  def isSubType(tp1: Type, tp2: Type): Boolean
+
+  /** Is `tp1` the same type as `tp2`? */
+  def isEqualType(tp1: Type, tp2: Type): Boolean
+
+  /** Is `tp` a case class? */
```
DarkDimius: add assumptions. for example `intersect` on `Kon` assumes that case classes cannot inherit each other.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+  /** The space of a not covered by b */
+  def minus(a: Space, b: Space): Space = (a, b) match {
+    case (Empty, _) => Empty
+    case (_, Empty) => a
+    case (Typ(tp1, _), Typ(tp2, _)) =>
+      if (isSubType(tp1, tp2)) Empty
+      else if (isSubType(tp2, tp1) && canDecompose(tp1))
+        minus(Or(partitions(tp1)), b)
+      else a
```
DarkDimius: See counterexample above. 
Maybe it may make sence to try to decompose `a` here and see if some of results of decomposition ` minus b` change.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+ *  The problem of unreachable patterns can be formulated as follows:
+ *
+ *      Is the space covered by a pattern a subspace of the space covered by previous patterns?
+ */
+
+
+/** space definition */
+sealed trait Space
+case object Empty extends Space
+case class Typ(tp: Type, decomposed: Boolean) extends Space
```
DarkDimius: document `decomposed`

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+  def minus(a: Space, b: Space): Space = (a, b) match {
+    case (Empty, _) => Empty
+    case (_, Empty) => a
+    case (Typ(tp1, _), Typ(tp2, _)) =>
+      if (isSubType(tp1, tp2)) Empty
+      else if (isSubType(tp2, tp1) && canDecompose(tp1))
+        minus(Or(partitions(tp1)), b)
+      else a
+    case (Typ(tp1, _), Kon(tp2, ss)) =>
+      if (isSubType(tp1, tp2)) minus(Kon(tp2, signature(tp2).map(Typ(_, false))), b)
```
DarkDimius: this case seems to be wrong: `a - b == fun(b) - fun(b)`, the implementation does not take into account a.
offtopic: https://www.kirupa.com/forum/showthread.php?328247-Are-there-names-for-x-and-y-in-x-y
but maybe this is a corner case. I propose to at least document it.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    case (_, Or(ss)) =>
+      ss.foldLeft(a)(minus)
+    case (Or(ss), _) =>
+      Or(ss.map(minus(_, b)))
+    case (Kon(tp1, ss), Typ(tp2, _)) =>
+      if (isSubType(tp1, tp2)) Empty
+      else if (simplify(a) == Empty) Empty
+      else if (isSubType(tp2, tp1) && canDecompose(tp1))
+        minus(Or(partitions(tp1)), b)
+      else a
```
DarkDimius: Not a bug, but potential improvement: 
```scala
trait A
trait B

abstract sealed case class Parent()
class Foo extends Parent with A
class Bar extends Parent with B
```

`Parent() - A` is known more precisely to be `Bar`

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    case (Kon(tp1, ss1), Kon(tp2, ss2)) =>
+      if (!isEqualType(tp1, tp2)) a
+      else if (ss1.zip(ss2).exists(p => simplify(intersect(p._1, p._2)) == Empty)) a
```
DarkDimius: all following 3 cases are non-obvious. it would be nice to document them

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+case object Empty extends Space
+case class Typ(tp: Type, decomposed: Boolean) extends Space
+case class Kon(tp: Type, params: List[Space]) extends Space
+sealed trait Point extends Space
+case class Var(sym: Symbol, tp: Type) extends Point
+case class Const(value: Constant, tp: Type) extends Point
+case class Or(spaces: List[Space]) extends Space
+
+/** abstract space logic */
+trait SpaceLogic {
```
DarkDimius: TODO: this class __desperately needs__ infrastructure that will check code coverage by tests.

src/dotty/tools/dotc/transform/patmat/Space.scala
DarkDimius: document that tp is erazed.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+      else if (roundUp) Typ(pat.tpe.stripAnnots, false)
+      else Empty
+    case Typed(pat @ UnApply(_, _, _), _) => project(pat)
+    case Typed(expr, _) => Typ(erase(expr.tpe.stripAnnots), true)
+    case _ =>
+      Empty
+  }
+
+  /* Erase a type binding according to erasure rule */
+  def erase(tp: Type): Type = {
```
DarkDimius: counter-example
```scala
       (null) match { case a: (List[Int]| Stack[Int]) => }
```
I propose to use real erasure: `TypeErasure.erasure(tp)`

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+case class Kon(tp: Type, params: List[Space]) extends Space
+sealed trait Point extends Space
+case class Var(sym: Symbol, tp: Type) extends Point
+case class Const(value: Constant, tp: Type) extends Point
+case class Or(spaces: List[Space]) extends Space
+
+/** abstract space logic */
+trait SpaceLogic {
+  /** Is `tp1` a subtype of `tp2`? */
+  def isSubType(tp1: Type, tp2: Type): Boolean
```
DarkDimius: there may be huge cans of worms here as you are using pre-erasure `<:<` on semi-erazed or fully-erazed types.
Maybe it would make sense to go into full erasure and replace sub typing by `derivesFrom` after deconstructing Or types into your OrSpaces.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+  }
+
+
+  /** Expose refined type to eliminate reference to type variables
+   *
+   *  A = B                      M { type T = A }        ~~>  M { type T = B }
+   *
+   *  A <: X :> Y                M { type T = A }        ~~>  M { type T <: X :> Y }
+   *
+   *  A <: X :> Y  B <: U :> V   M { type T <: A :> B }  ~~>  M { type T <: X :> B }
```
DarkDimius: typo: last B should be V

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+    val origin = doErase(tp)
+    if (origin =:= defn.ArrayType) tp else origin
+  }
+
+  /** Is `tp1` a subtype of `tp2`?  */
+  def isSubType(tp1: Type, tp2: Type): Boolean = {
+    // expose is important due to type bounds
+    // check SI-9657 and tests/patmat/gadt.scala
+    tp1 <:< expose(tp2)
```
DarkDimius: this seems like a hack to me. You are trying to mimic erazed semantics of instanceOf checks by non-erazed types.
If you'll go full-erasure you shouldn't have this problem.  
What I'm proposing is more-or-less `tp1.typeSymbol.derivesFrom(tp2.typeSymbol)`, but you'll need to additionally take OrTypes into account.

 I believe you can handle Or Types on a `Space` level by using 
`Space(A | B) == Space(A.erasure) union Space(A.erasure)`

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+  /** Parameter types of the case class type `tp`  */
+  def signature(tp: Type): List[Type] = {
+    val ktor = tp.classSymbol.primaryConstructor.info
+
+    val meth =
+      if (ktor.isInstanceOf[MethodType]) ktor
+      else
+      tp match {
+        case AppliedType(_, params) =>
```
DarkDimius: ping @odersky, this kind-of reimplements asSeenFrom for constructor. I guess there should be a better way to do it but @liufengyun says it does not work.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+          ktor.appliedTo(refined)
+        case _ =>
+          ktor
+      }
+
+    // refine path-dependent type in params. refer to t9672
+    meth.firstParamTypes.map(_.stripTypeVar).map { ptp =>
+      (tp, ptp) match {
+        case (TypeRef(ref1: TypeProxy, _), TypeRef(ref2: TypeProxy, name)) =>
+          if (ref1.underlying <:< ref2.underlying) TypeRef(ref1, name) else ptp
```
DarkDimius: same here. this is kind-of asSeenFrom.

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+        case _ => ptp
+      }
+    }
+  }
+
+  def partitions(tp: Type): List[Space] = {
+    val children = tp.classSymbol.annotations.filter(_.symbol == ctx.definitions.ChildAnnot).map { annot =>
```
DarkDimius: getAnnotation

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    meth.firstParamTypes.map(_.stripTypeVar).map { ptp =>
+      (tp, ptp) match {
+        case (TypeRef(ref1: TypeProxy, _), TypeRef(ref2: TypeProxy, name)) =>
+          if (ref1.underlying <:< ref2.underlying) TypeRef(ref1, name) else ptp
+        case _ => ptp
+      }
+    }
+  }
+
+  def partitions(tp: Type): List[Space] = {
```
DarkDimius: decompose

DarkDimius: add a comment that says that there is an assumption that `canDecompose(tp)`

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+
+  /** Refine tp2 based on tp1
+   *
+   *  E.g. if `tp1` is `Option[Int]`, `tp2` is `Some`, then return
+   *  `Some[Int]`.
+   *
+   *  If `tp1` is `path1.A`, `tp2` is `path2.B`, and `path1` is subtype of
+   *  `path2`, then return `path1.B`.
+   */
+  def refine(tp1: Type, tp2: Type): Type = (tp1, tp2) match {
```
DarkDimius: @odersky, I have a feeling that we have similar functionality already somewhere. Would you have a look?

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    val Match(sel, cases) = _match
+    // ignore selector type for now
+    // val selTyp = sel.tpe.widen.elimAnonymousClass.dealias
+
+    // starts from the second, the first can't be redundant
+    (1 until cases.length).foreach { i =>
+      // in redundancy check, take guard as false, take extractor as match
+      // nothing in order to soundly approximate
+      val prevs = cases.take(i).map { x =>
+        if (x.guard.isEmpty) project(x.pat, false)
```
DarkDimius: `intersect(project(x.pat, false), project(seltor))` will be more precise

src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -0,0 +1,574 @@
...
+    case _ => List(space)
+  }
+
+  /** Is `a` a subspace of `b`? Equivalent to `a - b == Empty`, but faster */
+  def subspace(a: Space, b: Space): Boolean = (a, b) match {
+    case (Empty, _) => true
+    case (_, Empty) => false
+    case (Or(ss), _) => ss.forall(subspace(_, b))
+    case (Typ(tp1, _), Typ(tp2, _)) =>
+      isSubType(tp1, tp2)
```
liufengyun: Can be refined if `tp1` is decomposable.

src/dotty/tools/dotc/typer/TypeAssigner.scala
liufengyun: This widen corresponds to the widen in the [typer](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/typer/Typer.scala#L1176):

``` Scala
      val tpt = TypeTree(AnnotatedType(arg1.tpe.widen, Annotation(annot1)))
      assignType(cpy.Typed(tree)(arg1, tpt), tpt)
```

Disable widening in typer breaks a lot of assumptions elsewhere, for now I think it's OK to just keep this change.

smarter: Interesting, I think you should open an issue about this so that we don't forget it at least.

jvican: I have a hunch that this change will disable things like `42.type` (SIP 23). If Dotty supports it in the future, we'll need to revisit this change...

liufengyun: Actually there is one issue created for this:

https://github.com/lampepfl/dotty/issues/1330

A lot of places depend on widening, currently there is no simple fix.

src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: It seems here we can't just reuse the existing type erasure by splitting or-types to or-spaces, e.g.  `M[Int | Boolean]` may not necessarily the same as `M[Int] | M[Boolean]`.

src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: This is addressed in the new PR by handling AND and OR-types explicitly. `TypeErasure` can't be simply reused here as I don't want `Int | Boolean` get erased to `Object` and the `prefix` in `TypeRef` removed.

src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: This is also addressed in the PR.

src/dotty/tools/dotc/transform/patmat/Space.scala
smarter: Maybe something like `asSeenFrom(tp2, tp1 & tp2)`

liufengyun: Just tried, no luck.

smarter: yeah `asSeenFrom` does not make sense here I meant to say something like `(tp1 & tp2).baseTypeWithArgs(tp2.classSymbol)`, no idea if that would work here.

liufengyun: No luck with `baseTypeWithArgs` either. Actually I also tried something like `tp1.asSeenFrom(tp1.prefix, tp1.prefix.classSymbol)` for the second case, but with no luck.

milessabin:
It would be good to know where the various open Scala GADT bugs stand wrt this, in particular [SI-5195](https://issues.scala-lang.org/browse/SI-5195?jql=text%20~%20%22gadt%22%20AND%20status%20%3D%20Open).

liufengyun:
@milessabin [SI-5195](https://issues.scala-lang.org/browse/SI-5195?jql=text%20~%20%22gadt%22%20AND%20status%20%3D%20Open) is about GADT type checking & inference in patmat, which is not covered by this PR. Exhaustivity & redundancy check happens after typer and only generate warnings.

milessabin:
I obviously read too much into "add GADT test" and the list of open GADT bugs above.

DarkDimius:
closing in favor of #1364

PR #1270, opened 2016-05-20
Fix bootstrap

odersky:
Fixes needed so that dotc can selectively compile some parts of itself, using TASTY as the pickled symbol format. Based on #1243.

### DIFF
--- a/src/dotty/tools/dotc/core/Contexts.scala
+++ b/src/dotty/tools/dotc/core/Contexts.scala
@@ -248,7 +248,7 @@ object Contexts {
       withPhase(phase.id)
 
     final def withPhaseNoLater(phase: Phase) =
-      if (ctx.phase.id > phase.id) withPhase(phase) else ctx
+      if (phase.exists && ctx.phase.id > phase.id) withPhase(phase) else ctx
 
     /** If -Ydebug is on, the top of the stack trace where this context
      *  was created, otherwise `null`.
--- a/src/dotty/tools/dotc/core/Symbols.scala
+++ b/src/dotty/tools/dotc/core/Symbols.scala
@@ -398,10 +398,10 @@ object Symbols {
 
     /** Subclass tests and casts */
     final def isTerm(implicit ctx: Context): Boolean =
-      (if(isDefinedInCurrentRun) lastDenot else denot).isTerm
+      (if (defRunId == ctx.runId) lastDenot else denot).isTerm
 
     final def isType(implicit ctx: Context): Boolean =
-      (if(isDefinedInCurrentRun) lastDenot else denot).isType
+      (if (defRunId == ctx.runId) lastDenot else denot).isType
 
     final def isClass: Boolean = isInstanceOf[ClassSymbol]
 
--- a/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
+++ b/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
@@ -485,4 +485,15 @@ object TastyFormat {
     case PRIVATEqualified => "PRIVATEqualified"
     case PROTECTEDqualified => "PROTECTEDqualified"
   }
+
+  /** @return If non-negative, the number of leading references of a length/trees entry.
+   *          If negative, minus the number of leading non-reference trees.
+   */
+  def numRefs(tag: Int) = tag match {
+    case VALDEF | DEFDEF | TYPEDEF | TYPEPARAM | PARAM | NAMEDARG | RETURN | BIND |
+         SELFDEF | REFINEDtype => 1
+    case RENAMED | PARAMtype => 2
+    case POLYtype | METHODtype => -1
+    case _ => 0
+  }
 }
--- a/src/dotty/tools/dotc/core/tasty/TreePickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/TreePickler.scala
@@ -217,8 +217,8 @@ class TreePickler(pickler: TastyPickler) {
     case tpe: RefinedType =>
       writeByte(REFINEDtype)
       withLength {
-        pickleType(tpe.parent)
         pickleName(tpe.refinedName)
+        pickleType(tpe.parent)
         pickleType(tpe.refinedInfo, richTypes = true)
       }
     case tpe: TypeAlias =>
--- a/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -153,7 +153,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
     /** Record all directly nested definitions and templates between current address and `end`
      *  as `OwnerTree`s in `buf`
      */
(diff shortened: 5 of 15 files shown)


src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -398,10 +398,10 @@ object Symbols {
 
     /** Subclass tests and casts */
     final def isTerm(implicit ctx: Context): Boolean =
-      (if(isDefinedInCurrentRun) lastDenot else denot).isTerm
+      (if (defRunId == ctx.runId) lastDenot else denot).isTerm
```
smarter: Why not change the definition of `isDefinedInCurrentRun` ?

src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -248,7 +248,7 @@ object Contexts {
       withPhase(phase.id)
 
     final def withPhaseNoLater(phase: Phase) =
-      if (ctx.phase.id > phase.id) withPhase(phase) else ctx
+      if (phase.exists && ctx.phase.id > phase.id) withPhase(phase) else ctx
```
smarter: Does this happen in practice?

src/dotty/tools/dotc/core/Symbols.scala
odersky: Because it's used elsewhere to mean: In a compilation unit that's currently being compiled.

src/dotty/tools/dotc/core/Contexts.scala
odersky: Yes, there are tests that break (the tests don't start a full compiler).

odersky:
We now can selectively compile individual dotc files and directories using Tasty to access the information of the other parts of dotc. #1274 demonstrates what I did to verify this.

odersky:
Rebased to master

src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -174,32 +174,22 @@ object TreeTransforms {
  }
 
   /** A helper trait to transform annotations on MemberDefs */
-  trait AnnotationTransformer extends MiniPhaseTransform with InfoTransformer {
+  trait AnnotationTransformer extends MiniPhaseTransform with DenotTransformer {
```
DarkDimius: Why is this change needed?
As annotation transformers also change types, isn't every annotation transformer and info-transformer?

src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -485,4 +485,15 @@ object TastyFormat {
     case PRIVATEqualified => "PRIVATEqualified"
     case PROTECTEDqualified => "PROTECTEDqualified"
   }
+
+  /** @return If non-negative, the number of leading references of a length/trees entry.
+   *          If negative, minus the number of leading non-reference trees.
+   */
+  def numRefs(tag: Int) = tag match {
```
DarkDimius: This method seem to only be used by tasty-unpickler. Maybe move it there?

src/dotty/tools/dotc/core/tasty/TreePickler.scala
```diff
@@ -217,8 +217,8 @@ class TreePickler(pickler: TastyPickler) {
     case tpe: RefinedType =>
       writeByte(REFINEDtype)
       withLength {
-        pickleType(tpe.parent)
```
DarkDimius: This is an incompatible change. Why was the order changed?

src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -145,19 +209,25 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
...
-          // Create a stub; owner might be wrong but will be overwritten later.
-          forkAt(addr).createSymbol()
-          val sym = symAtAddr(addr)
-          ctx.log(i"forward reference to $sym")
-          stubs += sym
-          sym
-      }
+    def readSymRef()(implicit ctx: Context): Symbol = symbolAt(readAddr())
+
+    /** The symbol at given address; createa new one if none exists yet */
```
DarkDimius: `createa` : typo

src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -393,44 +476,31 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table) {
...
-                  preExisting
-                case none =>
-                  ctx.newNakedSymbol(start.index)
-              }
-              val denot = ctx.SymDenotation(symbol = sym, owner = ctx.owner, name, flags, completer, privateWithin)
-              sym.denot = denot
-              sym
-            }
-        } // TODO set position
+              ctx.newClassSymbol(ctx.owner, name.asTypeName, flags, completer, privateWithin, coord = start.index)
```
DarkDimius: The only difference between branches is the name. Maybe try:
```scala
val symName = if (isClass) name.asTypeName else name
ctx.newClassSymbol(ctx.owner, name.asTypeName, flags, completer, privateWithin, coord = start.index)

```

src/dotty/tools/dotc/core/tasty/TreePickler.scala
odersky: To harmonize the layout. Non-trees now always come before trees in Tasy entries (except for some cases where trailing non-trees are optional as explained in numRefs.

src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
odersky: And whether create a symbol or a class symbol no? That prevented me from merging the branches.

src/dotty/tools/dotc/core/tasty/TastyFormat.scala
odersky: I think it is essentially an information about the format. So the place looks right to me. Otherwise put, if we change the format of some entries, this method needs to be updated in sync. With the information provided by it, we enable generic traversals of Tasty trees in other (future) modules as well.

src/dotty/tools/dotc/transform/TreeTransform.scala
odersky: Currently that's not true, but I see you argument. I..e if we left out the self type transform from FirstTransform it would not be an InfoTransformer. And that transform is in FirstTransform only by accident, it could well be elsewhere.

*However* one could argue that every annotation transform should affect annotated types, and annotated types can be infos of symbols so in that sense every annotation transform *is* an info transform.

*But* that would require a complete typemap of every type in every info transform, and we know this leads to cyclic references. So we can't do this at present.

Bottom line: When in doubt decouple the functionalities.

src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
DarkDimius: My bad. You're right.

odersky:
ready to merge?

DarkDimius:
LGTM

PR #1271, opened 2016-05-20
Fix bootstrap 2

odersky:
This differs from #1270 only in the last commit, where finger printing was removed. The plan is to merge this PR with master one day after #1270 to verify what the effect on compiler performance is.


smarter:
> Trying the dotc bootstrap revealed that Frozen logic can lead
to assertion errors. So we want to remove it, but then taking
fingerprints would no longer be cost-effective.

Is it not possible to fix `Frozen`?

odersky:
@smarter I don't see a way to fix `Frozen`.

DarkDimius:
> Trying the dotc bootstrap revealed that Frozen logic can lead to assertion errors.

would you elaborate about under which circumstances the Frozen logic is not working? Is it specific to Dotty bootstrap?

odersky:
It turned out that errors connected with Frozen were just early warnings of an unrelated problem with Tasty unpickling. So we do not need to drop it, after all.

Issue #1255, opened 2016-05-16
incorrect type checking in pattern matching


liufengyun:
Following code passes type checking in dotty:

``` Scala
object Test {
  def foo(x: Option[Int]) = x match {
    case Some(_: Double) => true
    case None => true
  }
}
```

In scalac it generates following error message:

```
error: scrutinee is incompatible with pattern type;
 found   : Double
 required: Int
    case Some(_: Double) => true
                 ^
one error found
```

DarkDimius:
I propose to make it both more general and simpler by decoupling the transformation from typer and doing it after pattern matcher.

I propose to intercept `sc.isInstanceOf[Sel]`: depending on the erased type of `sc` and `Sel` either:
 - report the error\warning if test is always failing;
 - replace the call be constant if the result is statically known;
 - leave as is.

Here is the table for transformation:

| sel\Sc      | Trait                           | Class                           | final Class                |
|-------------|---------------------------------|---------------------------------|----------------------------|
| Trait       | happens                         | happens                         | known statically |
| Class       | happens                         | false if classes are unrelated | known statically |
| Final class | false if classes are unrelated |  false if classes are unrelated | known statically |

DarkDimius:
I propose to make it both more general and simpler by decoupling the transformation from typer and doing it after pattern matcher.

I propose to intercept `sc.isInstanceOf[Sel]`: and, depending on the erased type of `sc` and `Sel`, either:
 - report the error\warning if test is always failing;
 - replace the call be constant if the result is statically known;
 - leave as is.
Here is the table for transformation:

| sel\Sc      | Trait                          | Class                          | final Class      |
|-------------|--------------------------------|--------------------------------|------------------|
| Trait       | happens                        | happens                | known statically |
| Class       | happens                        | always false if classes are unrelated | known statically |
| Final class | always false if classes are unrelated | always false if classes are unrelated | known statically |

liufengyun:
@felixmulder I found an interesting bug with scalac related to this issue, which you may want to add to the test set:

```
case class A(v: Int)
class B(v: Int) extends A(v)

object Test {
  val a = new B(1)
  a match { case A(_) => 1 }
}
```

The code above looks perfectly fine, but scalac generates an error:

```
error: constructor cannot be instantiated to expected type;
 found   : A
 required: B
  a match { case A(_) => 1 }
                 ^
one error found
```

However, if we annote `a` with `val a:A = new B(1)`, everything is fine.

[...]

PR #1275, opened 2016-05-24
Add partial evaluation of `isInstanceOf` mini-phase

felixmulder:
In #1255 it was suggested to add a mini-phase that would simply partially evaluate calls to `isInstanceOf` and issue warnings/errors where appropriate.

This PR adds the ability to discern if the transformed tree came from a `match` or from an `if` as well as the proper error messages and warnings.

The following example will yield an error:

```scala
val x = Option(1)
x match {
  case Some(_: Double) => // Error
}
```

While the following will yield a warning:

```scala
if (x.isInstanceOf[Double]) { ...
```


src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+  def phaseName = "isInstanceOfEvaluator"
+  override def transformTypeApply(tree: TypeApply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    val defn = ctx.definitions
+
+    /** Handles the four cases of statically known `isInstanceOf`s and gives
+     *  the correct warnings, or an error if statically known to be false in
+     *  match
+     */
+    def handleStaticallyKnown(tree: Select, scrutinee: Type, selector: Type, inMatch: Boolean, pos: Position): Tree =
+      if (!(scrutinee <:< selector) && inMatch) {
```
DarkDimius: it may be better to compute `scrutinee <:< selector` only once.

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+
+import dotty.tools.dotc.util.Positions._
+import TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+import core._
+import Contexts.Context, Types._, Constants._, Decorators._, Symbols._
+import TypeUtils._, TypeErasure._
+
+
+/** Implements partial `isInstanceOf` evaluation according to the matrix on:
+ *  https://github.com/lampepfl/dotty/issues/1255
```
smarter: Instead of referring to an issue, I'd appreciate it if the comment for this phase explained in more details what it did and why, as it is I have trouble figuring out what's going on here :).

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+     *  constant
+     */
+    def evalTypeApply(tree: TypeApply): Tree =
+      if (tree.symbol != defn.Any_isInstanceOf) tree
+      else tree.fun match {
+        case s: Select => {
+          val scrutinee = erasure(s.qualifier.tpe.widen)
+          val selector  = erasure(tree.args.head.tpe.widen)
+
+          val scTrait = scrutinee.typeSymbol is Flags.Trait
```
smarter: import Flags._

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+        rewrite(tree, to = false)
+      }
+
+    /** Rewrites the select to a boolean if `to` is false or if the qualifier
+     *  is a primitive.
+     *
+     *  If `to` is set to true and the qualifier is not a primitive, the
+     *  instanceOf is replaced by a null check, since:
+     *
+     *  `srutinee.isInstanceOf[Selector]` if `scrutinee eq null`
```
smarter: typo: srutinee -> scrutinee

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+        rewrite(tree, to = false)
+      } else {
+        ctx.warning(
+          s"will always yield false since `${scrutinee.show}` is not a subclass of `${selector.show}`",
+          tree.pos
+        )
+        rewrite(tree, to = false)
+      }
+
+    /** Rewrites the select to a boolean if `to` is false or if the qualifier
```
smarter: This logic is a bit weird `rewrite(tree, to = true)` does not mean `Rewrite tree to true` it means `Rewrite tree to true unless the type is nullable`

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+        rewrite(tree, to = false)
+      } else if (!(scrutinee <:< selector) && !inMatch) {
+        ctx.warning(
+          s"this will always yield false since `${scrutinee.show}` is not a subclass of `${selector.show}` (will be optimized away)",
+          pos
+        )
+        rewrite(tree, to = false)
+      } else if (scrutinee <:< selector && !inMatch) {
+        ctx.warning(
+          s"this will always yield true since `${scrutinee.show}` is a subclass of `${selector.show}` (will be optimized away)",
```
smarter: Will it _always_ yield true or only if the value is not null?

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,148 @@
...
+      } else {
+        ctx.warning(
+          s"will always yield false since `${scrutinee.show}` is not a subclass of `${selector.show}`",
+          tree.pos
+        )
+        rewrite(tree, to = false)
+      }
+
+    /** Rewrites the select to a boolean if `to` is false or if the qualifier
+     *  is a primitive.
```
smarter: you mean value class, not just primitive

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,166 @@
...
+          if (valueClassesOrAny) tree
+          else if (knownStatically)
+            handleStaticallyKnown(s, scrutinee, selector, inMatch, tree.pos)
+          else if (falseIfUnrelated && scrutinee <:< selector)
+            // scrutinee is a subtype of the selector, safe to rewrite
+            rewrite(s, to = true)
+          else if (falseIfUnrelated && !(selector <:< scrutinee))
+            // selector and scrutinee are unrelated
+            handleFalseUnrelated(s, scrutinee, selector, inMatch)
+          else if (happens) tree
```
DarkDimius: there's no difference between `happens` and `!happens`

tests/run/t6637.scala
```diff
@@ -1,8 +0,0 @@
-
-object Test extends dotty.runtime.LegacyApp {
```
DarkDimius: why was this test deleted?

felixmulder: In this test `List()` is evaluated to `List[Nothing]` which leads to `head.isInstanceOf[A with B]` will always be true since `Nothing` is a subtype of `A`.

The new mini-phase will report an error here, as such the test will fail because it won't compile and `println("ok")` will never run.

Of course, the mini-phase could just warn here...decide what you think is best and I'll adjust the PR.

DarkDimius: maybe it would be better to remove the `.isInstanceOf[A with B]` part, while keeping the other parts of the test?

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
felixmulder: Agreed, it was kept to showcase all cases - but I'll change it to and note this in a comment.

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -0,0 +1,166 @@
...
+    /** Rewrites the select to a boolean if `to` is false or if the qualifier
+     *  is a value class.
+     *
+     *  If `to` is set to true and the qualifier is not a primitive, the
+     *  instanceOf is replaced by a null check, since:
+     *
+     *  `scrutinee.isInstanceOf[Selector]` if `scrutinee eq null`
+     */
+    def rewrite(tree: Select, to: Boolean): Tree =
+      if (!to || !tree.qualifier.tpe.widen.derivesFrom(defn.AnyRefAlias))
```
DarkDimius: Counterexample: 
```scala
def foo = {println(1); 1}.isInstanceOf[Int]
```

need to check if `tpd.isPure(tree.qual)`. If not, rewrite to `Block(List(tree.qual), Literal(Constant(to)))`.

src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -91,11 +91,16 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
...
-      if (!to || !tree.qualifier.tpe.widen.derivesFrom(defn.AnyRefAlias))
-        Literal(Constant(to))
-      else
-        Apply(tree.qualifier.select(defn.Object_ne), List(Literal(Constant(null))))
+    def rewrite(tree: Select, to: Boolean): Tree = {
+      val expr =
+        if (!to || !tree.qualifier.tpe.widen.derivesFrom(defn.AnyRefAlias))
+          Literal(Constant(to))
+        else
+          Apply(tree.qualifier.select(defn.Object_ne), List(Literal(Constant(null))))
```
DarkDimius: if `tree.qualifier` is not pure, it will be computed twice

DarkDimius:
LGTM

PR #1277, opened 2016-05-25
fix newline in DOTTY_ROOT when cd produces output

DarkDimius:
Helps in some environments where cd is not silent.


smarter:
I have never heard of such an environment but LGTM :).

PR #1278, opened 2016-05-25
Fix errors with Javadoc 8

Blaisorblade:
`sbt dotty-interfaces/publishLocal` fails for me, fix enough errors to
make it work.


smarter:
LGTM, thank you for your contribution!

PR #1281, opened 2016-05-26
Fixes to lambdalift that prevent memory leaks.

DarkDimius:
CollectDependencies had incorrect handling of `Ident`s. If it had ever met
    an `Ident` to a symbol defined outside of current owner-chain(e.g.
    Predef.println) it would issue narrowTo(enclosingClass).

This is very conservative and leads to memory leaks even in trivial lambdas.

### DIFF
--- a/tests/run/t5375.scala
+++ b/tests/run/t5375.scala
@@ -1,3 +1,14 @@
+/** Hello fellow compiler developer.
+    if you are wondering why does test suite hang on this test
+    then it's likely that the lambda inside map has been compiled into static method
+    unfotrunatelly, as it is executed inside static object initializer,
+    it is executed inside class-loader, in a synchronized block that is not source defined.
+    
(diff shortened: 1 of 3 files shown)


DarkDimius:
@odersky please review.

tests/run/t5375.scala
```diff
@@ -1,3 +1,14 @@
...
+    if you are wondering why does test suite hang on this test
+    then it's likely that the lambda inside map has been compiled into static method
+    unfotrunatelly, as it is executed inside static object initializer,
+    it is executed inside class-loader, in a synchronized block that is not source defined.
+    
+    If the lambda will be static Test$#foo, calling it through a different thread would require grabbing the
+    lock inside classloader. Unlike if it not static and is called through This(Test).foo, no lock is grabbed.
+    
+    @DarkDimius  
+*/
```
odersky: :-)

odersky:
LGTM

PR #1283, opened 2016-05-27
Fix #1276: `_` binding in patternmatcher

### LINKED ISSUE #1276: Patternmatcher fails to bind `_: _*` correctly
When transforming:

```scala
val List(_: _*) = List(1)
```

`_` is evaluated to be an `Int` instead of a `Seq[Int]`.

felixmulder:
This fix changes the typer to infer a `A*` instead of `Seq[A]` for repeated arguments. Subsequent parts of the patternmatcher and checking has been changed to accommodate this.

Related issue: #1276

### DIFF
--- a/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -1094,6 +1094,7 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
       */
     object WildcardPattern {
       def unapply(pat: Tree): Boolean = pat match {
+        case Typed(_, arg) if arg.tpe.isRepeatedParam => true
         case Bind(nme.WILDCARD, WildcardPattern()) => true // don't skip when binding an interesting symbol!
         case t if (tpd.isWildcardArg(t))            => true
         case x: Ident                              => isVarPattern(x)
@@ -1157,8 +1158,8 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
 
       object TypeBound {
         def unapply(tree: Tree): Option[Type] = tree match {
-          case Typed(_, _)  => Some(tree.typeOpt)
-          case _                                      => None
+          case Typed(_, arg) if !arg.tpe.isRepeatedParam => Some(tree.typeOpt)
+          case _                                         => None
         }
       }
 
--- a/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -119,7 +119,12 @@ class TreeChecker extends Phase with SymTransformer {
     val prevPhase = ctx.phase.prev // can be a mini-phase
     val squahsedPhase = ctx.squashed(prevPhase)
     ctx.echo(s"checking ${ctx.compilationUnit} after phase ${squahsedPhase}")
-    val checkingCtx = ctx.fresh.setReporter(new ThrowingReporter(ctx.reporter))
+
+    val checkingCtx = ctx
+        .fresh
+        .setMode(Mode.ImplicitsEnabled)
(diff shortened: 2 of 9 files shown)


src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -119,7 +119,12 @@ class TreeChecker extends Phase with SymTransformer {
     val prevPhase = ctx.phase.prev // can be a mini-phase
     val squahsedPhase = ctx.squashed(prevPhase)
     ctx.echo(s"checking ${ctx.compilationUnit} after phase ${squahsedPhase}")
-    val checkingCtx = ctx.fresh.setReporter(new ThrowingReporter(ctx.reporter))
+
+    val checkingCtx = ctx
+        .fresh
+        .setMode(Mode.ImplicitsEnabled)
```
felixmulder: Added `setMode(Mode.ImplicitsEnabled)` here since `Compiler` uses that in its initial context.

tests/pos/t9765.scala
```diff
@@ -0,0 +1,7 @@
+case class A(v: Int)
+class B(v: Int) extends A(v)
+
+object Test {
+  val a = new B(1)
+  a match { case A(_) => 1 }
+}
```
liufengyun: Is there a typo? You probably mean SI-9795 here.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1559,30 +1560,31 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
...
         // [2] then we have to index the binder that represents the sequence for the remaining subpatterns, except for...
         // [3] the last one -- if the last subpattern is a sequence wildcard:
         //       drop the prefix (indexed by the refs on the preceding line), return the remainder
-        (    productElemsToN(binder, firstIndexingBinder)
-          ++ genTake(binder, expectedLength)
-          ++ lastTrees
-          ).toList
-      }
+
+        productElemsToN(binder, firstIndexingBinder) ++
```
DarkDimius: I find previous version easier to read.

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1157,8 +1158,8 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {thisTrans
 
       object TypeBound {
         def unapply(tree: Tree): Option[Type] = tree match {
-          case Typed(_, _)  => Some(tree.typeOpt)
-          case _                                      => None
+          case Typed(_, arg) if !arg.tpe.isRepeatedParam => Some(tree.typeOpt)
+          case _                                     => None
```
DarkDimius: indentation a bit off.

tests/pos/t9765.scala
felixmulder: Yes, thank you - fixed it :)

DarkDimius:
LGTM, thanks!

src/dotty/tools/dotc/transform/TreeChecker.scala
smarter: I don't think `TreeChecker` should have to do implicit searches, did you have to enable this to make some tests pass?

felixmulder:
@smarter - for some reason github couldn't find your comment so I'm reposting it here or maybe you deleted it? Anyway, for posterity...

> I don't think `TreeChecker` should have to do implicit searches, did you have to enable this to make some tests pass?

I had a look in `Compiler.scala` and figured that tree checking should have the same fresh context as the regular compiler. That being said - this change makes absolutely no difference now. Why? Because for some reason the `ImplicitsEnabled` mode is not retracted after the implicit search - so it's prevalent throughout the entire pipeline. I reckon this is a bug, but I think @DarkDimius is already aware of it - not sure whether to report it properly or not.

PR #1288, opened 2016-05-31
Implicit scope caching: bug fixes and performance improvements

smarter:
Review by @odersky


src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1146,11 +1146,17 @@ trait Applications extends Compatibility { self: Typer =>
...
+        def narrowByTrees(alts: List[TermRef], args: List[Tree], resultType: Type): List[TermRef] = {
+          val alts2 = alts.filter(alt =>
+            isDirectlyApplicable(alt, targs, args, resultType)
           )
+          if (alts2.isEmpty && !ctx.isAfterTyper)
+            alts.filter(alt =>
+              isApplicable(alt, targs, args, resultType)
+            )
+          else
+            alts2
```
odersky: That changes the logic of overloading resolution. With the change, we prefer methods that are applicable without implicit conversions to other methods. That's not as specced I think. What is the motivation for the change?

smarter: From the commit message of https://github.com/lampepfl/dotty/pull/1288/commits/7a060ba8ed52bc4a5762917c8f0cb1322542bc4a:
> If a directly applicable alternative exists, there is no need to try
every alternative (which would require searching for implicit
conversions for every argument that doesn't directly match). This should
not affect semantics since applicable overloads that require some
implicit conversions are dismissed in `narrowMostSpecific`

The motivation for this change is that I observed that many of the implicit searches when compiling dotty were caused by overloading resolution (e.g., every call to `Int#+` will check if the argument can be converted to `Char`,`Short`,`Double`, etc even if the argument is an `Int`)  It's possible that I missed a case where this does change the semantics, if so could you give an example?

odersky:
@smarter I was thinking something along these lines:
    
  trait A; trait B
    class C1 {
       def f(x: A): Unit
    }
    class C2 extends C1 {
       def f(x: B): Unit
    }
    object Test extends C2 {
      implicit def a2b(x: A): B = ???
      f(new A)
    }

Under the old rules this is ambiguous. Under the new rules, C1#f would win. Which is reasonable, but not the way I remember the spec. On the other hand, `scalac` also picks `C1#f`, so maybe that's what we should do anyway. Do you have an idea of the savings in compile rime this change gives us?

odersky:
Everything else LGTM

smarter:
> Do you have an idea of the savings in compile time this change gives us?

Not really, it seemed to shave at least a second on `dotty` but the measurements are very noisy, I'm fine with either keeping or dropping this change, in either case we should have a look at what the spec says

odersky:
The spec can be changed quite easily. And it's true that there are cases
where this could help a lot. So I think we should adopt the change.

On Fri, Jun 3, 2016 at 6:52 PM, Guillaume Martres <notifications@github.com>
wrote:

> Do you have an idea of the savings in compile time this change gives us?
>
> Not really, it seemed to shave at least a second on dotty but the
> measurements are very noisy, I'm fine with either keeping or dropping this
> change, in either case we should have a look at what the spec says
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly, view it on GitHub
> <https://github.com/lampepfl/dotty/pull/1288#issuecomment-223632562>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAwlVpjKfGtC8YHq2QgV91oj0MugoPbjks5qIFu7gaJpZM4IqmNI>
> .
>



-- 
Prof. Martin Odersky
LAMP/IC, EPFL

smarter:
OK, I'll add your example as a pos test to the commit that does the change and merge. Looking at the spec it seems we would just need to define what "directly applicable" means and use that in the overloading resolution secton

PR #1289, opened 2016-06-01
partest: Enable separate compilation

smarter:
partest can separately compile files based on their suffix (_1, _2,
...), it turns out that this feature was never enabled in the dotty
version of partest and no one noticed (it prints warnings in
./tests/partest-generated/gen.log which no one reads).

Enabling this revealed some latent separate compilation bugs that will need to
be fixed in the next few commits.

There are four (edit: actually more than that) test failures after this commit: https://gist.github.com/smarter/5cfff93170dae8864b9920e0d0398b63

I don't know if I'll have time to look at all of them, so it would be nice if people could assign themselves (ping @DarkDimius, @felixmulder, @liufengyun, @odersky, @VladimirNik) to try to fix one of them and push it to this PR (you can edit this issue to assign yourself):
- [x] `t1751` and `t294`
  - Assignee: @smarter
   - These are the same issue related to inference and Java annotations:
    ```scala
error: none of the overloaded alternatives of constructor SuiteClasses in class SuiteClasses with types
 (value: Class[_]*)SuiteClasses
 (value: Array[Class[_]])SuiteClasses
match arguments (Array[Nothing])
@SuiteClasses(Array())
^
```
- [X] `colltest5`
  - Assignee: @smarter
   - The value class `StringOps` has an overloaded method `++`, we replace overloading by numbers in the corresponding extension methods (`$extension0`, `$extension1`, ...) to avoid dealing with overloading after Typer (I'm no longer a fan of that scheme that we copied from scalac and considered replacing it by overloading). It seems to have gone wrong here (we emitted `$plus$plus$extension0(String, String)` when compiling the first file but when compiling the second file we looked for  `$plus$plus$extension1(String, String)`:
   ```scala
java.lang.NoSuchMethodError: strawman.collections.CollectionStrawMan5$StringOps$.$plus$plus$extension1(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;
```
- [X] `t7859`
  - Assignee: @VladimirNik 
  - The getter of `x` in the value class `A` should not be `private` because we need to access it from separate compilation units to unbox the value class:
  ```scala
java.lang.AssertionError: assertion failed: private getter x in class A in /home/smarter/opt/dotty/tests/partest-generated/run/t7859/A_1.scala accessed from value a in /home/smarter/opt/dotty/tests/partest-generated/run/t7859/B_2.scala
        at dotty.tools.dotc.transform.ExpandPrivate.ensurePrivateAccessible(ExpandPrivate.scala:75)
```
- [X] `t1186, t1235, t1254, t1642`
  - Assignee: @smarter
  - The classpath used for javac is incorrect
- [X] `volatile`
  - Assignee: @smarter
  - java filename should match public class name to make javac happy
- `t1263, t1745`
  - Assignee: ...
  - Emitted classfiles do not contain type parameter information used by javac, see #1303, moved to pending for now
- [x] `colltest4`
  - Assignee: @felixmulder
  - There is a bug in `CollectionStrawMan4`:
  ```scala
  object ListBuffer extends IterableFactory[ListBuffer] {
    def fromIterable[B](coll: Iterable[B]): ListBuffer[B] = coll match {
      case pd @ View.Partitioned(partition: View.Partition[B]) =>
        partition.distribute(new ListBuffer[B]())
        pd.forced.get.asInstanceOf[ListBuffer[B]]
```
  The cast is wrong, `pd.forced.get` returns a `List`, not a `ListBuffer`, this should be fixed both in `./tests/run/colltest4/CollectionStrawMan4_1.scala` and `./src/strawman/collections/CollectionStrawMan4.scala`, stacktrace: https://gist.github.com/smarter/5aab6857a20b0979b2909355171ab341
- [x] `t0288`
  - Assignee: @odersky 
  - This one seems tricky: it's a stale symbol error on the second run of dotty, the symbol is `test$Outer$Inner$$$outer` which is weird because Outer.Inner is a java inner class so it doesn't have an `$outer` field, someone familiar with how we deal with outer accessors should have a look.


src/dotty/tools/dotc/transform/ExpandPrivate.scala
```diff
@@ -93,6 +102,8 @@ class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { t
       case Apply(sel @ Select(_: Super, _), _)
       if sym.is(PrivateParamAccessor) && sel.symbol.is(ParamAccessor) && sym.name == sel.symbol.name =>
         sym.ensureNotPrivate.installAfter(thisTransform)
+      case _ if (isVCPrivateParamAccessor(sym)) =>
```
smarter: Just do:
```scala
case _ =>
  if (isVC...)
```
Instead of two `case _` cases

VladimirNik: Fixed. Thank you!

felixmulder:
@smarter: is this ready for review after the tests pass?

smarter:
Turns out there's more issues than the three I listed above, so not yet

smarter:
OK, after fixing another classpath issue and deferring to #1303 for two other tests, there is one remaining test failure:
- [ ] `t0288`
  - Assignee: ...
  - This one seems tricky: it's a stale symbol error on the second run of dotty, the symbol is `test$Outer$Inner$$$outer` which is weird because Outer.Inner is a java inner class so it doesn't have an `$outer` field, someone familiar with how we deal with outer accessors should have a look.

tests/run/colltest4/CollectionStrawMan4_1.scala
```diff
@@ -217,7 +217,7 @@ object CollectionStrawMan4 {
     def fromIterable[B](coll: Iterable[B]): ListBuffer[B] = coll match {
       case pd @ View.Partitioned(partition: View.Partition[B]) =>
         partition.distribute(new ListBuffer[B]())
-        pd.forced.get.asInstanceOf[ListBuffer[B]]
+        new ListBuffer[B] ++= pd.forced.get.asInstanceOf[List[B]]
```
smarter: You don't even need the cast, `pd.forced.get` works, but it would be interesting to know if this is what @odersky intented or if there's a design flaw somewhere. Also please squash your commits related to this into one commit

odersky: The code looks correct to me. The cast seems to be a left-over from a
previous version.

On Fri, Jun 3, 2016 at 2:45 PM, Guillaume Martres <notifications@github.com>
wrote:

> In tests/run/colltest4/CollectionStrawMan4_1.scala
> <https://github.com/lampepfl/dotty/pull/1289#discussion_r65700740>:
>
> > @@ -217,7 +217,7 @@ object CollectionStrawMan4 {
> >      def fromIterable[B](coll: Iterable[B]): ListBuffer[B] = coll match {
> >        case pd @ View.Partitioned(partition: View.Partition[B]) =>
> >          partition.distribute(new ListBuffer[B]())
> > -        pd.forced.get.asInstanceOf[ListBuffer[B]]
> > +        new ListBuffer[B] ++= pd.forced.get.asInstanceOf[List[B]]
>
> You don't even need the cast, pd.forced.get works, but it would be
> interesting to know if this is what @odersky <https://github.com/odersky>
> intented or if there's a design flaw somewhere. Also please squash your
> commits related to this into one commit
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/lampepfl/dotty/pull/1289/files/61ce7bb9be6a74d19b246e621f8fbc02581684e5..8f3b4bf6cc2927dc2b1d9e9515dd0451b5359233#r65700740>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAwlVoqkdv3GnuHJWa4SDmvsTQq6NGcfks5qICHWgaJpZM4IrGUX>
> .
>



-- 
Prof. Martin Odersky
LAMP/IC, EPFL

odersky:
Fixed issue with t0288 and rebased to latest master.

odersky:
@smarter Your commits LGTM. Would you like to review the rest?

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2645,6 +2645,9 @@ object Types {
       if (ctx.period != validSuper) {
         cachedSuper = tycon match {
           case tp: TypeLambda => defn.AnyType
+          case tp: TypeVar =>
+            // supertype not stable, since underlying might change
```
smarter: The supertype should be stable if `tp.inst.exists`

odersky: Good point. Fixed in latest commit.

smarter:
LGTM, good to have this in!

PR #1291, opened 2016-06-01
Add scala.Dynamic support.

nicolasstucki:


nicolasstucki:
Still need to add some tests and write some comments.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -553,6 +560,18 @@ trait Applications extends Compatibility { self: Typer =>
 
       fun1.tpe match {
         case ErrorType => tree.withType(ErrorType)
+        case TryDynamicCallType =>
+          tree match {
+            case Apply(Select(qual, name), args) if !isDynamicMethod(name) =>
+              if (untpd.isWildcardStarArgList(args)) {
+                ctx.error(name.toString + " does not support passing a vararg parameter", tree.pos)
+                tree.withType(ErrorType)
```
sjrd: Is there any reason to "preserve" that limitation?

nicolasstucki: The current implementation does what the scalac implementation does. Im not sure why the restriction was there in the first place. It could be just because it is harder to handle with `applyDynamicNamed`.

nicolasstucki:
All dynamic method (like `selectDynamic(name: String)`) receive as first parameter the name of the method. Currently it would be possible to define it as `selectDynamic(name: Int)` which will always generate an error at call site. I'm considering adding a new check that forces that first type to be a `String` at the declaration site and generates an error there (could also be a warning). Is anyone against adding that new restriction for some reason?

sjrd:
Yes, I'm against forcing `String`. You could have an implicit conversion from `String` to `T` that should kick in at that point. Scalac sort of supports that, although it's buggy for `updateDynamic`.

Speaking of which, it would be good to test that implicit conversions can be applied to the arguments at least. `js.Dynamic` is full of that requirement.

felixmulder:
This PR should solve #657 when merged

tests/run/dynamicDynamicTests.scala
```diff
@@ -1,3 +1,4 @@
+import dotty.language.dynamics
```
smarter: Unless this works very differently from dynamic in scalac, I think we should use `import scala.language.dynamics` for compatibility

nicolasstucki: I agree.

tests/neg/dynamicApplyDynamicTest8.scala
```diff
@@ -0,0 +1,9 @@
+import scala.language.dynamics
+
+class Foo extends scala.Dynamic {
+  def applyDynamicNamed(name: String)(args: (Sting, Int)*): String = ???
+}
+
+object DynamicTest {
+  def test: String = new Foo().bazApply("1" -> 2)  // error // error
```
nicolasstucki: Note to self: remove one of the `// error`s

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3021,6 +3021,11 @@ object Types {
 
   object ErrorType extends ErrorType
 
+  /* Type used to track Select nodes that could not resolve a member and their qualifier is a scala.Dynamic. */
+  abstract class TryDynamicCallType extends UncachedGroundType with ValueType
+
+  object TryDynamicCallType extends TryDynamicCallType
+
```
odersky: What happens if we try to map or fold over such a type? Did we think of that?

src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -68,6 +69,30 @@ trait Reporting { this: Context =>
   def featureWarning(msg: => String, pos: SourcePosition = NoSourcePosition): Unit =
     reporter.report(new FeatureWarning(msg, pos))
 
+  private[this] var reportedFeatures = Set[Symbol]()
```
odersky: Global vars are disallowed since they will cause races for parallel compilation.

odersky: This will make reportedFeatures part of every context (and there are LOTS of those). I think it would be better to make the var a part of the Reporter class.

src/dotty/tools/dotc/core/Types.scala
nicolasstucki: The idea is that it behaves like `ErrorType` on any operation until it is caught and handled. I considered making it a subtype of `ErrorType` but this seamed a bit dangerous and with this alternative there was only one place where I had to add it to a `case` to do the same as `ErrorType`.

nicolasstucki: Actually, now that I think about it, it should be `object TryDynamicCallType extends ErrorType`.

src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -206,6 +230,10 @@ abstract class Reporter extends interfaces.ReporterResult {
    */
   def errorsReported = hasErrors
 
+  private[this] var reportedFeaturesUseSites = Set[Symbol]()
```
odersky: It would be good to have a motivation why this is needed. In fact I am not yet sure we want to suppress feature warnings per use site.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -541,7 +542,22 @@ trait Applications extends Compatibility { self: Typer =>
...
+       * Mode.IgnoreNextDynamic will ensure that Apply(Select(foo, bar), baz) is not transformed
+       * into foo.selectDynamic(bar).apply(args).
+       * We remove Mode.IgnoreNextDynamic if the name of the called method if is one of the scala.Dynamic methods or
+       * if this node was generated by an assign (see typedAssign).
+       * See: Dynamic.scala
+       */
+      def generatedByTypedAssign(name: Name) = name == nme.update &&  ctx.mode.is(Mode.IgnoreNextDynamic)
+      val ctx1 = tree.fun match {
+        case Select(_, name) if isDynamicMethod(name) || generatedByTypedAssign(name) => ctx.retractMode(Mode.IgnoreNextDynamic)
+        case _                                                                        => ctx.addMode(Mode.IgnoreNextDynamic)
```
odersky: That's a bit heavy (both computationally, addMode is not super-cheap, and in terms for complexity). Is there another way to achieve this?

src/dotty/tools/dotc/reporting/Reporter.scala
nicolasstucki: Currently it does not suppress the warning. It is only there to make sure the explanation of how to avoid it is only printed once per site.

src/dotty/tools/dotc/typer/Applications.scala
nicolasstucki: Commit `e33e20a` removes the mode and adds `AssignProto` to make the checks directly on prototypes. This commit should be squashed before merging.

src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -430,6 +430,31 @@ object ProtoTypes {
...
+
+  /** A prototype for expressions that appear in the lhs position of an assign
+   *
+   *  lhs = (rhs: resType)
+   */
+  case class AssignProto(lhs: untpd.Tree, rhsType: Type, typer: Typer)(implicit ctx: Context)
+      extends UncachedGroundType with AssigningProto {
+    override def resultType(implicit ctx: Context) = rhsType
+
+    def isMatchedBy(tp: Type)(implicit ctx: Context) = ???
```
nicolasstucki: @odersky, i'm not sure what to return `AssignProto. isMatchedBy`.

src/dotty/tools/dotc/reporting/Reporter.scala
odersky: Ah, OK, I had missed that.

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -196,11 +196,16 @@ trait TypeAssigner {
   def selectionType(site: Type, name: Name, pos: Position)(implicit ctx: Context): Type = {
     val mbr = site.member(name)
     if (reallyExists(mbr)) site.select(name, mbr)
-    else {
+    else if (site <:< defn.DynamicType && !Dynamic.isDynamicMethod(name)) {
```
odersky: Cheaper test is: `site.derivesFrom(defn.DynamicClass)`

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -196,11 +196,16 @@ trait TypeAssigner {
...
+      TryDynamicCallType
+    } else {
       if (!site.isErroneous) {
         ctx.error(
           if (name == nme.CONSTRUCTOR) d"$site does not have a constructor"
-          else d"$name is not a member of $site", pos)
+          else if (site <:< defn.DynamicType) {
```
odersky: same here

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -316,7 +317,15 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
       if (tree.name.isTypeName) checkStable(qual1.tpe, qual1.pos)
-      typedSelect(tree, pt, qual1)
+      val select = typedSelect(tree, pt, qual1)
+      if (select.tpe != TryDynamicCallType) {
+        select
+      } else {
+        pt match {
+          case _: FunProto | _: AssignProto => select
+          case _                            => typedDynamicSelect(tree, pt)
+        }
```
odersky: I'd refactor to

    pt match {
       case _: FunProto | _: AssignProto => select
       case _  => 
         if (select.tpe eq TryDynamicCallType) typedDynamicSelect(tree, pt)
         else select
    }

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1079,7 +1095,13 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       .withType(dummy.nonMemberTermRef)
     checkVariance(impl1)
     if (!cls.is(AbstractOrTrait) && !ctx.isAfterTyper) checkRealizableBounds(cls.typeRef, cdef.pos)
-    assignType(cpy.TypeDef(cdef)(name, impl1, Nil), cls)
+    val cdef1 = assignType(cpy.TypeDef(cdef)(name, impl1, Nil), cls)
+    if (ctx.phase.isTyper && cdef1.tpe <:< defn.DynamicType && !ctx.dynamicsEnabled) {
```
odersky: again, can use `derivesFrom` instead of `<:<`

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1079,7 +1095,13 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+      val isRequired = parents1.exists(_.tpe =:= defn.DynamicType)
```
odersky: Cheaper test: `(_.tpe.isRef(defn.DynamicClass)`.

odersky: I was wondering about the isRequired logic. Is that how scalac does it?

src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -430,6 +430,31 @@ object ProtoTypes {
       (if (theMap != null) theMap else new WildApproxMap).mapOver(tp)
   }
 
+  trait AssigningProto extends ProtoType
+
```
odersky: I wonder whether you need the trait and the case class. Instead of all this, maybe you can do it as with AnyFunctionProto:

    @sharable object AssignProto extends UncachedGroundType with MatchAlways

Can you try this and see whether it would work?

src/dotty/tools/dotc/typer/Typer.scala
nicolasstucki: I reversed engineered that from the warning in scalac of all the test cases I have. As far as I can tell the conditions in scalac are equivalent but checked in a different way. I might also have taken some input from @sjrd of how they are expected to behave from the use cases in Scala.js.

nicolasstucki:
Updated

odersky:
LGTM, thanks!

PR #1296, opened 2016-06-02
add scala library as a submodule

### LINKED ISSUE #1290: SBT: report a warning if ./scala-scala is absent
Dotty test suite would fail if one didn't clone compatible version of stdlib.

It would be nice to give a warning for people running either `test` or `partests` suggesting the to do:

```
git clone -b dotty-library https://github.com/DarkDimius/scala.git ./scala-scala
```

liufengyun:
To synchronize the submodule:

    git submodule init
    git submodule update

Reference: https://git-scm.com/book/en/v2/Git-Tools-Submodules

Review @DarkDimius .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,9 +32,6 @@ tests/partest-generated/
 tests/locks/
 /test-classes/
 
-# Used in tests
-scala-scala
-
 # Ignore output files but keep the directory
 out/
 !out/.keep
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "scala-scala"]
+	path = scala-scala
+	url = https://github.com/DarkDimius/scala.git
+	branch = dotty-library


DarkDimius:
I'm fine with making it a submodule, butI don't think it fixes #1290, as people still need to *know* that they need to issue additional commands.

liufengyun:
I agree, while adding submodule makes the dependency explicit, an additional warning message is even better.

sjrd:
In my experience, it is easier to download it directly as part of the build. It causes less headache. Of course that's also because in Scala.js we have to download the appropriate version of Scala wrt the version we're building Scala.js for. So actually we end up with one subdirectory for each version of Scala we build against.

In dotty's case, there's only ever one specific version, so it's easier, and the submodule idea can work without any particular hassle.

DarkDimius:
Note, that branch submodules were added in Git 1.8.2, which was realeased in 2013. 
Git version prior to that will silently clone master and fail the test suite later.

PR #1297, opened 2016-06-02
Fix #1292: give position when failing to emit switch on annotated match

### LINKED ISSUE #1292: failed to emit @switch should give position
When compiling dotty with dotty, this warning shows up twice but without a position to indicate from where it is coming.

felixmulder:
This fix gives the position and refactors the code that gave off
warnings, but it also begs the question - should we be able to handle
emitting a switch for the following case:

```scala
(x: @switch) match {
  case 'a' if somePredicate(x) => // ...
  case 'b' => // ...
}
```

Currently if there is a guard, the patternmatcher will fail to emit
a switch. Scalac can handle these cases currently.

review: @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/parsing/Scanners.scala
+++ b/src/dotty/tools/dotc/parsing/Scanners.scala
@@ -922,8 +922,8 @@ object Scanners {
           getFraction()
         }
       } else (ch: @switch) match {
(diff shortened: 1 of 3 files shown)


DarkDimius:
LGTM

PR #1298, opened 2016-06-02
Switch organization from org.scala-lang to ch.epfl.lamp

smarter:
Publishing artifacts under ch.epfl.lamp is easier.

Review by @DarkDimius


DarkDimius:
LGTM. We've already got this group id from sonatype.

PR #1302, opened 2016-06-02
Properly report errors when cli flags are malformed

smarter:
Previously we returned an empty Reporter with no errors so partest
reported the test as a success.

Review by @DarkDimius


DarkDimius:
LGTM

Issue #1264, opened 2016-05-19
partest fails when testing a file that is also tested as part of a directory test


smarter:
@odersky originally noticed this issue while trying to add a test for `core/Types.scala` with different options than the `dotc_core` test.
To reproduce:

``` shell
$ mkdir dummy
$ echo "object Test { def main(args: Array[String]): Unit = {} }" > dummy/A.scala
```

Then in `test/dotc/tests.scala`:

``` scala
@Test def dummy1 = compileDir("./", "dummy")
@Test def dummy2 = compileFile("./", "dummy/A")
```

If you then run something like:

``` shell
$ sbt
> partest-only-no-bootstrap --verbose tests/partest-generated/dummy
```

You'll get a "compiled twice" error which can be explained by looking at the partest-generated folder:

``` shell
$ ls tests/partest-generated/pos/dummy/*.scala
A.scala  A_v1.scala
```

`CompilerTest#computeDestAndCopyFiles` copied `A.scala` to `A_v1.scala` to avoid a clash, but this assumes that `A.scala` is tested individually, whereas it is tested as part of the directory `dummy`, the result is that when we run the `dummy1` test, we try to compile both `A.scala` and `A_v1.scala`.

This should be fixable by changing `CompilerTest#computeDestAndCopyFiles` to always copy directories instead of files, we should have `tests/partest-generated/pos/dummy` and `tests/partest-generated/pos/dummy_v1`.

smarter:
We do not use partest anymore.

PR #1304, opened 2016-06-02
partest: correct line numbers in run stack traces

smarter:
partest adds a warning in a comment at the beginning of source files
that it copies, but this means that every line number displayed in a
stack trace is offset by 6. We can workaround this by moving the warning
to the end of the file.


DarkDimius:
The reason why this header is in the beggining of the file is to make it apparent that people should not modify the generated files.

I believe that moving the header to the end of the file defeats the purpose.

smarter:
> The reason why this header is in the beggining of the file is to make it apparent that people should not modify the generated files.

I understand that, but:
- The files are in a directory `partest-generated` which makes it obvious that they're generated
- Most test files are small so you should still be able to see the warning at the end
- If your text editor is not terrible, you should be able to recover an overwritten file
- Having wrong line number information can make you waste a lot of time trying to fix a bug in the incorrect place, I believe that's much worse thing than having your test file overwritten
- If we wanted to really fix this we should fix or rewrite partest so that we don't need to generate files and so that we don't need so many layers of abstractions which lead to hard to debug issues like https://github.com/lampepfl/dotty/issues/1264. This won't happen soon, but I want correct line information now, hence this patch.

DarkDimius:
Proposal from the dotty meeting: change the comment to still be a prepend, but be a single line prepend instead with scary amount of exclamation marks.

smarter:
That looks like a good solution to me since you shouldn't look at generated tests anyway, I'll update my PR

DarkDimius:
Please reopen this PR when you'll update it. Thanks.

smarter:
Looks like I can't reopen a pr after force-pushing a branch, so I'll open a new one

PR #1306, opened 2016-06-04
Merge the sbt compiler bridge as a subproject of dotty

smarter:
Note that if we want to run the bridge tests on every PR we should add the following to a Jenkins job:
```
sbt bridge/test
sbt bridge/scripted
```
The scripted tests take a fair amount of time (27 minutes here) so we may not want to do that right now, this will get better once we base ourselves on sbt 1.0.


smarter:
Review by @felixmulder

felixmulder:
Aside from those two comments LGTM

felixmulder:
LGTM (once those two comments have been answered)

felixmulder:
Also, great job! I think we can all agree that you're the MVP of the week! :tada:

DarkDimius:
I'd prefer to have commits in this PR squashed for several reasons:
 - they do not show up on github;
 - they make my local tooling(SourceTree) go into infinite cycle;
 - the last commit in any case screws the blame history(at least in my IDE) so I don't see much value in maintaining it given that it breaks tooling(both on github and locally).

Otherwise LGTM.

smarter:
> the last commit in any case screws the blame history
`git blame bridge/src/main/scala/xsbt/CompilerInterface.scala` seems to display the correct thing.

Anyway, seems like you're trying to use a bunch of broken tools, feel free to change this PR to make them work with your tools, but maybe you should change tools ;)

smarter:
If you do squash the commits, the commit message should at least include a link to https://github.com/smarter/dotty-bridge so that people can find the real history.

PR #1314, opened 2016-06-07
Add console to bridge

felixmulder:
This PR does two things:

1. Adds the ability to use `console` from within an sbt-session that is using the dotty-bridge
1. Creates an aggregated subproject in the build file so that we may publish the correct snapshot artifacts using `sbt dotty-core/publishLocal` or to deploy real snapshots: `sbt dotty-core/publish`

### DIFF
--- a/bridge/src/main/scala/xsbt/ConsoleInterface.scala
+++ b/bridge/src/main/scala/xsbt/ConsoleInterface.scala
@@ -1,0 +1,68 @@
+/* sbt -- Simple Build Tool
+ * Copyright 2008, 2009 Mark Harrah
+ */
+package xsbt
+
+import xsbti.Logger
+import scala.tools.nsc.{ GenericRunnerCommand, Interpreter, InterpreterLoop, ObjectRunner, Settings }
+import scala.tools.nsc.interpreter.InteractiveReader
+import scala.tools.nsc.reporters.Reporter
+import scala.tools.nsc.util.ClassPath
+
+import dotty.tools.dotc.repl.REPL
+import dotty.tools.dotc.repl.REPL.Config
+
+class ConsoleInterface {
+  def commandArguments(
+    args: Array[String],
+    bootClasspathString: String,
+    classpathString: String,
+    log: Logger
+  ): Array[String] = args
+
+  def run(args: Array[String],
+          bootClasspathString: String,
+          classpathString: String,
+          initialCommands: String,
+          cleanupCommands: String,
+          loader: ClassLoader,
+          bindNames: Array[String],
+          bindValues: Array[Any],
+          log: Logger
+  ): Unit = {
+    val completeArgs =
+      args                                    :+
+      "-bootclasspath" :+ bootClasspathString :+
+      "-classpath"     :+ classpathString
+
+    println("Starting dotty interpreter...")
+    val repl = ConsoleInterface.customRepl(
+      initialCommands :: Nil,
+      cleanupCommands :: Nil,
+      bindNames zip bindValues
+    )
+    repl.process(completeArgs)
+  }
+}
+
+object ConsoleInterface {
+  def customConfig(
+    initCmds: List[String],
+    cleanupCmds: List[String],
+    boundVals: Array[(String, Any)]
+  ) = new Config {
+    override val initialCommands: List[String] = initCmds
+    override val cleanupCommands: List[String] = cleanupCmds
+    override val boundValues: Array[(String, Any)] = boundVals
+  }
+
+  def customRepl(cfg: Config): REPL = new REPL {
+    override lazy val config = cfg
+  }
+
+  def customRepl(
+    initCmds: List[String],
+    cleanupCmds: List[String],
+    boundVals: Array[(String, Any)]
+  ): REPL = customRepl(customConfig(initCmds, cleanupCmds, boundVals))
+}
--- a/bridge/src/main/scala/xsbt/ConsoleInterface.scala
+++ b/bridge/src/main/scala/xsbt/ConsoleInterface.scala
@@ -9,6 +9,7 @@ import scala.tools.nsc.interpreter.InteractiveReader
 import scala.tools.nsc.reporters.Reporter
 import scala.tools.nsc.util.ClassPath
 
+import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.repl.REPL
 import dotty.tools.dotc.repl.REPL.Config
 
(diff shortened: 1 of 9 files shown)


bridge/src/main/scala/xsbt/ConsoleInterface.scala
```diff
@@ -0,0 +1,68 @@
...
+    classpathString: String,
+    log: Logger
+  ): Array[String] = args
+
+  def run(args: Array[String],
+          bootClasspathString: String,
+          classpathString: String,
+          initialCommands: String,
+          cleanupCommands: String,
+          loader: ClassLoader,
```
smarter: `loader` isn't passed to the REPL, is that a TODO? If yes that should be noted in a comment

bridge/src/main/scala/xsbt/ConsoleInterface.scala
```diff
@@ -0,0 +1,68 @@
...
+import dotty.tools.dotc.repl.REPL
+import dotty.tools.dotc.repl.REPL.Config
+
+class ConsoleInterface {
+  def commandArguments(
+    args: Array[String],
+    bootClasspathString: String,
+    classpathString: String,
+    log: Logger
+  ): Array[String] = args
```
smarter: Did you figure out what this method is for? Could you add a comment explaining why it makes sense to return just `args` here?

bridge/src/main/scala/xsbt/ConsoleInterface.scala
```diff
@@ -0,0 +1,68 @@
...
+    cleanupCmds: List[String],
+    boundVals: Array[(String, Any)]
+  ) = new Config {
+    override val initialCommands: List[String] = initCmds
+    override val cleanupCommands: List[String] = cleanupCmds
+    override val boundValues: Array[(String, Any)] = boundVals
+  }
+
+  def customRepl(cfg: Config): REPL = new REPL {
+    override lazy val config = cfg
```
smarter: Why lazy?

felixmulder: Must be lazy to override lazy val ;)

project/Build.scala
```diff
@@ -349,6 +351,12 @@ object DottyInjectedPlugin extends AutoPlugin {
      )
    )
 
+   lazy val `dotty-core` = project
```
smarter: I think the root project should be the one that aggregates the others (like in Ensime: https://github.com/ensime/ensime-server/blob/2.0/project/EnsimeBuild.scala#L243) but I don't know if that requires doing something special to get `compile`, `run`, etc to do the correct thing

bridge/src/main/scala/xsbt/ConsoleInterface.scala
felixmulder: In the scalac version it looks like it's just setting up the state of the compiler during this method call. Not sure if that makes sense for dotty...

bridge/src/main/scala/xsbt/ConsoleInterface.scala
felixmulder: Good point, it isn't necessary to pass the loader - but if a modified loader is passed somehow, it makes sense that it be propagated to the REPL. Will add that in the morning :) thanks!

project/Build.scala
felixmulder: That is the intention of `dotty-core`, once @sjrd is done with the restructuring this could simply be renamed to `dotty`.

src/dotty/tools/dotc/repl/CombinedClassLoader.scala
```diff
@@ -0,0 +1,37 @@
+package dotty.tools
+package dotc
+package repl
+
+import scala.util.Try
+
+class CombinedClassLoader(loaders: List[ClassLoader]) extends ClassLoader {
```
felixmulder: Yes, this guy needs to go away - not sure how yet though. Every other solution seems to create a circular dependency on sbt 👎

smarter: Why is this needed? Does scalac do something similar?

felixmulder: They just set it as the class loader - but that doesn't work for us. So I decided that it would be a good idea to merge the ClassLoaders and look through both of them when searching for classes.

But I don't like this solution - if you have a better one - I'm all ears.

smarter: Why doesn't it work for us? Can't we replace `parentClassLoader` in `val classLoader` with the classloader given by sbt?

smarter: In general, I would advise against making a custom subclass of ClassLoader because it can very easily not do what you think it should do, and it's hard to debug, but I don't know enough about the situation here to say for sure.

felixmulder:
This feels stable enough for review now - @DarkDimius

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -153,8 +155,10 @@ class CompilingInterpreter(out: PrintWriter, ictx: Context) extends Compiler wit
   */
   /** class loader used to load compiled code */
   val classLoader: ClassLoader = {
-    val parent = new URLClassLoader(compilerClasspath.toArray, parentClassLoader)
-    new AbstractFileClassLoader(virtualDirectory, parent)
+    lazy val parent = new URLClassLoader(compilerClasspath.toArray,
+                                         classOf[Interpreter].getClassLoader)
```
DarkDimius: why `Interpreter` instead of `CompilingInterpreter`?

felixmulder: Good question - I think this was originally written by @odersky

project/Build.scala
```diff
@@ -207,11 +209,15 @@ object DottyBuild extends Build {
         "org.scala-sbt" % "api" % sbtVersion.value % "test",
         "org.specs2" %% "specs2" % "2.3.11" % "test"
       ),
-      version := "0.1.1-SNAPSHOT",
-      // The sources should be published with crossPaths := false, the binaries
-      // are unused so it doesn't matter.
+      version :=
+        "0.1.1-" + VersionUtil.commitDate + "-" + VersionUtil.gitHash + "-SNAPSHOT",
```
DarkDimius: I'm not sure if it is a good idea to publish hundreds of distinct snapshots on maven and ask users to specify git-hash by hand. I'd prefer to publish one snapshot and have version information reported by REPL. Something similar to https://github.com/hmrc/sbt-git-stamp may be in handy for implementing this.

DarkDimius:
LGTM, otherwise.
I prefer to fast-track this PR, as publishing snapshots depends on it.

PR #1315, opened 2016-06-10
Fix #856: Handle try/catch cases as catch cases if possible.

### LINKED ISSUE #856: Optimize handling of `catch` blocks
Current code emitted by pattern-matcher is inefficient. 
What it currently does is:

```
try {<code>} 
catch { <cases>}
```

is compiled as if it this was written:

```
try {<code>} catch {
  case e => e 
    match {<cases>; case _ => throw e}
  }
```

nicolasstucki:
Previously they were all lifted into a match with the came cases.
Now the first cases are handled directly by by the catch. If one
of the cases can not be handled the old scheme is applied to to it
and all subsequent cases.

### DIFF
--- a/src/dotty/tools/dotc/transform/TryCatchPatterns.scala
+++ b/src/dotty/tools/dotc/transform/TryCatchPatterns.scala
@@ -1,0 +1,99 @@
+package dotty.tools.dotc
+package transform
+
+import core.Symbols._
+import core.StdNames._
+import ast.Trees._
+import core.Types._
+import dotty.tools.dotc.core.Decorators._
+import dotty.tools.dotc.core.Flags
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+import dotty.tools.dotc.util.Positions.Position
+
+/** Compiles the cases that can not be handled by primitive catch cases as a common pattern match.
+ *
+ *  The following code:
+ *    ```
+ *    try { <code> }
+ *    catch {
+ *      <tryCases> // Cases that can be handled by catch
+ *      <patternMatchCases> // Cases starting with first one that can't be handled by catch
+ *    }
+ *    ```
+ *  will become:
+ *    ```
+ *    try { <code> }
+ *    catch {
+ *      <tryCases>
+ *      case e => e match {
+ *        <patternMatchCases>
+ *      }
+ *    }
+ *    ```
+ *
+ *  Cases that are not supported include:
+ *   - Applies and unapplies
+ *   - Idents
+ *   - Alternatives
+ *   - `case _: T =>` where `T` is not `Throwable`
+ *
+ */
+class TryCatchPatterns extends MiniPhaseTransform {
+  import dotty.tools.dotc.ast.tpd._
+
+  def phaseName: String = "tryCatchPatterns"
+
+  override def runsAfter = Set(classOf[ElimRepeated])
+
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = tree match {
+    case Try(_, cases, _) =>
+      cases.foreach {
+        case CaseDef(Typed(_, _), guard, _) => assert(guard.isEmpty, "Try case should not contain a guard.")
+        case CaseDef(Bind(_, _), guard, _) => assert(guard.isEmpty, "Try case should not contain a guard.")
+        case c =>
+          assert(isDefaultCase(c), "Pattern in Try should be Bind, Typed or default case.")
+      }
+    case _ =>
+  }
+
+  override def transformTry(tree: Try)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    val (tryCases, patternMatchCases) = tree.cases.span(isCatchCase)
+    val fallbackCase = mkFallbackPatterMatchCase(patternMatchCases, tree.pos)
+    cpy.Try(tree)(cases = tryCases ++ fallbackCase)
+  }
+
+  /** Is this pattern node a catch-all or type-test pattern? */
+  private def isCatchCase(cdef: CaseDef)(implicit ctx: Context): Boolean = cdef match {
+    case CaseDef(Typed(Ident(nme.WILDCARD), tpt), EmptyTree, _)          => isSimpleThrowable(tpt.tpe)
+    case CaseDef(Bind(_, Typed(Ident(nme.WILDCARD), tpt)), EmptyTree, _) => isSimpleThrowable(tpt.tpe)
+    case _                                                               => isDefaultCase(cdef)
+  }
+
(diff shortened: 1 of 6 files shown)


src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  override def transformTry(tree: Try)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    val (tryCases, patternMatchCases) = tree.cases.span(isCatchCase)
+    val fallbackCase = mkFallbackPatterMatchCase(patternMatchCases, tree.pos)
+    cpy.Try(tree)(cases = tryCases ++ fallbackCase)
+  }
+
+  /** Is this pattern node a catch-all or type-test pattern? */
+  private def isCatchCase(cdef: CaseDef)(implicit ctx: Context): Boolean = cdef match {
+    case CaseDef(Typed(Ident(nme.WILDCARD), tpt), _, _)          => isSimpleThrowable(tpt.tpe)
```
DarkDimius: what about guards? Since this PR they will not be desugared by patmat, and will reach the backend and be ignored there.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+ *   - Idents
+ *   - Alternatives
+ *   - `case _: T =>` where `T` is not `Throwable`
+ *
+ */
+class TryCatchPatterns extends MiniPhaseTransform {
+  import dotty.tools.dotc.ast.tpd._
+
+  def phaseName: String = "tryCatchPatterns"
+
```
DarkDimius: it would be nice to add a postcondition here: 
 - cases in tries have no guards
 - cases in tries are either binds or type-tests.

tests/run/tryPatternMatch.check
```diff
@@ -0,0 +1,9 @@
+IllegalArgumentException: abc
+IllegalArgumentException
+IOException
```
DarkDimius: this shows a bug) `IOException` should NOT be printed in this testcase(guard is false))

tests/run/tryPatternMatch.scala
```diff
@@ -0,0 +1,53 @@
...
+object Test {
+  def main(args: Array[String]): Unit = {
+    var a: Int = 1
+
+    def testThrow(throwIt: => Unit): Unit = {
+      try {
+        throwIt
+      } catch {
+        // These cases will be compiled as catch cases
+        case e: IOException if e.getMessage == null  => println("IOException")
```
DarkDimius: this guard.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+  private def isCatchCase(cdef: CaseDef)(implicit ctx: Context): Boolean = cdef match {
+    case CaseDef(Typed(Ident(nme.WILDCARD), tpt), _, _)          => isSimpleThrowable(tpt.tpe)
+    case CaseDef(Bind(_, Typed(Ident(nme.WILDCARD), tpt)), _, _) => isSimpleThrowable(tpt.tpe)
+    case _                                                       => isDefaultCase(cdef)
+  }
+
+  private def isSimpleThrowable(tp: Type)(implicit ctx: Context): Boolean = tp match {
+    case tp @ TypeRef(pre, _) =>
+      val sym = tp.symbol
+      (pre == NoPrefix || pre.widen.typeSymbol.isStatic) &&
```
DarkDimius: I would add a comment expaining why you're doing this check(i guess you're checking that it wouldn't need outer-class check. Am I right?)

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+      !sym.is(Flags.Trait) &&
```
DarkDimius: why don't you support traits?

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+      tp <:< defn.ThrowableType
```
DarkDimius: tp.derivesFrom(defn.ThrowableClass) is faster.

DarkDimius:
@nicolasstucki please ping me when ready for next review round. Thanks.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
nicolasstucki: Because the JVM does not support them. At least this is true for the case that I added in the tests.

DarkDimius: yes, you are right. `Throwable` is a class.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,89 @@
...
+import dotty.tools.dotc.util.Positions.Position
+
+/** Compiles the cases that can not be handled by primitive catch cases as a common pattern match.
+ *
+ *  The following code:
+ *    ```
+ *    try { <code> }
+ *    catch {
+ *      <tryCases> // Cases that can be handled by catch
+ *      <paterMatchCases> // Cases starting with first one that can't be handled by catch
```
smarter: Typo: patternMatchCaes

smarter: Oh, same elsewhere, so I guess it's not a typo but intentional, still seems a bit weird

nicolasstucki: It was indeed a typo. Must have copy pasted it around.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
nicolasstucki: done

nicolasstucki:
@DarkDimius this PR is ready for another round of review.

src/dotty/tools/dotc/transform/TryCatchPatterns.scala
```diff
@@ -0,0 +1,99 @@
...
+  import dotty.tools.dotc.ast.tpd._
+
+  def phaseName: String = "tryCatchPatterns"
+
+  override def runsAfter = Set(classOf[ElimRepeated])
+
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = tree match {
+    case Try(_, cases, _) =>
+      cases.foreach {
+        case CaseDef(Typed(_, _), guard, _) => assert(guard.isEmpty, "Try case should not a guard.")
```
DarkDimius: A synonym of `have` is missing. Same on the next line.

DarkDimius:
LGTM, Thanks!

PR #1325, opened 2016-06-20
better handling of overloaded extractors

### LINKED ISSUE #1318: Type checking error in patmat (related to overloaded unapply)
Following code compiles fine in scalac:
```Scala
object Test {
  class S(i: Int)
  case class T(i: Int) extends S(i)

  object T {
    def unapply(s: S): Option[(Int, Int)] = Some(5, 6)
  }

  val s = new S(5)

  s match {
    case T(x, y) => println(x + y)
    case _ => println("not match")
  }
}
```

But it fails with dotty:
```
examples/auto-tupling2.scala:13: error: Pattern type (T1, T2) is neither a subtype nor a supertype of selector type Int
    case T(x, y) => println(x + y)
           ^
examples/auto-tupling2.scala:13: error: type mismatch:
 found   : Nothing(x)
 required: ?{ + : ? }
 both method Byte2byte in object Predef$ and method Character2char in object Predef$ convert from Nothing(x) to ?{ + : FunProto(y):Any }
    case T(x, y) => println(x + y)
                            ^
two errors found
```

Following change seems to fix the problem:

``` Scala
--- a/src/dotty/tools/dotc/typer/Applications.scala
+++ b/src/dotty/tools/dotc/typer/Applications.scala
@@ -709,11 +709,11 @@ trait Applications extends Compatibility { self: Typer =>
       // try first for non-overloaded, then for overloaded ocurrences
       def tryWithName(name: TermName)(fallBack: Tree => Tree)(implicit ctx: Context): Tree =
         tryEither {
-          implicit ctx => typedExpr(untpd.Select(qual, name), genericProto)
+          implicit ctx => typedExpr(untpd.Select(qual, name), specificProto)
         } {
           (sel, _) =>
             tryEither {
-              implicit ctx => typedExpr(untpd.Select(qual, name), specificProto)
+              implicit ctx => typedExpr(untpd.Select(qual, name), genericProto)
             } {
               (_, _) => fallBack(sel)
             }
```

What's the consideration to try the generic first, then the specific one?

liufengyun:
Fix #1318 , review @DarkDimius 

This fix only works in limited cases. For example, if we remove the comments in the test `tests/pos/i1318.scala`, compilation still fails.

It seems that a more general and correct approach is to use a try and backtrack approach to find the first overloaded `unapply` that type checks. I'm not sure if it's good to follow that approach.

### DIFF
--- a/src/dotty/tools/dotc/typer/Applications.scala
+++ b/src/dotty/tools/dotc/typer/Applications.scala
@@ -709,11 +709,11 @@ trait Applications extends Compatibility { self: Typer =>
       // try first for non-overloaded, then for overloaded ocurrences
       def tryWithName(name: TermName)(fallBack: Tree => Tree)(implicit ctx: Context): Tree =
         tryEither {
-          implicit ctx => typedExpr(untpd.Select(qual, name), genericProto)
+          implicit ctx => typedExpr(untpd.Select(qual, name), specificProto)
         } {
           (sel, _) =>
             tryEither {
-              implicit ctx => typedExpr(untpd.Select(qual, name), specificProto)
+              implicit ctx => typedExpr(untpd.Select(qual, name), genericProto)
             } {
               (_, _) => fallBack(sel)
             }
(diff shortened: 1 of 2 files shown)


DarkDimius:
LGTM.

PR #1331, opened 2016-06-24
fix incorrect widening in match #1330

### LINKED ISSUE #1330: Incorrect widening in typing Match and Annotated type
Following code should type check, but it failed:

``` Scala
object Test {
  val ok = "ok"
  val s: ok.type = ok match { case x => x }
  val ts: ok.type = (ok : @unchecked) match { case x => x }

  def foo: PartialFunction[Int, Unit] = {
    case x => ()
  }
}
```

Following change **almost** fixed the problem if tree checking is disabled:

``` Scala
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -722,8 +722,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         typed(desugar.makeCaseLambda(tree.cases, protoFormals.length) withPos tree.pos, pt)
       case _ =>
         val sel1 = typedExpr(tree.selector)
-        val selType = widenForMatchSelector(
-            fullyDefinedType(sel1.tpe, "pattern selector", tree.pos))
+        val selType = fullyDefinedType(sel1.tpe, "pattern selector", tree.pos)

         val cases1 = typedCases(tree.cases, selType, pt)
         val cases2 = harmonize(cases1).asInstanceOf[List[CaseDef]]
@@ -1173,7 +1172,8 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     if (ctx.mode is Mode.Type)
       assignType(cpy.Annotated(tree)(annot1, arg1), annot1, arg1)
     else {
-      val tpt = TypeTree(AnnotatedType(arg1.tpe.widen, Annotation(annot1)))
+      val tpt = TypeTree(AnnotatedType(arg1.tpe, Annotation(annot1)))
+      println("I'm here:" + tpt)
       assignType(cpy.Typed(tree)(arg1, tpt), tpt)
     }
   }
```

Tree checking shows that the type of the identifier `_` in `isDefinedAt` refers to the variable `x$1` in `$anonfun`, which caused an undefined symbol assertion failure:

``` Scala
  def $anonfun(x$1: Int): Unit =
    x$1 match {
      case x @ _ =>
        ()
    }
  def isDefinedAt(x$1: Int): Boolean =
    x$1 match {
      case x @ _ => true
      case _ @ _ => false
    }
```

The error is introduced in **expandSAM**, I'm going to fix this error.

liufengyun:
Fix #1330 . Review @DarkDimius .

### DIFF
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -176,9 +176,10 @@ object desugar {
             name = meth.name.defaultGetterName(n),
             tparams = meth.tparams.map(tparam => dropContextBound(toDefParam(tparam))),
             vparamss = takeUpTo(normalizedVparamss, n),
-            tpt = TypeTree(),
+            tpt = vparam.tpt,
             rhs = vparam.rhs
           ).withMods(Modifiers(mods.flags & AccessFlags, mods.privateWithin))
+
         val rest = defaultGetters(vparams :: vparamss1, n + 1)
         if (vparam.rhs.isEmpty) rest else defaultGetter :: rest
       case Nil :: vparamss1 =>
--- a/src/dotty/tools/dotc/core/tasty/TreePickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/TreePickler.scala
@@ -331,7 +331,7 @@ class TreePickler(pickler: TastyPickler) {
     tree match {
       case Ident(name) =>
         tree.tpe match {
-          case tp: TermRef => pickleType(tp)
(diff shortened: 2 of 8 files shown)


DarkDimius:
Hooraaay for @liufengyun for fixing the bug and to Ycheck for finding it.

liufengyun:
Oops, the fix introduces some pickling errors and TreeChecker assertion exceptions.

liufengyun:
As discussed in dotty meeting,  disable widening breaks assumptions in type inference and elsewhere. And type checking related to singleton types is unspecified.

As there's no simple fix for now, I close this PR and keep the issue open.

PR #1344, opened 2016-06-30
partest: correct line numbers in run stack traces

smarter:
partest adds a warning in a comment at the beginning of source files
that it copies, but this means that every line number displayed in a
stack trace is offset by 6. We can workaround this by making the warning
a single line with no newline at the end.


DarkDimius:
lgtm

PR #1345, opened 2016-06-30
Improve Xprint

nicolasstucki:
* Dont reprint a tree that hasn't changed.
* Highlight changes in yellow.

### DIFF
--- a/src/dotty/tools/dotc/Run.scala
+++ b/src/dotty/tools/dotc/Run.scala
@@ -2,14 +2,20 @@ package dotty.tools
 package dotc
 
 import core._
-import Contexts._, Periods._, Symbols._, Phases._, Decorators._
+import Contexts._
+import Periods._
+import Symbols._
+import Phases._
+import Decorators._
 import dotty.tools.dotc.transform.TreeTransforms.TreeTransformer
 import io.PlainFile
-import util.{SourceFile, NoSource, Stats, SimpleMap}
+import util._
 import reporting.Reporter
 import transform.TreeChecker
 import rewrite.Rewrites
 import java.io.{BufferedWriter, OutputStreamWriter}
+
+import scala.annotation.tailrec
 import scala.reflect.io.VirtualFile
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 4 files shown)


nicolasstucki:
This PR will add `Unchanged since ...` rather than reprinting the tree.
screen shot 2016-06-30 at 14 56 14

It will also show the parts of the tree that have changed in yellow.
screen shot 2016-06-30 at 14 55 57

DarkDimius:
Awesome job! 🌟

DarkDimius:
ping @adriaanm @lrytz, you may be interested port this.

xeno-by:
Wow not bad.

nicolasstucki:
New version will mark additions in green and modifications in yellow.
screen shot 2016-06-30 at 17 07 34

With `-Xprint-diff-del` deleted parts are inserted in red and modified parts are inserted in magenta (just before a yellow modification).
screen shot 2016-06-30 at 17 06 49

adriaanm:
Cool! Yeah we should port this over once 2.12.0 is out

src/dotty/tools/dotc/Run.scala
```diff
@@ -56,26 +64,103 @@ class Run(comp: Compiler)(implicit ctx: Context) {
...
+        ctx.echo("  Unchanged since " + phase)
+        last
+
+      case NoPrintedTree =>
+        ctx.echo(treeString)
+        SomePrintedTree(squashedPhase.toString, treeString)
+    }
+  }
+
+  private def mkColoredDiffTree(treeString: String, lastTreeSting: String): String = {
```
DarkDimius: it would be nice to factor this out to a module reusable from other parts of compiler. 
This logic doesn't seem specific to `Run`.

DarkDimius:
otherwise LGTM

src/dotty/tools/dotc/Run.scala
nicolasstucki: Extracted it to `dotty/tools/dotc/util/DiffUtil.scala`

DarkDimius:
LGTM. Glad to have this merged.

Issue #1341, opened 2016-06-29
A crazy idea: provide a reasonable toString for case classes


DarkDimius:
When having discussions with people in the community, I saw a misconception that as soon as compiler developers have some idea, it quickly gets implemented and shipped in the next release. At the same time, the misconception suggest that ideas from outsiders never get support and are never implemented.

The source of this misconception is that decisions that originate from compiler developers that do are not supported by majority in compiler team(s) are never publicly presented. 
This creates a biased distribution for ideas that are presented to public: most of them are already fleshed out, discussed in private and preliminary decision is frequently made. They have been privately discussed for month if not years in order to cover all the corner-cases of both abstract idea and practical implementation.

This issue exists to show discussion that normally happens in private, to demonstrate that quite a few of ideas that we have are not implemented, as `not good enough`. This is an example of such a idea, that may simplify everyday life of a particular developer (me) but I believe it's not good enough (yet?) to be implemented.

# The proposal

This proposal proposes a best-effort to make toString on case classes be JSON. 
Another good option is proposed by @densh in [comment](https://github.com/lampepfl/dotty/issues/1341#issuecomment-229387206) below.
I don't have preference over either of two currently, as his version has also some convincing advantages, such as repl compatibility. Additionally, AFAIK this is what Python does. 

## Introduction: 
I've been thinking about it for a year or so already, sometimes leaning towards "it's very easy to implement, and it makes common usage easier, so why not do it", while sometimes leaning towards "I like JSON as much as I like XML". 

Recently, we've had a lot of discussions with @felixmulder on how should dottydoc compiler part and web-part communicate. And it seemed that what he intended to implement is comparable in amount of code to implementing inside the compiler.

I'd like to finally have this discussion, mostly to gather opinions of people and see if we should do it.
If we don't do it in Dotty, I guess it's better to never do it. So why not have discussion now?

## Motivation
When discussing including serializer in standard library, in most cases this is seen as a bad idea, as there are multiple formats available and we do not want to bear the burden of supporting all of them. In particular there have always been an argument not to include JSON\XML\Protobuf serializer in standard library\compiler as we'll need to maintain it.

## Current status
Compiler generates a toString on case classes that calls into 
```scala
    ScalaRunTime..MODULE$._toString((Product)this);
```
That is implemented as
```scala
    x.productIterator.mkString(x.productPrefix + "(", ",", ")")
```

## Idea
[...]

PR #1346, opened 2016-06-30
[WIP] Add methods to abstract over structure of case classes

felixmulder:
Inspired by #1341, I decided to take a crack at improving case classes. These commits add two members to case classes:

* `def fields: List[String]`
* `def toMap: Map[String, Any]`

I think it is important to have a good way of abstracting over the members of case classes, this provides just that. One drawback to having this signature in the `toMap` method is that we loose type information. Ideally something like an `HMap` could be used here to retain type information.

There's room for improvement in this implementation as well - for instance `def fields` could of course be a `@static val fields` in the companion instead.

These two functions could in turn be used to improve the `toString` capabilities of case classes as well as various forms of pickling. Let me know your thoughts on this - if it feels like a step in the right direction, I wouldn't mind spending some time on this.


odersky:
I see lots of ideas for improvements to case classes. Instead of trying to do this piece-wise I would prefer to have a very comprehensive SIP detailing how case classes should in the future support generic programming. I'll open an issue for this to serve as a place where discussions can take place.

felixmulder:
Sounds good :)

sourcekick:
@odersky or someone else: Please provide the link to the new issue where this is/was being discussed further.

Blaisorblade:
@sourcekick I can't read Martin's mind, but the next issue, #1347, appears highly relevant.

Issue #1285, opened 2016-05-29
Singleton types should not be inferred for final vars


smarter:
```scala
class Test {
  final var x = false
  x = true
}
```
```scala
error: type mismatch:
 found   : Boolean(true)
 required: Boolean(false)
  x = true
      ^
```

sjrd:
So there is such a thing as a `final var` ... :anguished:

DarkDimius:
what does `final var` mean in the first place?

smarter:
Good question, I just found it in the wild in https://github.com/molecule-labs/molecule/search?utf8=%E2%9C%93&q=%22final+var%22 but didn't stop to think if it made any sense :). http://stackoverflow.com/questions/18222233/what-is-the-usage-of-a-final-var-in-scala says it's to prevent overriding.

DarkDimius:
> says it's to prevent overriding.

Mutable variables can't be overridden even if they are not final.

```scala
scala> class A{var s = 3}; class B extends A{override var s = 4}
<console>:10: error: overriding variable s in class A of type Int;
 variable s cannot override a mutable variable
       class A{var s = 3}; class B extends A{ override var s = 4}
```

smarter:
@adriaanm: any idea if the semantics of `final var` are different from the semantics of `var` ?

adriaanm:
As @DarkDimius said, currently, final is implied for a var currently. I can't think of any difference implied by the presence of the modifier. I guess you could defend a design where you can override a var so you can introduce validation logic in the setter, but you couldn't change the type and the storage would still have to be in the super class, so, not sure...

adriaanm:
The logic around widening inferred types is here: https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L839

PR #1348, opened 2016-07-01
Widen type of final vars #1285

jvican:
Addresses #1285.

When defining a final var, the type is not widened and, thus, it's as
specific as possible. However, when changing the value of the var, it
will give a compilation error because the constant types don't match.
This fix widens the type whenever it finds a final var.

### DIFF
--- a/src/dotty/tools/dotc/typer/Typer.scala
+++ b/src/dotty/tools/dotc/typer/Typer.scala
@@ -1008,7 +1008,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   def typedValDef(vdef: untpd.ValDef, sym: Symbol)(implicit ctx: Context) = track("typedValDef") {
     val ValDef(name, tpt, _) = vdef
     completeAnnotations(vdef, sym)
-    val tpt1 = checkSimpleKinded(typedType(tpt))
+    val tpt1 = widenIfFinalMutable(checkSimpleKinded(typedType(tpt)), sym)
     val rhs1 = vdef.rhs match {
       case rhs @ Ident(nme.WILDCARD) => rhs withType tpt1.tpe
       case rhs => typedExpr(rhs, tpt1.tpe)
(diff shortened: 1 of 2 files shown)


smarter:
I don't think this is the only thing that needs to be done to fix #1285 we determined that `var` and `final var` have (or at least should have) the exact same semantics, so we should either always set the `Final` flag on `var` or never set it, but setting it some of the time and then ignoring it doesn't make sense. In either case we should also warn the user when they write `final var` that they should just write `var`

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1018,6 +1018,16 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+    if (sym.is(allOf(Mutable, Final))) {
+      val widened = vdef.tpe.widen
+      vdef.withType(widened)
+      sym.info_=(widened)
+    }
+    vdef
+  }
+
   /** Add a @volitile to lazy vals when rewriting from Scala2 */
   private def patchIfLazy(vdef: ValDef)(implicit ctx: Context): Unit = {
```
odersky: There's a simpler way to do it. In Namer.scala

      def isInline = sym.is(Final, butNot = Method)

We just need to exclude Mutable here. E.g.

      def isInline = sym.is(Final, butNot = Method | Mutable)

tests/pos/WidenFinalVars.scala
```diff
@@ -0,0 +1,5 @@
+/* For more information, see #1285 */
+class Test {
+  final var x = false
+  x = true
+}
```
nicolasstucki: Missing a new line at the end of the file.

nicolasstucki:
You could change the commit message to `Fix #1285: Widen type of final vars.` to follow the GitHub conventions.

odersky:
Superseded by #1428

PR #1358, opened 2016-07-02
Implement -print

edmundnoble:


DarkDimius:
CLA [signed](http://www.lightbend.com/contribute/cla/scala/check/edmundnoble).
LGTM after tests pass.
Conflicts with #1345.

odersky:
The options in ScalaSettings have been imported wholesale from scalac. We still need to do the work and remove those options that are not honored by dotty. In the case of -print, I believe we should not 
implement it, as -Xprint is already available. So I would rather change the PR to remove -print from ScalaSettings.

odersky:
LHTM, thanks!

nicolasstucki:
@edmundnoble you should squash those 3 commits (or remove the ones that cancel each other).

PR #1377, opened 2016-07-09
Fix 1365: Fix bindings in patterns

odersky:
We need to compare pattern types with expected types in order
to derive knowledge about pattern-bound variables. This is done
using the mechanism of gadt bounds.

This is a subtle aspect of type inference that was so far missing from dotty. scalac does it 
differently, I would argue in a more ad-hoc way.

Review by @DarkDimius.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -813,14 +813,24 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
   private def narrowGADTBounds(tr: NamedType, bound: Type, isUpper: Boolean): Boolean =
     ctx.mode.is(Mode.GADTflexible) && {
     val tparam = tr.symbol
-    typr.println(s"narrow gadt bound of $tparam: ${tparam.info} from ${if (isUpper) "above" else "below"} to $bound ${bound.isRef(tparam)}")
-    !bound.isRef(tparam) && {
-      val oldBounds = ctx.gadt.bounds(tparam)
-      val newBounds =
-        if (isUpper) TypeBounds(oldBounds.lo, oldBounds.hi & bound)
-        else TypeBounds(oldBounds.lo | bound, oldBounds.hi)
-      isSubType(newBounds.lo, newBounds.hi) &&
-      { ctx.gadt.setBounds(tparam, newBounds); true }
+    typr.println(i"narrow gadt bound of $tparam: ${tparam.info} from ${if (isUpper) "above" else "below"} to $bound ${bound.isRef(tparam)}")
+    if (bound.isRef(tparam)) false
+    else bound match {
+      case bound: TypeRef
+      if bound.symbol.is(BindDefinedType) && ctx.gadt.bounds.contains(bound.symbol) &&
+         !tr.symbol.is(BindDefinedType) =>
+        // Avoid having pattern-bound types in gadt bounds,
+        // as these might be eliminated once the pattern is typechecked.
+        // Pattern-bound type symbols should be narrowed first, only if that fails
+        // should symbols in the environment be constrained.
+        narrowGADTBounds(bound, tr, !isUpper)
+      case _ =>
+        val oldBounds = ctx.gadt.bounds(tparam)
+        val newBounds =
+          if (isUpper) TypeBounds(oldBounds.lo, oldBounds.hi & bound)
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -448,11 +448,12 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
             }
-          case tref: TypeRef if tref.symbol.isClass && !ctx.isAfterTyper =>
-            val setBefore = ctx.mode is Mode.GADTflexible
-            tpt1.tpe.<:<(pt)(ctx.addMode(Mode.GADTflexible))
-            if (!setBefore) ctx.retractMode(Mode.GADTflexible)
           case _ =>
+            if (!ctx.isAfterTyper) {
+              val setBefore = ctx.mode is Mode.GADTflexible
+              tpt1.tpe.<:<(pt)(ctx.addMode(Mode.GADTflexible))
+              if (!setBefore) ctx.retractMode(Mode.GADTflexible)
```
DarkDimius: Why do you need to rectractMode? `ctx` here is not fresh and `addMode` should create a new one instead of mutating existing one.

src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -813,14 +813,24 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
+        // as these might be eliminated once the pattern is typechecked.
+        // Pattern-bound type symbols should be narrowed first, only if that fails
+        // should symbols in the environment be constrained.
+        narrowGADTBounds(bound, tr, !isUpper)
+      case _ =>
+        val oldBounds = ctx.gadt.bounds(tparam)
+        val newBounds =
+          if (isUpper) TypeBounds(oldBounds.lo, oldBounds.hi & bound)
+          else TypeBounds(oldBounds.lo | bound, oldBounds.hi)
+        isSubType(newBounds.lo, newBounds.hi) &&
```
DarkDimius:
```scala
if (!isSubType(newBounds.lo, newBounds.hi))
 ctx.gadt.setBounds(tparam, newBounds)

true
```

DarkDimius:
Otherwise LGTM

src/dotty/tools/dotc/typer/Typer.scala
felixmulder: This is from the GADT fix, where I didn't want to change the state of the context apart from relaxing the  solver for GADTs

src/dotty/tools/dotc/core/TypeComparer.scala
odersky: I would not describe it as hacky. The full code you propose would be:

     if (isSubType(newBounds.lo, newBounds.hi)) {
       ctx.gadt.setBounds(tparam, newBounds)
       true
     }
     else false

It's certainly longer. There are quite a few other places in the code base by now that use the same idiom.

src/dotty/tools/dotc/typer/Typer.scala
odersky: But as @DarkDimius notes it's a new context that gets created, so this should not be a problem.

DarkDimius:
It would be also nice to add a test.

odersky:
@DarkDimius There is a test: pos/1365.scala. Should something else be added to it?

DarkDimius:
@odersky, Sorry for confusion, I've left a comment on a wrong PR. It should have been in #1393.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -963,11 +980,30 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     assignType(cpy.ByNameTypeTree(tree)(result1), result1)
   }
 
+  /** Define a new symbol associated with a Bind or pattern wildcard and
+   *  make it gadt narrowable.
+   */
+  private def newPatternBoundSym(name: Name, info: Type, pos: Position)(implicit ctx: Context) = {
+    val flags = if (name.isTypeName) BindDefinedType else EmptyFlags
+    val sym = ctx.newSymbol(ctx.owner, name, flags | Case, info, coord = pos)
```
smarter: `Case` is described as:
```scala
 /** A case class or its companion object */
final val Case = commonFlag(17, "case")
final val CaseClass = Case.toTypeFlags
final val CaseVal = Case.toTermFlags
```
so I don't understand what it's supposed to do here

Issue #1375, opened 2016-07-08
Infinite loop in interpreter


DarkDimius:
pasting  http://scalapuzzlers.com/#pzzlr-034
```scala
val ints = List(1, 2, 3)
val accumulator: List[Int] = Nil

val leftReversed = (ints foldLeft accumulator) { (acc, elem) => elem :: acc }
val rightReversed = (ints foldRight accumulator) { (elem, acc) => acc :+ elem }
println(leftReversed)
println(rightReversed)

// ...and back
println((leftReversed /: accumulator) { (acc, elem) => elem :: acc })
println((rightReversed :\ accumulator) { (elem, acc) => acc :+ elem })
``` 
into interpreter sends it into infinite look after the `// ...and back` comment.
Additionally, we disable control-c, there's no easy way to kill it.

PR #1383, opened 2016-07-11
Fix 1375: Infinite loop in interpreter

kcsongor:
The code example in #1375 caused the REPL to get in an infinite loop, and eventually terminate in an out of memory error.

The cause of this was the way the `/` character was handled, or more precisely, the character following a `/`. Instead of peeking the next character, it was polled, and in case of a no match, was put back using the lazy stream cons, which doesn't play nicely with expressions of the form `stream = n #:: stream` (or I should say it can be surprising).

### DIFF
--- a/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
+++ b/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
@@ -77,10 +77,15 @@ object SyntaxHighlighting {
         (n: @switch) match {
           case '/'  =>
             if (remaining.nonEmpty) {
-              takeChar() match {
-                case '/' => eolComment()
-                case '*' => blockComment()
-                case x => newBuf += '/'; remaining = x #:: remaining
+              remaining.head match {
+                case '/' =>
+                  takeChar()
+                  eolComment()
+                case '*' =>
+                  takeChar()
+                  blockComment()
+                case x =>
+                  newBuf += '/'
               }
             } else newBuf += '/'
           case '='  =>


smarter:
@kcsongor :Thanks for your contribution! Before we can accept your patch, you'll need to sign the Scala CLA at http://www.lightbend.com/contribute/cla/scala

kcsongor:
@smarter: thanks, I've signed it

src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -38,7 +39,7 @@ object SyntaxHighlighting {
   def apply(chars: Iterable[Char]): Vector[Char] = {
     var prev: Char = 0
     var remaining  = chars.toStream
-    val newBuf     = new StringBuilder
+    val newBuf     = new mutable.StringBuilder
```
felixmulder: Since there is no immutable `StringBuilder`, this distinction feels slightly redundant.

src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -262,4 +268,5 @@ object SyntaxHighlighting {
 
     newBuf.toVector
   }
+
```
felixmulder: Maybe remove this addition for a better blame?

felixmulder:
Otherwise LGTM

kcsongor:
I've made the suggested changes, not sure if/how I can squash the commits in the github web editor

smarter:
You can't do it online, but you can do it locally with git and then push-force: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

kcsongor:
I wasn't near my own computer at the time, and I was hoping that I can squash them quickly, but anyway, it should be good now

felixmulder:
Great, thanks for the contribution :)

PR #1389, opened 2016-07-14
Changes to overloading

### LINKED ISSUE #1381: Ambiguous overloads with type parameter and implicit conversion.
```scala
dotty> object Foo {
         class Bar[T]
         implicit def const[T](x: T): Bar[T] = ???
       
         def bar[T](e: T): Any = ???
         def bar[T](e: Bar[T]): Any = ???
       
         val b: Bar[Int] = ???
         bar(b)
       } 
  ^
  bar(bar)
<console>:13: error: Ambiguous overload. The overloaded alternatives of method bar in object Foo$ with types
 [T](e: Foo.Bar[T])Any
 [T](e: T)Any
both match arguments (Foo.Bar[Int](Foo.bar))
```

```scala
scala> object Foo {
     |   class Bar[T]
     |   implicit def const[T](x: T): Bar[T] = ???
     | 
     |   def bar[T](e: T): Any = ???
     |   def bar[T](e: Bar[T]): Any = ???
     | 
     |   val b: Bar[Int] = ???
     |   bar(b)
     | }
defined object Foo
```

odersky:
Fix #1381: Overloading is now changed so that we first try without implicit searches.
Only if that leaves no applicable alternatives we try again with implicit search turned on.
This also fixes test case t2660, which got moved from neg to pos.

Review by @smarter or @nicolasstucki


smarter:
Is this as specified or should we make a note that we need to update the spec?

odersky:
We should make a note to update the spec. I am sure there are other small details that are different as well; notably in how the expected return type is handled.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1095,23 +1071,63 @@ trait Applications extends Compatibility { self: Typer =>
...
+              resolveOverloaded(alts2, pt) match {
+                case alt2 :: Nil => alt2
+                case _ => chosen
+              }
+          }
+        case _ => chosen
       }
+    }
+
+    var found = resolveOverloaded(alts, pt, Nil)(ctx.retractMode(Mode.ImplicitsEnabled))
```
nicolasstucki: Could avoid the `var` here.

nicolasstucki:
That is all.
Otherwise LGTM.

nicolasstucki:
Actually you should also add a test that involves implicits in one or both functions.

odersky:
rebased to master

src/dotty/tools/dotc/typer/Applications.scala
odersky: How would you propose to avoid it?

tests/pos/overloaded.scala
```diff
@@ -57,3 +57,30 @@ object overloaded {
...
+
+object Test2 {
+    trait A; trait B
+    class C1 {
+       def f(x: A): Unit = println("A")
+    }
+    class C2 extends C1 {
+       def f(x: B): Unit = println("B")
+    }
+    object Test extends C2 with App {
```
nicolasstucki: These tests should probably be in `run` to check if the correct overload of `f` is called.

nicolasstucki: Same for class `Foo` above.

tests/pos/overloaded.scala
```diff
@@ -47,4 +47,13 @@ object overloaded {
...
   val t6: String = r6
+
+  // test result disambiguation
+  trait A
+  trait B
+  class C extends A with B
+  def fr(x: A): A = x
+  def fr(x: B): B = x
+  val a: A = fr(new C)
+  val b: B = fr(new C)
```
nicolasstucki: I would do a `run` test to make sure that the correct overload is called.

odersky:
Another insufficient memory failure

/rebuild

odersky:
We seem to have memory problems again.

odersky:
@nicolasstucki Can you add the necessary tests? I won't have time to do it before leaving on my trip. Thanks!

nicolasstucki:
I'll add the tests on tomorrow.

odersky:
@nicolsstucki Thanks for adding the tests.

PR #1390, opened 2016-07-14
Correct typo in IsInstanceOfEvalautor, fixes #1352

### LINKED ISSUE #1352: Typer: Incorrent assumtions when typing pattern matches
Typer assumes that pattern matching in only doing dowcasts and rejects this valid code. This code is accepted by scalac:

```scala
trait A
trait B

abstract sealed class Parent
class Foo extends Parent with A
class Bar extends Parent with B

(null: Parent) match {
  case (_: A) | (_: B) => // dotty complains here that neither A or B are subtypes of Parent
}
```

felixmulder:
No comment.


DarkDimius:
LGTM

PR #1393, opened 2016-07-15
Fix #1386: Reduce double def errors

### LINKED ISSUE #1386: java.util.HashMap.size has a double signature.
```scala
dotty> new java.util.HashMap[Int, Int]().size 
    ^
new java.util.HashMap[Int, Int]().size
<console>:8: error: cannot merge
  method size: ()Int  and
  variable size: Int;
they are both defined in class HashMap but have matching signatures
  ()Int and
  Int
when seen as members of java.util.HashMap[Int, Int]
```

odersky:
Use additional disambiguation criteria before raising a double def error.
See for context: #1240.

Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/core/Denotations.scala
+++ b/src/dotty/tools/dotc/core/Denotations.scala
@@ -292,24 +292,41 @@ object Denotations {
             val sym1 = denot1.symbol
             val sym2 = denot2.symbol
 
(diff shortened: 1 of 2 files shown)


DarkDimius:
This PR lacks test. Did you forgot to push it?

DarkDimius:
t7775 died with OOM.

PR #1395, opened 2016-07-16
Fix #1378: Propagate more knowledge of result type into applications

### LINKED ISSUE #1378: (1, x => 2): (Int, Int => Int) fails type checking the _2
```scala
dotty> (1, x => 2): (Int, Int => Int) 
    ^
(1, x => 2): (Int, Int => Int)
<console>:5: error: missing parameter type for parameter x, expected = T2
```
```scala
scala> (1, x => 2): (Int, Int => Int)
res0: (Int, Int => Int) = (1,<function1>)
```
This issue is related to #1366 and is probably a further refinement of fix #1371.

This code was executed with #1371.

odersky:
In the presence of implicits, this a delicate dance. On the one hand, we need to propagate
knowledge of result types into applications in order to infer parameters as shown in i1378.scala.
On the other hand, we might go down a wrong path with this, if for instance we really meant to insert an implicit on the result type. 

Review by @nicolasstucki or @smarter.

### DIFF
--- a/src/dotty/tools/dotc/ast/TreeInfo.scala
+++ b/src/dotty/tools/dotc/ast/TreeInfo.scala
@@ -280,7 +280,7 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
 trait UntypedTreeInfo extends TreeInfo[Untyped] { self: Trees.Instance[Untyped] =>
   import TreeInfo._
 
-  def isFunctionWithImplicitParamType(tree: Tree) = tree match {
+  def isFunctionWithUnknownParamType(tree: Tree) = tree match {
     case untpd.Function(args, _) =>
       args.exists {
         case ValDef(_, tpt, _) => tpt.isEmpty
--- a/src/dotty/tools/dotc/ast/TreeInfo.scala
+++ b/src/dotty/tools/dotc/ast/TreeInfo.scala
@@ -278,7 +278,18 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
 }
 
 trait UntypedTreeInfo extends TreeInfo[Untyped] { self: Trees.Instance[Untyped] =>
-  // todo: fill with methods from TreeInfo that only apply to untpd.Tree's
+  import TreeInfo._
+
+  def isFunctionWithImplicitParamType(tree: Tree) = tree match {
+    case untpd.Function(args, _) =>
+      args.exists {
+        case ValDef(_, tpt, _) => tpt.isEmpty
+        case _ => false
+      }
+    case _ => false
+  }
+
+  // todo: fill with other methods from TreeInfo that only apply to untpd.Tree's
 }
 
 trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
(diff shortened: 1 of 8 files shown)


src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -278,7 +278,18 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
 }
 
 trait UntypedTreeInfo extends TreeInfo[Untyped] { self: Trees.Instance[Untyped] =>
-  // todo: fill with methods from TreeInfo that only apply to untpd.Tree's
+  import TreeInfo._
+
+  def isFunctionWithImplicitParamType(tree: Tree) = tree match {
```
smarter: I find the use of `implicit` in this context confusing, how about `unknown` instead?

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -553,6 +553,16 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       // a modified tree but this would be more convoluted and less efficient.
       if (proto.isTupled) proto = proto.tupled
 
+      // If some of the application's arguments are function literals without explicitly declared
+      // parameter types, and the expected type is a value type, relate the
+      // normalized result type of the application with the expected type through `<:<`.
+      // This can add more constraints which help sharpen the inferred parameter
+      // types for the argument function literal(s).
+      // This tweak is needed to make i1348 compile.
```
smarter: Did you mean `i1378` instead of `i1348` ?

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -553,6 +553,16 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       // a modified tree but this would be more convoluted and less efficient.
       if (proto.isTupled) proto = proto.tupled
 
+      // If some of the application's arguments are function literals without explicitly declared
+      // parameter types, and the expected type is a value type, relate the
```
smarter: This says `and the expected type is a value type` but I don't see that constraint checked in the code below

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -597,6 +597,13 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         untpd.TypeTree(defn.FunctionClass(args.length).typeRef), args :+ body), pt)
     else {
       val params = args.asInstanceOf[List[untpd.ValDef]]
+
+      pt match {
+        case pt: TypeVar if untpd.isFunctionWithImplicitParamType(tree) =>
+          isFullyDefined(pt, ForceDegree.noBottom)
```
smarter: A comment should clarify that this method is called for its side-effects, and why we don't need to do anything special when it return `false` (I think this is the only instance of `isFullyDefined` in the compiler where we ignore the return value)

odersky:
More info on failing stdlib test: If I run the test alone in partest it succeeds. Unfortunately, this looks like we might have a race. Difficult to say without more test output. I'll make a separate issue so that we can complete this PR.

DarkDimius:
See https://github.com/lampepfl/dotty/issues/1401#issuecomment-233464727 for the log of failure in CI.

nicolasstucki:
LGTM

PR #1398, opened 2016-07-17
Fix #1269: Typing and pattern matching of nested subclasses

### LINKED ISSUE #1269: Discrepancy in pattern matching type checking between dotty and scalac
Following code compiles in scalac, but in dotty it generates following error message:

```
error: Pattern type Patmat.this.module.O.LetR is neither a subtype nor a supertype of selector type Patmat.this.module.Tree
    case module.O.LetR() =>
```

``` Scala
trait Module {
  sealed abstract class Tree

  case class LetL() extends Tree

  object O {
    case class LetR() extends Tree
  }
}

class Patmat(val module: Module) {
  def patmat(tree: module.Tree) = tree match {
    case module.LetL() =>
    case module.O.LetR() =>
  }
}
```

odersky:
Two problems, one in typing and one in pattern matching were fixed.

Review by @liufengyun or @DarkDimius.

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -2993,10 +2993,6 @@ object Types {
     /** The class type with all type parameters */
     def fullyAppliedRef(implicit ctx: Context): Type = fullyAppliedRef(cls.typeRef, cls.typeParams)
 
-    def rebase(tp: Type)(implicit ctx: Context): Type =
-      if ((prefix eq cls.owner.thisType) || !cls.owner.isClass || ctx.erasedTypes) tp
-      else tp.substThis(cls.owner.asClass, prefix)
-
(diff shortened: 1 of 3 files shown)


DarkDimius:
I'd like to review this PR. I'm currently relying on `rebase` in linker and I need to make sure that new version works for me as intended.

odersky:
Is this ready to go in? It's been 2 weeks without any activity on this.

DarkDimius:
@odersky, after applying this change in linker, the behavior did not change(I'm getting the same call graphs on tests), which is good. 
Unfortunately, the running time increased by 1.4x. 

We need this change in for correctness though. So I propose to get the check in and discuss how we can get the speed back later. LGTM.

odersky:
Interesting that the running time went up so much. There might be some scope for caching asSeenFroms, similar to what we do for caching member denotations.

PR #1399, opened 2016-07-17
Fix typing of match expressions

odersky:
Allow matches between unapply types and selector type where a possible
overlap might exist. Review by @liufengyun.

### DIFF
--- a/src/dotty/tools/dotc/typer/Applications.scala
+++ b/src/dotty/tools/dotc/typer/Applications.scala
@@ -751,7 +751,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       if (subtp <:< tp) true
       else tp match {
         case tp: TypeRef if tp.symbol.isClass =>
-          tp.symbol.is(Trait) && isSubTypeOfParent(subtp, tp.parents.head)
+          tp.symbol.is(Trait) && isSubTypeOfParent(subtp, tp.firstParent)
         case tp: TypeProxy => isSubTypeOfParent(subtp, tp.superType)
         case _ => false
       }
--- a/src/dotty/tools/dotc/typer/Applications.scala
+++ b/src/dotty/tools/dotc/typer/Applications.scala
@@ -740,13 +740,19 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
 
     def fromScala2x = unapplyFn.symbol.exists && (unapplyFn.symbol.owner is Scala2x)
 
-    /** Can `subtp` be made to be a subtype of `tp`, possibly by dropping some
-     *  refinements in `tp`?
+    /** Is `subtp` a subtype of `tp` or of some generalization of `tp`?
+     *  The generalizations of a type T are the smallest set G such that
+     *
+     *   - T is in G
+     *   - If a typeref R in G represents a trait, R's superclass is in G.
+     *   - If a type proxy P is not a reference to a class, P's supertype is in G
      */
     def isSubTypeOfParent(subtp: Type, tp: Type)(implicit ctx: Context): Boolean =
       if (subtp <:< tp) true
       else tp match {
-        case tp: RefinedType => isSubTypeOfParent(subtp, tp.parent)
+        case tp: TypeRef if tp.symbol.isClass =>
+          tp.symbol.is(Trait) && isSubTypeOfParent(subtp, tp.parents.head)
+        case tp: TypeProxy => isSubTypeOfParent(subtp, tp.superType)
         case _ => false
       }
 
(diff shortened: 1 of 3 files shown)


smarter:
Could you add tests for this fix?

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -740,27 +740,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+     *   - If a typeref R in G represents a trait, R's superclass is in G.
+     *   - If a type proxy P is not a reference to a class, P's supertype is in G
      */
     def isSubTypeOfParent(subtp: Type, tp: Type)(implicit ctx: Context): Boolean =
       if (subtp <:< tp) true
       else tp match {
-        case tp: RefinedType => isSubTypeOfParent(subtp, tp.parent)
+        case tp: TypeRef if tp.symbol.isClass =>
+          tp.symbol.is(Trait) && isSubTypeOfParent(subtp, tp.parents.head)
+        case tp: TypeProxy => isSubTypeOfParent(subtp, tp.superType)
```
liufengyun: Why `RefinedType` is no longer relevant here? Is it possible that`tp.superType` is a `RefinedType` referencing a class, thus the recursive call becomes inconsistent with the specification that:

> If a type proxy P is not a reference to a class, P's supertype is in G

tests/neg/i1212.scala
```diff
@@ -0,0 +1 @@
+@ann class ann extends scala.annotation.Annotation // error: cyclic reference
```
liufengyun: This seems to be a slip. A relevant test case is expected here.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -740,27 +740,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+     *   - T is in G
+     *   - If a typeref R in G represents a trait, R's superclass is in G.
+     *   - If a type proxy P is not a reference to a class, P's supertype is in G
      */
     def isSubTypeOfParent(subtp: Type, tp: Type)(implicit ctx: Context): Boolean =
       if (subtp <:< tp) true
       else tp match {
-        case tp: RefinedType => isSubTypeOfParent(subtp, tp.parent)
+        case tp: TypeRef if tp.symbol.isClass =>
+          tp.symbol.is(Trait) && isSubTypeOfParent(subtp, tp.parents.head)
```
liufengyun: I tried to find docs about the first parent of a trait, there's nothing special said about that. Why we just handle the first parent? I guess I miss some tacit knowledge here.

smarter: The first parent is the only one which is a class and not a trait, if you don't specify one explicitly then `AnyRef` is added as a first parent of the trait

liufengyun: I assume `tp.symbol.is(Trait)` implies `tp.symbol.isClass` here. If true, it would read better if the guard is changed to `tp.symbol.is(Trait)` and adapt correspondingly the code.

liufengyun: Thanks  @smarter for confirmation, my intuition is that but not sure. Then maybe `tp.firstParent` is better than `tp.parents.head` here.

odersky: > I assume tp.symbol.is(Trait) implies tp.symbol.isClass here. If true, it would read better if the guard is changed to tp.symbol.is(Trait) and adapt correspondingly the code.

No that changes the logic. What we need is that if a TypeRef points to a non-trait class, we stop there, so we are not allowed to execute the following case about TypeProxys.

src/dotty/tools/dotc/typer/Applications.scala
odersky: `RefinedType` is a `TypeProxy`. Generally, the diagram at the top of file Types.scala is important for full understanding of type testing logic.

tests/neg/i1212.scala
odersky: I found out that the issue was closed before and piggy-backed on this one. Not according to ettiquette, I know. My only excuse is lack of time.

src/dotty/tools/dotc/typer/Applications.scala
liufengyun: Thanks @odersky , I see `RefinedType` is a `TypeProxy`. I mean `RefinedType` referencing a class may be caught by `case tp: TypeProxy`, while the comment says:

> If a type proxy P is **not a reference to a class**, P's supertype is in G

It seems to me the comment needs to be updated to be in accord with the code.

liufengyun:
Otherwise, LGTM

src/dotty/tools/dotc/typer/Applications.scala
liufengyun: I see now, by *reference* you mean only *TypeRef*, I was thinking any `TypeProxy` is referencing.

PR #1400, opened 2016-07-18
 HK reduction: Remove special-case for typerefs 

smarter:
The special case:
```scala
  case stripped: TypeRef =>
    stripped.symbol.is(BaseTypeArg)
```
is wrong because you might still want to reduce applications involving
TypeRefs which are not base class parameters, like in:
```scala
  class Foo[A]
  type Alias[X] = Foo[X]
  val x: Alias[Int] = ???
```
`Alias` is a TypeRef so before this commit `Alias[Int]` was never reduced
to `Foo[Int]`. It should have been:
```scala
  case stripped: TypeRef if stripped.symbol.is(BaseTypeArg) =>
    true
```
But even this is incorrect: it assumes that we can always safely reduce HK
applications involving base class parameters, this is not the case when
the parameter kind is different from the rhs kind as illustrated by
`i1181c.scala`. We fix this by simply dropping the special case.

Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/config/Config.scala
+++ b/src/dotty/tools/dotc/config/Config.scala
@@ -98,9 +98,8 @@ object Config {
   final val splitProjections = false
 
   /** If this flag is on, always rewrite an application `S[Ts]` where `S` is an alias for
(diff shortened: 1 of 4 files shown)


smarter:
Oops, had a type in a test, should be all good now.

odersky:
LGTM

Issue #1396, opened 2016-07-17
`_ => _` should be a valid way to write `Function1[_, _]`


smarter:
This works in scalac but not in dotty:
```scala
val x: _ => _ = (x: Int) => 1
val y: (_ <: Int) => _ = (x: Int) => 1
```
Real-world usage: https://github.com/functional-streams-for-scala/fs2/blob/series/0.9/core/src/main/scala/fs2/Chunk.scala#L321

smarter:
Note: I'm thinking that we could leave this one open for a bit as a "good first PR" in the spirit of https://starters.servo.org/

cswinter:
It looks like you cannot have existentials within infix types. E.g. this won't work either:
```scala
val x: Int Either _ = Left(10)
```

cswinter:
I suppose this is caused by an unrelated issue, but when I run the initial example using `Function1[_, _]` instead of `_ => _` with dotr (or dotc), the console will freeze for tens of seconds with CPU at 100% and then report this error (sometimes it OOMs instead):
```
scala> val x: Function1[_, _] = (s: String) => 1 
                                        ^
val x: Function1[_, _] = (s: String) => 1
<console>:4: error: type mismatch:
 found   : Int(1)
 required: 
```

odersky:
@cswinter Thanks for exploring this. The unrelated issue is fixed in #1402.

cswinter:
I've investigated some more, and I believe that at the core of this problem (and also of #1403) is an incomplete attempt of rejecting unbound wildcard types at the grammar level.  
Basically, wildcards are only allowed in the [`ArgType` production](https://github.com/lampepfl/dotty/blob/master/docs/SyntaxSummary.txt#L114) which is used instead of `Type` inside tuple types and type parameter lists, but which is not available to other types (including infix types). We can easily fix #1396 by e.g. widening the grammar through merging the `ArgType` and `SimpleType` productions. Of course then we also allow things like `val a: _ = ???` (this is currently rejected as unbound wildcard by scalac, and causes dotty to hang and eventually oom). Note that it is already possible to define constructs like this in dotty by working around the grammar restriction using singleton tuples (this is essentially what #1403 does). E.g. all of these are currently accepted by the parser:
```scala
val f: (_) => (_)          // compiles
val a: (_) = 1             // #1403. compiler hangs, then cryptic error message/oom
def wut(x: (_)): Int = 1   // compiles
wut(???)                   // compiler hangs, then oom
```
[...]

PR #1402, opened 2016-07-19
Fix wildcard protos

odersky:
Fixes issue discovered by @cswinter on #1396. Also fixes printing of tuple and function types with wildcards. Review by @smarter.


smarter:
If I write:
```scala
  val foo = (x: (_ <: String)) => (1: (_ >: Nothing))
```
Then I get after frontend:
```
    val foo: _ <: String => Int = {
      def $anonfun(x:  >: Nothing <: String): Int = 1
      closure($anonfun)
    }
```
Which compiles, but the type of the parameter of the `def` shouldn't be a wildcard, right?.

smarter:
I guess my example shouldn't even compile, in scalac we get `error: unbound wildcard type`

smarter:
LGTM, opened https://github.com/lampepfl/dotty/issues/1403 to keep track of the issue I noted above

PR #1406, opened 2016-07-20
Make -Xprint-diff an opt-in option.

nicolasstucki:
* -Xprint:[...] alone prints the trees without diffs.
* -Xprint-diff enables the diff on -Xprint.
* -Xprint-diff-del enables the diff with deletes on -Xprint.


nicolasstucki:
Review by @odersky

odersky:
LGTM, thanks for the quick reaction!

PR #1407, opened 2016-07-21
Fix #1401: Make sure all references are forwarded

### LINKED ISSUE #1401: Adding mutable.BufferLike to whitelist breaks partest
Uncommenting this line

    #./scala-scala/src/library/scala/collection/mutable/BufferLike.scala

in scala-collections-whitelist breaks partest, when run with all the other test but succeeds when `compileStdLib` is run as the only test. Unfortunately, the only debug info I can see is this:

    !! 306 - pos/compileStdLib                         [compilation failed]

    # Failed test paths (this command will update checkfiles)
    test/partest --update-check \
      /Users/odersky/workspace/dotty/tests/partest-generated/pos/compileStdLib
commit 9ccb47ae2e9ae34a0ebd6bafbf6f327b39fc8d4a

Any help in debugging this further would be appreciated.

odersky:
I minimized the problem to i1401.scala. If we swap Buffer and BufferLike, the test already succeeds without this PR. The difference between the two orders is that we are lacking some parameter forwarders in the error scenario. This PR fixes that.

### DIFF
--- a/src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala
+++ b/src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala
@@ -106,7 +106,8 @@ object Scala2Unpickler {
         // `denot.sourceModule.exists` provision i859.scala crashes in the backend.
         denot.owner.thisType select denot.sourceModule
       else selfInfo
-    denot.info = new TempClassInfo(denot.owner.thisType, denot.classSymbol, decls, ost) // first rough info to avoid CyclicReferences
+    val tempInfo = new TempClassInfo(denot.owner.thisType, denot.classSymbol, decls, ost)
+    denot.info = tempInfo // first rough info to avoid CyclicReferences
     var parentRefs = ctx.normalizeToClassRefs(parents, cls, decls)
     if (parentRefs.isEmpty) parentRefs = defn.ObjectType :: Nil
     for (tparam <- tparams) {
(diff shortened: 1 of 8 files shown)


odersky:
Review by @liufengyun ?

liufengyun:
Well done @odersky 👍  I don't feel confident for the type system internals, but I'll try to understand the fix. Maybe @smarter can help review as well?

odersky:
Rebased to master

src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala
```diff
@@ -106,7 +106,7 @@ object Scala2Unpickler {
         // `denot.sourceModule.exists` provision i859.scala crashes in the backend.
         denot.owner.thisType select denot.sourceModule
       else selfInfo
-    denot.info = ClassInfo(denot.owner.thisType, denot.classSymbol, Nil, decls, ost) // first rough info to avoid CyclicReferences
+    denot.info = new TempClassInfo(denot.owner.thisType, denot.classSymbol, decls, ost) // first rough info to avoid CyclicReferences
```
liufengyun: It's not clear to me where the `suspensions` of this TempClassInfo is executed. For the TempClassInfo in Namer, there's an explicit call `tempClassInfo.suspensions.foreach(_())`.

odersky: Good catch (they were never executed). I added a commit that fixes that. I could not come up with a test case specifically for classfile reading, but it seems obvious that the treatment
should be the same as in Namer.

liufengyun:
LGTM

liufengyun:
Thanks @odersky for fixing this tricky bug.

PR #1408, opened 2016-07-21
Add erasable phantom classes (old)

nicolasstucki:
Implementation details: https://gist.github.com/nicolasstucki/572c3ee016e966c26491990c6eac8b80


xeno-by:
Sounds interesting, but lacks any kind of description. What are these phantom types? Why would one want to use them?

nicolasstucki:
For now the PR is just for testing purposes. I will add a full description once it is more mature.

The basic idea is to have parameters/field that will get erased (term wise). This will be needed for efficient implementation of implicit evidences for the capability based effect system.

odersky:
In that case we should mark the title of the PR with [WIP]. I think the "in progress" label is not specific enough.

nicolasstucki:
Todo list:
- [x] Create `PhantomAny` and `PhantomNothing` types.
- [x] Create way to instantiate phantom types (synthetic `def phantasmic[P <: PhantomAny]: P`).
- [x] Infer phantom bounds in type parameters. 
- [x] Remove assumption that all types are subtype of `Any`.
- [x] Add `PhantomNothing` as subtype of any phantom type and make `Nothing` a subtype of a type only if that type is non phantom.
- [ ] Add phantom value erasure phase. 
  - [x] Create new phase before erasure.
  - [x] Erase function parameters. `def boo(i: Int)(p: Phantom)` -> `def boo(i: Int)()`
  - [x] Erase function arguments. `boo(2)(phantom)` -> `boo(2)()` 
  - [ ] Erase phantom statements. `class Boo { def p: MyPhantom = ...; def f() = ??? }` -> `class Boo { def f() = ??? }`
    - [x] With hacky implementation of `info` rewriting.
    - [ ] Replace hacky implementation.
- [ ] Disallow `phantasmic` and `phantasmic[PhantomNothing]`.
- [ ] Investigate if there is a use for the equivalent to `???` in phantom types.
- [x] Disallow type parameters that are bounded by mix of phantom and non phantom types.
- [x] Disallow curried groups of parameters with both phantom and non phantom type parameters.
- [ ] Disallow `|` and `&` types with mixes of phantom and non phantom types.
- [x] Make sure that `asInstanceOf` and `isInstanceOf` are not allowed on phantom types.
- [ ] Find a way to only allow instantiation of phantom types in some context (like private/protected constructor for classes).

japgolly:
This seems really cool, @nicolasstucki. Zero-overhead proofs. What particular scenarios did you have that motivated this feature?

nicolasstucki:
@japgolly the main motivation for me is to use them to implement a capability system. This would then be used to encode effect tracking (but this is still a bit far away). We could also potentially use this to reimplement `=:=` and other such constructs that we already have.

japgolly:
Nice!

But would you really be able to implement `=:=` with this? My understand from the desc was that that entire subclasses would be elided, meaning that it'd be awesome for building up proofs that you don't interact with once satisfied (eg. capability system), but wouldn't be applicable once you need to put methods in it. Would methods that don't reference `this`/ctor-args be kept and forwarded to some static location?

How would you imagine `=:=` working?

nicolasstucki:
Mabe `=:=` was not the best example. The proof of concept for can equal in `tests/pos/phantomEq.scala` is a better example.

Yes, all the phantom classes would be elided.
For now, phantom 'objects' can only have methods that create other static evidences. In general phantom classes disallow any construct that could be affected or can affect the program at runtime. If you which to put method on it, I believe that it should be done with an implicit class conversion. I will need to try this.

nicolasstucki:
@japgolly I added a proof of concept of `=:=` implemented with phantoms. It was quite straightforward but still does need a small change in design.
You can find them in `tests/pos/phantomEvidence.scala` and `tests/neg/phantomEvidence.scala`

nicolasstucki:
Review by @odersky.

Blaisorblade:
Based on the description only: this sounds pretty cool, but I have a few questions (if I might butt in).

> Expressions of phantom types can not be if branching expressions `if`, `match` and `try`.

What's the goal of this restriction? I'd have guessed this tries to ensure that erasure is semantics-preserving, but IIUC this criterion is not made explicit, and restrictions described seem both too conservative and too lax.

- It seems it would be enough to allow branches but force them to only return phantom types?
- Such branches should not perform side effects, but right now it sounds like erasable functions can perform side effects (printing) that are unconditionally erased.

In other systems (like Agda) erasability is IIUC an annotation to the type rather than part of the type itself, something closer to `Int @erasable` rather than `ErasableInt`.* Of course, that would still need to not be <: Any, but it would simplify reusing a type as an erasable one. Otherwise maybe I could use write `class PhantomCell[T](t: T) extends PhantomAny`?
For instance, what happens if I want to take an erasable vector length?

*I confess I'm not extremely familiar with the detail of those features and when they do/do not work, that's IIUC not very well documented.

nicolasstucki:
@Blaisorblade in general I've added restrictions that are not absolutely necessary but simplify the implementation. The current PR aims to give support all the required features to implement stuff with phantom types. The descriptions in the PR are mostly implementation details, at some point I will write a more theoretical description an do the link with the implementation. 

> * It seems it would be enough to allow branches but force them to only return phantom types?
> * Such branches should not perform side effects, but right now it sounds like erasable functions can perform side effects (printing) that are unconditionally erased.

For the `if` restrictions it is true that it could work if both branches of the `if` have phantom expressions. The branches will not have side effects by the restrictions on phantom type expressions (including this one), but then the condition must be checked for side effects. The other aspect of an if is that the if the branches are phantoms then the only it's type is important (as they will never execute anything), this implies that you get the intersection of those types. As there are other more explicit ways to get the same intersection types I just disabled them.

I'm not familiar with Agda, but the example that you give is uses dependent types which we don't have in Scala. 

> For instance, what happens if I want to take an erasable vector length?

It could be possible to create phantom types representing integer numbers and use those as in the vector type.

Blaisorblade:
Sorry for the premature critique then! I've also looked at your negative tests and the errors
"Phantom classes can not have expressions in statement position" and "Methods returning phantom values can not have statements" close the loopholes I had in mind.

> I'm not familiar with Agda, but the example that you give is uses dependent types which we don't have in Scala.

Yes, but the encodings of dependent types used in Haskell would work in Scala IIUC. It's enough to have something to erase.

-----

More importantly, there's one small (separate) problem with `S =::= T`: this design doesn't have a method to convert from `S` to `T` and viceversa, which is sometimes needed to use the evidence. It sounds like I can't do that without a cast, unlike now—so you need to verify the cast by hand (which might be acceptable), and this might produce a `checkcast` opcode with some runtime overhead (not sure). But I'm not sure that the current implementation can avoid a cast altogether anyway.

```
implicit class CastOps[S, T](eq: S =::= T) {
  @inline def cast(s: S): T = s // fails to typecheck, needs a cast
}
implicit class CastOps2[S](eq: S =::= S) { //Shouldn't be applicable in interesting examples, as the typechecker doesn't know the two arguments are the same.
  @inline def cast(s: S): S = s
}
```

xavier-fernandez:
It wouldn't make sense it has it's own reserved keyword?

```scala
phantom trait Foo 
```

nicolasstucki:
It would make sense to use the keyword `phantom`. But it would mean that we would need to reserve the string `phantom` for this feature that can be expressed by means of classes and types in it's natural way. Phantom classes are not a special implementation of classes, they are in fact just a version on which we add the necessary restrictions to be able to erase them without changing the runtime semantics. This semantics are defined in a similar way as `AnyRef` and `AnyVal` (and `js.Any` in Scala.js) define different semantics.

Jasper-M:
@Blaisorblade You would need to write `s.asInstanceOf[T]` in your example, but normally there should be no cast in the bytecode.

Blaisorblade:
> You would need to write s.asInstanceOf[T] in your example, but normally there should be no cast in the bytecode.

Oh right, casting to a type variable is erased (unless the variable has upper bounds?), and erasure adds casts elsewhere. You still need to verify the cast by hand, which is probably acceptable.

sirinath:
BTW, the following post on phantom types which was published in "This Month in Dotty" has disappeared from the website and docs: https://raw.githubusercontent.com/wiki/lampepfl/dotty/This-Month-in-Dotty.md

adriaanm:
Interesting! I would suggest writing a more detailed description up front is never a waste of time. It'll both help you refine the design, provide material for reuse in papers/thesis and create more opportunities for collaboration.

nicolasstucki:
Thanks for the suggestion. I will start putting together my technical notes in some document. I'll share it as soon as it has some well structured description.

sirinath:
Looking forward to the post.

nicolasstucki:
Added some implementation details in the [link](https://gist.github.com/nicolasstucki/572c3ee016e966c26491990c6eac8b80) in the first post.

compiler/src/dotty/tools/dotc/core/ScalaFunction.scala
```diff
@@ -0,0 +1,122 @@
...
+      new SimpleFunctionName(functionArity, isImplicit = false)
+    } else if (implicitFunctionArity != -1) {
+      new SimpleFunctionName(implicitFunctionArity, isImplicit = true)
+    } else if (name.phantomFunctionArity != -1) {
+      val phantomicity = new Phantomicity(name.toString.substring(tpnme.PhantomFunction.length))
+      new PhantomFunction(phantomicity, isImplicit = false)
+    } else if (name.implicitPhantomFunctionArity != -1) {
+      val phantomicity = new Phantomicity(name.toString.substring(tpnme.ImplicitPhantomFunction.length))
+      new PhantomFunction(phantomicity, isImplicit = true)
+    } else new NotAFunctionName(name)
```
OlivierBlanvillain: Is this case reachable in practice? (or should this throw an exception instead?)

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -107,33 +108,54 @@ class Definitions {
...
+   *      and Mi == '1' if the the parameter Ti is a phantom (Mi == '0' otherwise).
+   *
+   *  ImplicitPhantomFunctionM traits follow this template:
+   *
+   *      trait ImplicitPhantomFunctionM[T0,...,T{N-1}, R] extends Object with PhantomFunctionM[T0,...,T{N-1}, R] {
+   *        def apply(implicit $x0: T0, ..., $x{N_1}: T{N-1}): R
+   *      }
+   *
+   *      where M is a string of length N where each Char corresponds to a Ti
+   *      and Mi == '1' if the the parameter Ti is a phantom (Mi == '0' otherwise).
```
OlivierBlanvillain: the the

Do you think you could rephrase this not to leak the naming scheme? I also think `phantomicityOnChar` & `phantomicityOffChar` could be made private.

compiler/src/dotty/tools/dotc/core/NameOps.scala
```diff
@@ -231,13 +233,27 @@ object NameOps {
...
+      if (name.startsWith(prefix))
+        try name.drop(prefix.length).toString.toInt
+        catch { case ex: NumberFormatException => -1 }
+      else -1
+    }
+
+    private def testArityWithPhantoms(prefix: Name): Int = {
+      if (name.startsWith(prefix)) {
+        val phantomicity = name.drop(prefix.length)
+        if (phantomicity.exists(c => c != phantomicityOnChar && c != phantomicityOffChar)) -1
```
OlivierBlanvillain: I don't understand this condition

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -644,6 +644,8 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     val cond1 = typed(tree.cond, defn.BooleanType)
     val thenp1 = typed(tree.thenp, pt.notApplied)
     val elsep1 = typed(tree.elsep orElse (untpd.unitLiteral withPos tree.pos), pt.notApplied)
+    checkNotPhantomExpr("Cannot yield a phantom type in one of the if branches", thenp1)
```
OlivierBlanvillain: Question: could these checks be done in a miniphase after Typer instead of being mixed up?

compiler/src/dotty/tools/dotc/core/ScalaFunction.scala
nicolasstucki: No, it is used to check if the name is a valid function name with`ScalaFunction(name).exists`.

compiler/src/dotty/tools/dotc/core/Definitions.scala
nicolasstucki: Done, thanks

compiler/src/dotty/tools/dotc/core/NameOps.scala
nicolasstucki: It is to check if the suffix is valid. Already reformatted and made more explicit.

compiler/src/dotty/tools/dotc/typer/Typer.scala
nicolasstucki: I pushed all the checks in the `PhantomChecks` miniphase. But this one must be caught in typer as it would generate an expression of type `Any | PhantomAny`, then use it in inference and then all hell beaks loose.

nicolasstucki:
This PR is outdated. It was split into smaller parts:
* #2136
* https://github.com/dotty-staging/dotty/tree/implement-phantom-types-part-2
* https://github.com/dotty-staging/dotty/tree/implement-phantom-types-part-3
* https://github.com/dotty-staging/dotty/tree/implement-phantom-types-part-4

PR #1409, opened 2016-07-21
Fix parameterized typedefs with lambdas as rhs

odersky:
Previously the compiler crashed when faced with a parameterized typedef
that has a lambda as rhs. We fix this by refining the condition when
not to abstract in typeDefsig.

Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -934,15 +934,11 @@ class Namer { typer: Typer =>
   }
 
   def typeDefSig(tdef: TypeDef, sym: Symbol, tparamSyms: List[TypeSymbol])(implicit ctx: Context): Type = {
-    val isDerived = tdef.rhs.isInstanceOf[untpd.DerivedTypeTree]
-    //val toParameterize = tparamSyms.nonEmpty && !isDerived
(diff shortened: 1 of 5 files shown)


src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -954,7 +950,9 @@ class Namer { typer: Typer =>
       //
       // The scheme critically relies on an implementation detail of isRef, which
       // inspects a TypeRef's info, instead of simply dealiasing alias types.
-    val rhsType = abstracted(typedAheadType(tdef.rhs).tpe)
+
+    val isDerived = tdef.rhs.isInstanceOf[untpd.DerivedTypeTree]
+    val rhsType = abstracted(typedAheadType(tdef.rhs).tpe, canAbstract = !isDerived)
```
smarter: Having a boolean parameter that changes the behavior of a method to become the identity seems weird to me, I would just write:
```scala
val rhsType0 = typedAheadType(tdef.rhs).tpe
val rhsType = if (!isDerived) abstracted(rhsType0) else rhsType0
```

smarter:
If we're going to support that, we should also support multiple application, `T[Int][String]` currently fails with `error: Test.this.T[Int] does not take type parameters`

odersky:
@smarter Last two commits should address your comments.

smarter:
LGTM

PR #1410, opened 2016-07-22
Fix #1263: Suppress super initializer call for val parameters of traits.

### LINKED ISSUE #1263: NoSuchMethodError when overriding a trait constructor val parameter
Consider the following code:

~~~ scala
trait Foo(val s: String)

val foo = new Foo { override val s = "bar" }
~~~

It compiles fine. Maybe it should raise a compilation error?

At run-time we get the following exception:

~~~
java.lang.NoSuchMethodError: Initialization$Foo.initial$s()Ljava/lang/String;
        at Initialization$$anon$1.<init>(Initialization.scala:8)
        at Initialization$.<init>(Initialization.scala:8)
        at Initialization$.<clinit>(Initialization.scala)
        at Main$.main(Main.scala:5)
        at Main.main(Main.scala)
~~~

odersky:
Val-parameters of traits don't have an initializer, as other vals do. So
we cannot call the initializer in an initialization sequence of a subclass.

Fixes #1263. Review by @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/transform/Mixin.scala
+++ b/src/dotty/tools/dotc/transform/Mixin.scala
@@ -182,8 +182,8 @@ class Mixin extends MiniPhaseTransform with SymTransformer { thisTransform =>
         }
     }
 
-    def wasDeferred(sym: Symbol) =
-      ctx.atPhase(thisTransform) { implicit ctx => sym is Deferred }
+    def was(sym: Symbol, flags: FlagSet) =
+      ctx.atPhase(thisTransform) { implicit ctx => sym is flags }
 
     def traitInits(mixin: ClassSymbol): List[Tree] = {
       var argNum = 0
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/transform/Mixin.scala
```diff
@@ -220,23 +220,23 @@ class Mixin extends MiniPhaseTransform with SymTransformer { thisTransform =>
...
               else if (getter.is(Module))
                 New(getter.info.resultType, List(This(cls)))
               else Underscore(getter.info.resultType)
-            else transformFollowing(superRef(initializer(getter)).appliedToNone)
+            else initial
           // transformFollowing call is needed to make memoize & lazy vals run
           transformFollowing(DefDef(implementation(getter.asTerm), rhs))
         }
-        else if (isScala2x) EmptyTree
+        else if (isScala2x || was(getter, ParamAccessor)) EmptyTree
```
odersky: That's the only semantic change on this line. The rest are just refactorings.

DarkDimius:
LGTM

PR #1412, opened 2016-07-25
Fix #1284: Make classTag depend directly on erasure

### LINKED ISSUE #1284: Array of case objects does not compile due to missing ClassTag
```scala
case object A
case object B

object Test {
  Array(A, B)
}
```
```
error: No ClassTag available for (Object & Product0)'
  Array(A, B)
             ^
```

odersky:
In the end, a classTag reflects the erased version of a type. The only
condition for its generation should be that the erasure is stable under
possible instantiations.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -277,6 +277,22 @@ object TypeErasure {
           else tp1
       }
   }
+
+  /** Does the (possibly generic) type `tp` have the same erasure in all its
+   *  possible instantiations?
+   */
+  def hasStableErasure(tp: Type)(implicit ctx: Context): Boolean = tp match {
+    case tp: TypeRef =>
+      tp.info match {
+        case TypeAlias(alias) => hasStableErasure(alias)
+        case _: ClassInfo => true
+        case _ => false
+      }
+    case tp: PolyParam => false
+    case tp: TypeProxy => hasStableErasure(tp.superType)
+    case tp: AndOrType => hasStableErasure(tp.tp1) && hasStableErasure(tp.tp2)
+    case _ => false
+  }
 }
 import TypeErasure._
 
@@ -493,4 +509,6 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       println(s"no sig for $tp")
       throw ex
   }
+
+
 }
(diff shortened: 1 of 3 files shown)


odersky:
Review by @xeno-by ?

src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -479,15 +480,12 @@ trait Implicits { self: Typer =>
...
-                .appliedToType(tp)
-                .appliedTo(clsOf(tref))
-                .withPos(pos)
-            case _ =>
-          }
+          if (hasStableErasure(tp))
+            return ref(defn.ClassTagModule)
+              .select(nme.apply)
+              .appliedToType(tp)
+              .appliedTo(clsOf(erasure(tp)))
```
xeno-by: That's an interesting DSL! Does it cover the entire language, or it just contains a few most used utilities?

odersky: It covers about 30 ops, maybe half of them construct syntax trees. See `ast.tpd#TreeOps` for where it  is defined.

src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -493,4 +509,6 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       println(s"no sig for $tp")
       throw ex
   }
+
+
```
xeno-by: Why these newlines?

src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -479,15 +480,12 @@ trait Implicits { self: Typer =>
       formal.argTypes match {
         case arg :: Nil =>
           val tp = fullyDefinedType(arg, "ClassTag argument", pos)
-          tp.underlyingClassRef(refinementOK = false) match {
```
xeno-by: This call was previously returning `NoType` on some refined types. I'm not sure whether the new logic returns false on them.

odersky: It's not directly comparable. My point is the new implementation has some justification (stable erasure) where the old one hadn't.

xeno-by: Makes sense.

xeno-by:
To the best of my knowledge, lgtm.

PR #1413, opened 2016-07-25
Fix #1385: Temporarily lift 22 limit for functions

### LINKED ISSUE #1385: Lambdas with parametrized captured parameters may end with more than 22 parameters.
```scala
 def foo22[T](c1:T, c2:T, c3:T, c4:T, c5:T, c6:T, c7:T, c8:T, c9:T, c10:T, c11:T, c12:T, c13:T, c14:T, c15:T, c16:T, c17:T, c18:T, c19:T, c20:T, c21:T, c22:T): Int => T = {
    (a: Int) => { // This labda ends with 23 parameters (c1 to c22 and a)
      c22; c21; c20; c19; c18; c17; c16; c15; c14; c13; c12; c11; c10; c9; c8; c7; c6; c5; c4; c3; c2; c1
    }
  }
```

This case will fail because the inner lambda will have 23 parameter, which fails due to #1384.

odersky:
Explanatiions in comment in Definitions. Review by @nicolasstucki.

### DIFF
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -12,7 +12,13 @@ import collection.mutable
 import scala.reflect.api.{ Universe => ApiUniverse }
 
 object Definitions {
-  val MaxFunctionArity, MaxTupleArity = 22
+  val MaxTupleArity, MaxAbstractFunctionArity = 22
+  val MaxFunctionArity = 24
+    // Awaiting a definite solution that drops the limit altogether, 44 gives a safety
+    // margin over the previous 22, so that treecopiers in miniphases are allowed to
+    // temporarily create larger closures. This is needed in lambda lift where large closures
+    // are first formed by treecopiers before they are split apart into parameters and
+    // environment in the lambdalift transform itself.
 }
 
 /** A class defining symbols and types of standard definitions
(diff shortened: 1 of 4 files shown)


src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -12,7 +12,13 @@ import collection.mutable
 import scala.reflect.api.{ Universe => ApiUniverse }
 
 object Definitions {
-  val MaxFunctionArity, MaxTupleArity = 22
+  val MaxTupleArity, MaxAbstractFunctionArity = 22
+  val MaxFunctionArity = 24
```
nicolasstucki: `MaxFunctionArity` should be 44 as documented below.

nicolasstucki: Actually it is the doc that is off.

nicolasstucki:
That is all.

tests/pos/i1385.scala
```diff
@@ -0,0 +1,7 @@
+object Test {
+def foo22[T](c1:T, c2:T, c3:T, c4:T, c5:T, c6:T, c7:T, c8:T, c9:T, c10:T, c11:T, c12:T, c13:T, c14:T, c15:T, c16:T, c17:T, c18:T, c19:T, c20:T, c21:T, c22:T): Int => T = {
+    (a: Int) => { // This labda ends with 23 parameters (c1 to c22 and a)
```
xeno-by: Typo

nicolasstucki:
LGTM

PR #1416, opened 2016-07-26
-Xprint-diff-del: more meaningful colors

smarter:
When a line is modified, show the deleted part in red and the added part
in green instead of using magenta and yellow which are confusing.

Review by @nicolasstucki or @odersky


nicolasstucki:
LGTM

PR #1417, opened 2016-07-27
Implement Xprint-diff without external libraries.

nicolasstucki:


odersky:
@nicolasstucki Can you propose a reviewer?

nicolasstucki:
Review by @DarkDimius

nicolasstucki:
BTW, I found some bugs with the current implementation. I did not try to minimize them and add them to the issue tracker because this PR fixes them.

src/dotty/tools/dotc/util/DiffUtil.scala
```diff
@@ -30,38 +30,120 @@ object DiffUtil {
...
+    for (j <- y.indices)
+      F(0)(j) = d * j
+    for (i <- 1 to x.length) {
+      for (j <- 1 to y.length) {
+        val mtch = F(i - 1)(j - 1) + similarity(x(i-1), y(j-1))
+        val delete = F(i - 1)(j) + d
+        val insert = F(i)(j - 1) + d
+        F(i)(j) = mtch max insert max delete
+      }
+    }
```
julienrf: Why not use `Array.tabulate` rather than `Array.fill`?

nicolasstucki: It was to remain as close as possible to the pseudocode of the algorithm. I will change it now to save up two small traversals.

odersky:
Can this go in now?

nicolasstucki:
I should be ready to go in. In the last update I just changed a couple of variable names that Codacy suggested and I completely agreed with it in those cases.

src/dotty/tools/dotc/util/DiffUtil.scala
nicolasstucki: Implemented

DarkDimius:
LGTM

PR #1418, opened 2016-07-27
Fix #1353 : more friendly error message for case trait

### LINKED ISSUE #1353: Parser: report a readable error in case user tries to define a case-trait
SIP 25 disallows defining `case trait`s on syntax level. 
Currently reported error is `illegal start of definition` which isn't very informative.

liufengyun:
Now it says:

```
only `case class` or `case object` allowed
```

Review @DarkDimius .


odersky:
LGTM

PR #1421, opened 2016-07-27
Fix #1411: Give REPL output in correct order

### LINKED ISSUE #1411: REPL doesn't indent caret correctly in error messages
The REPL doesn't indent the caret correctly:
```
scala> val a = List(1)
a: scala.collection.immutable.List[Int] = List(1)
scala> a+1
  ^
a+1
<console>:6: error: type mismatch:
 found   : Int(1)
 required: String
```
It should point to the `1`

cswinter:
CompilingInterpreter was inverting output lines by prepending to a List.


odersky:
LGTM, thanks!

PR #1422, opened 2016-07-28
Add Function25 on top of #1413.

nicolasstucki:
Review by @odersky .

There is one case in scalacheck where there are 3 captured parameters on top of the explicit 22.

### DIFF
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -13,8 +13,8 @@ import scala.reflect.api.{ Universe => ApiUniverse }
 
 object Definitions {
   val MaxTupleArity, MaxAbstractFunctionArity = 22
-  val MaxFunctionArity = 24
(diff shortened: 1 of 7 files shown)


DarkDimius:
Adding functions by one does not scale. Can we make functions over 22 
arguments be created lazily and add a temporary check in refchecks that 
people do not use them with a nice error message?

Addionally, it would be great to add a postcondition to lambdalift that 
those functions do not survive.

odersky:
I agree that adding functions by one does not scale. But it's a temporary hack anyway. So I'd say let's add two more as a quick fix before we do the real one. @nicolasstucki can you do this please?

nicolasstucki:
I added functions up to 30, with this we can capture up to 8 additional parameters.

Has someone started looking into removing the parameter number limitation yet?

odersky:
LGTM, thanks! 

Nobody has started working on the general problem yet. I thought Olivier Blancpain could have a go at it when he starts in September.

PR #1425, opened 2016-07-29
Fix #1423: Fix owners of called methods in CollectSuperCalls.

### LINKED ISSUE #1423: Supertraits need to be filled in
Consider the following test:

```scala
    class B { def m: Int = 0 }
    class C extends B { override def m = 4 }
    trait T1 extends B { override def m = 1 }
    trait T2 extends T1 { override def m = 2 }
    trait T3 extends T1 { override def m = 3 }

    trait T4 extends T1
    class D extends B {
      def f() = println(super[B].m)
    }

    object Test extends C with T2 with T3 with T4 {
        println(m)
        println(super[T2].m)
        println(super[T3].m)
        println(super[T4].m)
        new D().f()
      }
    }
```

Running this gives:

    3
    2
    3
    Exception in thread "main" java.lang.IncompatibleClassChangeError: Interface method reference: T1.m()I, is in an indirect superinterface of Test$
    	at Test$.main(test.scala:17)
    	at Test.main(test.scala)

If I revert 468ff9c the example runs fine. The problem seems to be that the supertrait T1 got dropped even though there is a super reference (via `super[T4].m`) to it.

nicolasstucki:
Review by @odersky


src/dotty/tools/backend/jvm/CollectSuperCalls.scala
```diff
@@ -17,14 +17,14 @@ class CollectSuperCalls extends MiniPhaseTransform {
...
     tree match {
-      case Trees.Super(qual: This, mix) if mix.nonEmpty =>
+      case Select(Super(qual, mix), _) if mix.nonEmpty =>
         val classSymbol = qual.symbol.asClass.classSymbol
-        registerSuperCall(classSymbol, tree.tpe.baseClasses.head)
+        registerSuperCall(classSymbol, tree.symbol.owner.asClass)
       case _ =>
     }
-    super.transformSuper(tree)
+    super.transformSelect(tree)
```
smarter: Just return `tree`, the default implementation of `transformX` is always the identity

src/dotty/tools/backend/jvm/CollectSuperCalls.scala
```diff
@@ -17,14 +17,14 @@ class CollectSuperCalls extends MiniPhaseTransform {
 
   def phaseName: String = "collectSuperCalls"
 
-  override def transformSuper(tree: Super)(implicit ctx: Context, info: TransformerInfo): Tree = {
+  override def transformSelect(tree: tpd.Select)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
```
smarter: Usually we do `import tpd._` at the beginning of a class to avoid adding the prefix everywhere: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/ClassOf.scala#L20

src/dotty/tools/backend/jvm/CollectSuperCalls.scala
```diff
@@ -14,17 +14,18 @@ import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, Transforme
...
  */
 class CollectSuperCalls extends MiniPhaseTransform {
+  import tpd._
 
   def phaseName: String = "collectSuperCalls"
 
-  override def transformSuper(tree: Super)(implicit ctx: Context, info: TransformerInfo): Tree = {
+  override def transformSelect(tree: Select)(implicit ctx: Context, info: TransformerInfo): Tree = {
     tree match {
-      case Trees.Super(qual: This, mix) if mix.nonEmpty =>
```
odersky: Fruitless outer pattern match (and this matters at least a little bit, because Select is a common node). More efficient:

    tree.qualifier match { ... }

odersky:
LGTM

PR #1427, opened 2016-07-29
fix #1274: test for dotty bootstrap based on tasty

### LINKED ISSUE #1274: Include a test for selective dotty bootstrap
I have been running the script below locally to verify that bootstrap using Tasty works. It first compiles everything, then compiles each subdirectory individually, using Tasty to access info from other directories. Getting this to work took a while, so it would be good to have a test that verifies we can still do this. 

Can someone with more test-foo than I give it a stab? Here's the (tcsh) script I was using (run it from `dotty/bin`):

    # To run this script, make sure /classes is on the classpath!
    rm -rf /classes/dotty
    alias dc      'echo \!! ; java -Xms1g -Xmx3g dotty.tools.dotc.Main -d /classes \!*'
    cd ../src/dotty/tools
    pwd
    echo "compile self"
    dc *.scala */*.scala */*/*.scala */*/*/*.scala */*/*/*/*.scala */*/*/*/*/*.scala
    dc *.scala
    dc dotc/*.scala
    dc dotc/ast/*.scala
    dc dotc/config/*.scala 
    dc dotc/parsing/*.scala
    dc dotc/printing/*.scala
    dc dotc/repl/*.scala
    dc dotc/reporting/*.scala
    dc dotc/rewrite/*.scala
    dc dotc/transform/*.scala
    dc dotc/typer/*.scala
    dc dotc/util/*.scala

liufengyun:
Fix #1274.

The real change is to make user-supplied `-classpath` take precedence over libraries from the compiler runtime.

### DIFF
--- a/test/dotc/tests.scala
+++ b/test/dotc/tests.scala
@@ -1,10 +1,9 @@
 package dotc
 
 import test._
-import dotty.partest._
-import org.junit.Test
-import org.junit.experimental.categories._
+import org.junit.{Before, Test}
 
+import scala.reflect.io.Directory
 import scala.io.Source
 
 // tests that match regex '(pos|dotc|run|java|compileStdLib)\.*' would be executed as benchmarks.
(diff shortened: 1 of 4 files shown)


smarter:
I see what you're trying to do but this still seem like a very weird change to me, do we really want the scala bootclasspath to have a lesser priority than the scala user classpath?

liufengyun:
@smarter letting user to override the compiler runtime libs seems reasonable for following cases:

1. User wants to compile for a different JRE from the one that runs the compiler.
2. User wants to compile with a different version of `scala-collection` that the compiler depends on.

WDYT?

BTW, the change passes the test set if run locally. I'm still investigating why it fails on Jenkins.

smarter:
Would be interesting to hear from someone with more experience with classfile resolving like @adriaanm maybe ?

liufengyun:
Finally passed all tests. Review @smarter ?

smarter:
Note that scalac will stop using the bootclasspath with Java 9 where it's not available anymore apparently: https://github.com/scala/scala/pull/5385/commits/ce08ea5995f3d0cb7b35cb4bc84bdaa5d858ff5f

test/dotc/tests.scala
```diff
@@ -1,30 +1,29 @@
...
+  val defaultOutputDir = "./out/"
+
   val noCheckOptions = List(
 //        "-verbose",
 //         "-Ylog:frontend",
 //        "-Xprompt",
 //        "-explaintypes",
 //        "-Yshow-suppressed-errors",
+        "-d", defaultOutputDir,
         "-pagewidth", "160")
```
liufengyun: Note that I moved the `-d` option to `noCheckOptions`, so that the `dotc_core_nocheck` test case also generates class files under `./out/`, instead of the current directory.

liufengyun:
This is now rebased and test pass.

PR #1428, opened 2016-07-31
Fix #1285: Mutable vars are not to be considered constant

### LINKED ISSUE #1285: Singleton types should not be inferred for final vars

odersky:
Fixes #1285, replacing #1348. Review by @jvican.


jvican:
LGTM!

Issue #1403, opened 2016-07-19
Disallow unbound wildcard types like "val x: (_ <: Any) = 1"


smarter:
If you write:
```scala
val x: (_ <: Any) = 1
```
In `scalac` you get:
```scala
error: unbound wildcard type
```
But in dotty we get (after the compiler stops desperately searching for a matching implicit view for several seconds):
```scala
error: type mismatch:
 found   : Int(1)
 required:
```

Labelling this issue as `Parser` because that's where the error is emitted in scalac but I don't know if we wish to reimplement that logic in there or do something later.

PR #1431, opened 2016-08-01
Fix #1396, #1403: Properly handle unbound wildcard types

### LINKED ISSUE #1396: `_ => _` should be a valid way to write `Function1[_, _]`

cswinter:


smarter:
The neg test fails because it compiles, instead of commenting its content, you should put it in `tests/pending/neg`

odersky:
LGTM, thanks!

PR #1436, opened 2016-08-01
Fix #1372: Add handler for `PatDef`s to REPL

### LINKED ISSUE #1372: repl breaks multiple val definiton syntax`
taken from: http://scalapuzzlers.com/#pzzlr-017
```scala
class C
val x1,x2 = new C
val y1@y2 = new C
println(x1 == x2)
println(y1 == y2)
```

This code works in dotc, but does not work in repl: 
```scala
scala> val x1,x2 = new C
scala> val y1@y2 = new C
scala> println(x1 == x2)
        ^
println(x1 == x2)
<console>:8: error: not found: x1
```

cswinter:
Adds a new handler to `CompilingInterpreter` to extract variable definitions from `PatDef` statements, which allows multiple assignment syntax and pattern match assignments to work properly in the REPL.

Review by @odersky?

### DIFF
--- a/src/dotty/tools/dotc/repl/CompilingInterpreter.scala
+++ b/src/dotty/tools/dotc/repl/CompilingInterpreter.scala
@@ -222,7 +222,7 @@ class CompilingInterpreter(
           if (delayOutput)
             previousOutput ++= resultStrings.map(clean)
           else if (printResults || !succeeded)
-            resultStrings.map(x => out.print(clean(x)))
+            resultStrings.foreach(x => out.print(clean(x)))
           if (succeeded) {
             prevRequests += req
             Interpreter.Success
@@ -328,6 +328,7 @@ class CompilingInterpreter(
     private def chooseHandler(stat: Tree): StatementHandler = stat match {
       case stat: DefDef => new DefHandler(stat)
       case stat: ValDef => new ValHandler(stat)
+      case stat: PatDef => new PatHandler(stat)
       case stat @ Assign(Ident(_), _) => new AssignHandler(stat)
       case stat: ModuleDef => new ModuleHandler(stat)
       case stat: TypeDef if stat.isClassDef => new ClassHandler(stat)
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -662,29 +663,71 @@ class CompilingInterpreter(
...
+
+        s""" + "$prettyName: $varType = " + """ +
+          if_(s"$fullPath.asInstanceOf[AnyRef] != null") {
+            if_(s"$fullPath.toString().contains('\\n')") { "\"\\n\"" }
+            /* else */ { "\"\"" } +
+              s""" + $fullPath.toString() + "\\n" """
+          } /* else */ {
+            "\"null\\n\""
+          }
       }
```
odersky: I find there are too many levels of escaping to be able to understand it. Can we try to put everything between
one pair of """s and use only interpolation inside? That should also obviate the need for `if_`.

odersky:
Otherwise LGTM

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -662,29 +663,68 @@ class CompilingInterpreter(
...
-            req.fullPath(vname) +
-            ".toString().contains('\\n')) " +
-            " \"\\n\" else \"\") + " +
-            req.fullPath(vname) + ".toString() + \"\\n\") else \"null\\n\") ")
-        }
+        if (!shouldShowResult(req)) return
+
+        val resultExtractors =
+          for (varName <- boundNames)
+            yield resultExtractionCode(req, varName)
```
smarter: `val resultExtractors = boundNames.map(name => resultExtractionCode(req, name))` is more concise

src/dotty/tools/dotc/repl/CompilingInterpreter.scala
```diff
@@ -662,29 +663,68 @@ class CompilingInterpreter(
...
+        code.print(resultExtractors.mkString(""))
+      }
+
+      private def resultExtractionCode(req: Request, varName: Name): String = {
```
smarter: Avoid overloading when possible, you could name this method `resultExtractor` or something like that

smarter:
Thanks for making these changes, merged!

PR #1438, opened 2016-08-02
Fix #1437: handle build failure in bin/dotc.

### LINKED ISSUE #1437: bin/dotc silently ignores build failure
Following the [steps](https://github.com/lampepfl/dotty/wiki/Getting-Started#compiling-and-running-code) in getting started guide, I get the following error:

```
$ ./bin/dotc tests/pos/HelloWorld.scala
The script is going to build the required jar files
Building dotty-interfaces...
done
Building dotty...
done
Building tests...
done
New files detected in dotty, rebuilding...
done
New files detected in dotty-tests, rebuilding...
done
Error: Could not find or load main class dotty.tools.dotc.Main
```

Running `sbt compile` I get a unresolved dependency on jline. After fixing the dependency issue, I was able to run `bin/dotc` without issues.

I would like `bin/dotc` to give a more useful error message when the build fails.

olafurpg:


olafurpg:
I'm open for suggestion on how to better propagate the error message from the subprocess in the command substitution. I think the current is quite hacky, but I struggled to come up with a better solution.

felixmulder:
LGTM

PR #1450, opened 2016-08-12
Fix #1443: Replace toplevel TypeBounds with Any

### LINKED ISSUE #1443: Unbound wildcards cause assertion error/oom
E.g. `val a: _ = ???` causes compiler hang/OOM and `val a: List[_ <: _] = List(0)` crashes with a failed assertion in typer. Some more details on why this happens can be found in [this comment](https://github.com/lampepfl/dotty/issues/1396#issuecomment-234186015). This can only be replicated with dotc; REPL aborts compilation as soon as the frontend emits an "unbound wildcard type" error.

Further examples can be found in `tests/pending/neg/unboundWildcard.scala`, which should be moved to `tests/neg` once this issues has been resolved.

cswinter:
Is there a way to give a full path for "Any" that cannot be shadowed? I've not actually managed to come up with a situation where this would matter, but my current solution still seems inadequate.
Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -648,12 +648,17 @@ object Parsers {
     }
 
 /* ------------- TYPES ------------------------------------------------------ */
-    /** Same as [[typ]], but emits a syntax error if it returns a wildcard.
+    /** Same as [[typ]], but if this results in a wildcard it emits a syntax error and
+     *  returns a tree for type `Any` instead.
      */
     def toplevelTyp(): Tree = {
       val t = typ()
-      for (wildcardPos <- findWildcardType(t)) syntaxError("unbound wildcard type", wildcardPos)
-      t
+      findWildcardType(t) match {
+        case Some(wildcardPos) =>
+          syntaxError("unbound wildcard type", wildcardPos)
+          scalaAny
+        case None => t
+      }
     }
 
     /** Type        ::=  FunArgTypes `=>' Type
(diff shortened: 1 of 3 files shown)


odersky:
`_root_.scala.Any` should be safe enough.

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -648,12 +648,18 @@ object Parsers {
...
+     *  returns a tree for type `Any` instead.
      */
     def toplevelTyp(): Tree = {
       val t = typ()
-      for (wildcardPos <- findWildcardType(t)) syntaxError("unbound wildcard type", wildcardPos)
-      t
+      findWildcardType(t) match {
+        case Some(wildcardPos) =>
+          syntaxError("unbound wildcard type", wildcardPos)
+          Select(Select(Ident(Names.termName("_root_")), Names.termName("scala")),
```
smarter: instead of `Names.termName` and `Names.typeName` you should use the names already defined in [StdNames](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/core/StdNames.scala): `nme.ROOTPKG`, `nme.scala_` and `tpnme.Any`.

odersky: @cswinter Jave a look at the definition of `scalaUnit` in `untpd`. I propose to model a `scalaAny` like this instead of writing things out in the parser.

smarter:
LGTM!

PR #1452, opened 2016-08-15
Fix desugaring of Bind(WILDCARD, _).

### LINKED ISSUE #1432: `val _@a = 1` causes infinite loop
`val _@a = 1` will send the compiler into an infinite loop (where "infinity" is an amount of time > 30min).

nicolasstucki:


nicolasstucki:
Review by @DarkDimius

smarter:
LGTM

PR #1456, opened 2016-08-17
Fix #1444: Pass implicits to parameterless traits if needed

### LINKED ISSUE #1444: Inconsistency with implicit parameter list for traits
```scala
class Cls(implicit x:X)
class ClsImpl extends Cls //this works

trait Tr1(implicit x:X)
class TrtImpl extends Tr1 //Compiler: Error: parameterized trait Tr1 lacks argument list

trait Tr2()(implicit x:X)
class Tr2Impl extends Tr2() //this works

trait X
implicit object AnX extends X
```

The relevant method is traitInits in Mixin.

So it seems there are multiple different implementations of "argument list".
Is that because Mixin phase has to occur after Erasure?

odersky:
Fixes #1444. Review by @smarter. 

Based on #1434.

### DIFF
--- a/src/dotty/tools/dotc/transform/Mixin.scala
+++ b/src/dotty/tools/dotc/transform/Mixin.scala
@@ -189,16 +189,17 @@ class Mixin extends MiniPhaseTransform with SymTransformer { thisTransform =>
       var argNum = 0
       def nextArgument() = initArgs.get(mixin) match {
         case Some(arguments) =>
-          try arguments(argNum) finally argNum += 1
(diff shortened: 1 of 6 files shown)


odersky:
Only last 2 commits are new.

smarter:
LGTM, but SIP-25 should be updated to explicitly mention the semantics of trait implicit paramter lists since they're not obvious

odersky:
I added something to SIP 25. Still have to figure out where to publish it.

PR #1458, opened 2016-08-18
Fixes for comparing and avoiding singletons

odersky:
Two interesting problems involving singleton types exhibited by one test case. Review by @smarter.


tests/pending/pos/spec-t5545/S_1.scala
```diff
@@ -1,4 +0,0 @@
-trait F[@specialized(Int) T1, R] {
```
smarter: Why were the files in this directory deleted?

odersky: It was an oversight.  I'll add spec-t5545l to tests.

smarter:
LGTM

PR #1459, opened 2016-08-18
Refinements to auto-tupling

odersky:
Fixes a nasty interaction with implicit search ion qualifiers.


src/dotty/language.scala
```diff
@@ -4,13 +4,9 @@ object language {
...
 
-  /** Allow higher-kinded type syntax (not yet checked) */
-  val higherKinds = new Feature
-
   /** Keep union types */
   val keepUnions = new Feature
 
-  /** No auto tupling */
-  val noAutoTupling = new Feature
+  val f = "".contains("", (_: Int))
```
smarter: I don't think that was an intentional change

tests/pos/t2913.scala
```diff
@@ -53,3 +53,26 @@ object Main {
...
+                  // and AToRichA(a).foo("") is found.
+                  //
+                  // My reading of the spec "7.3 Views" is that `a.foo` denotes a member of `a`, so the view should
+                  // not be triggered.
+                  //
+                  // But perhaps the implementation was changed to solve See https://lampsvn.epfl.ch/trac/scala/ticket/1756
+
+  a.foo("a", "b") // Without implicits, a type error regarding invalid arity is generated at `foo(<error>"", "")`.
+                  // Typers#tryTypedApply:3274 only checks if the error is as the same position as `foo`, `"a"`, or `"b"`.
+                  // None of these po
```
smarter: This comment is unfinished

tests/pos/t2913.scala
```diff
@@ -53,3 +53,26 @@ object Main {
     ()
   }
 }
+
+object TestWithAutoTuling {
```
lare96: You forgot a "p" here

tests/pos/t2913.scala
odersky: Yes, I think it came from scalac that way, but I might be wrong.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -580,13 +592,11 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
               convertNewGenericArray(ConstFold(result))
             } { (failedVal, failedState) =>
-              val fun2 = tryInsertImplicitOnQualifier(fun1, proto)
-              if (fun1 eq fun2) {
-                failedState.commit()
-                failedVal
-              } else typedApply(
-                cpy.Apply(tree)(untpd.TypedSplice(fun2), proto.typedArgs map untpd.TypedSplice), pt)
-            }
+              def fail = { failedState.commit(); failedVal }
```
smarter: Should be `def fail()` since it's not pure

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -541,17 +541,29 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+
     def realApply(implicit ctx: Context): Tree = track("realApply") {
-      var proto = new FunProto(tree.args, IgnoredProto(pt), this)(argCtx(tree))
-      val fun1 = typedExpr(tree.fun, proto)
+      val originalProto = new FunProto(tree.args, IgnoredProto(pt), this)(argCtx(tree))
+      val fun1 = typedExpr(tree.fun, originalProto)
 
-      // Warning: The following line is dirty and fragile. We record that auto-tupling was demanded as
+      // Warning: The following lines are dirty and fragile. We record that auto-tupling was demanded as
       // a side effect in adapt. If it was, we assume the tupled proto-type in the rest of the application.
```
smarter: Shouldn't this comment be updated? It says "we assume the tupled proto-type in the rest of the application" but we use both `proto` and `originalProto` below

odersky: OK, I'll take care of this in #1460.

PR #1460, opened 2016-08-21
Make sure arguments are evaluated in the correct typer state.

odersky:
There's a tricky interaction with caching of typed arguments in FunProto types
and backtracking using different typer states. We might end up with a typed
argument that is evaluated in one typer state and that is used in another. The
problem is that the argument typing might have inserted type variables (maybe
by adding polymorphic implicit views) that are not registered in the typer
state in which the application is finally typed. In that case we will see
an "orphan poly parameter" in pickling.

The fix is to discard argument types if their typerstate is not committed
to the one in which the application is finally typed. To apply the fix we
need to track

 - for typer states: whether or not it was committed, and what its parent is.
 - for function prototypes: the typer state in which an argument with cached type
   was evaluated.

Test case is t1756.scala, which produced an "orphan poly parameter CI" before.

Based on #1459. First new commit is "Don't force in isErroneous check".

Review by @smarter.


src/dotty/tools/dotc/core/TyperState.scala
```diff
@@ -59,6 +59,20 @@ class TyperState(r: Reporter) extends DotClass with Showable {
...
   def commit()(implicit ctx: Context): Unit = unsupported("commit")
 
+  /** The typer state has already been committed */
+  def isCommitted: Boolean = false
+
+  /** Optionally, if this is a mutable typerstate, it's creator state */
+  def parent: Option[TyperState] = None
+
+  /** The closest ancestor of this typer state (including possible this typer state itself)
+   *  which is not yet committed.
```
smarter: The documentation should state what is returned when this state and all of its ancestors are committed, or this method should return an `Option`

src/dotty/tools/dotc/core/TyperState.scala
```diff
@@ -59,6 +59,20 @@ class TyperState(r: Reporter) extends DotClass with Showable {
...
+   */
+  def uncommittedAncestor: TyperState = parent match {
+    case Some(p) if p.isCommitted => p.uncommittedAncestor
+    case _ => this
```
smarter: If `this` is commited but its parent is not committed, we end up in this branch and return `this` instead of its parent, shouldn't we do something like this instead:
```scala
def uncommittedAncestor: Option[TyperState] =
  if (!this.isCommitted)
    Some(this)
  else
    parent.flatMap(_.isCommitted)
```

src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -181,27 +184,49 @@ object ProtoTypes {
 
     def argsAreTyped: Boolean = myTypedArgs.size == args.length
 
+    private def typedArg(arg: untpd.Tree, typerFn: untpd.Tree => Tree)(implicit ctx: Context): Tree = {
```
smarter: I would move this method below `allArgTypesAreCurrent`, with the other overloads of `typedArg`

smarter: I would also consider renaming it because having two methods called `typedArg` with the same number of arguments can be confusing

src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -181,27 +184,49 @@ object ProtoTypes {
...
+     *  typer state that is not yet committed into the one of the current context `ctx`.
+     *  This is necessary to avoid "orphan" PolyParams that are referred to from
+     *  type variables in the typed arguments, but that are not registered in the
+     *  current constraint. A test case is pos/t1756.scala.
+     *  @return True if all arguments have types (in particular, no types were forgotten).
+     */
+    def allArgTypesAreCurrent()(implicit ctx: Context): Boolean = {
+      evalState foreachBinding { (arg, tstate) =>
+        if (tstate.uncommittedAncestor.constraint ne ctx.typerState.constraint) {
+          println(i"need to invalidate $arg / ${myTypedArg(arg)}, ${tstate.constraint}, current = ${ctx.typerState.constraint}")
```
smarter: `typr.println` ?

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -546,7 +546,8 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       val fun1 = typedExpr(tree.fun, originalProto)
 
       // Warning: The following lines are dirty and fragile. We record that auto-tupling was demanded as
-      // a side effect in adapt. If it was, we assume the tupled proto-type in the rest of the application.
+      // a side effect in adapt. If it was, we assume the tupled proto-type in the rest of the application,
+      // until, possibly, we have to fall back to insert an implicit on thq qualifier.
```
smarter: So why does the code do `tryWithImplicitOnQualifier(fun1, originalProto)` and then fallbacks to `tryWithImplicitOnQualifier(fun1, proto)`, isn't `originalProto` the non-tupled one?

src/dotty/tools/dotc/core/TyperState.scala
odersky: Yes, you are right.

src/dotty/tools/dotc/typer/ProtoTypes.scala
odersky: Indeed. A left-over from debugging.

src/dotty/tools/dotc/typer/Applications.scala
odersky: We have to try with both tupled and untupled.

odersky:
Rebased to master. The test failures are all OOM, btw.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -574,6 +563,32 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+      /** Type application where arguments come from prototype, and no implicits are inserted */
+      def simpleApply(fun1: Tree, proto: FunProto)(implicit ctx: Context): Tree =
+        methPart(fun1).tpe match {
+          case funRef: TermRef =>
+            val app =
+              if (proto.allArgTypesAreCurrent())
+                new ApplyToTyped(tree, fun1, funRef, proto.typedArgs, pt)
+              else
+                new ApplyToUntyped(tree, fun1, funRef, proto, pt)(argCtx(tree))
+            convertNewGenericArray(ConstFold(app.result))
```
smarter: Some of this code is duplicated from `typedApply` and this line is especially non-obvious, it would be better to have `simpleApply` and `typedApply` call into a common method

odersky: As far as I can see, the code was moved out of typedApply, not duplicated.

src/dotty/tools/dotc/core/TyperState.scala
```diff
@@ -59,6 +59,19 @@ class TyperState(r: Reporter) extends DotClass with Showable {
...
+  /** The typer state has already been committed */
+  def isCommitted: Boolean = false
+
+  /** Optionally, if this is a mutable typerstate, it's creator state */
+  def parent: Option[TyperState] = None
+
+  /** The closest ancestor of this typer state (including possibly this typer state itself)
+   *  which is not yet committed, or which does not have a parent.
+   */
+  def uncommittedAncestor: TyperState =
```
DarkDimius: proposal: instead of storing the direct parent, lazily-cache `uncommittedAncestor` and test-update it every time it's updated:

```
private var lastUncommittedAncestor = parent

def uncommittedAncestor: TyperState = {
  if (lastUncommittedAncestor.isCommited && lastUncommittedAncestor.isDefined) { 
    lastUncommittedAncestor = lastUncommittedAncestor.lastUncommittedAncestor
  }
  lastUncommittedAncestor
}

odersky: Good point. I'll roll that into the next PR.

PR #1461, opened 2016-08-22
Fixes of GADTs and test recategorization.

odersky:
Fixes GADT logic so that GADT using code is now Y-checkable.  Also, I went through tests in pending/pos and reclassified as far as possible.

Based on #1460. 
First new commit is "Fix testing in tree checker".

Review by @liufengyun


tests/pos/gadts2.scala
```diff
@@ -8,9 +8,9 @@ object Test {
   case class Cell[a](var x: a) extends Term[a]
   final case class NumTerm(val n: Number) extends Term[Number]
 
-  def f[a](t: Term[a], c: Cell[a]): Unit = {
+  def f[A](t: Term[A], c: Cell[A]): Unit = {
     t match {
-      case NumTerm(n) => c.x = MyDouble(1.0)
+      case NumTerm(n) => c.x = MyDouble(1.0)  // problem is: this assignment is not type correct, since gadt variable is forgotten
```
liufengyun: From programmer's point of view, I'm wondering if it is possible to figure out in this context, `c` is of the type `Cell[Number]`.

I see that approach breaks the `definiteness` of a symbol's type -- which is a fundamental assumption. But it seems such cases are very common for GADTs, I'm thinking if it's possible to solve such problems in general.

If we could do that, then the compiler will be able to reject following code:

```Scala
sealed trait Expr[+T]
case class IntExpr(x: Int) extends Expr[Int]
case class BooleanExpr(b: Boolean) extends Expr[Boolean]

object Test {
    case (IntExpr(_), IntExpr(_)) =>
    case (BooleanExpr(_), BooleanExpr(_)) =>
    case (BooleanExpr(_), IntExpr(_)) =>                //  should reject, but accept
  }
}
```

A tentative idea is that, when typing pattern matching of GADT, we make a set of type constraints available (generated from `selector.tpe <:< pat.tpe`) for typing the body of a case.

- If the constraint set is unsatisfiable, the compiler generates an error (handles the code example above).
- If it's satisfiable, in typing the case body, we can use the constraint to get more specific types for GADT symbols (`JUST` for the current case).

This requires us to introduce the concept of `incarnation`: the type of a symbol in a specific context. In the typing, we first check if it's OK to type symbols normally. If not, we try type symbols with their `incarnated` type.

odersky: The comment is actually no longer correct. The example compiles.

But your example is something else: a more refined version of exhaustiveness checks, right?

liufengyun: I see the example compiles -- but based on `asInstanceOf` hack.

In thee code example I gave, I think the typer should reject the code -- a typing error, instead of unreachable clause in exhaustivity check.

odersky:
Going to merge now as a log of other fixes are depending on this.

liufengyun:
LGTM, I've no more comments.

odersky:
Thanks @liufengyun!

Issue #1457, opened 2016-08-18
3 incompatibilities with scalac


olafurpg:
See: https://github.com/olafurpg/dotty/pull/2/files

All three examples compile fine with scalac while dotc emits a compiler error. I don't know if these incompatibilities are intentional or known bugs.

odersky:
Hi Olaf, can you make a PR against dotty master and put the tests in pos/pending please?

olafurpg:
Done, see #1464.

PR #1464, opened 2016-08-23
Three incompatibilities with scalac.

olafurpg:
Using Scala 2.11.8 and dotty at bcfa3be8cfe2be.

cc/ #1457


odersky:
L (not so) GTM. Let's see what we can fix here.

PR #1465, opened 2016-08-23
Fix #1457: Three incompatbilities with scalac

### LINKED ISSUE #1457: 3 incompatibilities with scalac

odersky:
Two of these are unavoidable. I moved the tests to diabled/not-representable
and added in each case a comment to the main scala file detailing why there
is a deviation.

The last one (import-rewrite) is fixed.

review by @olafurpg


olafurpg:
LGMT. Would be interesting to know how much code in the wild will be affected by these differences.

odersky:
@olafurpg Yes, let's talk about how to bridge the gap here. I believe the rewriting tool and/or the scala compiler could at least give warnings for cases like these which are not representable. We might also be able to handle the "naming-resolution" problem specially in scala2mode.

odersky:
The naming resolution case is now supported under -language:Scala2, and gives a migration warning. 

There's still nothing we can do about the hkt case.

PR #1466, opened 2016-08-23
More run tests

odersky:
Add some run tests which should be passing


felixmulder:
Rebased this against master since it wasn't mergeable.

PR #1469, opened 2016-08-24
Fixes for scala.Dynamic

### LINKED ISSUE #1468: Missing type apply for scala.Dynamic
```scala
class Foo extends Dynamic {
  def applyDynamic[T](n: String)(): T = ???
}

new Foo().foo[Int]()
// should become
new Foo().applyDynamic[Int]("foo")()
```

The current behaviour is to convert the call to `applyDynamic("foo")[Int]()` which then fails type checking with an error message that is not useful.

nicolasstucki:


nicolasstucki:
Review by @odersky

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -592,11 +592,15 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       fun1.tpe match {
         case ErrorType => tree.withType(ErrorType)
         case TryDynamicCallType =>
-          tree match {
-            case tree @ Apply(Select(qual, name), args) if !isDynamicMethod(name) =>
-              typedDynamicApply(qual, name, args, pt)(tree)
-            case _ =>
-              handleUnexpectedFunType(tree, fun1)
+          tree.fun match {
```
odersky: Move up to line 603 into separate function.

src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -679,7 +683,18 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
           }
       case _ =>
     }
-    assignType(cpy.TypeApply(tree)(typedFn, typedArgs), typedFn, typedArgs)
+    if (typedFn.tpe eq TryDynamicCallType) {
+      (pt, typedFn) match {
```
odersky: Same here: Move the block to a separate function (so that each alternative of
the `if` is a single expression.

src/dotty/tools/dotc/typer/Dynamic.scala
```diff
@@ -30,10 +29,12 @@ object Dynamic {
...
   /** Translate selection that does not typecheck according to the normal rules into a applyDynamic/applyDynamicNamed.
-   *    foo.bar(baz0, baz1, ...)              ~~> foo.applyDynamic(bar)(baz0, baz1, ...)
-   *    foo.bar(x = bazX, y = bazY, baz, ...) ~~> foo.applyDynamicNamed("bar")(("x", bazX), ("y", bazY), ("", baz), ...)
+   *    foo.bar(baz0, baz1, ...)                       ~~> foo.applyDynamic(bar)(baz0, baz1, ...)
+   *    foo.bar[T0, ...](baz0, baz1, ...)              ~~> foo.applyDynamic[T0, ...](bar)(baz0, baz1, ...)
+   *    foo.bar(x = bazX, y = bazY, baz, ...)          ~~> foo.applyDynamicNamed("bar")(("x", bazX), ("y", bazY), ("", baz), ...)
+   *    foo.bar[T0, ...](x = bazX, y = bazY, baz, ...) ~~> foo.applyDynamicNamed[T0, ...]("bar")(("x", bazX), ("y", bazY), ("", baz), ...)
    */
-  def typedDynamicApply(qual: untpd.Tree, name: Name, args: List[untpd.Tree], pt: Type)(original: untpd.Apply)(
+  def typedDynamicApply(qual: untpd.Tree, name: Name, targsOpt: Option[List[untpd.Tree]], args: List[untpd.Tree], pt: Type)(original: untpd.Apply)(
```
odersky: I don't think you need an option here, as there are no empty type argument lists, right? 

Generally, Options are avoided in `dotc` when there is a more streamlined solution.

src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -203,8 +203,9 @@ trait TypeAssigner {
    */
   def selectionType(site: Type, name: Name, pos: Position)(implicit ctx: Context): Type = {
     val mbr = site.member(name)
-    if (reallyExists(mbr)) site.select(name, mbr)
-    else if (site.derivesFrom(defn.DynamicClass) && !Dynamic.isDynamicMethod(name)) {
+    lazy val canBeDynamicMethod = site.derivesFrom(defn.DynamicClass) && !Dynamic.isDynamicMethod(name)
+    if (reallyExists(mbr) && (mbr.accessibleFrom(site).exists || !canBeDynamicMethod)) site.select(name, mbr)
```
odersky: Can you explain the logic here? I am particularly wondering about the extra condition
    
    mbr.accessibleFrom(site).exists

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -318,10 +318,10 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
       pt match {
-        case _: FunProto | AssignProto => select
-        case _ =>
-          if (select.tpe eq TryDynamicCallType) typedDynamicSelect(tree, pt)
-          else select
+        case _ if select.tpe ne TryDynamicCallType => select
+        case _: FunProto | AssignProto             => select
+        case PolyProto(_,_)                        => select
+        case _                                     => typedDynamicSelect(tree.qualifier, tree.name, None, pt)
       }
```
odersky: I think this pattern match needs cleaning up. There must be a nicer way to express this!

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -514,7 +514,9 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
               case TryDynamicCallType =>
                 tree match {
```
odersky: I'd also move this block to a separate function.

src/dotty/tools/dotc/typer/Dynamic.scala
nicolasstucki: As far as I coul tell there are never empty argument lists. I added the option just to be absolutely sure. I will change it to use directly the list and see if something breaks.

src/dotty/tools/dotc/typer/TypeAssigner.scala
nicolasstucki: The issue here is that `site.member(name)` can return a private member that is not accessible from this type (for example a private def in one of the parents). These pass the the checks in `selectionType` and fail later when the accessibility of the member is checked. The condition above is there to check that the selected member is in fact accessible. In case it is not and that this is a Dynamic we need to try one of the dynamics transformation.

See run/t5040.scala

odersky: No, a private def in a parent will not show up as a member of a subclass. But there could be another accessibility problem, that's true. Still, I think it's wrong to put accessibility concerns here. That's what we have function accessibleSelectionType for, after all. It's illogical that accessibility enters into selectionType logic for Dynamic resolution but not in other ways. To give a comparison: Here's what happens with implicit search:

    scala> class C { protected def foo: Int = ??? } 
    defined class C
    scala> val c = new C 
    c: C = C@68f4865
    scala> c.foo 
    <console>:7: error: method foo in class C cannot be accessed as a member of C(c) from module class $iw$. Access to protected method foo not permitted because
     enclosing object $iw is not a subclass of
     class C where target is defined
    c.foo
      ^
    scala> class D { def foo: Int = ??? } 
    defined class D
    scala> implicit def c2d(x: C): D = new D 
    c2d: (x: C)D
    scala> c.foo 
    <console>:9: error: method foo in class C cannot be accessed as a member of C(c) from module class $iw$. Access to protected method foo not permitted because
     enclosing object $iw is not a subclass of
     class C where target is defined
    c.foo
      ^
    scala> c2d(c).foo 
    scala.NotImplementedError: an implementation is missing

As you see, the implicit is _not_ tried if a normal member is found but that member is inaccessible. I believe Dynamic should work the same way.

Keep in mind in all this that the concept of private is not related to the concept of accessibility (in either Scala or Java). It looks like it, but it isn't. A private definition is simply not a member in a subclass. Accessibility does not enter the picture here.

odersky:
Is there a way to see what was committed after my last review?

nicolasstucki:
Commits 34ef580  and 0ee74cc where added after the last review.

odersky:
LGTM now. Thanks!

PR #1472, opened 2016-08-24
[doc] Add support for usecases

felixmulder:
This PR solves the `@usecase` annotation, todo:

* [x] Symbols for `@usecase` code
* [x] Typecheck `@usecase` code
* [x] Make result available in `ctx.docbase`
* [x] Fix test failure when documenting `Iterator`(c8b3504)
* [x] Fix test failure when documenting `IterableLike` (7d5bd7c)
* [x] Move cooking of doc strings into dotty (currently in dottydoc)
* [x] Make sure that strings are properly cooked - i.e. no missing `$Coll` expansions (d561cb9)
* [x] Strip `@usecase` after typing
* [x] Stop double expansion - `isExpanded: Boolean`
* [x] Stop expanding in dottydoc
* [x] Clean up


felixmulder:
This is ready for review - @odersky perhaps?

src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -171,6 +171,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val Ybuildmanagerdebug = BooleanSetting("-Ybuild-manager-debug", "Generate debug information for the Refined Build Manager compiler.")
   val Ycompletion = BooleanSetting("-Ycompletion-debug", "Trace all tab completion activity.")
   val Ydocdebug = BooleanSetting("-Ydoc-debug", "Trace all scaladoc activity.")
+  val Ydocrun = BooleanSetting("-Ydoc-run", "Current run should be treated as a docrun, enables `@usecase` annotations in comments")
```
odersky: Is that option needed anytime one wants to do docs with @useCase? In that case it should not be a -Y option, we need something less obscure then? I am not sure why we need the option at all.

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1515,7 +1515,9 @@ object SymDenotations {
 
     /** Enter a symbol in given `scope` without potentially replacing the old copy. */
     def enterNoReplace(sym: Symbol, scope: MutableScope)(implicit ctx: Context): Unit = {
-      require((sym.denot.flagsUNSAFE is Private) ||  !(this is Frozen) || (scope ne this.unforcedDecls) || sym.hasAnnotation(defn.ScalaStaticAnnot))
+      lazy val isUsecase = ctx.settings.Ydocrun.value && sym.name.show.takeRight(4) == "$doc"
```
odersky: I think useCase can be a def instead of a lazy val, since it will be called at most once.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1456,6 +1462,42 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
   }
 
+  private def typedUsecases(syms: List[Symbol], owner: Symbol)(implicit ctx: Context): Unit = {
+    val relevantSyms = syms.filter(ctx.docbase.docstring(_).isDefined)
+    relevantSyms.foreach { sym =>
+      expandParentDocs(sym)
+      val usecases = ctx.docbase.docstring(sym).map(_.usecases).getOrElse(Nil)
+
+      usecases.foreach { usecase =>
+        enterSymbol(createSymbol(usecase.untpdCode))
```
odersky: Since you are entering symbols I would run typedUseCases in context with some fresh local scope. Would that not work?

felixmulder: I see - disregard my previous comment and I'll remove the flag and try this. :)

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1190,6 +1192,9 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     val self1 = typed(self)(ctx.outer).asInstanceOf[ValDef] // outer context where class members are not visible
     val dummy = localDummy(cls, impl)
     val body1 = typedStats(impl.body, dummy)(inClassContext(self1.symbol))
+
+    typedUsecases(body1.map(_.symbol), self1.symbol)(localContext(cdef, cls).setNewScope)
```
felixmulder: @odersky: using a fresh scope here, also got rid of the flags and changed the `lazy val` to a `def`.  So if this is correct - then this PR should be good to go :)

odersky:
@felixmulder I agree with your changes, but think nevertheless usecases should only be typechecked in a docrun. A usecase might be ill-typed, but a normal compiler run should treat this as just a comment. Let's discuss how to best achieve this.

felixmulder:
Superseded by #1502

PR #1475, opened 2016-08-25
Fix SymDenot.isStatic & isStaticOwner after lambdaLift.

DarkDimius:
The other approach would be to actually check for presence of $outer
field after lambda lift, but I'd prefer not to rely on side-effects here


odersky:
I don't think this is in the right spirit of things. "isStatic" is a property of denotations. When things get moved out to package level by flatten, they are thereby made static. We should not try to dissimulate this. So we need to work at the caller. Either ask whether the denotation was static before flatten and lambda lift, or indeed look at an outer field.

odersky:
Superseded by #1477.

PR #1476, opened 2016-08-25
Fix handling of superCalls.

DarkDimius:
1. There may be calls to super on non-this.
2. there may be calls to in-dirrect super-traits.


odersky:
Superseded by #1477(?)

PR #1479, opened 2016-08-26
Fix #1286: Warn on inexistent imports that are not used.

### LINKED ISSUE #1286: Trying to import a java class that doesn't exist from a package that exists does not produce an error
This shouldn't compile:

```scala
import java.io.Idontexist
object Test
```

nicolasstucki:


smarter:
Shouldn't this be an error instead of a warning?

nicolasstucki:
Given that they can be safely ignored without any semantic issue, it seemed that a warning was the reasonable choice.

src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -160,7 +160,16 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
     ast.Trees.flatten(reorderAndComplete(trees)(ctx.withPhase(thisTransformer.next)))
 
   override def transformOther(tree: Tree)(implicit ctx: Context, info: TransformerInfo) = tree match {
-    case tree: Import => EmptyTree
+    case tree: Import =>
+      val exprTpe = tree.expr.tpe
+      tree.selectors.foreach {
+        case ident @ Ident(name) if name != nme.WILDCARD && !exprTpe.member(name).exists =>
```
odersky: You can factor our the common code in a helper function `checkExists(name)`.

src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -160,7 +160,16 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
     ast.Trees.flatten(reorderAndComplete(trees)(ctx.withPhase(thisTransformer.next)))
 
   override def transformOther(tree: Tree)(implicit ctx: Context, info: TransformerInfo) = tree match {
-    case tree: Import => EmptyTree
+    case tree: Import =>
```
odersky: I think it would be better to move the whole check in Typer or PostTyper. FirstTransform is late for this kind of thing, it will not be called at all in IDEs until you build.

nicolasstucki: Moved to PostTyper.

nicolasstucki:
All comments have been addressed.

stuhood:
It is very handy for this check to differentiate between used and unused, thanks! But I agree with @smarter that it should be an error.

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -274,6 +274,18 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
               case tpe => tpe
             }
           )
+        case Import(expr, selectors) =>
+          val exprTpe = expr.tpe
+          def notExists(name: Name): Boolean =
+            name != nme.WILDCARD && !exprTpe.member(name).exists
```
odersky: I think you need to refine this so that you look for term names and for type names. 

I also agree that it should be an error.

nicolasstucki: Now it emits an error and it tests the name as a type name as well.

nicolasstucki: I also made it encode the name. This was needed for example in `scala.collections.immutable.::`.

src/dotty/language.scala
```diff
@@ -4,6 +4,16 @@ object language {
 
   class Feature
 
+  /** Enable scala.Dynamic */
```
odersky: Can we leave everything in scala.language? My latest PR #1550 removed dotty.language completely. 

Otherwise LGTM

nicolasstucki: The change on #1550 will break this PR. I will have to find another way to handle the members in the language module.

odersky: Yes, I would propose to have them all in scala.language, not dotty.language.

src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -275,6 +275,23 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
               case tpe => tpe
             }
           )
+        case Import(expr, selectors) =>
+          val exprTpe = expr.tpe
+          def checkIdent(ident: Ident): Unit = {
+            val name = ident.name.asTermName.encode
+            def isDottyLanguageFeature = { // TODO Add new language features to scala.language
```
odersky: Why make an exception for language features? I would doubt we need one, but if we do, this needs to be explained in a comment.

odersky:
Resolved that we need to solve the dotty language module issue (i.e. that it differs from scalac's) as part of this pull request.

nicolasstucki:
#1617 fixed the issue with the `language` module. This PR will probably need a rebase on #1617.

nicolasstucki:
Rebased. No longer depends on `dotty.language`.

PR #1481, opened 2016-08-26
Remove warnings from compliation of dotty.

nicolasstucki:
* Suppress warnings from the java Function caused by casts on erased specialized functions.
* Remove 1 dead code warning.
* Remove 1 import warning.

With this get from 372 warning each time dotty is compiled to 7 warnings.


DarkDimius:
LGTM

PR #1482, opened 2016-08-26
More tweaks to type inference

odersky:
It turned out we need to merge constraints when backtracking, after all.

### DIFF
--- a/src/dotty/tools/dotc/core/TyperState.scala
+++ b/src/dotty/tools/dotc/core/TyperState.scala
@@ -59,18 +59,10 @@ class TyperState(r: Reporter) extends DotClass with Showable {
   /** Commit state so that it gets propagated to enclosing context */
   def commit()(implicit ctx: Context): Unit = unsupported("commit")
 
-  /** The typer state has already been committed */
-  def isCommitted: Boolean = false
-
-  /** Optionally, if this is a mutable typerstate, it's creator state */
-  def parent: Option[TyperState] = None
-
   /** The closest ancestor of this typer state (including possibly this typer state itself)
    *  which is not yet committed, or which does not have a parent.
    */
-  def uncommittedAncestor: TyperState =
-    if (!isCommitted || !parent.isDefined) this
-    else parent.get.uncommittedAncestor
+  def uncommittedAncestor: TyperState = this
 
   /** Make type variable instances permanent by assigning to `inst` field if
    *  type variable instantiation cannot be retracted anymore. Then, remove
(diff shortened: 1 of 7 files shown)


odersky:
Rebased to master

odersky:
Review by @smarter?

tests/pos/isApplicableSafe.scala
```diff
@@ -0,0 +1,54 @@
...
+    //
+    // isApplicableSafe.scala:15: error: type mismatch:
+    // found   : A.this.Array[Nothing]
+    // required: A.this.Array[Symbol]
+    // xs = Array(Array())
+    //
+    // Here's the sequence of events that leads to this problem:
+    //
+    // 1. the outer Array.apply is overloaded, so we need to typecheck the inner one
+    //    without an expected prototype
```
smarter: What if instead of giving up when this fails, we tried to re-typecheck the inner call using every possible overload? Of course this would have to be done carefully to avoid ending up like the Swift typechecker which apparently gives up when trying to compile `let a: Double = -(1 + 2) + -(3 + 4) + 5` because of overloading and constraint resolutions: https://www.cocoawithlove.com/blog/2016/07/12/type-checker-issues.html

On a more general note, it seems that there is a tension in Scala caused by overloading: on one hand it significantly reduces the power of type inference, on the other hand it makes some APIs much nicer to use. So either we figure out how to get overloading to play well with type inference, or we come up with a more principled alternative to overloading that helps us keep our APIs nice.

odersky: Yeah, I am very nervous about trying all combinations. That typically leads to blowup in compile times. Then you either become too restrictive and can't even typecheck trivial expressions or your compile times go through the roof. In both cases there's little a programming can do.

smarter:
The commit message for https://github.com/lampepfl/dotty/pull/1482/commits/36648547105bc8ac53dd63a03821a27243af13cd says:
> the merge in TyperState is needed not just for
isApplicableSafe but also for (e.g. erased-lubs.scala)

Looks like a word is mising after `for`

smarter:
Otherwise LGTM

src/dotty/tools/dotc/core/TyperState.scala
```diff
@@ -128,25 +125,20 @@ extends TyperState(r) {
    */
   override def commit()(implicit ctx: Context) = {
     val targetState = ctx.typerState
-    assert(targetState eq previous)
     assert(isCommittable)
-    targetState.constraint = constraint
+    targetState.constraint =
+      if (targetState.constraint eq previousConstraint) constraint
+      else targetState.constraint & constraint
```
smarter: The commit message of https://github.com/lampepfl/dotty/pull/1482/commits/36648547105bc8ac53dd63a03821a27243af13cd explains why this is needed but this is tricky enough that it should also be explained in a comment here.

odersky:
I'll add the note wrt merging in the next PR.

PR #1491, opened 2016-08-31
Fix #1490: type test of union types via type alias

### LINKED ISSUE #1490: Incorrect type test of union types via type alias
``` Scala
class Base {
  type T = Int | Boolean
  def test(x: Object) = x.isInstanceOf[T]
}

object Test {
    val b = new Base
    println(b.test(Int.box(3)))                   
    println(b.test(Double.box(3.4)))    // should be false, but true
  }
}
```

liufengyun:
Make type erasure accept one additional parameter:

- `eraseOrType`: Whether erase OrType to its lowest uppper bound

### DIFF
--- a/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
+++ b/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
@@ -100,7 +100,7 @@ trait TypeTestsCasts {
          *  The transform happens before erasure of `argType`, thus cannot be merged
          *  with `transformIsInstanceOf`, which depends on erased type of `argType`.
          */
-        def transformOrTypeTest(qual: Tree, argType: Type): Tree = argType match {
+        def transformOrTypeTest(qual: Tree, argType: Type): Tree = argType.dealias match {
           case OrType(tp1, tp2) =>
             evalOnce(qual) { fun =>
               transformOrTypeTest(fun, tp1)
--- a/tests/run/i1490.check
+++ b/tests/run/i1490.check
@@ -1,0 +1,3 @@
+true
+true
+false
--- a/tests/run/i1490.scala
+++ b/tests/run/i1490.scala
@@ -1,0 +1,13 @@
+class Base {
+  type T = Int | Boolean
+  def test(x: Object) = x.isInstanceOf[T]
+}
+
+object Test {
+  def main(args: Array[String]) = {
+    val b = new Base
+    println(b.test(Int.box(3)))
+    println(b.test(Boolean.box(false)))
+    println(b.test(Double.box(3.4)))
+  }
+}


src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -112,9 +112,10 @@ object TypeErasure {
   /** The standard erasure of a Scala type. Value classes are erased as normal classes.
    *
    *  @param tp            The type to erase.
-  */
-  def erasure(tp: Type)(implicit ctx: Context): Type =
-    erasureFn(isJava = false, semiEraseVCs = false, isConstructor = false, wildcardOK = false)(tp)(erasureCtx)
+   *  @param eraseOrType   Whether erase OrType to its lowest uppper bound
```
smarter: I would write `Whether OrType should be erased to its lowest upper bound`

src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -112,9 +112,10 @@ object TypeErasure {
...
+   */
+  def erasure(tp: Type, eraseOrType: Boolean = true)(implicit ctx: Context): Type =
+    erasureFn(isJava = false, semiEraseVCs = false, isConstructor = false, wildcardOK = false)(tp, eraseOrType)(erasureCtx)
```
smarter: I think it would be more consistent to add this parameter to the `TypeErasure` class like `semiEraseVCs`, etc.

liufengyun: Indeed it's, I'll update the commit.

src/dotty/tools/dotc/core/TypeErasure.scala
liufengyun: Thanks, will address in the next commit.

odersky:
I am missing the context. Why do we need to make this distinction? The way it is done it adds a significant cost to compiler complexity. Adding a new boolean parameter to a core method should not be taken lightly, because you just have doubled all possible error scenarios!

smarter:
`transformIsInstanceOf` takes erased arguments but sometimes y`ou don't actually want to erase things because you're going to do transformations on the `instanceOf` checks like for union types and singleton types. The alternative would be to rewrite `transformIsInstanceOf` to expect non-erased arguments and only erase them after doing the rewritings it needs to do, that might be better indeed.

smarter:
By the way, the same thing is true in https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala which deals with erased values and so might end up doing an incorrect rewriting.

liufengyun:
> The alternative would be to rewrite transformIsInstanceOf to expect non-erased arguments and only erase them after doing the rewritings it needs to do, that might be better indeed.

@smarter That is how the previous implementation was done, but it doesn't handle the case in #1490 . A proper handling of #1490 would require us to recursively get rid of all type alias, that's part of the logic of type erasure. Either we need to find a way to reuse the type erasure code, or we have to duplicate some of its code.

odersky:
Question: What should the imlementation

    xs.isInstanceOf[List[A | B]] 

be? Erase | or not? I would assume erase. In which case, the whole thing does belong in isInstanceOfEvaluator. Can #1490 not be addressed by dealiasing?

smarter:
I think that inn your exaple, `A | B` is not erased if `eraseOrType` is false, but `List[A | B]` is erased to a raw `List` anyway so it doesn't matter

liufengyun:
I agree it's better to make the core simpler. Will follow the original approach by dealiasing.

smarter:
LGTM

PR #1492, opened 2016-09-01
Implement inline

odersky:
* [x] Basic inliner
 * [x] Keep original call alongside inlined code
 * [x] Print inlining history in error messages
 * [x] Support for separate compilation
 * [x] Support for outer this references in inlined code
 * [x] Create accessors for non-public members accessed from inlined code
 * [x] Inline closures that come from an inline method
 * [x] Use new `inline` keyword instead of annotation
 * [x] Clarify relation to SIP

Review by @xeno-by


src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -0,0 +1,200 @@
...
+  private val InlinedBody = new Attachment.Key[InlinedBody]
+
+  def attachBody(inlineAnnot: Annotation, tree: => Tree)(implicit ctx: Context): Unit =
+    inlineAnnot.tree.putAttachment(InlinedBody, new InlinedBody(tree))
+
+  def inlinedBody(sym: SymDenotation)(implicit ctx: Context): Tree =
+    sym.getAnnotation(defn.InlineAnnot).get.tree
+      .attachment(InlinedBody).body
+
+  private class Typer extends ReTyper {
```
DarkDimius: why is inliner a typer? Cant it be made it into a miniphase\part of post-typer?

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -402,37 +402,28 @@ class Namer { typer: Typer =>
...
-      ctx
+        addEnumConstants(mdef, sym)
+        ctx
+      case stats: Thicket =>
+        stats.toList.foreach(recur)
+        ctx
+      case _ =>
+        ctx
+    }
+    recur(stat)
```
felixmulder: @odersky - changed this to expand inside the recursive function instead. Since we can't get the attached docstrings on some expanded trees otherwise.

src/dotty/tools/dotc/typer/Inliner.scala
odersky: 1. inlining needs to be done during typing, since macro expansion depends on it, and macro expansion needs to e.g. create implicit values.

2. The inliner could possibly be formulated as another TreeMap instead of a full typer. I made it a type primarily for two reasons.

  2.1 That way we get constant folding for free (would have to be implemented explicitly in a treemap).
  2.2 That way we make sure the inlined code is type correct instead of failing during translation.

odersky:
I think we are done for now. Over to you, @xeno-by!

xeno-by:
@odersky Congrats with finishing the initial version of this fun project! I'm on vacation together with Mano and Denys until the end of the week, but will do my best to review after I return.

odersky:
Last two days commits

 - add inline function parameters
 - contain fixes for inline methods in inline arguments, which are needed for compiling nested fors
 - removes InlinInfo in favor of a simpler annotation-based scheme

src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -434,6 +431,36 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
...
+      (meth, targs, Nil)
+    case _ =>
+      (tree, Nil, Nil)
+  }
+
+  /** An extractor for closures, either contained in a block or standalone.
+   */
+  object closure {
+    def unapply(tree: Tree): Option[(List[Tree], Tree, Tree)] = tree match {
+      case Block(_, Closure(env, meth, tpt)) => Some(env, meth, tpt)
```
smarter: This is only correct for closures created by `desugar.makeClosure` and only for certain phases, we have similar pattern matches in other places that makes similar assumption and I don't like it either, I think `ExpandSAMs#transformBlock` is the only one which tries to be careful, maybe we could use that here:
```scala
case Block(stats @ (fn: DefDef) :: Nil, Closure(_, fnRef, tpt)) if fnRef.symbol == fn.symbol =>
```

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2551,6 +2577,24 @@ object Types {
         x => paramBounds mapConserve (_.subst(this, x).bounds),
         x => resType.subst(this, x))
 
+    /** Merge nested polytypes into one polytype. nested polytypes are normally not suported
```
smarter: Typo: suported

src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -0,0 +1,518 @@
...
+          if (refPart.symbol.isStatic) {
+            // Easy case: Reference to a static symbol
+            val accessorType = accessedType.ensureMethodic
+            val accessor = accessorSymbol(tree, accessorType).asTerm
+            val accessorDef = polyDefDef(accessor, tps => argss =>
+              rhs(refPart, tps, argss))
+            val accessorRef = ref(accessor).appliedToTypeTrees(targs).appliedToArgss(argss)
+            (accessorDef, accessorRef)
+          } else {
+            // Hard case: Reference needs to go via a dyanmic prefix
```
smarter: Typo: dyanmic -> dynamic

src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -0,0 +1,518 @@
...
+       *  @param accessedType The type of the accessed method or field, as seen from the access site.
+       *  @param rhs          A function that builds the right-hand side of the accessor,
+       *                      given a reference to the accessed symbol and any type and
+       *                      value arguments the need to be integrated.
+       *  @return The call to the accessor method that replaces the original access.
+       */
+      def addAccessor(tree: Tree, refPart: Tree, targs: List[Tree], argss: List[List[Tree]],
+                      accessedType: Type, rhs: (Tree, List[Type], List[List[Tree]]) => Tree)(implicit ctx: Context): Tree = {
+        val (accessorDef, accessorRef) =
+          if (refPart.symbol.isStatic) {
```
smarter: Couldn't this be extended to:
```scala
if (refPart.symbol.isStatic || refPart.symbol.isContainedIn(inlineMethod.owner) {
  ...
}
```
This way when we have:
```scala
    private val x: Int
    inline def get1 = x
```
we'll get:
```scala
def $inlineAccessor$get1$1: T = C.this.x
```
instead of:
```scala
def $inlineAccessor$get1$1(x$0: Test.C[T]): T = x$0.x
```

src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -0,0 +1,518 @@
...
+            if (tree.isTerm) {
+              val (methPart, targs, argss) = decomposeCall(tree)
+              addAccessor(tree, methPart, targs, argss,
+                  accessedType = methPart.tpe.widen,
+                  rhs = (qual, tps, argss) => qual.appliedToTypes(tps).appliedToArgss(argss))
+           } else {
+              // TODO: Handle references to non-public types.
+              // This is quite tricky, as such types can appear anywhere, including as parts
+              // of types of other things. For the moment we do nothing and complain
+              // at the implicit expansion site if there's a reference to an inaccessible type.
```
smarter: Can't we just disable access checks for type members in inlined blocks? We need accessors for term members because of the JVM but type members get erased anyway.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1735,15 +1780,23 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-        else if (tree.tpe <:< pt)
-          if (ctx.typeComparer.GADTused && pt.isValueType)
+        else if (tree.tpe <:< pt) {
+          if (pt.hasAnnotation(defn.InlineParamAnnot))
+            checkInlineConformant(tree, "argument to inline parameter")
+          if (Inliner.hasBodyToInline(tree.symbol) &&
+              !ctx.owner.ownersIterator.exists(_.isInlineMethod) &&
+              !ctx.settings.YnoInline.value &&
+              !ctx.isAfterTyper)
+            adapt(Inliner.inlineCall(tree, pt), pt)
```
smarter: This is the only call to `inlineCall` and it's in a branch with `tree.tpe <:< pt` this means that when `adaptToSubType` is necessary we never inline, for example `get` isn't inlined in `foo`:
```scala
object Test {
  inline def get: Int = 1

  def foo: Unit = {
    get
  }
}
```

src/dotty/tools/dotc/parsing/Scanners.scala
```diff
@@ -212,8 +213,22 @@ object Scanners {
     /** A buffer for comments */
     val commentBuf = new StringBuilder
 
+    private def handleMigration(keyword: Token): Token =
+      if (!isScala2Mode) keyword
+      else if (keyword == INLINE) treatAsIdent()
```
smarter: This means that the following doesn't work with `-language:Scala2` because `inline` is treated as an identifier:
```scala
inline def x = ...
```
you have to use `@inline` instead. We can fix this by only treating `INLINE` specially if the previous token is `AT`.

src/dotty/tools/dotc/ast/TreeInfo.scala
odersky: Not sure about this one. The question is, how should closures in other blocks be categorized? Is that still a closure or something else? For the moment I'd leave as is.

src/dotty/tools/dotc/typer/Inliner.scala
odersky: Not true for classes. They don't get erased.

src/dotty/tools/dotc/typer/Typer.scala
odersky: It turns out we cannot move this test earlier or we get inline errors. I fixed the problem by recursively calling adapt when converting to Unit. I think that's the only place where an inline was missing.

odersky:
Rebased to master

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1859,7 +1859,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       if (folded ne tree) return adaptConstant(folded, folded.tpe.asInstanceOf[ConstantType])
       // drop type if prototype is Unit
       if (pt isRef defn.UnitClass)
-        return tpd.Block(tree :: Nil, Literal(Constant(())))
+        return adapt(tpd.Block(tree :: Nil, Literal(Constant(()))), pt)
```
smarter: This is not enough, the Block itself is not inlinable, so adapting the block doesn't help:
```scala
object X {
  inline def foo: Int = 1

  def test: Unit = {
    X.foo
  }
}
```
`-Xprint:frontend`:
```scala
package <empty> {
  final lazy module val X: X$ = new X$()
  final module class X$() extends Object() { this: X.type => 
    @1 @inline() inline def foo: Int = 1
    def test: Unit = 
      {
        {
          X.foo
          ()
        }
      }
  }
}
```

odersky:
I have a tendency to merge this now. We already got fine-grained comments, and questions about more fundamental issues might just as well be expressed as separate issues. Since the PR is large it tends to get into conflicts with master quickly. It took me considerable time to rebase everything against current master.

What do people think?

src/dotty/tools/dotc/typer/Inliner.scala
smarter: Indeed, but does the JVM restrict access to classes in any way? As far as I know it only restricts access to the constructor of the class, but we take care of that by adding a term accessor for the constructor, not a type accessor.

smarter:
I think it would be good to get this code in soon, yeah

src/dotty/tools/dotc/typer/Inliner.scala
odersky: I thought access to private classes is caught by the JVM

smarter: Doesn't seem so, here's some Java code:
`Foo.java`:
```java
public class Foo {
  public class Inner {
    int get() { return 42; }
  }
  Inner inner = new Inner();
}
```
`Bar.java`:
```java
public class Bar {
  public static void main(String[] args) {
    Foo.Inner inner = (new Foo()).inner;
    System.out.println(inner.get());
  }
}
```

```shell
$ javac Foo.java
$ javac Bar.java
$ java Bar
42
$ edit Foo.java, make Inner private
$ javac Foo.java
$ java Bar
42
```

odersky: Inner classes indeed get the private modifier removed. But package inner classes are maintained in the JVM I th

smarter: Ah you're right indeed, the following breaks:
```java
package a;

public class Foo {
  public class Inner {
    public int get() { return 42; }
  }
  public Inner inner = new Inner();
}
```

```java
package b;

public class Bar {
  public static void main(String[] args) {
    a.Foo.Inner inner = (new a.Foo()).inner;
    System.out.println(inner.get());
  }
}
```

```shell
$ javac a/Foo.java
$ javac b/Bar.java
$ java b.Bar
$ Remove public from "public class Inner"
$ javac a/Foo.java
$ java b.Bar
Exception in thread "main" java.lang.IllegalAccessError: tried to access class a.Foo$Inner from class b.Bar
        at b.Bar.main(Bar.java:6)
```

I don't think adding a type alias is going to help us here since it's just going to get erased, I think we need to mark public in the bytecode all the package inner classes referenced from inline methods

smarter: Interestingly, `scalac` 2.11 and 2.12 always emit private inner classes with `ACC_PUBLIC`:
```scala
package a;

class Foo {
  private class Inner {
    def get = 42
  }
}
```
But `dotty` does emit them without `ACC_PUBLIC` unless we remove `private` from `private class Inner`.

Both `scalac` and `dotty` will prevent you from leaking a reference to private class:
```scala
package a;

class Foo {
  private class Inner {
    def get = 42
  }
  val inner = new Inner // error: non-private value inner refers to private class Inner
  type Duck = Inner // error: non-private type Duck refers to private class Inner
}
```

The easiest solution would be to always emit inner classes with `ACC_PUBLIC` as `scalac` does, @DarkDimius: do you think that this is reasonable ?

odersky:
LGTM

smarter:
LGTM too, with the following needing to be addressed as follow-ups:
- `inline` with `-language:Scala`: https://github.com/lampepfl/dotty/pull/1492#discussion_r81396243
- access to non-public types: https://github.com/lampepfl/dotty/pull/1492#discussion_r80314543

PR #1494, opened 2016-09-03
[WIP] fix encoding issues

martijnhoekstra:
the -encoding compiler flag wasn't honored at all.

Behind a WIP flag, because it depends on the earlier PR with the other test fixes for Windows, as it's very hard to build something isolated when the tests are still failing, and will need to get rabased and cleaned up at some point.


tests/run/utf16encoded.flags
```diff
@@ -0,0 +1 @@
+-encoding UTF16
```
smarter: .flags file are currently ignored by partest, and I'm not sure if there's any way to have a run test with a special set of flags currently.
/cc @DarkDimius : If we make a `run-special` folder like `pos-special`, will its content be treated as run-tests by partest?

martijnhoekstra: @Smarter That makes me realize this doesn't actually have to be a run test, as it won't compile anyway if it doesn't have the -encoding flag set to UTF-16. Should it be moved?

smarter: Yes, you should move it to pos-special and add a line next to https://github.com/lampepfl/dotty/blob/d84805299a42cb8d2c756aff5a117af24dbeaaf4/test/dotc/tests.scala#L113 like:
```scala
@Test def pos_utf16encoded = compileFile(posSpecialDir, "utf16encoded", List("-encoding", "UTF16"))
```

martijnhoekstra:
And that's a passing test-suite.

I'm particularly uncertain about what I did at https://github.com/lampepfl/dotty/pull/1494/commits/32819e2edc88dd06095704c04ed9c2dd0603386f

Should that be split up in the part where -encoding and default UTF8 encoding are honored, and the test renames? Should neg/valueClasses.scala be renamed rather than pos/valueclasses/?

I'm also not too happy with the manual bitwise copy, which seems out of place in https://github.com/lampepfl/dotty/pull/1494/commits/6bce106fea7ce10eefc864a6e7c1351675065880

sjrd:
I'll have a closer look at this during the week-end, on Windows at home.

test/test/TestREPL.scala
```diff
@@ -49,7 +49,7 @@ class TestREPL(script: String) extends REPL {
     out.close()
     val printed = out.toString
     val transcript = printed.drop(printed.indexOf(config.prompt))
-    if (transcript.toString != script) {
+    if (transcript.toString.lines.toList != script.lines.toList) {
```
sjrd: You can avoid the reification as `List`s with `sameElements`:
```scala
if (transcript.toString.lines.sameElements(script.lines))
```

test/test/TestREPL.scala
```diff
@@ -34,7 +34,7 @@ class TestREPL(script: String) extends REPL {
           while (lines.hasNext && lines.head.startsWith(continuationPrompt)) {
             val continued = lines.next
             output.println(continued)
-            buf append "\n"
+            buf append System.lineSeparator()
```
sjrd: Is this still needed, despite the comparison through `.lines` down below?

test/test/InterfaceEntryPointTest.scala
```diff
@@ -18,7 +18,7 @@ import scala.collection.mutable.ListBuffer
  */
 class InterfaceEntryPointTest {
   @Test def runCompilerFromInterface = {
-    val sources = List("./tests/pos/HelloWorld.scala")
+    val sources = List("./tests/pos/HelloWorld.scala").map(p => new java.io.File(p).getPath())
```
sjrd: I think it would be good to factor out this incantation as `normalizePath(path: String)`, since it is used 3 times (currently).

dottydoc/test/BaseTest.scala
```diff
@@ -12,7 +12,7 @@ import model.Package
 trait DottyTest {
   dotty.tools.dotc.parsing.Scanners // initialize keywords
 
-  implicit var ctx: FreshContext = {
+  implicit val ctx: FreshContext = {
```
sjrd: This change seems irrelevant to the PR/commit. It shouldn't appear in the commit.

test/test/CompilerTest.scala
```diff
@@ -413,7 +413,8 @@ abstract class CompilerTest {
     val flags = oldFlags.map(f => if (f == oldOutput) partestOutput else f) ++
       List(s"-classpath $partestOutput") // Required for separate compilation tests
 
-    getExisting(dest).isDifferent(source, flags, nerr) match {
+    val difference = getExisting(dest).isDifferent(source, flags, nerr)
+    difference match {
```
sjrd: Useless diff?

test/test/CompilerTest.scala
```diff
@@ -449,6 +450,7 @@ abstract class CompilerTest {
   /** Recursively copy over source files and directories, excluding extensions
     * that aren't in extensionsToCopy. */
   private def recCopyFiles(sourceFile: Path, dest: Path): Unit = {
+
```
sjrd: Useless diff?

test/test/CompilerTest.scala
```diff
@@ -449,11 +450,37 @@ abstract class CompilerTest {
...
+        }
+      }
+    }
+
+    processFileDir(sourceFile, { sf => 
       if (extensionsToCopy.contains(sf.extension)) {
         dest.parent.jfile.mkdirs
-        dest.toFile.writeAll("/* !!!!! WARNING: DO NOT MODIFY. Original is at: $sf !!!!! */",
-          sf.slurp())
+	      copyfile(sf, false)
```
sjrd: Indentation issue (tabs)

test/test/CompilerTest.scala
```diff
@@ -465,7 +492,7 @@ abstract class CompilerTest {
 
   /** Reads the existing files for the given test source if any. */
   private def getExisting(dest: Path): ExistingFiles = {
-    val content: Option[Option[String]] = processFileDir(dest, f => f.safeSlurp, d => Some(""))
+    val content: Option[Option[String]] = processFileDir(dest, f => try Some(f.slurp("UTF8")) catch {case io: java.io.IOException => Some(io.toString())}, d => Some(""))
```
sjrd: This ugly thing definitely need be factored out (also used a couple lines below).

What does this do, anyway? Shouldn't `safeSlurp` be responsible of this safety net?

src/dotty/tools/dotc/util/SourceFile.scala
```diff
@@ -36,9 +37,9 @@ object ScriptSourceFile {
 
 case class SourceFile(file: AbstractFile, content: Array[Char]) extends interfaces.SourceFile {
 
-  def this(_file: AbstractFile)                 = this(_file, _file.toCharArray)
-  def this(sourceName: String, cs: Seq[Char])   = this(new VirtualFile(sourceName), cs.toArray)
-  def this(file: AbstractFile, cs: Seq[Char])   = this(file, cs.toArray)
+  def this(_file: AbstractFile, codec: Codec) = this(_file, new String(_file.toByteArray, codec.charSet).toCharArray)
```
sjrd: I'm not sure using `Codec` is the best choice. `scala.io.*` is not really a nice API, and using it when we can do otherwise doesn't seem wise to me.

I suggest directly using `java.nio.Charset`.

test/test/ScannerTest.scala
```diff
@@ -16,8 +17,8 @@ class ScannerTest extends DottyTest {
   def scan(name: String): Unit = scan(new PlainFile(name))
 
   def scan(file: PlainFile): Unit = {
-    println("***** scanning " + file)
-    val source = new SourceFile(file)
+    //println("***** scanning " + file)
```
sjrd: This comment seems irrelevant to the PR.

sjrd:
Why renaming some pos tests with `pos_test` names? That seems like a workaround to a deeper issue. We should fix the original issue instead.

test/test/CompilerTest.scala
```diff
@@ -228,7 +228,7 @@ abstract class CompilerTest {
 
   private def expectedErrors(filePath: String): List[ErrorsInFile] = expectedErrors(List(filePath))
 
-  private def isNegTest(testPath: String) = testPath.contains(JFile.separator + "neg" + JFile.separator)
+  private def isNegTest(testPath: String) = testPath.contains("/neg/")
```
sjrd: Should this change have been part of another commit?

test/test/CompilerTest.scala
```diff
@@ -449,11 +450,37 @@ abstract class CompilerTest {
   /** Recursively copy over source files and directories, excluding extensions
     * that aren't in extensionsToCopy. */
   private def recCopyFiles(sourceFile: Path, dest: Path): Unit = {
-    processFileDir(sourceFile, { sf =>
+
+    def copyfile(file: SFile, bytewise: Boolean): Unit = {
+      if (bytewise) {
```
sjrd: Instead of having a boolean param, you should have two separate `def`s: `copyfile` and `copyfileBytewise`.

test/test/CompilerTest.scala
```diff
@@ -449,11 +450,37 @@ abstract class CompilerTest {
...
+        val in = file.inputStream()
+        val out = SFile(dest).outputStream()
+        val buffer = new Array[Byte](1024)
+        def loop(available: Int):Unit = {
+          if (available < 0) {()}
+          else {
```
sjrd: Consider just `if (available >= 0) {`

test/test/CompilerTest.scala
```diff
@@ -449,11 +450,37 @@ abstract class CompilerTest {
...
+            val read = in.read(buffer)
+            loop(read)
+          }
+        }
+        loop(0)
+        in.close()
+        out.close()
+      } else {
+        try {
+          SFile(dest)(scala.io.Codec.UTF8).writeAll((s"/* !!!!! WARNING: DO NOT MODIFY. Original is at: $file !!!!! */").replace("\\", "/"), file.slurp("UTF-8"))
```
sjrd: Why the `.replace("\\", "/")`?

test/test/CompilerTest.scala
```diff
@@ -449,11 +450,37 @@ abstract class CompilerTest {
...
+        } catch {
+          case unmappable: java.nio.charset.MalformedInputException => 
+            copyfile(file, true) //there are bytes that can't be mapped with UTF-8. Bail and just do a straight byte-wise copy without the warning header.
+        }
+      }
```
sjrd: Why this whole thing anyway? Files should always be copied bytewise, shouldn't they? They have an encoding before, and they should have the exact same encoding after. There is no need to decode and re-encode them.

The comment should be pure-ASCII anyway, so it doesn't really matter what encoding is used. It could actually be encoded as such (with US_ASCII) using a `replacement` char `?` for characters of `$file` that are unmappable in US-ASCII.

That would be much simpler, not to mention faster.

sjrd:
That's all I have.

Note: on my Windows machine, `sbt test` still fails on your branch. Is that expected?
```
[error] Failed: Total 166, Failed 19, Errors 0, Passed 147
[error] Failed tests:
[error]         dotc.tests
[error] (dotty/test:test) sbt.TestsFailedException: Tests unsuccessful
```

test/test/CompilerTest.scala
martijnhoekstra: `safeSlurp` is problematic in that it can't take an encoding as an argument, and uses the encoding when the `File` is created, and many ways to create a `File` don't have the opportunity to pass an encoding either (in this case `ifFile`).

That said, it can still go from this line.

test/test/CompilerTest.scala
martijnhoekstra: some path parts start with `u`. Then comment then contains `\u(something)`, which gets interpreted as an invalid unicode escape, failing compilation.

test/test/CompilerTest.scala
martijnhoekstra: the comment is pure-ASCII, but you can't write it to file if you don't know the encoding of the file; the encoding of the comment in UTF-16 or UTF-32 is unequal to the encoding in UTF-8.

martijnhoekstra:
No, on my Windows machine sbt test passes everything. I'll first get to work on your comments; that's plenty for now. Maybe after that we can see if there are failing tests in different environments?

test/test/CompilerTest.scala
martijnhoekstra: https://github.com/lampepfl/dotty/pull/1494/commits/32819e2edc88dd06095704c04ed9c2dd0603386f should probably be broken up in just the part that fixes dotty not honoring default and passed in -encoding arguments (which is an actual bug in dotty) and the parts that get the tests running on Windows (which is "just" a set of test-suite bugs).  WDYT?

test/test/CompilerTest.scala
sjrd: Ah ah!
Well, that deserves to be written as a comment above this line, then, IMO.

test/test/CompilerTest.scala
sjrd: True. I stupidly did not think about UTF-16.

That said, this code only applies to files in this repo, doesn't it? We shouldn't have UTF-16-encoded files in this repo, so it should be fine. It would be a different matter if this code were applied to arbitrary files in users' repositories, of course.

test/test/CompilerTest.scala
sjrd: That sounds good :)

test/test/CompilerTest.scala
martijnhoekstra: There is one in there now, namely the UTF-16 test - which tests whether the passed in -encoding compiler flag is honored. I didn't see any way to test that other than this. I could see if I can resort to tricks, finding an encoding/file that is both valid UTF-8, but doesn't compile as UTF-8, but is also a valid different encoding and does compile as that encoding. A third alternative is not having a test for it. None of the alternatives (this ugly thing which is only needed for one test, unclear tricks with a polyglot file encoding, or not having a test) are really appealing IMO.

sjrd: Easier: use a latin1 test instead of a UTF-8 test. Use a French `é` in some identifier, that's all it takes. I won't be valid UTF-8, but it will parse and compile as latin1. Besides, latin1 is a superset of US-ASCII, so the comment can stay US-ASCII.

martijnhoekstra: If it won't be valid UTF-8, then the file can't be slurped with UTF-8, and the file copy fails

sjrd: My comment was to use that in the potential variant where files are never slurped to being with. They are always copied byte-wise.

martijnhoekstra: Ah, I *think* I get what you mean: like this: https://github.com/martijnhoekstra/dotty/commit/42fc87a6c868b60a7b4545c9bd0a45c187f0a972 ?

sjrd: Yes, like that.

felixmulder:
As this improves the current state - we're merging this. Thanks for your contribution @martijnhoekstra !

martijnhoekstra:
@felixmulder thank you for merging it, and my sincere apologies for not coming back to this in a timely manner. I'll investigate the remainder of the cleanup that's still to be done, and make sure it gets solved.

felixmulder:
No need to apologize, your contribution improved the project -- therefore we merged it, simple as that. We'd be very grateful for further improvements on the windows side, as nobody at the lab currently uses windows.

Cheers,
Felix

PR #1499, opened 2016-09-08
Fix #1335: Generate null checks for extractors

### LINKED ISSUE #1335: Patmat: dotty does not generate null checks for extractors
```scala
object Test {
 case class MyTuple(a: Int, b: Int)
 def foo = MyTuple(1,2)
   val MyTuple(a,b) = null // should throw MatchError. Throws NPE instead.
   println(a + b)
 }
}
```
Scala2 style option-based pattern matching is handled correctly. This bug only happens with option-less pattern matching.

OlivierBlanvillain:


felixmulder:
LGTM

OlivierBlanvillain:
Hmm maybe I should have `mv tests/pos/t1335.scala tests/run/t1335.scala`

felixmulder:
Yes, that'd be great - just make a new PR for it (=

PR #1502, opened 2016-09-08
[doc] Usecases with docbase as param

felixmulder:
Cherry-picked the commit from @odersky's branch on inline that refactors `Attachment.Key[A]` => `Property.Key[A]` and implemented the docbase as a property - which will then cause the typechecking of usecases only when the docbase property is present in the context.

Other than that, not much has changed from #1472 

Review: @odersky


src/dotty/tools/dotc/core/Comments.scala
```diff
@@ -0,0 +1,423 @@
...
+     */
+    private def allInheritedOverriddenSymbols(sym: Symbol)(implicit ctx: Context): List[Symbol] = {
+      if (!sym.owner.isClass) Nil
+      else sym.allOverriddenSymbols.toList.filter(_ != NoSymbol) //TODO: could also be `sym.owner.allOverrid..`
+      //else sym.owner.ancestors map (sym overriddenSymbol _) filter (_ != NoSymbol)
+    }
+
+    class ExpansionLimitExceeded(str: String) extends Exception
+  }
+}
```
odersky: Why put this in core instead of dottydoc? A lot of the logic seems to be doc-specific. Proposal: Keep Comment as a pure data class as it was in Scanners. Implement the new logic as methods external to Comment. Comment contains only defs (the only val in it could as well be a def), so this is easy.

src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -68,6 +69,9 @@ object Contexts {
     /** The context base at the root */
     val base: ContextBase
 
+    /** Documentation base */
+    def getDocbase = property(DocContext)
```
odersky: Can we move this out of Context? The design should be ideally one of a plugin. That is, there is a context property, but the `Contexts` object knows nothing about it. I would do it more like we model `defn`: Put in the main Doc object:
    
     val DocBase = ...
     def getDocBase(implicit ctx: Context) = ctx.property(DocBase)

src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -573,10 +578,15 @@ object Contexts {
     }
   }
 
+  val DocContext = new Key[DocBase]
   class DocBase {
```
odersky: DocContext should be renamed to DocBase for consistency.
Both kinds of DocBase should also move to the dottydoc package, I think.

src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1515,7 +1515,9 @@ object SymDenotations {
 
     /** Enter a symbol in given `scope` without potentially replacing the old copy. */
     def enterNoReplace(sym: Symbol, scope: MutableScope)(implicit ctx: Context): Unit = {
-      require((sym.denot.flagsUNSAFE is Private) ||  !(this is Frozen) || (scope ne this.unforcedDecls) || sym.hasAnnotation(defn.ScalaStaticAnnot))
+      def isUsecase = ctx.property(DocContext).isDefined && sym.name.show.takeRight(4) == "$doc"
```
odersky: Move this line to a "isUserCase" method in the dottydoc package.

src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1456,6 +1462,45 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     traverse(stats)
   }
 
+  private def typedUsecases(syms: List[Symbol], owner: Symbol)(implicit ctx: Context): Unit =
```
odersky: I think expandParentDocs belongs in DottyDoc, not in Typer. So I would move typedUseCases there as well. Pass the typer as a paremeter, so it can call back to enterSymbol, createSymbol, typedStats.

src/dotty/tools/dotc/util/CommentParsing.scala
```diff
@@ -3,15 +3,10 @@
...
 
-package dotty.tools
-package dottydoc
-package model
-package comment
-
-import scala.reflect.internal.Chars._
-
-object CommentUtils {
+object CommentParsing {
```
odersky: Needs a comment what it does.

Also what's the motivation for keeping this in `util`? Is it used outside dottydoc as well?

src/dotty/tools/dotc/core/Comments.scala
felixmulder: @odersky: I think the basis of this review rests on the thoughts in this comment, as such let me try to address these concerns here in summary.

My reasons for keeping this in the compiler is for tooling - I think that's the same reason as to why this implementation was in `nsc` and not in Scaladoc for scalac.

When we have a presentation compiler - we're going to want expanded (or "cooked") comments for the standard library entities. 

The `Comment` in `core` as such - does not deal with things that dottydoc does - like `@group` and other dottydoc specific annotations - it only deals with expansion of definitions and the comment inheritance.

src/dotty/tools/dotc/core/Comments.scala
```diff
@@ -0,0 +1,446 @@
...
+
+    def docstrings: Map[Symbol, Comment] = _docstrings.toMap
+
+    def docstring(sym: Symbol): Option[Comment] = _docstrings.get(sym)
+
+    def addDocstring(sym: Symbol, doc: Option[Comment]): Unit =
+      doc.map(d => _docstrings += (sym -> d))
+  }
+
+  abstract case class Comment(pos: Position, raw: String)(implicit ctx: Context) { self =>
```
odersky: Passing a context into a class always poses the risk of space leaks. It's better not to make the context implicit, so that we can track precisely where it is used. It's even better not to pass a context into a class at all.

odersky:
LGTM!

PR #1509, opened 2016-09-14
Better `try` and `catch` messages

### LINKED ISSUE #1446: Improve error message for `try { } catch { }`
Currently the error is `error: class Unit in package scala does not take parameters`.
It should probably have a message that tells the user that he must at least have one `case`.

felixmulder:
Fixes #1446, and allows a `try` without a `catch` or `finally` as per the spec.

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1009,6 +1009,15 @@ object Parsers {
               in.nextToken()
               expr()
             } else EmptyTree
+
+          handler match {
+            case Block(Nil, EmptyTree) => syntaxError(
+              "`catch` block does not contain a valid expression, try adding a case like - `case e: Exception =>` to the block",
+              handler.pos
+            )
+            case _ =>
+          }
+
           val finalizer =
             if (in.token == FINALLY) { accept(FINALLY); expr() }
             else {
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1010,8 +1010,13 @@ object Parsers {
               expr()
             } else EmptyTree
           val finalizer =
-            if (handler.isEmpty || in.token == FINALLY) { accept(FINALLY); expr() }
-            else EmptyTree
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1009,6 +1009,15 @@ object Parsers {
               in.nextToken()
               expr()
             } else EmptyTree
+
+          handler match {
+            case Block(Nil, EmptyTree) => syntaxError(
+              "`catch` block does not contain a valid expression, try adding a case like - `case t: Throwable =>` to the block",
```
smarter: People should not use `Throwable` unless they know what they're doing, I would suggest `case e: Exception =>` instead

felixmulder: Done :)

tests/neg/emptyCatch.scala
```diff
@@ -0,0 +1,3 @@
+object Test {
+  try {} catch {} // error: `catch` block does not contain a valid expression, try adding a case like - `case t: Throwable =>` to the block
```
smarter: You left throwable here

felixmulder: Done - not sure why github is displaying the old code here...

src/dotty/tools/dotc/parsing/Parsers.scala
smarter: I'm not against that, but do we really want to recommend people to do that by default? `NonFatal` will catch `StackOverflowException` among other things

sjrd: It doesn't:
```scala
scala> try { throw new StackOverflowError } catch { case scala.util.control.NonFatal(e) => println("I caught you!") }
java.lang.StackOverflowError
  at .liftedTree1$1(<console>:8)
  ... 33 elided
```
`NonFatal` is *precisely* meant as a good default, to catch all the things you should catch, and none of those you shouldn't.

smarter: Ha you're right, I googled to find the doc and didn't realize I was looking at the 2.9 version :) http://www.scala-lang.org/api/2.9.3/scala/util/control/NonFatal$.html

PR #1511, opened 2016-09-14
Fix #1510: Fix error message when abstract member not implemented.

### LINKED ISSUE #1510: Unformated tree in error emitted when abstract member is not defined.
```scala
abstract class Foo {
  def f(i: Int): Unit
}
class Bar extends Foo {
  def f(i: String): Unit = ???
}
```

```
error: class Bar needs to be abstract, since def f: (i: Int)Unit is not defined 
(Note that TypeRef(TermRef(ThisType(TypeRef(NoPrefix,<root>)),scala),Int) does not match TypeRef(TermRef(ThisType(TypeRef(NoPrefix,scala)),Predef),String))
class Bar extends Foo {
      ^
```

Note that the types are printed as trees.

nicolasstucki:


nicolasstucki:
With this fix the error message in #1510 becomes:
```
error: class Bar needs to be abstract, since def f: (i: Int)Unit is not defined 
(Note that Int does not match String)
class Bar extends Foo {
      ^
```

odersky:
LGTM

PR #1514, opened 2016-09-15
Fix #1513: misaligned by name type parameter type bounds

### LINKED ISSUE #1513: Fixed a type parameter "by name" should be equivalent than fixing it with a value
In the following example I would expect `list.at2(Proxy[Succ[Succ[Zero]]])` and `list.at1[N = Succ[Succ[Zero]]]` to be equivalent, only the first one compiles.

```scala
object Test {
  // Heterogeneous lists and natural numbers as defined in shapeless.

  sealed trait HList
  sealed trait ::[H, T <: HList] extends HList
  sealed trait HNil extends HList

  sealed trait Nat
  sealed trait Succ[P <: Nat] extends Nat
  sealed trait Zero extends Nat

  // Accessor type class to compute the N'th element of an HList L.

  trait Accessor[L <: HList, N <: Nat] { type Out }
  object Accessor {
    type Aux[L <: HList, N <: Nat, O] = Accessor[L, N] { type Out = O }

    // (H :: T).At[Zero] = H
    implicit def caseZero[H, T <: HList]: Aux[H :: T, Zero, H] = ???

    // T.At[N] = O => (H :: T).At[Succ[N]] = O
    implicit def caseN[H, T <: HList, N <: Nat, O]
      (implicit a: Aux[T, N, O]): Aux[H :: T, Succ[N], O] = ???
  }

  case class Proxy[T]()

  def at1[NN <: Nat, OO]              (implicit e: Accessor.Aux[String :: HNil, NN, OO]): OO = ???
  def at2[NN <: Nat, OO](p: Proxy[NN])(implicit e: Accessor.Aux[String :: HNil, NN, OO]): OO = ???

  // N is fixed by a value
  at2(Proxy[Zero]): String

  // N is fixed as a type parameter (by name)
  at1[NN = Zero]: String
}
```

OlivierBlanvillain:


OlivierBlanvillain:
@odersky could you have a look?

odersky:
What about if the named parameter arguments appear in an order different from the named parameters? The `gaps` solution does not seem to cater for that? It would be good to have some more test cases. Named type args broke sometimes during the last hk types transition and have not been brought back fully.

OlivierBlanvillain:
I think the out of order cases are also covered, I tried to test for all permutations in t1513b.scala.

I added a second commit for a corner case where things would go wrong if all parameters are specified by name, but out of order.

odersky:
LGTM

PR #1528, opened 2016-09-20
WIP: delegate dotty compiler info to sbt

MasseGuillaume:


project/Build.scala
```diff
@@ -237,6 +239,25 @@ object DottyBuild extends Build {
     ).
     settings(ScriptedPlugin.scriptedSettings: _*).
     settings(
+      cleanBridge := {
```
smarter: Add a comment explaining that this is necessary until https://github.com/sbt/sbt/issues/2402 is fixed

PR #1536, opened 2016-09-24
Make positions fit for meta

odersky:
Simplifications and checks for consistent positions in parse trees.

Review by @liufengyun.

### DIFF
--- a/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/src/dotty/tools/dotc/ast/Desugar.scala
@@ -632,7 +632,7 @@ object desugar {
     val selector = makeTuple(params.map(p => Ident(p.name)))
 
     if (unchecked)
-      Function(params, Match(Annotated(New(ref(defn.UncheckedAnnotType)), selector), cases))
+      Function(params, Match(Annotated(selector, New(ref(defn.UncheckedAnnotType))), cases))
     else
       Function(params, Match(selector, cases))
   }
@@ -661,7 +661,7 @@ object desugar {
    *      tree @cls
    */
   def makeAnnotated(cls: Symbol, tree: Tree)(implicit ctx: Context) =
-    Annotated(untpd.New(untpd.TypeTree(cls.typeRef), Nil), tree)
+    Annotated(tree, untpd.New(untpd.TypeTree(cls.typeRef), Nil))
 
   private def derivedValDef(original: Tree, named: NameTree, tpt: Tree, rhs: Tree, mods: Modifiers)(implicit ctx: Context) = {
     val vdef = ValDef(named.name.asTermName, tpt, rhs)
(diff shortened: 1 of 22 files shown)


src/dotty/tools/dotc/parsing/JavaParsers.scala
```diff
@@ -576,35 +584,32 @@ object JavaParsers {
...
       atPos(cdef.pos) {
+        assert(cdef.pos.exists)
         ModuleDef(cdef.name.toTermName,
           makeTemplate(List(), statics, List(), false)).withMods((cdef.mods & (Flags.AccessFlags | Flags.JavaDefined)).toTermFlags)
       }
 
-    private val wild     = Ident(nme.WILDCARD) withPos Position(-1)
-    private val wildList = List(wild) // OPT This list is shared for performance.
+    private def wild     = Ident(nme.WILDCARD)
+    private def wildList = List(wild) // OPT This list is shared for performance.
```
smarter: Not shared anymore since it's a def, it could be removed

xeno-by:
@odersky @liufengyun Could you elaborate on the motivation behind this pull request?

odersky:
We want parse trees to be as close to source as possible and we want them to have predictable positions. After all positions is all we have to relate to meta trees.

src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -505,7 +505,7 @@ object desugar {
       val clsTmpl = cpy.Template(tmpl)(self = clsSelf, body = tmpl.body)
       val cls = TypeDef(clsName, clsTmpl)
         .withMods(mods.toTypeFlags & RetainedModuleClassFlags | ModuleClassCreationFlags)
-      Thicket(modul, classDef(cls))
+      Thicket(modul, classDef(cls).withPos(mdef.pos))
```
odersky: No, the withPos applies to the classDef here, not the thicket.

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -63,15 +63,16 @@ object Parsers {
       atPos(Position(start, end, point))(t)
 
     def atPos[T <: Positioned](start: Offset, point: Offset)(t: T): T =
-      atPos(start, point, in.lastOffset max start)(t)
+      if (in.lastOffset > start) atPos(start, point, in.lastOffset)(t) else t
```
odersky: No, that's actually a common case. It always happens when the generated tree does not consume any source. In that case we don't set a position and wait for the position to be set by the enclosing node. I'll add a comment to explain.

felixmulder:
@odersky: If we want to merge the syntax highlighting with the scanner/parser we should give comments a token and position as well - not sure if it should be a part of this PR though...

odersky:
@felixmulder Let's discuss comment handling separately.

PR #1538, opened 2016-09-25
Simplify trees

odersky:
Get rid of Pair and SelectorFromTypeTree nodes. Neither had an important role to play, so the result is shorter and simpler code than what we had before.


DarkDimius:
👍

odersky:
@DarkDimius It's in the doc comment of the Import tree.

odersky:
@darldimius Do you want to do the review?

smarter:
LGTM, though it may be nicer to have an extractor for an import rename:
```scala
  object ImportRename {
    def unapply[T >: Untyped](thicket: Thicket[T]): Option[(Ident[T], Ident[T])] = thicket match {
      case Thicket((name @ Ident(_)) :: (rename @ Ident(_)) :: Nil) => Some(name, rename)
      case _ => None
    }
  }
```

odersky:
@smarter I am a bit concerned Import selector analysis might be time-critical, that's why I preferred the non-extractor solution, because it can be optimized better.

PR #1539, opened 2016-09-26
Drop modifiers

odersky:
The end goal of this PR is to get rid of the Modifiers structure and to represent each modifier as a tree of its own. This is more IDE friendly and simplifies treatment of trees. 

@DarkDimius can you comment on this? Also, have a look at my comments in BackendInterface which require action.

### DIFF
--- a/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -70,7 +70,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
   type Bind            = tpd.Bind
   type New             = tpd.New
   type Super           = tpd.Super
-  type Modifiers       = tpd.Modifiers
+  type Modifiers       = Null
   type Annotation      = Annotations.Annotation
   type ArrayValue      = tpd.JavaSeqLiteral
   type ApplyDynamic    = Null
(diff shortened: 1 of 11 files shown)


src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -944,7 +944,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
   }
 
   object ValDef extends ValDefDeconstructor {
-    def _1: Modifiers = field.mods
+    def _1: Modifiers = tpd.Modifiers(field.symbol)
```
odersky: Note that modifiers will be re-created as a new object everytime mods is called. Is it possible to split accessor into two accessors, one for flags and another for annotations? If not, an alternative would be to let Symbol inherit from Modifiers.

src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -249,17 +249,6 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
   /** Is this case guarded? */
   def isGuardedCase(cdef: CaseDef) = cdef.guard ne EmptyTree
 
-  /** True iff definition is a val or def with no right-hand-side, or it
-   *  is an abstract typoe declaration
```
odersky: This function was unused, superseded by NavigateAST.

src/dotty/tools/backend/jvm/DottyBackendInterface.scala
xeno-by: Do you also plan to retain original positions of modifiers as written in the source file?

odersky: In the parse-trees yes. But these are not preserved in typed trees or TASTY.

DarkDimius: Grepping the source code it seems that _1 is never used. Try returning `null` to ensure it?

felixmulder:
I tried this locally and everything seems to be fine. I'll try to rebuild this again.

DarkDimius:
LGTM

Issue #1175, opened 2016-03-14
def in structural type is rejected, but val is accepted


samuelgruetter:
``` scala
object StructuralTypes {
  type A = { def foo: Int }
  type B = { val foo: Int }
}
```

dotty rejects this example:

```
[error] /home/sam/Documents/git/scalameta/tutorial/src/main/resources/StructuralTypes.scala:2: error: refinement method foo without matching type in parent Object
[error]   type A = { def foo: Int }
[error]                  ^
[error] one error found
```

scalac accepts it, and I guess dotty should as well?

odersky:
We have not decided yet to what degree dotty will support this.


On Mon, Mar 14, 2016 at 2:48 PM, Samuel Gruetter <notifications@github.com>
wrote:

> object StructuralTypes {
>   type A = { def foo: Int }
>   type B = { val foo: Int }
> }
>
> dotty rejects this example:
>
> [error] /home/sam/Documents/git/scalameta/tutorial/src/main/resources/StructuralTypes.scala:2: error: refinement method foo without matching type in parent Object
> [error]   type A = { def foo: Int }
> [error]                  ^
> [error] one error found
>
> scalac accepts it, and I guess dotty should as well?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lampepfl/dotty/issues/1175>.
>



-- 
Martin Odersky
EPFL

olafurpg:
If structural types will not be supported, I think we will need a good migration story for patterns where structural types are used as type classes.

For example, consider the `Closeable` structural type in [better-files](https://github.com/pathikrit/better-files/blob/a45f9056ce80581c41c033a79756ec15629cb01e/core/src/main/scala/better/files/package.scala#L8-L10):

```scala
  type Closeable = {
    def close(): Unit
  }
```

The `Closeable` structural type is used to pimp unrelated types such as `java.io.Closeable` and `scala.io.Source`, like this

```scala
  implicit class CloseableOps[A <: Closeable](resource: A) {
    /**
      * Lightweight automatic resource management
      */
    def autoClosed: ManagedResource[A]
  }
```

The best rewrite I could come up with for this situation was changing `Closeable` to a type class and manually provide typeclass instances for each type.

```scala
trait Closeable[T] {
  def close(): Unit
}
object Closeable {
  // Note. `extends Closeable[java.io.Closeable]` is not correct because then
  // we lose the original type
[...]

PR #1550, opened 2016-10-01
True union types

odersky:
Keep union types, instead of approximating them by their dominators. This change surfaced
several other errors which were due to interactions of union types with other features.

Review by @smarter


tests/neg/OrType.scala
```diff
@@ -2,7 +2,7 @@ class B(val x: Int)
 class C(val x: Double)
 
 object Test{
-  def bar(x: B | C): Int | Double = x.x
+  def bar(x: B | C): Int | Double = x.x  // error
```
DarkDimius: why is it an error now?

smarter: I assume this is caused by the change to `findMember`, we now always approximate union prefixes, I think this should be reconsidered since it considerably diminishes the usability of union types, it also breaks the nice symmetry with selection on intersection types.

smarter: @odersky If you don't have the time to investigate the issues with `union-types-narrow-prefix` I could have a look next week.

odersky: @smarter Do you have a realistic scenario where the utility of union types is diminished?

odersky: @darkdimius The previous findMember operation on union types would often fail to instantiate member types properly. I wonder whether that's an issue in the linker as well, since linker sems to use union types pervasively.

smarter: I think it might help in some situations where people currently use structural types, for example https://github.com/lampepfl/dotty/issues/1175#issuecomment-242684995.
The proposed solution is to use a typeclass but if you assume that everyone should extend `java.io.Closeable` but still want to handle a few special cases that do not implement the interface but still follows its contract (like `scala.io.Source`) you could do:
```scala
type Closeable = java.io.Closeable | scala.io.Source
```

smarter: Another use case is reducing boilerplate in pattern matching without introducing intermediate traits:
```scala
sealed trait Tree
class Assign(lhs: Tree, rhs: Tree) extends Tree
class Plus(lhs: Tree, rhs: Tree) extends Tree

(x: Tree) match {
  case x: (Assign | Plus) => foo(x.lhs, x.rhs)
}
```

odersky: @smarter I pushed `union-types-narrow-prefix` to staging. I am a bit skeptical this one can be made to work, though. In particular I don't know what the theoretical foundations of doing it this way would be.

smarter: I also assume that there might be nice use cases when interfacing with JS APIs, https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#union-types has some examples.

smarter: @odersky I wonder if we could somehow make something like https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/Splitter.scala be a part of Typer to avoid having to select on a union type at all

odersky: @smarter Both examples make use of the fact that the alternatives have "coincidentally" members with the same type. Both also require possibly expensive splitting under the cover. Maybe it's more honest to require one match for `Assign` and another for `Plus`. For now, it looks to me like the use cases are close to code smells.

odersky: >  I wonder if we could somehow make something like https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/Splitter.scala be a part of Typer to avoid having to select on a union type at all

Yes, I believe we'd need something like that to make findMember work correctly without widening the qualifier. This would work if findMember was only called when typing a Select node in source code. Unfortunately, it's called in a lot of other situations as well...

odersky: To summarize it, the change is on the safe, conservative side. It's a simple way to avoid bugs and keep result types small (no propagation of disjunctions). It requires sometimes manual splitting, hence more verbose code. But it seems that code can always transformed through manual splitting to fit the new model.

odersky: Regarding the linker, if disjunction propagation through selection is required, the linker could do the splitting itself.

odersky: Another note: The change led to a 5% increase in compilation speed on the junit tests. Not a scientific benchmark, so it might not be a very exact number. But I was fearing it would slow compilation down, so it's nice to see an indication to the contrary.

odersky: @smarter Actually, even splitting Select nodes is non-trivial. Can you come up with a safe splitting scheme that guarantees all types make sense. I couldn't, because splitting changes path identity. I think that highlights some of the problem we are facing here.

nicolasstucki:
Regarding `dotty.language` you should also remove it from `Definitions`. There is currently a `lazy val LanguageModuleRef = ctx.requiredModule("dotty.language")` and a `lazy val Scala2LanguageModuleRef = ctx.requiredModule("scala.language")`.

tests/neg/OrType.scala
sjrd: > I also assume that there might be nice use cases when interfacing with JS APIs

I don't think widening the prefix of member selection will be an issue for JS APIs. Also, I agree with @odersky that the use cases look like code smell to me.

smarter:
Rebasing before merging to make sure the tests still pass.

PR #1555, opened 2016-10-04
Migrate wiki to docs dir

felixmulder:
This PR migrates the wiki to the docs dir so that we can create better documentation in the long run. It also adds the blog posts concerning dotty that are currently featured on http://dotty.epfl.ch/blog.

Once merged the website will be set up to poll the repo and host the documenation on http://dotty.epfl.ch/docs


felixmulder:
@OlivierBlanvillain - would you mind reviewing?

smarter:
Can we have the same disqus comment thread for the blog posts which are both on scala-lang and on the dotty website?

felixmulder:
If we want to - we just have to choose the same thread ID IIUC.

felixmulder:
But I agree with whoever said it, that most discussion is still on HN/Reddit

smarter:
Maybe we should link to a HN or Reddit thread instead of having disqus comment then.

felixmulder:
@odersky - I'd like to work on the site to render this. Is there anything else we should consider before merging this?

PR #1556, opened 2016-10-06
Improve message when directory is passed dotc.

olafurpg:
Previously, running ./bin/dotc <directory> caused a null pointer
exception. Now, the error is:

`error: expected file, received directory <directory>`

### DIFF
--- a/src/dotty/tools/dotc/Run.scala
+++ b/src/dotty/tools/dotc/Run.scala
@@ -29,10 +29,14 @@ class Run(comp: Compiler)(implicit ctx: Context) {
   var units: List[CompilationUnit] = _
 
   def getSource(fileName: String): SourceFile = {
-    val encoding = ctx.settings.encoding.value
     val f = new PlainFile(fileName)
-    if (f.exists) new SourceFile(f, Codec(encoding))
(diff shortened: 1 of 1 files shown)


src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -30,7 +30,7 @@ object Formatting {
         catch {
           case NonFatal(ex) => s"[cannot display due to $ex, raw string = $toString]"
         }
-      case _ => arg.toString
+      case _ => s"$arg"
```
felixmulder: It does not look like it - @olafurpg, could you remove this from the PR?

felixmulder: I see, however, I still don't think we should be allowing `null.show`.

felixmulder: It's arguably better if this blows up :)

olafurpg: Fine, I've removed the change.

olafurpg:
I've rebased now.

PR #1560, opened 2016-10-06
Harmonize PolyType and TypeLambda

odersky:
Push as much as we can into common superclass GenericType. Review by @smarter.

Update: I managed to completely eliminate any distinction between PolyType and TypeLambda. They are the same type now, and GenericType as a supertype could also be eliminated.

### DIFF
--- a/docs/syntax-summary.txt
+++ b/docs/syntax-summary.txt
@@ -96,7 +96,7 @@ grammar.
   ClassQualifier    ::=  `[' id `]'
 
   Type              ::=  FunArgTypes `=>' Type                                  Function(ts, t)
-                      |  HkTypeParamClause `->' Type						    TypeLambda(ps, t)
+                      |  HkTypeParamClause `=>' Type						    TypeLambda(ps, t)
                       |  InfixType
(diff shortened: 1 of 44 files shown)


smarter:
Needs to be rebased after the inline merge

odersky:
Rebased to master

odersky:
In theory this commit paves the way to allow polymorphic types as value types. A test of some
of the basic functionality is in pending/pos/polyTypes. This does not work yet,  because `adapt` forces every polytype it sees to be fully applied. Enabling this functionality should be a SIP and separate PR.

smarter:
Rebased after the union-type PR merge

smarter:
LGTM, I wonder if we can merge `LambdaParam` and `PolyParam` ?

odersky:
> LGTM, I wonder if we can merge LambdaParam and PolyParam ?

Yes, interesting question. PolyParam is a reference to a parameter, whereas LambdaParam is a surrogate for a symbol, i.e. a definition. So it's not obvious, but still worth thinking about it.

PR #1561, opened 2016-10-06
Fix build failure after merging #1492

smarter:
The one-parameter constructor of SourceFile was removed in #1494

Review by @felixmulder


felixmulder:
LGTM

smarter:
To be super-correct we should probably store the encoding in the Tasty section instead of using the encoding passed by the user, but let's worry about this when our build works.

PR #1565, opened 2016-10-06
Small fixes to some warnings/comment typos

Blaisorblade:
Just my 2 cents—things I noticed while taking a look at the architecture.


felixmulder:
@Blaisorblade - let me know if you're still working on this. Otherwise we'd be happy to merge it :)

Blaisorblade:
Thanks—let me at least finish #1572, I might find more :-).

I've also looked at other warnings but they're trickier. IIUC I could fix the ones about `final case class` uses by lifting those classes out, but not sure where to place those classes.

Blaisorblade:
#1572's done, so go ahead! (No further time for now for this, too many deadlines coming).

PR #1578, opened 2016-10-11
use usr/bin/env bash since it's more general then /bin/bash

MasseGuillaume:


felixmulder:
Thanks for taking the time to look at this. LGTM!

PR #1582, opened 2016-10-11
linking cbt usage, merge after #1581

cvogt:


felixmulder:
Thanks @cvogt - looks great: http://lampepfl.github.io/dotty/docs/usage/cbt-projects.html

PR #1585, opened 2016-10-12
Fix #1552: REPL file loading

### LINKED ISSUE #1552: Load file in repl -  eof expected but '}' found
The example below failed with `<console>:8: error: eof expected but '}' found;}^`
```scala
def foo(): String = { 
  "foo" 
}

class B {

}
```
But when i place braces in one line - it's work
```scala
def foo(): String = { "foo" }

class B {}
```

dos65:


felixmulder:
Thank you @dos65 for taking both the time to report this issue - as well as fixing it! LGTM

PR #1587, opened 2016-10-13
Fix Tasty positions

odersky:
Make Tasty positions read back exactly the way they were serialized.

The aim is to turn -Yprintpos on and to get exact correspondence between before-pickling and after-pickling.

### DIFF
--- a/src/dotty/tools/dotc/core/tasty/PositionPickler.scala
+++ b/src/dotty/tools/dotc/core/tasty/PositionPickler.scala
@@ -36,9 +36,9 @@ class PositionPickler(pickler: TastyPickler, addrsOfTree: tpd.Tree => List[Addr]
       if (it.hasNext) Some(it.next) else None
   }
 
-  def header(addrDelta: Int, hasStartDelta: Boolean, hasEndDelta: Boolean) = {
+  def header(addrDelta: Int, hasStartDelta: Boolean, hasEndDelta: Boolean, hasPoint: Boolean) = {
     def toInt(b: Boolean) = if (b) 1 else 0
-    (addrDelta << 2) | (toInt(hasStartDelta) << 1) | toInt(hasEndDelta)
+    (addrDelta << 3) | (toInt(hasStartDelta) << 2) | (toInt(hasEndDelta) << 1) | toInt(hasPoint)
   }
 
   def picklePositions(roots: List[Tree])(implicit ctx: Context) = {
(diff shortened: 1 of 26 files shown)


src/dotty/tools/dotc/core/tasty/TreePickler.scala
```diff
@@ -438,15 +438,10 @@ class TreePickler(pickler: TastyPickler) {
...
-      case tree: Inlined =>
-        // Why drop Inlined info when pickling?
-        // Since we never inline inside an inlined method, we know that
-        // any code that continas an Inlined tree is not inlined itself.
-        // So position information for inline expansion is no longer needed.
-        // The only reason to keep the inline info around would be to have fine-grained
-        // position information in the linker. We should come back to this
-        // point once we know more what we would do with such information.
-        pickleTree(Inliner.dropInlined(tree))
+      case Inlined(call, bindings, expansion) =>
```
smarter: As an alternative to save space, could we just pickle the `call` and re-inline it in the unpickler?

odersky: It might be possible in some situations, but it does not look straightforward. In other situations (e.g. inlined trees materialized by implicit macros) it's impossible.

src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -752,6 +760,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
       val parentRefs = ctx.normalizeToClassRefs(parents.map(_.tpe), cls, cls.unforcedDecls)
       val self =
         if (nextByte == SELFDEF) {
+          val selfStart = currentAddr
```
smarter: Unused `val`

src/dotty/tools/dotc/core/tasty/PositionPickler.scala
```diff
@@ -31,29 +48,44 @@ class PositionPickler(pickler: TastyPickler, addrOfTree: tpd.Tree => Option[Addr
...
+              //println(i"pickling $x with $pos at $addr")
+              pickleDeltas(addr.index, pos)
             case _ =>
+              //println(i"no address for $x")
           }
         }
+        //else if (x.pos.exists) println(i"skipping $x")
         x match {
-          case x: MemberDef @unchecked => traverse(x.symbol.annotations, x.pos)
+          case x: MemberDef @unchecked => traverse(x.symbol.annotations.map(_.tree))
```
smarter: Could be `x.symbol.annotations.foreach(ann => traverse(ann.tree))` to avoid allocating an intermediate list

PR #1592, opened 2016-10-13
Fix #1590: Eliminate wildcards when approximating a type

### LINKED ISSUE #1590: java.lang.Error: type is not fully defined
This minimized case class:

```scala
case class W[T](seq: Option[Option[T]] = Option.empty)
```

Stops the compilation with the following exception:

```
Exception in thread "main" java.lang.Error: internal error: type of right-hand side Option[A] is not fully defined, pos = <58..98>
```

([full stack trace](https://gist.github.com/OlivierBlanvillain/7594cd2d17f4f6dde879cdb0770c81d1))

odersky:
Fixes #1590. Type variables should never be instantiated to types
containing wildcards. 

Review by @smarter

### DIFF
--- a/src/dotty/tools/dotc/core/ConstraintHandling.scala
+++ b/src/dotty/tools/dotc/core/ConstraintHandling.scala
@@ -177,7 +177,25 @@ trait ConstraintHandling {
           case tp: RefinedType if param occursIn tp.refinedInfo => tp.parent
           case tp: WildcardType =>
             val bounds = tp.optBounds.orElse(TypeBounds.empty).bounds
-            if (fromBelow == (variance >= 0)) bounds.lo else bounds.hi
+            // Try to instantiate the wildcard to a type that is known to conform to it.
+            // This means:
+            //  If fromBelow is true, we minimize the type overall
+            //  Hence, if variance < 0, pick the maximal safe type: bounds.lo
+            //           (i.e. the whole bounds range is over the type)
+            //         if variance > 0, pick the minimal safe type: bounds.hi
+            //           (i.e. the whole bounds range is under the type)
+            //         if variance == 0, pick bounds.lo anyway (this is arbitrary but in line with
+            //           the principle that we pick the smaller type when in doubt).
+            //  If fromBelow is false, we maximize the type overall and reverse the bounds
+            //  if variance != 0. For variance == 0, we still minimize.
+            //  In summary we pick the bound given by this table:
+            //
+            //  variance    | -1  0   1
+            //  ------------------------
+            //  from below  | lo  lo  hi
+            //  from above  | hi  lo  lo
+            //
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/core/ConstraintHandling.scala
```diff
@@ -162,7 +162,8 @@ trait ConstraintHandling {
   /** Solve constraint set for given type parameter `param`.
    *  If `fromBelow` is true the parameter is approximated by its lower bound,
    *  otherwise it is approximated by its upper bound. However, any occurrences
-   *  of the parameter in a refinement somewhere in the bound are removed.
+   *  of the parameter in a refinement somewhere in the bound are removed. Also
+   *  wildcard types in bounds are approximated by their upper or lower bounds.
    *  (Such occurrences can arise for F-bounded types).
```
smarter: This should be moved before the sentence you added

smarter:
With this fix, we end up with the equivalent of:
```scala
case class W[T](seq: Option[Option[T]] = Option.empty[Option[Any]])
```
But what we really want is:
```scala
case class W[T](seq: Option[Option[T]] = Option.empty[Option[T]])
```
The difference is apparent once you try to use the class:
```scala
case class W[T](seq: Option[Option[T]] = Option.empty)

object Test {
  W[Int]()
}
```
This compiles with scalac but with this patch we get:
```scala
-- [E007] Type Mismatch Error: tests/pos/i1590.scala -------------------------------------------------------------------
4 |  W[Int]()
  |  ^^^^^^^^
  |  found:    Option[Option[Any]]
  |  required: Option[Option[Int]]
  |  
```

odersky:
We now get `Option[Option[Nothing]]` which is what `scalac` does as well. I remember there was a long discussion on JIRA somewhere why we cannot do better. The gist is that the default argument is not required to conform to the parameter type, so constraining it to be a subtype is wrong.

src/dotty/tools/dotc/core/ConstraintHandling.scala
```diff
@@ -177,7 +177,25 @@ trait ConstraintHandling {
...
             val bounds = tp.optBounds.orElse(TypeBounds.empty).bounds
-            if (fromBelow == (variance >= 0)) bounds.lo else bounds.hi
+            // Try to instantiate the wildcard to a type that is known to conform to it.
+            // This means:
+            //  If fromBelow is true, we minimize the type overall
+            //  Hence, if variance < 0, pick the maximal safe type: bounds.lo
+            //           (i.e. the whole bounds range is over the type)
+            //         if variance > 0, pick the minimal safe type: bounds.hi
+            //           (i.e. the whole bounds range is under the type)
+            //         if variance == 0, pick bounds.lo anyway (this is arbitrary but in line with
```
smarter: Ah nevermind, I misread the condition.

PR #1593, opened 2016-10-13
Fixes to Type#isRef

smarter:
Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -121,7 +121,7 @@ object Types {
         }
       case this1: RefinedType if stripRefinements => this1.parent.isRef(sym, stripRefinements)
       case this1: RecType => this1.parent.isRef(sym, stripRefinements)
-      case this1: HKApply => this1.superType.isRef(sym, stripRefinements)
+      case this1: HKApply => this1.superType.isRef(sym, stripRefinements) && this1.lowerBound.isRef(sym, stripRefinements)
       case _ => false
     }
 
--- a/src/dotty/tools/dotc/core/Types.scala
+++ b/src/dotty/tools/dotc/core/Types.scala
@@ -106,20 +106,22 @@ object Types {
       case _ => false
     }
 
-    /** Is this type a (possibly refined or applied or aliased) type reference
-     *  to the given type symbol?
-     *  @sym  The symbol to compare to. It must be a class symbol or abstract type.
-     *        It makes no sense for it to be an alias type because isRef would always
-     *        return false in that case.
+    /** Is this type a (possibly aliased) type reference to the given type symbol?
+     *
+     *  @param sym               The symbol to compare to. It must be a class symbol or abstract type.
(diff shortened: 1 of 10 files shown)


smarter:
Alright, the tests pass now :)

src/dotty/tools/dotc/core/Types.scala
```diff
@@ -106,20 +106,22 @@ object Types {
...
         this1.info match { // see comment in Namer#typeDefSig
-          case TypeAlias(tp) => tp.isRef(sym)
+          case TypeAlias(tp) => tp.isRef(sym, stripRefinements)
           case _ =>  this1.symbol eq sym
         }
-      case this1: RefinedOrRecType => this1.parent.isRef(sym)
-      case this1: HKApply => this1.superType.isRef(sym)
+      case this1: RefinedType if stripRefinements => this1.parent.isRef(sym, stripRefinements)
+      case this1: RecType => this1.parent.isRef(sym, stripRefinements)
+      case this1: HKApply => this1.superType.isRef(sym, stripRefinements) && this1.lowerBound.isRef(sym, stripRefinements)
```
odersky: If `stripRefinements == false`, it seems we should also stop at `RecType` and `HKApply` since they are conceptually analogous to `RefinedType`.

odersky: I tend to think it would be better to split the method in two. `isRef` and `isRefInstance` (?). `isRefInstance` could presumably just be this:

    def isRefInstance(sym: Symbol)(implicit ctx: Context): Boolean = dealias match {
       case this1: TypeRef => this1.symbol eq sym
       case this1: RefinedOrRecType => this1.parent.isRef(sym)
       case this1: HKApply => this1.superType.isRef(sym) && this1.lowerBound.isRef(sym)
       case _ => false
    }

odersky: And `isRef` would just be the remaining cases. That seems simpler logic and avoids the boolean arguments in the calls.

smarter: Having two methods would be nice yes, I avoided that so that cases like refined aliases of refined aliases of refined instances work, but that might not be important. An additional difficulty is that we cannot use `dealias` because of `Namer#typeDefSig`. Another option would be to have:
```scala
private[this] def isRef(sym: Symbol, stripRefinements: Boolean = false)(...) = ...
def isRef(sym: Symbol) = isRef(sym)
def isAppliedRef(sym: Symbol) = isRef(sym, true)
```

felixmulder:
@smarter - please update :)

allanrenucci:
@smarter I tried the two test cases in this PR on latest master and they pass. Should I open a new PR with the two test cases and close this one?

nicolasstucki:
@smarter apart for these two test cases do you have another way to reproduce it?

milessabin:
This is the sort of thing I had in mind for `checkTypes`.

odersky:
@smarter What to do with this? Do we still need this?

smarter:
It'd have to be redone from scratch anyway so we can close this one.

PR #1595, opened 2016-10-14
Fix #1567: Widen private constructor in value class

### LINKED ISSUE #1567: AnyVal Private constructor inlined (assertion error)
This subtle example minimized from ScalaTest code base requires two files and a particular compilation order:

*1.scala:*
```scala
final class PosZInt private (val value: Int) extends AnyVal

object PosZInt {
  def from(value: Int): Option[PosZInt] =
    if (value >= 0) Some(new PosZInt(value)) else None
}
```

*2.scala:*
```scala
object Test {
  scala.util.Try(PosZInt.from(1).get)
}
```

`run 2.scala 1.scala`

Leads to:

```
assertion failed: private constructor PosZInt in class PosZInt in 1.scala accessed from method $anonfun in object Test in 2.scala
```

I'm guessing the generated code is correct, so maybe it should be enough to remove / relax the assertion?

odersky:
Private or protected constructors of value classes need to be widenened
to public in order to enable boxing anywhere.

Technically we should also do something about qualified private constructors, but since we
want to get rid of them anyway it's urgent.

Review by @OlivierBlanvillain

### DIFF
--- a/src/dotty/tools/dotc/transform/ExtensionMethods.scala
+++ b/src/dotty/tools/dotc/transform/ExtensionMethods.scala
@@ -32,6 +32,9 @@ import SymUtils._
  * in [[ElimErasedValueType]].
  * This is different from the implementation of value classes in Scala 2
  * (see SIP-15) which uses `asInstanceOf` which does not typecheck.
+ *
+ * Finally, if the constructor of a value class is private pr protected
+ * it is widened to public.
  */
 class ExtensionMethods extends MiniPhaseTransform with DenotTransformer with FullParameterization { thisTransformer =>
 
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/transform/ExtensionMethods.scala
```diff
@@ -32,6 +32,9 @@ import SymUtils._
  * in [[ElimErasedValueType]].
  * This is different from the implementation of value classes in Scala 2
  * (see SIP-15) which uses `asInstanceOf` which does not typecheck.
+ *
+ * Finally, if the constructor of a value class is private pr protected
```
smarter: typo: pr -> or

PR #1596, opened 2016-10-14
Fix #1531: Ignore private members when looking for abstract ones

### LINKED ISSUE #1531: Abstract members wrongly implemented by class arguments
The following code compiles:

```scala
trait T {
  def f: Int
}

class A(f: Int) extends T
```

I would expect to fail with following error like it does in scalac:

`error: class A needs to be abstract, since def f: => Int is not defined`

odersky:
Private members do not override abstract ones. So when looking for
abstract members we need to search with `nonPrivateMember`, not
`member`.

Fixes #1531. Review by @smarter.


smarter:
LGTM

PR #1597, opened 2016-10-14
Fix #1540: overloaded get and isDefined in option-less patmat

### LINKED ISSUE #1540: Overloaded `isDefined` or `get` breaks option-less pattern matching
Overloading `isDefined` or `get` as in the following example crashes dotc on this [assertion](https://github.com/lampepfl/dotty/blob/80cbe114cc5d641d66512c58402eca4ae69f9f2a/src/dotty/tools/dotc/transform/Erasure.scala#L335):

```scala
class Casey1(val a: Int) {
  def isDefined: Boolean = true
  def isDefined(x: Int): Boolean = ???
  def get: Int = a
  def get(x: Int): String = ???
}
object Casey1 { def unapply(a: Casey1) = a }

object Test {
    val c @ Casey1(x) = new Casey1(0)
    assert(x == c.get)
  }
}
```

liufengyun:
Fix #1540 .

When `isDefined` and `get` are overloaded, make sure the translated code use the correct version of the methods.

### DIFF
--- a/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -240,7 +240,7 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
           val isDefined = extractorMemberType(prev.tpe, nme.isDefined)
 
           if ((isDefined isRef defn.BooleanClass) && getTp.exists) {
-            // isDefined and get maybe overloaded
+            // isDefined and get may be overloaded
             val getDenot = prev.tpe.member(nme.get).suchThat(_.info.isParameterless)
             val isDefinedDenot = prev.tpe.member(nme.isDefined).suchThat(_.info.isParameterless)
 
--- a/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -240,17 +240,21 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
           val isDefined = extractorMemberType(prev.tpe, nme.isDefined)
 
           if ((isDefined isRef defn.BooleanClass) && getTp.exists) {
-            val tmpSym = freshSym(prev.pos, prev.tpe, "o")
-            val prevValue = ref(tmpSym).select("get".toTermName).ensureApplied
+            // isDefined and get maybe overloaded
+            val getDenot = prev.tpe.member(nme.get).suchThat(_.info.isParameterless)
+            val isDefinedDenot = prev.tpe.member(nme.isDefined).suchThat(_.info.isParameterless)
 
-              Block(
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -240,17 +240,21 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
...
 
-              Block(
-                List(ValDef(tmpSym, prev)),
-                // must be isEmpty and get as we don't control the target of the call (prev is an extractor call)
-                ifThenElseZero(
-                  ref(tmpSym).select(nme.isDefined),
-                  Block(List(ValDef(b.asTerm, prevValue)), next)
-                )
+            val tmpSym = freshSym(prev.pos, prev.tpe, "o")
+            val prevValue = ref(tmpSym).select(getDenot.namedType).ensureApplied
```
odersky: I don't think that can work. getDenot has a type that's dependent on `prev`. But the denotation gets added to another prefix, `tmpSym`. You could try instead:

    ref(tmpSym).select(getDenot.symbol).ensureApplied

liufengyun: Thanks for the tip @odersky !

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -240,17 +240,21 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
...
-                  Block(List(ValDef(b.asTerm, prevValue)), next)
-                )
+            val tmpSym = freshSym(prev.pos, prev.tpe, "o")
+            val prevValue = ref(tmpSym).select(getDenot.symbol).ensureApplied
+
+            Block(
+              List(ValDef(tmpSym, prev)),
+              // must be isEmpty and get as we don't control the target of the call (prev is an extractor call)
+              ifThenElseZero(
+                ref(tmpSym).select(isDefinedDenot.namedType),
```
liufengyun: Thanks for the feedback @odersky , I was wondering why the second succeeded without change. Now I added another test that fails my initial attempt.

odersky:
LGTM, thanks!

src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -240,17 +240,21 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
           val isDefined = extractorMemberType(prev.tpe, nme.isDefined)
 
           if ((isDefined isRef defn.BooleanClass) && getTp.exists) {
-            val tmpSym = freshSym(prev.pos, prev.tpe, "o")
-            val prevValue = ref(tmpSym).select("get".toTermName).ensureApplied
+            // isDefined and get maybe overloaded
```
liufengyun: Thanks  for the catch @smarter , it's fixed in the latest commit.

PR #1598, opened 2016-10-14
Fix #1515: Don't narrow gadt bounds when frozen

### LINKED ISSUE #1515: Pattern matching on GATD generating `.asInstanceOf[Nothing]`
The following example:

```scala
sealed trait Trait[T]

final case class Case[T](e: T) extends Trait[T]

object Demo {

    def f[H](t: Trait[H]): Unit =
      t match {
        case Case(e) => println(Some(e))
      }

    f(Case(1))

  }
}
```

Fails at run time with `java.lang.ClassCastException: java.lang.Integer cannot be cast to scala.runtime.Nothing$`.

odersky:
Fixes #1515. Review by @smarter.


smarter:
LGTM, that was indeed one of the things I wanted to fix in GADTs :).

PR #1599, opened 2016-10-14
Fix #1570: Allow inline parameters as inline args

### LINKED ISSUE #1570: Question on semantics of inlined parameters.
```scala
inline def foo(inline n: Int) = bar(n)
inline def bar(inline n: Int) = n
```

currently has the following error 
```
error: argument to inline parameter must be a constant expression or a function
  inline def foo(inline n: Int) = bar(n)
                                      ^
```

I would assume that `n` should be considered as a constant as at call site it will be a constant and `foo` is going to be inlined. Hence this should be allowed. Is this correct? Or, I'm missing something else?

odersky:
Inline parameters can always be passed to other
inline parameters.

Fixes #1570. Review by @nicolasstucki

### DIFF
--- a/src/dotty/tools/dotc/core/Flags.scala
+++ b/src/dotty/tools/dotc/core/Flags.scala
@@ -544,6 +544,9 @@ object Flags {
   /** An inline method */
   final val InlineMethod = allOf(Inline, Method)
 
+  /** An inline parameter */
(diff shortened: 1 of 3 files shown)


nicolasstucki:
LGTM

Issue #1589, opened 2016-10-13
Missing error messages


felixmulder:
We have a multitude of errors that need to be ported to the [new scheme](http://scala-lang.org/blog/2016/10/14/dotty-errors.html) - you'll find these under "Error Messages" below. But first, a tutorial:
# HOWTO on creating new messages

Once you've identified an error you want to replace, you should decide what semantic information is needed in the message - i.e. the position, perhaps a tree or some types. For this example we're going to be implementing the error message for the following error:

``` scala
case class Foo // you're not allowed to have a case class without a single parameter list!
```

So we create a file with this and compile it using `./bin/dotc test.scala`. This gives us the error message:

```
case class Foo
           ^
           case class needs to have at least one parameter list
```

So now we search the repository for "case class needs to have" - I do this with `git grep -n "case class needs to have"`.

And now we find that in `src/dotty/tools/dotc/ast/Desugar.scala` we have the following line:

``` scala
ctx.error("case class needs to have at least one parameter list", cdef.pos)
```

Cool - now we know what to replace. So next step is to add a `case class` to `src/dotty/tools/dotc/reporting/diagnostic/messages.scala`. The case class that we add should extend `Message`:

``` scala
abstract class Message(val errorId: Int) {
  /** The `msg` contains the diagnostic message e.g:
    *
    * > expected: String
    * > found:    Int
    *
    * This message wil be placed underneath the position given by the enclosing
    * `MessageContainer`
    */
  def msg: String

  /** The kind of the error message is something like "Syntax" or "Type
    * Mismatch"
    */
  def kind: String

  /** The explanation should provide a detailed description of why the error
    * occurred and use examples from the user's own code to illustrate how to
    * avoid these errors.
    */
  def explanation: String
  ...
}
```

Now we name our case class something useful and give it a unique `errorId`:

``` scala
case class CaseClassMissingParamList() extends Message(4)
```

Since we can see that the call to `ctx.error` took the position of a `cdef: TypeDef`, we could perhaps use this in the constructor of our `Message`?

``` scala
case class CaseClassMissingParamList(cdef: untpd.TypeDef) extends Message(4)
```

It might also be that we're going to be manipulating `cdef` tree, as such we should pass a `Context` to get all the implicit operations:

``` scala
case class CaseClassMissingParamList(cdef: untpd.TypeDef)(implicit ctx: Context)
extends Message(4)
```

Alright! Now we can start adding things to our `Messsage`:

``` scala
[...]

PR #1601, opened 2016-10-16
Missing error messages: Parsers.scala: 2064

sebastianharko:
Tackles [1589](https://github.com/lampepfl/dotty/issues/1589). More specifically, it provides a detailed explanation for the error message in Parsers.scala:2064 : "early definitions are not supported; use trait parameters instead".

The example code in the explanation message is taken from [this presentation](http://files.meetup.com/18712511/Scala2016.pdf) (page 17).

Let me know whether the explanation is too long or needs improvements :-)


felixmulder:
Thanks for taking the time! LGTM

PR #1607, opened 2016-10-18
Fix #1605: don't inline methods that have errors

### LINKED ISSUE #1605: Inlining invalid method crashes compiler
With:

``` scala
object Test {
  def foo = inlineMe

  inline def inlineMe = 1 + x // where x is not defined anywhere
}
```

the compiler crashes.

felixmulder:
Review: @smarter

### DIFF
--- a/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/src/dotty/tools/dotc/typer/Inliner.scala
@@ -146,7 +146,7 @@ object Inliner {
               addAccessor(tree, methPart, targs, argss,
                   accessedType = methPart.tpe.widen,
                   rhs = (qual, tps, argss) => qual.appliedToTypes(tps).appliedToArgss(argss))
-           } else {
+            } else {
               // TODO: Handle references to non-public types.
               // This is quite tricky, as such types can appear anywhere, including as parts
               // of types of other things. For the moment we do nothing and complain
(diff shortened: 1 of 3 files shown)


DarkDimius:
nice title)

felixmulder:
It inlined it ;)

smarter:
I think I would put the check at the place where we call `makeInlineable` in `registerInlineInfo` instead of deep into `makeInlineable` itself.

felixmulder:
@smarter - done

src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -191,7 +191,9 @@ object Inliner {
           sym.updateAnnotation(LazyBodyAnnotation { _ =>
             implicit val ctx: Context = inlineCtx
             val tree1 = treeExpr(ctx)
```
odersky: Better use reporter.errorCount

src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -190,8 +190,7 @@ object Inliner {
           val inlineCtx = ctx
           sym.updateAnnotation(LazyBodyAnnotation { _ =>
             implicit val ctx: Context = inlineCtx
-            val tree1 = treeExpr(ctx)
-            makeInlineable(tree1)
+            ctx.withNoError(treeExpr(ctx))(makeInlineable)
```
felixmulder: @odersky - updated with your suggestion

odersky:
LGTM

PR #1608, opened 2016-10-18
Record syntactic information about modifiers

liufengyun:
This PR defines `trait Mod` to record source information about modifiers.

Note that `privateWithin` is already present in `Modifiers`, thus it's not repeated in Mod definitions. 

Review @odersky .

### DIFF
--- a/src/dotty/tools/dotc/ast/untpd.scala
+++ b/src/dotty/tools/dotc/ast/untpd.scala
@@ -94,9 +94,42 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
   class InfixOpBlock(leftOperand: Tree, rightOp: Tree) extends Block(leftOperand :: Nil, rightOp)
 
   // ----- Modifiers -----------------------------------------------------
+  /** Mod is intended to record syntactic information about modifiers, it's
+    * NOT a replacement of flagsets.
+    *
+    * For any query about semantic information, check `flags` instead.
+    */
+  sealed trait Mod extends Positioned
 
- /** Modifiers and annotations for definitions
-   *  @param flags          The set flags
+  object Mod {
+    case class Private() extends Mod
+
+    case class Protected() extends Mod
+
+    case class Val() extends Mod
+
+    case class Var() extends Mod
+
+    case class Implicit() extends Mod
+
+    case class Final() extends Mod
+
+    case class Sealed() extends Mod
+
+    case class Override() extends Mod
+
+    case class Abstract() extends Mod
+
+    case class Lazy() extends Mod
+
+    case class Inline() extends Mod
+
+    case class Type() extends Mod
+  }
+
+  /** Modifiers and annotations for definitions
+    *
+    *  @param flags          The set flags
    *  @param privateWithin  If a private or protected has is followed by a
    *                        qualifier [q], the name q, "" as a typename otherwise.
    *  @param annotations    The annotations preceding the modifiers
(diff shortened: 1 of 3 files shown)


src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1488,10 +1501,12 @@ object Parsers {
         mods
 
     private def addModifier(mods: Modifiers): Modifiers = {
-      val flag = flagOfToken(in.token)
+      val tok = in.token
+      val flag = flagOfToken(tok)
+      val mod = atPos(in.offset) { in.nextToken(); modOfToken(tok) }
```
odersky: shorter:

    val mod = atPos(in.skipToken()) { modOfToken(tok) }

Same for similar places below.

src/dotty/tools/dotc/ast/untpd.scala
```diff
@@ -94,17 +94,51 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
...
+    * NOT a replacement of flagsets.
+    *
+    * For any query about semantic information, check `flags` instead.
+    */
+  sealed trait Mod extends Positioned
 
- /** Modifiers and annotations for definitions
-   *  @param flags          The set flags
+  object Mod {
+    case class Private() extends Mod
```
odersky: My hunch is that this can be done slightly more elegantly as follows:

 - Have `Mod` take a `flags` field and have every case class set it. E.g.

           case class Protected() extends Mod(Flags.Protected)

Have a new method in Modifiers:

     def | (mod: Mod) =
        withFlags(mod.flags).withAddedMod(mod)
        
Then, only use `|` instead of setting flags and mods separately. I believe that would make redundant much of the current code. WDYT?

liufengyun: That's a good idea, I'll do as you suggested, thanks a lot.

liufengyun: @odersky I just checked the parser code, whether a semantic flag gets added depends on some compatibility check (see `addFlag`). But the syntactic `Mod` should be always added.

It's also possible to refactor the flag compatibility check to `Modifiers`, but the error reporting depends on compiler state (`lastErrorOffset`), make the refactoring difficult.

odersky: @liufengyun I think this can be accommodated in addFlag:

      if (compatible(mods.flags, flag)) mods | mod
      else mods.withAddedMod(mod)

liufengyun: @odersky I just pushed a commit addressing the review feedback. Note that `addFlag` is also used for non-syntactic flags(such as `trait`, `method`), so I added another `addMod` method.

And it turns out that we always need to check the added flags, thus ` def | (mod: Mod) = ..` is not useful.

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1819,9 +1826,13 @@ object Parsers {
      */
     def defOrDcl(start: Int, mods: Modifiers): Tree = in.token match {
       case VAL =>
-        patDefOrDcl(start, posMods(start, mods), in.getDocComment(start))
+        val mod = atPos(in.skipToken()) { Mod.Val() }
+        val modPos = atPos(start) { mods.withAddedMod(mod) }
```
odersky: Two comments: 

 - Why the name `modPos`? It reads like "the positon of mod" to me. Maybe use `mods1` instead,  or inline the expression.

- I don't think you need an `atPos(start)` here. The position of the modifiers is set automatically to the union of `mods` and `mod`; so you don't need to set the position manually.

Otherwise LGTM.

odersky:
Good to have this in!

PR #1611, opened 2016-10-18
Improved error messages in Desugar.scala

ShaneDelmore:


src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,66 @@ object messages {
...
+  case class ImplicitCaseClass(cdef: untpd.TypeDef)(implicit ctx: Context)
+    extends Message(11) {
+    val kind = "Syntax"
+
+    val msg = hl"""|A ${"case class"} may not be defined as ${"implicit"}"""
+
+    val explanation =
+      hl"""|implicit classes may not be case classes. Instead use a plain class:
+           |
+           |implicit class ${cdef.name}...""".stripMargin
```
ShaneDelmore: Should I try to create the class definition here using args from the AST?  Or is that excessive/difficult?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,66 @@ object messages {
...
+    val msg = hl"""|An ${"implicit class"} may not be top-level"""
+
+    val explanation =
+    hl"""|implicit class are not allowed to be defined at the top-level.
+         |
+         |Declare ${cdef.name} inside of an ${"object"} or ${"package object"}
+         |then import the members of the object at the use site if needed, for example:
+         |
+         |package object Implicits {
+         |  implicit class ${cdef.name}...
```
ShaneDelmore: I would have liked to add the args here but did not see how to grab them.  Can I use scala.meta for this?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,66 @@ object messages {
...
+         |}
+         |
+         |import Implicits.${cdef.name}""".stripMargin
```
felixmulder: Nice! I like the usage of the programmers own code here - makes for something that's easy to grok.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Perhaps just give a quick reason as to why - i.e. "There may not be any method, member or object in scope with the same name as the implicit class and a case class automatically gets a companion object with the same name."

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,66 @@ object messages {
...
+    val kind = "Syntax"
+
+    val msg = hl"""|An ${"object"} must not have a ${"self type"}"""
+
+    val explanation =
+      hl"""|objects ${mdef.name} must not have a ${"self type"}:
+           |
+           |Consider these alternative solutions:
+           |  Extend the object with the desired type
+           |  Use a trait or a class instead of an object""".stripMargin
```
felixmulder: As this is how you'd solve the problem, I think it is totally reasonable advice :)

When you say "Extend the object with the desired type", are you implying `trait T; object O extends T` and placing the self type in `T`?

If so, perhaps phrase it as "Let the object extend a trait containing the self type"

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Currently, using scala.meta for this is not possible. It might be in the future!

Hmm, I can't verify this right now but something like:

1. check if `cdef.rhs` is a `Template`
2. if so get the constr (a `DefDef`)

Check what `.show` reveals on the constructor. Should be something like: `def <init>(param1: X, param2: Y)`

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: If you want to, I mean I'd at least shorten it to not include the body of the definition (i.e. `rhs`). I would say that a similar method to what was described bellow would work for this too, and as such you can factor it out to a common method.

It is also possible to have a look at `dottydoc/src/.../DocASTPhase.scala` to get a sense of how the new doctool does it. Bear in mind that the doctool is working with a typed tree..

felixmulder:
@ShaneDelmore - wait, scratch that. You're in the `classDef` method, and on `Desugar.scala:253` you already have basically all you'll need at least for the implicit class one.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: That's great, I did not know the reasoning behind this until you explained it.

ShaneDelmore:
I almost have what I wanted.  Say I have
```
implicit case class Foo(bar: Int) {
  def baz = bar.toString
}
```

What I will get out of the AST is  
```
implicit case class Foo(private[this] val bar: Int) extends {
  def baz = bar.toString
}
```

I am worried the difference may confuse the users it is intended to help.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,66 @@ object messages {
...
+
+    val msg = hl"""|An ${"implicit class"} may not be top-level"""
+
+    val explanation =
+    hl"""|implicit class are not allowed to be defined at the top-level.
+         |
+         |Declare ${cdef.name} inside of an ${"object"} or ${"package object"}
+         |then import the members of the object at the use site if needed, for example:
+         |
+         |package object Implicits {
```
smarter: I wouldn't recommend (or even mention) package objects, they have tricky semantics and are not commonly what you want to do.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,86 @@ object messages {
         |$code2
         |""".stripMargin
   }
+
+  val implicitClassRestrictionsText =
+    """For a full list of restrictions on implicit classes visit
+      |http://docs.scala-lang.org/overviews/core/implicit-classes.html""".stripMargin
```
ShaneDelmore: Is there a highlight available for URLs?  The highlighting on this treats http: as plain text and //restOfTheUrl as a comment.

Also, do highlighted sections compose?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,86 @@ object messages {
...
+
+  case class TopLevelImplicitClass(cdef: untpd.TypeDef)(implicit ctx: Context)
+    extends Message(10) {
+    val kind = "Syntax"
+
+    val msg = hl"""|An ${"implicit class"} may not be top-level"""
+
+    val explanation = {
+      val TypeDef(name, impl @ Template(constr0, parents, self, _)) = cdef
```
ShaneDelmore: This is duplicated from Desugar.scala.  Is it preferred to duplicate it here to keep the coupling looser or should I instead add more parameters to the case class and pass along explicitely what I need, name, impl, constr0?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,91 @@ object messages {
...
+           |To resolve the conflict declare ${cdef.name} inside of an ${"object"} then import the class
+           |from the object at the use site if needed, for example:
+           |
+           |object Implicits {
+           |  implicit class ${cdef.name}($exampleArgs)$exampleBody
+           |}
+           |
+           |import Implicits.${cdef.name}""".stripMargin
+    }
+  }
```
ShaneDelmore: Example output:
```
-- [E010] Syntax Error: test.scala -------------------------------------------------------------------------------------
2 |implicit class Foo(n: Int)
  |^^^^^^^^^^^^^^^^^^^^^^^^^^
  |An implicit class may not be top-level

Explanation
===========
There may not be any method, member or object in scope with the same name as the
implicit class and a case class automatically gets a companion object with the same name
created by the compiler which would cause a naming conflict if it were allowed.

For a full list of restrictions on implicit classes visit
http://docs.scala-lang.org/overviews/core/implicit-classes.html

To resolve the conflict declare Foo inside of an object then import the class
from the object at the use site if needed, for example:

object Implicits {
  implicit class Foo(val n: Int)
}

import Implicits.Foo
```
```
-- [E010] Syntax Error: test.scala -------------------------------------------------------------------------------------
4 |implicit class Foo(bar: Int, baz: String) {
  |^
  |An implicit class may not be top-level
5 |  def baz = bar.toString
6 |
7 |  def qux = baz.take(5)
8 |}

Explanation
===========
There may not be any method, member or object in scope with the same name as the
implicit class and a case class automatically gets a companion object with the same name
created by the compiler which would cause a naming conflict if it were allowed.

For a full list of restrictions on implicit classes visit
http://docs.scala-lang.org/overviews/core/implicit-classes.html

To resolve the conflict declare Foo inside of an object then import the class
from the object at the use site if needed, for example:

object Implicits {
  implicit class Foo(val bar: Int, val baz: String) {
    def baz = bar.toString

    def qux = baz.take(5)
  }
}

import Implicits.Foo
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,91 @@ object messages {
...
+    val kind = "Syntax"
+
+    val msg = hl"""|A ${"case class"} may not be defined as ${"implicit"}"""
+
+    val explanation =
+      hl"""|implicit classes may not be case classes. Instead use a plain class:
+           |  example: implicit class ${cdef.name}...
+           |
+           |$implicitClassRestrictionsText""".stripMargin
+  }
```
ShaneDelmore:
```
-- [E011] Syntax Error: test.scala -------------------------------------------------------------------------------------
14 |  implicit case class Foo(bar: Int, baz: String)
   |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |  A case class may not be defined as implicit

Explanation
===========
implicit classes may not be case classes. Instead use a plain class:
  example: implicit class Foo...

For a full list of restrictions on implicit classes visit
http://docs.scala-lang.org/overviews/core/implicit-classes.html
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,91 @@ object messages {
...
+      val ModuleDef(name, tmpl) = mdef
+      val ValDef(_, selfTpt, _) = tmpl.self
+      hl"""|objects must not have a ${"self type"}:
+           |
+           |Consider these alternative solutions:
+           |  - Create a trait or a class instead of an object
+           |  - Let the object extend a trait containing the self type:
+           |      example: object $name extends ${selfTpt.show}""".stripMargin
+    }
+  }
```
ShaneDelmore:
```
-- [E012] Syntax Error: test.scala -------------------------------------------------------------------------------------
13 |  self: Foo =>
   |  ^^^^^^^^^
   |  objects must not have a self type

Explanation
===========
objects must not have a self type:

Consider these alternative solutions:
  - Create a trait or a class instead of an object
  - Let the object extend a trait containing the self type:
      example: object Test extends Foo
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: All interpolated strings in `hl"..."` get syntax highlighting. If you want to interpolate `implicitClassRestrictions` in you highlighted block I'd wrap it in a `NoColor(...)` `Highlight`. Already "colored" strings do not get interpolated by the interpolator.

See: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/printing/Highlighting.scala#L55

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Good question, I think there's a risk that more people will need to print the class def and as such perhaps we should factor this out into a method call that takes the `cdef` - a small bit of duplication here is fine in order for it to be easy to use for new contributors IMHO.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +319,91 @@ object messages {
...
+      val exampleArgs = constr0.vparamss(0).map(_.withMods(Modifiers()).show).mkString(", ")
+      val exampleBodyDefs = impl.body.collect {
+        case t if !t.isEmpty => t.show
+      }
+      val exampleBody = if (exampleBodyDefs.isEmpty) ""
+      else exampleBodyDefs.mkString(
+        start = " {\n    ",
+        sep = "\n\n    ",
+        end = "\n  }"
+      )
```
felixmulder: I would actually prefer this to have an else clause that is something like ` {\n  ...\n  }`.

My motivation for this is if we have an implicit class with a lot of members, they will all get printed and bloat the message. For copy-pastability (is that even a word?), I think it's fine if we leave out the complete method body.

felixmulder:
Super impressive PR, keep up the good work :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: Sounds good. I will try to make the suggested updates this evening.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
smarter: I think it would be nice to make it more explicit that the `import Implicits.Foo` is a at the use site, for example by writing:
```scala
object Implicits {
  implicit class Foo(val bar: Int, val baz: String){
 ...
 }
}

...

// At the use site:
import Implicits.Foo
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: Now that the explanation body has been simplified there isn't really much here worth extracting.

PR #1613, opened 2016-10-19
Add error message - Parsers.scala:1180

jyotman94:
Part of #1589.
Explanation for the error on line Parsers.scala:1180.
Adds the sematic object for the error `illiegal start of simple expression`.
@felixmulder kindly review.

### DIFF
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -482,4 +482,22 @@ object messages {
         |but an uninitialized var definition
         """.stripMargin
   }
+
+  case class IllegalStartSimpleExpr(illegalToken: String)(implicit ctx: Context) extends Message(17) {
+    val kind = "Syntax"
+    val msg = "illegal start of simple expression"
+    val explanation = {
+      hl"""|An expression yields a value. In the case of the simple expression, this error
+           |commonly occurs when there's a missing parenthesis or brace. The reason being
+           |that a simple expression is one of the following:
+           |
+           |- Block
+           |- Expression in parenthesis
+           |- Identifier
+           |- Object creation
+           |- Literal
+           |
+           |which cannot start with ${Red(illegalToken)}.""".stripMargin
+    }
+  }
 }
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1177,7 +1177,7 @@ object Parsers {
         case _ =>
           if (isLiteral) literal()
           else {
-            syntaxErrorOrIncomplete("illegal start of simple expression")
+            syntaxErrorOrIncomplete(IllegalStartSimpleExpr(tokenString(in.token)))
             errorTermTree
           }
       }


src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,13 @@ object messages {
         |$code2
         |""".stripMargin
   }
+
+  case class IllegalExpressionStart()(implicit ctx: Context) extends Message(14) {
```
felixmulder: I would call this something like: `IllegalStartSimpleExpr` since it is simple expressions and not all valid expressions. I'm assuming you chose 14 to coordinate with Shane's PR? :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -318,4 +318,13 @@ object messages {
...
+    val kind = "Syntax"
+    val msg = "illegal start of simple expression"
+    val explanation = {
+      hl"""|The expression you've provided is not valid because
+           |an expression in Scala always returns a value of some type""".stripMargin
```
felixmulder: If the explanation is printed, the user is most likely a beginner. In this case we need to educate them by providing a hint as to how they can make their program valid. If I don't know that expressions are things which return values then I, the user, need to be given some examples.

This error also occurs frequently when a user has forgotten a parenthesis somewhere which makes it sort of tricky to  give a 100% concrete solution in this case. Perhaps in the future we could have some machinery to discover this.

I would try to write something similar to:

```
An expression yields a value. In the case of the simple expression, this error
commonly occurs when there's a missing parenthesis or brace. The reason being
that a simple expression is one of the following:

- Block
- Expression in parenthesis
- Identifier
- Object creation
- Literal

Which cannot start with a `}` or a `)`.
```

felixmulder:
Also a quick note on commit messages, we prefer the form:

```
Add explanation for "illegal start of simple expression"
```

instead of:

```
Added explanation for "illegal start of simple expression"
```

:)

felixmulder:
Otherwise things are looking great :+1:

jyotman94:
Thanks @felixmulder  all points noted :) I'll try to give a better explanation using the one you've suggested. Do you want me to include code samples also to explain both - probable wrong ways and the correct ways? But there could be many examples for that.

Also yes I kept the message number 14 to coordinate with Shane's PR.

felixmulder:
@jyotman94 - hmm, yeah I don't know, that's really subjective right? IMHO this explanation probably doesn't need any code examples - but you know, have fun with it and experiment :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: It won't produce an exception if the token is defined no - so it shouldn't be a problem. Good thinking on including user code :+1:

jyotman94:
@felixmulder I've made the changes. Kindly review.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -400,4 +400,22 @@ object messages {
            |  ((${nestedRepresentation}))""".stripMargin
     }
   }
+
+  case class IllegalStartSimpleExpr(illegalString: String)(implicit ctx: Context) extends Message(14) {
```
felixmulder: I would rename `illegalString` to `illegalToken` :)

felixmulder:
So when I said the the git commit should be in "imperative form" I meant do a rebase and reword the commit. If you didn't know how, it can be done like this:

```
$ git remote add upstream git@github.com:lampepfl/dotty.git
$ git fetch upstream
$ git rebase -i upstream/master
```

Then choose to reword your commit message and call it something like: `Add explanation for "illegal start of simple expression"`. Makes sense? After that everything looks good to me, great job!

jyotman94:
Sorry, actually I got confused whether you referred to my commit message or the pull request title earlier so I had just renamed the pull request title. I've made the new changes. Thank you for being so patient :)

felixmulder:
No problem - thank you for putting in all the hard work :)

Once this passes CI, I'll merge it - cheers!

felixmulder:
Hey @jyotman94 - could you rebase so I can merge? :) (Don't forget to check that your `errorId` is correct)

jyotman94:
@felixmulder done!

felixmulder:
Great! Thanks a lot for the PR :)

PR #1615, opened 2016-10-20
Add error messages - Parsers.scala:626 and Parsers.scala:1492 

sebastianharko:
Error messages for Parsers.scala:626 and Parsers.scala:1492.

**Note**: I may have to bump the error message ids - depending on whether the other [PR](https://github.com/lampepfl/dotty/pull/1613) gets merged before this one.

### DIFF
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -400,4 +400,51 @@ object messages {
            |  ((${nestedRepresentation}))""".stripMargin
     }
   }
+
+  case class RepeatedModifier(modifier: String)(implicit ctx:Context) extends Message(14) {
+    val kind = "Syntax"
+
+    val msg = hl"""repeated modifier $modifier"""
+
+    val code1 = hl"""private private val Origin = Point(0, 0)"""
+
+    val code2 = hl"""private final val Origin = Point(0, 0)"""
+
+    val explanation =
+      hl"""This happens when you accidentally specify the same modifier twice.
+           |
+           |Example:
+           |
+           |$code1
+           |
+           |instead of
+           |
+           |$code2
+           |
+           |""".stripMargin
+  }
+
+  case class InterpolatedStringError()(implicit ctx:Context) extends Message(15) {
+    val kind = "Syntax"
+
+    val msg = "error in interpolated string: identifier or block expected"
+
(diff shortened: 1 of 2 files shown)


felixmulder:
So the test that is failing is: `./tests/neg/inlinevals.scala`

You can try compiling it with `run tests/neg/inlinevals.scala` from sbt. I think the issue is that the position has changed from when it was using `syntaxError`. That method uses `in.offset` as the pos IIRC

sebastianharko:
@felixmulder thanks for the hint ! I fixed it :-)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -400,4 +400,51 @@ object messages {
...
+           |
+           |Example:
+           |
+           |$code1
+           |
+           |instead of
+           |
+           |$code2
+           |
+           |""".stripMargin
```
felixmulder: @liufengyun has been working on modifiers with positions, and I'm thinking maybe we could enhance this further to say make the following recommendation:

```
This happens when you accidentally specify the same modifier twice:

private private lazy val x = ...

instead of:

private lazy val x = ...
```

Where this would be the user's code without the duplicate modifier. What do you think @liufengyun - can this be done with the new modifier objects?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -400,4 +400,51 @@ object messages {
...
+  }
+
+  case class InterpolatedStringError()(implicit ctx:Context) extends Message(15) {
+    val kind = "Syntax"
+
+    val msg = "error in interpolated string: identifier or block expected"
+
+    val code1 = "s\"$new Point(0,0).toString\""
+
+    val code2 = "s\"${new Point(0,0).toString}\""
```
felixmulder: No need for `.toString` in a string interpolation, it gets called automagically :)

felixmulder:
Hmm, where the CI at?

felixmulder:
Hey @adriaanm and @SethTisue - could you check if something's wrong? Pretty please :)

adriaanm:
Looks like you're getting the regular sbt launcher and not xsbt.

adriaanm:
There's also a DNS DDOS going on.

adriaanm:
sbt-extras bug should be resolved (https://github.com/scala/scala-jenkins-infra/pull/194)

PR #1616, opened 2016-10-20
Update error message at Parsers.scala:1901

ljdelight:
This one is about a 'missing return type' when we can't infer a type. I used tests/neg/i871.scala to verify the change

### DIFF
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -275,10 +275,10 @@ object messages {
 
     val explanation = ""
   }
- 
+
   case class EarlyDefinitionsNotSupported()(implicit ctx:Context) extends Message(9) {
     val kind = "Syntax"
-   
+
     val msg = "early definitions are not supported; use trait parameters instead"
 
     val code1 =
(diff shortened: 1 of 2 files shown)


ljdelight:
FYI I'm behind at least one PR that will conflict. Noting this since I'm not familiar with the capabilities of the CI

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -400,4 +400,15 @@ object messages {
...
+  case class MissingReturnType()(implicit ctx:Context) extends Message(14) {
+    val kind = "Syntax"
+    val msg = "missing return type"
+    val explanation =
+      hl"""The return type is missing and needs to be clarified. For example, `area` returns a ${"Double"}:
+        |
+        |trait Shape {
+        |  def area: Double
+        |}""".stripMargin
+  }
```
felixmulder: So first off, you're correctly stating that if a trait method does not have a return type it needs to be clarified. But why is this? It's an abstract definition; we need to know what it's type is when we implement it somewhere else.

I also have a different nitpick, so the error messages obey a flag called `-pagewidth X`, where `X` is the max number of columns that should be printed in the terminal. As such, I would suggest that you split line 408 to be no more than 80 characters (i.e. newline after `` `area` ``).

Good job otherwise, and thanks for cleaning up the trailing whitespace :heart:

felixmulder:
Oh and also, second nitpick! 

Make your commit message subject line a bit shorter, put the rest in the body - example:

```
Update error message for missing return type

This one is about...
# Please enter the commit message for your changes...
```

Cheers!

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ljdelight: Thanks for the comments. How does this read?
`"""An abstract definition must have a return type. For example:
        |
        |trait Shape {
        |  def area: Double // abstract definition returning a ${"Double"}
        |}"""`

felixmulder: Yes that looks fine, make the change and this is approved for merge :)

felixmulder:
LGTM - a rebase and an update on the `errorId` and we're good to go :)

PR #1617, opened 2016-10-20
Introduce scalaShadowing package

odersky:
The `scalaShadowing` package is used to safely modify classes and
objects in scala so that they can be used from dotty. They will
be visible as members of the `scala` package, replacing any objects
or classes with the same name. But their binary artifacts are
in `scalaShadowing` so they don't clash with the same-named `scala`
members at runtime.

Review by @nicolasstucki

### DIFF
--- a/src/dotty/Pair.scala
+++ b/src/dotty/Pair.scala
@@ -1,5 +1,0 @@
-package dotty
-
-class Pair[T, U](x: T, y: U) {
-
-}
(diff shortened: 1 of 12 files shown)


nicolasstucki:
LGTM and it fixed the issue in #1479.

nicolasstucki:
I assume that we could also shadow `scala.Predef` and merge it with `dotty.Predef`.

PR #1621, opened 2016-10-23
Add error message - Comments.scala:128

thiagoandrade6:
This commit adds the semantic object for the ```proper definition not found``` error.
It is part of the [https://github.com/lampepfl/dotty/issues/1589](https://github.com/lampepfl/dotty/issues/1589)

### DIFF
--- a/src/dotty/tools/dotc/core/Comments.scala
+++ b/src/dotty/tools/dotc/core/Comments.scala
@@ -9,6 +9,7 @@ import util.Positions._
 import util.CommentParsing._
 import util.Property.Key
 import parsing.Parsers.Parser
+import reporting.diagnostic.messages.ProperDefinitionNotFound
 
 object Comments {
   val ContextDoc = new Key[ContextDocstrings]
@@ -125,7 +126,7 @@ object Comments {
               val newName = (tree.name.show + "$" + codePos + "$doc").toTermName
               untpd.DefDef(newName, tree.tparams, tree.vparamss, tree.tpt, tree.rhs)
             case _ =>
-              ctx.error("proper definition was not found in `@usecase`", codePos)
+              ctx.error(ProperDefinitionNotFound(), codePos)
               tree
           }
         }
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -543,5 +543,42 @@ object messages {
            |
            |""".stripMargin
   }
- 
+
+  case class ProperDefinitionNotFound()(implicit ctx: Context) extends Message(20) {
+    val kind = "Definition Not Found"
+    val msg = hl"""|Proper definition was not found in ${"@usecase"}"""
+
+    val noUsecase = "def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That]): That"
+
+    val usecase =
+      """|/** Map from List[A] => List[B]
+        |  *
+        |  * @usecase def map[B](f: A => B): List[B]
+        |  */
+        |def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That]): That""".stripMargin
+
+    val explanation = {
+      hl"""|${"@usecase"} are only supported for ${"def"}s. They exist because with Scala's
+           |advanced type-system, we sometimes end up with seemingly scary signatures.
+           |
+           |Let's see an example using the `map`function:
+           |
+           |${"List(1, 2, 3).map(2 * _) // res: List(2, 4, 6)"}
+           |
+           |It's very straight forward to understand and use, but has such a scary signature:
+           |
+           |$noUsecase
+           |
+           |In order to mitigate this and ease the usage of such functions we have the ${"@usecase"}
+           |annotation for docstrings. Which can be used like this:
+           |
+           |$usecase
+           |
+           |Now when creating the docs, the method signature is substituted by the
+           |${"@usecase"} with a reader-friendly version. The compiler makes sure that it is valid.
+           |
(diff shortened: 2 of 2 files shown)


thiagoandrade6:
Ok, this is my first PR over here, so go easy with me :)
I have some doubts related to the ```explanation``` of this error. I particularly believe we should have something there, but I have no idea. I would appreciate any help.

src/dotty/tools/dotc/core/Comments.scala
```diff
@@ -2,8 +2,12 @@ package dotty.tools
...
 package core
 
-import ast.{ untpd, tpd }
-import Decorators._, Symbols._, Contexts._, Flags.EmptyFlags
+import ast.{tpd, untpd}
+import Decorators._
+import Symbols._
+import Contexts._
+import Flags.EmptyFlags
+import dotty.tools.dotc.reporting.diagnostic.messages.ProperDefinitionNotFound
```
felixmulder: I think a lot of these changes can be left out, basically the only addition here should be something like:

```scala
// you should be able to omit most of the path and get something like:
import reporting.diagnostic.messages.ProperDefinitionNotFound
// or even just:
import reporting.diagnostic.messages._
```

whichever floats your goat :)

src/dotty/tools/dotc/core/Comments.scala
```diff
@@ -125,7 +129,7 @@ object Comments {
               val newName = (tree.name.show + "$" + codePos + "$doc").toTermName
               untpd.DefDef(newName, tree.tparams, tree.vparamss, tree.tpt, tree.rhs)
             case _ =>
-              ctx.error("proper definition was not found in `@usecase`", codePos)
+              ctx.error(ProperDefinitionNotFound(), codePos)
```
felixmulder: To get a good sense of what's going on here for the error message's explanation there are a couple of things you ( or the user) need(s) to know:

1. What are `@usecase`?
2. Why do we care that we're getting an `untpd.DefDef` here?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -500,4 +500,11 @@ object messages {
            |which cannot start with ${Red(illegalToken)}.""".stripMargin
     }
   }
+
+  case class ProperDefinitionNotFound()(implicit ctx: Context) extends Message(18) {
+    val kind = "Definition Not Found"
+    val msg = hl"""|A proper definition was not found in ${"@usecase"}"""
+    val explanation = ""
+  }
+
```
felixmulder: Usecases are a workaround for ugly signatures that the user of a library does not need to know about. For instance:

```
trait List[+A] {
  def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That] : That = ???
}
```

now the users of the list are going "dafuq :confused:". Basically the implicit builder will effectively build a collection of the expected type - but the user just assumes that's what happens when you map - and doesn't need to know that.

So if we do this:

```scala
/** Map from List[A] => List[B]
  *
  * @usecase def map[B](f: A => B): List[B]
  */
def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That]): That = ???
```

when building the documentation, the doc tool (which is a sort of compiler) will switch out the definition of `map` for the one defined in the `@usecase`. As such, the user reading the docs, won't see the ugly definition and be blissfully unaware :)

Because of this, the error occurs when the user has not written a `def` but something else (like `val map: List[B]` or w/e).

So perhaps the explanation can contain some of this information?

thiagoandrade6: Great explanation, thanks for your time to do that, I will be working on it and should push a fix later :)

thiagoandrade6:
@felixmulder I have updated the code based on your explanation, take a look and let me know if you agree. Also, there is another PR with approved state with the same number(18), perhaps a merge should occur.

felixmulder:
You're right, update yours to 19? I'll have a look at your changes first thing in the morning :)

thiagoandrade6:
yeah, I have rebased with master, merged and now it's 19!
Sorry the delay, but besides timezone, I still have to work 8 hours a day:( but we will get used to it and I will help as much as I can :)

felixmulder:
Don't worry about the time @thiagoandrade6 - there's no rush :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +512,25 @@ object messages {
...
+           |${"def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That] : That"}
+           |
+           |Could be written using ${"@usecase"} as follow:
+           |
+           |${"@usecase def map[B](f: A => B): List[B]"}
+           |
+           |This way you can hide the ugly definition by providing an easy readable version.
+           |
+           |""".stripMargin
+    }
```
thiagoandrade6: Well done again :) Thank you!

thiagoandrade6:
@felixmulder Take a look in the code whenever you can. Basically I got your message but I have rewritten some parts of that. I believe anyone who got such error, will now have get rid of that with a very good explanation :)

felixmulder:
@thiagoandrade6 - thanks for your patience, I'll merge as soon as the CI passes :)

thiagoandrade6:
@felixmulder thanks for make my life easier, helping me to understand more about how the compile works. I have other several questions, but I will read more the code, docs, talks, whatever exist and asking you on demand. For now let's move on to the next PR :)

felixmulder:
No worries - for questions not regarding a specific PR, I'll refer you to our [gitter-channel](http://gitter.im/lampepfl/dotty) :) happy coding!

Issue #1612, opened 2016-10-18
DuplicateBind error triggered by same variable name in alternative pattern


ShaneDelmore:
I believe the issue is in Typer.transform.  Here is an example snippet and output:

```
object Test {
  def g(p:(Int,Int)) = p match {
    case (10,n) | (n,10) => println(n) 
    case _ => println("nope")
  }
}
```

On compile this error is returned:

```
-- [E005] Naming Error: test.scala -------------------------------------------------------------------------------------
9 |    case (10,n) | (n,10) => println(n)
  |                   ^
  |                   duplicate pattern variable: `n`

longer explanation available when compiling with `-explain`
```

ShaneDelmore:
Tagging @DarkDimius and @liufengyun at @felixmulder request.

odersky:
There's clearly an error here, but not the one that's flagged. scalac says:

error: illegal variable in pattern alternative

ShaneDelmore:
@odersky This is actually how I found this bug, I was attempting to improve the error message for "illegal variable in pattern alternative" and found I could not trigger the error message in Dotty as I could in Scala.

odersky:
@ShaneDelmore You want to give it a try to add the error message? I believe it should be in Typer. The thing to watch out for are Bind nodes that bind a name other than `_` inside Alternative nodes. 
Maybe add a Mode bit that we are inside an alternative that gets tested in the Bind node?

ShaneDelmore:
I will take a look at it this weekend and see if I can make any headway on it.

ShaneDelmore:
I started looking into this bug last weekend and think I should be able to come up with a fix.  Besides the error message, is there any reason not to just allow multiple variable bindings if they exist in every alternative with the same name and same type?

felixmulder:
This sounds like a judgement call for @odersky, the spec doesn't specifically prohibit this. It might make the desugaring inside `PatternMatcher` harder - unclear.

odersky:
Duplicate of #3332.

PR #1623, opened 2016-10-25
#1589: Improve error message for WrongNumberOfArgs

ShaneDelmore:


src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,38 @@ object messages {
...
+             |  If you need to hold more types in a list then you need to combine them
+             |  into another data type that can contain the number of types you need,
+             |  In this example one solution would be to use a Tuple:
+             |  val tuple2: Tuple2[Int, String = (1, "one)
+             |    List[(Int, String)] = List(tuple2)""".stripMargin
+      else
+        hl"""|You have not supplied enough type parameters
+             |  If you specify one type parameter then you need to specify every type parameter.""".stripMargin
+    }
+  }
```
ShaneDelmore: Example message for too many type args
```
-- [E018] Syntax Error: test.scala -------------------------------------------------------------------------------------
16 |  var bad = List[Int, String]
   |            ^^^^^^^^^^^^^^^^^
   |Too many type arguments for ([scala$collection$immutable$List$apply$$A] => (xs: A*)scala.collection.immutable.List[A])(List.apply)
   |expected: [A]
   |actual:   [Int, String]

Explanation
===========
You have supplied too many type parameters

For example List takes a single type parameter (List[A])
  If you need to hold more types in a list then you need to combine them
  into another data type that can contain the number of types you need,
  In this example one solution would be to use a Tuple:
  val tuple2: Tuple2[Int, String = (1, "one)
    List[(Int, String)] = List(tuple2)
```

ShaneDelmore: Example message for not enough type args
```
-- [E018] Syntax Error: test.scala -------------------------------------------------------------------------------------
17 |  var bad2 = Map[(Int, String)]
   |             ^^^^^^^^^^^^^^^^^^
   |Not enough type arguments for ([scala$collection$generic$GenMapFactory$apply$$A, scala$collection$generic$GenMapFactory$apply$$B] =>
   |  (elems: (A, B)*)scala.collection.immutable.Map[A, B]
   |)(Map.apply)
   |expected: [A, B]
   |actual:   [Tuple2[Int, String]]

Explanation
===========
You have not supplied enough type parameters
  If you specify one type parameter then you need to specify every type parameter.
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
         |}""".stripMargin
   }
 
+  case class WrongNumberOfArgs(fntpe: Type, argKind: String, expectedArgs: List[TypeParamInfo], actual: List[untpd.Tree])(implicit ctx: Context)
+    extends Message(18) {
+    val kind = "Syntax"
+    val expectedCount = expectedArgs.length
+    val actualCount = actual.length
+    val msgPrefix = if (actualCount > expectedCount) "Too many" else "Not enough"
```
felixmulder: I would wrap this in `NoColor` so that it's not highlighted in the `msg`.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+
+    val expectedArgString = (fntpe.widen match {
+      case pt: MethodOrPoly => pt //ensure we return a type that will have useful typeParms
+      case _ => fntpe
+    }).typeParams.map(_.paramName.show.split("\\$").last).mkString("[", ", ", "]")
```
felixmulder: this match expression is the same as just taking: `fntpe.widen.typeParams`

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+
+    val actualArgString = actual.map(_.show.split("\\.").last).mkString("[", ", ", "]")
+
+    val msg =
+      hl"""|$msgPrefix ${argKind} arguments for $fntpe
```
felixmulder: Instead of `$fntpe`, I would do: `${fntpe.termSymbol.showFullName}` which will give you something like: `scala.collection.immutable.List.apply` which you can then add the correct parameter list to (if you want).

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+    val explanation = {
+      if (actualCount > expectedCount)
+        hl"""|You have supplied too many type parameters
+             |
+             |For example List takes a single type parameter (List[A])
+             |  If you need to hold more types in a list then you need to combine them
+             |  into another data type that can contain the number of types you need,
+             |  In this example one solution would be to use a Tuple:
+             |  val tuple2: Tuple2[Int, String = (1, "one)
+             |    List[(Int, String)] = List(tuple2)""".stripMargin
```
felixmulder: Maybe kill the left padding on these lines and add a newline before the code?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+
+    val explanation = {
+      if (actualCount > expectedCount)
+        hl"""|You have supplied too many type parameters
+             |
+             |For example List takes a single type parameter (List[A])
+             |  If you need to hold more types in a list then you need to combine them
+             |  into another data type that can contain the number of types you need,
+             |  In this example one solution would be to use a Tuple:
+             |  val tuple2: Tuple2[Int, String = (1, "one)
```
felixmulder: missing `]` and `"`?

Also perhaps add `${"..."}` around the code to get highlighting, or just move it to variable and interpolate that :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
         |}""".stripMargin
   }
 
+  case class WrongNumberOfArgs(fntpe: Type, argKind: String, expectedArgs: List[TypeParamInfo], actual: List[untpd.Tree])(implicit ctx: Context)
+    extends Message(18) {
```
felixmulder: This will be number 19 if merged first :)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -547,4 +547,30 @@ object messages {
...
+           |    }
+           |
+           |  could be implemented by moving each alternative into a separate case:
+           |    def g(pair: (Int,Int)): Int = pair match {
+           |      case (1, n) => n
+           |      case (n, 1) => n
+           |      case _ => 0
+           |    }
+           |
+           |""".stripMargin
```
felixmulder: That's fine :) Also, if you want - you can take a stab at fixing #1612 as a more challenging task!

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+           |      case (1, n) | (n, 1) => n
+           |      case _ => 0
+           |    }
+           |
+           |  could be implemented by moving each alternative into a separate case:
+           |    def g(pair: (Int,Int)): Int = pair match {
+           |      case (1, n) => n
+           |      case (n, 1) => n
+           |      case _ => 0
+           |    }
```
felixmulder: Interpolate the code in this guy too? We love syntax highlighting :unicorn: :rainbow:

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
smarter: I would just write `val tuple2 = (1, "one")`, even if you wanted to write the type explicitly you would write `(Int, String)`, not `Tuple2[Int, String]`

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
smarter: As much as possible, we should avoid playing with regexps on names in error messages, this way lies madness. If there are issues with the way we print names, they should be fixed in a common place, not worked around in every error message.
A slightly nicer way:
```scala
import NameOps._
//...
}).typeParams.map(_.paramName.unexpandedName.show).....
```
Technically this is still incorrect, `unexpandedName` should only be called on a name of a symbol with the flag `ExpandedName` but we have no way of checking that here, I think the correct solution would be to add a `def simpleParamName` method to `TypeParamInfo`, but this could be mentioned in a `TODO` comment and done later.

felixmulder: I second the `TODO` approach - and perhaps an issue?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
smarter: Extra indentation of two spaces compared to the previous line

ShaneDelmore: Oops.  I was using indentation to try to group related or nested lines of code, but is the concensus that I should just left align everything?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: I was hoping I could fix this but so far don't have a good enough idea of the flow through the compiler to know where the right fix is.  I will spend some time on it this weekend most likely though.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: Nice simplification

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: That is great, I was hunting around for something like it but didn't find it.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: No problem.  I was trading off indentation for vertical space as I worried we didn't want the messages too long. Newlines work just as well for me though.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: Any reason for me not to just add def simpleParamName in this PR?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Yes please :) That's been the policy thus far.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,64 @@ object messages {
...
+    val expectedCount = expectedArgs.length
+    val actualCount = actual.length
+    val msgPrefix = if (actualCount > expectedCount) "Too many" else "Not enough"
+
+    val expectedArgString = (fntpe.widen match {
+      case pt: MethodOrPoly => pt //ensure we return a type that will have useful typeParms
+      case _ => fntpe
+    }).typeParams.map(_.paramName.show.split("\\$").last).mkString("[", ", ", "]")
+
+    val actualArgString = actual.map(_.show.split("\\.").last).mkString("[", ", ", "]")
```
ShaneDelmore: I remember why I had no luck with asSeenFrom, one of the four callers of this function has an untyped arg tree.  Should I split the message into two, one for typed callers and one for untyped callers, or just leave the full path?

felixmulder: If it doesn't have a type (there's a `typeOpt` function available to trees as well as `hasType`) - go for the full type - otherwise `asSeenFrom`?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: No luck, it doesn't return the same result. on the below code
```
object Test {
  case class Bar[D, E](baz: D, qux: E)
  var bad3: Bar[Int] = null
}
```

My original solution returned "Test.Bar" where fntpe.termSymbol returns nonSensicalType

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -514,63 +514,74 @@ object messages {
...
-    }).typeParams.map(_.paramName.show.split("\\$").last).mkString("[", ", ", "]")
+    val expectedArgString = fntpe.widen.typeParams.map(_.paramName.unexpandedName.show).mkString("[", ", ", "]")
+
+    val actualArgString = actual.map(_.show).mkString("[", ", ", "]")
 
-    val actualArgString = actual.map(_.show.split("\\.").last).mkString("[", ", ", "]")
+    val prettyName = fntpe.termSymbol match {
+      case NoSymbol => fntpe.show
+      case symbol   => symbol.showFullName
+    }
```
ShaneDelmore: I would like to make another PR to add def prettyName (or something like it) to a few types.  I will leave it as a follow up unless you prefer it as part of this PR.

With these changes
```
object Test {
  var bad = List[Int, String]
  case class Bar[D, E](baz: D, qux: E)
  var bad3: Bar[Int] = null
}
```
results in
```
-- [E019] Syntax Error: test.scala -------------------------------------------------------------------------------------
16 |  var bad = List[Int, String]
   |            ^^^^^^^^^^^^^^^^^
   |            Too many type arguments for scala.collection.immutable.List.apply[A]
   |            expected: [A]
   |            actual:   [Int, String]

Explanation
===========
You have supplied too many type parameters

For example List takes a single type parameter (List[A])
If you need to hold more types in a list then you need to combine them
into another data type that can contain the number of types you need,
In this example one solution would be to use a Tuple:

val tuple2: (Int, String) = (1, "one")
val list: List[(Int, String)] = List(tuple2)

-- [E019] Syntax Error: test.scala -------------------------------------------------------------------------------------
19 |  var bad3: Bar[Int] = null
   |            ^^^^^^^^
   |            Not enough type arguments for Test.Bar[D, E]
   |            expected: [D, E]
   |            actual:   [Int]

Explanation
===========
You have not supplied enough type parameters
If you specify one type parameter then you need to specify every type parameter.
```

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,75 @@ object messages {
...
   }
 
+  case class WrongNumberOfArgs(fntpe: Type, argKind: String, expectedArgs: List[TypeParamInfo], actual: List[untpd.Tree])(implicit ctx: Context)
+    extends Message(19) {
+    val kind = "Syntax"
+    val expectedCount = expectedArgs.length
+    val actualCount = actual.length
+    val msgPrefix = if (actualCount > expectedCount) "Too many" else "Not enough"
+
+    val expectedArgString = fntpe.widen.typeParams.map(_.paramName.unexpandedName.show).mkString("[", ", ", "]")
```
felixmulder: This was supposed to have a `//TODO` right?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -512,4 +513,75 @@ object messages {
         |}""".stripMargin
   }
 
+  case class WrongNumberOfArgs(fntpe: Type, argKind: String, expectedArgs: List[TypeParamInfo], actual: List[untpd.Tree])(implicit ctx: Context)
+    extends Message(19) {
```
felixmulder: The `errorId` has to be updated - to `21` I think - so 22 for the other one

felixmulder: Sorry, merged @AndrewZurn's, so 22 and 23 - you're next I promise :)

ShaneDelmore: 220 and 230 it is 👍

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ShaneDelmore: I added it now.

felixmulder:
@ShaneDelmore - seems there is still a conflict in `messages.scala`, rebase?

ShaneDelmore:
I rebased and resolved the merge conflict.

PR #1627, opened 2016-10-25
fix #1626: inconsistent names in prettyprint

### LINKED ISSUE #1626: Inconsistent names in pretty-print
Compile following code with `bin/dotc -Xprint:lambdaLift posfix/test.scala`:

``` Scala
object Test {
     lazy val a = 8 * 9
     print(a)
  }
}
```

We get following output:

``` Scala
package <empty> {

  // ... elided

      {
        lazy var a$lzy1: dotty.runtime.LazyInt = new dotty.runtime.LazyInt()
        print(scala.Int.box(Test.a(a$lzy1)))    //  <==== No `Test.a` now, Should be `Test.a$1`
      }

  // ... elided

    private lazy <accessor> def a$1(a$lzy1$2: dotty.runtime.LazyInt): Int =
      if a$lzy1$2.initialized() then a$lzy1$2.value() else Test.a$lzyINIT1(a$lzy1$2)
  }
}
```

Note that in the output, `Test.a` should be `Test.a$1`.

liufengyun:
I'm not sure if `sym.originalName` is intended here. Change it to `sym.name` fixes #1626 .

### DIFF
--- a/compiler/src/dotty/tools/dotc/Run.scala
+++ b/compiler/src/dotty/tools/dotc/Run.scala
@@ -15,6 +15,7 @@ import reporting.Reporter
 import transform.TreeChecker
 import rewrite.Rewrites
 import java.io.{BufferedWriter, OutputStreamWriter}
+import printing.XprintMode
 
 import scala.annotation.tailrec
 import scala.reflect.io.VirtualFile
@@ -95,7 +96,7 @@ class Run(comp: Compiler)(implicit ctx: Context) {
     val unit = ctx.compilationUnit
     val prevPhase = ctx.phase.prev // can be a mini-phase
     val squashedPhase = ctx.squashed(prevPhase)
-    val treeString = unit.tpdTree.show
+    val treeString = unit.tpdTree.show(ctx.withProperty(XprintMode, Some(())))
 
     ctx.echo(s"result of $unit after $squashedPhase:")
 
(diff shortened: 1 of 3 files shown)


liufengyun:
I propose this to be merged. In addition to the confusion in #1626 , inner class names are displayed `Inner` instead of `A$Inner` after flattening with `-Xprint:all`. WDYT @odersky ?

compiler/src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -271,6 +271,18 @@ object Contexts {
       if (this.settings.YtraceContextCreation.value)
         creationTrace = (new Throwable).getStackTrace().take(20)
 
+    /** Whether current prettyprint is for -Xprint
+     *  -Xprint will print `sym.name` instead of `sym.originalName`
+     */
+    private var _isXprintMode = false
```
odersky: This is far too heavyweight. You are adding a field to every context which will have to copied millions of times. There are more lightweight options: 

 - Add a bit to Mode
 - Add a property 

I think adding a property is preferred as it adds the least footprint and the creation and access is not performance critical (adding and testing a mode bit is faster than adding and testing a property).

Also, the logic turns Contexts into mutable objects, which completely changes their role in the compiler and is at odds about all the assumptions we make about them.

liufengyun: Thanks a lot for the explanation and tip, it makes perfect sense. I've now updated the PR based on the `property`-based approach.

compiler/src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -284,6 +284,22 @@ object Contexts {
       println("=== end context creation trace ===")
     }
 
+    /** Whether the compiler is currently doing -Xprint
```
liufengyun: Thanks @odersky , I see what you mean, now it's much neater than the previous attempt 485df22.

odersky:
Now it's perfect, thanks!

PR #1629, opened 2016-10-26
Update error msg for Parsers.scala:1738 and 1739

ljdelight:
These error messages are for aux ctors needing non-implicit params. I'm testing the error messages with this code

```
class Square(val len: Int) {
  // typical cases:
  //  (1) this() parens are forgotten, (2) argument set as implicit
  def this(implicit width: Double) = this(0)
  def this = { this(4) }
}

class MyList(val s: String) {
  def this
```

### DIFF
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -715,4 +715,21 @@ object messages {
            |""".stripMargin
     }
   }
+
+  case class AuxConstructorNeedsNonImplicitParameter()(implicit ctx:Context) extends Message(26) {
+    val kind = "Syntax"
+    val msg = "auxiliary constructor needs non-implicit parameter list"
+    val explanation =
+      hl"""Only the primary constructor is allowed an ${"implicit"} parameter list;
+        |auxiliary constructors need a non-implicit parameter list. When a primary
+        |constructor has an implicit argslist, auxiliary constructors that call the
+        |primary constructor must specify the implicit value.
+        |
+        |To resolve this issue check for:
+        | - forgotten parenthesis on ${"this"} (${"def this() = // ..."})
+        | - auxiliary constructors specify the implicit value
+        |""".stripMargin
+  }
+
+
 }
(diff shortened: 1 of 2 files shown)


sebastianharko:
@ljdelight Awesome 👍 - quick tip: you can get the 'error printer' of dotc or dotr to syntax highlight the example code you gave (class Person): see previous PRs related to errors :-)

ljdelight:
Neat! I had tried to get syntax highlighting using those inline `val explanation = "foobar code is "${def func()}"`, but it would only work with single-line insertions. I didn't realize the use of other variables and inserting those had a change on behavior, but it works great using similar to 
```
val code = "class Person ..."
val explanation = "text ... ${code}"
```

I'm not sure how it works, and thanks for the tip

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -678,4 +678,25 @@ object messages {
...
+
+    val explanation =
+      hl"""An auxiliary constructor requires a non-implicit parameter list. Auxiliary
+        |constructors are allowed to call either auxiliary or primary constructors,
+        |where the primary constructor is allowed implicit parameters. A common mistake
+        |is to forget the parenthesis on `this` (eg `def this() = ...`).
+        |
+        |Here's an example class with primary and auxiliary constructors:
+        |
+        |$code""".stripMargin
```
felixmulder: Perhaps the reason as to why somebody would create an auxiliary constructor with an implicit parameter is because they get the error: `could not find implicit value for parameter X: Type`.

It might be worth mentioning that with auxiliary constructors calling a primary constructor that has an implicit argslist, the implicit value has to be explicitly specified at the callsite.

ljdelight: I'm a bit over my head (learning scala from coursera, on week 7) with creating an example for the other case. I've been looking at `math.Ordering` and I cannot create a simple example; and my example in this PR actually does against the general consensus of not using basic types in `implicit`. Do you think I should remove the example and not have one? I'm considering adding more text to the explanation to help with the case of an implicit primary constructor, thoughts @felixmulder?

felixmulder: For the sake of progress - perhaps omit the `implicit` example?

ljdelight: Updated the text :+1:

felixmulder:
Thanks @ljdelight! :tada:

PR #1630, opened 2016-10-26
Add error message Parsers.scala:1620 missing type keyword

m-sp:


m-sp:
signed the CLA

felixmulder:
Hi @m-sp - thanks for taking the time to make this PR, we'll try to get to it as soon as possible. I also think it would be best if @smarter reviewed this.

Cheers,
Felix

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -677,5 +677,18 @@ object messages {
            |
            |$fixedVarInAlternative""".stripMargin
     }
+  }  
+    
+  case class TypeParamsTypeExpected(mods:Modifiers, identifier:TermName)(implicit ctx:Context) extends Message(24) {
```
smarter: We should always add a space between `:` and a type signature, e.g. `mods: Modifiers`

m-sp:
thanks for the review @smarter, i addressed your comments

felixmulder:
@m-sp: thanks for the contribution! :tada:

PR #1631, opened 2016-10-26
Fix nightly builds

felixmulder:
This commit removes the setting to not publish docs for the `dotty` project.

I suggest either adding a task-key that makes `publishLocal` faster, or simply overriding the setting when publishing locally instead of hardcoding it.


smarter:
LGTM

PR #1633, opened 2016-10-26
Add error message IdentifierExpected

thiagoandrade6:
This commit adds the semantic object fir the `identifier expected` error.
It is part of the https://github.com/lampepfl/dotty/issues/1589

I believe this PR fixes:
- parsing/JavaParsers.scala:233
- Parsers.scala:319

### DIFF
--- a/src/dotty/tools/dotc/parsing/JavaParsers.scala
+++ b/src/dotty/tools/dotc/parsing/JavaParsers.scala
@@ -21,6 +21,7 @@ import Symbols._
 import ast.Trees._
 import Decorators._
 import StdNames._
+import dotty.tools.dotc.reporting.diagnostic.messages.IdentifierExpected
 import dotty.tools.dotc.util.SourceFile
 import util.Positions._
 import annotation.switch
@@ -230,7 +231,7 @@ object JavaParsers {
         case AppliedTypeTree(_, _) | Select(_, _) =>
           tree
         case _ =>
-          syntaxError("identifier expected", tree.pos)
+          syntaxError(IdentifierExpected(tree.show), tree.pos)
           errorTypeTree
       }
     }
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -691,4 +691,28 @@ object messages {
                            |${s"trait A[${mods.flags} type $identifier]"}
                            |""".stripMargin
   }
+
+  case class IdentifierExpected(identifier: String)(implicit ctx: Context) extends Message(25) {
+    val kind = "Syntax"
+
+    val msg = "identifier expected"
+
+    val wrongIdentifier = s"def foo: $identifier = {...}"
+
+    val validIdentifier = s"def foo = {...}"
+
+    val explanation = {
+      hl"""|A valid identifier expected, but `$identifier` found.
+           |Let the compiler infer the type for you.
+           |For example, instead of:
+           |
+           |$wrongIdentifier
+           |
+           |Write your code like:
+           |
+           |$validIdentifier
+           |
+           |""".stripMargin
+    }
+  }
 }
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -316,7 +316,7 @@ object Parsers {
       case id @ Select(qual, name) =>
         cpy.Select(id)(qual, name.toTypeName)
       case _ =>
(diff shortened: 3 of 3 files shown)


thiagoandrade6:
@felixmulder take a look when you have some time please =)

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -678,4 +676,33 @@ object messages {
...
+           |
+           |But you also have the option to omit it completely.
+           |
+           |$typeImplicit
+           |
+           |This time the compiler will infer this type for you.
+           |
+           |""".stripMargin
+    }
+  }
```
felixmulder: So I have a couple of comments/questions on this explanation:

* Does it make sense for all cases?
* Could you show the user their code and the correct way to rewrite it? E.g: `def foo: int = { ... }` => `def foo = { ... }`
* There are some weird things when it comes to the language like `"Because of Scala's stronger type"` - wat? Have a look through it and I'm sure you'll find places to correct :)
*  The line breaking is increasing, first line is something like 60 chars, the second 65 and third 90+, try to even it out.
* The line with "Here we explicitly" is too long (84 chars). We want a line to be max 80 chars (the standard terminal width).

src/dotty/tools/dotc/parsing/JavaParsers.scala
```diff
@@ -21,8 +20,10 @@ import Symbols._
 import ast.Trees._
 import Decorators._
 import StdNames._
+import dotty.tools.dotc.reporting.diagnostic.messages.IdentifierExpected
 import dotty.tools.dotc.util.SourceFile
 import util.Positions._
+
```
thiagoandrade6: no need indeed...

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
thiagoandrade6: Ok, I'm going through your cases and explain to you what I thought right?

1 - Actually, to be honest, the only piece of code I could reproduce this error was this:

```
def foo: this = "bar"
```

and it goes to the `Parsers` class.

Is there any other case I could reproduce it, so I can make sure nothing was left behind?

2 - Get the user code, you mean I could use the `Tree` to show the user wrong code and give the correct one? In fact I could pass the `Tree` inside the class and use it. Don't know how to play with `Tree`s but I will get used to it.

3 - yeah, that one was difficult to write. I mean, Scala has a strong type system, which is good, everything has a type where it could be inferred by the compiler or not(in case the user provided it).

4 & 5 - No problem :)

thiagoandrade6:
Hi @felixmulder take a look on the changes I have done. This time, minimal changes have been applied, no new lines, no import messing, nothing but the required code :)
Also I tried to be simple with the explanation. Take a look if that satisfies.
Thanks for your time :)
Cheers!

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -678,4 +678,28 @@ object messages {
...
+    val validIdentifier = s"def foo = {...}"
+
+    val explanation = {
+      hl"""|A valid identifier is expected, but $identifier was found.
+           |Let the compiler infer the type for you.
+           |For example, instead of:
+           |
+           |$wrongIdentifier
+           |
+           |Write your sentence to:
```
felixmulder: "sentence to" should be "code like"

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -678,4 +678,28 @@ object messages {
...
+    val kind = "Syntax"
+
+    val msg = "identifier expected"
+
+    val wrongIdentifier = s"def foo: $identifier = {...}"
+
+    val validIdentifier = s"def foo = {...}"
+
+    val explanation = {
+      hl"""|A valid identifier is expected, but $identifier was found.
```
felixmulder: maybe quotes around `$identifier` so the line becomes:

``hl"""|A valid identifier expected, but `$identifier` found.``

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
thiagoandrade6: ahhh of course, we are not talking about essay :)

thiagoandrade6:
@felixmulder all set, including merge.

felixmulder:
Great - merging as soon as CI passes :)

thiagoandrade6:
@felixmulder thank you! I will be looking for other issues to help :)

PR #1635, opened 2016-10-27
Add error message - _* syntax - Parsers.scala:1082

AndrewZurn:


AndrewZurn:
#1589

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -678,4 +678,20 @@ object messages {
...
+    val explanation =
+      hl"""|Expected * in '_*' operator.
+           |
+           |If you were to have a method such as:
+           |${"def func(args: Int*) = _ // do something"}
+           |
+           |Then you would provide arguments to it as such:
+           |${"func(Seq(1, 2, 3): _*)"}
+           |""".stripMargin
+  }
```
felixmulder: This can also happen in:

```scala
case class Foo(i: Int*)

(null: Foo) match {
  case Foo(i: _) => // ...
}
```

And providing `func` with arguments could also be done like: `func(1, 2, 3)` right? Basically I think your explanation is good, but try to explain what the `_*` construct does :)

AndrewZurn:
@felixmulder - let me know if that looks a little better. Think I might have captured the essence of it a little better, but please let me know if you would like to see some more work done.

felixmulder:
Looks great! Thanks for the PR :tada:

PR #1636, opened 2016-10-28
Restructuring repo and issues

felixmulder:
Partest was not compiling everything in directories, it was compiling the top-level scala files in the directories.

Which revealed some nasty bugs. Specifically:
- pos/dotc
- pos/dotty
- pos/java-interop
- pos/repl
- pos/tools

These previously passed because only the top-level was compiled, but now do not. Stale symbols galore.

Could I get some help on this branch?

cc: @odersky, @DarkDimius 

PS - since most of these are contained within one another (dotty is a superset of repl etc) I hope it's just one bug.


project/Build.scala
```diff
@@ -221,13 +232,6 @@ object DottyBuild extends Build {
...
       addCommandAlias(
-        "packageAll",
-        ";dotty-interfaces/package" +
-        ";dotty-compiler/package" +
-        ";dotty-library/package" +
-        ";dotty-compiler/test:package"
-      ) ++
-      addCommandAlias(
         "partest",
         ";packageAll" +
```
sjrd: I guess that then `;packageAll` can be removed from all the command aliases :)

project/Build.scala
```diff
@@ -12,7 +12,7 @@ object DottyBuild extends Build {
   val baseVersion = "0.1"
   val isNightly = sys.env.get("NIGHTLYBUILD") == Some("yes")
 
-  val jenkinsMemLimit = List("-Xmx1300m")
+  val jenkinsMemLimit = List("-Xmx1500m")
```
DarkDimius: why is more memory needed after those changes?

project/Build.scala
sjrd: Don't use `Def.sequential` unless you know exactly what you are doing.

`Def.sequential` does not give much *guarantee* in terms of sequentiality. In particular, if from X you depend on task A which depends on a sequence of B, C; but at the same time X depends on C (via another path of dependencies); then B and C will probably still run in parallel.

project/Build.scala
sjrd: I think the commit message explains it.

smarter: See commit message.

DarkDimius: Would you mind explaining the bug?
https://github.com/lampepfl/dotty/pull/1636/commits/7f4f387f220478ce4f8f13b99c7d16f0c254339d 
does a lot of changes but I don't see anything particularly related to `-deep` or something similar.

The biggest tests that we had so far were stdlib and dotty. Stdlib relies on whitelist, and `dotty` did use `-deep` flag. If this flag works unreliably, it would be nice to know as other forks depend on it.

smarter: The bugfix is at: https://github.com/lampepfl/dotty/pull/1636/commits/7f4f387f220478ce4f8f13b99c7d16f0c254339d#diff-ca99a4570465b856d9f97804352f6851R167
compare with the implementation of sources in https://github.com/scala/scala-partest/blob/7c265b38352f146f941a873b153cbfc22d7b7afd/src/main/scala/scala/tools/partest/nest/Runner.scala#L427

smarter:
Remaining issues:
- [ ] pickling/unpickling difference in `tasty_dotc_repl` and other tests: https://gist.github.com/smarter/52a6757915ed2a19104ed7e63e1e6734 (before pickling we have `([+X0] => Vector[X0])[Char]`, after pickling we have `Vector[Char]`, could be fixed by homogenizing maybe.
- [ ] stale symbol in `tasty_tools` and others: https://gist.github.com/smarter/14c26ce25e1ed61f4cc86cc9c0ccd58d

project/Build.scala
DarkDimius: The question still persists. Why does partest need substantially more memory to run?

I'm sure that previous September we were able to compile dotty & stdlib using only 750mb of memory without GC-related issues(i had to work on a very slow machine that had 2GB of memory for a week and I've decreased -Xmx to 750mb). In March, you  https://github.com/lampepfl/dotty/commit/acf05353a599de11e466b5e35cd003a2f6aa5312 have bumped the upper limit to 1.3Gb. Now we bump it to 1.5GB. 

Since last September, tests didn't grow much. Yes, dotty is now 20% bigger, white list is negligibly bigger. Tests do not explain why dotty now needs _twice_ more to pass the test suite, compared to situation 1 year ago.

felixmulder: Previously dotty was not compiled properly by partest. It was only able to compile top-level files (since the compiler itself was on the class path).

It never recursed obeying `-deep`.

DarkDimius: @felixmulder, I do understand that partest did the wrong thing. What I mean is that the current situation indicates that we use too much memory. Note, that even scalac runs entire test suite including partest, perfectly fine under `-Xmx1024m`(that's what is used by CI AFAIK).

The baselines, that seem reasonable to me are: 
 - normal hello-world like tests shouldn't need more than 100mb of memory.
 - dotty & stdlib tests shouldn't need more than 0.5 gigs of memory.

Assuming our CI runs 4 compiler threads, this means that in worst case, we would never need more than 1.2 Gb, and we'll only need 1.2 Gb if we're unlucky and dotty & stlib are compiled concurrently. If you'll have a look at the log output of the build failed with GC https://scala-ci.typesafe.com/job/dotty-master-validate-partest/2417/consoleFull, you'll see that these two tests _did not_ run concurrently.

felixmulder:
Yay, now the unit tests are passing again. Hopefully it's just that pesky stale symbol left now.

project/Build.scala
```diff
@@ -180,6 +181,28 @@ object DottyBuild extends Build {
         )
       }.evaluated,
 
+      // Override run to be able to run compiled classfiles
+      run := {
```
felixmulder: I thought you would say that - but waited until you did. I'll patch it.

felixmulder: fixed :+1:

smarter: Thanks!

smarter:
Rebased (more commits to come).

smarter:
@felixmulder: Remaining known issues:
- The directory structure is inconsistent (some projects have `main/scala/`, some don't), would be nice to fix this in this PR
- Bootstrapping is broken (again!) `./compiler/dotty.jar` and `./compiler/dotty-lib.jar` are empty. This could be fixed in a separate PR once this one is green.

@odersky : Commits you may want to review:
- do not force symbols of annotations when unpickling: https://github.com/lampepfl/dotty/pull/1636/commits/4124a4fc2b1ab9347d5e82d2da1537e3fc90ed50
- https://github.com/lampepfl/dotty/pull/1636/commits/b9982d876c0ceecdbf1a6074d5ad3ead9394de00
- https://github.com/lampepfl/dotty/pull/1636/commits/13692aa7cec74cf316f6836c81336945d6c3701a
- https://github.com/lampepfl/dotty/pull/1636/commits/f32eefcd37fb1900c5137ead3bcc0272726b5f01
- https://github.com/lampepfl/dotty/pull/1636/commits/c6fa1a4e270eba3f41da201a71e6841979fb8992
- https://github.com/lampepfl/dotty/pull/1636/commits/e77868cb4870bb455235e6467687bfcb98a07def
- https://github.com/lampepfl/dotty/pull/1636/commits/1b09bff981243a5697bee84a306290cfd53aadb4
- https://github.com/lampepfl/dotty/pull/1636/commits/3b83061b7ace97deac473963abe7c2dd2f1659aa
- https://github.com/lampepfl/dotty/pull/1636/commits/0412f515beb9e86c77ce23be83b494fb550f812f
- https://github.com/lampepfl/dotty/pull/1636/commits/21935f9a16737fcf8c631b3afbec7d0926c55e87
- https://github.com/lampepfl/dotty/pull/1636/commits/9f6d8b2d37fec3f3fea2bb4e3a395026e894d35d
- https://github.com/lampepfl/dotty/pull/1636/commits/bded3aa9698c54ebeb99c4367485788ddf28333f

smarter:
@felixmulder Also it would be really nice if `sbt run` was an alias for `sbt dotc` to not break existing workflows.

smarter:
> java.lang.OutOfMemoryError: GC overhead limit exceeded

This is with half the number of threads and 1.5GB of heap. We must be leaking memory somehow.

smarter:
At some point the JVM gets stuck doing full GCs that free almost no memory:
```scala
[Full GC (Ergonomics) [PSYoungGen: 340391K->0K(341504K)] [ParOldGen: 1023992K->1023926K(1024000K)] 1364384K->1023926K(1365504K), [Metaspace: 42841K->42841K(1087488K)], 1.3911473 secs] [Times: user=8.66 sys=0.00, real=1.39 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->27594K(341504K)] [ParOldGen: 1023926K->1023959K(1024000K)] 1194934K->1051554K(1365504K), [Metaspace: 42841K->42841K(1087488K)], 1.8500764 secs] [Times: user=13.59 sys=0.03, real=1.85 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->49471K(341504K)] [ParOldGen: 1023959K->1023756K(1024000K)] 1194967K->1073228K(1365504K), [Metaspace: 42841K->42841K(1087488K)], 1.7650672 secs] [Times: user=12.98 sys=0.03, real=1.77 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->57394K(341504K)] [ParOldGen: 1023756K->1023628K(1024000K)] 1194764K->1081023K(1365504K), [Metaspace: 42841K->42841K(1087488K)], 1.7648224 secs] [Times: user=12.97 sys=0.04, real=1.77 secs] 
ok 140 - pos/i1137-1                             

[Full GC (Ergonomics) [PSYoungGen: 171008K->20300K(341504K)] [ParOldGen: 1023628K->1023862K(1024000K)] 1194636K->1044163K(1365504K), [Metaspace: 42841K->42841K(1087488K)], 2.2582140 secs] [Times: user=16.62 sys=0.04, real=2.26 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->34607K(341504K)] [ParOldGen: 1023862K->1023758K(1024000K)] 1194870K->1058366K(1365504K), [Metaspace: 42847K->42847K(1087488K)], 1.5830262 secs] [Times: user=11.62 sys=0.03, real=1.58 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->47778K(341504K)] [ParOldGen: 1023758K->1023695K(1024000K)] 1194766K->1071474K(1365504K), [Metaspace: 42849K->42849K(1087488K)], 1.7812954 secs] [Times: user=12.99 sys=0.06, real=1.78 secs] 

[Full GC (Ergonomics) [PSYoungGen: 171008K->62948K(341504K)] [ParOldGen: 1023695K->1023683K(1024000K)] 1194703K->1086631K(1365504K), [Metaspace: 42849K->42849K(1087488K)], 1.6414215 secs] 
```
I can't reproduce this locally, ideas?

DarkDimius:
try running with  `-XX:AdaptiveSizePolicyOutputInterval=1` on the CI. This should explain why it runs so many `Ergonomics` gc-s.

smarter:
I've dumped the heap of the running JVM on the CI while it's spending all its time trying to do GC: http://martres.me/tmp/dump.hprof but haven't analyzed it yet. I've also dumped some statistics of the live objects on the JVM at that point: http://martres.me/tmp/histolive3
Here's what the top looks like:

| num | #instances| size    | class name |
| --- | --------- | ------------- | ----------------------------------------------------------------- |
| 1:  |   1091834 |   297.839 MB  |   [C |
| 2:  |   71771   |   96.3677 MB  |   [Ljava.lang.Object; |
| 3:  |   2715099 |   62.1437 MB  |   scala.collection.immutable.$colon$colon |
| 4:  |   10912   |   60.7552 MB  |   [B |
| 5:  |   402214  |   30.6865 MB  |   java.util.zip.ZipEntry |
| 6:  |   441263  |   30.2991 MB  |   dotty.tools.dotc.core.SymDenotations$SymDenotation |
| 7:  |   32333   |   28.7728 MB  |   [I |
| 8:  |   1818687 |   27.751 MB   |   dotty.tools.dotc.core.tasty.TastyBuffer$Addr |
| 9:  |   862514  |   26.3218 MB  |   dotty.tools.dotc.core.Types$CachedRefinedType |
| 10: |   997778  |   22.8373 MB  |   java.lang.String |
| 11: |   901523  |   20.6342 MB  |   scala.collection.mutable.DefaultEntry |
| 12: |   530744  |   20.2463 MB  |   dotty.tools.dotc.ast.Trees$Ident |
| 13: |   402165  |   18.4097 MB  |   scala.reflect.io.FileZipArchive$FileEntry$1 |
| 14: |   598117  |   18.2531 MB  |   dotty.tools.dotc.core.Types$CachedTypeAlias |
| 15: |   394378  |   18.0532 MB  |   dotty.tools.dotc.core.Types$CachedMethodType |
| 16: |   136654  |   17.724 MB   |   dotty.tools.dotc.core.Contexts$InitialContext |
| 17: |   376390  |   17.2298 MB  |   dotty.tools.dotc.ast.Trees$Apply |
| 18: |   521255  |   15.9074 MB  |   dotty.tools.dotc.core.Types$PolyParam |
| 19: |   515886  |   15.7436 MB  |   dotty.tools.dotc.core.Scopes$ScopeEntry |
| 20: |   320749  |   14.6827 MB  |   dotty.tools.dotc.ast.Trees$Select |
| 21: |   98580   |   12.0337 MB  |   dotty.tools.dotc.core.SymDenotations$ClassDenotation |
| 22: |   43044   |   11.3244 MB  |   [Lscala.collection.mutable.HashEntry; |
| 23: |   341227  |   10.4134 MB  |   java.util.HashMap$Node |
| 24: |   429543  |   9.83146 MB  |   dotty.tools.dotc.core.Symbols$Symbol |
| 25: |   300668  |   9.17566 MB  |   scala.collection.mutable.LinkedEntry |
| 26: |   240531  |   9.17553 MB  |   dotty.tools.dotc.ast.Trees$TypeTree |
| 27: |   169674  |   9.06157 MB  |   dotty.tools.dotc.ast.Trees$ValDef |
| 28: |   193489  |   8.85722 MB  |   dotty.tools.dotc.core.Types$TermRefWithSignature |
| 29: |   193347  |   8.85072 MB  |   dotty.tools.dotc.core.Types$TermRefWithFixedSym |
| 30: |   164612  |   8.79123 MB  |   dotty.tools.dotc.core.Types$PolyType |
| 31: |   156511  |   7.16451 MB  |   dotty.tools.dotc.core.Types$ImplicitMethodType |
| 32: |   111518  |   6.80652 MB  |   dotty.tools.dotc.ast.Trees$DefDef |
| 33: |   292028  |   6.68399 MB  |   dotty.tools.dotc.util.Positions$Position |
| 34: |   398729  |   6.08412 MB  |   dotty.tools.dotc.core.tasty.TastyName$NameRef |
| 35: |   117554  |   5.3812 MB   |   dotty.tools.dotc.ast.Trees$Block |
| 36: |   76893   |   5.27982 MB  |   dotty.tools.dotc.core.TypeComparer |
| 37: |   114997  |   5.26414 MB  |   dotty.tools.dotc.ast.untpd$Modifiers |

smarter:
Looks like I'm on the right track and aggressive null-ing in `testUnpickler` really helped :sparkles:, let's try increasing the number of threads used again...

nightscape:
@smarter You might know that tool already, but I just had to do an analysis of a memory leak, and [Eclipse MAT](http://www.eclipse.org/mat/) helped greatly there.
You just feed it the heap dump and it can automatically search for leak suspects (i.e. objects with transitive links to a set of objects that consumes most of the memory).

smarter:
@nightscape Thanks, I'll have a look, so far I've only used visualvm

smarter:
I think the correct thing to do here is to set up partest so that tests with`-Ytest-pickler` are never run in parallel with anything else, they just use too much heap, and there's no easy way to avoid that. We could first run every test that is not a pickling test as we do now, then run every pickling test sequentially.

smarter:
Alright, I think I'm done now, it would be very helpful if people using IDEs could test this PR and see if everything still works with the new directory structure (you'll probably have to reimport the project).

smarter:
@felixmulder `./bin/dotc` and `./bin/dotr` are broken, even if I delete `.packages`, the generated `.packages` is wrong sometimes: it may contain test jars because your calls to build_jar do not avoid them.

smarter:
(I think the whole `.packages` idea is wrong, it will break everytime we change version number for example, but now may not be the time to fix it)

felixmulder:
In Berlin, but I've been following the progress. The bin scripts should be
made to not contain any test jars. Previously it was needed for the type
stealer IIRC. If it isn't something you want to do I'll deal with it
tomorrow evening or Monday morning :) thanks for all the hard work guys!

nicolasstucki:
I tried it with IntelliJ. I could generate the IntelliJ project and use it without any issues.

felixmulder:
This has now been tried with both Eclipse and IntelliJ (@sjrd & @nicolasstucki). So should be good to go.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -881,7 +881,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
         readType() match {
           case path: TypeRef => TypeTree(path)
           case path: TermRef => ref(path)
-          case path: ThisType => This(path.cls)
+          case path: ThisType => untpd.This(untpd.EmptyTypeIdent).withType(path)
```
odersky: This one LGTM

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -1026,7 +1026,10 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
             case APPLIEDtpt =>
               AppliedTypeTree(readTpt(), until(end)(readTpt()))
             case ANDtpt =>
-              AndTypeTree(readTpt(), readTpt())
+              val tpt1 = readTpt()
+              val tpt2 = readTpt()
+              // FIXME: We need to do this instead of "AndType(tpt1, tpt2)" to avoid self-type cyclic reference in tasty_tools
+              untpd.AndTypeTree(tpt1, tpt2).withType(AndType(tpt1.tpe, tpt2.tpe))
```
odersky: These changes look good, but I wonder whether we should generalize them: Never use & or simplify when dealing with type trees. This looks logically consistent. Have you tried this?

smarter: Haven't tried it yet, mostly because I wasn't sure if I should do the same with `|`, etc and wanted to get this PR in first.

smarter:
Alright, rebasing on top of master with https://github.com/lampepfl/dotty/pull/1725 merged seems to have helped with the remaining memory issues, now partest consistently runs in ~15 minutes. Merging now before something else breaks! :)

PR #1645, opened 2016-10-31
Error message Parsers.scala:307

thiagoandrade6:
This commit updates error message for Parsers.scala:307
It is part of the #1589

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -304,7 +304,7 @@ object Parsers {
       case Typed(Ident(name), tpt) =>
         makeParameter(name.asTermName, tpt, mods) withPos tree.pos
       case _ =>
-        syntaxError(s"not a legal $expected", tree.pos)
+        syntaxError(NotALegalFormalParameter(expected), tree.pos)
         makeParameter(nme.ERROR, tree, mods)
     }
 
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -677,8 +677,8 @@ object messages {
            |
            |$fixedVarInAlternative""".stripMargin
     }
-  }  
-    
+  }
+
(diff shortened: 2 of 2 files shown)


thiagoandrade6:
@felixmulder I don't know if it's too simple or if I'm missing something. I could reproduce this writing the following pieces of code:

```
Map(1 => 123)
val sum = (foo: Int, bar Int) => foo + bar //note the missing `:` between bar and Int
```

You would add something else to this explanation? Or should I consider a different scenario?

felixmulder:
from the comment on `implicitClosure` you need to handle:

```
Expr         ::= implicit Id `=>' Expr
BlockResult  ::= implicit Id [`:' InfixType] `=>' Block
```

The method `convertToParam` is also called from `templateStatSeq()`, as such - this also applies to self-types.

Blaisorblade:
@thiagoandrade6 Sorry for missing your PR! What I had until now is this:

https://github.com/lampepfl/dotty/compare/master...Blaisorblade:topic/1645-parser-error

I haven't written the explanations, but I think the one for self-types is probably going to be different, so I've used a separate case class. I've added there a comment explaining a bit about what the parser is doing with self-types IIUC.
I'll leave the rest to you!

thiagoandrade6:
@Blaisorblade thanks, I will check out that.

thiagoandrade6:
@felixmulder I got confused about the syntax described in the comments you have pointed out, but looking at the `docs` I think I could understand a bit more how you describe stuffs.
So, I will be contributing to the @Blaisorblade repo with the explanation. Do you think my explanation here is valid for the case of `Illegal formal parameter`? The case of self type, it's another history :)

I was looking on this SO to see more about the self-type error...

[http://stackoverflow.com/questions/4774743/scala-self-type-member-of-type-parameter-error](http://stackoverflow.com/questions/4774743/scala-self-type-member-of-type-parameter-error)

Do you think I could use it as the basis of the self type explanation?

thiagoandrade6:
@Blaisorblade I have closed my PR, I think it's better to proceed with your code. Also I'm going to study some resources and I'm afraid I will not be able to help you on that, sorry!

Issue #1579, opened 2016-10-11
Adapt the sbt bridge for the new error messages


felixmulder:
The sbt bridge currently does not use the new scheme for error messages.

It simply delegates the error message (not the complete error) to the sbt reporter. To fix this we should be able to simply do the following:
- Adapt `src/dotty/.../ConsoleReporter.scala` to produce a complete string of the whole error message before printing
- Hand the above string to the delegate in `bridge/.../DelegatingReporter.scala`
- Check that the pretty error messages are now available without hickups in an sbt example project

slothspot:
@felixmulder if nobody works on this, I'm going to start.

> Check that the pretty error messages are now available without hickups in an sbt example project
Does it mean that I need to contribute to smarter/dotty-example-project code with errors?

felixmulder:
Hi @slothspot - you're more than welcome to work on this :)

You will probably not need to do modify anything in the example project. But you'll need to tweak the bridge and you'll need to do `publishLocal` to check the results. Let us know on gitter if you need any help!

PR #1646, opened 2016-10-31
[WIP] #1579 Adapt the sbt bridge for the new error message

slothspot:
I started adaptation of sbt bridge.
Right now I have following output for next samples:
`case class Foo`
results in (with explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:10:
[error] case class Foo
[error]            ^^^
[error] A case class must have at least one parameter list
[error] Foo must have at least one parameter list, if you would rather
[error] have a singleton representation of Foo, use a "case object".
[error] Or, add an explicit `()' as a parameter list to Foo.
```
results in (without explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:10:
[error] case class Foo
[error]            ^^^
[error] A case class must have at least one parameter list
```

`println("msg", 123, None)`
results in (with explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:23:
[error]     println("msg", 123, None)
[error]     ^^^^^^^
[error] none of the overloaded alternatives of method println in object Predef with types
[error]  (x: Any)Unit
[error]  ()Unit
[error] match arguments (String("msg"), Int(123), None.type)
```
results in (without explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:23:
[error]     println("msg", 123, None)
[error]     ^^^^^^^
[error] none of the overloaded alternatives of method println in object Predef with types
[error]  (x: Any)Unit
[error]  ()Unit
[error] match arguments (String("msg"), Int(123), None.type)
```

```
case class Bar(s: String,
  i: Itn,
  b: Boolean)
```
results in (with explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:13:
[error]   i: Itn,
[error]      ^^^
[error] not found: type Itn
[error] An identifier for `type Itn` is missing. This means that something
[error] has either been misspelt or you're forgetting an import
```
results in (without explain):
```
[error] /Users/dmitry/Projects/dotty-example-project/src/main/scala/Main.scala:13:
[error]   i: Itn,
[error]      ^^^
[error] not found: type Itn
```

At this moment I would like to receive suggestions about errors output format:
should it be the same as used by dotty or proposed one is acceptable? (I opt for second because of sbt messages decoration)

### DIFF
--- a/bridge/src/main/scala/xsbt/DelegatingReporter.scala
+++ b/bridge/src/main/scala/xsbt/DelegatingReporter.scala
@@ -9,12 +9,13 @@ import reporting._
 import reporting.diagnostic.MessageContainer
 import reporting.diagnostic.messages
 import core.Contexts._
-
 import xsbti.{Maybe, Position}
 
 final class DelegatingReporter(delegate: xsbti.Reporter) extends Reporter
   with UniqueMessagePositions
-  with HideNonSensicalMessages {
+  with HideNonSensicalMessages
+  with MessageRendering {
+  import MessageContainer._
 
   override def printSummary(implicit ctx: Context): Unit = delegate.printSummary()
 
@@ -25,36 +26,23 @@ final class DelegatingReporter(delegate: xsbti.Reporter) extends Reporter
         case _: messages.Warning => xsbti.Severity.Warn
         case _                   => xsbti.Severity.Info
       }
-    val pos =
-      if (cont.pos.exists) Some(cont.pos)
-      else None
-
-    val file =
-      if (cont.pos.source.file.exists) Option(cont.pos.source.file.file)
-      else None
-
-    val offset0 = pos.map(_.point)
 
     val position = new Position {
-      def line: Maybe[Integer] = maybe(pos.map(_.line))
-      def lineContent: String = pos.map(_.lineContent).getOrElse("")
-      def offset: Maybe[Integer] = maybeInt(offset0)
-      def pointer: Maybe[Integer] = offset
-      def pointerSpace: Maybe[String] = maybe(offset0.map(" " * _))
-      def sourceFile: Maybe[java.io.File] = maybe(file)
-      def sourcePath: Maybe[String] = maybe(file.map(_.getPath))
+      def line: Maybe[Integer] = Maybe.nothing()
+      def lineContent: String = ""
+      def offset: Maybe[Integer] = Maybe.nothing()
+      def pointer: Maybe[Integer] = Maybe.nothing()
+      def pointerSpace: Maybe[String] = Maybe.nothing()
+      def sourceFile: Maybe[java.io.File] = Maybe.nothing()
+      def sourcePath: Maybe[String] = Maybe.nothing()
     }
 
-    delegate.log(position, cont.message, severity)
-  }
+    val sb = new StringBuilder()
+    sb.append(messageAndPos(cont.contained, cont.pos, diagnosticLevel(cont)))
+    if (ctx.shouldExplain(cont) && cont.contained.explanation.nonEmpty) {
+      sb.append(explanation(cont.contained))
+    }
 
-  private[this] def maybe[T](opt: Option[T]): Maybe[T] = opt match {
-    case None => Maybe.nothing[T]
-    case Some(s) => Maybe.just[T](s)
-  }
-  import java.lang.{ Integer => I }
-  private[this] def maybeInt(opt: Option[Int]): Maybe[I] = opt match {
-    case None => Maybe.nothing[I]
-    case Some(s) => Maybe.just[I](s)
+    delegate.log(position, sb.toString(), severity)
   }
 }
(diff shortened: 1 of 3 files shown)


felixmulder:
I think it should present the same output as dotc. We worked hard to make sure that the output from dotc is readable and well separated such that it is easy to know where errors end and begin. There's no point in better error messages if you can't tell where one starts and the other begins :smiley:

Instead of implementing a "new" reporter, I would look at adapting the `ConsoleReporter` such that you can get the complete String and pass it to the delegating reporter. Perhaps this could simply be accomplished by refactoring `doReport` and changing `printMessageAndPos` to `messageAndPos` returning a String.

Thus you've eliminated the need for maintaining two reporters and improvements made to `ConsoleReporter` will carry over to the sbt reporter :+1:

felixmulder:
Since we have little to no control over the sbt-API, it is possible that you'll need to for instance give sbt an empty position so that you can reuse the one that the `ConsoleReporter` gives you - but that will reveal itself during the implementation :) happy coding!

slothspot:
@felixmulder thanks for your feedback and suggestions.
I have same things on my mind and started hack on ConsoleReporter already.
Need to figure out how sbt handles a location and new lines at the end of a log message, through.

slothspot:
@felixmulder I've updated PR, now output is like this:

Without -explain:
screen shot 2016-11-03 at 22 30 56

With -explain:
screen shot 2016-11-03 at 22 31 38

felixmulder:
Or you know what - I see the point of moving them out into a separate object. Keep them separate - but make it a `trait` that you then mix into both the `ConsoleReporter` and `DelegatingReporter`. That way, you keep the rendering logic separate from the entity actually doing the printing.

I'd call the trait something like `MessageRendering` so that it becomes:

`class ConsoleReporter(...) extends Reporter with ... with MessageRendering { .. }`

src/dotty/tools/dotc/reporting/MessageUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+import core.Contexts.Context
+import core.Decorators._
+import printing.Highlighting.{Blue, Red}
+import diagnostic.{Message, MessageContainer, NoExplanation}
+import diagnostic.messages._
+import util.SourcePosition
+
+import scala.collection.mutable
+
+object MessageUtils {
```
felixmulder: So this guy becomes `MessageRendering`

src/dotty/tools/dotc/reporting/MessageUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+        .map { case (prefix, line) =>
+          val lnum = Red(" " * math.max(0, maxLen - prefix.length) + prefix)
+          hl"$lnum$line"
+        }
+
+    val (before, after) = pos.beforeAndAfterPoint
+    (render(before), render(after), maxLen)
+  }
+
+  def columnMarker(pos: SourcePosition, offset: Int)(implicit ctx: Context):String = {
```
felixmulder: Space after `:`

src/dotty/tools/dotc/reporting/MessageUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+      val prefix = s"-- ${errId}${kind}: $file "
+
+      prefix +
+        ("-" * math.max(ctx.settings.pageWidth.value - stripColor(prefix).length, 0))
+    }).show else ""
+
+  def explanation(m: Message)(implicit ctx: Context): String = {
+    val sb = new StringBuilder(hl"""|
+                      |${Blue("Explanation")}
+                                    |${Blue("===========")}""".stripMargin)
```
felixmulder: Weird indentation here, ident under the `|` signs

src/dotty/tools/dotc/reporting/MessageUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+    if (pos.exists) {
+      val (srcBefore, srcAfter, offset) = sourceLines(pos)
+      val marker = columnMarker(pos, offset)
+      val err = errorMsg(pos, msg.msg, offset)
+      sb.append((srcBefore ::: marker :: err :: srcAfter).mkString("\n"))
+    } else sb.append(msg.msg)
+    sb.toString
+  }
+
+  def diagnosticLevel(cont: MessageContainer):String = {
```
felixmulder: Space after `:` here as well

slothspot:
Done

felixmulder:
Great job! We really appreciate this PR as it greatly improves the experience when using the [sbt-dotty](github.com/felixmulder/sbt-dotty) plugin or the [dotty-example-project](github.com/smarter/dotty-example-project) :tada:

PR #1654, opened 2016-11-03
Add error messages: no class or object after case keyword, expected class or object definition

jozi-k:


felixmulder:
Hey @jozi-k - sign the CLA please: https://www.lightbend.com/contribute/cla/scala

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -4,7 +4,7 @@ package parsing
 
 import scala.collection.mutable.ListBuffer
 import scala.collection.immutable.BitSet
-import util.{ SourceFile, SourcePosition }
+import util.{SourceFile, SourcePosition}
```
felixmulder: Unnecessary change?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -729,4 +729,11 @@ object messages {
            | - auxiliary constructors specify the implicit value
            |"""
   }
+
+  case class NoClassOrObjectAfterCase()(implicit ctx: Context)
+  extends Message(27) {
+    val kind = "Syntax"
+    val msg = hl"Only ${"class"} or ${"object"} are allowed after ${"case"}."
+    val explanation = ""
```
felixmulder: Perhaps explain why this is in the explanation? Perhaps a small `case class` and `case object` explanation here.

src/dotty/tools/dotc/parsing/Parsers.scala
jozi-k: Reverted in new version.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: I think the explanation should be something explaining to the user why (s)he can't create a `case trait`. Perhaps you don't need to explain case classes/objects but the user should be given some alternatives to fix the error.

For example: (s)he could use `abstract case class`

jozi-k: This is not about `case trait` only. You could write `case val`, `case match` or `case foo`. So in general I think explanation, that only class/object are allowed after case keyword is sufficient. Or we can match token after `case` and if it is `trait`, then print another error message which will inform only about alternatives to `case trait`.

felixmulder: Yes, `case trait` is just an example of what the user can write. What I'm trying to say is that the user should (if possible) be given the right solution to the problem.

I think that `msg` saying: `"Only ${"class"} or ${"object"} are allowed after ${"case"}."` is perfectly valid and sufficient.

And I think you should go for your second suggestion :) the point of having better error messages is indeed that they should be better! Do a lookahead at the next token and see if you can give some meaningful help in the `explanation` val. Sounds good?

src/dotty/tools/dotc/parsing/Parsers.scala
felixmulder: Should have been squished into 1282d77f41976355dc00dd87d62f2c28c3e5b90f, but whatever - it's not the end of the world.

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
jozi-k: The only possible explanation I can think of is for `case trait`. Do you think there can be any more meaningful token used after case?

felixmulder: Here are a few possible scenarios, starting with the one you mentioned:

1. next token is `TRAIT`, give explanation on case traits not being a thing
2. next token is `t' such that `t >= CHARLIT  && t =< IDENTIFIER` you can't have `case` followed by these outside a match expression
3. same for `TRUE` and `FALSE`

Have a look at [Tokens.scala](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/parsing/Tokens.scala) for more tokens that could follow. This is somewhat open ended, so feel free to limit yourself to a few scenarios and  we'll detail this in #1589 :)

felixmulder:
This needs to be rebased against master - that's why there's a huuuuuge diff ^^

felixmulder:
Something seems to be very wrong with this PR - might be Github's fault or a rebase gone bad. Anyway - we'd love to see a revisit to this if you still have time for it.

Cheers,
Felix

PR #1666, opened 2016-11-06
Fix #1653: Check "no inherit from final" earlier.

### LINKED ISSUE #1653: crash with assertion failed ..
$ cat abc.scala
```
trait Foo {
  def foo() = new Unit with Foo
}
```

$ dotc abc.scala
```
error while transforming () extends Unit() with Foo {}
error while transforming final class $anon() extends Unit() with Foo {}
error while transforming {
  final class $anon() extends Unit() with Foo {}
  new Unit with Foo{...}(): Unit & Foo
}
error while transforming def foo(): Unit & Foo =
  {
    final class $anon() extends Unit() with Foo {}
    new Unit with Foo{...}(): Unit & Foo
  }
error while transforming () extends Object {
  def foo(): Unit & Foo =
    {
      final class $anon() extends Unit() with Foo {}
      new Unit with Foo{...}(): Unit & Foo
    }
}
error while transforming @scala.annotation.internal.SourceFile("abc.scala") <trait> trait Foo() extends Object {
  def foo(): Unit & Foo =
    {
      final class $anon() extends Unit() with Foo {}
      new Unit with Foo{...}(): Unit & Foo
    }
}
error while transforming package <empty> {
  @scala.annotation.internal.SourceFile("abc.scala") <trait> trait Foo() extends Object {
    def foo(): Unit & Foo =
      {
        final class $anon() extends Unit() with Foo {}
        new Unit with Foo{...}(): Unit & Foo
      }
  }
}

	at dotty.tools.dotc.transform.SyntheticMethods.dotty$tools$dotc$transform$SyntheticMethods$$valueHashCodeBody$1(SyntheticMethods.scala:141)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticRHS$1$1.apply(SyntheticMethods.scala:87)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticRHS$1$1.apply(SyntheticMethods.scala:87)
	at dotty.tools.dotc.ast.tpd$.polyDefDef(tpd.scala:203)
	at dotty.tools.dotc.ast.tpd$.DefDef(tpd.scala:173)
	at dotty.tools.dotc.transform.SyntheticMethods.syntheticDef$1(SyntheticMethods.scala:96)
	at dotty.tools.dotc.transform.SyntheticMethods.dotty$tools$dotc$transform$SyntheticMethods$$syntheticDefIfMissing$1(SyntheticMethods.scala:70)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticMethods$1.apply(SyntheticMethods.scala:190)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticMethods$1.apply(SyntheticMethods.scala:190)
	at scala.collection.immutable.List.flatMap(List.scala:327)
```

odersky:
The test case is an illegal inheritance from a primitive value class,
which makes this an illegal value class. Previously, the error was
detected by refchecks, but crashes occurred earlier (when generating
synthesized methods) or at the same phase block (in extension methods).

The problem is avoided by moving the test to Namer.

Review by @smarter.

### DIFF
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -666,6 +666,8 @@ class Namer { typer: Typer =>
        * (1) It must be a class type with a stable prefix (@see checkClassTypeWithStablePrefix)
        * (2) If may not derive from itself
        * (3) Overriding type parameters must be correctly forwarded. (@see checkTypeParamOverride)
+       * (4) The class is not final
+       * (5) If the class is sealed, it is defined in the same compilation unit as the current class
        */
       def checkedParentType(parent: untpd.Tree, paramAccessors: List[Symbol]): Type = {
         val ptype = parentType(parent)(ctx.superCallContext)
@@ -684,7 +686,14 @@ class Namer { typer: Typer =>
           }
           else if (!paramAccessors.forall(checkTypeParamOverride(pt, _)))
             defn.ObjectType
-          else pt
+          else {
+            val pclazz = pt.typeSymbol
+            if (pclazz.is(Final))
+              ctx.error(em"cannot extend final $pclazz", cls.pos)
+            if (pclazz.is(Sealed) && pclazz.associatedFile != cls.associatedFile)
+              ctx.error(em"cannot extend sealed $pclazz in different compilation unit", cls.pos)
+            pt
+          }
         }
       }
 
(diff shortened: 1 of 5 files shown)


smarter:
#1670 could be fixed in the same way

DarkDimius:
I propose to reconsider this implementation. 
This check can safely go into Infotransformer of RefChecks instead of making typer\namer responsible of even more stuff.

PR #1669, opened 2016-11-06
Fix #1668 - add encoding for default docs layout

### LINKED ISSUE #1668: Docs - html page have no encoding
For example [here](http://dotty.epfl.ch/docs/internals/contexts.html) at bottom of context's internals list occurs invalid characters

dos65:


felixmulder:
LGTM :tada:

Issue #1647, opened 2016-10-31
crash with "java.lang.IllegalArgumentException: requirement failed"


helloqirun:
$ cat abc.scala
```
class ann {
  @ann({ def baz })
  def foo()
}
```

$ dotc abc.scala
```
-- [E018] Syntax Error: abc.scala --------------------------------------------------------------------------------------
3 |  @ann({ def baz })
  |                ^
  |                missing return type

longer explanation available when compiling with `-explain`
-- [E018] Syntax Error: abc.scala --------------------------------------------------------------------------------------
4 |  def foo()
  |           ^
  |           missing return type

longer explanation available when compiling with `-explain`

exception occurred while typechecking abc.scala

exception occurred while compiling abc.scala
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:207)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.enterNoReplace(SymDenotations.scala:1548)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.enter(SymDenotations.scala:1538)
	at dotty.tools.dotc.typer.NamerContextOps$class.enter(Namer.scala:34)
	at dotty.tools.dotc.core.Contexts$Context.enter(Contexts.scala:57)
	at dotty.tools.dotc.typer.Namer.enterSymbol(Namer.scala:334)
	at dotty.tools.dotc.typer.Namer.dotty$tools$dotc$typer$Namer$$recur$1(Namer.scala:417)
	at dotty.tools.dotc.typer.Namer.indexExpanded(Namer.scala:427)
<snipped>
```

liufengyun:
Report confirmed, thanks @helloqirun !

odersky:
This has to do with the fact that the annotation argument is typed with the enclosing class as owner, whereas the owner should be the localDummy of the template. Does not look so easy to fix though.

PR #1677, opened 2016-11-07
Use inline flag instead of @inline annotation

odersky:
Convert `@inline` annotations to `inline` flags, not the
other way round as was done before.

This is the first step to fixing #1647, which requires a reorganization how
annotations are handled. 

I wanted to get the irregular treatement of inline annotations out of the way before trying this. 

Review by @smarter. Feel free to cherry-pick for getting the new repo structure to work.

### DIFF
--- a/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -751,10 +751,7 @@ object SymDenotations {
     //    def isOverridable: Boolean = !!! need to enforce that classes cannot be redefined
     def isSkolem: Boolean = name == nme.SKOLEM
 
-    def isInlineMethod(implicit ctx: Context): Boolean =
-      is(Method, butNot = Accessor) &&
-      !isCompleting &&   // don't force method type; recursive inlines are ignored anyway.
-      hasAnnotation(defn.InlineAnnot)
+    def isInlineMethod(implicit ctx: Context): Boolean = is(InlineMethod, butNot = Accessor) 
 
     // ------ access to related symbols ---------------------------------
 
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -521,6 +521,7 @@ class Namer { typer: Typer =>
     mergeCompanionDefs()
     val ctxWithStats = (ctx /: stats) ((ctx, stat) => indexExpanded(stat)(ctx))
     createCompanionLinks(ctxWithStats)
+    //stats foreach enterAnnotations
     ctxWithStats
   }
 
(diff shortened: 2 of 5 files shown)


src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -751,10 +751,7 @@ object SymDenotations {
     //    def isOverridable: Boolean = !!! need to enforce that classes cannot be redefined
     def isSkolem: Boolean = name == nme.SKOLEM
 
-    def isInlineMethod(implicit ctx: Context): Boolean =
-      is(Method, butNot = Accessor) &&
-      !isCompleting &&   // don't force method type; recursive inlines are ignored anyway.
-      hasAnnotation(defn.InlineAnnot)
+    def isInlineMethod(implicit ctx: Context): Boolean = is(InlineMethod, butNot = Accessor)
```
smarter: Is the `butNot` still necessary? Do accessors get the inline flag too somehow?

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -521,6 +521,7 @@ class Namer { typer: Typer =>
     mergeCompanionDefs()
     val ctxWithStats = (ctx /: stats) ((ctx, stat) => indexExpanded(stat)(ctx))
     createCompanionLinks(ctxWithStats)
+    //stats foreach enterAnnotations
```
smarter: Stray comment?

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -568,19 +569,9 @@ class Namer { typer: Typer =>
...
-          // create a @inline annotation. Currently, the inlining trigger
-          // is really the annotation, not the flag. This is done so that
-          // we can still compile inline methods from Scala2x. Once we stop
-          // being compatible with Scala2 we should revise the logic to
-          // be based on the flag. Then creating a separate annotation becomes unnecessary.
-          denot.addAnnotation(Annotation(defn.InlineAnnot))
-          addInlineInfo(denot, original)
+          if (cls == defn.InlineAnnot && denot.is(Method, butNot = Accessor))
+            denot.setFlag(Inline)
+          if (denot.isInlineMethod) addInlineInfo(denot, original)
```
smarter: This means that `addInlineInfo` will be called for every annotation if the method is marked inline, previously `hasInlineAnnot` prevented that.

src/dotty/tools/dotc/core/SymDenotations.scala
odersky: They might if we have an accessor of an inline val. Better play it safe.

odersky:
Merging now, because there's more to come...

PR #1678, opened 2016-11-07
Resync copy of the type diagram in docs

Blaisorblade:
Also add a note that the two diagrams should be kept in sync, both ways,
so there's a chance that the copies are kept in sync.

Alternatively one could drop a copy.


smarter:
LGTM, thanks!

Issue #1670, opened 2016-11-06
Synth method generation failed on invalid value class


dos65:
Compiler crashes on [synthetic methods generation](https://github.com/lampepfl/dotty/blob/70a8426912f6f256536470a046cdf86827ab2c93/src/dotty/tools/dotc/transform/SyntheticMethods.scala#L141) on value class with more that one parameter
```scala
class A(a:Int, b:Int) extends AnyVal 
```
```
error while transforming (a: Int, b: Int) extends AnyVal() { 
  private[this] val a: Int
  private[this] val b: Int
}
error while transforming @scala.annotation.internal.SourceFile("/home/dos65/projects/tmp/dotty_samples/3.scala") final class A(a: Int, b: Int)
   extends
 AnyVal() { 
  private[this] val a: Int
  private[this] val b: Int
}
error while transforming package <empty> {
  @scala.annotation.internal.SourceFile("/home/dos65/projects/tmp/dotty_samples/3.scala") final class A(a: Int, b: Int)
     extends
   AnyVal() { 
    private[this] val a: Int
    private[this] val b: Int
  }
  final lazy module val A: A$ = new A$()
  final module class A$() extends Object() { this: A.type =>}
}

exception occurred while compiling /home/dos65/projects/tmp/dotty_samples/3.scala
Exception in thread "main" java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:151)
	at dotty.tools.dotc.transform.SyntheticMethods.dotty$tools$dotc$transform$SyntheticMethods$$valueHashCodeBody$1(SyntheticMethods.scala:141)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticRHS$1$1.apply(SyntheticMethods.scala:87)
	at dotty.tools.dotc.transform.SyntheticMethods$$anonfun$syntheticRHS$1$1.apply(SyntheticMethods.scala:87)
	at dotty.tools.dotc.ast.tpd$.polyDefDef(tpd.scala:200)
	at dotty.tools.dotc.ast.tpd$.DefDef(tpd.scala:170)
	at dotty.tools.dotc.transform.SyntheticMethods.syntheticDef$1(SyntheticMethods.scala:96)
	at dotty.tools.dotc.transform.SyntheticMethods.dotty$tools$dotc$transform$SyntheticMethods$$syntheticDefIfMissing$1(SyntheticMethods.scala:70)
[...]

Issue #1642, opened 2016-10-29
Disallow having a value classe whose underlying value is itself


helloqirun:
dotc crashes when compiling the follow code. 

$ cat abc.scala

```
class Test2(val valueVal: Test2) extends AnyVal
```

$ dotc abc.scala
output:
https://gist.github.com/helloqirun/d09e36c276832140e7c74d1f26378a14

smarter:
For very long output, you can paste them on https://gist.github.com/ and put a link in your issue.

helloqirun:
Thanks, updated.

PR #1682, opened 2016-11-07
Fix checks related to value classes

liufengyun:
Fix checks related to value classes: #1670 #1642 

Note: despite the discussion today, the simplest fix is to move some `RefChecks` to `typer`. The TODO comment in the typer already points the direction for the migration:

```
+
+
+    // check value class constraints
+    RefChecks.checkDerivedValueClass(cls, body1)
+
     cdef1

     // todo later: check that
     //  1. If class is non-abstract, it is instantiatable:
     //  - self type is s supertype of own type
     //  - all type members have consistent bounds
     // 2. all private type members have consistent bounds
     // 3. Types do not override classes.
     // 4. Polymorphic type defs override nothing.

```

And I'd argue the complexity is manageable -- typer is not becoming more complex. And having checks earlier simplifies error handling and later processing.

Review @odersky .


src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -720,6 +720,9 @@ object RefChecks {
           case List(param) =>
             if (param.is(Mutable))
               ctx.error("value class parameter must not be a var", param.pos)
+
+            if (isDerivedValueClass(param.info.typeSymbol))
+              ctx.error("value class may not wrap another user-defined value class", param.pos)
```
smarter: No. That's completely fine and supported in Dotty. The only issue is when there is a cycle, this happens a value class underlying type (or the underlying type of its underlying type or ...) is itself. There is a `checkNonCyclic` method in `ExtensionMethods` that is supposed to do this but is currently disabled, try enabling it:
https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/transform/ExtensionMethods.scala#L148-L150

liufengyun: Thanks for the catch @smarter , I'm going to enable that.

src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -135,7 +135,7 @@ class TreeChecker extends Phase with SymTransformer {
     }
   }
 
-  class Checker(phasesToCheck: Seq[Phase]) extends ReTyper {
+  class Checker(phasesToCheck: Seq[Phase]) extends ReTyper with NoChecking {
```
odersky: That's a drastic change. Why disable all checking in TreeChecker?

If it's just checkDerivedValue class that should be disabled during tree checking then it's better to 
override that method directly.

odersky:
Otherwise LGTM

src/dotty/tools/dotc/transform/TreeChecker.scala
liufengyun: @odersky this is now addressed in the latest commit.

odersky:
LGTM

PR #1685, opened 2016-11-08
Fix #1665: Check that != has an operand on the left.

### LINKED ISSUE #1665: crash on compilable code: scala.MatchError: Ident...

$ cat abc.scala
```
import scala.reflect.runtime.universe._
import scala.reflect.runtime.{currentMirror => cm}
object Test extends App {
  val mutant = new { val x = 2 }
  val c = cm.classSymbol(mutant.getClass)
  !=(c)
  println(c.fullName)
  c.info.toString.lines
    .filter(_ != "  private var bitmap$init$0: Boolean") foreach println
}
```

$ dotc abc.scala
```
exception while transforming Test.!=(Test.c) of class class dotty.tools.dotc.ast.Trees$Apply # 285
exception while transforming () extends Object() with App {
  val mutant: Object =
    {
      final class $anon() extends Object() {
        val x: Int = 2
      }
      new Object{...}(): Object
    }
  val c: scala.reflect.runtime.package.universe.ClassSymbol =
    scala.reflect.runtime.currentMirror.classSymbol(Test.mutant.getClass())
  Test.!=(Test.c)
  println(Test.c.fullName)
  augmentString(Test.c.info.toString()).lines.filter(
    {
      def $anonfun(_$1: String): Boolean = _$1.!=("  private var bitmap$init$0: Boolean")
      closure($anonfun)
    }
  ).foreach[Unit^](
    {
      def $anonfun(x: Any): Unit^ = println(x)
      closure($anonfun)
    }
  )
} of class class dotty.tools.dotc.ast.Trees$Template # 306
exception while transforming @scala.annotation.internal.SourceFile("abc.scala") final module class Test$() extends Object() with App {
  val mutant: Object =
    {
      final class $anon() extends Object() {
        val x: Int = 2
      }
      new Object{...}(): Object
    }
  val c: scala.reflect.runtime.package.universe.ClassSymbol =
    scala.reflect.runtime.currentMirror.classSymbol(Test.mutant.getClass())
  Test.!=(Test.c)
  println(Test.c.fullName)
  augmentString(Test.c.info.toString()).lines.filter(
    {
      def $anonfun(_$1: String): Boolean = _$1.!=("  private var bitmap$init$0: Boolean")
      closure($anonfun)
    }
  ).foreach[Unit^](
    {
      def $anonfun(x: Any): Unit^ = println(x)
      closure($anonfun)
    }
  )
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 307
  @scala.annotation.internal.SourceFile("abc.scala") final module class Test$() extends Object() with App {
    val mutant: Object =
      {
        final class $anon() extends Object() {
          val x: Int = 2
        }
        new Object{...}(): Object
      }
    val c: scala.reflect.runtime.package.universe.ClassSymbol =
      scala.reflect.runtime.currentMirror.classSymbol(Test.mutant.getClass())
    Test.!=(Test.c)
    println(Test.c.fullName)
    augmentString(Test.c.info.toString()).lines.filter(
      {
        def $anonfun(_$1: String): Boolean = _$1.!=("  private var bitmap$init$0: Boolean")
        closure($anonfun)
      }
    ).foreach[Unit^](
      {
        def $anonfun(x: Any): Unit^ = println(x)
        closure($anonfun)
      }
    )
  }
} of class class dotty.tools.dotc.ast.Trees$PackageDef # 308

Exception in thread "main" scala.MatchError: Ident($bang$eq) (of class dotty.tools.dotc.ast.Trees$Ident)
        at dotty.tools.dotc.transform.InterceptedMethods.qual$lzycompute$1(InterceptedMethods.scala:99)
        at dotty.tools.dotc.transform.InterceptedMethods.qual$1(InterceptedMethods.scala:99)
        at dotty.tools.dotc.transform.InterceptedMethods.transformApply(InterceptedMethods.scala:106)
        at dotty.tools.dotc.transform.TreeTransforms$TreeTransformer.goApply(TreeTransform.scala:602)
        at dotty.tools.dotc.transform.TreeTransforms$TreeTransformer.transformUnnamed(TreeTransform.scala:1005)
        at dotty.tools.dotc.transform.TreeTransforms$TreeTransformer$$anonfun$transform$2.apply(TreeTransform.scala:1179)
        at dotty.tools.dotc.transform.TreeTransforms$TreeTransformer$$anonfun$transform$2.apply(TreeTransform.scala:1171)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:132)
<snipped>
```

nicolasstucki:


tests/neg/i1665-nonfuzzy.scala
```diff
@@ -0,0 +1,5 @@
+
+object Test {
+  !=(1) // error: Binary operation != should have an operand on the left hand side
```
DarkDimius: why? this is supported by scalac.

src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -96,14 +96,22 @@ class InterceptedMethods extends MiniPhaseTransform {
         s"that means the intercepted methods set doesn't match the code")
       tree
     }
-    lazy val Select(qual, _) = tree.fun
+    lazy val qual = tree.fun match {
+      case Select(qual, _) => qual
+      case Ident(name) => EmptyTree
```
DarkDimius: this should also handle idents. Here's an example illustrating how to do it: https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/backend/jvm/DottyBackendInterface.scala#L428-L438

tests/neg/i1665-nonfuzzy.scala
nicolasstucki: In scalac there is a warning and then it is converted to `true`

```
<console>:12: warning: comparing values of types type and Int using `!=' will always yield true
       !=(1)
         ^
res0: Boolean = true
```
I could:
* Make it a warning and collapse it to `true` if in Scala 2 mode.
* Or make it a warning and collapse it to `true` in all cases.

src/dotty/tools/dotc/transform/InterceptedMethods.scala
nicolasstucki: I doubt that doing this will keep the semantics stated in the last comment.

src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -96,14 +96,31 @@ class InterceptedMethods extends MiniPhaseTransform {
...
     val Any_!= = defn.Any_!=
     val rewrite: Tree = tree.fun.symbol match {
       case Any_## =>
           poundPoundValue(qual)
       case Any_!= =>
-          qual.select(defn.Any_==).appliedToArgs(tree.args).select(defn.Boolean_!)
+        assert(tree.args.size == 1)
+        val lhs = qual.tpe
+        val rhs = tree.args.head.tpe
+        if (!(lhs <:< rhs || rhs <:< lhs) && rhs != defn.NullType && lhs != defn.NullType) {
```
nicolasstucki: Here if I don't add `rhs != defn.NullType && lhs != defn.NullType` I start getting failures when comparing `java.lang.Object` some other types including types in `scala.collection`. 
@DarkDimius I wonder if this is a bug or am I testing it the wrong way?

DarkDimius: I'm not sure I understood what failures did you get. Could you please elaborate?

nicolasstucki: For example if I run `partest`without the `rhs != defn.NullType && lhs != defn.NullType` I get the following two warnings (with `partest-only-no-bootsrtap` I only get the second one):

```
[info] Test test.TestBCode.nullChecks started
-- Warning: 
class Foo {
  def foo(x: AnyRef): Int = {
    val bool = x == null
    if (x != null) 1
    else 0
  }
}
                  
5 |    if (x != null) 1
  |       ^^^^^^^^^^^
  |       comparing values of types Object and Null using `!=' will always yield true
```

```
[info] Test dotc.tests.repl_all started
input differs from transcript:
scala> val Const,x = 0
Const: Int = 0
x: Int = 0
scala> val (Const, List(`x`, _, a), b) = (0, List(0, 1337, 1), 2)
-- Warning: <console> --------------------------------------------------------------------------------------------------
6 |val (Const, List(`x`, _, a), b) = (0, List(0, 1337, 1), 2)
  |                 ^
  |         comparing values of types scala.collection.immutable.List[Int] and Null using `!=' will always yield true
a: Int = 1
b: Int = 2
scala> val a@b = 0
a: Int @unchecked = 0
b: Int @unchecked = 0
scala> :quit
```

Are those non nullable `AnyRef`s?

DarkDimius: I think this is a bug in `IsInstanceOfEvaluator`.

odersky:
We should have a discussion whether we need IsInstanceOfEvaluator or whether this can be rolled into multiversal equality checks.

src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -96,14 +96,24 @@ class InterceptedMethods extends MiniPhaseTransform {
         s"that means the intercepted methods set doesn't match the code")
       tree
     }
-    lazy val Select(qual, _) = tree.fun
+    lazy val qual = tree.fun match {
+      case Select(qual, _) => qual
+      case ident @ Ident(_) =>
+        ident.tpe match {
+          case TermRef(prefix: TermRef, name) =>
```
DarkDimius: name isn't used in either of two cases.

tests/pos/i1665.scala
```diff
@@ -0,0 +1,12 @@
+
+import scala.reflect.runtime.universe._
+import scala.reflect.runtime.{currentMirror => cm}
+
+object Test extends App {
+  val mutant = new { val x = 2 }
+  val c = cm.classSymbol(mutant.getClass)
```
DarkDimius: why does this test need _scala2_ reflection?

src/dotty/tools/dotc/transform/InterceptedMethods.scala
nicolasstucki: Removed

tests/pos/i1665.scala
nicolasstucki: Just because it is the code that was in issue #1665.

felixmulder: @nicolasstucki - could you update this test to use Java reflection instead? Cheers from Dima

nicolasstucki: I removed the test as the other test file contains all the tests needed.

DarkDimius:
LGTM

PR #1689, opened 2016-11-09
Fix #1688: don't allow override of synthetic or primitive classes

### LINKED ISSUE #1688: crash on compilable code: assertion failed: Cannot create ClassBType for phantom type Lscala/Null

$ cat abc.scala
```
package scala
sealed trait Null
```

$ dotc abc.scala
```
java.lang.AssertionError: assertion failed: Cannot create ClassBType for phantom type Lscala/Null;
	at scala.tools.nsc.backend.jvm.BTypes$ClassBType.checkInfoConsistency(BTypes.scala:620)
	at scala.tools.nsc.backend.jvm.BTypes$ClassBType.info_$eq(BTypes.scala:607)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.scala$tools$nsc$backend$jvm$BTypesFromSymbols$$setClassInfo(BTypesFromSymbols.scala:142)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols$$anonfun$classBTypeFromSymbol$4.apply(BTypesFromSymbols.scala:63)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols$$anonfun$classBTypeFromSymbol$4.apply(BTypesFromSymbols.scala:57)
	at scala.collection.MapLike$class.getOrElse(MapLike.scala:128)
	at scala.collection.AbstractMap.getOrElse(Map.scala:59)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.classBTypeFromSymbol(BTypesFromSymbols.scala:57)
	at scala.tools.nsc.backend.jvm.BCodeHelpers$BCInnerClassGen$class.getClassBTypeAndRegisterInnerClass(BCodeHelpers.scala:243)
<snipped>
```

felixmulder:
review by: @DarkDimius

### DIFF
--- a/src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
+++ b/src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
@@ -1,0 +1,41 @@
+package dotty.tools.dotc
+package transform
+
+import TreeTransforms._
+import core._
+import Contexts._, Definitions._, Decorators._
+
+/** Checks if the user is trying to illegally override synthetic classes,
+  * primitives or their boxed equivalent
+  */
+class CheckIllegalOverrides extends MiniPhaseTransform {
+  import ast.tpd._
+
+  override def phaseName = "checkPhantom"
+
+  override def transformTypeDef(tree: TypeDef)(implicit ctx: Context, info: TransformerInfo) = {
+    val defn = ctx.definitions
+    val tpe = tree.tpe
+    val fullName = tree.symbol.showFullName
+
+    val syntheticClasses = defn.syntheticCoreClasses.map(_.showFullName)
+    val primitives = defn.ScalaValueClasses().map(_.showFullName)
+    val boxedPrimitives = defn.ScalaBoxedClasses().map(_.showFullName)
+
+    val illegalOverride =
+      if (syntheticClasses.contains(fullName))
+        Some("synthetic")
+      else if (primitives.contains(fullName))
+        Some("primitive")
+      else if (boxedPrimitives.contains(fullName))
+        Some("boxed primitive")
+      else
+        None
+
+    illegalOverride.foreach { overriden =>
+      ctx.error(i"Cannot define symbol overriding $overriden class `$tpe`", tree.pos)
+    }
+
+    tree
+  }
+}
--- a/src/dotty/tools/dotc/Compiler.scala
+++ b/src/dotty/tools/dotc/Compiler.scala
@@ -50,6 +50,7 @@ class Compiler {
            new CheckReentrant),     // Internal use only: Check that compiled program has no data races involving global vars
       List(new RefChecks,           // Various checks mostly related to abstract members and overriding
            new CheckStatic,         // Check restrictions that apply to @static members
+           new CheckIllegalOverrides, // Check restrictions that apply to phantom types and value classes
            new ElimRepeated,        // Rewrite vararg parameters and arguments
            new NormalizeFlags,      // Rewrite some definition flags
            new ExtensionMethods,    // Expand methods of value classes with extension methods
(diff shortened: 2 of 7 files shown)


src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
```diff
@@ -0,0 +1,41 @@
...
+package transform
+
+import TreeTransforms._
+import core._
+import Contexts._, Definitions._, Decorators._
+
+/** Checks if the user is trying to illegally override synthetic classes,
+  * primitives or their boxed equivalent
+  */
+class CheckIllegalOverrides extends MiniPhaseTransform {
```
DarkDimius: I wouldn't call this an override, it's more of a re-definition. But I'm also fine with the current name.

src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
```diff
@@ -0,0 +1,41 @@
...
+  import ast.tpd._
+
+  override def phaseName = "checkPhantom"
```
DarkDimius: It would be nice to have phaseName and class name in sync.

src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  override def transformTypeDef(tree: TypeDef)(implicit ctx: Context, info: TransformerInfo) = {
```
DarkDimius: I'd propose to slightly change the way such checks are done. 
If you make this be a mini-phase transform, the other mini-phases in the same block couldn't rely on the facts that you've checking, as they run interleaved with this phase.

As you see, your implementation only needs the symbol and this means that you can make it into an SymTransformer which checks the invariants. Checking in a SymTransformer enforces that even phases in the same mini-phase block that are after you may rely on invariants that you introduce.

src/dotty/tools/dotc/transform/CheckIllegalOverrides.scala
```diff
@@ -0,0 +1,41 @@
...
+    val illegalOverride =
+      if (syntheticClasses.contains(fullName))
+        Some("synthetic")
+      else if (primitives.contains(fullName))
+        Some("primitive")
+      else if (boxedPrimitives.contains(fullName))
+        Some("boxed primitive")
+      else
+        None
+
```
DarkDimius: It would be also nice to check that no-one tries to introduce a companion to those classes.
Eg, there actually is a class `Nothing$` that isn't source defined but backend relies on it existing.

felixmulder: Since we're comparing the full paths of the symbols we get this for free, but I added a neg test `i1688c.scala` just to make sure.

felixmulder:
@DarkDimius - addressed your feedback :)

odersky:
Also, we can't forbid defining primitive value classes here because we do define them when we bootstrap!

smarter:
I think the correct way to fix #1688 is to change the backend to issue an error and gracefully exit instead of asserting when trying to define `scala.Null` or `scala.Nothing`

odersky:
@smarter Or change the frontend to disallow definitions of such classes. Which might prevent other possible crashes.

felixmulder:
(Already said this in part on a commit, the emailed link sent me there - reiterating:)

@smarter - IIRC when I was looking at this, changing the backend required changing the shared backend that we're depending on from artefact.

I could change the `enterSymbol` function to take this into account by not allowing to enter symbols of the same primitives / boxed primitives. WDYT?

> Also, we can't forbid defining primitive value classes here because we do define them when we bootstrap!

I thought `FrontEnd` threw these out in `discardAfterTyper`? Or are we defining them somewhere else?

odersky:
> I thought FrontEnd threw these out in discardAfterTyper? Or are we defining them somewhere else?

Ah, yes, I had not thought of that. But that means a test later on is redundant, right? I.e. if it fires it would be a (weird) internal compiler error.

odersky:
`enterSymbol` is a possibility but I think it's even better to do it in Desugar. Let me close this and do a new PR with the idea.

felixmulder:
IIUC if it fires, it just means that the class that got through with the fully qualified name `scala.Char` was not a value class :)

PR #1690, opened 2016-11-09
Fix benchmarks and add multiple mini benchmark tests

liufengyun:
The benchmarks has been down due to some weird error. A single line change fix it: [link](https://github.com/lampepfl/dotty/compare/master...dotty-staging:bench?expand=1#diff-28f8d722cae3da74d903fc63040c9d36R68)

```
-  @Test def pickle_pickleOK = compileDir(testsDir, "pickling", testPickling)
+  @Test def pickle_pickleOK = compileFiles(testsDir + "pickling/", testPickling)
```

And I've added several mini tests which covers implicit search, exhaustivity check, tailrec, etc. We can just put any mini test to the folder `bench/tests`, it will be automatically tracked in the benchmarks.


liufengyun:
Review @felixmulder and @DarkDimius ?

bench/src/test/scala/Benchmarks.scala
```diff
@@ -74,11 +76,24 @@ object BenchTests extends OnlineRegressionReport {
   def setup =
     performance of "dotty" in {
       measure.method("stdlib") in {
-        using(Gen.unit("test")) curve "stdlib" in { r => stdLib }
+        // maybe scalac curve later
+        using(Gen.unit("test")) curve "dotty" in { r => stdLib }
```
liufengyun: @DarkDimius The `curve` is to support multiple curves in a chart. Yes, I should keep it the same for `stdlib` and `dotty-src` for backward data compatibility.

liufengyun: @DarkDimius the name for `stdlib` and `dotty-src` are now restored.

odersky:
Status?

liufengyun:
@DarkDimius  This is one is up to you to review, thanks :)

DarkDimius:
LGTM

PR #1695, opened 2016-11-10
Fix #1662: Add missing case for singleton

### LINKED ISSUE #1662: crash with scala.MatchError: Skolem...
$ cat abc.scala
```
class Lift {
  def apply(f: F0)
  class F0
  object F0 { implicit def f2f0(String): F0 = ??? }
  (new Lift)("")
}
```

$ dotc abc.scala
```
-- [E018] Syntax Error: abc.scala --------------------------------------------------------------------------------------
3 |  def apply(f: F0)
  |                  ^
  |                  missing return type

-- Error: abc.scala ----------------------------------------------------------------------------------------------------
5 |  object F0 { implicit def f2f0(String): F0 = ??? }
  |                                      ^
  |                                      ':' expected but ')' found

Exception in thread "main" scala.MatchError: Skolem(529116035) (of class dotty.tools.dotc.core.Types$CachedSkolemType)
        at dotty.tools.dotc.ast.tpd$.singleton(tpd.scala:343)
        at dotty.tools.dotc.ast.tpd$.ref(tpd.scala:328)
        at dotty.tools.dotc.ast.tpd$.singleton(tpd.scala:344)
        at dotty.tools.dotc.ast.tpd$.ref(tpd.scala:328)
        at dotty.tools.dotc.typer.Implicits$ImplicitSearch$$anonfun$typedImplicit$1$1$$anonfun$apply$26.apply(Implicits.scala:602)
        at dotty.tools.dotc.typer.Implicits$ImplicitSearch$$anonfun$typedImplicit$1$1$$anonfun$apply$26.apply(Implicits.scala:600)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:132)
```

odersky:
Review by @liufengyun

### DIFF
--- a/src/dotty/tools/dotc/ast/tpd.scala
+++ b/src/dotty/tools/dotc/ast/tpd.scala
@@ -306,8 +306,8 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
       case _ =>
         false
     }
-    typeIsElidable || 
-    tp.symbol.is(JavaStatic) || 
+    typeIsElidable ||
+    tp.symbol.is(JavaStatic) ||
     tp.symbol.hasAnnotation(defn.ScalaStaticAnnot)
   }
 
@@ -343,6 +343,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   def singleton(tp: Type)(implicit ctx: Context): Tree = tp match {
     case tp: TermRef => ref(tp)
     case tp: ThisType => This(tp.cls)
+    case tp: SkolemType => singleton(tp.narrow)
     case SuperType(qual, _) => singleton(qual)
     case ConstantType(value) => Literal(value)
   }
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -343,6 +343,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   def singleton(tp: Type)(implicit ctx: Context): Tree = tp match {
     case tp: TermRef => ref(tp)
     case tp: ThisType => This(tp.cls)
+    case tp: SkolemType => singleton(tp.narrow)
     case SuperType(qual, _) => singleton(qual)
```
liufengyun: I thought `SkolemType` is only used for existential types, but it seems not to be the case. Could you briefly explain a little bit the usage of `SkolemType`, or add a comment to `class SkolemType`?

`RecThis` is also a `SingletonType`, I guess it can be safely ignored here.

Issue #1692, opened 2016-11-10
By-name params used only in a lazy field should be cleared to prevent leaks


Blaisorblade:
Noticed in #1691 — scalac, but not dotty (as pointed out by @smarter), compiles the following so the initializer for `byLazyValMsg` clears `byNameMsg`. This is *not* just an optimization, but is needed for correctness to prevent memory leaks:

```scala
class TestByNameLazy(byNameMsg: => String) {
  lazy val byLazyValMsg = byNameMsg
}
```

Resulting code with Scalac (as decompiled with jad)—note `byName = null;`:

```java
public class TestByNameLazy
{

    private String byLVal$lzycompute()
    {
        synchronized(this)
        {
            if(!bitmap$0)
            {
                byLVal = (String)byName.apply();
                bitmap$0 = true;
            }
            BoxedUnit _tmp = BoxedUnit.UNIT;
        }
        byName = null;
        return byLVal;
    }

    public String byLVal()
    {
        return bitmap$0 ? byLVal : byLVal$lzycompute();
    }

    public TestByNameLazy(Function0 byName)
    {
        this.byName = byName;
        super();
    }

    private final Function0 byName;
    private String byLVal;
    private volatile boolean bitmap$0;
}
```

> Is this documented anywhere?

Maybe but not to my knowledge; I've seen similar patterns often enough and Dragos described this in https://github.com/scala/scala/pull/922 as:

> Less known fact: lazy values null-out their dependent values is they're accessed only from
their initializer.

I found that PR by searching JIRA https://issues.scala-lang.org/browse/SI-6092 (which is not about this "feature" missing altogether, just about a bug with it).

odersky:
@DarkDimius Let's discuss how this can be best achieved. Looks not easy to me.

DarkDimius:
This can be achived by: having a pass before lazy vals that marks that `private val byNameMsg` is only used inside this lazy val and making `lazyVals` null out the private function field in such cases.

allanrenucci:
Needed to make `collections-strawman` tests pass

PR #1696, opened 2016-11-10
Make sure messages are lazily evaluated until `report` in `Reporter`

felixmulder:
In this PR I make sure that messages are lazily evaluated to the same point in the code it was previously. Namely in `Reporter.scala` in the method `Reporter#report`. The message is forced within this method, WDYT - should it be made lazy as well?

Review by @odersky

### DIFF
--- a/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -99,7 +99,7 @@ object Parsers {
     /** Issue an error at given offset if beyond last error offset
       *  and update lastErrorOffset.
       */
-    def syntaxError(msg: Message, offset: Int = in.offset): Unit =
+    def syntaxError(msg: => Message, offset: Int = in.offset): Unit =
       if (offset > lastErrorOffset) {
         syntaxError(msg, Position(offset))
         lastErrorOffset = in.offset
@@ -108,7 +108,7 @@ object Parsers {
     /** Unconditionally issue an error at given position, without
       *  updating lastErrorOffset.
       */
-    def syntaxError(msg: Message, pos: Position): Unit =
+    def syntaxError(msg: => Message, pos: Position): Unit =
       ctx.error(msg, source atPos pos)
 
   }
@@ -215,23 +215,23 @@ object Parsers {
       }
     }
 
-    def warning(msg: Message, sourcePos: SourcePosition) =
+    def warning(msg: => Message, sourcePos: SourcePosition) =
       ctx.warning(msg, sourcePos)
 
-    def warning(msg: Message, offset: Int = in.offset) =
+    def warning(msg: => Message, offset: Int = in.offset) =
       ctx.warning(msg, source atPos Position(offset))
 
-    def deprecationWarning(msg: Message, offset: Int = in.offset) =
+    def deprecationWarning(msg: => Message, offset: Int = in.offset) =
       ctx.deprecationWarning(msg, source atPos Position(offset))
 
     /** Issue an error at current offset taht input is incomplete */
-    def incompleteInputError(msg: Message) =
+    def incompleteInputError(msg: => Message) =
       ctx.incompleteInputError(msg, source atPos Position(in.offset))
 
     /** If at end of file, issue an incompleteInputError.
      *  Otherwise issue a syntax error and skip to next safe point.
      */
-    def syntaxErrorOrIncomplete(msg: Message) =
+   def syntaxErrorOrIncomplete(msg: => Message) =
       if (in.token == EOF) incompleteInputError(msg)
       else {
         syntaxError(msg)
--- a/src/dotty/tools/dotc/reporting/Reporter.scala
+++ b/src/dotty/tools/dotc/reporting/Reporter.scala
@@ -38,16 +38,16 @@ trait Reporting { this: Context =>
     reporter.report(new Info(msg, pos))
 
   def deprecationWarning(msg: => Message, pos: SourcePosition = NoSourcePosition): Unit =
-    reporter.report(msg.deprecationWarning(pos))
+    reporter.report(new DeprecationWarning(msg, pos))
 
   def migrationWarning(msg: => Message, pos: SourcePosition = NoSourcePosition): Unit =
-    reporter.report(msg.migrationWarning(pos))
+    reporter.report(new MigrationWarning(msg, pos))
 
   def uncheckedWarning(msg: => Message, pos: SourcePosition = NoSourcePosition): Unit =
-    reporter.report(msg.uncheckedWarning(pos))
+    reporter.report(new UncheckedWarning(msg, pos))
 
   def featureWarning(msg: => Message, pos: SourcePosition = NoSourcePosition): Unit =
-    reporter.report(msg.featureWarning(pos))
+    reporter.report(new FeatureWarning(msg, pos))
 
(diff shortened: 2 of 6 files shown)


felixmulder:
The PR also adds a `persist` method to `Message` in order to allow for persisting messages without also accidentally persisting contexts resulting in memory leaks :+1:

felixmulder:
Also @Blaisorblade - if you're interested in having a look I'd appreciate it :)

odersky:
I am afraid we should not evaluate on `report` either. Only leaf reporters (e.g. ConsoleReporter and ThrowingReporter) are allowed to force this. Previously this was achieved by making the message string a lazy val in Diagnostics.

felixmulder:
Sounds reasonable - I'll update the PR with your feedback - thanks Martin.

test/dotty/tools/dotc/reporting/TestMessageLaziness.scala
```diff
@@ -0,0 +1,37 @@
...
+  ctx = ctx.fresh.setReporter(new NonchalantReporter)
+
+  class NonchalantReporter(implicit ctx: Context) extends Reporter
+  with UniqueMessagePositions with HideNonSensicalMessages {
+    def doReport(d: MessageContainer)(implicit ctx: Context) = ???
+
+    override def report(d: => MessageContainer)(implicit ctx: Context) = ()
+  }
+
+  case class LazyError() extends Message(1000) {
```
Blaisorblade: 👍

src/dotty/tools/dotc/reporting/diagnostic/Message.scala
```diff
@@ -47,45 +56,62 @@ abstract class Message(val errorId: Int) { self =>
...
+  }
+}
+
+/** An extended message keeps the contained message from being evaluated, while
+  * allowing for extension for the `msg` string
+  *
+  * This is useful when we need to add additional information to an existing
+  * message.
+  */
+class ExtendMessage(_msg: () => Message)(f: String => String) { self =>
```
Blaisorblade: Maybe `MappedMessage` or `TransformedMessage` would be even clearer.

src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -236,7 +240,7 @@ abstract class Reporter extends interfaces.ReporterResult {
     override def default(key: String) = 0
   }
 
-  def report(d: MessageContainer)(implicit ctx: Context): Unit =
+  def report(d: => MessageContainer)(implicit ctx: Context): Unit =
```
Blaisorblade: 1. `MessageContainer` still takes messages by-name so I doubt it needs to be by-name itself.
2. If there's some reason I'm missing (totally possible), other uses of `MessageContainer` would have to also become by-name—starting from `isHidden`. A test might be harder, so one would have to grep for uses of `MessageContainer`.

src/dotty/tools/dotc/reporting/diagnostic/Message.scala
felixmulder: Sure :)

src/dotty/tools/dotc/reporting/Reporter.scala
felixmulder: I think I'll investigate this with...more tests! :open_mouth:

Blaisorblade:
_IIUC_ `Reporter#report` does not force the Message thunk itself, even if the PR description says so—that's delegated to `doReport`, so the PR might still pass @odersky's comment.

To know I inspected overrides of `doReport`, and found two problematic cases:
- mixin `HideNonSensicalMessages` forces the message to check `isNonSensical`, hence so does `Reporter.fromSimpleReporter` (which is just a forwarder). I'm not sure that's legal.
- `StoreReporter` might force the message in debug mode (if `Printers.typr` is changed, but the logic isn't documented), but it violates the new invariants on leaks — if it can't call `persist`, it needs to be documented as "leaky" and used with care (only held as long as the context is alive anyway).

Conversely, `ConsoleReporter` forces `Message` as allowed, and `ThrowingReporter` does not force `Message` itself.

This sort of complicated invariant is why I advocated dynamic checking via assertions, at least in some (existing?) debug mode (beyond the testcase, which are a good step): even with your new ScalaDocs on leaks, we need to trust reviewers to maintain a non-local invariant. Some other solution might work, but enforcing the invariant isn't.

Also, to prevent leaks, `MessageContainer.contained` should probably still clear `msgFn` (so it should be a var, or `contained` could be a lazy val once #1692 is fixed), and `contained` should store the `persist`ed message. Or does the PR address this somehow? Otherwise, holding onto a `MessageContainer` will also leak contexts.

felixmulder:
Just to update this - I've been hacking on this throughout the day. But I'm not ready to update the PR yet, still have a few kinks to work out.

src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -99,7 +99,7 @@ object Parsers {
     /** Issue an error at given offset if beyond last error offset
       *  and update lastErrorOffset.
       */
-    def syntaxError(msg: Message, offset: Int = in.offset): Unit =
+    def syntaxError(msg: => Message, offset: Int = in.offset): Unit =
```
Blaisorblade: 👍 to this whole commit (without having rechecked other occurrences of `Message`).

Blaisorblade: (Note that questions on `StoreReporter` and so on are still probably on, but all that commit does makes sense).

felixmulder: Yes, the store reporter will force if debugging - but I think we can live with that :)

Blaisorblade: FWIW, IMHO we can live with sharp edges as long as we document them when we notice them, and you already added some comments in this direction so 👍.
I'm just asking for a warning comment on the store reporter lest people use it in other scenarios, as I think Murphy's law applies to software evolution. Copy-pasting this would be enough, if drafting the text is the problem:
```scala
/** Beware this can leak memory, see https://github.com/lampepfl/dotty/pull/1696#issuecomment-259739205 */
```

Blaisorblade: > But sure, maybe it could use another comment :)

👍 . "Private to the compiler" doesn't help compiler contributors, neither core ones nor contributors outside EPFL. I also had no clue `StoreReporter` is for debugging only.

I'm also confused by the pushback — ideally one documents more pitfalls, not fewer (or even better, makes the comments unnecessary by removing them).

I'm aware there are bad comments, and double-checked checklists of comment downsides (on http://wiki.c2.com/?CommentCostsAndBenefits)—and I'm not sure I see a downside here (other than "time to write", which is why I offered text).

felixmulder: Yeah, didn't mean to push back - but at home right now ^^. The PR has been updated to document this in a2354dd :+1:

Blaisorblade: > but at home right now ^^

Oh sorry didn't mean to interrupt. Only time bound I care for: comments are best addressed before the issue/PR is closed, or they should be moved to a new issue lest they're forgot.

odersky:
LGTM 👍

PR #1698, opened 2016-11-10
Add an error message for incorrect sequence wildcard position Parsers.scala:1425 and Parsers.scala:1079

maseev:
This pull request is related to the following [issue](https://github.com/lampepfl/dotty/issues/1589).

### DIFF
--- a/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -789,4 +789,27 @@ object messages {
     val msg = "unreachable code"
     val explanation = ""
   }
+
+  case class SeqWildcardPatternPos()(implicit ctx: Context) extends Message(31) {
+    val kind = "Syntax"
+    val msg = "`_*' can be used only for last argument"
+    val explanation = {
+      val code =
+        """def sumOfTheFirstTwo(list: List[Int]): Int = list match {
+          |  case List(first, second, x:_*) => first + second
+          |  case _ => 0
+          |}"""
+      hl"""|Sequence wildcard pattern is expected at the end of an argument list.
+           |This pattern matches any remaining elements in a sequence.
+           |Consider the following example:
+           |
+           |$code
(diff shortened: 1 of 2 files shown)


src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -787,4 +787,24 @@ object messages {
     val msg = "unreachable code"
     val explanation = ""
   }
+
+  case class IncorrectSequenceWildcardPatternPosition()(implicit ctx: Context) extends Message(31) {
```
felixmulder: Could we make this name a little shorter? How about: `SeqWildcardPatternPos`?

src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -787,4 +787,24 @@ object messages {
...
+    val msg = "`_*' can be used only for last argument"
+    val explanation =
+      hl"""|Sequence wildcard pattern is expected at the end of an argument list.
+           |This pattern matches any remaining elements in a sequence.
+           |Consider the following example:
+           |
+           |def sumOfTheFirstTwo(list: List[Int]): Int = list match {
+           |  case List(first, second, x:_*) => first + second
+           |  case _ => 0
+           |}
```
felixmulder: To get highlighting on this part pull it out to a variable like:

```scala
val code = "def sumOfThe..."

hl"""|Sequence wildcard pattern is expected at the end of an argument list.
     |This pattern matches any remaining elements in a sequence.
     |Consider the following example:
     |
     |$code
     |
     |..."""
```

And for the code under `Calling:`

interpolate it like:

```
|Calling:
|
|${"sumOfTheFirstTwo(List(1, 2, 10))"}
|
|would give 3 as a result"""
```

this will highlight the code.

maseev:
@felixmulder PTAL

felixmulder:
LGTM thanks!

Issue #1683, opened 2016-11-08
Closures parsed as self-type annotation


Blaisorblade:
I'm not sure I care and I can't conceive a use case (at least, not without involving some weird macro), but the following is valid for scalac and rejected (at parse-time) by dotty:

```scala
class Foo {
  (a: Int, b: String) => a
}
```

The reason is that dotty parses (what looks like) an expression, then sees `=>` and "reinterprets" the parsed expression as a self-type annotation.

I'm not so concerned about the feature, I'm more concerned about aligning the grammar with the implementation.

I noticed this while trying to document errors for self-type annotations.

So this gives something like:
```
> run -explain tests/Foo.scala
[...]
[info] Running dotty.tools.dotc.Main -explain tests/Foo.scala
-- Error: tests/Foo.scala ----------------------------------------------------------------------------------------------
2 |  (a: Int, b: String) => a
  |  ^^^^^^^^^^^^^^^^^^^
  |  not a legal self type clause
-- [E006] Missing Identifier Error: tests/Foo.scala --------------------------------------------------------------------
2 |  (a: Int, b: String) => a
  |                         ^
  |                         not found: a

Explanation
===========
An identifier for `a` is missing. This means that something
has either been misspelt or you're forgetting an import


two errors found
java.lang.RuntimeException: Nonzero exit code returned from runner: 1
```

BTW I don't like "missing identifier", I think it's trying to say "unbound identifier" (I have a local patch I'll send at some point).

liufengyun:
Thanks for reporting @Blaisorblade . This is related to #1424.

Unlike #1424, it seems the parser can do a better job here, as it's possible to know syntactically it's not a self-annotation, but a lambda def.

Blaisorblade:
Thanks I missed that! Also Martin there claims that handling this in Scalac is very tricky, so maybe it shouldn't be fixed. Making sure the grammar rejects this might be enough.

odersky:
> BTW I don't like "missing identifier", I think it's trying to say "unbound identifier" (I have a local patch I'll send at some point).

I don't like that either. But "unbound" is probably too academic for some. How about

    No definition for `a` was found.

?

odersky:
I forgot to say: "Unbound Identifier" looks fine as a name for this class of error messages. But the explanation should say something else.

odersky:
I think we should not try to overengineer fixing this. The scalac code doing this is quite impenetrable. I think "not a legal self type clause" is perfectly fine as an error message.

[...]

PR #1699, opened 2016-11-11
Clarify error for unbound identifiers and add/fix a few comments

Blaisorblade:
I think "Missing identifier" is not a good error, @odersky agreed and suggested "No definition for `a` was found." (https://github.com/lampepfl/dotty/issues/1683#issuecomment-259480649), and I replaced "definition" with "declaration".

I also tried improving the explanation.

Also a few other minor issues in separate commits (I'd PR them separately but IIUC that'd be noise to you).


odersky:
LGTM, thanks!

PR #1705, opened 2016-11-13
Add colon after method type in printer

rethab:
Changes
```scala
scala> def id(x: 4): 4 = x
id: (x: Int(4))Int(4)
```
to
```scala
scala> def id(x: 4): 4 = x
id: (x: Int(4)): Int(4)
```

Contrary to the github markdown, the REPL would not highlight the method type w/o this change.

### DIFF
--- a/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -159,10 +159,14 @@ class PlainPrinter(_ctx: Context) extends Printer {
         "<noprefix>"
       case tp: MethodType =>
         def paramText(name: TermName, tp: Type) = toText(name) ~ ": " ~ toText(tp)
+        def typeColon(resultType: Type): Text = resultType match {
+          case _: TypeRef => ": "
+          case _ => "" // eg. methods with implicit parameters go here in first pass
+        }
         changePrec(GlobalPrec) {
           (if (tp.isImplicit) "(implicit " else "(") ~
             Text((tp.paramNames, tp.paramTypes).zipped map paramText, ", ") ~
-          ")" ~ toText(tp.resultType)
+          ")" ~ typeColon(tp.resultType) ~ toText(tp.resultType)
         }
       case tp: ExprType =>
         changePrec(GlobalPrec) { "=> " ~ toText(tp.resultType) }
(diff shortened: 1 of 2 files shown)


Blaisorblade:
Hi and thanks for the PR! If I may comment as an outsider, I think the type syntax itself is intended, and is even described in the language spec (http://scala-lang.org/files/archive/spec/2.12/03-types.html#method-types); maybe the highlighting should be changed.
The syntax for method types is hard to see, because they aren't value types (again see the spec) and IIUC can't appear in code as types of some value.

Your example doesn't work in Scala 2, but a variant does and gives a colon-less output in both REPLs. I can confirm that the `Int` in `(x: Int)Int` is not highlighted and that seems a bug a bug.
```scala
scala> def id(x: Int): Int = x
id: (x: Int)Int
```

smarter:
I think we should just print it like this:
```scala
scala> def id(x: Int): Int = x
def id(x: Int): Int
```
Anything else is confusing for users.

smarter:
...which is exactly what https://github.com/scala/scala/pull/5498 is about

rethab:
@Blaisorblade thanks for reference to the spec and your comments. That would mean we're simply missing the space. I just tested that and if there is a space, then the higlighting works correctly (the colon is not necessary for that).

@smarter consequently, should the same thing apply for `val`s? `var`s?

smarter:
> @smarter consequently, should the same thing apply for vals? vars? 

Yes.

rethab:
I have updated the code to print the following:
```scala
scala> def id(x: Int): Int = 10
def id: (x: Int): Int
scala> def idi(x: Int)(implicit y: Int): Int = x + y
def idi: (x: Int)(implicit y: Int): Int
scala> val a: Int = 10
val a: Int = 10
scala> var b: Int = 10
var b: Int = 10
```

@smarter What do you think?

smarter:
Looks nice! I think I would go further and print `def id(x: Int): Int` instead of `def id: (x: Int): Int`

rethab:
@smarter sorry I missed that in your example above. Fixed now!

smarter:
As the test failures indicate, you need to update the check files in https://github.com/lampepfl/dotty/tree/master/tests/repl to the new format (you can run just these tests locally using `test-only -- --tests=repl_all` from sbt)

felixmulder:
Completed in https://github.com/lampepfl/dotty/pull/1742, thanks @rethab !

PR #1711, opened 2016-11-15
Fix isNullable when `<:<` term refs.

nicolasstucki:
Note that without the fix console tests will fail and without
the warning there is no way to test the fix.

### DIFF
--- a/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -527,7 +527,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
           }
           isSubType(hi1, tp2) || compareGADT
         case _ =>
-          def isNullable(tp: Type): Boolean = tp.dealias match {
+          def isNullable(tp: Type): Boolean = tp.widenDealias match {
             case tp: TypeRef => tp.symbol.isNullableClass
             case tp: RefinedOrRecType => isNullable(tp.parent)
             case AndType(tp1, tp2) => isNullable(tp1) && isNullable(tp2)


src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -103,7 +103,13 @@ class InterceptedMethods extends MiniPhaseTransform {
...
           poundPoundValue(qual)
       case Any_!= =>
-          qual.select(defn.Any_==).appliedToArgs(tree.args).select(defn.Boolean_!)
+        val lhs = qual.tpe
+        val rhs = tree.args.head.tpe
+        if (!(lhs <:< rhs || rhs <:< lhs)) {
+          ctx.warning(ex"comparing values of types ${lhs.widenDealias} and ${rhs.widenDealias} using `!=' will always yield true",
+              tree.pos)
+        }
+        qual.select(defn.Any_==).appliedToArgs(tree.args).select(defn.Boolean_!)
```
odersky: Why put that test here? And why is it done only on != but not on ==?

nicolasstucki: Yes, the tests is incomplete, in scalac there is a different waring for those cases:
```
scala> x != y
<console>:18: warning: A and B are unrelated: they will most likely always compare unequal
       x != y
```

Now I'm convinced that these should not be a separate mechanism from multiversal equality.

There is also the question of the `NullType <:< X` tests failing for `X`s that are term refs to some subtype of `AnyRef`. In all the currently tests those term refs are widen first and therefore we never have false negatives. Is this the correct way to handle those cases or the change to `isNullabe` should be done?

odersky: I think the isNullable change is good. TermRefs should be nullable if their underlying typrs are nullable. So I propose to just keep this change and drop the rest.

nicolasstucki: I agree.

odersky:
LGTM 👍

PR #1719, opened 2016-11-16
Fix #1707: Survive non-existing positions in parser

### LINKED ISSUE #1707: crash on compilable code: at dotty.tools.dotc.util.Positions$Position$.start$extension
scalac can compile it but dotc crashes.

$ cat abc.scala
```
object DepBug {
  class A {
    class B
    def mkB = new B
    def m(b: B) = b
  }
  trait Dep {
    val a: A
    val b: a.B
  }
  val dep = new {
    val a = new A
    val b = a mkB
  }
  def useDep(d: Dep) {
    import d._
    a m (b)
  }
  {
    import dep._
    a m (b)
  }
  dep.a m (dep b)
}
```

$ dotc abc.scala
```
-- Error: abc.scala ----------------------------------------------------------------------------------------------------
15 |  def useDep(d: Dep) {
   |                     ^
   |                     '=' expected but '{' found
-- Error: abc.scala ----------------------------------------------------------------------------------------------------
19 |  {
   |  ^
   |  '}' expected but '{' found

exception occurred while parsing abc.scala

	at dotty.tools.dotc.util.Positions$Position$.start$extension(Positions.scala:40)
	at dotty.tools.dotc.parsing.Parsers$Parser.simpleExprRest(Parsers.scala:1199)
	at dotty.tools.dotc.parsing.Parsers$Parser.simpleExpr(Parsers.scala:1186)
	at dotty.tools.dotc.parsing.Parsers$Parser$$anonfun$8.apply(Parsers.scala:1137)
	at dotty.tools.dotc.parsing.Parsers$Parser$$anonfun$8.apply(Parsers.scala:1128)
	at dotty.tools.dotc.parsing.Parsers$Parser.postfixExpr(Parsers.scala:1124)
<snipped>
```

odersky:
Some trees, which do not consume input have unassigned positions (so that
they can fit in whatever range they are integrated). It's therefore risky
to take the start or end of a parsed tree's position. This commit guards
against the case where the position of the tree does not exist.

Fixes #1707. Review by @liufengyun


odersky:
Previous PR fell short; replaced with a scheme that's more general and thorough.

odersky:
Rebased to master to eliminate that pesky test failure.

rethab:
@odersky are you sure this fixes #1705 ? Seems rather unrelated

smarter:
Indeed, it's actually https://github.com/lampepfl/dotty/issues/1707, `neg/i1705.scala` should be renamed to `neg/i1707.scala`

odersky:
Test file renamed and rebased to master

liufengyun:
LGTM @odersky , changes are straight-forward.

PR #1720, opened 2016-11-16
fix #1701: disqualify package name in type name resolution

### LINKED ISSUE #1701: accepts invalid code with undefined type
``` bash
$ scalac-2.12.0 t.scala
t.scala:2: error: not found: type test
  def test (x : test) : Int
                ^
one error found
$ 
$ dotc -version

Dotty compiler version 0.01 -- (c) 2013 LAMP/EPFL
$ 
$ dotc t.scala
Warning: mocking up superclass for module class test
$ 
$ cat t.scala
trait Foo {
  def test (x : test) : Int
} 
$ 
```

liufengyun:
Fix #1701 : disqualify package name in type name resolution

Review @odersky

### DIFF
--- a/src/dotty/tools/dotc/core/Symbols.scala
+++ b/src/dotty/tools/dotc/core/Symbols.scala
@@ -428,7 +428,7 @@ object Symbols {
     final def entered(implicit ctx: Context): this.type = {
       assert(this.owner.isClass, s"symbol ($this) entered the scope of non-class owner ${this.owner}") // !!! DEBUG
       this.owner.asClass.enter(this)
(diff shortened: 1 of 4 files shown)


odersky:
I think here we should try to prevent the problem at the root. Can you try to change line 431 in `Symbols.scala` to the following:

      if (this.is(Module, butNot = Package)) this.owner.asClass.enter(this.moduleClass)

Then we won't have add more clauses to `qualifies`.

liufengyun:
@odersky I tried and it works!

odersky:
LGTM :-)

PR #1722, opened 2016-11-16
Fix #1708: duplicate symbols in package

### LINKED ISSUE #1708: crash on compilable code:at dotty.tools.dotc.core.Contexts$FreshContext.setOwner
scalac can compile it but dotc crashes.

$ cat abc.scala
```
package foo { trait identifier }
package foo { package identifier {} }
```

$ dotc abc.scala
```

	at dotty.tools.dotc.core.Contexts$FreshContext.setOwner(Contexts.scala:451)
	at dotty.tools.dotc.core.Contexts$Context.superOrThisCallContext(Contexts.scala:359)
	at dotty.tools.dotc.core.Contexts$Context.superCallContext(Contexts.scala:337)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedClassDef$1.apply(Typer.scala:1200)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedClassDef$1.apply(Typer.scala:1198)
	at dotty.tools.dotc.typer.Typer.typedClassDef(Typer.scala:1198)
	at dotty.tools.dotc.typer.Typer.typedNamed$1(Typer.scala:1434)
	at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:1482)
<snipped>
```

felixmulder:
review: @odersky

### DIFF
--- a/src/dotty/tools/dotc/typer/Namer.scala
+++ b/src/dotty/tools/dotc/typer/Namer.scala
@@ -21,6 +21,7 @@ import Inferencing._
 import transform.ValueClasses._
 import TypeApplications._
 import language.implicitConversions
+import reporting.diagnostic.messages._
 
 trait NamerContextOps { this: Context =>
 
@@ -264,7 +265,7 @@ class Namer { typer: Typer =>
       def preExisting = ctx.effectiveScope.lookup(name)
       if (ctx.owner is PackageClass)
         if (preExisting.isDefinedInCurrentRun)
-          errorName(s"${preExisting.showLocated} is compiled twice")
+          errorName(s"${preExisting.showLocated} has already been compiled\nonce during this run")
         else name
       else
         if ((!ctx.owner.isClass || name.isTypeName) && preExisting.exists)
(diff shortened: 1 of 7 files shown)


felixmulder:
Fixed failing test, this error was not caught previously:

```
-- Error: tests/neg/i1643.scala --------------------------------------
1 |trait T extends Array {
  |      ^
  |      cannot extend final class Array

```

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -343,8 +344,17 @@ class Namer { typer: Typer =>
       case Select(qual: RefTree, _) => createPackageSymbol(qual).moduleClass
     }
     val existing = pkgOwner.info.decls.lookup(pid.name)
+
     if ((existing is Package) && (pkgOwner eq existing.owner)) existing
-    else ctx.newCompletePackageSymbol(pkgOwner, pid.name.asTermName).entered
+    else {
+      /** If there's already an existing type, then the package is a dup of this type */
+      val existingTpe = pkgOwner.info.decls.lookup(pid.name.toTypeName)
+      if (existingTpe != NoSymbol) {
```
odersky: More idiomatic: `if (existingType.exists)`

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -343,8 +344,17 @@ class Namer { typer: Typer =>
...
+        ctx.error(PkgDuplicateSymbol(existingTpe), pid.pos)
+        ctx.newCompletePackageSymbol(pkgOwner, (pid.name ++ "$termDup").toTermName).entered
```
odersky: I propose we use the naming scheme `pid.name ++ "$_error_` here.

src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -261,10 +262,10 @@ class Namer { typer: Typer =>
         ctx.error(msg, tree.pos)
         name.freshened
       }
-      def preExisting = ctx.effectiveScope.lookup(name)
+      lazy val preExisting = ctx.effectiveScope.lookup(name)
```
odersky: I don't think we need a lazy val here. preExisting is only used once on each path. Note that defs are more compact (in terms of byte code length) and efficient than lazy vals.

felixmulder:
Done with suggested changes :+1:

odersky:
LGTM, thanks!

PR #1724, opened 2016-11-17
Fix #1648: don't define companion object for java.lang.Object

### LINKED ISSUE #1468: Missing type apply for scala.Dynamic

### LINKED ISSUE #1648: crash with assertion failed: attempt to parse java.lang.Object from classfile
$ cat abc.scala
```
class Foo { Object[A] }

```

$ dotc abc.scala
```
-- [E006] Missing Identifier Error: abc.scala --------------------------------------------------------------------------
1 |class Foo { Object[A] }
  |                   ^
  |                   not found: type A

exception caught when loading module class Object$: java.lang.AssertionError: assertion failed: attempt to parse java.lang.Object from classfile

Exception in thread "main" java.lang.AssertionError: assertion failed: attempt to parse java.lang.Object from classfile
	at dotty.tools.dotc.core.classfile.ClassfileParser$ConstantPool.getSuperClass(ClassfileParser.scala:1017)
	at dotty.tools.dotc.core.classfile.ClassfileParser.parseParents$1(ClassfileParser.scala:103)
	at dotty.tools.dotc.core.classfile.ClassfileParser.parseClass(ClassfileParser.scala:119)
	at dotty.tools.dotc.core.classfile.ClassfileParser.run(ClassfileParser.scala:54)
	at dotty.tools.dotc.core.ClassfileLoader.load(SymbolLoaders.scala:259)
	at dotty.tools.dotc.core.ClassfileLoader.doComplete(SymbolLoaders.scala:255)
	at dotty.tools.dotc.core.SymbolLoader.complete(SymbolLoaders.scala:205)
	at dotty.tools.dotc.core.SymDenotations$LazyType$$anon$1.complete(SymDenotations.scala:1903)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.completeFrom(SymDenotations.scala:210)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.info(SymDenotations.scala:188)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.ensureCompleted(SymDenotations.scala:327)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.flags(SymDenotations.scala:136)
	at dotty.tools.dotc.core.SymDenotations$ModuleCompleter.complete(SymDenotations.scala:1941)
<snipped>
```

liufengyun:
Fix #1648: `java.lang.Object` has no static methods, there's no need to define module object for it.

### DIFF
--- a/src/dotty/tools/dotc/core/Definitions.scala
+++ b/src/dotty/tools/dotc/core/Definitions.scala
@@ -194,6 +194,13 @@ class Definitions {
     val cls = ctx.requiredClass("java.lang.Object")
     assert(!cls.isCompleted, "race for completing java.lang.Object")
     cls.info = ClassInfo(cls.owner.thisType, cls, AnyClass.typeRef :: Nil, newScope)
+
+    // The companion object doesn't really exist, `NoType` is the general
+    // technique to do that. Here we need to set it before completing
+    // attempt to load Object's classfile, which causes issue #1648.
+    val companion = JavaLangPackageVal.info.decl(nme.Object).symbol
+    companion.info = NoType // to indicate that it does not really exist
+
     completeClass(cls)
   }
   def ObjectType = ObjectClass.typeRef
--- a/tests/neg/i1648.scala
+++ b/tests/neg/i1648.scala
@@ -1,0 +1,1 @@
+class Foo { Object[A] } // error // error


liufengyun:
It seems the fix works. I'm not sure if it's the right way to do it. Could you please advise, @odersky ?

src/dotty/tools/dotc/core/SymbolLoaders.scala
```diff
@@ -86,18 +86,27 @@ class SymbolLoaders {
...
   def enterClassAndModule(
       owner: Symbol, name: PreName, completer: SymbolLoader,
       flags: FlagSet = EmptyFlags, scope: Scope = EmptyScope)(implicit ctx: Context): Unit = {
     val clazz = enterClass(owner, name, completer, flags, scope)
-    val module = enterModule(
-      owner, name, completer,
-      modFlags = flags.toTermFlags & RetainedModuleValFlags,
-      clsFlags = flags.toTypeFlags & RetainedModuleClassFlags,
-      scope = scope)
+
```
odersky: It's a way to do it, but not very elegant to test this for every class and module that's entered. 

I played with an alternative strategy, which also seems to work and is less intrusive: IN line 196 of Definitions, where we define ObjectClass, insert:

    val companion = JavaLangPackageVal.info.decl(nme.Object).symbol
    companion.info = NoType // to indicate that it does not really exist

odersky: Note: The `NoType` technique is not an accident. It is used for all companion objects that are entered (because we see their file name) but turn out to not exist after class loading. Here we have to do it before class loading because once we try to load Object's classfile it is already too late.

liufengyun: Thanks @odersky , this is much more elegant 👍 . I've pushed a new commit adopting the approach.

odersky:
LGTM

PR #1728, opened 2016-11-21
Linker call graph with DCE

nicolasstucki:


DarkDimius:
Before merging it would be nice to do a proper rebase of historical commits that introduced the callgraph, as those commits contain a lot of knowledge.

nicolasstucki:
This is the first attempt to make all callgraph/dce tests work on partest. Currently on my machine `pos/t807.scala` fails without a clear reason.

PR #1733, opened 2016-11-22
Disable scaladoc generation for dotty-compiler

smarter:
We still publish an empty dotty-compiler_2.11-javadoc.jar to make
sonatype happy.

Review by @felixmulder


felixmulder:
LGTM

PR #1734, opened 2016-11-22
[WIP] add drone

felixmulder:
Drone is our new CI infrastructure based on docker images instead of VMs in Jenkins - the address is http://dotty-ci.epfl.ch. All members of the lampepfl group have access to the internals of this site.

### DIFF
--- a/compiler/test/dotty/partest/DPConsoleRunner.scala
+++ b/compiler/test/dotty/partest/DPConsoleRunner.scala
@@ -105,9 +105,9 @@ extends SuiteRunner(testSourcePath, fileManager, updateCheck, failed, javaCmdPat
     val seqResults =
       if (!sequentialTests.isEmpty) {
         val savedThreads = sys.props("partest.threads")
-        sys.props("partest.threads") = "1"
+        sys.props("partest.threads") = "2"
 
-        NestUI.echo(s"## we will run ${sequentialTests.length} tests sequentially")
+        NestUI.echo(s"## we will run ${sequentialTests.length} tests using ${PartestDefaults.numThreads} thread(s)")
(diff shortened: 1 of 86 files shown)


smarter:
> [error] (dotty-interfaces/compile:doc) java.io.IOException: Cannot run program "javadoc": error=2, No such file or directory

You need javadoc in your docker image I guess

felixmulder:
It seems to work fine now, I'm merging this so that we can have both CIs running at the same time to try out the differences.

This will also help @nicolasstucki - since everything above 100 commits gets ignored by the other CI.

felixmulder:
Also - note that the new CI will re-run tests on master to see if anything broke after merge.

compiler/test/dotty/partest/DPConsoleRunner.scala
```diff
@@ -105,9 +105,9 @@ extends SuiteRunner(testSourcePath, fileManager, updateCheck, failed, javaCmdPat
     val seqResults =
       if (!sequentialTests.isEmpty) {
         val savedThreads = sys.props("partest.threads")
-        sys.props("partest.threads") = "1"
+        sys.props("partest.threads") = "2"
 
-        NestUI.echo(s"## we will run ${sequentialTests.length} tests sequentially")
+        NestUI.echo(s"## we will run ${sequentialTests.length} tests using ${PartestDefaults.numThreads} thread(s)")
```
smarter: You did not update the documentation and the variable names to reflect the fact that these tests are not run sequentially anymore.

felixmulder: d'oh! On it

PR #1738, opened 2016-11-23
Make tests depend on environment variables for classpath jars

felixmulder:
sbt adds the correct jars to classpath and the tests depend on
`packageAll` which creates these. When using something else, however,
these together with `sbt-interfaces` do not get propagated from the
build.

To remedy this and make the testing a bit more flexible, we now
take these from `sys.props` instead, see `tests/dotty/Jars.scala`.

If the props aren't defined we fall back to the ones default to sbt.

review by: @odersky

### DIFF
--- a/compiler/test/dotty/partest/DPConsoleRunner.scala
+++ b/compiler/test/dotty/partest/DPConsoleRunner.scala
@@ -114,9 +114,16 @@ extends SuiteRunner(testSourcePath, fileManager, updateCheck, failed, javaCmdPat
     val seqResults =
       if (!limitResourceTests.isEmpty) {
         val savedThreads = sys.props("partest.threads")
-        sys.props("partest.threads") = limitedThreads
+        sys.props("partest.threads") = {
+          assert(
+            savedThreads == null || limitedThreads.toInt <= savedThreads.toInt,
+            """|Should not use more threads than the default, when the point
+               |is to limit the amount of resources""".stripMargin
+          )
+          limitedThreads
+        }
 
-        NestUI.echo(s"## we will run ${limitResourceTests.length} tests using ${PartestDefaults.numThreads} thread(s)")
+        NestUI.echo(s"## we will run ${limitResourceTests.length} tests using ${PartestDefaults.numThreads} thread(s) in parallel")
         val res = super.runTestsForFiles(limitResourceTests, kind)
 
         if (savedThreads != null)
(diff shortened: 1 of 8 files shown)


compiler/test/dotty/tools/dotc/EntryPointsTest.scala
```diff
@@ -20,9 +21,9 @@ class EntryPointsTest {
   private val sources =
     List("../tests/pos/HelloWorld.scala").map(p => new java.io.File(p).getPath())
   private val dottyInterfaces =
-    new java.io.File("../interfaces/dotty-interfaces-0.1-SNAPSHOT.jar").getPath
+    new java.io.File(Jars.dottyInterfaces).getPath
   private val dottyLibrary =
-    new java.io.File("../library/target/scala-2.11/dotty-library_2.11-0.1-SNAPSHOT.jar").getPath
+    new java.io.File(Jars.dottyLib).getPath
```
smarter: Note that the whole point of this test and `InterfaceEntryPointTest` is to give an example of how to use the compilers entry points for external projects; so relying on extra files and/or hardcoded path is not great.

felixmulder: Sure, but do you have an alternative?

project/Build.scala
```diff
@@ -15,6 +15,7 @@ object DottyBuild extends Build {
   val isNightly = sys.env.get("NIGHTLYBUILD") == Some("yes")
 
   val jenkinsMemLimit = List("-Xmx1500m")
+  val ciLimitedThreads = List("-Ddotty.tests.limitedThreads=2")
```
smarter: Is this necessary if this is the default?

compiler/test/dotty/partest/DPConsoleRunner.scala
```diff
@@ -88,27 +88,35 @@ extends SuiteRunner(testSourcePath, fileManager, updateCheck, failed, javaCmdPat
...
+    * when run in parallel with too many other tests.
+    *
+    * This number could be increased on the CI, but might fail locally if
+    * scaled too extreme - override with:
+    *
+    * `-Ddotty.tests.limitedThreads=X` or by changing the `ciLimitedThreads` option
+    * in `project/Build.scala`.
+    */
+  def limitResourceFlags = List("-Ytest-pickler")
+  private val limitedThreads = sys.props.get("dotty.tests.limitedThreads").getOrElse("2")
```
smarter: You should make sure that `limitedThreads` is always less than or equal to `partest.threads`

compiler/test/dotty/partest/DPConsoleRunner.scala
```diff
@@ -88,27 +88,35 @@ extends SuiteRunner(testSourcePath, fileManager, updateCheck, failed, javaCmdPat
...
     val seqResults =
-      if (!sequentialTests.isEmpty) {
+      if (!limitResourceTests.isEmpty) {
         val savedThreads = sys.props("partest.threads")
-        sys.props("partest.threads") = "2"
+        sys.props("partest.threads") = limitedThreads
 
-        NestUI.echo(s"## we will run ${sequentialTests.length} tests using ${PartestDefaults.numThreads} thread(s)")
-        val res = super.runTestsForFiles(sequentialTests, kind)
+        NestUI.echo(s"## we will run ${limitResourceTests.length} tests using ${PartestDefaults.numThreads} thread(s)")
```
smarter: This message is slightly different from the message below, either add `in parallel` here or remove it below

compiler/test/dotty/tools/dotc/EntryPointsTest.scala
felixmulder: Fixed

felixmulder:
Superseded by https://github.com/lampepfl/dotty/pull/1739 sot that Martin can add his fixes for Eclipse

PR #1739, opened 2016-11-23
Environmentally friendly tests

felixmulder:

PR #1740, opened 2016-11-23
Fix #1737: Enable GADT checking for objects

### LINKED ISSUE #1737: Pattern matching on types having type parameters or type members
The following code does not compile:

~~~ scala
  sealed trait Foo[A]
  case object FooI extends Foo[Int]
  case class FooS(b: Boolean) extends Foo[String]

  def algFoo[A](foo: Foo[A]): A =
    foo match {
      case FooI => 42
      case FooS(b) => "foo"
    }
~~~
~~~scala
  sealed trait Bar { type Out }
  case object BarI extends Bar { type Out = Int }
  case class BarS(b: Boolean) extends Bar { type Out = String }

  def algBar(bar: Bar): bar.Out =
    bar match {
      case BarI => 42
      case BarS(b) => "bar"
    }
~~~

The errors are:

~~~
[error] -- [E007] Type Mismatch Error: -------------------
[error] 18 |      case FooI => 42
[error]    |                   ^^
[error]    |                   found:    Int(42)
[error]    |                   required: A
[error]    |                   
[error] -- [E007] Type Mismatch Error: -------------------
[error] 28 |      case BarI => 42
[error]    |                   ^^
[error]    |                   found:    Int(42)
[error]    |                   required: bar.Out
[error]    |                   
[error] -- [E007] Type Mismatch Error: -------------------
[error] 29 |      case BarS(b) => "bar"
[error]    |                      ^^^^^
[error]    |                      found:    String("bar")
[error]    |                      required: bar.Out
[error]    |                      
[error] three errors found
~~~

The first example (with `Foo`) compiles fine with scalac. In Dotty, the error is raised only in the case of a `case object`, not a `case class`.
The second example (with `Bar`) fails to compile with both scalac and dotc.

odersky:
Review  by @julienrf

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1830,7 +1830,9 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         ctx.typeComparer.GADTused = false
         if (ctx.mode is Mode.Pattern) {
           tree match {
-            case _: RefTree | _: Literal if !isVarPattern(tree) =>
+            case _: RefTree | _: Literal
+            if !isVarPattern(tree) &&
+               !(tree.tpe <:< pt)(ctx.addMode(Mode.GADTflexible)) =>
               checkCanEqual(pt, wtp, tree.pos)(ctx.retractMode(Mode.Pattern))
             case _ =>
           }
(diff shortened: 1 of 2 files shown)


odersky:
This is based on #1739. Only the last commit is new and needs to be reviewed.

smarter:
last commit LGTM

julienrf:
Looks great, thanks for the quick fix!

PR #1742, opened 2016-11-24
Fix colons in printer

felixmulder:
This PR takes off where @rethab started and fixes defs not being properly printed when desugared to `TypeExpr`.

It does not fix the underlying issues with the REPL. The current strategy employed by the REPL is to gather all the types that were generated during a compilation and tie them to the correct names. This proves an issue though, because the types need to get printed using special hacks.

Ideally we would do an actual compile run and then just show the types without any distortion. But this I leave up to a later point in time when we're doing a more complete overhaul.

review: @OlivierBlanvillain

### DIFF
--- a/tests/repl/vars.check
+++ b/tests/repl/vars.check
@@ -1,8 +1,8 @@
 scala> var x = 0
-x: Int = 0
+var x: Int = 0
 scala> x = x + 1
 x: Int = 1
 scala> x *= 2
 scala> x
-res2: Int = 2
+val res2: Int = 2
 scala> :quit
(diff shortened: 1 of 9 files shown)


OlivierBlanvillain:
LGTM :+1:

tests/repl/vars.check
```diff
@@ -1,8 +1,8 @@
 scala> var x = 0
-x: Int = 0
+var x: Int = 0
 scala> x = x + 1
 x: Int = 1
 scala> x *= 2
 scala> x
-res2: Int = 2
+val res2: Int = 2
```
felixmulder: Indeed, the REPL currently has semantics that are not well defined. There's no spec for it :sweat_smile: 

But, if you want to have a closer look at its semantics, by all means! We'd be happy to get something that is clearer :)

PR #1744, opened 2016-11-25
Try fix #1743: typing of ValDef with anonymous class

### LINKED ISSUE #1743: Problem with typing of ValDef with anonymous class
Following code fails to type check in dotc, which is the second issue reported in #1707:

```Scala
object DepBug {
  val dep = new {
    val a = 3
    val b = 4
  }

  dep.a + dep.b
}
```

liufengyun:
In typing of blocks, it's fine to reference local type symbols. Fix #1743.


odersky:
Not sure we can do it that way. Avoidance is a tricky issue...

odersky:
In principle, we should not let a locally defined type escape in the type of a block.

liufengyun:
Thanks @odersky , so I guess in this case the only right solution is to resort to structural types, which we have to decide whether to support or not.

I close it for now.

PR #1746, opened 2016-11-27
replaced 'none' with 'None' and 'nil' with 'Nil' when pattern matching

rethab:
small cleanup.


odersky:
In fact, using none/nil was fully intentional. It conveys intent without imposing a runtime cost. In each case, we know that the case is none or nil, but don't trust the pattern matcher to act on that knowledge by eliminating a costly check. Compilers have to be fast and the only way to achieve that is attention to every detail.

rethab:
Interesting! Thanks a lot for the explanation.

PR #1749, opened 2016-11-27
fix #1748: desugaring with StringContext in PatDef

### LINKED ISSUE #1748: Desugaring problem with StringContext in PatDef
Following code failed to compile in Dotty, but it's fine in Scala:

```Scala
object Test {
    object q {
      def unapply(args: Any*): Option[(Any, Any)] =
    }
  }

  def f(defn: Any): Any = {
    val q"class $name extends $parent" =  defn
    println(name)
    println(parent)
  }
}
```

It turns out in Desugaring, a case for `Thicket` is missing.

liufengyun:
Fix #1748: desugaring with StringContext in PatDef

The method `getVariables` forgets the case `Thicket`.

Review @felixmulder

### DIFF
--- a/tests/pos/i1784.scala
+++ b/tests/pos/i1784.scala
@@ -1,0 +1,14 @@
+object Test {
+  implicit class Foo(sc: StringContext) {
+    object q {
+      def unapply(args: Any*): Option[(Any, Any)] =
+        Some((sc.parts(0), sc.parts(1)))
+    }
+  }
+
+  def f(defn: Any): Any = {
+    val q"class $name extends $parent" =  defn
+    println(name)
+    println(parent)
+  }
+}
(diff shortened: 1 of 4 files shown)


tests/pos/i1784.scala
```diff
@@ -0,0 +1,14 @@
...
+        Some((sc.parts(0), sc.parts(1)))
+    }
+  }
+
+  def f(defn: Any): Any = {
+    val q"class $name extends $parent" =  defn
+    println(name)
+    println(parent)
+  }
+}
```
felixmulder: Perhaps this should be a run-test instead since you're printing things? :)

felixmulder:
Checkfile not correct?

liufengyun:
Let me try adding a newline at the end of the check file.

PR #1752, opened 2016-11-28
fix publishing of dotty-library

liufengyun:
Fix publishing of `dotty-library`. Review @felixmulder .


felixmulder:
LGTM, now we just need to fix the nightly builds to work with the new structure.

felixmulder:
fixed nightlies :+1:

liufengyun:
Thanks @felixmulder !

PR #1758, opened 2016-11-30
Drop function 22 limit

odersky:
Functions with more than 22 parameters are now automatically converted to functions taking
a single object array parameter.

This has been achieved using two tricks:

 - Create function types of arbitrary arities on demand.
    This was achieved by tweaking the scope of the scala package.

 - Eliminate function types of large arity by mapping them to a new
    FunctionXXL type that takes all parameters in a single object array.
    This was achieved by tweaking erasure.

Other things I have tried that did not work out well:

 - Use a single function type in typer. The problem with this
   one which could not be circumvented was that existing higher-kinded
   code with e.g. Functor assumes that Functon1 is a binary type constructor.

 - Have a late phase that converts to FunctonXXL instead of
   doing it in erasure. The problem with that one was that
   potentially every type could be affected, which was ill-suited
   to the architecture of a miniphase.

Review by @OlivierBlanvillain ?


odersky:
The PR number #1758 is interesting: 17 + 5 is 22 and 17+5+8 = 30, our previous limit of the arity of temporary functions :wink:

sledorze:
fun all way down :)

odersky:
> Looks great! I've tried running the full test suite with MaxImplementedFunctionArity = 3 which revealed are quite a few failing test cases, for instance:

I should have mentioned: Only the implemented function traits have tupled and curried methods. The synthetic ones lack them.

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -646,6 +684,8 @@ class Definitions {
     tp.derivesFrom(NothingClass) || tp.derivesFrom(NullClass)
 
   def isFunctionClass(cls: Symbol) = isVarArityClass(cls, tpnme.Function)
+  def isUnimplementedFunctionClass(cls: Symbol) =
+    isFunctionClass(cls) && cls.name.functionArity >= MaxImplementedFunctionArity
```
odersky: yes, indeed.

library/src/scala/FunctionXXL.scala
```diff
@@ -0,0 +1,17 @@
...
+** /____/\___/_/ |_/____/_/ | |                                         **
+**                          |/                                          **
+\*                                                                      */
+// GENERATED CODE: DO NOT EDIT. See scala.Function0 for timestamp.
+package scala
+
+/** A function with all parameters grouped in an array. */
+trait FunctionXXL {
+
+  def apply(xs: Array[Object]): Object
```
odersky: FunctionXXL is only used after erasure, when Any is mapped to Object anyway. So I prefer to be explicit.

odersky:
I think we should add the testing setup in another PR (if we get around to it). To recall: We'd need to move all `curried` and `tupled` functions into decorators. Then we should be able to reduce `MaxImplementedFunctionArity` at will.

Issue #1681, opened 2016-11-07
REPL: uppercase strings have different colors than lowercase


rethab:
```scala
scala> "lowercase"
res16: String = lowercase
scala> "UPPERCASE"
res17: String = UPPERCASE
scala> "lower UPPER lower"
res18: String = lower UPPER lower
```

![2016-11-07_414x156](https://cloud.githubusercontent.com/assets/2222044/20077830/54f7a2e0-a53e-11e6-8c80-6ef378304276.png)

smarter:
I think the correct thing to do here is to print the string with `"` around it, as a bonus this means that the output of the repl is valid Scala, but it also means that we have to escape our output, e.g.
```scala
scala> "\""
res0: String = " // Before
res0: String = "\"" // After
```
I think that's a trade-off worth making

liufengyun:
Thanks for reporting @rethab . @felixmulder can you have a look at this one?

felixmulder:
I believe my time is best spent elsewhere at the moment. The issue stems from how we render results. The main culprit is [resultExtractor](https://github.com/lampepfl/dotty/blob/master/src/dotty/tools/dotc/repl/CompilingInterpreter.scala#L683), which needs to be changed.

This method needs to correctly render strings. But there are a number of issues/questions that will arise when somebody starts to look at this:

1. Should I just render strings differently?
2. What about higheir kinded types containing strings? `List[String]`?
3. How do I deal with multiline strings? Do I escape newlines there?

I think the issue needs to clarify these three points before somebody starts working on this. After that I think that it is an excellent entry-level task that could be handled by outside contributors.

Blaisorblade:
Regarding 1. and 2., you indeed want to print values as valid Scala, and `toString` simply can't do that. You don't even need parametric types like `List[String]`, a case class containing `String` is enough.

```
scala> case class A(a: String); A("hi there")
defined class A
res0: A = A(hi there)
```

Maybe there's some local REPL-only solution, but the wider issue remains.

Indeed, many languages with REPLs (say Haskell or Lisps) distinguish how the REPL shows values as expressions (say Haskell `Show` or Racket `print`) from how to present values to users (Haskell has `putStr` for string output, Lisp has `display`). Scala lacks a `Show` typeclass.

If you want an existing solution to this (and other problems), there's among others @lihaoyi's PPrint which was born for his Ammonite REPL (and I imagine there are other ports inspired by Haskell's `Show`). @lihaoyi do you have recommendations?

[...]

PR #1761, opened 2016-11-30
[REPL] Add show capability to common types

felixmulder:
This PR adds a `trait Show[-T]` to the Dotty library in order to print things in a better way for both the REPL and potentially other serializations. It also makes sure that the REPL shows all variables instead of `toString`ing them directly.

For now, there's a default show instance that calls `toString` so that there's a fallback in case of a `Show` instance not existing for a particular type. But potentially we could disallow printing things without a `Show` instance in the future...

Since we don't have control over the stdlib at the moment, we're stuck with providing default typeclass instances in the `dotty.Show` object. But ideally once @OlivierBlanvillain's HList implementation is in, perhaps we should investigate making a `Show` instance for `DottyProduct` or something similar so that we can have showing of all case classes :balloon: 

A known shortcoming in this PR is with regards to union types in e.g:

```scala
scala> List(1, 2, "three")
val res0: List[Int | String] = List(1, 2, three)
// instead of: List(1, 2, "three")
```

I tried to make the default instance down cast things if possible, but that put me into overflows for certain cases. Perhaps I could try harder if we really want this before bootstrapped library...

review: @odersky | @smarter 

merging this solves: #1681

### DIFF
--- a/library/src/dotty/Show.scala
+++ b/library/src/dotty/Show.scala
@@ -7,20 +7,27 @@ trait Show[-T] {
   def show(t: T): String
 }
 
+/** Ideally show would only contain `defaultShow` and the pimped generic class,
+  * but since we can't change the current stdlib, we're stuck with providing
+  * default instances in this object
+  */
 object Show {
   private[this] val defaultShow = new Show[Any] {
     def show(x: Any) = x.toString
   }
 
+  /** This class implements pimping of all types to provide a show method.
+    * Currently it is quite permissive, if there's no instance of `Show[T]` for
+    * any `T`, we default to `T#toString`.
+    */
   implicit class ShowValue[V](val v: V) extends AnyVal {
     def show(implicit ev: Show[V] = defaultShow): String =
       ev.show(v)
   }
 
   implicit val stringShow = new Show[String] {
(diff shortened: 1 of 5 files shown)


library/src/dotty/Show.scala
```diff
@@ -0,0 +1,102 @@
...
+    }) + "'"
+  }
+
+  implicit def showList[T](implicit st: Show[T]): Show[List[T]] = new Show[List[T]] {
+    def show(xs: List[T]) =
+      if (xs.isEmpty) "Nil"
+      else "List(" + xs.map(_.show).mkString(", ") + ")"
+  }
+
+  implicit val showNil: Show[List[Nothing]] = new Show[List[Nothing]] {
```
felixmulder: I did, but I didn't push it! (=

It actually gives a better representation of list:

```scala
scala> Nil
val res0: scala.collection.immutable.Nil.type = Nil
scala> List()
val res1: List[Nothing] = List()
```

instead of printing `Nil` for both.

smarter:
See also the discussion in https://github.com/scala/scala/pull/5222

odersky:
LGTM 👍

PR #1762, opened 2016-11-30
[WIP] Specialize functions

odersky:
I believe that with the same machinery that eliminates the function 22 limit we can also specialize function types.

The existsing FunctionalInterfaces was moved to erasure as a first step. Note that FunctionalInterfaces does not specialize function types, it only specializes closures. Used from dotty, this has no effect, it's an interop feature only. What this PR is about is to obtain full function specialization.


compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,85 @@
...
+import core.NameOps._
+import ast.Trees._
+import SymUtils._
+import core.Definitions.PerRun
+import dotty.tools.dotc.ast.tpd
+import collection.{ mutable, immutable }
+import collection.mutable.{ LinkedHashMap, LinkedHashSet, TreeSet }
+
+/**
+ *  Rewires closures to implement more specific types of Functions.
```
smarter: Could you add a comment explaining what prevents this from being a mini-phase?

DarkDimius:
It's not clear to me why the full function specialization can't be done in two miniphaphases:
 - First one replaces super-interfaces for classes and closures that extend Function1 to specialized variants before erasure
 - second one dispatches to more appropriate apply overload.

Neither of those would affect type signatures and neither of those would need to go into erasure.

DarkDimius:
Compared to the current state, we don't transform super-traits and we don't dispatch to appropriate apply. I think both of those can be easily implemented in a miniphase with an InfoTransofrmer that changes parents of classes and does not force cyclic reference errors as it does not need to ever go deep.

odersky:
Let's wait discussing until the full implementation is there.

odersky:
Let's close this for now and try the miniphase approach.

PR #1763, opened 2016-11-30
Fix emission of annotations

### LINKED ISSUE #1741: JUnit doesn't work with dotty-bridge
I have a Dotty project uses JUnit:

```Scala
libraryDependencies += "com.novocode" % "junit-interface" % "0.11" % "test"
testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v")
```

With the test file:

```Scala
import org.junit.Test
import org.junit.Assert._

class MacrosTest {
  @Test def hello = {
    assertTrue(true)
  }
}
```

When run `sbt test`, nothing happens.  When compile with scalac, everything works normally as confirmed by @smarter .

smarter:
Review by @DarkDimius

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -206,7 +206,6 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
   implicit val ConstantClassTag: ClassTag[Constant] = ClassTag[Constant](classOf[Constant])
   implicit val ClosureTag: ClassTag[Closure] = ClassTag[Closure](classOf[Closure])
 
-  /* dont emit any annotations for now*/
   def isRuntimeVisible(annot: Annotation): Boolean = {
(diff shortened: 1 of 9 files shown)


smarter:
Looks like this unmasked an issue with the inline body annotations, investigating...

DarkDimius:
Backend-wise LGTM

smarter:
This still isn't enough to fix #1741, turns out [sbt uses the incremental compilation analysis to know which methods have special annotations and need to be run with a test interface](https://github.com/sbt/sbt/blob/0.13/main/actions/src/main/scala/sbt/Tests.scala#L288), but I never implemented support for annotations in incremental compilation: https://github.com/lampepfl/dotty/blob/b9350f40990ce07ba7614a0448a98abd7075abe8/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala#L501, so I know what I need to do next!

smarter:
Alright, #1741 is fixed by the latest commit, and I added a test to test that the tests are testing something :)

liufengyun:
Well done @smarter 👍  !

smarter:
Ping @DarkDimius , would you mind review the remaining commits?

DarkDimius:
LGTM

PR #1769, opened 2016-12-02
Fix #1751: Make dominator work after erasure

### LINKED ISSUE #1751: crash on compilable code: scala.MatchError: List() (of class scala.collection.immutable.Nil$)
$ cat abc.scala
```
trait Break { protected val break: Int; }
case class BreakImpl(protected val break: Int) extends Break {}
object Test {
  def f2(x: Break) = x match {
    case BreakImpl(x) => BreakImpl
    case _ => -1
  }
  def f4(x: Any) = x match {
    case BreakImpl(x) => x
    case _ => -1
  }
    val break = BreakImpl(22)
    assert(f2(break) == 22)
    assert(f4(break) == 22)
  }
}
```

$ dotc abc.scala
```
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@445b84c0] of class class dotty.tools.dotc.ast.Trees$Apply # 734
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@61a52fbd] of class class dotty.tools.dotc.ast.Trees$Apply # 736
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@233c0b17] of class class dotty.tools.dotc.ast.Trees$Block # 743
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@63d4e2ba] of class class dotty.tools.dotc.ast.Trees$DefDef # 744
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@7bb11784] of class class dotty.tools.dotc.ast.Trees$TypeDef # 746
exception while typing [cannot display due to scala.MatchError: List() (of class scala.collection.immutable.Nil$), raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@5594a1b5] of class class dotty.tools.dotc.ast.Trees$PackageDef # 822

Exception in thread "main" scala.MatchError: List() (of class scala.collection.immutable.Nil$)
        at dotty.tools.dotc.core.TypeOps$class.dominators$1(TypeOps.scala:196)
        at dotty.tools.dotc.core.TypeOps$class.approximateOr$1(TypeOps.scala:242)
        at dotty.tools.dotc.core.TypeOps$class.orDominator(TypeOps.scala:256)
        at dotty.tools.dotc.core.Contexts$Context.orDominator(Contexts.scala:57)
        at dotty.tools.dotc.core.Types$OrType.join(Types.scala:2233)
        at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:453)
        at dotty.tools.dotc.core.Types$Type.findMember(Types.scala:580)
<snipped>
```

odersky:
i1751.scala shows a case where we need to compute the approximation
of an or-type during erasure. This can lead to an empty set of common
classes because Any does not exist anymore after erasure.

Review by @liufengyun ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeOps.scala
@@ -197,6 +197,9 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
       case c :: rest =>
         val accu1 = if (accu exists (_ derivesFrom c)) accu else c :: accu
         if (cs == c.baseClasses) accu1 else dominators(rest, accu1)
+      case Nil => // this case can happen because after erasure we do not have a top class anymore
+        assert(ctx.erasedTypes)
+        defn.ObjectClass :: Nil
     }
 
     def mergeRefined(tp1: Type, tp2: Type): Type = {
--- a/tests/pos/i1751.scala
+++ b/tests/pos/i1751.scala
@@ -1,0 +1,17 @@
+trait Break { protected val break: Int; }
+case class BreakImpl(protected val break: Int) extends Break {}
+object Test {
+  def f2(x: Break) = x match {
+    case BreakImpl(x) => BreakImpl
+    case _ => -1
+  }
+  def f4(x: Any) = x match {
+    case BreakImpl(x) => x
+    case _ => -1
+  }
+  def main(args: Array[String]): Unit = {
+    val break = BreakImpl(22)
+    assert(f2(break) == 22)
+    assert(f4(break) == 22)
+  }
+}


DarkDimius:
Our of curiosity: where do we construct an Or type during erasure?

odersky:
> Our of curiosity: where do we construct an Or type during erasure?

We don't (in this case), actually.It's a type that was left in the unerased tree. The error happens with the third line below in the stacktrace: 

    override def typedApply(tree: untpd.Apply, pt: Type)(implicit ctx: Context): Tree = {
      val Apply(fun, args) = tree
      if (fun.symbol == defn.dummyApply)

As you can see, it takes the `symbol` of the previous unerased tree.

odersky:
Also relevant is the comment on this Config variable:

    /** Check that certain types cannot be created in erasedTypes phases.
     *  Note: Turning this option on will get some false negatives, since it is
     *  possible that And/Or types are still created during erasure as the result
     *  of some operation on an existing type.
     */
    final val checkUnerased = true

liufengyun:
@odersky It looks good to me. Following minimised version of test is more intuitive:

```Scala
object Test {
  def f(x: Int) = if (x > 0) -1 else "String"

  def h = f(3) == 22      //  `== ` triggers the exception
}
```

PR #1770, opened 2016-12-02
[DO NOT MERGE] 	Illustrate issue with TreeTypeMap and miniphases

smarter:
Ping @DarkDimius @odersky, I'm not sure how to proceed from here.

The first commit illustrates the issue:
> When running TreeTypeMap in a miniphase, the tree we're transforming has
already been transformed by all the miniphases in the same group, but
the symbols we're looking at might be older than the trees, this means
that we can miss transforming some symbols as illustrated in details by
the added testcase which currently fails.

The second commit contains a hack to show that my diagnostic of the issue above is correct, but I don't know how to fix this properly:
> With this commit, tests/pos/treetypemap-miniphase.scala now compiles,
but this is only a hack: there are other instances of TreeTypeMap in the
compiler and they would all need to be run with "atGroupEnd", this is
tricky since we are not always calling TreeTypeMap in a context where we
have access to the current TreeTransformer (for example when we create a
TreeTypeMap inside TypeMap to deal with annotations), ideas and
alternative fixes welcome.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ElimByName.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ElimByName.scala
@@ -80,7 +80,9 @@ class ElimByName extends MiniPhaseTransform with InfoTransformer { thisTransform
             val inSuper = if (ctx.mode.is(Mode.InSuperCall)) InSuperCall else EmptyFlags
             val meth = ctx.newSymbol(
                 ctx.owner, nme.ANON_FUN, Synthetic | Method | inSuper, MethodType(Nil, Nil, argType))
-            Closure(meth, _ => arg.changeOwner(ctx.owner, meth))
+            atGroupEnd { implicit ctx: Context =>
+              Closure(meth, _ => arg.changeOwner(ctx.owner, meth))
+            }
         }
         ref(defn.dummyApply).appliedToType(argType).appliedTo(argFun)
       case _ =>
--- a/tests/pos/treetypemap-miniphase.scala
+++ b/tests/pos/treetypemap-miniphase.scala
@@ -1,0 +1,38 @@
+class Test {
+  def byName(param: => Any) = {}
+
+  def test: Unit = {
+
+    // 1. In ResolveSuper#transformTemplate we add "def <(that: Int): Boolean =
+    //    Foo.super[Ordered].<(that)", etc to Foo, these methods are inserted
+    //    into a new decls scope using enteredAfter in MixinOps#implementation
+
+    // 2. In ElimByName#transformApply we call `changeOwner` on the tree
+    //    containing Foo, which will use a TreeTypeMap
+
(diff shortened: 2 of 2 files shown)


DarkDimius:
In short: we have already encountered this issue and we have introduced `changeOwnerAfter` that is supposed to be robust against this(which in turn introduced `Denotation.transformAfter`). See https://github.com/lampepfl/dotty/commit/979ee0ff41c2a957b22e886ea43ad4f070a9777d

If I remember correctly @odersky proposed to keep `changeOwner` though it is unsafe because it is faster.

DarkDimius:
Similarly: entered is generaly unsafe while enteredAfter is more robust.

compiler/src/dotty/tools/dotc/transform/ElimByName.scala
```diff
@@ -80,7 +80,9 @@ class ElimByName extends MiniPhaseTransform with InfoTransformer { thisTransform
             val inSuper = if (ctx.mode.is(Mode.InSuperCall)) InSuperCall else EmptyFlags
             val meth = ctx.newSymbol(
                 ctx.owner, nme.ANON_FUN, Synthetic | Method | inSuper, MethodType(Nil, Nil, argType))
-            Closure(meth, _ => arg.changeOwner(ctx.owner, meth))
+            atGroupEnd { implicit ctx: Context =>
+              Closure(meth, _ => arg.changeOwner(ctx.owner, meth))
```
DarkDimius: try `changeOwnerAfter`. It's supposed to fix the issue. No need for `atGroupEnd`.

DarkDimius:
The correct solution to this issue is to only change denotations\owners in InfoTransormer. Than neither of those functionalities would have been needed.

odersky:
> If I remember correctly @odersky proposed to keep changeOwner though it is unsafe because it is faster.

No, it's the opposite. `changeOwnerAfter` is faster. I kept `changeOwner` to make sure it was debugged, because we will need it in cases where we duplicate a tree. I think now it's time to switch to `changeOwnerAfter` in `elimByName`.

But there are other tree type maps, for instance in tailrec and extensionMethods. So we need to be careful with those.

smarter:
> The correct solution to this issue is to only change denotations\owners in InfoTransormer. Than neither of those functionalities would have been needed.

That's what I thought too, but that seems like a major refactoring since many phases change denotations while transforming trees.

smarter:
> But there are other tree type maps, for instance in tailrec and extensionMethods. So we need to be careful with those.

In particular, every `TypeMap` might instantiate a `TreeTypeMap` to map over an annotation that can end up creating new symbols: https://github.com/lampepfl/dotty/blob/47d208448e614125446c7f294f8231c3fb7108d6/compiler/src/dotty/tools/dotc/core/Types.scala#L3493

felixmulder:
Closed in favor of opening an issue

PR #1772, opened 2016-12-04
Fix #1702: handle ConstantType in TypeComparer

### LINKED ISSUE #1702: method with singleton type mismatches same singleton type
As the below example shows, a method that takes a singleton type `Int(1)` cannot be applied with `1`.

```scala
scala> def id(x: 1): 1 = x
id: (x: Int(1))Int(1)
scala> id(1)
-- [E007] Type Mismatch Error: <console> -----------------------------------
6 |id(1)
  |   ^
  |   found:    Int(1)
  |   required: Int(1)
  |
```

rethab:
@smarter please review, thanks!


smarter:
Thanks, looks good to me! @odersky: any objection to putting this in firstTry?

odersky:
>  @odersky: any objection to putting this in firstTry?

No, I think that's where it should go. LGTM.

PR #1778, opened 2016-12-07
Fix #1773: handle patterns in interpolated string

### LINKED ISSUE #1773: PatDef desugaring problem
Following code compiles fine in scalac but failed to compile in dotty. The cause is mostly like to be unmatched cases when desugaring patdef, as in #1748 .

```Scala
object Test {
    object q {
      def unapply(arg: Any): Option[(Any, Any)] =
    }
  }

    val q"class ${name: String} extends ${parent: String}" = new Object
    println(name)
    println(parent)
  }
}
```

liufengyun:
Fix #1773: handle patterns in interpolated string

Now following patterns can be correctly desugared and type checked:

```Scala
val q"class ${name: String} extends ${parent: String}" = ...
```

Review @odersky

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -916,7 +916,11 @@ object desugar {
         val elems = segments flatMap {
           case ts: Thicket => ts.trees.tail
           case t => Nil
+        } map {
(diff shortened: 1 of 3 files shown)


odersky:
Otherwise LGTM

liufengyun:
Thanks @odersky , comment addressed and CI passed. Merge now.

PR #1780, opened 2016-12-08
fix #1779: support $_ and $_id  in interpolated string

### LINKED ISSUE #1779: Parsing error with `$_` and `$_id` in interpolated strings
Dotty cannot parse `$_` and `$_id`, while Scalac only supports `$_`.

```Scala
object Test {
    object q {
      def unapply(arg: Any): Option[(Any, Any)] =
    }
  }

    val q"class $_ extends $_parent" = new Object
    println(_parent)
  }
}
```

liufengyun:
Fix #1779: support $_ and $_id  in interpolated string.

Review @odersky .


odersky:
Allowing `$_parent` is technically a deviation from `scalac`'s behavior. But I think your scheme is more consistent because we already allow underscores elsewhere in $-bound identifiers. 

LGTM.

PR #1783, opened 2016-12-11
Fix exhaustivity check test after repo restructuring

liufengyun:
The exhaustivity check tests don't run after repo restructuring. Review @felixmulder .


felixmulder:
LGTM

PR #1785, opened 2016-12-12
Fix #1784: allow to omit types for local implicit vals

### LINKED ISSUE #1784: Don't demand explicit types for local implicit vals
A local implicit val can be referred to only from text following the valdef. Therefore, it
does not need an explicit type for type inference reasons. By allowing to elide the type
we make porting from Scala 2.x easier and avoid busywork.

odersky:
Review by @felixmulder ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/PathResolver.scala
+++ b/compiler/src/dotty/tools/dotc/config/PathResolver.scala
@@ -143,7 +143,7 @@ object PathResolver {
       println(Defaults)
     }
     else {
-      implicit val ctx: Context = (new ContextBase).initialCtx // Dotty deviation: implicits need explicit type
+      implicit val ctx = (new ContextBase).initialCtx // Dotty deviation: implicits need explicit type
       val ArgsSummary(sstate, rest, errors) =
         ctx.settings.processArguments(args.toList, true)
       errors.foreach(println)
--- a/compiler/src/dotty/tools/dotc/config/PathResolver.scala
+++ b/compiler/src/dotty/tools/dotc/config/PathResolver.scala
@@ -143,7 +143,7 @@ object PathResolver {
       println(Defaults)
     }
     else {
-      implicit val ctx = (new ContextBase).initialCtx // Dotty deviation: implicits need explicit type
+      implicit val ctx = (new ContextBase).initialCtx
       val ArgsSummary(sstate, rest, errors) =
         ctx.settings.processArguments(args.toList, true)
       errors.foreach(println)
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3378,7 +3378,7 @@ object Types {
 
     /** Map this function over given type */
     def mapOver(tp: Type): Type = {
-      implicit val ctx = this.ctx // Dotty deviation: implicits need explicit type
+      implicit val ctx = this.ctx
       tp match {
         case tp: NamedType =>
           if (stopAtStatic && tp.symbol.isStatic) tp
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3378,7 +3378,7 @@ object Types {
 
     /** Map this function over given type */
     def mapOver(tp: Type): Type = {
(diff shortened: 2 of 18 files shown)


sjrd:
To exercise this in bootstrap, you could remove all the `: Position` I had to add in `JSCodeGen` on local `implicit val pos`'es, for example this one: https://github.com/lampepfl/dotty/blob/master/compiler/sjs/backend/sjs/JSCodeGen.scala#L130

odersky:
> To exercise this in bootstrap, you could remove all the : Position I had to add in JSCodeGen on local implicit val pos'es, for example this one:

@sjrd Done. I saw they were quite a few of those.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3378,7 +3378,7 @@ object Types {
 
     /** Map this function over given type */
     def mapOver(tp: Type): Type = {
-      implicit val ctx: Context = this.ctx // Dotty deviation: implicits need explicit type
+      implicit val ctx = this.ctx // Dotty deviation: implicits need explicit type
```
sjrd: Looks like the comment is obsolete :P

compiler/src/dotty/tools/dotc/config/PathResolver.scala
```diff
@@ -143,7 +143,7 @@ object PathResolver {
       println(Defaults)
     }
     else {
-      implicit val ctx: Context = (new ContextBase).initialCtx // Dotty deviation: implicits need explicit type
+      implicit val ctx = (new ContextBase).initialCtx // Dotty deviation: implicits need explicit type
```
sjrd: Another obsolete comment.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -46,7 +46,13 @@ object ErrorReporting {
             errorMsg(msg, cx.outer)
         }
       } else msg
-    errorMsg(ex.show, ctx)
+
+      if (cycleSym.is(Implicit, butNot = Method) && cycleSym.owner.isTerm)
+        em"""cyclic reference involving implicit $cycleSym
```
odersky: @felixmulder Line 2 and 3 of this message might be moved to an explanation. But we can also leave it like this, I think.

felixmulder:
@sjrd @odersky: note that the `JSCodeGen.scala` is not part of the sources being compiled in tests.

The sources for the Scala.JS IR were added as a source dependency (by sbt) and compiled by sbt with dotty compiler. The resulting class files were previously erroneously put on the classpath during bootstrap.

Therefore the Scala.JS backend is no longer compiled with the tests. Ask me today and I'll give a more thorough explanation in person.

felixmulder:
LGTM. Added a proper error message as per @odersky's recommendation.

odersky:
@felixmulder I decided to revert the last commit. I think moving the explanation away into a separate step is counter-productive here because the error message alone is too opaque to tell you anything. So I think in this case it's actually better to give the explanation as part of the error message.

felixmulder:
@odersky - that's fine by me :+1:

smarter:
Ping @olafurpg since this affect scalafix

olafurpg:
Thanks for the ping @smarter. I will update the ExplicitImplicit rewrite to allow optional type annotations for local implicit vals. To be sure I understand correctly, a val is considered local if and only if it's not defined inside a template?

```scala
class/trait/object A {
  // not OK
  implicit val x =???
  def/val/var foo = {
    // OK
   implicit val x = ???
  }
}
```

odersky:
On Thu, Dec 15, 2016 at 4:48 PM, Ólafur Páll Geirsson <

> Thanks for the ping @smarter <https://github.com/smarter>. I will update
> the ExplicitImplicit rewrite to allow optional type annotations for local
> implicit vals. To be sure I understand correctly, a val is considered local
> if and only if it's not defined inside a template?
>
> class/trait/object A {
>   // not OK
>   implicit val x =???
>   def/val/var foo = {
>     // OK
>    implicit val x = ???
>   }
> }
>
>
Yes, that's right. Some local values still need to be given types, namely
if their right-hand side requires an implicit search. But that can be
handled locally by dotty, I think.

PR #1801, opened 2016-12-14
Fix #1790: Change by-name pattern matching.

### LINKED ISSUE #1790: Whitelisting scala.util.Try fails to compile.
This fails to compile with the following error:
```scala
178 |      case Failure(e) => f(e)
    |                           ^
    |                           found:    T(e)
    |                           required: Throwable
    | 
```

Where the code is:
```scala
def transform[U](s: T => Try[U], f: Throwable => Try[U]): Try[U] =
    try this match {
      case Success(v) => s(v)
      case Failure(e) => f(e)
    } catch {
      case NonFatal(e) => Failure(e)
    }
```

odersky:
New implementation following the scheme ~~outlined in #1790~~.
Update: The new set of rules is in #1805.

Review by @liufengyun, since it affects the pattern matcher.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -24,7 +24,6 @@ object desugar {
 
   /** Names of methods that are added unconditionally to case classes */
   def isDesugaredCaseClassMethodName(name: Name)(implicit ctx: Context): Boolean =
-    name == nme.isDefined ||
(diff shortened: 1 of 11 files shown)


odersky:
@DarkDimius please take a look at the changes in PatternMatcher of the last commit. I think I figured out how to determine the arity of the match in all cases, but it would be good to have this double-checked.

DarkDimius:
LGTM, seems like the right way to do it.

PR #1807, opened 2016-12-15
Specialize Functions

felixmulder:
This PR is a bit of a work in progress, but I'd like feedback on the code from @DarkDimius and/or @odersky.

Todo:

- [x] Specialize unobscured Function1 subclasses
- [x] Test for presence of specialized apply
- [x] Test for absence of call to boxing runtime

PS: got stuck at Zurich airport, so haven't run the full test-suite yet. Letting our CI do the heavy lifting instead..


felixmulder:
Perhaps it would be a good idea to move these phases to after erasure, since

```scala
class Func1 extends Function1[Int, Int] {
  def apply(i: Int) = i + 1
}

// this cast will be typed as `Int => Int` until erasure
(new Func1: Function1[Int, Int])(1)
```

and the transformation doesn't catch it currently.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+    returnTypes = Set(defn.DoubleClass,
+                      defn.FloatClass,
+                      defn.IntClass,
+                      defn.LongClass)
+
+    Function1 = ctx.requiredClass("scala.Function1")
+
+    blacklisted = Set(
+      "scala.compat.java8.JFunction1",
+      "scala.runtime.AbstractFunction1",
```
DarkDimius: why do you need to blacklist those 3? As those functions don't know the actual type parameters of parent class, they shoundn't be specialized.

DarkDimius: I.e. Aren't they already caught by https://github.com/lampepfl/dotty/pull/1807/files#diff-3a80aa3cb499ce4bfec8753eabcb1796R74?

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private object ShouldTransformDenot {
+    def unapply(cref: ClassDenotation)(implicit ctx: Context): Option[(ClassDenotation, Type, Type, Type)] = {
+      def collectFunc1(xs: List[Type])(implicit ctx: Context): Option[(Type, Type, Type)] =
+        xs.collect {
+          case func1 @ RefinedType(RefinedType(parent, _, t1), _, r)
+            if func1.isRef(Function1) => (t1, r, func1)
+        }.headOption
+
+      collectFunc1(cref.info.parentsWithArgs).flatMap { case (t1, r, func1) =>
```
DarkDimius: use `baseTypeWithArgs` instead of rolling over all parents.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+          blacklisted.exists(sym => cref.symbol.derivesFrom(sym))
+        ) None
+        else Some((cref, t1, r, func1))
+      }
+    }
+  }
+
+  private object ShouldTransformTree {
+    def unapply(tree: Template)(implicit ctx: Context): Option[(Tree, Type, Type)] =
+      tree.parents
```
DarkDimius: use baseTypeWithArgs instead of rolling over all parents.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+      .map { t => (t.tpe, t) }
+      .collect {
+        case (tp @ RefinedType(RefinedType(parent, _, t1), _, r), func1)
+        if tp.isRef(Function1) &&
+           argTypes.contains(t1.typeSymbol) &&
```
DarkDimius: this can be extracted and shared with  `ShouldTransformDenot`

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+      }
+      .headOption
+  }
+
+  private def specializedName(name: Name, t1: Type, r: Type)(implicit ctx: Context): Name =
+    name.specializedFor(List(t1, r), List(t1, r).map(_.typeSymbol.name))
+
+  def transformDenot(cref: ClassDenotation, t1: Type, r: Type, func1: Type)(implicit ctx: Context): SingleDenotation = {
+    val specializedFunction: TypeRef =
+      ctx.requiredClassRef(functionPkg ++ specializedName(functionName, t1, r))
```
DarkDimius: `requiredClassRef` would fail compilation if there's no such class. I don't think it should be used here.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+      alteredScope.enter(specializedApply)
+      alteredScope
+    }
+
+    val ClassInfo(prefix, cls, parents, decls, info) = cref.classInfo
+    val newInfo = ClassInfo(prefix, cls, replaceFunction1(in = parents), specializeApply(decls), info)
+    cref.copySymDenotation(info = newInfo)
+  }
+
+  private def transformTree(tmpl: Template, func1: Tree, t1: Type, r: Type)(implicit ctx: Context) = {
```
DarkDimius: I'd define it inline, as you use it once and the name is less informative then TransformTemplate.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private def specializedName(name: Name, t1: Type, r: Type)(implicit ctx: Context): Name =
+    name.specializedFor(List(t1, r), List(t1, r).map(_.typeSymbol.name))
+
+  def transformDenot(cref: ClassDenotation, t1: Type, r: Type, func1: Type)(implicit ctx: Context): SingleDenotation = {
+    val specializedFunction: TypeRef =
+      ctx.requiredClassRef(functionPkg ++ specializedName(functionName, t1, r))
+
+    def replaceFunction1(in: List[TypeRef]): List[TypeRef] =
+      in.foldRight(List.empty[TypeRef]) { (tp, acc) =>
```
DarkDimius: you are recreating a list, removing all advantages of structural sharing.

DarkDimius: use mapConserve

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private def transformTree(tmpl: Template, func1: Tree, t1: Type, r: Type)(implicit ctx: Context) = {
+    val specializedFunc1 =
+      TypeTree(ctx.requiredClassRef(functionPkg ++ specializedName(functionName, t1, r)))
+
+    val parents = tmpl.parents.foldRight(List.empty[Tree]) { (t, trees) =>
+      (if (func1 eq t) specializedFunc1 else t) :: trees
+    }
+
+    val body = tmpl.body.foldRight(List.empty[Tree]) {
```
DarkDimius: use mapConserve

compiler/test/dotty/tools/dotc/transform/SpecializeFunction1Tests.scala
```diff
@@ -0,0 +1,94 @@
...
+
+      val applys = methods
+        .collect {
+          case m if m.name == "apply$mcII$sp" => m
+          case m if m.name == "apply" => m
+        }
+        .map(_.name)
+        .toList
+
+      assert(
```
DarkDimius: it would be also nice to check that the actual implementation is in the right apply.
Ie, `apply$mcII$sp` isn't a forwarder to generic `apply` that unboxes.

DarkDimius:
> Perhaps it would be a good idea to move these phases to after erasure, since

I don't think so. After erasure the information is already lost, and I don't see why you cant translate this block now...

>
>// this cast will be typed as `Int => Int` until erasure
>(new Func1: Function1[Int, Int])(1)

Where is a cast here? There shound be no cast needed here at all.

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -62,8 +62,10 @@ class Compiler {
            new ExplicitOuter,       // Add accessors to outer classes from nested ones.
            new ExplicitSelf,        // Make references to non-trivial self types explicit as casts
            new CrossCastAnd,        // Normalize selections involving intersection types.
-           new Splitter),           // Expand selections involving union types into conditionals
-      List(new VCInlineMethods,     // Inlines calls to value class methods
+           new Splitter,           // Expand selections involving union types into conditionals
+           new SpecializeFunction1), // Specialized Function1 by replacing super with specialized super
```
felixmulder: I was unsure of what impact the interleaved phases would have here. `DispatchToSpecializedApply` needs to know that the identifier type has a specialized apply before dispatching to it - so the addition to the decls (in `SpecializeFunction1`) needs to happen before this phase's transformation is applied.

compiler/src/dotty/tools/dotc/transform/DispatchToSpecializedApply.scala
```diff
@@ -0,0 +1,29 @@
...
+  import ast.Trees._
+  import ast.tpd
+
+  val phaseName = "dispatchToSpecializedApply"
+
+  override def transformApply(tree: tpd.Apply)(implicit ctx: Context, info: TransformerInfo) =
+    tree match {
+      case Apply(select @ Select(id, nme.apply), arg :: Nil) =>
+        val params = List(arg.tpe, tree.tpe)
+        val specializedApply = nme.apply.specializedFor(params, params.map(_.typeSymbol.name))
```
felixmulder: should this check be done in `specializedFor`?

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
```diff
@@ -0,0 +1,164 @@
...
+package dotc
+package transform
+
+import TreeTransforms.{ MiniPhaseTransform, TransformerInfo }
+import core._
+import Contexts.Context, Types._, Decorators._, Symbols._, DenotTransformers._
+import Denotations._, SymDenotations._, Scopes._, StdNames._, NameOps._, Names._
+import ast.tpd
+
+class SpecializeFunction1 extends MiniPhaseTransform with DenotTransformer {
```
felixmulder: we agreed to start there :)

But I could make it generic, albeit we only have specializations for `JFunction0`, `JFunction1` and `JFunction2`

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
felixmulder: I wanted to keep things apart so that it would be easier to get a quick overview of the workings of `SpecializeFunction1`, but perhaps it was taking it a bit far...

felixmulder:
The cast stops the optimization from applying

compiler/src/dotty/tools/dotc/transform/DispatchToSpecializedApply.scala
DarkDimius: yes) would be nice to have an extra boolean arg that checks if it should force creation.

compiler/src/dotty/tools/dotc/transform/SpecializeFunction1.scala
DarkDimius: yes:-)

DarkDimius:
but why is cast there in the first place? There shouldn't be any casts needed.

felixmulder:
Without the cast (as should be in normal code) - the optimization applies and all is good. With the cast the optimization doesn't kick in - so the question arises: should it?

felixmulder:
So - you're totally right, there aren't any casts needed. But *if* there's a cast. The optimization doesn't apply.

felixmulder:
@DarkDimius - changes discussed today incorporated into the PR.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+  val phaseName = "specializeFunction1"
+
+  // Setup ---------------------------------------------------------------------
+  private[this] val functionName = "JFunction".toTermName
+  private[this] val functionPkg  = "scala.compat.java8.".toTermName
+  private[this] var argTypes: Set[Symbol] = _
+  private[this] var retTypes: Set[Symbol] = _
+
+  override def prepareForUnit(tree: Tree)(implicit ctx: Context) = {
+    retTypes = Set(defn.UnitClass,
```
DarkDimius: why not take primitive value classes from ctx.definitons?
Same with argTypes.

compiler/test/dotty/tools/dotc/transform/SpecializeFunctionsTests.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      clsNode.methods.asScala
+        .find(_.name == "<init>")
+        .map { m =>
+          checkForBoxing(instructionsFromMethod(m), source)
+        }
+        .getOrElse(assert(false, "Could not find constructor for object `Test`"))
+    }
+  }
+}
```
DarkDimius: it would be nice to have more run-tests that check if classes&lambdas inherit the right specialized functions using reflection.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+                   defn.LongClass,
+                   defn.DoubleClass,
+                   /* only for Function1: */
+                   defn.FloatClass)
+    this
+  }
+
+  // Transformations -----------------------------------------------------------
+
+  /** Transforms all classes extending `Function1[-T1, +R]` so that
```
DarkDimius: Outdated comment?

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  private def specializedName(name: Name, args: List[Type], ret: Type)(implicit ctx: Context): Name = {
+    val typeParams = args :+ ret
+    name.specializedFor(typeParams, typeParams.map(_.typeSymbol.name), Nil, Nil)
+  }
+
+  // Extractors ----------------------------------------------------------------
+  private object ShouldTransformDenot {
+    def unapply(cref: ClassDenotation)(implicit ctx: Context): Option[Seq[SpecializationTarget]] =
```
DarkDimius: I think it's very simple implementation that doesn't warrant pattern-matching that allocates options.
Additionally, it's used only once. I'd inline it.

DarkDimius: If you want to keep it as a pattern, maybe consider using unapplySeq

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+    */
+  def transform(ref: SingleDenotation)(implicit ctx: Context) = ref match {
+    case cref @ ShouldTransformDenot(targets) => {
+      val specializedSymbols: Map[Symbol, (Symbol, Symbol)] = (for (SpecializationTarget(target, args, ret, original) <- targets) yield {
+        val arity = args.length
+        val specializedParent = ctx.getClassIfDefined {
+          functionPkg ++ specializedName(functionName ++ arity, args, ret)
+        }
+
+        val specializedApply: Symbol = {
```
DarkDimius: what if specializedParent doesn't exist?

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+        .specializedFor(typeParams, typeParams.map(_.typeSymbol.name), Nil, Nil)
+
+      val interface = ctx.getClassIfDefined(functionPkg ++ interfaceName)
+
+      if (interface.exists) Some {
+        SpecializationTarget(interface, args, ret, sym)
+      }
+      else None
+    }
+    .flatten.toList
```
DarkDimius: why toList. Do you care what iterable kind you get?
Conversion costs time and memory

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+  // Transformations -----------------------------------------------------------
+
+  /** Transforms all classes extending `Function1[-T1, +R]` so that
+    * they instead extend the specialized version `JFunction$mp...`
+    */
+  def transform(ref: SingleDenotation)(implicit ctx: Context) = ref match {
+    case cref @ ShouldTransformDenot(targets) => {
+      val specializedSymbols: Map[Symbol, (Symbol, Symbol)] = (for (SpecializationTarget(target, args, ret, original) <- targets) yield {
+        val arity = args.length
+        val specializedParent = ctx.getClassIfDefined {
```
DarkDimius: shoulnd't this class always exist?
Wasn't it already extracted in `targets`?

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+        specializedSymbols.values.foreach { case (_, apply) =>
+          alteredScope.enter(apply)
+        }
+        alteredScope
+      }
+
+      def replace(in: List[TypeRef]): List[TypeRef] =
+        in.map { tref =>
+          val sym = tref.symbol
+          specializedSymbols.get(sym).map { case (specializedParent, _) =>
```
DarkDimius: I this this is a very inefficient implementation.

Instead of building a `Seq[(Symbol, (Symbol, Symbol)]` and then converting it and operating it,
you can use the fact that both `specializedSymbols` and `parents` maintain the same order and traverse them together.

This will allow for an implementation that doesn't allocate any tuples, doesn't convert collections and traverses collections only once.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+            specializedParent.typeRef
+          }
+          .getOrElse(tref)
+        }
+
+      val ClassInfo(prefix, cls, parents, decls, info) = cref.classInfo
+      val newParents = replace(parents)
+      val newInfo = ClassInfo(prefix, cls, newParents, specializeApplys(decls), info)
```
DarkDimius: I think you don't need power of entire `DenotTransformer`. InfoTransformer will do.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  /** Transform the class definition's `Template`:
+    *
+    * - change the tree to have the correct parent
+    * - add the specialized apply method to the template body
+    * - forward the old `apply` to the specialized version
+    */
+  override def transformTemplate(tree: Template)(implicit ctx: Context, info: TransformerInfo) =
+    tree match {
+      case tmpl @ ShouldTransformTree(targets) => {
```
DarkDimius: same here.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,231 @@
...
+        val symbolMap = (for ((tree, SpecializationTarget(target, args, ret, orig)) <- targets) yield {
```
DarkDimius: same here, but even worse -> you flatten maps.

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /** Dispatch to specialized `apply`s in user code */
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo) = {
+    import ast.Trees._
+    tree match {
+      case Apply(select @ Select(id, nme.apply), arg :: Nil) => {
+        val params = List(arg.tpe, tree.tpe)
+        val specializedApply = specializedName(nme.apply, params)
+        val hasOverridenSpecializedApply = id.tpe.decls.iterator.exists {
```
DarkDimius: why iterate through all by hand instead of performing member lookup?
Why do you need to check if `sym is override`?

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,158 @@
...
+          sym => sym.is(Flags.Override) && (sym.name eq specializedApply)
+        }
+
+        if (hasOverridenSpecializedApply) tpd.Apply(tpd.Select(id, specializedApply), arg :: Nil)
```
DarkDimius: what's wrong about dispatching to an abstract one?

compiler/src/dotty/tools/dotc/transform/SpecializedApplyMethods.scala
```diff
@@ -19,122 +21,163 @@ class SpecializedApplyMethods extends MiniPhaseTransform with InfoTransformer {
...
+      specApply(func1, defn.UnitType,    List(defn.DoubleType)),
+      specApply(func1, defn.BooleanType, List(defn.DoubleType)),
+      specApply(func1, defn.IntType,     List(defn.DoubleType)),
+      specApply(func1, defn.FloatType,   List(defn.DoubleType)),
+      specApply(func1, defn.LongType,    List(defn.DoubleType)),
+      specApply(func1, defn.DoubleType,  List(defn.DoubleType))
+    )
+    func2 = defn.FunctionClass(2)
+    func2Applys = List(
+      specApply(func2, defn.UnitType,    List(defn.IntType, defn.IntType)),
```
smarter: Please use loops :sob:

felixmulder:
@DarkDimius - this is nearing the end now. I found a few bugs and the last piece of the puzzle is when compiling `tests/run/lift-and-unlift.scala`. I get:

```
*** error while checking tests/run/lift-and-unlift.scala after phase arrayConstructors ***

exception occurred while compiling tests/run/lift-and-unlift.scala
Exception in thread "main" java.lang.AssertionError: assertion failed: symbols differ for f3.apply$mcII$sp
was                 : method apply$mcII$sp
alternatives by type:  of types
qualifier type      : PartialFunction[Int, Int](f3)
tree type           : ((x$0: Int)Int)(f3.apply$mcII$sp) of class class dotty.tools.dotc.core.Types$TermRefWithSignature
```

Can't seem to figure this out, but perhaps we can take care of it on Tuesday.

DarkDimius:
This is a caching issue.
Enabling `checkCacheMembersNamed`  will show the real error:
```
Exception in thread "main" java.lang.AssertionError: assertion failed: cache inconsistency: cached: val <none>, computed method apply$mcII$sp, name = apply$mcII$sp, owner = class PartialFunction
```

Duhemm:
Superseded by #3306

PR #1811, opened 2016-12-15
Fix #1810 LazyVals should use changeOwnerAfter.

### LINKED ISSUE #1810: local lazy val containing case class fails to compile, issue with TreeTypeMap
I was asked to provide an example of `TreeTypeMap` being broken in some other phase than `ElimByName`, so here you go :).
```scala
class Test {
  def test: Unit = {
    lazy val x: Int = {
      case class Foo(x: Int)
      1
    }
  }
}
```
Fails with the mysterious:
```scala
constructor throws in class throws does not take type parameters
```
Stack trace after using `-Xprompt`: https://gist.github.com/d121067571cdcbf9bac26cf8028f6002

DarkDimius:


DarkDimius:
@odersky please review.
I still think that we need a stronger case before changing fundamentals such as TreeTypeMap.

odersky:
LGTM

PR #1813, opened 2016-12-16
Fix #1806: Define outer accessors at the right phase

### LINKED ISSUE #1806: Bug with inner classe inheritance.
The following code 
```scala
trait A {
  class Inner
 }  
trait B extends A {
  class Inner extends super.Inner
}
```
fails to compile with this error:
```scala
6 |               class Inner extends super.Inner
  |                                   ^^^^^^^^^^^
  |                            cyclic inheritance: class Inner extends itself
```
This pattern can be found in `scala-scala/src/library/scala/collection/parallel/ParSeqLike.scala` with `protected trait Accessor[R, Tp] extends super.Accessor[R, Tp]`.

Even worse, the compiler crashes when compiling
```scala
trait A {
  class Inner
 }  
trait B extends A {
  class Inner extends super.Inner
}
```

odersky:
Some outer accessors were defined at phase explicitOuter,
but were entered into the scope of their enclosing class only
at phase explicitOuter + 1. This turned them to stale symbols
when trying to access them at a later run, because at their
initially valid phase they were not found as members of
their owner.

Fixes #1806. Review by @nicolasstucki

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -24,7 +24,6 @@ object desugar {
 
   /** Names of methods that are added unconditionally to case classes */
   def isDesugaredCaseClassMethodName(name: Name)(implicit ctx: Context): Boolean =
-    name == nme.isDefined ||
     name == nme.copy ||
     name == nme.productArity ||
     name.isSelectorName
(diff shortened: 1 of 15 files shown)


odersky:
In fact this was based by accident on top of #1801.

nicolasstucki:
LGTM

PR #1815, opened 2016-12-16
Fix #1797: Allow case class params with names _1, _2, ...

### LINKED ISSUE #1797: Whitelisting scala.Tuple1 fails to compile.
The error message is:
```scala
19 |case class Tuple1[@specialized(Int, Long, Double) +T1](_1: T1)
   |                                                       ^
   |                                                       value _1 is already defined as method _1: => T1
```

The same happens with all the 22 tuples.

odersky:
This was not possible before because it clashed with the automatically
generated name of the accessor. We now allow it, by simply taking the
parameter(accessor) itself as the case class accessor if it already has
that name. But you still cannot write

    case class C(_2: Int, _1: String)

nor should you be able to do this.


nicolasstucki:
LGTM

PR #1819, opened 2016-12-17
Fix #1803: Infer type parameters of anonymous class parents from expected type

### LINKED ISSUE #1803: Whitelisting scala.collection.mutable.ObservableBuffer fails to compile.
Whitelisting `scala/collection/mutable/ObservableBuffer.scala` will fail with:
```scala
73 |    })
   |     ^
   |     found:    scala.collection.script.Reset with scala.collection.mutable.Undoable{...}
   |     required: scala.collection.script.Message[A] & scala.collection.mutable.Undoable
```

where 
```scala
trait ObservableBuffer[A] extends Buffer[A] with Publisher[Message[A] with Undoable] {
  ...
  abstract override def clear(): Unit = {
    super.clear()
    publish(new Reset with Undoable {
      def undo() { throw new UnsupportedOperationException("cannot undo") }
    })
  }
  ...
}

case class Reset[+A]() extends Message[A]
```

if the `Reset` (i.e. `publish(new Reset[A] with Undoable {...`) type parameter is added explicitly it compiles. It looks like the type parameter of `Reset` is not inferred.

The same applies for `scala/collection/mutable/ObservableMap.scala`, `scala/collection/mutable/ObservableSet.scala`

odersky:
If a parent type of an anonymous class is an Ident or Select which refers
to a parameterized type, use the expected type to infer its type parameters.

Fixes #1803.

Review by @nicolasstucki  ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -119,6 +119,18 @@ object Inferencing {
     }
   }
 
+  /** If `tree` has a PolyType, infer its type parameters by comparing with expected type `pt` */
+  def inferTypeParams(tree: Tree, pt: Type)(implicit ctx: Context): Tree = tree.tpe match {
+    case poly: PolyType =>
+      val (poly1, tvars) = constrained(poly, tree)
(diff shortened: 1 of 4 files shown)


nicolasstucki:
LGTM

PR #1821, opened 2016-12-17
Fix #1820: make sure outer of traits implemented

### LINKED ISSUE #1820: $outer of trait not implemented if the class doesn't need $outer
Following code runs in Scalac, but failed to run in Dotty. Inspecting the generated code shows that the `$outer` for `A.Inner` is not implemented in the top-leve class `Inner`.

```Scala
class A {
  val a = "a"
  trait Inner {
    def f = println(a)
    def h = 3
  }
}

class Inner extends O.a.Inner

object O {
  val a = new A

    (new Inner).f
  }
}
```

A fix is in order, will push soon.

liufengyun:
Fix #1820: make sure outer of traits implemented.

Review @DarkDimius


compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
```diff
@@ -81,14 +81,19 @@ class ExplicitOuter extends MiniPhaseTransform with InfoTransformer { thisTransf
         !needsOuterAlways(cls) &&
         impl.existsSubTree(referencesOuter(cls, _)))
       ensureOuterAccessors(cls)
-    if (hasOuter(cls)) {
+
+    val hasOuterFlag = hasOuter(cls)
```
odersky: Can we rename `hasOuterFlag` to `clsHasOuter`, or something like this?

I was reading `hasOuterFlag` as "has the Outer flag set" but of course there is no such flag.
Otherwise LGTM.

liufengyun: @odersky I've renamed `hasOuterFlag` to `clsHasOuter` in c2bc637 .

PR #1823, opened 2016-12-18
Fix #1789: Fix signature of unit result type

### LINKED ISSUE #1789: Whitelisting scala.sys.BooleanProp crashes the compiler.
If `scala/sys/BooleanProp.scala` is added to the whitelist `scala-collections.whitelist` the  with 
```scala
dotty.tools.dotc.reporting.diagnostic.messages$Error: method clear is already defined as getter clear: ()scala.runtime.BoxedUnit
```

`ConstantImpl` is implementing `clear`, `enable`, `disable`, `toggle` and fails on each one.

```scala
class ConstantImpl(val key: String, val value: Boolean) extends BooleanProp {
    ...
    val clear, enable, disable, toggle = ()
    ...
  }
```

where 
```scala
trait BooleanProp extends Prop[Boolean] {
  def enable(): Unit
  def disable(): Unit
  def toggle(): Unit
}
```

If the code in `ConstantImpl` is replaced with `def`s for each member it fixes the issue.

odersky:
was scala.BoxedUnit, should be scala.Unit. Fixes #1789. Review by @nicolasstucki ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2275,11 +2275,13 @@ object Types {
 
     protected def resultSignature(implicit ctx: Context) = try resultType match {
       case rtp: MethodicType => rtp.signature
-      case tp => Signature(tp, isJava = false)
+      case tp =>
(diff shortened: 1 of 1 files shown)


nicolasstucki:
LGTM

PR #1826, opened 2016-12-18
Make more parts of stdlib compile

odersky:


odersky:
The continuous-integration failure is because the drone does not pull updated versions of the scala-scala stdlib.

odersky:
@felixmulder There's a dottyDoc failure, which causes the junit test to fail:

error] Test dotty.tools.dottydoc.TestWhitelistedCollections.arrayHasDocumentation failed: java.lang.AssertionError: assertion failed: invalid prefix PolyType(List(1), List(T), List(TypeBounds(TypeRef(ThisType(TypeRef(NoPrefix,scala)),Nothing), TypeRef(ThisType(TypeRef(NoPrefix,scala)),Any))), RefinedType(RefinedType(TypeRef(ThisType(TypeRef(NoPrefix,scala)),PartialFunction), scala$PartialFunction$$A, TypeAlias(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,scala)),package)/withSig(Signature(List(),)),Throwable), -1)), scala$PartialFunction$$B, TypeAlias(PolyParam(T), 1))), took 8.569 sec
[error]     at scala.Predef$.assert(Predef.scala:165)
[error]     at dotty.tools.dotc.core.Types$NamedType.<init>(Types.scala:1386)
[error]     at dotty.tools.dotc.core.Types$TypeRef.<init>(Types.scala:1759)

Also, we need to pull in the latest scala-scala so that local CI succeeds.

smarter:
Could the CI run git pull on scala-scala first?

felixmulder:
It could - but better to have a version local to the image and then update
it. I'll add that soonish

odersky:
Rebased to master. Also, blacklisted SystemProperties again which apparently caused the dottydoc failure.

felixmulder:
The CI can now update scala-scala.

felixmulder:
Regarding the doc-tool failure - I'll have a look at it tomorrow morning and add the commit to this PR

felixmulder:
I tried enabling the SystemProperties and compiling it with dottydoc - doesn't seem to crash the compiler anymore. But there is a stack overflow when compiling std lib from dottydoc. Not sure if it is relevant to the PR currently. Let me know if there's anything else I can do to help out here.

odersky:
@felixmulder On validate-junit it still says:

[error] Test dotty.tools.dottydoc.TestWhitelistedCollections.arrayHasDocumentation failed: java.lang.AssertionError: assertion failed: invalid prefix PolyType(List(1), List(T), List(TypeBounds(TypeRef(ThisType(TypeRef(NoPrefix,scala)),Nothing), TypeRef(ThisType(TypeRef(NoPrefix,scala)),Any))), RefinedType(RefinedType(TypeRef(ThisType(TypeRef(NoPrefix,scala)),PartialFunction), scala$PartialFunction$$A, TypeAlias(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,scala)),package)/withSig(Signature(List(),)),Throwable), -1)), scala$PartialFunction$$B, TypeAlias(PolyParam(T), 1))), took 15.693 sec
[error]     at scala.Predef$.assert(Predef.scala:165)

Same for continuous-integration/drone/pr. I don't see a stackoverfloa anywhere (?).

But it might well be unrelated to sys/SystemProperties.

felixmulder:
Hmm, not sure - the CI (both of them) give different results then when I run it locally on my machine. On my machine I get three stack overflows. One when compiling the standard library (`compileStdLib`) and then twice again when running the dottydoc tests - which also compile the whitelist, so that seemed reasonable. On the CI it is instead as you say, some invalid prefix. This seems to be the offending line:

```
[error]     at dotty.tools.dottydoc.core.DocASTPhase.dotty$tools$dottydoc$core$DocASTPhase$$membersFromSymbol$1(DocASTPhase.scala:49)
```

Which is this line: https://github.com/lampepfl/dotty/blob/master/doc-tool/src/dotty/tools/dottydoc/core/DocASTPhase.scala#L49

I'm going to try it again with a clean repo.

odersky:
Current status: We have some bootstrap related issues, and at least two unrelated transform problems (one in tailrec, the other probably in elimByName). And then there are some files that compile only if there's no deep subtype check, we have to figure whether this is normal or a bug. 

But I think the current state is ready to go in. Who wants to give it a shot for reviewing?

smarter:
I'll review tomorrow

odersky:
@smarter I did not want to do that because it's ugly and inefficient, and, in 99.99% of cases, not needed.

odersky:
I could not find out what the problems with whitelisting Function1 and Function2 are. the junit test passes but partest fails with a cyclic reference error. Can somebody with more partest fu figure out where that comes from? 

Anyway, if we would compile Function1/2 as part of a bootstrap it would be done sequentially, so I don't see much of an issue. 

That means the only real oustanding issue are the three tailcall failures. It would be good to adress those.

odersky:
We get a project loading failed on continuous integration. @felixmulder any way to do the equivalent of /rebuild here?

smarter:
Fabien is on it.

felixmulder:
Fabien tells us it should be fixed now. Looking into it further. Restarting
the jobs is easy using either the drone CLI tool or from the web interface.

Cheers,
Felix

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1640,13 +1640,19 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-  /** Add apply node or implicit conversions. Two strategies are tried, and the first
-   *  that is successful is picked. If neither of the strategies are successful, continues with
-   *  `fallBack`.
+  /** Potentially add apply node or implicit conversions. Before trying either,
+   *  if the function is applied to an empty parameter list (), we try
+   *
+   *  0th strategy: If `tree` overrides a nullary method, mark the prototype
+   *                so that the argument is dropped and return `tree` itself.
+   *
+   *  After that, two strategies are tried, and the firs that is successful is picked.
```
smarter: typo: firs -> first

compiler/src/dotty/tools/dotc/typer/FrontEnd.scala
```diff
@@ -75,7 +84,11 @@ class FrontEnd extends Phase {
     }
     unitContexts foreach (parse(_))
     record("parsedTrees", ast.Trees.ntrees)
-    unitContexts.foreach(enterSyms(_))
```
smarter: What's wrong with foreach here?

odersky: We need to keep track of `remaining`.

PR #1831, opened 2016-12-19
Fix JUnit compilation of dotty by dotty

smarter:
This was broken in 06a3d47ea9fd1b67b3acba9d115a16d18549e377 when the
-deep argument was removed, meaning that dotty compiled with dotty was
not tested at all.


smarter:
Turns out that this actually still work! Review by @felixmulder

felixmulder:
LGTM

PR #1832, opened 2016-12-19
Fix #1812, Symbols.mapSymbols shouldn't replace denotations

### LINKED ISSUE #1812: Issue with Symbols#mapSymbols leads to various errors when using TreeTypeMap
Every usage of `TreeTypeMap` is affected, even if run using `atGroupEnd` like we do in `ExtensionMethods`, so this is a different issue than https://github.com/lampepfl/dotty/pull/1770:
```scala
class FF[R] {
  def compose(): R = ???
}

class Test(x: Int) extends AnyVal {
  def method: Unit = {
    class Bla
    class Foo extends FF[Bla] {
      override def compose() = super[FF].compose()
    }
  }
}
```
This fails `-Ycheck:tailrec` with:
```scala
java.lang.AssertionError: assertion failed:
found:    Bla
required: Bla'
where:    Bla  is a class in method method
          Bla' is a class in method method$extension

tree = super[FF].compose()
```

It's pretty hard to understand what's going on here but I think I've narrowed it down to [`mapSymbols`](https://github.com/lampepfl/dotty/blob/7775aab98ce98151d198e88bb676158f58f0cf57/compiler/src/dotty/tools/dotc/core/Symbols.scala#L302), recall that it works in three steps:
1. Create new naked symbols for each symbol in `original`
2. Set the denotation of each of the copies to the original denotation
3. Create a new denotation for each of the copy where we set their info using the `TreeTypeMap`, etc.

The issue happens in 3., to create the new infos we map over the old infos (`info = ttmap1.mapType(oinfo)`) for each symbol in order, but when we map over the info of one mapped symbol, we may end up looking at  another mapped symbol whose denotation is still set to the original denotation, even if we don't end up taking any decision based on that symbol denotation we may end up corrupting some cache (for example, `NamedType` cache denotations).

More specifically, in the example above:
1. `class Foo` is in a method that becomes an extension method, so its symbol ends up being transformed by `TreeTypeMap`
2. In [`TreeTypeMap#withMappedSyms`](https://github.com/lampepfl/dotty/blob/7775aab98ce98151d198e88bb676158f58f0cf57/compiler/src/dotty/tools/dotc/ast/TreeTypeMap.scala#L179) we call `mapSymbols` with every declaration in `Foo`, in the list of declarations `compose` appears before `FF$R`, this is important.
3. Inside `mapSymbols` we get to the step 3 explained above, we map the info of `compose` which is `()Foo.this.FF$R`, this is done successfully but note that at this point the denotation of the new symbol for `FF$R` is still incorrect, and `Foo.this.FF$R` is a `NamedType` which means that we cache its denotation.
4. Still inside `mapSymbols` we correctly set the denotation of the new symbol for `FF$R` but it's too late.

Later in `TreeChecker` when we look at the type of `super[FF].compose()` we get the old cached denotation which points to the old symbol for `Bla`, thus the error.

DarkDimius:
It will use lazy types instead. The current version transforms a type, with
    a context that has denotations that may be forcefully replaced by
    mapSymbols. Types created during this transformation may cache denots, that
    are-to-be replaced. This is very problematic as this method is called from
    TreeTypeMap.withMappedSyms in a fixed-point cycle, creating new symbols on
    every iteration. Those cached denotations could make types keep symbols
    from previous iterations indefinitely.

The changed version does not transform the types eagerly, and instead makes
    them lazy. Assuming there are no cycles, this should ensure correct
    ordering. Unfortunatelly, at this point in the compiler we basically always
    touch everything, and we can't even transform the info of denotation
    without this denotations info. We basically have a chicked&egg problem
    here. To solve it, I use the same trick as used by other lazy types by
    assigning an approximation of future type first. This allows to pass the
    tests and makes dotty more robust, but I suspect this isn't a complete fix
    and new similar bugs may arrive.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -313,10 +313,7 @@ trait Symbols { this: Context =>
             newNakedSymbol[original.ThisName](original.coord)
         }
       val ttmap1 = ttmap.withSubstitution(originals, copies)
-      (originals, copies).zipped foreach {(original, copy) =>
-        copy.denot = original.denot // preliminary denotation, so that we can access symbols in subsequent transform
-      }
-      (originals, copies).zipped foreach {(original, copy) =>
+      (originals, copies).zipped foreach { (original, copy) =>
         val odenot = original.denot
         val oinfo = original.info match {
           case ClassInfo(pre, _, parents, decls, selfInfo) =>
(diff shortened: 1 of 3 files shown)


smarter:
Could you add the testcase from #1812 to verify that the fix work?

DarkDimius:
@odersky please review.

odersky:
LGTM 👍

PR #1833, opened 2016-12-19
Fix #1793: allow multiversal comparisons between Null and X

### LINKED ISSUE #1793: `WeakReference` content can't be compared with `null`.
```scala
def unapply[T <: AnyRef](wr: WeakReference[T]): Option[T] = {
    val x = wr.underlying.get
    if (x != null) Some(x) else None
  }
```
fails with 
```scala
 33 |    if (x != null) Some(x) else None
    |       ^^^^^^^^^^^
    |       Values of types wr.underlying.java$lang$ref$WeakReference$$T and Null cannot be compared with == or !=
```

This happens when whitelisting `scala/ref/WeakReference.scala`.

felixmulder:
`Eq[Null, X]` and `Eq[X, Null]` should not force `X` to have an `Eq`-instance, it should be enough that `X` is not a subtype of AnyVal.

This PR amends this and adds the failing `WeakReference` sourcefile to the whitelist.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Implicits.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Implicits.scala
@@ -669,10 +669,28 @@ trait Implicits { self: Typer =>
             case result: AmbiguousImplicits => true
             case _ => false
           }
+
+        // Is Eq[X, Null] or Eq[Null, X] where !(X <:< AnyVal)?
+        def eqNullable(tp1: Type, tp2: Type): Boolean = {
+          val other =
+            if (tp1.stripTypeVar eq defn.NullType) tp2
+            else if (tp2.stripTypeVar eq defn.NullType) tp1
+            else NoType
+
+          (other ne NoType) && !other.derivesFrom(defn.AnyValClass)
+        }
+
         def validEqAnyArgs(tp1: Type, tp2: Type) = {
           List(tp1, tp2).foreach(fullyDefinedType(_, "eqAny argument", pos))
-          assumedCanEqual(tp1, tp2) || !hasEq(tp1) && !hasEq(tp2) ||
-            { implicits.println(i"invalid eqAny[$tp1, $tp2]"); false }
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -669,10 +669,28 @@ trait Implicits { self: Typer =>
...
+            else if (tp2.stripTypeVar eq defn.NullType) tp1
+            else NoType
+
+          (other ne NoType) && !other.derivesFrom(defn.AnyValClass)
+        }
+
         def validEqAnyArgs(tp1: Type, tp2: Type) = {
           List(tp1, tp2).foreach(fullyDefinedType(_, "eqAny argument", pos))
-          assumedCanEqual(tp1, tp2) || !hasEq(tp1) && !hasEq(tp2) ||
-            { implicits.println(i"invalid eqAny[$tp1, $tp2]"); false }
```
odersky: I'd leave it as it is and change assumedCanEqual instead, so that it returns true if one type is Null and the other is not a value type.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -669,10 +669,28 @@ trait Implicits { self: Typer =>
             case result: AmbiguousImplicits => true
             case _ => false
           }
+
+        // Is Eq[X, Null] or Eq[Null, X] where !(X <:< AnyVal)?
+        def eqNullable(tp1: Type, tp2: Type): Boolean = {
+          val other =
+            if (tp1.stripTypeVar eq defn.NullType) tp2
```
odersky: better: `tp1.isRef(defn.NullClass)`.

felixmulder:
@odersky - updated with your suggested changes :+1:

odersky:
LGTM 👍

PR #1834, opened 2016-12-19
Fix #1828: add warning when patternmatching on generics

### LINKED ISSUE #1828: Missing unchecked warning when pattern matching on a type parameter
```
def remove[S](a: S | Int, f: Int => S):S = a match {
  case a: S => a
  case a: Int => f(a)
}

val t: Int | String = 5
val t1 = remove[String](t, _.toString)
```

This compiles fine, but at runtime, the call to `remove` will throw a `java.lang.ClassCastException` (Integer cannot be cast to String)

felixmulder:
review by: @smarter

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -357,6 +357,9 @@ class Definitions {
       List(AnyClass.typeRef), EmptyScope)
   def SingletonType = SingletonClass.typeRef
 
+  lazy val ListType: TypeRef = ctx.requiredClassRef("scala.collection.immutable.List")
+  def ListClass(implicit ctx: Context) = ListType.symbol.asClass
+
   lazy val SeqType: TypeRef = ctx.requiredClassRef("scala.collection.Seq")
   def SeqClass(implicit ctx: Context) = SeqType.symbol.asClass
 
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
@@ -137,13 +137,14 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
           /** Check if the selector's potential type parameters will be erased, and if so warn */
           val selTypeParam = tree.args.head.tpe.widen match {
             case tp @ AppliedType(_, arg :: _) =>
-              // If the type is `Array[X]` where `X` extends AnyVal or `X =:=
-              // Any`, this shouldn't yield a warning:
-              val isArray = tp.isRef(defn.ArrayClass)
-              val unerased = arg.derivesFrom(defn.AnyValClass) || arg.isRef(defn.AnyClass)
-              val hasAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
+              // If the type is `Array[X]` where `X` extends AnyVal
+              val anyValArray = tp.isRef(defn.ArrayClass) && arg.derivesFrom(defn.AnyValClass)
+              // param is: Any | AnyRef | java.lang.Object
+              val topType = defn.ObjectType <:< arg
+              // has @unchecked annotation to suppress warnings
+              val hasUncheckedAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
 
-              if (!hasAnnot && !(isArray && unerased)) ctx.uncheckedWarning(
+              if (!topType && !hasUncheckedAnnot && !anyValArray) ctx.uncheckedWarning(
                 ErasedType(hl"""|Since type parameters are erased, you should not match on them in
                                 |${"match"} expressions."""),
                 tree.pos
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
@@ -136,7 +136,7 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
 
           // Check if the selector's potential type parameters will be erased, and if so warn
           val selTypeParam = tree.args.head.tpe.widen match {
-            case tp @ AppliedType(_, arg :: _) =>
+            case tp @ AppliedType(_, args @ (arg :: _)) =>
               // If the type is `Array[X]` where `X` is a primitive value
               // class. In the future, when we have a solid implementation of
               // Arrays of value classes, we might be able to relax this check.
@@ -145,6 +145,16 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
               val topType = defn.ObjectType <:< arg
               // has @unchecked annotation to suppress warnings
               val hasUncheckedAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
+
+              // Shouldn't warn when matching on a subclass with underscore
+              // params or type binding
+              val matchingUnderscoresOrTypeBindings = args.forall(_ match {
+                case tr: TypeRef =>
+                  tr.symbol.is(BindDefinedType)
+                case TypeBounds(lo, hi) =>
+                  (lo eq defn.NothingType) && (hi eq defn.AnyType)
+                case _ => false
+              }) && selector <:< scrutinee
 
               // we don't want to warn when matching on `List` from `Seq` e.g:
               // (xs: Seq[Int]) match { case xs: List[Int] => ??? }
@@ -160,13 +170,16 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
                 hasSameTypeArgs
               }
 
-              if (!topType && !hasUncheckedAnnot && !matchingSeqToList && !anyValArray) {
-                ctx.uncheckedWarning(
-                  ErasedType(hl"""|Since type parameters are erased, you should not match on them in
-                                  |${"match"} expressions."""),
-                  tree.pos
-                )
-              }
+              val shouldWarn =
+                !topType && !hasUncheckedAnnot &&
+                !matchingUnderscoresOrTypeBindings && !matchingSeqToList &&
+                !anyValArray
+
+              if (shouldWarn) ctx.uncheckedWarning(
+                ErasedType(hl"""|Since type parameters are erased, you should not match on them in
+                                |${"match"} expressions."""),
+                tree.pos
+              )
               true
             case _ =>
               if (tree.args.head.symbol.is(TypeParam)) {
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
@@ -146,11 +146,27 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
               // has @unchecked annotation to suppress warnings
               val hasUncheckedAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
 
-              if (!topType && !hasUncheckedAnnot && !anyValArray) ctx.uncheckedWarning(
-                ErasedType(hl"""|Since type parameters are erased, you should not match on them in
-                                |${"match"} expressions."""),
-                tree.pos
-              )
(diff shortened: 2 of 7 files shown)


compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -147,8 +147,11 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
 
           val inMatch = s.qualifier.symbol is Case
 
-          if (valueClassesOrAny) tree
-          else if (knownStatically)
+          if (valueClassesOrAny) {
+            if (selector eq defn.ObjectType)
+              ctx.warning(i"abstract type pattern is unchecked since it is eliminated by erasure", tree.pos)
```
smarter: This should be a Message class

smarter: `@unchecked` should turn the warning off

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -147,8 +147,11 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
 
           val inMatch = s.qualifier.symbol is Case
 
-          if (valueClassesOrAny) tree
-          else if (knownStatically)
+          if (valueClassesOrAny) {
+            if (selector eq defn.ObjectType)
```
smarter: Why is this limited to the selector being `Object`?

felixmulder: Check the definition of `valueClassesOrAny`. If the selector is also `Object` here, we are dealing with a generic type param that's been erased.

We should also check to see that it was not originally Object either.

smarter:
We also need warnings when pattern matching on something that will be erased like `List[Int]`

felixmulder:
@smarter: fixed your suggestions

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -126,6 +127,25 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
 
           val selClassNonFinal = selClass && !(selector.typeSymbol is Final)
           val selFinalClass    = selClass && (selector.typeSymbol is Final)
+          val selTypeParam     = tree.args.head.tpe.widen match {
+            case AppliedType(tycon, args) =>
+              ctx.uncheckedWarning(
+                ErasedType(hl"""|Since type parameters are erased, you should not match on them in
```
odersky: There's one exception to this, and that's arrays.

    case xs: Array[Int]

_is_ testable. We should (a) check that it is indeed tested and (b) adapt the error condition.

felixmulder: Good catch! I'll fix this ASAP :)

retronym:
A nuance of the Scalac implementation of this is that it allows you to match on List[Int] without warning if the expected type (ie, the scrutinee type) is Seq[Int]. 

There are a number of bugs that I've got half fixed on a branch, which will at least have some good test cases for you. I'll dig that up when I'm not on my phone.

felixmulder:
@retronym - thanks, that will surely come in handy!

With the nuanced implementation isn't it possible for the user to define a subtype of `Seq[Int]` upcast it and then match on `List[Int]`?

i.e:

```scala
class Foo extends Seq[Int] { ... }

(new Foo: Seq[Int]) match {
  case xs: List[Int] => ... // no warning
}
```

retronym:
@felixmulder The rationale is that either if the erased instanceof check succeeds, `List[Int]` is the correct type. Basically, it is a shorthand for saying that the folllowing cast can be statically proven to be sound.

```scala
def (x: Seq[Int]) = x match {
  case xs: List[_] is xs.length > 0 => ... x.asInstanceOf[List[Int]].head + 1
}
```

My WIP to better test and implement our unchecked warnings in in https://github.com/scala/scala/pull/4366

retronym:
One more thing: we can also silence the warning with `_: List[Int @unchecked]`

felixmulder:
@odersky - I think this is ready to go in once there's been a final review :)

odersky:
The current implementation gives a false warning for

    val ss: Seq[Int] = ???
    ss match {
      case ss: List[Int] => ???
    }

We get:

/Users/odersky/workspace/dotty/tests/neg> dotc match.scala -unchecked -explain
-- [E035] Erased Type Unchecked Warning: match.scala ---------------------------
5 |    case ss: List[Int] => ???
  |                       ^
  |      abstract type pattern is unchecked since it is eliminated by erasure

Explanation
===========
Since type parameters are erased, you should not match on them in match expressions.

This will not be easy to fix with the chosen approach. But it is essential that it is fixed. Otherwise, signalling false negatives will simply mean that people don't play attention to the real problems.

odersky:
One possible fix might be to refine the translation of the patterm matcher. Here, we currently see:

    val ss: scala.collection.Seq[Int] = ???
    {
      case val selector11: scala.collection.Seq[Int] = Test.ss
      {
        def case11(): Nothing = 
          {
            def matchFail11(): Nothing = throw new MatchError(selector11)
            if selector11.isInstanceOf[scala.collection.immutable.List[Int]]
               then
             
              {
                case val ss: scala.collection.immutable.List[Int] = 
                  selector11.asInstanceOf[
                    scala.collection.immutable.List[Int](ss)
                  ]
                {
                  {
                    ???
                  }
                }
              }
             else matchFail11()
          }

If the pattern matcher could figure out that the argument is always an `Int`, it could use

            if selector11.isInstanceOf[scala.collection.immutable.List[_]]

instead of currently:

            if selector11.isInstanceOf[scala.collection.immutable.List[Int]]

DarkDimius:
@odersky, 
I think this false negative can be fixed inside `IsInstanceOfEvaluator`, by observing that `List` derrives from `Seq` and that they share the same type argument.

odersky:
Rebased to master

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -127,6 +133,36 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
...
 
+          /** Check if the selector's potential type parameters will be erased, and if so warn */
+          val selTypeParam = tree.args.head.tpe.widen match {
+            case tp @ AppliedType(_, arg :: _) =>
+              // If the type is `Array[X]` where `X` extends AnyVal
+              val anyValArray = tp.isRef(defn.ArrayClass) && arg.derivesFrom(defn.AnyValClass)
+              // param is: Any | AnyRef | java.lang.Object
+              val topType = defn.ObjectType <:< arg
+              // has @unchecked annotation to suppress warnings
+              val hasUncheckedAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
```
odersky: Two of these conditions seem ad-hoc to me. Why exclude `Object` and arrays of `AnyVal` element types? I would suspect these are just symptoms and the real exclusion criteria are wider.

odersky:
@felixmulder Can you figure out why tests fail now? It seems we have some REPL test failures after rebasing.

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
felixmulder: The `anyValArray` check is to make sure we allow: `xs.isInstanceOf[Array[X]]` where `X` is a primitive. However, before it would have allowed any value class - I constrained it to only allow checking on primitives.

The `topType` check is to allow explicit checking on top types, eg: 

- `xs.isInstanceOf[Array[Any]]`
- `xs.isInstanceOf[Array[AnyRef]]`
- `xs.isInstanceOf[Array[Object]]`

felixmulder:
The tests were failing because after the rebase, the error number of `ErasedType` changed.

odersky:
Needs to be rebased to master. Also need to verify that 

    val ss: Seq[Int] = ???  
    ss match {
      case ss: List[Int] => ???
    }

does not give a spurious error. If that's the case LGTM.

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -147,7 +185,7 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
 
           val inMatch = s.qualifier.symbol is Case
```
odersky: This is not a correct discrimination for isInstanceOf's used in pattern matching! It would also cover case objects and enum cases. We need a more robust check. Maybe have a special, second `isInstanceOf` symbol?

felixmulder: @odersky: This logic was proposed by @DarkDimius when I implemented the original `IsInstanceOfEvaluator`, he could perhaps know of some reason why this isn't a problem. But the basic idea was that:

`val` bindings created inside a pattern match desugaring would get the `case` flag. This was so that we can know that they were specifically created by the `PatternMatcher`. 

I'd have to verify that this does indeed conflict with `case object`s. 

If you're sure (or if I get to verifying before you respond), then if I've understood your proposal correctly:

- `PatternMatcher` should be changed to emit a special `caseIsInstanceOf` symbol (or similar)
- `IsInstanceOfEvaluator` should intercept these and replace them with either regular `isInstanceOf` checks or the evaluated expressions.

    should it **exclusively** do this for the special symbol, or do we still want the optimization to be performed elsewhere?


Last - do you want this fix included in this PR or can it be separate? The PR doesn't otherwise touch the original logic. I'd open a second PR  with the implementation during the day so that we don't forget.

odersky: The `caseIsInstanceOf` idea looks right (exclusive or not: no idea, propose something :-) It could be a separate PR.

felixmulder:
@odersky - added the missing `Seq => List` nuance suggested by Jason. If all looks good, please approve and I'll move to fixing the `isMatch` logic.

felixmulder:
@odersky - actually one more caveat. `(x: Any) match { case List[_] => ??? }` still warns. I'll fix this, then we should be good to go.

felixmulder:
Rebased

compiler/src/dotty/tools/dotc/transform/IsInstanceOfEvaluator.scala
```diff
@@ -128,6 +134,67 @@ class IsInstanceOfEvaluator extends MiniPhaseTransform { thisTransformer =>
...
+              // has @unchecked annotation to suppress warnings
+              val hasUncheckedAnnot = arg.hasAnnotation(defn.UncheckedAnnot)
+
+              // Shouldn't warn when matching on a subclass with underscore
+              // params or type binding
+              val matchingUnderscoresOrTypeBindings = args.forall(_ match {
+                case tr: TypeRef =>
+                  tr.symbol.is(BindDefinedType)
+                case TypeBounds(lo, hi) =>
+                  (lo eq defn.NothingType) && (hi eq defn.AnyType)
```
odersky: `eq` usually works, but is not in general a reliable method to compare types, since hash consing is not guaranteed. I recommend to use instead:

    lo.isRef(defn.NothingClass) && hi.isRef(defn.AnyClass)

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -357,6 +357,9 @@ class Definitions {
       List(AnyClass.typeRef), EmptyScope)
   def SingletonType = SingletonClass.typeRef
 
+  lazy val ListType: TypeRef = ctx.requiredClassRef("scala.collection.immutable.List")
+  def ListClass(implicit ctx: Context) = ListType.symbol.asClass
+
```
odersky: This should not be needed (see below).

felixmulder:
Needs `subtypeWithArgs` before progress can be made.

felixmulder:
Since `IsInstanceOfEvaluator.scala` no longer exists, this PR is obsolete. Issue #1828 still remains to be solved in a different manner.

PR #1835, opened 2016-12-19
Add git pull scala-scala to list of test commands

felixmulder:
This should fix the issue with scala-scala not being updated

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,9 +1,10 @@
 pipeline:
   test:
-    image: lampepfl/dotty:0.2
+    image: lampepfl/dotty:latest
     pull: true
     commands:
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -1,9 +1,10 @@
 pipeline:
   test:
-    image: lampepfl/dotty:0.2
+    image: lampepfl/dotty:latest
     pull: true
     commands:
       - ln -s /var/cache/drone/scala-scala scala-scala
+      - (cd scala-scala && git pull)
```
DarkDimius: I think you shound't be reinventing the wheel and call 
`update DarkDimius scala scala-scala dotty-library`

defined scripts/common.

You'll get all the advantages of already existing system, the most important of which: it'll show last commits in the pulled repo.

PR #1840, opened 2016-12-20
Call graph with DCE

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/linker/callgraph/CallGraphBuilder.scala
```diff
@@ -0,0 +1,620 @@
...
+    if (ret1 eq null) {
+      // not yet computed
+      val upd = reachableTypes.items.filter(tp => tp.tp.member(x).exists)
+      typesByMemberNameCache.put(x, upd)
+      upd
+    } else ret1
+  }
+
+  private def addCast(from: Type, to: Type) = {
+    if (!(from <:< to) && to.classSymbols.forall(!_.derivesFrom(defn.NothingClass))) {
```
smarter: You probably want `frozen_<:<`

compiler/src/dotty/tools/dotc/transform/linker/callgraph/CallGraphBuilder.scala
```diff
@@ -0,0 +1,620 @@
...
+  }
+
+  private def registerParentModules(tp: Type, from: CallInfoWithContext): Unit = {
+    @tailrec def register(tp1: Type): Unit = {
+      if ((tp1 ne NoType) && (tp1 ne NoPrefix)) {
+        if (tp1.widen ne tp1) registerParentModules(tp1.widen, from)
+        if (tp1.dealias ne tp1) registerParentModules(tp1.dealias, from)
+        if (tp1.termSymbol.is(Module)) {
+          addReachableType(new TypeWithContext(tp1.widenDealias, parentRefinements(tp1.widenDealias)), from)
+        } else if (tp1.typeSymbol.is(Module, Package)) {
```
smarter: write `butNot = Package` instead of just `Package` otherwise it's confusing

DarkDimius:
by that time it shouldn’t matter: after typer everything is 
instantiated.

compiler/src/dotty/tools/dotc/transform/linker/callgraph/CallGraphBuilder.scala
DarkDimius: by that time it shouldn’t matter: after typer everything is instantiated.

compiler/src/dotty/tools/dotc/transform/linker/callgraph/CallGraphBuilder.scala
```diff
@@ -156,21 +160,20 @@ class CallGraphBuilder(collectedSummaries: Map[Symbol, MethodSummary], mode: Int
...
-        if (tp1.widen ne tp1) registerParentModules(tp1.widen, from)
-        if (tp1.dealias ne tp1) registerParentModules(tp1.dealias, from)
-        if (tp1.termSymbol.is(Module)) {
-          addReachableType(new TypeWithContext(tp1.widenDealias, parentRefinements(tp1.widenDealias)), from)
-        } else if (tp1.typeSymbol.is(Module, butNot = Package)) {
-          val t = ref(tp1.typeSymbol).tpe
-          addReachableType(new TypeWithContext(t, parentRefinements(t)), from)
-        }
-        register(tp1.normalizedPrefix)
+  private val registeredTypes = mutable.Set.empty[Type]
```
DarkDimius: does it speed up?
Actually you could use reachableTypes instead of this set.

nicolasstucki: Yes, it does. `reachableTypes` does not help much in this case.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -143,8 +144,12 @@ class ClassfileParser(
       setClassInfo(moduleRoot, staticInfo)
     }
 
-    classRoot.hack = result
-    moduleRoot.hack = result
+    result match {
+      case result @ Some(_: DottyUnpickler) =>
+        classRoot.dottyUnpickler = result.asInstanceOf[Option[DottyUnpickler]]
```
smarter: I see we have the same kind of problems :). https://github.com/dotty-staging/dotty/commit/b552b8353ef077991a0a5eb9eae11fb1cb8b8523

DarkDimius:
Not in time for release, rescheduling to next one.

nicolasstucki:
Closing this for now, it will need a bit of work while rebasing.

LifeIsStrange:
Any update? @nicolasstucki

vaclavsvejcar:
@nicolasstucki Any chance that this will make it into final _Scala 3_ release?

nicolasstucki:
Unfortunately, it will not be in the final *Scala 3* release.

vaclavsvejcar:
@nicolasstucki Are there any technical difficulties that need to be solved? Or something concrete where some help is needed?

nicolasstucki:
Mostly it was left out due to complexity. The call graph needs to know some information about how the code is compiled to bytecode. This is quite complex on its own it needs knowledge of the rest of the compiler. Additionally the current encoding are not stable and changing from version to version. If someone wants to revive this, I would strongly suggest waiting at least for Scala 3.1 to be out.

PR #1842, opened 2016-12-21
Added error messages - Parsers.scala:712

adamtrousdale:


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -5,10 +5,9 @@ package diagnostic
 
 import dotc.core._
 import Contexts.Context, Decorators._, Symbols._, Names._, NameOps._, Types._
-import util.{SourceFile, NoSource}
-import util.{SourcePosition, NoSourcePosition}
+import util.SourcePosition
```
felixmulder: This line shouldn't need to be added IIRC.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -5,10 +5,9 @@ package diagnostic
...
 import config.Settings.Setting
-import interfaces.Diagnostic.{ERROR, WARNING, INFO}
+import interfaces.Diagnostic.{ERROR, INFO, WARNING}
```
felixmulder: This one shouldn't need to be changed either.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -901,4 +900,23 @@ object messages {
     val msg = hl"trying to define package with same name as `$existing`"
     val explanation = ""
   }
+
+  case class ExistentialTypesNoLongerSupported()(implicit ctx: Context) extends Message(34) {
+    val kind = "Syntax"
+    val msg = "Existential types are no longer supported - use a wildcard or dependent type instead"
+    val explanation =
+      hl"""| The use of existential types is no longer supported and a wildcard or dependent type should be used instead
```
felixmulder: This row is too long, max should be 80 characters, split it so that it is shorter than 80 (preferably even shorter).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -901,4 +900,23 @@ object messages {
...
+           | For example:
+           |
+           | Instead of using ${"forSome"} to specify a type variable
+           |
+           | ${"List[T forSome { type T }]"}
+           |
+           | Try using a wildcard type variable
+           |
+           | ${"List[_]"}
+           |"""
```
felixmulder: Throughout this explanation, there's no need for there to be a padding of whitespace after the `|`-signs. None of the other messages should be doing this.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
adamtrousdale: These imports that were removed were unused, that's why I removed them. `util.SourcePosition` still needs to be included as it is in use in the MessageContainers.

I can put the imports back in if that's what you want

adamtrousdale: No the one below shouldn't have changed, I've fixed that. That might have been the IDE trying to tidy up for me

felixmulder:
LGTM - thanks for the PR! :tada:

PR #1847, opened 2016-12-21
Topic/533 parse error

b-studios:
This pull request deals with "Parsers.scala:533". It is part of #1589.

## Message Before

```
-- Error: examples/parser533.scala ---------------------------------------------
3 |} // error
  |^
  |`.' expected
-- [E025] Syntax Error: examples/parser533.scala -------------------------------
2 |  type X = FooBar22.this // error
  |           ^^^^^^^^^^^^^
  |           identifier expected

Explanation
===========
An identifier expected, but `FooBar22.this` found. This could be because
`FooBar22.this` is not a valid identifier. As a workaround, the compiler could
infer the type for you. For example, instead of:

def foo: FooBar22.this = {...}

Write your code like:

def foo = {...}



two errors found
```

## Message After
```
-- [E036] Syntax Error: examples/parser533.scala -------------------------------
2 |  type X = FooBar22.this // error
  |           ^^^^^^^^^^^^^
  |           Expected an additional member selection after the keyword this

Explanation
===========
Paths of imports and type selections must not end with the keyword this.

Maybe you forgot to select a member of this? As an example, in the
following context:
  trait Outer {
    val member: Int
    type Member
    trait Inner {
      ...
    }
  }

- this is a valid import expression using a path
  import Outer.this.member
  //               ^^^^^^^

- this is a valid type using a path
  type T = Outer.this.Member
  //                 ^^^^^^^


one error found
```

In addition to adding the required error message this pull request also fixes some inconsistencies with referring to identifier literals. While in the doc-comments one could find `Ident`, `ident`, `ident()`, `Id` and `id`, in the [spec](https://github.com/lampepfl/dotty/blob/master/docs/syntax-summary.txt) identifiers are always referred to as `id`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -540,7 +540,7 @@ object Parsers {
       def handleThis(qual: Ident) = {
         in.nextToken()
         val t = atPos(start) { This(qual) }
-        if (!thisOK && in.token != DOT) syntaxError(DanglingThisInPath(), start)
+        if (!thisOK && in.token != DOT) syntaxError(DanglingThisInPath(), t.pos)
         dotSelectors(t, finish)
       }
       def handleSuper(qual: Ident) = {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -540,7 +540,7 @@ object Parsers {
       def handleThis(qual: Ident) = {
         in.nextToken()
         val t = atPos(start) { This(qual) }
-        if (!thisOK && in.token != DOT) syntaxError("`.' expected")
+        if (!thisOK && in.token != DOT) syntaxError(DanglingThisInPath(), start)
         dotSelectors(t, finish)
       }
       def handleSuper(qual: Ident) = {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -504,7 +504,7 @@ object Parsers {
     def selector(t: Tree): Tree =
       atPos(startOffset(t), in.offset) { Select(t, ident()) }
 
-    /** Selectors ::= ident { `.' ident()
+    /** Selectors ::= id { `.' id }
      *
      *  Accept `.' separated identifiers acting as a selectors on given tree `t`.
      *  @param finish   An alternative parse in case the next token is not an identifier.
@@ -515,7 +515,7 @@ object Parsers {
       if (t1 ne t) t1 else dotSelectors(selector(t), finish)
     }
 
-    /** DotSelectors ::= { `.' ident()
+    /** DotSelectors ::= { `.' id }
      *
      *  Accept `.' separated identifiers acting as a selectors on given tree `t`.
      *  @param finish   An alternative parse in case the token following a `.' is not an identifier.
@@ -528,9 +528,9 @@ object Parsers {
     private val id: Tree => Tree = x => x
 
     /** Path       ::= StableId
-     *              |  [Ident `.'] this
+     *              |  [id `.'] this
      *
-     *  @param thisOK   If true, [Ident `.'] this is acceptable as the path.
+     *  @param thisOK   If true, the path can end with the keyword `this`.
      *                  If false, another selection is required after the `this`.
      *  @param finish   An alternative parse in case the token following a `.' is not an identifier.
      *                  If the alternative does not apply, its tree argument is returned unchanged.
@@ -565,20 +565,20 @@ object Parsers {
       }
     }
 
-    /** MixinQualifier ::= `[' Id `]'
+    /** MixinQualifier ::= `[' id `]'
     */
     def mixinQualifierOpt(): Ident =
       if (in.token == LBRACKET) inBrackets(atPos(in.offset) { typeIdent() })
       else EmptyTypeIdent
 
-    /** StableId ::= Id
-     *            |  Path `.' Id
+    /** StableId ::= id
+     *            |  Path `.' id
      *            |  [id '.'] super [`[' id `]']`.' id
      */
     def stableId(): Tree =
       path(thisOK = false)
 
-    /** QualId ::= Id {`.' Id}
+    /** QualId ::= id {`.' id}
     */
     def qualId(): Tree =
       dotSelectors(termIdent())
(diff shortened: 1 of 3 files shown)


b-studios:
@felixmulder @Blaisorblade I am happy to fix the conflicts, once the PR has been reviewed.

felixmulder:
Paolo is probably a good reviewer for this PR in my opinion :) so I'll delegate to him ;)

Blaisorblade:
I'll try but I can't reply very timely these days...

b-studios:
Sorry @Blaisorblade, for a split-second while mentioning you, I forgot how busy you are these days. Maybe someone else can review this PR? Cheers and happy holidays!

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -901,4 +901,31 @@ object messages {
     val msg = hl"trying to define package with same name as `$existing`"
     val explanation = ""
   }
+
+  case class DangelingThisInPath()(implicit ctx: Context) extends Message(34) {
```
Blaisorblade: `Dangeling` -> `Dangling`. (And given the delay, maybe recheck the ID 34).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -901,4 +901,31 @@ object messages {
...
+    val kind = "Syntax"
+    val msg = hl"""Expected an additional member selection after the keyword ${"this"}"""
+
+    val importCode =
+      """import MyClass.this.member
```
Blaisorblade: Maybe use `OuterClass` instead of `MyClass`? And even add `class OuterClass { class InnerClass { ... } }` as context to show the point of `MyClass.this`? I ask because it took me a (small) while to get the point `MyClass.this` (I'll concede I might also be rusty).

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -504,7 +504,7 @@ object Parsers {
     def selector(t: Tree): Tree =
       atPos(startOffset(t), in.offset) { Select(t, ident()) }
 
-    /** Selectors ::= ident { `.' ident()
+    /** Selectors ::= id { `.' id }
```
Blaisorblade: Good commit. Possible further idea: mention `id` in the comment for `isIdent`, also given they're not consistent? So authors of other comments can also find the right terminal name to use in comments?
(Well, being consistent might be even better but I'm not sure what to pick).
But that shouldn't hold this PR.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
Blaisorblade: Indeed we're up to 35, so right now you'll need a 36. https://github.com/lampepfl/dotty/blob/bf9888edcd96f40f11260e6ad874eb27382f77ba/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala#L926

Blaisorblade:
LGTM apart from the comments (and even those are mostly optional) — sorry for the delay, I hadn't realized the review would be so quick!

b-studios:
@Blaisorblade Thanks for the review. I provided some more context to the code examples in 21194b

Blaisorblade:
The CI failure is on http://dotty-ci.epfl.ch/lampepfl/dotty/475/1 at line 6164-6173.* That seems to have to do with a testcase from 18bc638c276c56e497bd39b58c9f3acf8b09f9e0 in #1850 — for some reason only 1 error is being output instead of 2.

Does this relate to switching error message from strings, or (in some complicated way) to adding `, start` as argument to `syntaxError`? I'm hoping the latter, but no clue why except if the `if` is skipped in 
https://github.com/b-studios/dotty/blob/065377cadd9634b6fb48da561f1f504082fd06f5/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L111-L112.

BTW, with that change the caret is not on `this` (line 6167 in logs), and I'm not sure I like that.

*OT for @jvican, assuming you picked Drone: I can't find a way to link to specific log lines in Drone logs like I'd do in these cases with Travis. Am I missing something? Or any chance you could please forward this feature request to the right people?

b-studios:
I verified that, in deed, only one error is produced in presence of `syntaxError(DanglingThisInPath(), start)`, while there are two with `syntaxError(DanglingThisInPath())`. This is the second error which is missing:

```scala
3 |}
  |^
  |Expected an additional member selection after the keyword this
```

felixmulder:
@Blaisorblade, @b-studios, haven't looked too closely at this, but FYI:

error messages are pruned by [UniqueMessagePositions.scala](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/reporting/UniqueMessagePositions.scala) and [HideNonSensicalMessages.scala](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/reporting/HideNonSensicalMessages.scala). If positions are changed when modifying error messages, this is where they would be pruned out.

The caret is placed on the `point` of a position if it is focused, otherwise it will be a sort of underline of carets from `pos.start`  to `pos.end`.

Regarding Drone, http://discourse.drone.io/ is probably the appropriate place to make a feature request. They also have a gitter: https://gitter.im/drone/drone

Cheers,
Felix

b-studios:
Thanks @felixmulder for the caret explanation. Changing the error call to `syntaxError(DanglingThisInPath(), t.pos)` will print as

```scala
2 |  type X = FooBar22.this // error
  |           ^^^^^^^^^^^^^
  |           Expected an additional member selection after the keyword this
```
@Blaisorblade Do you like that better?

However, the second error message "identifier expected" is still pruned away.

Blaisorblade:
> However, the second error message "identifier expected" is still pruned away since it is located at the same message position.

FWIW, looking at the two messages, I'd say pruning away "identifier expected" is an improvement :-), and the new position is much better.
In the future, I'd show the old error in the commit message to clarify the goal of the change.

TL;DR. I'd remove the second `// error` in the testcase (I'm guessing that comment marks expected errors), get a green build and merge.

> Regarding Drone, http://discourse.drone.io/ is probably the appropriate place to make a feature request. They also have a gitter: https://gitter.im/drone/drone

Thanks, done here:
http://discourse.drone.io/t/feature-request-allow-links-to-specific-log-lines/150

b-studios:
@felixmulder Build now is successful. Is there something more I can do (like squashing, etc.), or are we ready to merge?

felixmulder:
I'd like the merge commit rebased away - otherwise LGTM :+1:

b-studios:
@felixmulder I rebase the merge commit away and also squashed the typo-commit.

felixmulder:
Great, thanks a bunch! You're awesome :tada:

PR #1848, opened 2016-12-21
Fix #1795: Avoid infinite recursion between member and asSeenFrom

### LINKED ISSUE #1795: rectypes à la OCaml
In OCaml there is a compiler flag (-rectypes) to allow recursive types. If a use invoke the OCaml compiler with ```ocaml -rectypes test.ml``` I am able to compile the following ```test.ml``` file:

```OCaml
type 'a t1 = O1
type 'b t2 = O2
type t1' = t2' t1 and t2' = t1' t2

let x: t1' = O1
```

In Scala/Dotty this is unfortunately not possible:

```Scala
sealed trait T1 {type M1}

case object o1 extends T1 

sealed trait T2 {type M2}

case object o2 extends T2

package object Test {
  type TestT1 = T1 {type M1 = TestT2}
  type TestT2 = T2 {type M2 = TestT1}
  val x: TestT1 = o1
}
```

I am wondering if it would make sense to allow something like this (maybe be explicitly importing something like scala.language.RecursiveTypes).

I would like to do things like the following (which IMHO helps to be more modular, i.e., functions that work on ```('var, 'cond) expr``` can be used in more contexts):
```OCaml
module Types = struct
  type ('var, 'cond) expr = [ 
    `Var of 'var
  | `IfThenElse of 'cond * ('var, 'cond) expr * ('var, 'cond) expr
  ]

  type 'expr cond_basic = [
    `Eq of 'expr * 'expr
  ]

  type 'expr cond_ext = [
    `Neq of 'expr * 'expr
  ]

  type 'expr cond = [
    'expr cond_basic 
  | 'expr cond_ext
  ]
end

module Test1 = struct
  type var = string
  type expr = (var, cond) Types.expr
  and  cond = expr Types.cond

  let example_condition: cond = `Eq (`Var "Hallo", `Var "Welt")
  let example_condition: cond = `Neq (`Var "Hallo", `Var "Welt")
end

module Test2 = struct
  type var = int
  type expr = (var, cond) Types.expr
  and cond = expr Types.cond_basic

  let example_condition: cond = `Eq (`Var 5, `Var 6)
  (* This does not compile and should not compile 
   * let example_condition: cond = `Neq (`Var 5, `Var 6)
   *)
end
```

odersky:
Review by @smarter ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeOps.scala
@@ -10,7 +10,8 @@ import NameOps._
 import Decorators._
 import StdNames._
 import Annotations._
-import util.SimpleMap
+import config.Config
+import util.{SimpleMap, Property}
 import collection.mutable
 import ast.tpd._
 
@@ -67,7 +68,10 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
           if (thiscls.derivesFrom(cls) && pre.baseTypeRef(thiscls).exists) {
             if (theMap != null && theMap.currentVariance <= 0 && !isLegalPrefix(pre)) {
               ctx.base.unsafeNonvariant = ctx.runId
-              AnnotatedType(pre, Annotation(defn.UnsafeNonvariantAnnot, Nil))
+              pre match {
+                case AnnotatedType(_, ann) if ann.symbol == defn.UnsafeNonvariantAnnot => pre
+                case _ => AnnotatedType(pre, Annotation(defn.UnsafeNonvariantAnnot, Nil))
+              }
             }
             else pre
           }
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -67,7 +68,10 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
           if (thiscls.derivesFrom(cls) && pre.baseTypeRef(thiscls).exists) {
             if (theMap != null && theMap.currentVariance <= 0 && !isLegalPrefix(pre)) {
               ctx.base.unsafeNonvariant = ctx.runId
-              AnnotatedType(pre, Annotation(defn.UnsafeNonvariantAnnot, Nil))
+              pre match {
+                case AnnotatedType(_, ann) if ann.symbol == defn.UnsafeNonvariantAnnot => pre
```
smarter: Unrelated change?

compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -554,4 +560,10 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
 
 object TypeOps {
   @sharable var track = false // !!!DEBUG
+
+  /** When a property with this key is set in a context, it limit the number
+   *  of recursive member searches. If the limit is reached, findMember returns
+   *  NoDenotation.
+   */
+  val findMemberLimit = new Property.Key[Int]
```
smarter: Since `PendingFindMemberLimit` is a constant I would just make this a boolean

compiler/src/dotty/tools/dotc/core/TypeOps.scala
odersky: No it's related. Before we were running into the member/asSeenFrom mutual recursion we stackoverflowed here. Avoiding this by not piling up the same annotation makes sense independently.

smarter:
LGTM

PR #1849, opened 2016-12-21
Topic/syntax spec docs

b-studios:
This PR ports the syntax spec from this [tex document](https://github.com/lampepfl/dotty/blob/master/docs/syntax-summary.txt) to Markdown and fixes a couple of typos and minor bugs.

@smarter What do you think?


b-studios:
I purposfully reengineered the history to make it possible to review the changes. Of course I will squash the commits later to eventually merge.

b-studios:
The two changes that might require reviewing are bac774d and cdc4442.

smarter:
Awesome! You should also add a link to it from https://github.com/lampepfl/dotty/blob/master/docs/docs/index.md, also the tex version should be deleted to avoid duplication.

smarter:
> The two changes that might require reviewing are bac774d and cdc4442.

@odersky, could you have a look please?

b-studios:
Ah, one thing I was not sure about: I deleted the tree annotations that can be found in the tex version, like
```diff
-|  Refinement                                             RefinedTypeTree(EmptyTree, refinement)
+|  Refinement
```

Not sure, whether you want to keep them.

odersky:
Looks great. Definitely better to have this in markdown. 

I would leave the tree annotations in this version of the document, though. The reason is that they give us a good overview what parsed constructs map into. When we export the document to a language spec, they would be removed, of course.

odersky:
LGTM, Thanks!

PR #1852, opened 2016-12-22
fix #1750: check problematic owner first to avoid cyclic reference

### LINKED ISSUE #1750: crash on compilable code: cyclic reference involving class Visitor

$ cat abc.scala
```
trait Lang1 {
  trait Exp
  trait Visitor { def f(left: Exp): Unit }
  class Eval1 extends Visitor { self =>
    def f(left: Exp) = ()
  }
}
trait Lang2 extends Lang1 {
  class Visitor extends Eval1 { Visitor =>
  }
}
```

$ dotc abc.scala
```
exception while transforming () extends Lang2.this.Eval1() {} of class class dotty.tools.dotc.ast.Trees$Template # 174
exception while transforming class Visitor() extends Lang2.this.Eval1() {} of class class dotty.tools.dotc.ast.Trees$TypeDef # 175
exception while transforming () extends Object with Lang1 {
  class Visitor() extends Lang2.this.Eval1() {}
} of class class dotty.tools.dotc.ast.Trees$Template # 176
exception while transforming @scala.annotation.internal.SourceFile("abc.scala") <trait> trait Lang2() extends

Object with Lang1 {
  class Visitor() extends Lang2.this.Eval1() {}
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 177
  @scala.annotation.internal.SourceFile("abc.scala") <trait> trait Lang1()
     extends
   Object {
    <trait> interface trait Exp() extends Object {}
    <trait> interface trait Visitor() extends Object {
      def f(left: Lang1.this.Exp): Unit
    }
    class Eval1() extends Object() with Lang1.this.Visitor {
      def f(left: Lang1.this.Exp): Unit = ()
    }
  }
  @scala.annotation.internal.SourceFile("abc.scala") <trait> trait Lang2()
     extends
   Object with Lang1 {
    class Visitor() extends Lang2.this.Eval1() {}
  }
} of class class dotty.tools.dotc.ast.Trees$PackageDef # 178

Exception in thread "main" dotty.tools.dotc.core.Types$CyclicReference: cyclic reference involving class Visitor
	at dotty.tools.dotc.core.Types$CyclicReference$.apply(Types.scala:3838)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation$$anonfun$baseTypeRefOf$2.apply(SymDenotations.scala:1720)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation$$anonfun$baseTypeRefOf$2.apply(SymDenotations.scala:1709)
	at dotty.tools.dotc.reporting.Reporting$class.conditionalTraceIndented(Reporter.scala:132)
	at dotty.tools.dotc.core.Contexts$Context.conditionalTraceIndented(Contexts.scala:57)
	at dotty.tools.dotc.reporting.Reporting$class.debugTraceIndented(Reporter.scala:127)
	at dotty.tools.dotc.core.Contexts$Context.debugTraceIndented(Contexts.scala:57)
<snipped>
```

liufengyun:
Fix #1750: check problematic owner first to avoid cyclic reference.

I'm not sure if this is the right fix, could you please advise @odersky ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/Compiler.scala
+++ b/compiler/src/dotty/tools/dotc/Compiler.scala
@@ -44,13 +44,13 @@ class Compiler {
       List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
       List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
       List(new PostTyper),          // Additional checks and cleanups after type checking
+      List(new RefChecks,           // Various checks mostly related to abstract members and overriding
+           new CheckStatic),        // Check restrictions that apply to @static members
       List(new sbt.ExtractAPI),     // Sends a representation of the API of classes to sbt via callbacks
       List(new Pickler),            // Generate TASTY info
       List(new FirstTransform,      // Some transformations to put trees into a canonical form
            new CheckReentrant),     // Internal use only: Check that compiled program has no data races involving global vars
-      List(new RefChecks,           // Various checks mostly related to abstract members and overriding
-           new CheckStatic,         // Check restrictions that apply to @static members
-           new ElimRepeated,        // Rewrite vararg parameters and arguments
+      List(new ElimRepeated,        // Rewrite vararg parameters and arguments
            new NormalizeFlags,      // Rewrite some definition flags
            new ExtensionMethods,    // Expand methods of value classes with extension methods
            new ExpandSAMs,          // Expand single abstract method closures to anonymous classes
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -804,7 +804,18 @@ class RefChecks extends MiniPhase { thisTransformer =>
       checkOverloadedRestrictions(cls)
       checkParents(cls)
       checkCompanionNameClashes(cls)
-      checkAllOverrides(cls)
+
+      def ownerPossibleProblematic(cls: Symbol) = {
+        !cls.owner.is(Flags.PackageClass) && cls.owner.isClass &&
+          cls.owner.info.baseClasses.exists(_.info.member(cls.name).exists)
+      }
+
+      // If owner is possibly problematic, check owner first.
+      // If there are overriding errors with owner, stop checking current to avoid cyclic reference.
+      // See neg/i1750.scala.
+      if (ownerPossibleProblematic(cls)) checkAllOverrides(cls.owner)
+
+      if (!ctx.reporter.hasErrors) checkAllOverrides(cls)
       tree
     } catch {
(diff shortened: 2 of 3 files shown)


odersky:
I don't think /rebuild works. You have to restart from the web interface, clicking on Details first.

odersky:
@felixmulder I tried restarting from the web interface but that did not do anything either. I think it would be good to resurrect /rebuild. Sometimes it is really essential.

felixmulder:
We have an alternative way to restart tasks currently.

```
$ drone build start lampepfl/dotty <buildNbr>
```

The build number can be found in two ways, either by:

```
$ drone build list lampepfl/dotty
```

or by clicking through the web interface. I restarted this job by running the first command for build number 392.

felixmulder:
@liufengyun:

```
!!  86 - neg/i1750.scala                           [expected compilation failure]
```

liufengyun:
Thanks @felixmulder . I was wondering why it failed on drone, while pass locally -- the fix is exactly for #1750.

smarter:
I would advise against moving `ExtractAPI` without thinking deeply about the consequences. For example `ElimRepeated` is going to replace `T*` by `Seq[T]`, which means that if the user replaces `def foo(x: Foo*)` by `def foo(x: Seq[Foo])` in some source file, this will not be considered to be an API change, but this is wrong since it will break everything that tries to call `foo` assuming it takes a `Foo*`.

liufengyun:
@smarter the comment makes sense, a true concern. Let's try the other way around.

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -44,13 +44,13 @@ class Compiler {
       List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
       List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
       List(new PostTyper),          // Additional checks and cleanups after type checking
+      List(new RefChecks,           // Various checks mostly related to abstract members and overriding
```
felixmulder: Add a comment here for why it was moved up?

liufengyun: @felixmulder Thanks for looking at it. I tend to agree moving RefChecks before Pickler is a huge change we should try our best to avoid. Let's wait for some better fix that deprecate this one.

odersky:
@liufengyun Why

odersky:
@liufengyun What was the reason of moving RefChecks before ExtractDependencies? Did you see a crash in ExtractDependencies?

liufengyun:
@odersky : exactly as you noticed, there is `CyclicReference` exception in `ExtractApi`.

liufengyun:
superseded by #1913 , close this now.

PR #1855, opened 2016-12-23
Rewrite scala.annotation.internal.SourceFile in Java

smarter:
This fixes a NullPointerException when compile SourceFile.scala with the
sbt-based bootstrap (if the sbt bootstrap was already merged, you could
test this by running
  dotty-compiler-bootstrapped/run ./library/src/scala/annotation/internal/SourceFile.scala
).

The stack trace was
https://gist.github.com/smarter/b2126f895a8458506e58710c93bc9342

Review by @odersky


smarter:
Turns out, this isn't enough to avoid the NPE, I'll investigate more

PR #1858, opened 2016-12-24
Fix #1716: avoid crash in type checking f[_]

### LINKED ISSUE #1716: Exception while typechecking and compiling pattern matching with @
```
object Fail {
  case class So[+T]()

    So[Int]() match {
      case p@So[_] => Console.println("Yep")
      case _       => Console.println("Nope")
    }
  }
}
```

Compiler says,

```
exception occurred while typechecking /path/to/Src.scala
exception occurred while compiling /path/to/Src.scala
```

Works well if underscore in pattern matching is replaced with a specific type.

liufengyun:
Fix #1716: avoid crash in type checking `f[_]`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
@@ -368,7 +368,12 @@ trait TypeAssigner {
         }
         else {
           val argTypes = args.tpes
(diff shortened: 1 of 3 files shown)


odersky:
I think it's workable that way, but it's more consistent to do these things in Parser, where we already exclude other occurrences of wildcard arguments.

liufengyun:
Superseded by #1860 , close now.

PR #1859, opened 2016-12-25
Fix #1857: Interpolate type variables before implicit search

### LINKED ISSUE #1857: implicit summon fails in dotty, works in 2.*
The code below compiles and runs under 2.11.8 and 2.12.1, but fails under dotty with:

```
[info] Compiling 1 Scala source to /home/nmrp3/devel/turingatemyhamster/commandeer/target/scala-2.11/classes...
/home/nmrp3/devel/turingatemyhamster/commandeer/src/main/scala/commandeer/RunMe.scala:32: error: no implicit argument of type DSL found for parameter dsl of method apply in object CommandeerDSL
    val kevin = CommandeerDSL(null.asInstanceOf[Foo])
                                                     ^
/home/nmrp3/devel/turingatemyhamster/commandeer/src/main/scala/commandeer/RunMe.scala:34: error: value Bar is not a member of Nothing(kevin)
    val bar = kevin.Bar("bob", 3)
                    ^
two errors found
[error] (compile:compileIncremental) Compilation failed
```

I'm attempting to summon an implicit instance of a trait, but getting back type information specific to the instance that I've summoned, beyond that in the trait that I summoned it through. There may be a better way to do this with dotty, but I don't know it.

```scala
package commandeer


trait CommandeerDSL[Host] {
  trait Operation[T]
  type Op[T] <: Operation[T]
}

object CommandeerDSL {
  def apply[Host, DSL <: CommandeerDSL[Host]](host: Host)(implicit dsl: DSL): DSL = dsl
}
 
trait Foo {
  def bar(a: String, b: Int): Double
}
 
object Foo {
  implicit val fooDSL: FooDSL = new FooDSL {}
}
 
trait FooDSL extends CommandeerDSL[Foo] {
  sealed trait FooOperation[T] extends Operation[T]
  type Op[T] = FooOperation[T]

  case class Bar(a: String, b: Int) extends FooOperation[Double]
}
 
object RunMe {
    println("Hi Mum")

    val kevin = CommandeerDSL(null.asInstanceOf[Foo])
    println(s"Found DSL for Foo: $kevin")
    val bar = kevin.Bar("bob", 3)
    println(s"Made a bar: $bar")
  }
}
```

odersky:
Interpolating type variables before resolving implicit parameters means that
more companion objects can become eligible for the implicit scope. An example
demonstrating this is i1857.scala. The spec mandates this behavior because it
says that type arguments are inferred before implicit parameter resolution takes place.

Review by @smarter?

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -275,7 +275,7 @@ object Inferencing {
     if (constraint.uninstVars exists qualifies) interpolate()
   }
 
-  /** Instantiate undetermined type variables to that type `tp` is
+  /** Instantiate undetermined type variables so that type `tp` is
    *  maximized and return None. If this is not possible, because a non-variant
    *  typevar is not uniquely determined, return that typevar in a Some.
    */
@@ -302,7 +302,7 @@ object Inferencing {
    *  +1 means: only covariant occurrences
    *  0 means: mixed or non-variant occurrences
    *
-   *  Note: We intentionally use a relaxed version of variance here,
+   *  Note 1: We intentionally use a relaxed version of variance here,
    *  where the variance does not change under a prefix of a named type
    *  (the strict version makes prefixes invariant). This turns out to be
    *  better for type inference. In a nutshell, if a type variable occurs
@@ -311,6 +311,10 @@ object Inferencing {
    *     (U? >: x.type) # T
    *
    *  we want to instantiate U to x.type right away. No need to wait further.
+   *
+   *  Note 2: Parameters of implicit method types are assumed to be non-variant here.
+   *  This is necessary to prevent them from getting interpolated before an implicit
+   *  parameter search.
    */
   private def variances(tp: Type, include: TypeVar => Boolean)(implicit ctx: Context): VarianceMap = Stats.track("variances") {
     val constraint = ctx.typerState.constraint
@@ -324,6 +328,11 @@ object Inferencing {
           if (v == null) vmap.updated(t, variance)
           else if (v == variance || v == 0) vmap
           else vmap.updated(t, 0)
+        case t: ImplicitMethodType =>
+          val saved = variance
+          variance = 0
+          val vmap1 = try foldOver(vmap, t.paramTypes) finally variance = saved
+          apply(vmap1, t.resultType)
         case _ =>
           foldOver(vmap, t)
       }
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1697,8 +1697,13 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
 
   def adapt(tree: Tree, pt: Type, original: untpd.Tree = untpd.EmptyTree)(implicit ctx: Context): Tree = /*>|>*/ track("adapt") /*<|<*/ {
     /*>|>*/ ctx.traceIndented(i"adapting $tree of type ${tree.tpe} to $pt", typr, show = true) /*<|<*/ {
+      def shouldInterpolate(tp: Type) = tp match {
+        case tp: PolyType => false
+        case tp: MethodType => tp.isImplicit
+        case _ => true
+      }
       if (tree.isDef) interpolateUndetVars(tree, tree.symbol)
-      else if (!tree.tpe.widen.isInstanceOf[MethodOrPoly]) interpolateUndetVars(tree, NoSymbol)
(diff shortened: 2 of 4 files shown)


smarter:
I think this is a too early point to do the instantiation, for example it breaks:
```scala
object Test {
  def foo1[T](x: T)(implicit ev: T): Nothing = ???

  def test1: Unit = {
    implicit val ii: Int = 42
    implicit val ss: String = "foo"

    foo1(10) // ambiguous implicit because T=Any
  }
}
```
Whereas previously we correctly inferred `T=Int`. 

We already have some logic to deal with type variables that appear in implicit parameters' types: https://github.com/lampepfl/dotty/blob/119725799675ee00d6d3374771765b88e4de67bc/compiler/src/dotty/tools/dotc/typer/Typer.scala#L1846

I've simplified the testcase to show where things stop working:
```scala
class Wrapper[Host]


class SubWrapper extends Wrapper[Foo]

class Foo
object Foo {
  implicit val fooWrapper: SubWrapper = new SubWrapper
}


object RunMe {
    def wrap1[A, W <: Wrapper[Foo]](host: A)(implicit w: W): W = w
    def wrap2[A, W <: Wrapper[A]](host: A)(implicit w: W): W = w

    val f: Foo = new Foo

    val a = wrap1(f) // works with master
    val b = wrap2(f) // doesn't work with master
  }
}
```

In `wrap2(f)` we end up calling `instantiateSelected(W, List(A))`, `W` is bounded by `Wrapper[A]` but `IsFullyDefinedAccumulator` will not look at the bounds of `W` but stop because `force.appliesTo(W)` returns false. I think the correct fix involves tweaking `IsFullyDefinedAccumulator` but I'm not sure how.

odersky:
@smarter I have a fix based on tweaking `interpolateUndetVars`. I am not sure how to do it by tweaking `isFullyDefinedAccumulator`. The problem here is that in the original example `DSL` is not fully defined, (nor should it be), and `Host` does not even appear in the implicit search. Should `Host` be always inferred before doing the implicit search? I actually doubt that. In fact, I would argue that the implicit search will determine `DSL`, which will in turn determine `Host`, so inferring `Host` before the search is wrong!

With the last fix, we fix the incompatibility "by accident". Since `Host` does not appear anywhere in the type of the implicit method, it will be instantiated. By contrast, if the result type of the method contained `Host` in invariant position, it would not be instantiated, and the implicit would not be found.

I realize it is deeply unsatisfactory that the implicit scope depends on such subtleties of type inference. Maybe the right answer is to go the other way: Instance types of type variables never contribute to the implicit scope, only their bounds do. Then i1857.scala (and I suspect quite a few others) would give compile errors. We'd need to be sure we can catch them with @olafurpg 's scalafix. 

WDYT?

smarter:
> Maybe the right answer is to go the other way: Instance types of type variables never contribute to the implicit scope, only their bounds do. Then i1857.scala (and I suspect quite a few others) would give compile errors. We'd need to be sure we can catch them with @olafurpg 's scalafix.

I would advise against going in this direction, this sound similar to what we had a year and a half ago which lead to issues when using the standard library like https://github.com/lampepfl/dotty/issues/553 and more generally https://github.com/lampepfl/dotty/issues/739. Being able to add the implicits to the code using scalafix would help, but it would be deeply unsatisfying if we can't get implicit search to work in some simple uses of the standard library.

I like the current approach because it is more flexible than what `scalac` but in most common situations will produce the same result.

> Should Host be always inferred before doing the implicit search? I actually doubt that. In fact, I would argue that the implicit search will determine DSL, which will in turn determine Host, so inferring Host before the search is wrong!

I would say that yes, `Host` should be instantiated before doing the implicit search because:
1. It has been constrained by a previous parameter list
2. It is referred to indirectly by the type `DSL` (since `DSL <: CommandeerDSL[Host]`)

So we just need a way to collect all type variables which are referred to indirectly, and we already almost have this in the fixed point computation of `variances`: https://github.com/lampepfl/dotty/blob/119725799675ee00d6d3374771765b88e4de67bc/compiler/src/dotty/tools/dotc/typer/Inferencing.scala#L334-L335

I think we could replace `IsFullyDefinedAccumulator` by a call to `variances` and some processing of the result, this might require some tweaking of `variances`.

smarter:
I did not manage to unify `IsFullyDefinedAccumulator` and `variances` but here's an alternative fix that shows what I had in mind, let me know what you think: https://github.com/lampepfl/dotty/pull/1861

odersky:
I think #1861 is more promising than this one, so let's close it. I did not see how #739 would have been invalidated by the fix but #553 certainly does become invalidated. I still don't like the tight intergration between type variable integration and implicit scope but maybe that boat has already sailed.

PR #1860, opened 2016-12-25
Fix #1716: Don't allow wildcards as type arguments to methods

### LINKED ISSUE #1716: Exception while typechecking and compiling pattern matching with @

odersky:
Wildcards don't make sense as type arguments to methods, and I believe to keep things
simple this should also apply to method type arguments in patterns.

The best way to enforce this is to make use of the existing infrastructure in the parser
for topLevelTypes.

This is an alternate fix for #1716 more in line with the way we exclude wildcards in types already.

Review by @liufengyun ?


tests/pending/neg/i1716.scala
```diff
@@ -0,0 +1,9 @@
+object Fail {
```
liufengyun: Seems a slip here: move the test from `tests/pending/neg` to `tests/neg`

PR #1861, opened 2016-12-26
Alternative fix for #1857: Consider tvars in bounds of uninstantiated tvars

smarter:
This is an alternative to https://github.com/lampepfl/dotty/pull/1859, review by @odersky

<hr>

Given:
```scala
  val f: Foo = new Foo
  def wrap2a[A, W <: Wrapper[A]](host: A)(implicit w: W): W = ???
  wrap2a(f)
```

We need to instantiate `A` before doing the implicit search for `w`
because the type of `w` indirectly depends on `A` and `A` was constrained by the first parameter list. This was not
done before because we did not traverse the constraint bounds of an
uninstantiated type variable.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -79,8 +79,9 @@ object Inferencing {
       case _: WildcardType | _: ProtoType =>
         false
       case tvar: TypeVar
-      if !tvar.isInstantiated && ctx.typerState.constraint.contains(tvar) =>
-        force.appliesTo(tvar) && {
+        if !tvar.isInstantiated && ctx.typerState.constraint.contains(tvar) =>
+        val forceApplies = force.appliesTo(tvar)
+        if (forceApplies) {
           val direction = instDirection(tvar.origin)
           if (direction != 0) {
             //if (direction > 0) println(s"inst $tvar dir = up")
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -95,8 +96,13 @@ object Inferencing {
             if (minimize) instantiate(tvar, fromBelow = true)
             else toMaximize = true
           }
-          foldOver(x, tvar)
         }
+        foldOver(x && forceApplies,
```
odersky: I think this is a good fix, but the logic can be kept closer to the original one, and I think that's both shorter and clearer. How about:

        force.appliesTo(tvar) && {
           val direction = instDirection(tvar.origin)
            if (direction != 0) {
              //if (direction > 0) println(s"inst $tvar dir = up")
             instantiate(tvar, direction < 0)
           }
           else {
             val minimize =
               force.minimizeAll ||
               variance >= 0 && !(
                 force == ForceDegree.noBottom &&
                 defn.isBottomType(ctx.typeComparer.approximation(tvar.origin, fromBelow = true)))
              if (minimize) instantiate(tvar, fromBelow = true)
              else toMaximize = true
            }
            tvar.isInstantiated || foldOver(x, ctx.typerState.constraint.entry(tvar.origin))    
         }

odersky:
I agree it's better to do it this way compared to #1859, because `interpolateUndetVars` is really more an optimization and we should not hand semantics on it. There's still an open question, though: `IsFullyDefinedAccumulator` is called in several situations and only one is `instantiateSelected`. Are we sure it's the right behavior in the other calling contexts as well? We'd need to see an argument why that would be the case. 

Also, this has to be documented, in `isFullyDefinedAccumulator` and all callers.

smarter:
> IsFullyDefinedAccumulator is called in several situations and only one is instantiateSelected. Are we sure it's the right behavior in the other calling contexts as well? We'd need to see an argument why that would be the case.

I agree, we should try to come up with some examples where this makes a difference.

allanrenucci:
@smarter #1857 is fixed. Can we close this?

odersky:
I am closing for now.

PR #1864, opened 2016-12-31
Fix #1732: Special treatment for bottom type by-name args

### LINKED ISSUE #1732: AbstractMethodError on break.
```scala
object Test {
  import scala.util.control.Breaks
    Breaks.breakable {
      Breaks.break
    }
  }
}
```

```
Exception in thread "main" java.lang.AbstractMethodError: Test$$$Lambda$1/1018081122.apply$mcV$sp()V
	at scala.util.control.Breaks.breakable(Breaks.scala:38)
	at Test$.main(Test.scala:6)
	at Test.main(Test.scala)
```

odersky:
If a by-name arg has a bottom type, we need to create a closure with the
result type of the formal parameter, or else specialization with
FunctionalInterfaces will fail.

Review by @nicolasstucki ?


nicolasstucki:
LGTM

PR #1869, opened 2017-01-02
Fix #1868: Fix a logic error in a match

### LINKED ISSUE #1868: Crash when annotating type parameter of inner method

$ cat abc.scala
```scala
class Test {
  def outer: Unit = {
    def inner[@specialized T]() = {}
  }
}
```
$ dotc abc.scala
```

Exception in thread "main" scala.MatchError: Some(val <local Test>) (of class scala.Some)
	at dotty.tools.dotc.typer.Typer.annotCtx$lzycompute$1(Typer.scala:1147)
	at dotty.tools.dotc.typer.Typer.dotty$tools$dotc$typer$Typer$$annotCtx$1(Typer.scala:1145)
	at dotty.tools.dotc.typer.Typer$$anonfun$completeAnnotations$2.apply(Typer.scala:1157)
	at dotty.tools.dotc.typer.Typer$$anonfun$completeAnnotations$2.apply(Typer.scala:1157)
	at dotty.tools.dotc.typer.Typer.completeAnnotations(Typer.scala:1157)
<...>
```

odersky:


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1150,7 +1150,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
           // classes defined in a such arguments should not be entered into the
           // enclosing class.
           c.exprContext(mdef, exprOwner)
-        case None => c
```
smarter: This is a perfect example of why we shouldn't turn off exhaustiveness warnings when guards are used ;) (cf: https://github.com/scala/scala/pull/5617)

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1150,7 +1150,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
           // classes defined in a such arguments should not be entered into the
```
smarter: while you're at it, you could fix the typo here: "in a such" -> "in such"

odersky:
The other PR got merged, so this one can go in as well.

PR #1872, opened 2017-01-03
Fix compileStdLib test

nicolasstucki:
The issue was that all the files stdlib files where copied in `partest-generated/pos/compileStdLib` directly (i.e. directories flattend). This implies that files with the same names would overwrite each other (like `package.scala` and `Set.scala`).

### DIFF
--- a/compiler/test/dotc/scala-collections.blacklist
+++ b/compiler/test/dotc/scala-collections.blacklist
@@ -64,3 +64,20 @@
 #    |illegal redefinition of standard class AnyVal
 # (This is intended)
 
+../scala-scala/src/library/scala/collection/parallel/Tasks.scala
+# java.lang.StackOverflowError
+
+../scala-scala/src/library/scala/reflect/package.scala
+# 63 |  private[scala] def materializeClassTag[T](): ClassTag[T] = macro ???
+#    |                                                             ^^^^^
+#    |                                                             not found: macro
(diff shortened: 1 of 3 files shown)


compiler/test/dotc/tests.scala
```diff
@@ -245,7 +245,7 @@ class tests extends CompilerTest {
   }
 
 
-  @Test def compileStdLib = compileList("compileStdLib", stdlibFiles, "-migration" :: "-Yno-inline" :: scala2mode)
+  @Test def compileStdLib = compileList("compileStdLib", stdlibFiles.map(_.replace("../scala-scala/src/library/", "")), "-migration" :: "-Yno-inline" :: scala2mode, Directory("../scala-scala/src/library/"))
   @Test def compileMixed = compileLine(
```
liufengyun: This line looks a little strange, it seems the change can cancel each other? Otherwise, LGTM.

nicolasstucki: Updated with a less intrusive change.

liufengyun: I've a minor question here: for other calls of `compileList`, we ignore the option `sourceDir`. It would be nice to add a comment for `compileList` to document the condition when this option can be ignored.

liufengyun:
LGTM

PR #1879, opened 2017-01-05
Fix #1878: Generate fields for final vars.

### LINKED ISSUE #1878: Fields for final var not found.
When we declare a `final var` and then assign a new value we get a `NoSuchFieldError`.
```scala
object Test {
    NoStackTrace2._noSuppression = true
  }
}

object NoStackTrace2 {
  final var _noSuppression: Boolean = false
}
```

```scala
java.lang.NoSuchFieldError: _noSuppression$$local
	at NoStackTrace2$._noSuppression_$eq(test.scala:8)
	at Test$.main(test.scala:3)
	at Test.main(test.scala)
```

In the byte code we get:
```java
public final class NoStackTrace2$ {
    ...
    public boolean _noSuppression() {
        return false;
    }
    public void _noSuppression_$eq(boolean x$1) {
        this._noSuppression$$local = x$1;
    }
}
```

Which makes me think that we are handling `final var` as a `final val` at some point because the  `false` initial value was inlined in the getter and the field is not there.

nicolasstucki:


odersky:
LGTM!

odersky:
@smarter @nicolasstucki Actually, it should be the other way round. Final has no place indicating constness, we now have Inline for that. So what we should do is 

 - Make sure that final vals read from Scala 2 have the inline flag set.
 - As a temporary measure also set Inline for final vals in Dotty source, so that we can 
    for the moment stay compatible with Scalac.
 - Base all constness logic on inline, not final.

@nicolasstucki Can you roll back this PR and implement the scheme outlined above? I believe a lot of it is already implemented that way, but it would be good to check.

nicolasstucki:
Will do.

Small refinement to the first rule: it is only if the RHS of the final value is idempotent.

PR #1880, opened 2017-01-05
Improve whitelist infrastructure.

nicolasstucki:


nicolasstucki:
This PR aims to make it easer to rebase the callgraph branch and to unify all the whitelist logic in one place (previously repeated in 3 different sub-projects).

@DarkDimius, you should be aware of this change.

compiler/test/dotty/tools/StdLibSources.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import scala.io.Source
+
+object StdLibSources {
+
+  private final val stdLibPath = "../scala-scala/src/library/"
+
+  def blacklistFile: String = "./test/dotc/scala-collections.blacklist"
+
+  def whitelisted: List[String] = (all.toSet -- blacklisted).toList
```
odersky: Shorter: `all.diff(blackListed)`

odersky: I have one concern: For debugging, it would be good to have the old whitelist around. For many hard to find bugs I bisected whitelist to come up with a minimal failing set. With the new automatic way to get all files in scala's stdlib we can no longer do that. Can we have a config option like `useExplicitWhiteList` which, if set to true would use the old whitelist file instead of the computed one? Then to debug one could set this option to true and then get to work on the whitelist file for triaging.

nicolasstucki: Added `useExplicitWhiteList` in `StdLibSources.scala` to allow the use of a whitelist. In the current scheme if `useExplicitWhiteList == true` and the whitelist file does not exist, it will be generated from the blacklist. If it exists it will simply use it.

PR #1881, opened 2017-01-05
Implement structural type member access

odersky:
New scheme for implementing structural type member access. The main motivation is to use one scheme that works for Java reflection as well as for user-defined access methods, such as accessing database fields where rows are represented as hashmaps. 

Details will be explained in a separate issue.


tests/pos/zoo2.scala
```diff
@@ -0,0 +1,44 @@
...
+}
+def newLion: Lion = new {
+  type Food = Meat
+  def eats(food: Meat) = ()
+  def gets = newMeat
+}
+val milka = newCow
+val leo = newLion
+leo.eats(milka)
+}
```
liufengyun: The test error suggests that `import scala.reflect.Projector.reflectiveProjector` is missing in this test.

odersky:
The new scheme does not use implicit resolution, which makes it simpler and more robust. The old scheme suffered from the following drawback:

    import ... .reflectiveProjector
    val x: Any = new Record(...)
     x.name   // uses reflective access and fails, since records don't have a field called "name".

odersky:
I think this is ready for review now. Who wants to review?

liufengyun:
I'm happy to review. Maybe @nicolasstucki can also have a look, as it's related to `dynamic`?

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -629,6 +629,28 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
...
+   */
+  def isStructuralTermSelect(tree: Tree)(implicit ctx: Context) = tree match {
+    case tree: Select =>
+      def hasRefinement(qualtpe: Type): Boolean = qualtpe.dealias match {
+        case RefinedType(parent, rname, rinfo) =>
+          rname == tree.name && tree.tpe.widen <:< rinfo || hasRefinement(parent)
+        case tp: TypeProxy =>
+          hasRefinement(tp.underlying)
+        case tp: OrType =>
+          hasRefinement(tp.tp1) || hasRefinement(tp.tp2)
```
smarter: Aren't we mixing up the body of `OrType` and `AndType` here? to select `foo`, `foo` needs to be present in both sides of an `OrType`, but only one side of an `AndType`. In any case, we need some testcases for structural type selection on union/intersection types

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -26,6 +26,9 @@ object Definitions {
    *  else without affecting the set of programs that can be compiled.
    */
   val MaxImplementedFunctionArity = 22
+
+  /** The maximal arity of a function thta can be accessed as member of a structrual type */
```
smarter: typo: thta -> that

compiler/src/dotty/tools/dotc/typer/Dynamic.scala
```diff
@@ -100,4 +110,55 @@ trait Dynamic { self: Typer with Applications =>
...
+
+    def fail(reason: String) =
+      errorTree(tree, em"Structural access not allowed on method $name because it $reason")
+
+    tree.tpe.widen match {
+      case tpe: MethodType =>
+        if (tpe.isDependent)
+          fail(i"has a dependent method type")
+        else if (tpe.paramNames.length > Definitions.MaxStructuralMethodArity)
+          fail(i"takes too many parameters")
```
smarter: I would expand this message to say "Structural types only support methods taking up to ${Definitions.MaxStructuralMethodArity} arguments", to keep people from guessing.

library/src/scala/reflect/Selectable.scala
```diff
@@ -0,0 +1,73 @@
+package scala.reflect
+
+class Selectable(val receiver: Any) extends AnyVal with scala.Selectable {
```
smarter: I would call this `ReflectiveSelectable` or something like that, it's easy to miss the prefix.

smarter:
To expand on what @liufengyun suggested, currently `{ def get: Int}`, is syntactic sugar for `AnyRef { def get: Int }`, which means that the type is unusable without an implicit conversion like `reflectiveSelectable`, but I think it would make more sense to have `{ def get: Int }` be syntactic sugar for `Structural { def get: Int }`, this encourages people to not use the reflective stuff by default, but the issue is that when we do want to use reflection we would need an implicit conversion like this:
```scala
AnyRef { $REFINEMENTS }
         -->
Structural { $REFINEMENTS }
```
I don't think we can do that in the library by itself, but maybe with some compiler magic?

liufengyun:
It seems that we don't need reflection to implement `structural types`.

If we transform the method below:

```Scala
def f(x: { def get: Int}) = x.get
```

to 

```Scala
def f(x: Structural @{ def get: Int}) = x.selectDynamic("get").asInstanceOf[Int]
```

At the calling site, we can transform the call:

```
f(Some(5))
```

to the following:

```Scala
val x = Some(5)
val struct = new Structural {
    def selectDynamic(name: String): Any = 
        if (name == "get") x.get else error 
    }
}
f(struct)
```

Following the approach above, it seems we don't need reflection at all.

Of course, for backward-compatibility this is not a good idea.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
odersky: It needs to be present on both sides, but one side might be a regular class member. As long as one side comes from a refinement, the access is reflection based. For an AndType, it's the other way round. A single class member on one side is sufficient for regular access.

smarter: Ah I see! The `AndType` case makes sense, the `OrType` case seems suspicious but I don't think it matters since we no longer allow calling a method on a union type that matches more than one symbol, e.g. the following doesn't compile:
```scala
class Closeable {
  def close(): Unit = {}
}


object Test {
  import scala.reflect.Selectable.reflectiveSelectable

  def f(closeable: (Selectable { def close(): Unit }) | Closeable) =
    closeable.close() // error: value `close` is not a member of (Selectable{close: ()Unit} | Closeable)(closeable)

    f(new Closeable)
  }
}
```

library/src/scala/reflect/Selectable.scala
```diff
@@ -0,0 +1,73 @@
...
+            x5.asInstanceOf[Object])
+      case 7 => (x0: Any, x1: Any, x2: Any, x3: Any, x4: Any, x5: Any, x6: Any) =>
+        mth.invoke(receiver,
+            x0.asInstanceOf[Object],
+            x1.asInstanceOf[Object],
+            x2.asInstanceOf[Object],
+            x3.asInstanceOf[Object],
+            x4.asInstanceOf[Object],
+            x5.asInstanceOf[Object],
+            x6.asInstanceOf[Object])
```
nicolasstucki: Why do we require functions with fixed sizes? Could we not return an object with an `apply(args: Object*)` and then pass the args directly to the `invoke`?

odersky: The current solution has the advantage that the result of a Selectable is a drop in replacement of the original call. If we bunched arguments in an array, this would mean we also have the change the calling context. So it's more complicated. Selectable has lots of restrictions anyway, so one more does not really matter.

nicolasstucki: Ok.

Other related question: is the limit of 7 parameters an arbitrary limit or is there some factor that influenced this limit?

odersky:
> It seems that we don't need reflection to implement structural types.

Have a look at Gille Dubochet's thesis and at the "WhiteOak" paper he discusses. The WhiteOak paper uses essentially what you propose to avoid reflection. But the details are hairy. So hairy in fact, that the publicly available WhiteOak implementation does not follow the scheme described in their paper and uses reflection instead.

liufengyun:
Thanks for the pointer @odersky , I was thinking I must have missed something there :)

liufengyun:
Three code examples from [the paper](https://infoscience.epfl.ch/record/138931/files/2009_structural.pdf) that fail the reflection-based approach:

**Test1**: compiles, but runtime error
```Scala
import scala.reflect.Selectable.reflectiveSelectable

object Test1 {
  def g(x: Any{ def + (i: Int): Int }) = x + 2
  def main(args: Array[String]): Unit = g(3)
}
```

**Test2**: doesn't compile
```
import scala.reflect.Selectable.reflectiveSelectable

object Test2 {
  def g[T](x: { def f(a: T): Boolean }, t: T) = x.f(t)

  g[Int](new { def f(a: Int) = true }, 4)
  g[Any](new { def f(a: Any) = true }, 5)
}
```

**Test3**: compiler crash
```Scala
import scala.reflect.Selectable.reflectiveSelectable

object Test3 {
  def g(x: { type T ; def t: T ; def f(a: T): Boolean }) = x.f(x.t)
  g(new { type T = Int; def t = 4; def f(a:T) = true })
  g(new { type T = Any; def t = 4; def f(a:T) = true })
}
```

I don't think these cases are important, no one will ever write such code. Just document here for information.

odersky:
@liufengyun Can you make an issue from Test 3? we should track that one.

odersky:
@liufengyun 

> I have a concern about platform compatibility -- how can user use structural types in a platform-independent way(native, scala.js)?

I believe they would each have different implementations of `scala.reflect.Selectable`. That should do it, right?

odersky:
@liufengyun @smarter Regarding the meaning of `new {...}` I think there's a larger question: how are values of structural types generated? Right now we do whatever and then slap a cast to a structural type on it. There should be more systematic ways to do this, probably related to bijections from HMaps/HList. /cc @OlivierBlanvillain.

For the moment I would leave structural creation rules unchanged, waiting for learning more about the issues.

liufengyun:
@odersky 

>> I have a concern about platform compatibility -- how can user use structural types in a platform-independent way(native, scala.js)?

>I believe they would each have different implicit implementations of `scala.reflect.Selectable`. That should do it, right?

Yes, if a library is specially designed for `scala.js` or `native`, they can import the platform-specific implementation of `Selectable` implementation. However, if a library is intended for all platforms, it seems difficult to do that without changing the `import scala.xxx.Selectable` in library source code for each platform.

odersky:
@liufengyun 

> However, if a library is intended for all platforms, it seems difficult to do that without changing the import scala.xxx.Selectable in library source code for each platform.

That's precisely what I suggest: Change the implementation `scala.reflect.Selectable` for each platform. That's the only possible route.

odersky:
Any more suggestions, or can we merge?

PR #1883, opened 2017-01-06
Fix #1877: Add forwarders for primitive/generic mixins.

### LINKED ISSUE #1877: Missing bridge for abstract primitive method implemented with polymorphic method.
Found this issue while executing code on dotty compiled standard library, specifically on calling the method `value` on a `BooleanPropImpl`. The following code is a minimized version of the issue:

```scala
object Test {
    (new BooleanPropImpl2: BooleanProp2).value
  }
}

class BooleanPropImpl2 extends PropImpl2[Boolean](true) with BooleanProp2

trait BooleanProp2 {
  def value: Boolean
}

class PropImpl2[T](x: T) {
  def value: T = x
}
```

failing with 

```scala
java.lang.AbstractMethodError: Method BooleanPropImpl2.value()Z is abstract
	at BooleanPropImpl2.value(main.scala)
	at Test$.main(main.scala:8)
```

In the bytecode it is clear that class `PropImpl2` implements `public Object value()` and class `BooleanProp2` defines `boolean value();`, but there is no implementation of `boolean value()` in `BooleanPropImpl2`.

nicolasstucki:


tests/run/mixin-primitive-on-generic-1.scala
```diff
@@ -0,0 +1,19 @@
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    println((new Foo: Baz).value1)
+    println((new Foo: Baz).value2)
+  }
+}
+
+class Foo extends Bar[Boolean](true) with Baz
```
DarkDimius: need a test with a value class, also it would be nice to check that the method called would be the unboxed one.

nicolasstucki: Done in `mixin-primitive-on-generic-4.scala`and `mixin-primitive-on-generic-5.scala` and I checked that the method called is the unboxed one.

DarkDimius:
LGTM

PR #1885, opened 2017-01-06
Fix dummy scala-library

smarter:
To be useful, the dummy scala-library:
- needs to be published with crossPaths off (the "_2.11" path of the
artefact name), like the real scala-library
- should depend on dotty-library and not just scala-library, since this
is what is needed to compile dotty programs

Review by @felixmulder


felixmulder:
LGTM :tada:

PR #1889, opened 2017-01-08
Add dummy scala{-compiler,-reflect,p}

smarter:
This is needed to make dotty-compiled projects work with sbt 0.13.13, the
other needed fix is https://github.com/sbt/sbt/pull/2897

Review by @felixmulder


felixmulder:
Thanks :)

PR #1890, opened 2017-01-09
Stop emitting fields for inlined fields.

nicolasstucki:
Also move final val inlining logic to Inline as mentioned in #1879.


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1185,6 +1186,32 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+      case Block(_, t1) => skipBlocks(t1)
+      case _ => t
+    }
+    def isFinalInlinableVal(sym: Symbol): Boolean = {
+      sym.is(Final, butNot = Mutable) &&
+      isIdempotentExpr(vdef.rhs) /* &&
+      ctx.scala2Mode (stay compatible with Scala2 for now) */
+    }
+    val sym = vdef.symbol
+    skipBlocks(vdef.rhs) match {
```
odersky: No that's not right. The right hand side can also be an Ident and a Select with ConstantType.

odersky: I think the right body is:

       val sym = vdef.symbol
       sym.info match {
          case info: ConstantType if isFinalInlinableVal(sym) => sym.flags |= Inline
          case _ =>
       }

odersky:
Also, I think it's cleaner to revert the original commit(s) of #1879 and then add the new changes.

nicolasstucki:
Ok, I will revert the changes and clean up the history.

nicolasstucki:
Reverted  63d68bf and reimplemented cleanly.

odersky:
LTTM 👍

PR #1894, opened 2017-01-10
Partially fix Java interop for emitted inner classes

smarter:
The backend uses `rawname` to define the "inner name" of an InnerClass
entry in a classfile, this should be the simple name of the class before
any mangling takes place but before this commit, we put the mangled name
after flatten there.

Fixing this allows Java code to reference dotty inner classes, except if
they're defined in objects which is still broken until
https://github.com/DarkDimius/scala/pull/4 is merged and a new backend
is published.

Review by @DarkDimius


compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -560,7 +560,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def javaBinaryName: Name = toDenot(sym).fullNameSeparated("/") // addModuleSuffix(fullNameInternal('/'))
     def javaClassName: String = toDenot(sym).fullName.toString// addModuleSuffix(fullNameInternal('.')).toString
     def name: Name = sym.name
-    def rawname: Name = sym.name // todo ????
+    def rawname: Name = sym.name(ctx.withPhase(ctx.flattenPhase))
```
DarkDimius: should be initial phase, instead of flatten.

Issue #1895, opened 2017-01-11
Bringing a symbol to a new run is broken


smarter:
Summarizing what we discussed last month:
A common pattern in `InfoTransformer` like `ExtensionMethods` is to create new symbols like `foo$extension` in the current phase and to add them to the declarations of a class in the next phase. This causes problem when we try to bring forward the denotation corresponding to such a symbol to a subsequent run, because `stillValid` considers a denotation to be valid if it's valid in its owner at the first phase it's defined, but at its first phase the symbol has no owner yet!
Here is a testcase:
```scala
scala> class Foo(x: Int) extends AnyVal { def foo: Int = 1 }
defined class Foo
scala> new Foo(1).foo
Exception in thread "main" dotty.tools.dotc.core.Denotations$StaleSymbol: stale symbol; method foo$extension#33232 in module class Foo$, defined in Period(12..33, run = 4), is referred to in run Period(13..13, run = 6)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.staleSymbolError(Denotations.scala:932)
...
```

We can extend `stillValid` to also check the phase right after the first phase but that's still not enough because InfoTransformers are lazy and the InfoTransformer for the next phase might never have run in the initial run where the symbol was created! I've only been able to reproduce this using the sbt-based bootstrap of dotty by running `dotty-compiler-bootstrapped/test-only -- --tests=repl_all`. I currently work around this by explicitly forcing the denotation of extension methods at the phase where they're entered in their owner, but this is just a hack.

@DarkDimius suggested that we always recompute denotations from pickling, doing this may require a lot of work.

odersky:
The test case works now.

smarter:
The test case only works because of the workaround in https://github.com/dotty-staging/dotty/commit/51a458efeeebfeed6c357d56cf8afe5b06e86724, I believe the underlying problem is still here and could manifest itself in other ways.

smarter:
Reopening because I think this is important, but feel free to close with stat:revisit

PR #1896, opened 2017-01-11
Add sbt-based bootstrap

smarter:
This PR might be the hardest I've worked on just because of all the deep and subtle issues in dotty it forced me to figure out (especially #1770, #1856, #1895), but is now finally ready! ~(Well, almost, the backend needs to updated with https://github.com/DarkDimius/scala/pull/4 first and we need to get https://github.com/lampepfl/dotty/pull/1894 in, but other than that it should be complete)~.

This adds two new project to the sbt build: dotty-library-bootstrapped and dotty-compiler-bootstrapped. These projects use the same source files as dotty-library and dotty-compiler but are compiled using dotty itself. The main usecase for this is that we can now run the JUnit tests (which are _not_ just a subset of the tests run by partest, for example the REPL tests are only run through JUnit) with a bootstrapped compiler:

```
$ sbt
> publishLocal # Non-bootstrapped dotty needs to be published first
> dotty-bootstrapped/test
```

But this also allows one to experiment with a bootstrapped dotty in a much nicer way. This also replaces the previous bootstrap mechanism used for partest: `partest` and `partest-only` will now go through `dotty-compiler-bootstrapped`. Since partest bootstrapping has been broken for several months now, I consider that a win (we could also keep the old mechanism around but only if someone volunteers to maintain it).


smarter:
Review by @felixmulder for the sbt/tests related stuff and by @odersky for everything else :).

smarter:
Every day is a new adventure, apparently compiling dotty (with scalac) is enough to fill up the code cache of the JVM now? http://dotty-ci.epfl.ch/lampepfl/dotty/530/2
```scala
CodeCache: size=131072Kb used=124845Kb max_used=125852Kb free=6226Kb
 bounds [0x00007f0cd0d0f000, 0x00007f0cd8d0f000, 0x00007f0cd8d0f000]
 total_blobs=27273 nmethods=26706 adapters=462
 compilation: disabled (not enough contiguous free space left)
OpenJDK 64-Bit Server VM warning: CodeCache is full. Compiler has been disabled.
OpenJDK 64-Bit Server VM warning: Try increasing the code cache size using -XX:ReservedCodeCacheSize=
```

retronym:
> Every day is a new adventure, apparently compiling dotty (with scalac) is enough to fill up the code cache of the JVM now? http://dotty-ci.epfl.ch/lampepfl/dotty/530/2

We decided to handle bootstrapping outside of the SBT build in the Scala build for this reason. We use `sbt .... publishLocal; sbt -Dstarr.version=....` instead.

smarter:
I think I can handle it by restricting sbt task parallelism when I bootstrap (right now it's happily compiling/publishing everything at the same time)

smarter:
Oh, I think what's going is that it's compiling dotty twice at the same time, I thought sbt whole job was to not do that sort of things:
```scala
[info] Compiling 296 Scala sources and 1 Java source to /drone/src/github.com/lampepfl/dotty/compiler/target/scala-2.11/classes...
...
[info] Compiling 296 Scala sources and 1 Java source to /drone/src/github.com/lampepfl/dotty/compiler/target/scala-2.11/classes...
```

DarkDimius:
DarkDimius/scala#4 was merged

project/Build.scala
```diff
@@ -251,8 +242,8 @@ object DottyBuild extends Build {
 
       // enable verbose exception messages for JUnit
       testOptions in Test += Tests.Argument(
-        TestFrameworks.JUnit, "-a", "-v",
-        "--run-listener=dotty.tools.ContextEscapeDetector"
+        TestFrameworks.JUnit, "-a", "-v"/*,
+        "--run-listener=dotty.tools.ContextEscapeDetector"*/
```
felixmulder: Does this need to be disabled still?

project/Build.scala
```diff
@@ -371,17 +418,36 @@ object DottyBuild extends Build {
...
+    settings(
+      // FIXME: this breaks the CI (OutOfMemory errors) somehow
+      // The non-bootstrapped compiler needs to be published first.
+      // update <<= update.dependsOn(Def.task {
+      //   // This crazy stuff is equivalent to `dotty/publishLocal`, if we just did
+      //   // `publishLocal in dotty` the aggregated projects would be ignored.
+      //   val s = state.value
+      //   val e = Project.extract(s)
+      //   e.runAggregated(publishLocal in dotty, s)
+      // })
```
felixmulder: comment in empty settings?

smarter: Yeah I should comment the whole thing

project/Build.scala
smarter: Until we have a published new backend and https://github.com/lampepfl/dotty/pull/1894 is merged, yes. (I'll wait for these things to happen before merging this PR)

smarter:
Rebased to master which contains https://github.com/lampepfl/dotty/pull/1912 which was the last blocker to get this in

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -319,7 +319,17 @@ class Definitions {
...
+  // Dotty deviation: we cannot use a lazy val here because lazy vals in dotty
+  // will return "null" when called recursively, see #1856.
+  def DottyPredefModuleRef = {
+    if (_DottyPredefModuleRef == null) {
+      _DottyPredefModuleRef = ctx.requiredModuleRef("dotty.DottyPredef")
+      assert(_DottyPredefModuleRef != null)
+    }
+    _DottyPredefModuleRef
+  }
+  private[this] var _DottyPredefModuleRef: TermRef = _
```
odersky: Aside: Following @ichoran's example, I tend to use `myDottyPredefModuleRef` for private vars that have a getter `DottyPredefModuleRef`. But _... notation works as well.

compiler/test/dotty/partest/DPConsoleRunner.scala
```diff
@@ -398,8 +398,13 @@ class DPTestRunner(testFile: File, suiteRunner: DPSuiteRunner) extends nest.Runn
   override def extraClasspath =
     suiteRunner.fileManager.asInstanceOf[DottyFileManager].extraJarList ::: super.extraClasspath
 
+
+  // FIXME: Dotty deviation: error if return type is emitted:
```
odersky: emitted -> omitted

compiler/src/dotty/tools/dotc/core/Definitions.scala
smarter: OK, let's standardize on my

compiler/test/dotty/partest/DPConsoleRunner.scala
smarter: Fixed.

PR #1899, opened 2017-01-12
compiler support for syntactic macros

liufengyun:
## Summary

Add compiler support for scalameta-based metaprogramming:

- [x] **quasiquotes**
- [x] **annotation macros**
- [x] **syntactic def macros**

Engineering considerations:

- minimal changes to compiler (typer in particular)
- no performance penalty if macros are not used

This PR only contains the minimal changes to the compiler, the main implementation is located in [Eden](https://github.com/liufengyun/eden).

With this PR, a lot of syntactic def macros in ScalaTest can be defined.

### 1. A typer for meta block

A separate typer for typing meta block in order to minimize changes to the standard typer.

### 2. `macro` flag and inlining

In implementing type parameter support for def macros, I found `inlining` makes both the implementation of _inliner_ and _macros_ [complex](https://github.com/liufengyun/eden/issues/4).

Thus, I use the flag `<macro>` to tell whether a method call needs expansion or not. During transformation of macro definitions, I replace the flag `<inline>` with `<macro>` to avoid inlining from happening.

This avoids making changes to inliner to specially handle type parameters of def macros, also makes def macros implementation simpler.

If this scheme is acceptable, I would suggest introduce `macro` as a modifier to methods to denote def macros. *Inline methods* are themselves a category of useful and well-founded macros,  the keyword `inline` fits that purpose well.

## Examples

Following code is an annotation macro define case classes from a XML schema:

```Scala
@xsd("macros/tests/schema.xsd")
object schema

import schema._
val note = new Note("Vassily", "Pupkin", "hello", "This is a test!")
```

Following code demonstrates the usage of type parameters inside def macro definition:

```Scala
inline def is[T](a: Any): Boolean = meta {
  q"$a.isInstanceOf[$T]"
}

assert(scope.is[String]("hello"))
```

Review @xeno-by @odersky


smarter:
Cool stuff!
- There are no tests currently, I think good test coverage is really important here.
- Why is `eden` a separate project? What happens when it is not on the classpath?
- The standard Scala library contains two macros, `StringContext.f` and `reflect.materializeClassTag`,`ClassTag` is already handled specially by dotty, but we will crash on any code that uses the `f` interpolator. Can the work done in this PR help us support the `f` interpolator in dotty?

liufengyun:
> There are no tests currently, I think good test coverage is really important here.

There are a lot of tests located in Eden, I'll think how to test the features without introducing eden as a dependency.

> Why is eden a separate project? What happens when it is not on the classpath

Eden is a separate project, because we don't want the compiler to depend on scala.meta. If eden is not in classpath, it doesn't influence normal compilation. But macros will not be expanded, macro definitions will not be transformed.

> StringContext.f and reflect.materializeClassTag

I'll have a look at  that. My hunch is that`StringContext.f` is possible, `reflect.materializeClassTag` can't be handled if it depends on type information.

smarter:
> reflect.materializeClassTag can't be handled if it depends on type information.

I don(t think it should be necessary anyway because we already handle classtags in typer: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Implicits.scala#L528

smarter:
> But macros will not be expanded, macro definitions will not be transformed.

Shouldn't we at least print an error explaining how to enable eden?

smarter:
> I'll think how to test the features without introducing eden as a dependency

It could be a dependency only for the tests:
```
`libraryDependency += "org.foo" % "project" % "0.1" % "test"
```

smarter:
Could you add a doc page to https://github.com/lampepfl/dotty/tree/master/docs/docs/usage explaining how to use the macro support? It doesn't have to be very long, even a basic howto would be very helpful.

liufengyun:
> Shouldn't we at least print an error explaining how to enable eden?

I agree, an error message is required there.

> Doc on how to use the macro support

Sure, I'll do in a separate PR  once this gets merged, as I need to update Eden as well -- there's a mutual dependency. Currently I've a fork of Dotty, which Eden depends on.

>  we already handle classtags in typer

The implementation of macros is based on scala.meta, where the macro definition has no access to the `context`. It's possible to make `context` available, but that needs a change in the specification.  However, this feature may cause confusion to programmers, as now in the meta body they can have both scala.meta trees and dotty trees.

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -39,12 +39,14 @@ class Compiler {
...
    *     plain SymDenotation, as opposed to a UniqueRefDenotation.
    */
-  def phases: List[List[Phase]] =
+  def phases(implicit ctx: Context): List[List[Phase]] =
     List(
       List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
       List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
       List(new PostTyper),          // Additional checks and cleanups after type checking
       List(new sbt.ExtractAPI),     // Sends a representation of the API of classes to sbt via callbacks
+      List(macros.newTransform:_*), // Transform macros definitions, may be Nil when scala.meta.eden not in -toolpath
```
xeno-by: In scalac, we've been saying `macro definitions`, `macro implementations`, etc, using a singular noun "macro" if it precedes another noun.

compiler/src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -389,6 +389,30 @@ object Contexts {
...
+    lazy val classloader: ClassLoader = {
+      val classpath = ctx.platform.classPath.asURLs.toArray
+      new java.net.URLClassLoader(classpath, getClass.getClassLoader)
+    }
+
+    /** Are macros enabled? */
+    lazy val macrosEnabled: Boolean = {
+      try {
+        Class.forName("scala.meta.eden.package", false, toolClassloader)
+        Class.forName("scala.meta.Tree", false, classloader)
```
xeno-by: Do we need to explicitly check for presence of `scala.meta.Tree`? Is it even possible to write a meta block (and, consequently, use it) without referencing scala.meta?

compiler/src/dotty/tools/dotc/macros/Typer.scala
```diff
@@ -0,0 +1,95 @@
...
+ *    T : scala.meta.Type
+ *    a : scala.meta.Term
+ *
+ *  If it's an annotation macro, e.g.
+ *
+ *    class main extends StaticAnnotation {
+ *      inline def apply(defn: Any): Any = meta { ... }
+ *    }
+ *
+ *  The parameter `defn` gets the type `scala.meta.Tree` inside the meta block.
```
xeno-by: I think that the current implementation in scalameta/paradise uses `scala.meta.Stat`.

compiler/src/dotty/tools/dotc/macros/Typer.scala
```diff
@@ -0,0 +1,95 @@
...
+ *  Inside a meta block, e.g.
+ *
+ *    inline def is[T](a: Any): Any = meta {
+ *      q"$a.isInstanceOf[$T]"
+ *    }
+ *
+ *  The variables will get following types:
+ *
+ *    T : scala.meta.Type
+ *    a : scala.meta.Term
```
xeno-by: Could you also mention `this` here?

compiler/src/dotty/tools/dotc/macros/Typer.scala
```diff
@@ -0,0 +1,95 @@
...
+      tryEither {
+        implicit ctx => {
+          val name = if (tree.name.isTermName) tree.name.toTypeName else tree.name
+          super.typedIdent(untpd.cpy.Ident(tree)(name = name), pt)
+        }
+      } { (_, _) =>
+        failedState.commit()
+        failedVal
+      }
+    }
```
xeno-by: Am I right in thinking that this logic makes sure that, inside meta blocks, we can refer to type parameters via term names?

compiler/src/dotty/tools/dotc/macros/macros.scala
```diff
@@ -0,0 +1,95 @@
...
+  def expandAnnot(mdef: MemberDef)(implicit ctx: Context): Tree =
+    forObject("scala.meta.eden.expand.package").apply("expandAnnot", mdef, ctx).asInstanceOf[Tree]
+
+  /** Does the given definition have annotation macros? */
+  def hasAnnotMacros(mdef: MemberDef)(implicit ctx: Context): Boolean = {
+    mdef.mods.annotations.filter(isAnnotMacros).headOption.nonEmpty
+  }
+
+  /** Is the given annotation an annotation macro? */
+  def isAnnotMacros(ann: untpd.Tree)(implicit ctx: Context): Boolean = {
```
xeno-by: How about calling this `isAnnotMacro`?

compiler/src/dotty/tools/dotc/macros/macros.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /** Is the given symbol `scalameta.package.apply`? */
+  def isMetaApply(sym: Symbol)(implicit ctx: Context): Boolean = {
+    val meta = ctx.requiredModule("scala.meta.package")
+    meta.info.member("apply".toTermName).symbol == sym
+  }
+
+  def metaTreeType(implicit ctx: Context): Type = {
+    ctx.requiredClassRef("scala.meta.Tree")
```
xeno-by: I noticed that in this pull request some `ctx.requiredClassRef` calls are sprinkled around in random places and some are pulled into dedicated methods. What's the convention with respect to that in the Dotty compiler? In scalac, we typically put stuff like that in a dedicated object called `definitions`.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -631,12 +631,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+              val treeTpd = convertNewGenericArray(ConstFold(app.result))
+
+              // expand def macros after type checking
+              // Note: macros that expand to functions cannot be immediately applied.
+              // This is a practical constraint, as currying is also disallowed for macros
+              if (fun1.symbol.is(Flags.Macro) && !pt.isInstanceOf[ApplyingProto]) {
+                if (ctx.macrosEnabled)
+                  typed(macros.expandDef(treeTpd), pt)
+                else
+                  errorTree(treeTpd, s"can't expand the macro ${fun1.symbol.show}, make sure `Eden` is in -classpath")
```
xeno-by: Do we need to have eden on classpath now that https://github.com/scalameta/scalameta/pull/585 is merged?

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -631,12 +631,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+                else
+                  new ApplyToUntyped(tree, fun1, funRef, proto, pt)(argCtx(tree))
+              val treeTpd = convertNewGenericArray(ConstFold(app.result))
+
+              // expand def macros after type checking
+              // Note: macros that expand to functions cannot be immediately applied.
+              // This is a practical constraint, as currying is also disallowed for macros
+              if (fun1.symbol.is(Flags.Macro) && !pt.isInstanceOf[ApplyingProto]) {
+                if (ctx.macrosEnabled)
+                  typed(macros.expandDef(treeTpd), pt)
```
xeno-by: Am I right in thinking that you're using `expandDef` to expand def macros? Maybe we could think of a different name, because `expandDef` can be thought to mean `expand definition`?

compiler/src/dotty/tools/dotc/macros/Typer.scala
liufengyun: Yes, exactly.

xeno-by: How do you control that this behavior is only enabled for type parameters of enclosing inline methods?

compiler/src/dotty/tools/dotc/typer/Applications.scala
liufengyun: We need it to be on the `-toolcp` path or `-classpath`. For sbt users, it seems to me the latter is much simpler, just a normal dependency on Eden suffices.

For command-line programmers, there's no simple way to enable macros -- as all the transitive dependencies of `Eden` (thus `scalameta`) have to be in `-classpath` or `-toolcp` for reflection, which is a long list of JARs.

xeno-by: Can we use `sbt-assembly` to simplify the `-toolcp` experience here?

compiler/src/dotty/tools/dotc/macros/Typer.scala
liufengyun: Yes, we do need more sanity checks to print nice message for crazy programs.

compiler/src/dotty/tools/dotc/typer/Applications.scala
liufengyun: That is a big win for command line users -- they can provide the fat JAR to `-classpath`, so both definition and expansion will be no problem. Then it seems to me `-toolcp` is not really needed.

compiler/src/dotty/tools/dotc/core/Contexts.scala
liufengyun: Normally, we expect a meta block to return meta trees. Of course, we can check for `scala.meta.package$` instead.

liufengyun: Sorry I misunderstood initially. Checking `scala.meta.Tree` has the benefit of reducing performance penalty in type checking when `Eden` is in `-toolcp`, but no macros are defined or used at all -- thus `scala.meta.Tree` is not available.

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -360,7 +360,16 @@ class Namer { typer: Typer =>
   /** Expand tree and store in `expandedTree` */
   def expand(tree: Tree)(implicit ctx: Context): Unit = tree match {
     case mdef: DefTree =>
-      val expanded = desugar.defTree(mdef)
+      val expanded = mdef match {
+        case mdef: MemberDef if ctx.macrosEnabled && macros.hasAnnotMacros(mdef) =>
+          macros.expandAnnotMacro(mdef) match {
+            case Block(stats, _) =>                 // introducing of same level symbols
```
odersky: I think it would be cleaner if `expandAnnotMacro` returned a `Thicket` directly. Why the detour via `Block`? What happens with its `expr` part?

liufengyun: The return code might be a block like `$class1; $object1` ([link](https://github.com/liufengyun/eden/blob/master/macros/src/main/scala/Annotations.scala#L29)). `Thicket` is not available in syntax.

Indeed dropping `expr` in blocks looks dubious. Here it involves a convention in annotation macros on how to introduce same level definitions. Only based on that protocol, we can drop `expr`. For safety reasons, some checks and warnings are in order here.

odersky: I think it would be better we isolated this point in `expandMacro`. `Namer` should deal with `Thicket`, not with `Block`. It's very important to have only the cleanest representations in core modules such as `Namer` or `Typer`, or we will quickly drown in complexity. So any dirt should to be taken care of at the interfaces if possible.

liufengyun: Ah, I see your point, it makes perfect sense.

compiler/src/dotty/tools/dotc/macros/Typer.scala
```diff
@@ -0,0 +1,98 @@
...
+        implicit ctx => {
+          if (tree.name.isTermName) {
+            val name = tree.name.toTypeName
+            super.typedIdent(untpd.cpy.Ident(tree)(name = name), pt)
+          }
+          else fail()
+        }
+      } { (_, _) => fail() }
+    }
+
```
liufengyun: Thanks for the tip, transform the macro definitions before typing checking seems to be a much more elegant approach, I'll try that.

odersky: @liufengyun Have a look at `untpd.UntypedTreeMap`, that should come in handy for the pre-typer transformation.

odersky:
What's the status of this PR?

liufengyun:
Now the macro transform is moved to pre-typer. The test fails because of a performance regression. I'm still investigating the performance problem. Otherwise, it's ready.

liufengyun:
@odersky this is now ready for review again. The performance problem is related to adding a lazy field to `Context`, now I moved them to `ContextBase` (thanks @DarkDimius) and it solves the problem.

compiler/src/dotty/tools/dotc/macros/macros.scala
```diff
@@ -0,0 +1,138 @@
...
+  def expandDefMacro(tree: tpd.Tree)(implicit ctx: Context): Tree = {
+    val res = forObject("scala.meta.eden.expand.package").apply("expandDefMacro", tree, ctx).asInstanceOf[Tree]
+    debug.println(i"def macro expansion: $tree expands to $res")
+    res
+  }
+
+  /** Transform macro definitions in class definition
+   *
+   *  @param tree  the tree that may contain macro definition, the tree may be a thicket
+   *  @returns     the transformed tree
```
xuwei-k: `s/@returns/@return`

liufengyun: Thanks @xuwei-k , it's now fixed.

liufengyun:
Close this for now, a new PR will be made later.

PR #1900, opened 2017-01-13
Implement @binaryCompatible and make dotty-interfaces be Scala-defined.

DarkDimius:


DarkDimius:
This is intended to be `@tailrec` - like annotation that will fail compilation that checks that emmited methods and their signatures are the same as written by users, which means: no methods synthesized and no signatures changed.

The intended use-case is providing binary compatibility across major Scala versions by ensuring that a class or a trait is compiled `as written by developer`.

DarkDimius:
A formal SIP will follow shortly.

interfaces/src/dotty/tools/dotc/interfaces/SimpleReporter.scala
```diff
@@ -0,0 +1,19 @@
...
+
+/** Report errors, warnings and info messages during the compilation process
+ *
+ * You should implement this interface if you want to handle the diagnostics
+ * returned by the compiler yourself.
+ *
+ * See the method `process` of `dotty.tools.dotc.Driver` for more information.
+ */
+@binaryCompatible
+trait SimpleReporter {
```
smarter: The spec says that `@binaryCompatible` should not change the public API but a simple trait like that will get a method:
```java
    }
```
in dotty and in Scala 2.12, but not in Scala 2.11. In Scala 2.12 it will additionally get a method:
```java
    default public static /* synthetic */ void report$(SimpleReporter $this, Diagnostic diag) {
        $this.report(diag);
    }
```

All of these things affect the public API.

DarkDimius: > default public void $init$() {
    }

Empty trait constructor methods are indeed checked by hand.

>   default public static /* synthetic */ void report$(SimpleReporter $this, Diagnostic diag) {
        $this.report(diag);
    }

why would it be emmited if there's no default implementation provided(this method is abstract)?

smarter: I retract my comment, the code I was compiling had a non-abstract method, that's why these methods existed. (`$init$` seems to always be generated in dotty but that should be considered a bug)

interfaces/src/dotty/tools/dotc/interfaces/SourcePosition.scala
```diff
@@ -0,0 +1,55 @@
...
+ * As a convenience, we also provide methods that return the line and the column
+ * corresponding to each offset.
+ *
+ * User code should not implement this interface, but it may have to
+ * manipulate objects of this type.
+ */
+@binaryCompatible
+trait SourcePosition {
+  /** @return Content of the line which contains the point */
+  def lineContent(): String
```
smarter: Since all of these methods are pure, I think it's fine to rewrite them without any parameter lists when converting to Scala.

project/Build.scala
```diff
@@ -129,13 +129,14 @@ object DottyBuild extends Build {
     settings(publishing)
 
 
-  lazy val `dotty-interfaces` = project.in(file("interfaces")).
+  lazy val `dotty-interfaces`: Project = project.in(file("interfaces")).
     settings(sourceStructure).
+    dependsOn(`dotty-library`).
```
smarter: We shouldn't do this. We need `binaryCompatible` to be on the classpath when compiling `dotty-interfaces` but it shouldn't be a runtime dependency.

library/src/scala/annotation/binaryCompatible.scala
```diff
@@ -0,0 +1,7 @@
+package scala.annotation
+
+import scala.annotation.meta._
+
+/** https://github.com/scala/scala.github.com/pull/491 */
```
smarter: Wrong SIP.

project/Build.scala
```diff
@@ -129,13 +129,14 @@ object DottyBuild extends Build {
...
     settings(
       // Do not append Scala versions to the generated artifacts
       crossPaths := false,
       // Do not depend on the Scala library
-      autoScalaLibrary := false,
```
smarter: We shouldn't do this either. Now you're tying `dotty-interfaces` to a particular version of `scala-library` even though `scala-library` is not necessary to use `dotty-interfaces`, only to compile it.

DarkDimius: yes, that's why this commit is called a HACK. 
I propose to discuss on a Dotty meeting what should be the proper way to do it build-wise.

smarter:
You can try adding `dotty-library` as a compile-time only dependency using a custom ivy configuration: http://stackoverflow.com/a/21516954/348497

smarter:
I wonder if we should also not emit Tasty sections for `@binaryCompatible` classes since we don't have any binary-compat requirement for it yet.

allanrenucci:
Closing with `revisit` label

PR #1901, opened 2017-01-15
Adopt scala's scheme for root import hiding

odersky:
scalac hides a root import from Predef if there is an explicit Predef import.
We now do the same (previously we did this only if the overriding import undefined
something, using a `x => _` syntax). To avoid cycles and races one had to be very careful
not to force import symbols too early, so we now compare the name before the symbol proper.

All this is likely temporary - the comment of ImportInfo#unimported points to a different,
more systematic solution.

Note: The difference in semantics showed up and caused problems for the current collection strawman.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ImportInfo.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ImportInfo.scala
@@ -15,17 +15,20 @@ object ImportInfo {
     val selectors = untpd.Ident(nme.WILDCARD) :: Nil
     def expr = tpd.Ident(refFn())
     def imp = tpd.Import(expr, selectors)
-    new ImportInfo(imp.symbol, selectors, isRootImport = true)
+    new ImportInfo(imp.symbol, selectors, None, isRootImport = true)
   }
 }
 
 /** Info relating to an import clause
- *  @param   sym        The import symbol defined by the clause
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/typer/ImportInfo.scala
```diff
@@ -15,17 +15,20 @@ object ImportInfo {
...
+ *  @param   sym          The import symbol defined by the clause
+ *  @param   selectors    The selector clauses
+ *  @param   symNameOpt   Optionally, the name of the import symbol. None for root imports.
+ *                        Defined for all explicit imports from ident or select nodes.
+ *  @param   isRootImport true if this is one of the implicit imports of scala, java.lang,
+ *                        scala.Predef or dotty.DottyPredef in the start context, false otherwise.
  */
-class ImportInfo(symf: => Symbol, val selectors: List[untpd.Tree], val isRootImport: Boolean = false)(implicit ctx: Context) {
+class ImportInfo(symf: => Symbol, val selectors: List[untpd.Tree],
+                 symNameOpt: Option[TermName], val isRootImport: Boolean = false)(implicit ctx: Context) {
```
smarter: You could put a `if (isRootImport) assert(symNameOpt.isEmpty)` here to enforce what the documentation says.

PR #1903, opened 2017-01-16
Fix NPE in Implicits

odersky:
We got an NPE when compiling the collection strawman under -Yno-imports. We did not
foresee that the contextual implicit scope stack could be empty.

This was found when compiling collection strawman under -Yno-imports


odersky:
I am merging this without review because this is a blocker for the collection strawman effort.

PR #1904, opened 2017-01-16
Fix another NPE when compiling under -Yno-imports

odersky:


odersky:
I am also merging this now, in order to make progress on collection strawman.

smarter:
These two PRs LGTM to but we would be benefit from at least one test that exerces `-Yno-imports`.

PR #1906, opened 2017-01-16
Fix #1905: Duplicate bridge

### LINKED ISSUE #1905: Duplicate symbol on bridge generation
We get a duplicate generated symbol for `fromIterableWithSameElemType` in class `ArrOps` when compiling iarray.scala:

    import scala.reflect.ClassTag

    class Arr[T](private val underlying: scala.Array[T]) extends AnyVal

    trait SeqMonoTransforms[+A, +Repr] extends Any {
      protected[this] def fromIterableWithSameElemType(): Repr
    }

    class ArrOps[A](val xs: Arr[A]) extends AnyRef with SeqMonoTransforms[A, Arr[A]] {
      def fromIterableWithSameElemType(): Arr[A] = xs
    }

    object Test {
      def main(args: Array[String]) =
        println(new ArrOps(new Arr(Array(1, 2, 3))).fromIterableWithSameElemType)
    }

If we compile this with -Ycheck:erasure, we get:

exception occurred while compiling iarray.scala
Exception in thread "main" class dotty.tools.dotc.reporting.diagnostic.messages$Error at iarray.scala:9: method fromIterableWithSameElemType is already defined as method fromIterableWithSameElemType: ()ErasedValueType(Arr, Object)
	at dotty.tools.dotc.reporting.Reporting$class.error(Reporter.scala:85)
	at dotty.tools.dotc.core.Contexts$Context.error(Contexts.scala:57)
	at dotty.tools.dotc.typer.Checking$$anonfun$checkDecl$1$1.doubleDefError$1(Checking.scala:548)
	at dotty.tools.dotc.typer.Checking$$anonfun$checkDecl$1$1.apply(Checking.scala:551)
	at dotty.tools.dotc.typer.Checking$$anonfun$checkDecl$1$1.apply(Checking.scala:540)

odersky:
The problem with #1905 was that we checked "new" overriding relationships
at phase erasure + 1. This is wrong when we have ErasedValueTypes because
these do not compare correctly with their underlying type. So we get spurious
mismatches which force in turn spurious bridge generation. The fix is to
compute newly overridden symbols at phase elimErasedValueTypes + 1, i.e. one
phase later.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Erasure.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Erasure.scala
@@ -600,6 +600,7 @@ object Erasure extends TypeTestsCasts{
     // this implementation doesn't check for bridge clashes with value types!
     def addBridges(oldStats: List[untpd.Tree], newStats: List[tpd.Tree])(implicit ctx: Context): List[tpd.Tree] = {
       val beforeCtx = ctx.withPhase(ctx.erasurePhase)
+      val afterCtx = ctx.withPhase(ctx.elimErasedValueTypePhase.next)
       def traverse(after: List[Tree], before: List[untpd.Tree],
                    emittedBridges: ListBuffer[tpd.DefDef] = ListBuffer[tpd.DefDef]()): List[tpd.DefDef] = {
         after match {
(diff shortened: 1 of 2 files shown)


DarkDimius:
@odersky, please see my comments in https://github.com/lampepfl/dotty/issues/1905 - those errors aren't spurious.

PR #1909, opened 2017-01-22
Fix #1687: postpone computations in tailrec until they are needed.

### LINKED ISSUE #1687: Exponential compilation time of "1" + 1 + "1" + ...
The following complies in 11s:

```scala
object O {
  def f: String = {
    "1" + 1 + "1" + 1 +
    "1" + 1 + "1" + 1 +
    "1" + 1 + "1" + 1 +
    "1" + 1 + "1" + 1 +
    "1" + 1 + "1" + 1 +
    "1" + 1 + "1" + 1
  }
}
```

DarkDimius:
Previous version precomputed everything needed to make the final decision.
This was fast-path for method that will be tail-rec transformed.
Unfortunatelly, in case decision was not to tail-rec optimize it could have
led to exponential number of transformations.

Now, the fast-path is for methods that will not.


DarkDimius:
ping @mukel

odersky:
LGTM

PR #1913, opened 2017-01-29
Fix #1750: Alternative fix for cyclic references due to illegal class overrides

### LINKED ISSUE #1750: crash on compilable code: cyclic reference involving class Visitor

odersky:
Illegal class overrides are fundamentally at odds with the way dotty
represents types and therefore can cause lots of low-level problems.

Two measures in this commit:

First, we detect direct illegal class overrides on completion instead of
during RefChecks. Break the override by making the previously
overriding type private.

This fixes i1750.scala, but still fails for indirect overrides between
two unrelated outer traits/classes that are inherited by the same class or trait.
We fix this by catching the previously thrown ClassCastException
in both ExtractAPI and RefChecks.

Test case for indirect overrides is in i1750a.scala.

Review by @liufengyun

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/OverridingPairs.scala
+++ b/compiler/src/dotty/tools/dotc/transform/OverridingPairs.scala
@@ -6,6 +6,7 @@ import Flags._, Symbols._, Contexts._, Types._, Scopes._, Decorators._
 import util.HashSet
 import collection.mutable
 import collection.immutable.BitSet
+import typer.ErrorReporting.cyclicErrorMsg
 import scala.annotation.tailrec
 
 /** A module that can produce a kind of iterator (`Cursor`),
@@ -122,10 +123,18 @@ object OverridingPairs {
       if (nextEntry ne null) {
         nextEntry = decls.lookupNextEntry(nextEntry)
         if (nextEntry ne null) {
-          overridden = nextEntry.sym
-          if (overriding.owner != overridden.owner && matches(overriding, overridden)) {
-            visited += overridden
-            if (!hasCommonParentAsSubclass(overriding.owner, overridden.owner)) return
+          try {
+            overridden = nextEntry.sym
+            if (overriding.owner != overridden.owner && matches(overriding, overridden)) {
+              visited += overridden
+              if (!hasCommonParentAsSubclass(overriding.owner, overridden.owner)) return
+            }
+          }
+          catch {
+            case ex: CyclicReference =>
+              // See neg/i1750a for an example where a cyclic error can arise.
+              // The root cause in this example is an illegal "override" of an inner trait
+              ctx.error(cyclicErrorMsg(ex), base.pos)
           }
         } else {
           curEntry = curEntry.prev
--- a/tests/neg/overrides.scala
+++ b/tests/neg/overrides.scala
@@ -34,11 +34,6 @@ package p2 { // all being in the same package compiles fine
     }
   }
 
-  abstract class T3 extends T2 {
-    class A {   // error: classes cannot be overridden
-      bug()
-    }
-  }
 }
 
 class A[T] {
--- a/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
+++ b/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
@@ -6,6 +6,7 @@ import core._, core.Decorators._
 import Annotations._, Contexts._, Flags._, Phases._, Trees._, Types._, Symbols._
 import Names._, NameOps._, StdNames._
 import typer.Inliner
+import typer.ErrorReporting.cyclicErrorMsg
 
 import dotty.tools.io.Path
 import java.io.PrintWriter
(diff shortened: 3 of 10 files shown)


odersky:
The CI failure Iis an OOM during dotty bootstrap.

compiler/src/dotty/tools/dotc/transform/OverridingPairs.scala
```diff
@@ -122,10 +123,18 @@ object OverridingPairs {
...
+            if (overriding.owner != overridden.owner && matches(overriding, overridden)) {
+              visited += overridden
+              if (!hasCommonParentAsSubclass(overriding.owner, overridden.owner)) return
+            }
+          }
+          catch {
+            case ex: CyclicReference =>
+              // See neg/i1750a for an example where a cyclic error can arise.
+              // The root cause in this example is an illegal "override" of an inner trait
+              ctx.error(cyclicErrorMsg(ex), base.pos)
```
liufengyun: As a general rule in development, may I understand that `try/catch` can only be used as the last resort? Just want to know.

tests/neg/overrides.scala
```diff
@@ -34,11 +34,6 @@ package p2 { // all being in the same package compiles fine
     }
   }
 
-  abstract class T3 extends T2 {
-    class A {   // error: classes cannot be overridden
-      bug()
-    }
-  }
 }
```
liufengyun: why remove the class here?

odersky: It now masks the other errors because the error is raised now in typer, whereas the others would be raised in RefChecks.

compiler/src/dotty/tools/dotc/transform/OverridingPairs.scala
odersky: It's complicated. Cyclic references can only be handled with try-catch, because at the points they are raised we typically do not have a way to back out cleanly. So if a cyclic reference is a possibility we need to handle it with try/catch. And we should try to minimize the places where cyclic references can be raised.

I would think twice before we introduce other exceptions and handle them with try/catch.

tests/neg/overrides.scala
liufengyun: I see, it makes sense.

odersky:
We still get an OOM on dotty bootstrapped. @smarter @DarkDimius do we even need this test anymore, now that we bootstrap already as part of the build?

smarter:
> We still get an OOM on dotty bootstrapped. @smarter @DarkDimius do we even need this test anymore, now that we bootstrap already as part of the build?

Yes, the bootstrap can now be done through sbt, but it still has to be done explicitly by using the `dotty-compiler-bootstrapped` subproject. This is exactly what `partest-only` now does, it runs partest inside `dotty-compiler-bootstrapped` instead of running it inside `dotty-compiler`.

The OOM we're getting here is similar to what I experienced while working on the sbt bootstrap, I thought I fixed them by increasing the heap size that sbt can use in https://github.com/lampepfl/dotty/pull/1896/commits/239a850f2d89a78e53386e3f9b8959be74bb9e65, but it looks like it wasn't enough. To recap: the OOM does not happen in one of the forked JVM where we run the tests, instead it happens in the original JVM used by sbt to run its tasks. Since `Runtime.getRuntime.availableProcessors` returns 40 on our CI, sbt can run up to 40 tasks at the same time, some of these tasks can be compile tasks where scalac compiles dotty, or dotty compiles itself, this easily adds up to a lot of heap usage. We could fix this by reducing the parallelism used by sbt but that would slow down our tests, instead I propose to try increasing the heap used by sbt again. We should also check if running dotty through sbt leaks memory, but that's less urgent.

smarter:
... In fact you can disregard everything I said in the previous commit because this branch is not rebased on top of master, so it should not be using the new sbt-based bootstrap and the new CI settings, I'll rebase it to master and everything should be fine.

odersky:
@smarter Thanks for rebasing!

PR #1914, opened 2017-01-30
Fix incorrect flatten of thicket inside interpolated string in UntpdTreeMap

liufengyun:
Previously, the thickets in `InterpolatedString` will be flattened during UntpdTreeMap. This commit addresses this problem.

Review @odersky .

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/untpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/untpd.scala
@@ -476,7 +476,7 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
       case SymbolLit(str) =>
         cpy.SymbolLit(tree)(str)
       case InterpolatedString(id, segments) =>
-        cpy.InterpolatedString(tree)(id, transform(segments))
+        cpy.InterpolatedString(tree)(id, segments.map(transform(_)))
       case Function(args, body) =>
         cpy.Function(tree)(transform(args), transform(body))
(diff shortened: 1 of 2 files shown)


liufengyun:
@odersky the diff is as the following:

```
InterpolatedString(q,List(Thicket(Literal(Constant(hello )), Block(List(),Literal(Constant(2017)))), Literal(Constant(!))))
InterpolatedString(q,List(Literal(Constant(hello )), Block(List(),Literal(Constant(2017))), Literal(Constant(!))))
```

odersky:
@liufengyun I would argue there should be no `Thicket` in the interpolated string example.

liufengyun:
In desugaring, we have related code as follows:

```Scala
      case InterpolatedString(id, segments) =>
        val strs = segments map {
          case ts: Thicket => ts.trees.head
          case t => t
        }
        val elems = segments flatMap {
          case ts: Thicket => ts.trees.tail
          case t => Nil
        } map {
          case Block(Nil, expr) => expr // important for interpolated string as patterns, see i1773.scala
          case t => t
        }

        Apply(Select(Apply(Ident(nme.StringContext), strs), id), elems)
```

I understand that `Thicket` here is used to hold the `(str, elem)` pair. If we remove the thicket during TreeMap, it seems it's impossible to do desugaring, which will operate on the pairs to produce `StringContext(strs).id(elems)`.

odersky:
Got it. Thanks for explaining. So this LGTM.

PR #1918, opened 2017-01-31
Improvements to implicits

odersky:
The PR contains an optimization and a generalization of implicits

Optimization: We now avoid recomputation of companion refs in many cases
Generalization: We allow implicit call-by-name parameters

There are also some improvements to internal diagnostics regarding implicits.
There a largish test case, which explores some schemes for generic programming over ADTs.

Review by @smarter, @OlivierBlanvillain ?

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1765,12 +1765,8 @@ object Parsers {
               TypeTree()  // XX-METHOD-INFER
             } else {
               accept(COLON)
-              if (in.token == ARROW) {
-                if (owner.isTypeName && !(mods is Local))
-                  syntaxError(s"${if (mods is Mutable) "`var'" else "`val'"} parameters may not be call-by-name")
-                else if (imods.hasFlags)
-                  syntaxError("implicit parameters may not be call-by-name")
-              }
+              if (in.token == ARROW && owner.isTypeName && !(mods is Local))
+                syntaxError(s"${if (mods is Mutable) "`var'" else "`val'"} parameters may not be call-by-name")
               paramType()
             }
           val default =
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -6,6 +6,7 @@ import Texts._, Types._, Flags._, Names._, Symbols._, NameOps._, Constants._, De
 import Contexts.Context, Scopes.Scope, Denotations.Denotation, Annotations.Annotation
 import StdNames.{nme, tpnme}
 import ast.Trees._, ast._
+import typer.Implicits._
 import config.Config
 import java.lang.Integer.toOctalString
 import config.Config.summarizeDepth
@@ -483,6 +484,23 @@ class PlainPrinter(_ctx: Context) extends Printer {
         tree.fallbackToText(this)
     }
   }.close // todo: override in refined printer
+
+  def toText(result: SearchResult): Text = result match {
+    case result: SearchSuccess =>
+      "SearchSuccess: " ~ toText(result.ref) ~ " via " ~ toText(result.tree)
+    case result: NonMatchingImplicit =>
+      "NoImplicitMatches"
+    case result: DivergingImplicit =>
+      "Diverging Implicit"
+    case result: ShadowedImplicit =>
+      "Shadowed Implicit"
+    case result: FailedImplicit =>
+      "Failed Implicit"
+    case result: AmbiguousImplicits =>
+      "Ambiguous Implicit: " ~ toText(result.alt1) ~ " and " ~ toText(result.alt2)
+    case _ =>
+      "?Unknown Implicit Result?"
+  }
 
   private var maxSummarized = Int.MaxValue
 
--- a/tests/run/generic/Color.scala
+++ b/tests/run/generic/Color.scala
@@ -1,0 +1,30 @@
+package generic
+
+import Shapes._
+
+/** enum Color {
+ *    case Red
+ *    case Green
+ *    case Blue
+ *  }
+ */
+sealed trait Color extends Enum
+
+object Color extends EnumValues[Color](3) {
+
+  private def $new(tag: Int, name: String) = new Color {
+    def enumTag = tag
+    override def toString = name
+    registerEnumValue(this)
+  }
+
+  val Red: Color = $new(0, "Red")
+  val Green: Color = $new(1, "Green")
+  val Blue: Color = $new(2, "Blue")
+
+  implicit val ColorShape: Color `shaped` EnumValue[Color] =
+    new (Color `shaped` EnumValue[Color]) {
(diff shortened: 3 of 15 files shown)


compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -484,6 +485,23 @@ class PlainPrinter(_ctx: Context) extends Printer {
...
+      "NoImplicitMatches"
+    case result: DivergingImplicit =>
+      "Diverging Implicit"
+    case result: ShadowedImplicit =>
+      "Shadowed Implicit"
+    case result: FailedImplicit =>
+      "Failed Implicit"
+    case result: AmbiguousImplicits =>
+      "Ambiguous Implicit: " ~ toText(result.alt1) ~ " and " ~ toText(result.alt2)
+    case _ =>
```
odersky: Looks like we can

odersky:
Some current advantages of `enum` as a syntax:

 - less verbose syntax
 - gives us a robust way to find all children of a class. Even using sealed we still have problems with that because exploration can cause cycles.
 - Gives us an efficient way to create enums with many values.
 - Opens the potential to represent some enums as Java enums, which helps interop
 - Give us a lightweight way to hide implementation classes. E.g. We'd like `Cons`, `Nil` to be
   of type List (instead of a subtype) because this helps sometimes with type inference. 

Could this be done with Scala meta? I doubt it since the natural enum syntax is not currently legal Scala. Also there's a "meta discussion" to be had about Scala meta. Language workbenches are very appealing academically but often turn to nightmares in practice. Google for "The Lisp Curse" for details.

tests/run/generic/Shapes.scala
```diff
@@ -0,0 +1,22 @@
...
+  case class Fst[+F](x: F) extends Sum[F, Nothing]
+  case class Snd[+S](x: S) extends Sum[Nothing, S]
+
+  case class Prod[+P1, +P2](fst: P1, snd: P2)
+
+  case class Singleton[SI](value: SI)
+
+  case class EnumValue[E](tag: Int)
+
+  trait shaped[SH1, SH2] extends unfolds[SH1, SH2]
```
smarter: What's the difference between shaped and unfolds?

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
odersky: No, in fact we cannot. There are too many other cases we'd have to handle. E.g. ExplainedSearchResult. Sometimes it's better not to seal.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1765,12 +1765,8 @@ object Parsers {
               TypeTree()  // XX-METHOD-INFER
             } else {
               accept(COLON)
-              if (in.token == ARROW) {
-                if (owner.isTypeName && !(mods is Local))
-                  syntaxError(s"${if (mods is Mutable) "`var'" else "`val'"} parameters may not be call-by-name")
-                else if (imods.hasFlags)
-                  syntaxError("implicit parameters may not be call-by-name")
```
xeno-by: Why was this prohibited in scalac?

xeno-by:
@OlivierBlanvillain @odersky As defined in the examples, the syntax for enums cannot be implemented in scala.meta. In new-style macro annotations, we only accept annottees written using regular Scala syntax. 

At the moment, we don't have plans to turn scala.meta into a language workbench. The main goal of the project is to provide syntactic and semantic API for standard Scala to be used in developer tools and macros.

tests/run/generic/Color.scala
```diff
@@ -0,0 +1,30 @@
...
+ */
+sealed trait Color extends Enum
+
+object Color extends EnumValues[Color](3) {
+
+  private def $new(tag: Int, name: String) = new Color {
+    def enumTag = tag
+    override def toString = name
+    registerEnumValue(this)
+  }
```
xeno-by: Can we have a named subclass of `Color` defined somewhere in the companion object and used here? I think that classnames of anonymous classes are quite ugly.

tests/run/generic/Enum.scala
```diff
@@ -0,0 +1,18 @@
+package generic
+
+import Shapes.Singleton
+
+trait Enum {
+  def enumTag: Int
```
xeno-by: What use cases do you envision for `enumTag`? Back then, @densh and I planned to use tags to speed up pattern matching. Do you think that'd make sense?

tests/run/generic/Enum.scala
```diff
@@ -0,0 +1,18 @@
...
+}
+
+trait FiniteEnum extends Enum
```
xeno-by: What's that?

tests/run/generic/Enum.scala
```diff
@@ -0,0 +1,18 @@
...
+
+abstract class EnumValues[E <: Enum](numVals: Int) {
+  private var myValues = new Array[AnyRef](numVals)
+
+  def registerEnumValue(v: E) =
+    myValues(v.enumTag) = v
```
xeno-by: Can we somehow restrict the visibility of this method?

tests/run/generic/List.scala
```diff
@@ -0,0 +1,89 @@
...
+  abstract case class Cons[T](hd: T, tl: List0[T]) extends List0[T] {
+    def enumTag = 0
+  }
+  object Cons {
+    def apply[T](x: T, xs: List0[T]): List0[T] = new Cons(x, xs) {}
+    implicit def ConsShape[T]: Cons[T] `shaped` Prod[T, List0[T]] =
+      new (Cons[T] `shaped` Prod[T, List0[T]]) {
+        def toShape(x: Cons[T]) = Prod(x.hd, x.tl)
+        def fromShape(p: Prod[T, List0[T]]) = new Cons(p.fst, p.snd) {}
+      }
```
xeno-by: @odersky Am I right in thinking that you'd like the `XXXShape` implicits to be derived by macros?

xeno-by: And, in general, which parts of the `enum` proposal do you see implementable as macros, and what you'd like to hardcode in the compiler?

tests/run/generic/List.scala
```diff
@@ -0,0 +1,89 @@
...
+        case x: Nil[T] => Snd(x)
+      }
+      def fromShape(x: Sum[Cons[T], Nil[T]]) = x match {
+        case Fst(c) => c
+        case Snd(n) => n
+      }
+    }
+}
+
+/** enum List[T] {
```
xeno-by: Is the variance modifier omitted here on purpose?

tests/run/generic/List.scala
odersky: @xeno-by The `XXXShape` implicits can be done by the compiler. The part I need from macros are the implicits in Test.scala.

tests/run/generic/List.scala
odersky: There are actually two sources, one without variance for List0, the other with variance for List1.

PR #1919, opened 2017-01-31
Fix #1915 Synthetic function traits need NoInits flag

### LINKED ISSUE #1915: FunctionXXL NoSuchMethodError on `$init$` method.
When instantiating a `FunctionXXL` in:
```scala
object Test {
    println(new IntFunction26().apply(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26))
  }
}

class IntFunction26 extends Function26[Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int] {
  def apply(x1: Int, x2: Int, x3: Int, x4: Int, x5: Int, x6: Int, x7: Int, x8: Int, x9: Int, x10: Int, x11: Int, x12: Int, x13: Int, x14: Int, x15: Int, x16: Int, x17: Int, x18: Int, x19: Int, x20: Int, x21: Int, x22: Int, x23: Int, x24: Int, x25: Int, x26: Int) = 42
}
```
a `NoSuchMethodError` is thrown at
```scala
java.lang.NoSuchMethodError: scala.FunctionXXL.$init$()V
	at IntFunction26.<init>(IntFunction26.scala:7)
	at Test$.main(IntFunction26.scala:3)
	at Test.main(IntFunction26.scala)
```

In theory the `FunctionXXL` should be a trait with the `NoInitsTrait` flags, hence the call to `$init$` should not exist.

odersky:
Fixes #1915. Review by @nicolasstucki

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -766,6 +766,8 @@ class Definitions {
 
   lazy val PhantomClasses = Set[Symbol](AnyClass, AnyValClass, NullClass, NothingClass)
 
+  lazy val NoInitClasses = PhantomClasses + FunctionXXLClass
+
   def isPolymorphicAfterErasure(sym: Symbol) =
      (sym eq Any_isInstanceOf) || (sym eq Any_asInstanceOf)
 
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -130,7 +130,7 @@ class Definitions {
           ClassInfo(ScalaPackageClass.thisType, cls, ObjectType :: parentTraits, decls)
       }
     }
-    newClassSymbol(ScalaPackageClass, name, Trait, completer)
+    newClassSymbol(ScalaPackageClass, name, Trait | NoInits, completer)
   }
 
   private def newMethod(cls: ClassSymbol, name: TermName, info: Type, flags: FlagSet = EmptyFlags): TermSymbol =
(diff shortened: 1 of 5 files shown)


nicolasstucki:
@odersky, I had a the same issue and I believe [this change](https://github.com/lampepfl/dotty/pull/1408/files#diff-9cb74d44450b0176fbe43789a63779e1R178) fixes it. It seams that when a function (0 to 22) class is not bootstrapped it is missing the `NoInitsTrait` flag. The question here would be why it is missing and is that an issue for other traits coming from Scala 2?

odersky:
@nicolasstucki Can you be more specific what change you had in mind? The commit you point to is huge...

nicolasstucki:
The link should be pointing to the line.  The change is in `superCallOpt` of `transform/Mixin.scala`. Ignore the change from `defn.PhantomClasses` to `defn.NotRuntimeClasses`.

```scala
def superCallOpt(baseCls: Symbol): List[Tree] = superCalls.get(baseCls) match {
       .....
        case None =>
 -        if (baseCls.is(NoInitsTrait) || defn.PhantomClasses.contains(baseCls)) Nil
 +        // FIXME: function classes should have the NoInitsTrait flags
 +        if (baseCls.is(NoInitsTrait) || defn.NotRuntimeClasses.contains(baseCls) || defn.FunctionXXLClass == baseCls) Nil
          else {
            //println(i"synth super call ${baseCls.primaryConstructor}: ${baseCls.primaryConstructor.info}")
            transformFollowingDeep(superRef(baseCls.primaryConstructor).appliedToNone) :: Nil
```

odersky:
@nicolasstucki I see that patches the problem. But why single out `FunctionXXL`. In what way is it different from other traits we read from Scala2?

odersky:
@nicolasstucki I have figured it out. It's complicated. See my comment on the last commit.

nicolasstucki:
LGTM

Issue #1730, opened 2016-11-21
Disallow Array[Nothing], Array[Null], ... to avoid ClassCastException in some cases


smarter:
See also https://issues.scala-lang.org/browse/SI-7453

In `scalac`:
```scala
scala> if (false) Array("qwe") else Array()
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;
```
In dotty, thankfully:
```scala
scala> if (false) Array("qwe") else Array() 
res1: Array[String] | Array[Nothing] = [Ljava.lang.Object;@1dd92fe2
```
But still in dotty:
```scala
scala> val x = if (false) Array("qwe") else Array() 
x: Array[String] | Array[Nothing] = [Ljava.lang.Object;@5454d35e
scala> val y: Array[_ <: String] = x 
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;
```

We either need to disallow `Array[Nothing]`, `Array[Null]`, etc or we need to erase `Array[_ <: X]` to `Object` instead of `[X;`.

It would also be much nicer if we could infer the empty array above to have type `Array[String]` instead of `Array[Nothing]`, I don't know what prevents us from doing this currently.

odersky:
Erasing Array to Object is very expensive computationally, and it would be surprising for users. So I am much in favor of disallowing Array[Nothing] and Array[Null].

odersky:
One quite radical idea would be to treat `Nothing` and `Null` not as classes but as abstract types. That means we could not generate a `ClassTag` for them and therefore it would be impossible to create instances of `Array[Nothing] and `Array[Null]`

sjrd:
If we disallow `classTag[Null]`, then we should also disallow `classOf[Null]`. This is extreme, as now we have concrete types we cannot a `classOf`/`classTag` of anymore. "Treating them as abstract types" seems completely unjustified, other than as a workaround for this implementation limitation related to arrays. This seems wrong to me.

odersky:
Is `classOf[Null]` used anywhere?

I have my doubts because in the end there is no class `Null` on the JVM. We fake one, but it's a hack.

retronym:
Here are a few usages I found in projects I've checked out. Seems fairly rare.

```
⚡ ack --scala 'classOf\[Null\]'

json4s/scalabeans/src/main/scala/org/scalastuff/scalabeans/sig/ScalaTypeCompiler.scala
[...]

PR #1921, opened 2017-02-01
Fix #1907: Improve error message

### LINKED ISSUE #1907: Union type found when inferring `Array.empty`
```scala
import java.io.File

object Test {
  Some(new File("."))
    .map(_.listFiles).getOrElse(Array.empty)
    .map(_.listFiles)
}
```
Error message:

```
value `map` is not a member of Array[java.io.File] | Array[T]
where:    T is a type variable which is an alias of Array[T]
```

odersky:
Two fixes:

 - we avoid the 5-levels of array tag by synthesizing array tags at the same time as other class tags
 - we bail when asked to generate a `ClassTag[Nothing]` or `ClassTag[Null]`.

The second point is likely contentious. It does forbid some programs that would make sense such as

    def f(x: Any) = ???
    f(Array(): _*)

But it also avoids many situations that would lead to surprising results or confusing errors later.

@retronym What is your take on this?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
+++ b/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
@@ -79,12 +79,12 @@ trait ConstraintHandling {
       if (Config.failOnInstantiationToNothing) assert(false, msg)
       else ctx.log(msg)
     }
-    constr.println(i"adding $description")
+    constr.println(i"adding $description in ${ctx.typerState.hashesStr}")
     val lower = constraint.lower(param)
     val res =
       addOneBound(param, bound, isUpper = true) &&
       lower.forall(addOneBound(_, bound, isUpper = true))
-    constr.println(i"added $description = $res")
+    constr.println(i"added $description = $res in ${ctx.typerState.hashesStr}")
     res
   }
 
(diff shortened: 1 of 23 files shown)


odersky:
This is based on #1918. Only the last 3 commits are new. The contentious commit about not synthesizing ClassTags of bottom types is 	e34555f.

retronym:
I'll need to page this back in to comment meaningfully, but for now, I'll dump some links to related discussions about soundness issues with arrays and/or class tags of bottom types:

https://issues.scala-lang.org/browse/SI-9754
https://issues.scala-lang.org/browse/SI-5353
https://issues.scala-lang.org/browse/SI-7453

odersky:
@retronym Thanks for the links. They are all treated OK under this commit. I am not sure whether they would have been soundness issues before, since dotc with its union types behaves differently than scalac. So for instance i1907.scala is compiled by scalac but rejected by dotc.

smarter:
This also indirectly solves https://github.com/lampepfl/dotty/issues/1730, I think it's fine to require the user to be explicit when dealing with `Array[Nothing]`/`Array[Null]`, though the error message as is might be confusing so we should not forget to add a detailed `-explained` message for it at some point.

odersky:
@smarter It would solve #1730, but only if we also disallow forming a ClassTag[Nothing] or ClassTag[Null] manually. I think it would be a good idea to do this, though.

smarter:
Yes, you're right, otherwise the soudness hole is still there.

Issue #1723, opened 2016-11-16
`checkNoPrivateLeaks` is done too late, code with inferred private types might break


smarter:
`checkNoPrivateLeaks` replaces private types in signatures by dealiasing them if possible, but this is done in `PostTyper` which means that the following code does not compile (works in scalac):
```scala
class A {
  private type Foo = Int

  def foo: Foo = 1
}
class B extends A {
  foo
}
```
```scala
cannot take signature of => B.this.Foo
-- Error: ./try/1130x.scala ----------------------------------------------
7 |  foo
  |  ^^^
  |  cannot resolve reference to type B(B.this).Foo
  |  the classfile defining the type might be missing from the classpath
```

This is a big deal because inferred types might contain private types in perfectly innocent code:
```scala
class A {
  private val x: List[Int] = List(1)
  def foo = x.head // foo inferred type is this.x.scala$collection$immutable$List$$A
}

class B extends A {
  foo
}
```
```scala
cannot take signature of => B.this.x.scala$collection$immutable$List$$A
-- Error: ./try/i1130x.scala -------------------------------------------------------------------------------------------
7 |  foo
  |  ^^^
  |  cannot resolve reference to type B.this.x.type.scala$collection$immutable$List$$A
  |  the classfile defining the type might be missing from the classpath
```

I think we really need inference to avoid private types.

/cc @odersky

smarter:
Fixed by #1922, closing.

Issue #1644, opened 2016-10-29
crash for nested class with message: "no path exists from class Inner to class..."


helloqirun:
$ cat abc.scala

```
class X(val i: Int) extends AnyVal { class Inner(val q: Int) { def plus = i + q } }
object Test extends App {
  val x = new X(11)
  val i = new x.Inner(22)
  assert(i.plus == 33)
}
```

$ dotc abc.scala

```
exception while typing X.this of class class dotty.tools.dotc.ast.Trees$This # 151
exception while typing X.this.i of class class dotty.tools.dotc.ast.Trees$Select # 152
exception while typing X.this.i.+ of class class dotty.tools.dotc.ast.Trees$Select # 153
exception while typing X.this.i.+(Inner.this.q) of class class dotty.tools.dotc.ast.Trees$Apply # 367
exception while typing def plus: Int = X.this.i.+(Inner.this.q) of class class dotty.tools.dotc.ast.Trees$DefDef # 368
exception while typing class Inner(q: Int) extends Object() { 
  <accessor> def q: Int
  def plus: Int = X.this.i.+(Inner.this.q)
  private <accessor> def $outer: X
  final def X$Inner$$$outer: X = Inner.this.$outer
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 463
exception while typing @scala.annotation.internal.SourceFile("abc.scala") final class X(i: Int) extends AnyVal() { 
  <accessor> def i: Int
  class Inner(q: Int) extends Object() { 
    <accessor> def q: Int
    def plus: Int = X.this.i.+(Inner.this.q)
    private <accessor> def $outer: X
    final def X$Inner$$$outer: X = Inner.this.$outer
  }
  override def hashCode(): Int = X.hashCode$extension(this)()
  override def equals(x$0: Any): Boolean = X.equals$extension(this)(x$0)
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 465
exception while typing package <empty> {
  @scala.annotation.internal.SourceFile("abc.scala") final class X(i: Int) extends AnyVal() { 
    <accessor> def i: Int
    class Inner(q: Int) extends Object() { 
      <accessor> def q: Int
      def plus: Int = X.this.i.+(Inner.this.q)
      private <accessor> def $outer: X
      final def X$Inner$$$outer: X = Inner.this.$outer
    }
    override def hashCode(): Int = X.hashCode$extension(this)()
[...]

PR #1922, opened 2017-02-01
Fix-#1723: Avoid private leaks on completion

odersky:
As #1723 demonstrates, doing this at PostTyper is too late.

Review by @smarter. How about having a look at #1644 soon in return?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -733,7 +733,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
         // no longer necessary.
       goto(end)
       setPos(start, tree)
-      ta.avoidPrivateLeaks(sym, tree.pos)
+      sym.info = ta.avoidPrivateLeaks(sym, tree.pos)
       tree
     }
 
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -343,6 +343,7 @@ object Checking {
         fail(i"$sym cannot have the same name as ${cls.showLocated} -- class definitions cannot be overridden")
         sym.setFlag(Private) // break the overriding relationship by making sym Private
       }
+    avoidPrivateLeaks(sym, sym.pos)
```
smarter: So far, no method in `Checking` mutates anything, it would be nice to keep this pattern. I suggest doing what `checkNonCyclic` does and returning the potentially updated `sym.info`, i.e.:
```scala
if (!sym.is(SyntheticOrPrivate) && sym.owner.isClass)
  checkNoPrivateLeaks(sym, sym.pos)
else
  sym.info
```

odersky: > So far, no method in Checking mutates anything

The line just above the `avoidPrivateLeaks` call is also a mutation (admittedly, only in the error case).

odersky: I have now refactored. You were right, the new organization is more coherent.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -733,7 +733,7 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
         // no longer necessary.
       goto(end)
       setPos(start, tree)
-      Checking.avoidPrivateLeaks(sym, tree.pos)
+      ta.avoidPrivateLeaks(sym, tree.pos)
```
smarter: Missing `sym.info =`

odersky: Oops

PR #1923, opened 2017-02-01
avoid using macro-based StringContext.f to ease bootstrap

liufengyun:
In two places we used the macro-based `f` string interpolator, which will not expand during bootstrap, thus the generated code for `TastyPrinter` will not be actually runnable.

Review @felixmulder .

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
@@ -31,8 +31,10 @@ class TastyPrinter(bytes: Array[Byte])(implicit ctx: Context) {
   def nameRefToString(ref: NameRef): String = nameToString(tastyName(ref))
 
   def printNames() =
-    for ((name, idx) <- tastyName.contents.zipWithIndex)
-      println(f"$idx%4d: " + nameToString(name))
+    for ((name, idx) <- tastyName.contents.zipWithIndex) {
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
```diff
@@ -47,7 +49,8 @@ class TastyPrinter(bytes: Array[Byte])(implicit ctx: Context) {
     def unpickle(reader: TastyReader, tastyName: TastyName.Table): Unit = {
       import reader._
       var indent = 0
-      def newLine() = print(f"\n ${index(currentAddr) - index(startAddr)}%5d:" + " " * indent)
+      val length = "%5d:".format(index(currentAddr) - index(startAddr))
```
felixmulder: Since `newLine` is called twice and `length` is defined as a val, then I believe length would have the same value even though `currentAddr` and `startAddr` might mutate. This should therefore be inside the def like:

```scala
def newLine() = {
  val length = "%5d:".format(index(currentAddr) - index(startAddr))
  print(s"\n $length" + " " * indent)
}
```

liufengyun: Good catch @felixmulder , thanks a lot 🍷 !

felixmulder: Otherwise LGTM :)

PR #1926, opened 2017-02-01
Fix #1747: Tweak for <:< between method types

### LINKED ISSUE #1747: Can't implement or override java method with type parameters.
The following code will not compile because it can't match the types method types.
```scala
class Coll[E] extends java.util.Collection[E] {
  def toArray[T](a: Array[T]): Array[T] = ???
  ...
}


-- Error: Coll.scala -----------------------------------------------------------
2 |   def toArray[T](a: Array[T]): Array[T] = ???
  |       ^
  |overriding method toArray in trait Collection of type [T] => (x$0: Array[T])Array[T];
  | method toArray of type [T] => (a: Array[T])Array[T] has incompatible type
-- Error: Coll.scala -----------------------------------------------------------
1 |class Coll[E] extends java.util.Collection[E] {
  |      ^
  |class Coll needs to be abstract, since def toArray: [T] => (x$0: Array[T])Array[T] is not defined 
  |(Note that Array[T] does not match Array[T]: their type parameters differ)

two errors found
```

The same happens when trying to override the method:
```scala
class Coll[E] extends java.util.AbstractSet[E] {
  override def toArray[T](a: Array[T]): Array[T] = ???
  ...
}


-- Error: Coll.scala -----------------------------------------------------------
2 |  override def toArray[T](a: Array[T]): Array[T] = ???
  |               ^
  |overriding method toArray in class AbstractCollection of type [T] => (x$0: Array[T])Array[T];
  | method toArray of type [T] => (a: Array[T])Array[T] has incompatible type

one error found
```

odersky:
If one of two method types originates from Java, the two
types might match yet have different parameter signatures,
because of the way Array[T] is handled differently in Java
and Scala. The tweak takes this pecularity into account.

Review by @nicolasstucki


tests/pos/i1747.scala
```diff
@@ -0,0 +1,3 @@
+abstract class Coll[E] extends java.util.Collection[E] {
+  def toArray[T](a: Array[T]): Array[T] = ???
```
nicolasstucki: It would probably be a good idea to make this a `run` test to make sure that the erased signatures are also consistent in the bytecode.

odersky: @nicolasstucki Yes, I agree. Can I ask you to add this? Thanks!

retronym: I don't think that its sound to treat them as equivalent. You can't safely provide a generic Scala Array to Java:

```scala
scala> def foo[T](al: java.util.ArrayList[T], a: Array[T]) = al.toArray(a)
<console>:12: error: type mismatch;
 found   : Array[T]
 required: Array[? with Object]
Note: T >: ? with Object, but class Array is invariant in type T.
You may wish to investigate a wildcard type such as `_ >: ? with Object`. (SLS 3.2.10)
       def foo[T](al: java.util.ArrayList[T], a: Array[T]) = al.toArray(a)
                                                                        ^
```

You need to bound things to:

```scala
scala> def foo[T <: AnyRef](al: java.util.ArrayList[T], a: Array[T]) = al.toArray(a)
foo: [T <: AnyRef](al: java.util.ArrayList[T], a: Array[T])Array[T with Object]
```

The overriding method can't add a bound to the type parameter, but can express the restriction as:

```scala
abstract class C[E] extends java.util.Collection[E] {
   override def toArray[T](a: Array[T with Object]): Array[T with Object] = ???
}
```

We also can't add bridges to paper over the difference because of the mutability of arrays.

nicolasstucki: Additionally, the `pos` test is not testing anything as it is an abstract class and it could be defining new abstract members.
```scala
abstract class Coll[E] extends java.util.Collection[E] {
  def toArray[T](a: Array[T]): Array[T] = ???
}
```

This is in fact generating the following method in the bytecode.

```scala
public Object toArray(Object a) { ... }
```
I added run tests that reproduce the issue (the CI will fail).

odersky:
In light of @retronym's comments it seems we should reject this PR and close #1747, or change it to demand a better error message. Everyone agreed?

nicolasstucki:
Yes

PR #1928, opened 2017-02-01
Fix #1569: Fix logic for by-name parameters of inline methods

### LINKED ISSUE #1569: Inline def with `inline x: => X` parameters do not respect semantics.
``` scala
inline def foo(inline n: => Int) = n + n
foo({ println("foo"); 42 })
```

should print two times `"foo"` and return 84 but currently it just returns 84 without printing anything.

odersky:
As #1569 shows, by-name parameters should be treated in the same way
as by-value parameters. If the expression is not pure, install a binding.


nicolasstucki:
LGTM

PR #1929, opened 2017-02-02
support pre-typer transform of trees

liufengyun:
Currently `mergeCompanionDefs` assume `ModuleDef` is expanded
exactly to `module val; module class`. To enable more
general pre-typer transform, we need to weaken the assumption
to that the first two elements of the thicket must be
`module val; module class`.

This change will enable transform def macros defined inside
an object, as well as other pre-typer transforms.

Review @odersky .


odersky:
LGTM

PR #1931, opened 2017-02-02
Fix #1501 - Check trait inheritance condition

### LINKED ISSUE #1501: Class extended to a trait fail on invokespecial when mixed in.
``` scala
class A {
  def foo: Int = 1
}

trait B extends A

abstract class D {
  def foo: Int
}

class C extends D with B

object Test {
    println(new C().foo)
  }
}
```

will fail with:

```
java.lang.VerifyError: Bad invokespecial instruction: current class isn't assignable to reference class.
Exception Details:
  Location:
    C.foo()I @1: invokespecial
  Reason:
    Error exists in the bytecode
  Bytecode:
    0x0000000: 2ab7 0013 ac                           

    at Test$.main(foo.scala:16)
    at Test.main(foo.scala)
    at scala.reflect.internal.util.ScalaClassLoader$$anonfun$run$1.apply(ScalaClassLoader.scala:70)
    at scala.reflect.internal.util.ScalaClassLoader$class.asContext(ScalaClassLoader.scala:31)
    at scala.reflect.internal.util.ScalaClassLoader$URLClassLoader.asContext(ScalaClassLoader.scala:101)
    at scala.reflect.internal.util.ScalaClassLoader$class.run(ScalaClassLoader.scala:70)
    at scala.reflect.internal.util.ScalaClassLoader$URLClassLoader.run(ScalaClassLoader.scala:101)
    at scala.tools.nsc.CommonRunner$class.run(ObjectRunner.scala:22)
    at scala.tools.nsc.ObjectRunner$.run(ObjectRunner.scala:39)
    at scala.tools.nsc.CommonRunner$class.runAndCatch(ObjectRunner.scala:29)
    at scala.tools.nsc.ObjectRunner$.runAndCatch(ObjectRunner.scala:39)
    at scala.tools.nsc.MainGenericRunner.runTarget$1(MainGenericRunner.scala:65)
    at scala.tools.nsc.MainGenericRunner.run$1(MainGenericRunner.scala:87)
    at scala.tools.nsc.MainGenericRunner.process(MainGenericRunner.scala:98)
    at scala.tools.nsc.MainGenericRunner$.main(MainGenericRunner.scala:103)
    at scala.tools.nsc.MainGenericRunner.main(MainGenericRunner.scala)

```

odersky:
We need to check a coherence condition between the superclass
of a trait and the superclass of an inheriting class or trait.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -603,12 +603,25 @@ trait Checking {
 
   /** Given a parent `parent` of a class `cls`, if `parent` is a trait check that
    *  the superclass of `cls` derived from the superclass of `parent`.
+   *
+   *  An exception is made if `cls` extends `Any`, and `parent` is a Java class
+   *  that extends `Object`. For instance, we accept code like
+   *
+   *      ... extends Any with java.io.Serializable
+   *
+   *  The standard library relies on this idiom.
    */
-  def checkTraitInheritance(parent: Symbol, cls: ClassSymbol, pos: Position)(implicit ctx: Context): Unit =
+  def checkTraitInheritance(parent: Symbol, cls: ClassSymbol, pos: Position)(implicit ctx: Context): Unit = {
     parent match {
-      case parent: ClassSymbol if parent.is(Trait) && !cls.superClass.derivesFrom(parent.superClass) =>
-        ctx.error(em"illegal trait inheritance: super${cls.superClass} does not derive from $parent's super${parent.superClass}", pos)
+      case parent: ClassSymbol if parent is Trait =>
+        val psuper = parent.superClass
+        val csuper = cls.superClass
+        val ok = csuper.derivesFrom(psuper) ||
+          parent.is(JavaDefined) && csuper == defn.AnyClass && psuper == defn.ObjectClass
+        if (!ok)
+          ctx.error(em"illegal trait inheritance: super$csuper does not derive from $parent's super$psuper", pos)
       case _ =>
+    }
   }
 }
 
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -448,7 +448,6 @@ object Checking {
       }
       stats.foreach(checkValueClassMember)
     }
-
   }
 }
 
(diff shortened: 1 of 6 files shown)


smarter:
The following compiles in dotty but not scalac, is this intentional?
```scala
class A
class SubA extends A
trait TA extends A
trait TSubA extends SubA

class Foo extends TA with TSubA
```
In dotty the last line becomes after typer:
```scala
class Foo() extends SubA() with TA with TSubA
```
So it's correct, but in scalac we get:
```scala
 class Foo extends A with TA with TSubA
```
So it complains (`superclass A is not a subclass of the superclass SubA ...`). 

If this behavior is intended the [spec](http://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html#templates) will need to be updated:
> It is possible to write a list of parents that starts with a trait reference, e.g. mt1 with … with mtn. In that case the list of parents is implicitly extended to include the supertype of mt1 as first parent type.

DarkDimius:
I think that the question @smarter asked is incomplete without discussing order of constructor evaluation, ie: in which order should super-initializers be called for `Foo`?

DarkDimius:
Class Foo _has_ to start by calling the super-call of SubA, which whould invoke A and SubA.

IE the starting sequence is fixed:
```
Object
A
SubA
```
Now, we have flexibility to continue by calling `TA` and `TSubA`. 

I think this is a valid and expected order for what was created by typer and should be allowed.

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -601,6 +600,29 @@ trait Checking {
...
+   *
+   *  The standard library relies on this idiom.
+   */
+  def checkTraitInheritance(parent: Symbol, cls: ClassSymbol, pos: Position)(implicit ctx: Context): Unit = {
+    parent match {
+      case parent: ClassSymbol if parent is Trait =>
+        val psuper = parent.superClass
+        val csuper = cls.superClass
+        val ok = csuper.derivesFrom(psuper) ||
+          parent.is(JavaDefined) && csuper == defn.AnyClass && psuper == defn.ObjectClass
```
smarter: I think this might be too permissive: as far as I can tell, `scalac` only allows allows this for two Java classes: `Serializable` and `Comparable` which are treated as if they were traits that extend `Any`: https://github.com/scala/scala/blob/cac6383e6658dc5956540b76b9b46c3b664774ac/src/reflect/scala/reflect/internal/Definitions.scala#L365-L366 If we allow any Java class to be used in universal traits, then we break the definition of universal trait.

odersky: OK, I'll change that.

odersky:
@darkdimius Right. I just verified that this does does not lead to inconsistencies when classes take parameters.

odersky:
We should note that this requires a spec change.

odersky:
Why don't we see a drone result here? validate-main is pending as usual, but the drone should still be running, right?

smarter:
No clue, I haven't seen that on other PRs, try git commit --amend && git push --force ?

felixmulder:
Merging this broke master: http://dotty-ci.epfl.ch/lampepfl/dotty/900

PR #1933, opened 2017-02-02
Weaken assumption in mergeCompanionDefs

liufengyun:
Previously `mergeCompanionDefs` assumes that if the attachment of class `Foo` is as follows:

    x :: y :: tdef @ TypeDef(_, templ)

Then `tdef` must be `Foo$`. When there are multiple pre-typer transforms,
this is not necessarily true. For example, an annotation macro expansion may
expand a non-case class `Foo` to `class Foo; object FooA`.

We need to check the name of `tdef` to be equal to `Foo$`.

Review @odersky .

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -491,7 +491,8 @@ class Namer { typer: Typer =>
         if (cdef.isClassDef) {
           classDef(name) = cdef
           cdef.attachmentOrElse(ExpandedTree, cdef) match {
-            case Thicket(cls :: mval :: (mcls @ TypeDef(_, _: Template)) :: crest) =>
+            case Thicket(cls :: mval :: (mcls @ TypeDef(mname, _: Template)) :: crest)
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -491,7 +491,8 @@ class Namer { typer: Typer =>
         if (cdef.isClassDef) {
           classDef(name) = cdef
           cdef.attachmentOrElse(ExpandedTree, cdef) match {
-            case Thicket(cls :: mval :: (mcls @ TypeDef(_, _: Template)) :: crest) =>
+            case Thicket(cls :: mval :: (mcls @ TypeDef(mname, _: Template)) :: crest)
+            if name ++ nme.MODULE_SUFFIX == mname =>
```
smarter: `NameOps` has a `moduleClassName` method to do that

liufengyun: Thanks for the tip @smarter 👍

odersky:
LGTM

PR #1935, opened 2017-02-02
Fix slack integration and allow empty to gh-pages

felixmulder:


project/scripts/genDocs
```diff
@@ -57,10 +57,10 @@ if [ "$1" = "test" ]; then
 
     # add all contents of $PWD to commit
     git add -A
-    git commit -m "Update gh-pages site for $GIT_HEAD"
+    git commit -m "Update gh-pages site for $GIT_HEAD" || true
```
DarkDimius: instead of allowing it to silently fail, it would be nice to make it a separate task

felixmulder: Currently trying to get something else to work, will log here instead later instead of silently failing

PR #1920, opened 2017-01-31
Fix #1916 - fix erasure of xxl closures

odersky:
Fixes #1916. Review by @nicolasstucki

PR #1936, opened 2017-02-03
Fix #1916: fix erasure of implicit xxl closures

### LINKED ISSUE #1916: Lambda with more than 22 parameter fails cast to FunctionXXL
If the code of `tests/pos/functionXXL.scala` is executed, the there is a ClassCastException thrown at:

```scala
java.lang.ClassCastException: Test$$$Lambda$1/2051853139 cannot be cast to scala.FunctionXXL
	at Test$.<init>(functionXXL.scala:28)
	at Test$.<clinit>(functionXXL.scala)
	at Test.main(functionXXL.scala)
```

nicolasstucki:
This commit extends the fix done in #1920 to implicit closures.


odersky:
LGTM

Issue #1805, opened 2016-12-15
Name-based pattern matching rules


odersky:
This is an attempt to come up with a sane set of rules for name-based pattern matching. To recall: In dotty, name-based is all we have, there is no special treatment of type classes anymore. 

The purpose of the rules is, given a pattern `P(P1, .., Pn)`, to produce the expected types `PT1,..., PTn` against which the argument patterns `P1, .., Pn` will be matched. Let `X` be the expected type of the whole pattern match. 

1. If the companion object `P` contains a member `def unapply(x: T): U` such that `X` conforms to `T` then

    1. If `U` is a subtype of a product type `Product{n}[T1, ... Tn]` with arity `n` then `PTi = Ti`, for `i = 1,..,n`.
    2. If `U` has parameterless (`def` or `val`) members `isEmpty: Boolean` and `get: V`,
      then: 

          1. if there is exactly one argument (P1) then `PT1 = V`. 
          2. if there are zero or more than one arguments, then `V` must derive from a product type
             `Product{n}[V1, ..., Vn]` and `PTi = Vi`, for `i = 1,..,n`.
    3. If `U` is of type `Boolean` and `n = 0`, OK
    4. Otherwise error.

2. If the precondition (1) does not hold (i.e. no `unapply` method with the right argument type), but the companion object `P` contains a member `def unapplySeq(x: T): U` where `X` conforms to `T` then

    1. If `U` derives from `Seq[V]` for some type `V`, then `PTi = V` for `i = 1,..,n`, or `PTn = V*` if `Pn` is a repeated parameter.
    2. If `U` has parameterless (`def` or `val`) members `isEmpty: Boolean` and `get: S`,
      where `S <: Seq[V]` for some type `V`, then `PTi = V` for `i = 1,..,n`, or `PTn = V*` if `Pn` is a repeated parameter.
    3. Otherwise error.

**Note 1**: There are some subtleties with overloading which are skimmed over here. See the comment of typer.Applications#trySelectUnapply for a more detailed explanation. 

**Note 2:** A consequence of the rules is that a selector type may match patterns of different arities using the same `unapply`. One arity could trigger a product based selection (using rule 1.i), the other a get/isEmpty selection (using rule 1.ii). Scala parser combinators actually exercise that functionality. A test case is in `pos/Patterns.scala`.

Blaisorblade:
These rules don't make `Option` special, but still make `Seq` special. Could that change in the same spirit?
[...]

PR #1938, opened 2017-02-03
Change case class desugaring and decouple Products and name-based-pattern-matching

OlivierBlanvillain:
This PR changes the name based pattern matching the following way:

Eligibility condition is to extend a `NameBasedPattern` trait (instead of extending `ProductN`). 
The pattern arity and types are determined by looking at implemented `_1` to `_N` methods, where `N` is the arity of the last (subsequently) implemented method with such shape (`N` was before obtained from the `ProductN` superclass).

As a side effect, this PR lifts the 22 limitation on case classes.

@DarkDimius I remember you mentioning a change in this vein, what are you thoughts?


DarkDimius:
While I like the marker-trait approach and I'd be in favor of it if we were designing a new language, I'd prefer if we don't go for it now.

The reason is that scalac actually supports name-based pattern matching, though it wasn't documented, and there are some code-bases that use it(e.g. parser combinators). The proposed scheme is incompatible between the two approaches.

The intermediate ground could be to temporarily support both under -language:Scala2 mode, but it would mean implementing both schemes, which would only increase complexity.

I think there's a possibility to remain compatible with scalac, by having a scheme that is very close to it, like the current one used by dotty. While I agree that it would be nice to simplify it, I believe, we should remain compatible.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -344,14 +345,35 @@ object desugar {
...
     // neg/t1843-variances.scala for a test case. The test would give
     // two errors without @uncheckedVariance, one of them spurious.
-    val caseClassMeths =
-      if (isCaseClass) {
-        def syntheticProperty(name: TermName, rhs: Tree) =
-          DefDef(name, Nil, Nil, TypeTree(), rhs).withMods(synthetic)
+    val caseClassMeths = {
+      def syntheticProperty(name: TermName, rhs: Tree) =
+        DefDef(name, Nil, Nil, TypeTree(), rhs).withMods(synthetic)
+      // The override here is less than ideal: user defined productArity / productElement
```
DarkDimius: why? I think it's used a lot in userland and I don't (yet?)  see a value in removing it.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -344,14 +345,35 @@ object desugar {
...
+      // methods would be silently ignored. This is necessary to compile `scala.TupleN`.
+      // The long term solution is to remove `ProductN` entirely from stdlib.
+      def productArity =
```
DarkDimius: there's now a LOT more code synthesized per case-class. I'm worried about both the compile-time footprint and the runtime code size.

DarkDimius: I'd prefer to keep all those methods in super-classes, as it makes it a lot easier for JIT to omptimize a callsite that sees products of the same arity. In your case, it would be non optimizable by contemporary VMs, to the best of my knowledge.

DarkDimius: ping @sjrd I think this would also affect Scala.js size.

OlivierBlanvillain: I think this is what scalac does, but I guess we could keep `ProductN` when `N < 22` and synthesize only after that.

OlivierBlanvillain:
@DarkDimius I think I'm confusing names, if name based pattern matching refers to `def unapply(a: A): A` / `isEmpty` & `get`, then it's not be affected by this PR (and the trait should be called something else than `NameBasedPattern`). The changes of this PR affect the condition under which `dotty` does it's "case class" pattern matching, the one without allocation nor `isEmpty` check (is there a another name for this one?).

Let me summarize my understanding to be sure we are on the same page.

The following works as expected on `scalac` / `dotty-master` / this PR:

```scala
class A(e1: Int, e2: Int, e3: Int) {
  def isEmpty = false
  def get     = this
  def _1      = e1
  def _2      = e2
  def _3      = e3
}

object A {
  def unapply(a: A): A = a
}
```

Let's now consider `case class A(e1: Int, e2: Int, e3: Int)` desugared by the different compiler.

1. Desugared by `scalac`:

```scala
class A(e1: Int, e2: Int, e3: Int) extends Product // + case flag

object A {
  def unapply(a: A): Option[(Int, Int, Int)] = Some((a.e1, a.e2, a.e3))
}
```

2. Desugared by `dotty/master`:

```scala
class A(e1: Int, e2: Int, e3: Int) extends Product3[Int, Int, Int]

object A {
  def unapply(a: A): A = a
}
```

3. Desugared by this PR:

```scala
class A(e1: Int, e2: Int, e3: Int) extends Product with NameBasedPattern

object A {
  def unapply(a: A): A = a
}
```

Pattern matching on 1. with dotty compiles but goes through the unapply, thus allocating a Tuple3 and an Option (can be observed with `val a :: b = List(1)`). Pattern matching on 2. and 3. won't compile with scalac, and works as expected with `dotty/master` / this PR respectively.

So this `NameBasedPattern` does not affect compatibility. Furthermore, adding `isEmpty` & `get` methods to the desugaring of case classes should be enough to achieve compatibility both from master and from here.

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -794,18 +794,17 @@ class Definitions {
...
-  def productArity(tp: Type)(implicit ctx: Context) =
-    if (tp derivesFrom ProductType.symbol)
-      tp.baseClasses.find(isProductClass) match {
-        case Some(prod) => prod.typeParams.length
-        case None => -1
-      }
-    else -1
+  /** Is this type eligible for name based pattern matching?
+   *
+   *  That means either extending `scala.ProductN` or `NameBasedPattern`.
```
DarkDimius: I'd capitalize OR.
It would be also nice to "one day" make `scala.ProductN` inherit `NameBasedPattern`. 
That would simplify the rule even more.

DarkDimius:
@OlivierBlanvillain yes, we didn't speak about the same thing indeed
>if name based pattern matching refers to def unapply(a: A): A / isEmpty & get

Yes, this is a name-based pattern-matching is refers to unapply that doesn't return a `Option`. 
https://github.com/scala/scala/pull/2848 introduced it initially. 

I'd propose to rename this pr to along the lines "change case class desugaring and decouple Products and name-based-pattern-matching".
I'd also propose to keep ProductN classes for the sake of bytecode size and reduction of number of possible virtual call targets. Ie, after my proposal the correct desugaring would be 

```
class A(e1: Int, e2: Int, e3: Int) extends Product3[Int, Int, Int] with NameBasedPattern
```

This change would also remove most of the new code introduced in this PR as well as the workaround(I guess?).

compiler/src/dotty/tools/dotc/ast/Desugar.scala
sjrd: Can someone summarize what the effective changes to generated code are, e.g., as snippets of Scala or Java code of what's generated per case class?

OlivierBlanvillain: This is `case class A[T1, T2](e1: T1, e2: T2)` [compiled by 2.12](https://gist.github.com/OlivierBlanvillain/106edb8b6cc93c3066167ac12f2fdf39) vs [compiled by dotty/master](https://gist.github.com/OlivierBlanvillain/31d517d6f03757e012a74820284b4f88).

The PR replaces the replace the `ProductN` superclasses by `Product` which require synthesizing `productElements`, so the discussion is about reintroducing [these lines](https://gist.github.com/OlivierBlanvillain/106edb8b6cc93c3066167ac12f2fdf39#file-2-12-bytecode-java-L65-L86) in dotty.

OlivierBlanvillain:
> This change would also remove most of the new code introduced in this PR as well as the workaround(I guess?).

@DarkDimius We would still need it them for arity 23 and up. It makes sense to keep ProductN below that, I just need to add a test case to make sure that the synthesized code is covered :)

odersky:
@OlivierBlanvillain Why the change? Is about the 22 restriction, or are there other reasons?

OlivierBlanvillain:
@odersky Yes, it's about the 22 restriction. I need something like this for my hlist branch where I synthesize TupleN for N > 22 and erase them to arrays. An alternative would be to also synthesize & erase ProductN as needed.

odersky:
So that means with the current PR case classes stop implementing ProductN traits once they have more than 22 parameters? in that case I'd prefer to generalize Product instead.

odersky:
I like the idea of disentangling ProductN and name-based pattern matching, since the future status of ProductN is unclear. But I am reluctant to introduce yet another marker trait. Can we try instead with `Product` instead of `ProductN`?

OlivierBlanvillain:
Things mostly work using `scala.Product` instead of a new marker trait for "product pattern matching" (the one tuples and other case classes that calls `_i` directly).

This is non trivial because of types such as `Option` and `List` that are already valid return type for unapplys but also extend `Product`, thus becoming candidate for two different patterns.

In the last commit I swapped the priority of `isProductMatch` and `isGetMatch` tests which is enough to make things work for `Option` and `List`. However it breaks the "Scala-parser-combinators use case" where a single extractor is used with two different type of patterns. I could not find a way to make this one work without breaking something else... Could we deprecate this corner case?

DarkDimius:
In presence of both modes I would prefer the name-based patternatching to 
be preferred, as it is more efficient.

I also think that we should not break source compatibility like the last 
change does. It makes it impossible to compile into efficient bytecode 
without source-incompatible changes to public api of the library. Its as if 
we are propagating the problem we didn't solve to library authors.

odersky:
I agree with @DarkDimius that I'd prefer product matches to take precedence over get matches.

> This is non trivial because of types such as Option and List that are already valid return type for unapplys but also extend Product, thus becoming candidate for two different patterns.

I thought the test would be "extends Product and has the right number of `_i` selectors. As far as I can see, neither `List` nor `Option` has `_i` selectors?

OlivierBlanvillain:
I've found a way to make `Product` work without breaking anything.

Actually `List` & `Option` where considered valid Product patterns with arity 0. A simple fix is to make the inequality [on this line](https://github.com/lampepfl/dotty/pull/1938/commits/b59156300ff91dc134fa8c16fde6a624b4cdaf8e#diff-3027c21e3a4dfdce5454d6632ff8bbc3R58) strict, which adds a "you need at least a `_1`" constraint.

odersky:
> Actually List & Option where considered valid Product patterns with arity 0.

Ouch. I had not thought of that. I guess that swings the balance in favor of using a new subtrait of `Product` as our marker instead. What's a good name for it, I wonder?

OlivierBlanvillain:
If we ignore this zero arity case we could even avoid using a marker trait altogether by saying that the `def _1` method serves as a marker for product pattern matching. It would be more consistent with the rest of the language which seams is mostly name based.

Also, it looks like scalac desugars `case class C()` with a `def unapply(c: C): Boolean`, could we do the same?

odersky:
Sent from my iPhone

> On 24 Feb 2017, at 10:50, Olivier Blanvillain <notifications@github.com> wrote:
> 
> If we ignore this zero arity case we could even avoid using a marker trait altogether by saying that the def _1 method serves as a marker for product pattern matching. It would be more consistent with the rest of the language which seams is mostly name based.
> 
> Also, it looks like scalac desugars case class C() with a def unapply(c: C): Boolean, could we do the same?
> 
Yes let's do that. that means we can leave the other criterion to be "extends Product and has exactly the right number of constructors". I'd tend to still require Product for safety.

odersky:
What's the status here?

OlivierBlanvillain:
The PR is in sync with the latest comments:

- The change in [this line](https://github.com/lampepfl/dotty/pull/1938/files#diff-3027c21e3a4dfdce5454d6632ff8bbc3R58) updates the `isProductMatch` condition to have both a `_1` method and a `Product` superclass. 

- Dotty already generates a `def unapply(c: C): Boolean` for zero arity case classes (see test case in latest commit)

The changes are mergeable from my side!

OlivierBlanvillain:
Here another attempt to formalize Dotty's pattern-matching (base on #1805). This covers the *Extractor Patterns* [section of the spec](https://www.scala-lang.org/files/archive/spec/2.12/08-pattern-matching.html#extractor-patterns). Dotty support 4 different extractor patterns: Boolean Pattern, Product Pattern, Seq Pattern and Name Based Pattern.

#### Boolean Pattern

- Extractor defines `def unapply(x: T): Boolean`
- Pattern-matching on exactly `0` patterns


#### Product Pattern

- Extractor defines `def unapply(x: T): U`
- `U <: Product` (could be removed, kept for safetly?)
- `N > 0` is the maximum number of consecutive (parameterless `def` or `val`) `_1: P1` ... `_N: PN` members in `U`
- Pattern-matching on exactly `N` patterns with types `P1, P2, ..., PN`


#### Seq Pattern

- Extractor defines `def unapplySeq(x: T): U`
- `U` has (parameterless `def` or `val`) members `isEmpty: Boolean` and `get: S`
- `S <: Seq[V]`
- Pattern-matching on any number of pattern with types `V, V, ..., V`


#### Name Based Pattern

- Extractor defines `def unapply(x: T): U`
- `U` has (parameterless `def` or `val`) members `isEmpty: Boolean` and `get: S`
- If there is exactly `1` pattern, pattern-matching on `1` pattern with type `S`
- Otherwise `N > 1` is the maximum number of consecutive (parameterless `def` or `val`) `_1: P1` ... `_N: PN` members in `U`
- Pattern-matching on exactly `N` patterns with types `P1, P2, ..., PN`

In case of ambiguities, *Product Pattern* is preferred over *Name Based Pattern*.

OlivierBlanvillain:
@DarkDimius Could you give your opinion on this last comment?

odersky:
The spec looks accurate to me. One question concerns the almost equivalence of name-based and product patterns. Should we enforce that the result type of a `get` is a subtype of `Product`? In this case we could merge most of the two cases into one condition.

OlivierBlanvillain:
Indeed that would be nice, either by adding a `Product` on one side or removing it on the other (the alternative being checking for the presence of a `_1` member). I would lean towards removing `Product` altogether for the following reasons:

- It shouldn't break existing code (adding a requirement would)
- It completely solves the `extends Product with Serializable` problem
- `Product` is a weird marker trait: it requires `productElement` & `productArity` which are never in the implementation
- Less coupling with the stdlib

odersky:
OK, let's drop Product from the name based matching rules. For the moment we should still generate a Product for case classes, because existing code uses productArity and productElement. But dropping Product from the pattern matching rules makes it easier to change that later.

OlivierBlanvillain:
Updated to decouple `Product` and pattern-matching

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -846,17 +848,9 @@ class Definitions {
   }
 
   def isProductSubType(tp: Type)(implicit ctx: Context) =
-    (tp derivesFrom ProductType.symbol) && tp.baseClasses.exists(isProductClass)
+    Applications.extractorMemberType(tp, nme._1).exists
```
odersky: The `isProductSubType` method should be renamed and moved to `Applications`.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -48,12 +48,12 @@ object Applications {
...
-   *  for a pattern with `numArgs` subpatterns>
-   *  This is the case of `tp` is a subtype of the Product<numArgs> class.
+   *  for a pattern with `numArgs` subpatterns?
+   *  This is the case of `tp` has members `_1` to `_N` where `N == numArgs`.
    */
   def isProductMatch(tp: Type, numArgs: Int)(implicit ctx: Context) =
-    0 <= numArgs && numArgs <= Definitions.MaxTupleArity &&
-    tp.derivesFrom(defn.ProductNType(numArgs).typeSymbol)
+    numArgs > 0 && defn.isProductSubType(tp) &&
+    productSelectorTypes(tp).size == numArgs
```
odersky: Use `productArity` instead?

PR #1939, opened 2017-02-03
[doc] html enhancements

felixmulder:
There were a couple of things missing / forgotten, this is what's fixed in this PR:

- ability to expand docstrings using a `[+]` symbol
- type paramters added to entity title (top of page)
- return type references rendered correctly for (most) types
- type aliases now get their docstrings and references to their rhs if applicable
- adds constructors to the docs

To verify or generate the site, simply clone this branch and issue `sbt genDocs`, this will put a site in `docs/_site/index.html`.

Caveats
======
Constructors are unfortunately not modelled as `def`s therefore, their docstrings are lost. This should be fixed in a future PR.

Some references that were previously broken, are unfortunately - still broken. This PR does not aim to address this.

Comments to reviewer(s)
===================
I would like to get this in before Monday, so it'd be grrrreat if one of you guys could review this quickly - I don't need reviews from all of you (unless you REALLY REALLY want to :heart:)


felixmulder:
@olivierblanvillain: agreed :)

PR #1941, opened 2017-02-05
Better positions for infix operations

smarter:
Review by @odersky

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -750,12 +750,14 @@ object desugar {
           case Tuple(args) => args mapConserve assignToNamedArg
           case _ => right :: Nil
         }
-        Apply(Select(left, op.name), args)
+        val selectPos = Position(left.pos.start, op.pos.end, op.pos.start)
+        Apply(Select(left, op.name).withPos(selectPos), args)
       } else {
         val x = ctx.freshName().toTermName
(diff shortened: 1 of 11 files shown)


smarter:
@felixmulder Can we get rid of validate-main? It's marked pending on all new PRs so they don't get green checkmarks

felixmulder:
I agree, new bot tomorrow :)

odersky:
I wonder whether this is not a partial solution. What do we do about definition sites, e.g.

     val `&` = ...   ?

I am asking to better evaluate whether we should fix this in the trees (which might become costly in terms of resources), or whether we should fix this by reading the source and correcting for occurrences of "`" there.

smarter:
This PR does several things so I'm not sure what your pointing to with your example, could you be more specific?

odersky:
I mean, replace `Name` with `Ident` in the `...Op` nodes. That's OK as far as it goes, but if we have to do the same thing for `MemberDef` nodes we should discuss in more depth, and should discuss both together.

smarter:
OK, I see. Using`Ident` in `MemberDef` would certainly make my job easier, but I can see how this could be costly, and going back to sources shouldn't be too hard, but I think this can be discussed separately from this PR: for infix operations, reconstructing everything from source is much harder since we don't pickle the Position.point of Select nodes (another thing which would make my job easier :)). It's also much less costly since this only affects untyped trees, and it seems necessary to not misinterpet things like ``Int `&` String`` (see testcase).

odersky:
@smarter Yes, the explanations make sense. LGTM

PR #1943, opened 2017-02-05
Positioned#initialPos: Union the position of every children

smarter:
Previously we missed some children, one consequence of this is that the
position of the typed tree corresponding to the lambda "z => 1" did not
contain the position of "z".

Review by @odersky


odersky:
LGTM

PR #1952, opened 2017-02-06
Trees#namePos: Correct position for objects

smarter:
Review by @odersky, can you think of other situations where the tree name after typer is different from the source name?


xeno-by:
@smarter What about backticks?

smarter:
@xeno-by `pos.point` should correspond to the first character in the name that is not a backtick, so it should be okay

odersky:
LGTM

odersky:
> Review by @odersky, can you think of other situations where the tree name after typer is different from the source name?

I don't think there are other instances (module encode/decode)

PR #1953, opened 2017-02-07
Fix to wildapprox

odersky:
The test case f-bounded-case-class.scala exhibited a StackOverflow in wildApprox before
the fixes. The problem was due to F-bounds.


tests/run/t10170.scala
```diff
@@ -0,0 +1,7 @@
+object Test {
```
smarter: A check-file corresponding to this test is missing.

PR #1957, opened 2017-02-07
Fix dotr/dotc scripts

OlivierBlanvillain:
This PR fixes the dotr/dotc scripts.

@felixmulder Do you have a better idea than the added line in `.drone.yml`? It looks a bit brutal, but [this comment in Build.scala](https://github.com/dotty-staging/dotty/blame/22022a245be2b7dd235acf9f130f3100199b7d83/project/Build.scala#L122) suggests that these tests need to be run in isolation.


.drone.yml
```diff
@@ -34,6 +34,7 @@ pipeline:
 matrix:
   TEST:
     - test
+    - dotty-bin-tests/test
```
smarter: Not super-nice, but you could do `;test;dotty-bin-tests/test`

felixmulder: Yeah, I agree with Guillaume here - no need for this to be a separate process.

felixmulder:
Would you mind adding a test that verifies correct behaviour on incorrect `.packages` file?

I believe that if we change versions, the `.packages` gets corrupted. Perhaps the `clean` command should also delete the `.packages` file?

OlivierBlanvillain:
@felixmulder I know nothing of `.packages`, but I'm guessing this is not the "correct behaviour on incorrect `.packages` file":

```bash
$ bin/dotc
find: `/home/yolo/workspace/dotty/bin/../interfaces/target/dotty-interfaces-0.1.1-SNAPSHOT.jar': No such file or directory
find: `/home/yolo/workspace/dotty/compiler/target/scala-2.11/dotty-compiler_2.11-0.1.1-SNAPSHOT.jar': No such file or directory
find: `/home/yolo/workspace/dotty/bin/../library/target/scala-2.11/dotty-library_2.11-0.1.1-SNAPSHOT.jar': No such file or directory
find: `/home/yolo/workspace/dotty/library/target/scala-2.11/dotty-library_2.11-0.1.1-SNAPSHOT-tests.jar': No such file or directory
Error: Could not find or load main class dotty.tools.dotc.Main
```

felixmulder:
So the scripts create a file called `.packages` in the root of the project with a format that the scripts read. If you delete the `.packages` file and re-run the script, it should create a valid one.

OlivierBlanvillain:
@felixmulder something like this: https://github.com/lampepfl/dotty/pull/1957/commits/ca6563a2d47d5a340f13c97b570afb923101de8a?

felixmulder:
One option would be to screw with the version numbers in the packages file.

OlivierBlanvillain:
I manually corrupted the file with `sed -i s/olivier/yolo/g .packages` and the script failed (this was https://github.com/lampepfl/dotty/pull/1957#issuecomment-278014206)

OlivierBlanvillain:
@felixmulder ec1011a handls `.package` corruption

liufengyun:
@felixmulder : Sure, I think it's better to have that change in:

```
+      - ln -s /var/cache/drone/ivy2 $HOME/.ivy2
        - ./scripts/update-scala-library
 -      - sbt -J-Xmx4096m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m -Ddotty.drone.mem=4096m -ivy /var/cache/drone/ivy2 "${TEST}"
 +      - sbt -J-Xmx4096m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m -Ddotty.drone.mem=4096m "${TEST}"

```

@OlivierBlanvillain could you please add that?

OlivierBlanvillain:
```sh
$ drone sign lampepfl/dotty
Error: you must provide the Drone server address.
$ export DRONE_SERVER=dotty-ci.epfl.ch
$ drone sign lampepfl/dotty
Error: you must provide your Drone access token.
```

PR #1958, opened 2017-02-08
Add "enum" construct

odersky:
This is a prototype implementation to add an "enum" construct to Scala. Scala enums give a more concise notation for 

 - enums as in Java
 - ADTs
 - GADTs

A specification of what's implemented is provided by #1970.

Current status:

 - [x] First implementation
 - [x] Some test cases
 - [x] A specification
 - [x] A decision whether we want to go ahead with this

If we want to go ahead with this, the following todos should be broken out as separate issues.

 - An implementation of generic programming in the style of SYB. We need to clarify first exactly
    what we want from the compiler
 - A way to interop with Java enumerations
 - Update the exhaustiveness checking in the pattern matcher
 - Optimize pattern matcher to use enumTags


tests/run/enum-Color.scala
```diff
@@ -0,0 +1,13 @@
+enum Color {
+  case Red
+  case Green
+  case Blue
+}
```
viktorklang: Will

    enum Color { case Red, Green, Blue }

expand to the equivalent of the above?

odersky: Not yet. I was thinking it would be a nice abbreviation, but not decided yet.

odersky:
@felixmulder The build failure comes from my changes to how doc comments are treated in Parser. The changes are in b20c1be. I noted a lot of `docstring` parameters passed, but they were all in perfect sync with the `start` position that we also pass, and they were all computed as `in.getDocString(start)`. I foolishly assumed that `getDocString` was referentially transparent and could be called at the point where we need the docstring instead. But that seems to be a wrong assumption. My attempted fix in 4e0dc0e did not help. 

So, we could back out, which is slightly messy but doable. But maybe it's best to go forward instead. Can you make `getDocString` referentially transparent, so that the timing when we call it does not matter? I would simply manage a map from a token start offset to the docstring that immediately proceed it. Then, when we have start positions of trees (which are exact) we automatically have their docstrings as well, and a lot of redundancy and passing things around becomes unnecessary. WDYT?

felixmulder:
I think that originally I wanted to protect from all cases of misuse. E.g:

```scala
trait T {
  /** Cheeky docstring that shouldn't appear in `C` */
}
class C
```

ce40517 is a bit more permissive and allows this misuse...

Another alternative to this fix could be to make docstring its own token and attach it as part of the AST, but I guess that would be added complexity to the parser.

odersky:
> I think that originally I wanted to protect from all cases of misuse. E.g:

    trait T {
       /** Cheeky docstring that shouldn't appear in `C` */
    }
    class C

But if you associate doc comments to start positions of tokens following them, you can do that. The cheeky doc comment would be associated with `}' not with `class`. So it would not qualify.

Re: stdlib I forgot to push a new one. (the old one uses `enum` in one place, which is not a keyword).

tests/run/enum-Color.scala
vicaba: `enum Color { case Red, Green, Blue }`  is the word "case" needed in the abbreviation?

odersky: We definitely should keep `case`

viktorklang:
+💯

-- 
Cheers,
√

odersky:
Regarding shorthand for simple enums: Instead of

    enum Color { case Red, Green, Blue }

should it rather be:

    enum Color { case Red | Green | Blue }

? That would align it with pattern syntax.

odersky:
We now have the syntax

    enum Color { case Red, Green, Blue }

supported. 

I propose we defer all other todos to separate issues and skip straight to the decision. Do we want to go ahead with this concept? If everyone is still positive, let's review the code and get this in.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -410,17 +429,26 @@ object desugar {
...
         val parent =
           if (constrTparams.nonEmpty ||
               constrVparamss.length > 1 ||
               mods.is(Abstract) ||
               constr.mods.is(Private)) anyRef
+          else
             // todo: also use anyRef if constructor has a dependent method type (or rule that out)!
-          else (constrVparamss :\ classTypeRef) ((vparams, restpe) => Function(vparams map (_.tpt), restpe))
+            (constrVparamss :\ (if (isEnumCase) applyResultTpt else classTypeRef)) (
+              (vparams, restpe) => Function(vparams map (_.tpt), restpe))
```
liufengyun: I may miss some point here, it seems to me the fold is useless here, because `constrVparamss` can only be of size 0?

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -288,17 +291,26 @@ object desugar {
...
+    val isEnum = mods.hasMod[Mod.Enum]
+    val isEnumCase = isLegalEnumCase(cdef)
     val isValueClass = parents.nonEmpty && isAnyVal(parents.head)
       // This is not watertight, but `extends AnyVal` will be replaced by `inline` later.
 
-    val constrTparams = constr1.tparams map toDefParam
+    val originalTparams =
+      if (isEnumCase && parents.isEmpty) {
+        if (constr1.tparams.nonEmpty)
+          ctx.error(em"case with type parameters needs extends clause", constr1.tparams.head.pos)
```
liufengyun: What if the parent takes no type parameters, while the child has type parameters, like the following case:
```Scala
enum Base {
  case Bar[T](x: T)
}
```

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2080,11 +2098,73 @@ object Parsers {
...
+        TypeDef(clsName, Template(constr, parents, EmptyValDef, Nil))
+          .withMods(addMod(mods, enumMod)).setComment(in.getDocComment(start))
+      }
+      newLineOptWhenFollowedBy(LBRACE)
+      val modDef = atPos(in.offset) {
+        val body = inBraces(enumCaseStats)
+        ModuleDef(modName, Template(emptyConstructor, Nil, EmptyValDef, body))
+          .withMods(mods)
+      }
+      Thicket(clsDef :: modDef :: Nil)
```
odersky: I tried multi-level expansion first, including flattening, but it did not work. Essentially the compiler got confused with symbol attachments and I did not find a good way around it.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
odersky: Good point. An example for this are HLists:

    enum HList {
       case HCons[Hd, Tl](hd: Hd, tl: Tl)
       case HNil
     }

I think we want to allow this.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
odersky: size 0 or 1

odersky: size 0 or 1, actually.

liufengyun: Ah, I see, thanks, obviously I've some problems with basic math :)

odersky:
Rebased to master

odersky:
As far is I'm concerned we are done for this part. We already had a review by @liufengyun but if people want to review either everything or just the later commits please assign to yourself as reviewer.

liufengyun:
One thing missing is exhaustivity check, which can be done in another PR.

According to our current algorithm, as long as the parent class symbol carries a list of children symbols (case vals and case classes), the algorithm will be able to handle.

My concern here is that in parser, it's clear what the children are, and it's easy to get them. But after desugaring, it seems to require a lot of assumptions about desugaring in order to recover that information.

PR #1962, opened 2017-02-09
Factor out logic for scala functions.

nicolasstucki:
This PR factors out code from #1408 aimed to make an abstraction on top of all different kinds of functions. Implemented `Function0` to `Function22`, synthetic `FunctionN` with `N > 22`, `ImplicitFunctionN` and erased `FunctionXXL`. #1408 extends it with `PhantomFunctionM` and `ImplicitPhantomFunctionM`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -635,13 +635,9 @@ class Definitions {
       FunctionType(args.length, isImplicit).appliedTo(args ::: resultType :: Nil)
     def unapply(ft: Type)(implicit ctx: Context) = {
       val tsym = ft.typeSymbol
-      val isImplicitFun = isImplicitFunctionClass(tsym)
-      if (isImplicitFun || isFunctionClass(tsym)) {
-        val targs = ft.argInfos
-        val numArgs = targs.length - 1
-        if (numArgs >= 0 && FunctionType(numArgs, isImplicitFun).symbol == tsym)
-          Some(targs.init, targs.last, isImplicitFun)
-        else None
+      if (isFunctionClass(tsym)) {
+        val targs = ft.dealias.argInfos
+        Some(targs.init, targs.last, tsym.name.isImplicitFunction)
       }
       else None
     }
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -108,32 +109,28 @@ class Definitions {
...
           }
           else (MethodType, Nil)
-        val applyMeth =
-          decls.enter(
-            newMethod(cls, nme.apply,
-              methodType(argParams.map(_.typeRef), resParam.typeRef), Deferred))
-        denot.info =
-          ClassInfo(ScalaPackageClass.thisType, cls, ObjectType :: parentTraits, decls)
+
+        decls.enter(newMethod(cls, nme.apply, methodType(tParamTpes.init, tParamTpes.last), Deferred))
```
OlivierBlanvillain: With a `.init` and `.last` here I'm not actually sure I prefer `tParamTpes` to `argParams` & `resParam`

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -688,26 +684,14 @@ class Definitions {
   lazy val AbstractFunctionType = mkArityArray("scala.runtime.AbstractFunction", MaxImplementedFunctionArity, 0)
   val AbstractFunctionClassPerRun = new PerRun[Array[Symbol]](implicit ctx => AbstractFunctionType.map(_.symbol.asClass))
   def AbstractFunctionClass(n: Int)(implicit ctx: Context) = AbstractFunctionClassPerRun()(ctx)(n)
-  private lazy val ImplementedFunctionType = mkArityArray("scala.Function", MaxImplementedFunctionArity, 0)
+  lazy val ImplementedFunctionType = mkArityArray("scala.Function", MaxImplementedFunctionArity, 0)
```
OlivierBlanvillain: What about keeping this private and adding:?

```scala
lazy val Function0Type = ImplementedFunctionType(0)
lazy val Function1Type = ImplementedFunctionType(1)
```

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -780,7 +760,8 @@ class Definitions {
    *  trait gets screwed up. Therefore, it is mandatory that FunctionXXL
    *  is treated as a NoInit trait.
    */
-  lazy val NoInitClasses = PhantomClasses + FunctionXXLClass
+  def isNoInitClass(cls: Symbol) =
+    cls.is(NoInitsTrait) || PhantomClasses(cls) || ScalaFunction(cls).exists
```
OlivierBlanvillain: Is this `PhantomClasses.contains(cls)`?

compiler/src/dotty/tools/dotc/core/Definitions.scala
nicolasstucki: Now I agree, the change was done on an intermediate version where it saved a considerable amount of code. This is clearly not the case anymore. I will change it back.

compiler/src/dotty/tools/dotc/core/Definitions.scala
nicolasstucki: This would work nicely. Thought will define them as `def ` as `ImplementedFunctionType` is an array.

nicolasstucki:
Yes, in this PR the advantages does not really exist. It is only when adding the functions with phantom parameters that this starts to help. I will redesign it to to be closer to the original code to reduce the cognitive load while making it easy to add the new types of functions.

nicolasstucki:
In this PR we:
* Add `isSyntheticFunction` checks for synthetic functions such as FunctionN for N > 22 and ImplicitFunctionN for N >= 0.
* Add `erasedFunctionClass` to get the erased version of synthetic functions.
* Change the semantics of `isFunctionClass` to return  true if it is any kind of FunctionN or ImplicitFunctionN.

odersky:
LGTM! Great cleanup. 👍

PR #1963, opened 2017-02-09
The dotty-bot

felixmulder:
This bot is meant to be our new build automation hero. Currently it does the CLA checking of all commits (both commiters and authors), there's no restriction on X commits, it handles the pagination gracefully.

With this merged, I can tell Fabien to host it for us so that we can snip the cord to the old CI infra for good.


smarter:
Shouldn't it be compiled and tested by the CI?

smarter:
Also, as much as I like mono-repositories, this is completely separate from dotty and could be its own project really.

felixmulder:
It should indeed!

Whether or not it should be a part of the mono-repo, I'll leave that up to discussion here. Since it's its own separate sbt-project, it shouldn't be bothersome for the rest of the project, and improvements will be noticed by contributors.

Going to sleep some more on it, cheers

Felix

PR #1966, opened 2017-02-10
Fix #1965: add proper testing infrastructure for reporting tests

### LINKED ISSUE #1965: Improve error messages tests by providing unit test support
Currently, testing error messages is done by using check files in the `./tests/repl/` dir. This has several drawbacks, one of the largest being that these are hard to keep in sync since they are only tested by diffing.

It would be great to have a unit test that would work along the lines of:

```scala
@Test def typeMismatch =
  checkErrors (
    """|class Foo {
       |  def bar: Int = "IIEK"
       |}""".stripMargin,
  )
  .expect { implicit ctx =>
    val defn = ctx.definitions
    TypeMismatch(defn.StringType, defn.IntType) :: Nil
  }
```

To do this, we need to extend `DottyTest` and provide the context with a custom reporter that aggregates the errors instead of rendering them to stdout.

felixmulder:


compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
```diff
@@ -0,0 +1,70 @@
...
+        buffer.clear()
+
+        val ctx = capturedContext
+        capturedContext = null
+
+        new Report(xs, ctx)
+      }
+  }
+
+  ctx = freshReporter(ctx)
```
OlivierBlanvillain: I would move the mutation stuff at the very top of the trait. BTW could you make it a class instead? (it feels wrong to have class → trait → class inheritance)

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Assert._
+
+trait ErrorMessagesTest extends DottyTest {
+
+  class Report(messages: List[Message], ictx: Context) {
+    def expect(f: (Context, List[Message]) => Unit): Unit = {
+      f(ictx, messages)
+    }
+
+    def expectNoErrors: Unit = this.isInstanceOf[FailedReport]
```
OlivierBlanvillain: Did you mean `asInstanceOf`?

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    ctx.fresh.setReporter(new CapturingReporter)
+
+  def checkMessages(source: String): Report = {
+    checkCompile("frontend", source) { (_,ictx) => () }
+    val rep = ctx.reporter.asInstanceOf[CapturingReporter].toReport
+    ctx = freshReporter(ctx)
+    rep
+  }
+
+  def messageCount(expected: Int, messages: List[Message]): Unit =
```
OlivierBlanvillain: `assertMessageCount`

OlivierBlanvillain:
Looks good otherwise :)

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
felixmulder: No, but I should probably rename `FailedReport` => `EmptyReport`

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
felixmulder: Maybe `DottyTest` should be a trait instead?

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
felixmulder: :+1:

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
OlivierBlanvillain: Then I don't understand the statement, this computes a `Boolean` and converts it to `Unit` regardless of it's value?

PR #1968, opened 2017-02-11
Change 'overrides nothing' to report via Message (see #1965)

ennru:
'overrides nothing' is now reported as Message as descibed in #1589 (even though it is not listed in that issue.
I took the liberty to split the error into two different messages:
* overrides nothing - when name doesn't exist
* method foo has a different signature than the overridden declaration - when the name exists, but is not overriden by the declaration

Please review the explaining texts thoroughly as I'm not really sure what to call things properly.

I noticed `SingleDenotation.showDcl` creates non-valid code e.g. `def foo(i: Int)Unit` without the colon. Is there a better way to print declarations in the error messages?

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -21,6 +21,9 @@ import DenotTransformers._
 
 object RefChecks {
   import tpd._
+  import reporting.diagnostic.Message
+  import reporting.diagnostic.messages._
+
 
   private def isDefaultGetter(name: Name): Boolean =
     name.isTermName && name.asTermName.defaultGetterIndex >= 0
@@ -622,15 +625,12 @@ object RefChecks {
       if (member.isAnyOverride && !(clazz.thisType.baseClasses exists (hasMatchingSym(_, member)))) {
         // for (bc <- clazz.info.baseClasses.tail) Console.println("" + bc + " has " + bc.info.decl(member.name) + ":" + bc.info.decl(member.name).tpe);//DEBUG
 
-        val nonMatching = clazz.info.member(member.name).altsWith(alt => alt.owner != clazz && !alt.is(Final))
-        def issueError(suffix: String) =
-          ctx.error(i"$member overrides nothing$suffix", member.pos)
+        val nonMatching = clazz.info.member(member.name).altsWith(alt => alt.owner != clazz)
         nonMatching match {
           case Nil =>
-            issueError("")
+            ctx.error(OverridesNothing(member), member.pos)
           case ms =>
-            val superSigs = ms.map(_.showDcl).mkString("\n")
-            issueError(s".\nNote: the super classes of ${member.owner} contain the following, non final members named ${member.name}:\n${superSigs}")
+            ctx.error(OverridesNothingButNameExists(member, ms), member.pos)
         }
         member.resetFlag(Override)
         member.resetFlag(AbsOverride)
--- a/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
+++ b/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
@@ -57,10 +57,10 @@ class ErrorMessagesTests extends ErrorMessagesTest {
 
       assertMessageCount(1, messages)
       val OverridesNothing(member) :: Nil = messages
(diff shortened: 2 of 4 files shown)


compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -42,4 +42,47 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+        |}
+      """.stripMargin
+    }
+    .expect { (ictx, messages) =>
+      implicit val ctx: Context = ictx
+      val defn = ictx.definitions
+
+      assertMessageCount(1, messages)
+      val OverridesNothing(member) :: Nil = messages
+      assertEquals("bar", member.name.toString)
```
felixmulder: Doesn't really matter in this case, but show automatically decodes "weird" names - so let's set a good precedence :+1:

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -622,15 +625,12 @@ object RefChecks {
       if (member.isAnyOverride && !(clazz.thisType.baseClasses exists (hasMatchingSym(_, member)))) {
         // for (bc <- clazz.info.baseClasses.tail) Console.println("" + bc + " has " + bc.info.decl(member.name) + ":" + bc.info.decl(member.name).tpe);//DEBUG
 
-        val nonMatching = clazz.info.member(member.name).altsWith(alt => alt.owner != clazz && !alt.is(Final))
-        def issueError(suffix: String) =
-          ctx.error(i"$member overrides nothing$suffix", member.pos)
+        val nonMatching = clazz.info.member(member.name).altsWith(alt => alt.owner != clazz)
```
felixmulder: Just looking at the compressed diff here - so I can't see the reason to cut `!alt.is(Final)`

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -42,4 +42,47 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+    .expect { (ictx, messages) =>
+      implicit val ctx: Context = ictx
+      val defn = ictx.definitions
+
+      assertMessageCount(1, messages)
+      val OverridesNothingButNameExists(member, sameName) :: Nil = messages
+      // check expected context data
+      assertEquals("bar", member.name.toString)
+      assertEquals(3, sameName.size)
+      assert(sameName.forall(_.symbol.name.toString == "bar"),
```
felixmulder: `show`

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -42,4 +42,47 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+      """.stripMargin
+    }
+    .expect { (ictx, messages) =>
+      implicit val ctx: Context = ictx
+      val defn = ictx.definitions
+
+      assertMessageCount(1, messages)
+      val OverridesNothingButNameExists(member, sameName) :: Nil = messages
+      // check expected context data
+      assertEquals("bar", member.name.toString)
```
felixmulder: `show`

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
ennru: The former error message listed only non-final methods with the same name, but I think it is useful to have even final methods in the "not eligible" list of methods so that one can easily see it as the modifiers are included.

felixmulder: Sure, makes sense :)

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
ennru: Ah, than I reckon I should use `member.name.show` in the message and explanation, as well - or does `hl` do that for me?

felixmulder: `hl` should do that for you

felixmulder:
Regarding `showDcl`, it should probably be fixed outside of this PR eventually.

felixmulder:
Thanks @ennru! :tada:

PR #1972, opened 2017-02-13
TreeMap/TreeAccumulator: proper context for inlined trees

smarter:
This was already be done in TreeTraverser but should also be done in
TreeMap and TreeAccumulator for ctx.error(..., tree.pos) to not use
completely incorrect positions inside inlined trees.


compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -1077,6 +1077,13 @@ object Trees {
     /** Hook to indicate that a transform of some subtree should be skipped */
     protected def skipTransform(tree: Tree)(implicit ctx: Context): Boolean = false
 
+    /** For untyped trees, this is just the identity.
+     *  For typed trees, a context derived form `ctx` that records `call` as the
```
odersky: derived from

PR #1973, opened 2017-02-13
Ennru forward reference error

ennru:
@felixmulder
Another awsome error message giving more relevant context. 

There seem to be other checks for this error around that expect the old message, but I did not find out how to run those tests.

I removed the `ctx.debuglog` as that symbol is now included in the message.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1034,4 +1034,23 @@ object messages {
            |"""
   }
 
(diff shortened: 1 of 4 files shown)


felixmulder:
@liufengyun - debug test failing in this build. Could you investigate?

liufengyun:
@felixmulder I just restarted the test, seems to be some non-determinism. I have reduced the probability it can happen last week. If it happens > 5%, I am happy to disable it for the moment, until I figure out a solution.

felixmulder:
@liufengyun - I'll be at the office in an hour, let's talk about it then. I'd like to eliminate the non-determinism if possible: in a couple of years this knowledge could be lost.

felixmulder:
Thanks @ennru !

PR #1974, opened 2017-02-13
Avoid accidental captures of Context

smarter:
These issues where detected using the made-for-the-occasion `CheckCaptures` phase available at https://github.com/dotty-staging/dotty/commits/checkClosures (see last two commits for more information). Ideally, `CheckCaptures` or something like it would be part of the compiler to prevent future accidental captures, but the current implementation requires adding more than 150 annotations to our codebase which is not very practical, see also https://github.com/TiarkRompf/scala-escape for a much more advanced and safer alternative that could be an inspiration for future work.

Most of the fixed issues are hypothetical but https://github.com/lampepfl/dotty/commit/1ccee1f0926bbeff01a0aed2c5126ba13cf889ea caused a `StaleSymbolError` in the Language Server and https://github.com/lampepfl/dotty/commit/08cd627597b106b37c45c0151d76a3ba1967ef55 apparently fixing an error message.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Annotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Annotations.scala
@@ -117,11 +117,17 @@ object Annotations {
       }
 
     /** Create an annotation where the symbol and the tree are computed lazily. */
-    def deferredSymAndTree(sym: => Symbol, treeFn: Context => Tree)(implicit ctx: Context): Annotation =
+    def deferredSymAndTree(symf: Context => Symbol, treeFn: Context => Tree)(implicit ctx: Context): Annotation =
       new LazyAnnotation {
-        lazy val symf = sym
+        private[this] var mySym: Symbol = _
 
-        override def symbol(implicit ctx: Context): Symbol = symf
+        override def symbol(implicit ctx: Context): Symbol = {
+          if (mySym == null) {
+            mySym = symf(ctx)
+            assert(mySym != null)
+          }
+          mySym
+        }
         def complete(implicit ctx: Context) = treeFn(ctx)
       }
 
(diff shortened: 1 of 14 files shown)


compiler/src/dotty/tools/dotc/typer/ImportInfo.scala
```diff
@@ -27,14 +27,14 @@ object ImportInfo {
  *  @param   isRootImport true if this is one of the implicit imports of scala, java.lang,
  *                        scala.Predef or dotty.DottyPredef in the start context, false otherwise.
  */
-class ImportInfo(symf: => Symbol, val selectors: List[untpd.Tree],
+class ImportInfo(symf: Context => Symbol, val selectors: List[untpd.Tree],
                  symNameOpt: Option[TermName], val isRootImport: Boolean = false)(implicit ctx: Context) {
```
odersky: can we delete this `ctx` parameter?

compiler/src/dotty/tools/dotc/core/Annotations.scala
```diff
@@ -117,11 +117,17 @@ object Annotations {
...
-    def deferredSymAndTree(sym: => Symbol, treeFn: Context => Tree)(implicit ctx: Context): Annotation =
+    def deferredSymAndTree(symf: Context => Symbol, treeFn: Context => Tree)(implicit ctx: Context): Annotation =
       new LazyAnnotation {
-        lazy val symf = sym
-
-        override def symbol(implicit ctx: Context): Symbol = symf
+        private[this] var mySym: Symbol = _
+
+        override def symbol(implicit ctx: Context): Symbol = {
+          if (mySym == null) {
```
odersky: This one looks fishy. Should we not recompute the symbol here maewhen its validity period does not fall in the current run? ImportInfo has a similar problem, but there I think one can argue that it does not matter, as ImportInfos are either global (live forever) or are local to one compilation unit. But for annotations I don't think we can make that argument.

smarter:
@odersky: The last two commits should address your comments.

smarter:
I added two more commits to address the remaining context captures issues

odersky:
LGTM 👍

PR #1977, opened 2017-02-14
Fix #1976: Hack to support scala.xml's $scope

### LINKED ISSUE #1976: XML typing issue - not sure if valid
```scala
object Test {
  import scala.xml._
  val node = <node>{ "whatever " }</node>
}
```

```
-- [E006] Unbound Identifier Error: local/xml.scala ----------------------------
3 |  val nodes: Node = <node>{ "whatever" }</node>
  |                     ^
  |                     not found: $scope
```

odersky:
This is a gross hack to support the need for a $scope
binding when XML literals are created. It should go away
once we phase out XML literals.  A library-based
solution should use implicits instead.

Note: scalac uses another hack which is more elaborate. But 
since this is temporary anyway I think the simplest thing that works
is fine.


felixmulder:
Updated the branch to see if @liufengyun fix to debug tests works, fingers crossed

PR #1981, opened 2017-02-14
Remove debugging for now, too many spurious failures

felixmulder:
Unfortunately the problems with the debug tests have persisted. We need to revert it.


felixmulder:
duplicate

PR #1984, opened 2017-02-15
Fix #1747: Improve error message for Scala/Java type mismatch

### LINKED ISSUE #1747: Can't implement or override java method with type parameters.

odersky:


compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -111,11 +111,23 @@ object ErrorReporting {
...
+        case tp: PolyType =>
+          tp.derivedPolyType(resType = dropJavaMethod(tp.resultType))
+        case tp: JavaMethodType =>
+          MethodType(tp.paramNames, tp.paramTypes, dropJavaMethod(tp.resultType))
+        case tp => tp
+      }
+      val found1 = dropJavaMethod(found)
+      val expected1 = dropJavaMethod(expected)
+      if ((found1 eq found) != (expected eq expected1) && (found1 <:< expected1))
+        "\n (Note that Scala's and Java's representation of this type differs)"
```
nicolasstucki: This might not be enough for the user to understand the issue. In the example we would get 
```
overriding method toArray in trait Collection of type [T](x$0: Array[T])Array[T];
       method toArray of type [T](a: Array[T])Array[T] has incompatible type
       (Note that Scala's and Java's representation of this type differs)
```

Would it be possible to hint that to make the representation match one would need to write  
`def toArray[T](a: Array[T with Object]): Array[T with Object]`?

nicolasstucki:
That is all. Otherwise LGTM.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
odersky: I thought about it, but it is really tricky and would require quite a bit of code. In short interoperating with Java using arrays requires a lot of context knowledge. Hard to condense that in an error message.

PR #1985, opened 2017-02-15
Change 'mixed left- and right-associative operators' to Message

ennru:
@felixmulder 
This becomes a habit. Next syntax error to the Message scheme, with more explanation from the language specs.
Not really sure if `in.name` is the correct name of the second operator in second part of`infixOps`.


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1086,4 +1086,37 @@ object messages {
...
+          |  (all letters)
+          |  |
+          |  ^
+          |  &
+          |  = !
+          |  < >
+          |  :
+          |  + -
+          |  * / %
+          |  (all other special characters)
```
felixmulder: Hmm, just reading this with my noob-glasses on, I'd think that letters had the greatest precedence. Then I'd see the sentence on the next line...

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1086,4 +1086,37 @@ object messages {
...
   }
 
+  case class MixedLeftAndRightAssociativeOps(op1: Name, op2: Name, op2LeftAssoc: Boolean)(implicit ctx: Context)
+    extends Message(41) {
+    val kind = "Syntax"
+    val op1Asso = if (op2LeftAssoc) "which is right-associative" else "which is left-associative"
+    val op2Asso = if (op2LeftAssoc) "which is left-associative" else "which is right-associative"
+    val msg = s"${hl"`${op1}`"} (${op1Asso}) and ${hl"`${op2}`"} ($op2Asso) have same precedence and may not be mixed"
+    val explanation =
+      s"""|The operators ${hl"${op1}"} and ${hl"${op2}"} are used as infix operators in the same expression,
```
felixmulder: why not use the `hl` interpolator? You still have to interpolate values for them to be highlighted. If there is some interpolated part of the message that gets highlighted incorrectly, you can wrap it in `NoColor(..)` or some other "color" that you find more suitable.

Anyway, choose the lesser evil - if `hl` requires more boilerplate, then of course `s` is preferable.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -404,14 +404,13 @@ object Parsers {
 
     var opStack: List[OpInfo] = Nil
 
-    def checkAssoc(offset: Int, op: Name, leftAssoc: Boolean) =
+    def checkAssoc(offset: Int, op: Name, leftAssoc: Boolean, op2: Name) =
```
felixmulder: I'm assuming you want to avoid a larger diff, but it looks weird that `op2` is at the end. If you want to refactor that - put it in a second commit and we'll merge both of them unsquished.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: You're right. I added a bit more from [the language spec](http://scala-lang.org/files/archive/spec/2.11/06-expressions.html#infix-operations).
Would be an idea to provide links to the spec or Stack overflow...

felixmulder:
Github seems to struggle with your rebased PRs, it might be due to the email you're signing commits with not being associated with your Github account.

I rebased your branch and merged it as #2007 instead, hope that's ok with you!

ennru:
@felixmulder_ Thank you for pointing out my wrong git user as the problem, I was wondering what was going on.
Sorry for the extra work this resulted in for you.

felixmulder:
No worries :)

PR #1993, opened 2017-02-17
Treat implicit by-name arguments as lazy values

odersky:
With the previous rules, the two test cases produce a
diverging implicit expansion. We avoid this by creating
for every implicit by-name argument of type T a lazy implicit
value of the same type. The implicit value is visible for all
nested implicit searches of by-name arguments. That way,
we tie the knot and obtain a recursive lazy value instead of
a diverging expansion.


odersky:
@milessabin: I got this idea from our conversation at Scala.IO. Does it correspond to what you had in mind?

PR #1994, opened 2017-02-17
Fix #1991: Use classtag where available in unapply

### LINKED ISSUE #1991: ClassTag-based pattern matching does not work with extractors
```scala
import scala.reflect.ClassTag

class A[Foo](implicit tag: ClassTag[Foo]) {
  object ExtractFoo {
    def unapply(foo: Foo): Boolean = true
  }

  def isFoo(x: Any) = x match {
    case ExtractFoo() => true
    case _ => false
  }
}

object Test {
    assert((new A[String]).isFoo("foo")) // OK
    assert(!(new A[String]).isFoo(42)) // OK in scalac, fails in Dotty
  }
}
```

The ClassTag handling is done in [`typedTyped`](https://github.com/lampepfl/dotty/blob/6df672c7e7be65d7be1cd6524c610aed4f35178c/compiler/src/dotty/tools/dotc/typer/Typer.scala#L525-L533) but the extractor handling is done in [`typedUnapply` where we create a `Typed` node using `tpd.Typed` which does not call `typedTyped`](https://github.com/lampepfl/dotty/blob/6df672c7e7be65d7be1cd6524c610aed4f35178c/compiler/src/dotty/tools/dotc/typer/Applications.scala#L919)

[This is needed to get the dotty-compiled shared backend to have the correct runtime behavior.](https://github.com/DarkDimius/scala/blob/40bdc7b65a766f0fbe8ee9ccd744647d3d58a7ef/src/compiler/scala/tools/nsc/backend/jvm/BackendInterface.scala#L57-L88)

odersky:


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1123,7 +1130,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     val body1 = typed(tree.body, pt1)
     typr.println(i"typed bind $tree pt = $pt1 bodytpe = ${body1.tpe}")
     body1 match {
-      case UnApply(fn, Nil, arg :: Nil) if tree.body.isInstanceOf[untpd.Typed] =>
+      case UnApply(fn, Nil, arg :: Nil) if tree.body.isInstanceOf[untpd.Typed] && !body1.tpe.isError =>
```
smarter: This is now too restrictive, it means that the following does not typecheck:
```scala
  def isFoo(x: Any) = x match {
    case x @ ExtractFoo() =>
      val b: Foo = x
      true
```
Because at this point we have:
```scala
tree.body = Apply(Ident(ExtractFoo),List())
```

(Note that explicitly writing `x @ (_: Foo)` does not work either because then `tree.body` is an instance of `untpd.Parens`, not `untpd.Typed`)

smarter:
Your commit LGTM, feel free to merge after reviewing mine.

odersky:
@smarter Your change makes sense. Thanks!

PR #1995, opened 2017-02-17
checkNoPrivateLeaks: Use correct position for errors

smarter:
Previously we never used the `pos` argument of `checkNoPrivateLeaks` and
instead used `sym.pos`, this makes a difference for calls to
`avoidPrivateLeaks` coming from `TreeUnpickler` where we should use
`tree.pos` instead.


odersky:
And it gets simpler that way, too! Nice.

PR #1996, opened 2017-02-18
Fix #1990: Handle inlining where this proxies change types

### LINKED ISSUE #1990: ExplicitOuter failure when calling inline method defined in inner class of superclass
Triggered while trying to compile the shared backend with dotty, here's a minimization:
```scala
class A {
  class Foo {
    inline def inlineMeth: Unit = {
      new Bar
    }
  }
  class Bar
}

class B extends A {
  (new Foo).inlineMeth
}
```

odersky:
The new situation in the test was that outer of the inlined method
was `A` but its as-seen-from type is a subtype `B`. Outer path logic
needs to be adapted to that case.


compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
```diff
@@ -355,14 +355,21 @@ object ExplicitOuter {
...
+     *  @param  outOfContext  When true, we take the `path` in code that has been inlined
+     *                        from somewhere else. In that case, we need to stop not
+     *                        just when `toCls` is reached exactly, but also in any superclass
+     *                        of `treeCls`. This compensates the `asSeenFrom` logic
+     *                        used to compute this-proxies in Inliner.
      */
-    def path(toCls: Symbol, start: Tree = This(ctx.owner.lexicallyEnclosingClass.asClass)): Tree = try {
+    def path(toCls: Symbol,
+             start: Tree = This(ctx.owner.lexicallyEnclosingClass.asClass),
+             outOfContext: Boolean = true): Tree = try {
```
smarter: Looks like this parameter is always set to true, is that intended?

odersky: No, that was an oversight.

compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
```diff
@@ -355,14 +355,21 @@ object ExplicitOuter {
     }
 
     /** The path of outer accessors that references `toCls.this` starting from
-     *  the context owner's this node.
+     *  node `start`, which defaults to the context owner's this node.
+     *  @param  outOfContext  When true, we take the `path` in code that has been inlined
+     *                        from somewhere else. In that case, we need to stop not
```
smarter: But we always have `outOfContext = true` in `transformSelect`, not just in code that has been inlined.
Shouldn't the condition be something like `!tpd.enclosingInlineds.isEmpty` instead?

odersky: Note that C_<OUTER> nodes are only created for inline proxies.

smarter: Oh, I missed that, makes sense now.

smarter:
I found the fix a bit suspicious so I played a bit with the testcase and got it to crash again:
```scala
class A { self =>
  class Foo {
    inline def inlineMeth: Unit = {
      println(self)
    }
  }
}

class C extends A {
  class B extends A
}

object Test {
    val c = new C
    val b = new c.B

    (new b.Foo).inlineMeth
  }
}
```
Stack trace: https://gist.github.com/smarter/75ee9b3da1ffe127721bb93535bd8483

smarter:
Progress! Now I can only get it to crash at runtime, I apologize for the devious code :).
```scala
trait A { self =>
  class Foo {
    inline def inlineMeth: Unit = {
      println(self)
    }
  }
}

class A2 extends A {
  class C extends Foo with A

  val c = new C
  (new c.Foo).inlineMeth
}

object Test {
    new A2
  }
}
```
```scala
$ dotr Test
Exception in thread "main" java.lang.ClassCastException: A$Foo cannot be cast to A2$C
        at A2.<init>(i1990a.scala:13)
        at Test$.main(i1990a.scala:18)
        at Test.main(i1990a.scala)
```
The crash happens because the outer path stops too early:
```scala
        val Foo_this: A2.this.c.Foo = new A2.this.c.Foo()
        val A_this: A2.this.C(A2.this.c) = 
          Foo_this.asInstanceOf[A2.this.C(A2.this.c)]
```
If I replace `class C extends Foo with A` by `class C extends A`, the path is correct:
```scala
        val Foo_this: A2.this.c.Foo = new A2.this.c.Foo()
        val A_this: A2.this.C(A2.this.c) = 
          Foo_this.A$Foo$$$outer.asInstanceOf[A2.this.C(A2.this.c)]
```

DarkDimius:
Those bugs are very similar from ones that we had in linker.
here's another test for you to try:

```scala
trait A { self =>
  type T
  val field: T
  def bar(f: T): T
  class Foo {
    def inlineMeth: T = {
      bar(field)
    }
  }
}

class A2 extends A {
  type T = Int
  val field = 1
  override def bar(a: Int) = a + 1
  class C extends Foo with A

  val c = new C
  val ret = bar((new c.Foo).inlineMeth) // how many boxings? 
  val resI: Int = ret
}

object Test {
    new A2
  }
}
```

odersky:
@darkdimius That test does not typecheck. Is that intentional?

PR #1997, opened 2017-02-18
Fix off-by-one error in forward reference checking

odersky:


nicolasstucki:
I added the test for #2005 in an additional commit.

Issue #2000, opened 2017-02-19
Override method with implicit methods.


nicolasstucki:
Currently we are unable to override a method with an implicit method as the implicit method is not a subtype of the plain method. This is in used in the [definitions of the synthetic symbols for `ImplicitFunctionN`](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/Definitions.scala#L125) but it does not work properly. I found the following two issues
 * Dotty complains when we try to override a method with its equivalent implicit method.
 * We are forced to implement `ImplicitFunctionN.apply` without the `implicit`. Making the function class implementing `ImplicitFunctionN` behave as a `FunctionN`.

```scala
object Test {
  def main(args: Array[String]): Unit = {
    implicit val x: String = "hello"
    val i0: Int = new Fun0().apply

    val i1: Int = new Fun1().apply
    //                ^^^^^^^^^^^^
    //                missing arguments for method apply in class Fun1
    //                follow this method with `_' if you want to treat it as a partially applied function

    val i2: Int = new Fun2().apply // works as expected
  }
}

class Fun0 extends ImplicitFunction1[String, Int] {
  def apply(implicit x: String): Int = 42
  // overriding method apply in trait Function1 of type (v1: String)Int;
  //   method apply of type (implicit x: String)Int has incompatible type
}

class Fun1 extends ImplicitFunction1[String, Int] {
  def apply(x: String): Int = 43
}

class Fun2 {
  def apply(implicit x: String): Int = 44
}

class A {
  def foo(x: String): Unit = ()
}

class B extends A {
  def foo(implicit x: String): Unit = ()
  // overriding method foo in trait A of type (x: String)Unit;
  //   method foo of type (implicit x: String)Unit has incompatible type
}
```

odersky:
That's intentional. Implicit methods should not be able to override normals methods. ImplicitFunctionN is an exception. it should not undergo this check.

odersky:
And, I am not sure at all we want to make implicit function classes extensible. Make them final instead?

retronym:
I believe that taking the `implicit` modifier as significant in subtyping is incorrect.

Here's an example in scalac that lets you override without the `override` keyword:

```
scala> class C[A] { def foo(a: A) = "c" }; class D extends C[String] { def foo(implicit s: String) = "d" }; new D().foo("")
defined class C
defined class D
res3: String = d
```

[...]

PR #2008, opened 2017-02-20
Decouple ImplicitFunctionN from FunctionN.

nicolasstucki:
This removes an illegal method override mentioned in #2000.


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2003,20 +2003,29 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-              // but this prevents case blocks from implementing polymorphic partial functions,
-              // since we do not know the result parameter a priori. Have to wait until the
-              // body is typechecked.
-              return cpy.Closure(tree)(Nil, id, TypeTree(pt)).withType(pt)
-            case _ =>
-          }
+        case Closure(Nil, id @ Ident(nme.ANON_FUN), _) =>
+          val adaptToPrototype =
+            if (defn.isFunctionType(pt)) {
+              // convert implicit function to function
```
odersky: Not sure about the change in this logic. Questions:

 - Do we want to convert implicit functions to functions, or to other SAM types? Does this case even arise
    as an implicit function is always immediately applied?
 - What would happen if we kept the previous code here?

nicolasstucki: It is from implicit function to function, previously this worked because implicit functions where a subtype of functions.

A concrete case that fails without this change is 
```scala
-- [E007] Type Mismatch Error: /Users/nicolasstucki/GitHub/dotty/tests/partest-generated/run/implicitFuns.scala --------
148 |        println(if (thisTransaction.isAborted) "aborted" else s"result: $res")
    |                                                                              ^
    |                                                                        found:    implicit Transaction => Unit
    |                                                                        required: Transaction => Any
```

where

```scala
  def transaction[T](op: Transaction => T) = {
    ...
  }

    transaction {
      implicit thisTransaction =>
        val res = f1(args.length)
        println(if (thisTransaction.isAborted) "aborted" else s"result: $res")
    }
  }
```

Is there a better way to implement this coercion from implicit function to function?

odersky:
@nicolasstucki Thanks for explaining! So we need to make implicit functions compatible with normal functions after all. This means we have the following choices:

 - add a specific conversion, as in this PR. This is conceptually heavy. Every new kind of conversion adds considerable complexity to the spec. 
 - drop implicit modifier on apply method of implicit function, but treat it as an implicit anyway. This is also irregular.
 - Make implicit functions subtypes of functions, but omit the check on implicit overrides. This is the status quo. To keep it bullet proof we have to keep implicit functions effectively final (which they are, as of now, because of the impossibility to define `apply`).

To come to a resolution here I think it's best to answer first the question whether implicit functions should be final or not. Making them non-final gives us a lot of new power. For instance:

     class Ref[T](var elem: T) extends implicit Dummy => T {
        def apply(implicit d: Dummy): T = elem
     }

     val r = new Ref(22)
     val y = r    // y has type Int, not Ref[Int]!

Super powerful. But do we want it?

nicolasstucki:
I do not have any use cases for this. We should still double check that with #2014 we can not override the apply of an implicit function with `def apply(d: Dummy)` (in the example above), this was perviously possible.

odersky:
> We should still double check that with #2014 we can not override the apply of an implicit function with def apply(d: Dummy) (in the example above), this was perviously possible.

I tried it in the REPL: Now it's no longer possible.

nicolasstucki:
Then this PR is irrelevant now.

PR #2011, opened 2017-02-20
HMap test case

OlivierBlanvillain:
@DarkDimius this is the test case you requested

### DIFF
--- a/tests/run/hmap.scala
+++ b/tests/run/hmap.scala
@@ -1,0 +1,117 @@
+trait Tuple
+trait ::[H, T <: Tuple] extends Tuple
+case class TupleCons[H, T <: Tuple](h: H, t: T) extends ::[H, T]
+final case object TNil extends Tuple
+
+// Type level natural numbers -------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


tests/run/hmap.scala
```diff
@@ -0,0 +1,117 @@
...
+  def main(args: Array[String]): Unit = {
+    type MapType1 =
+      HEntry["name",   String]  ::
+      HEntry["genre",  Boolean] ::
+      HEntry["moneyz", Int]     ::
+      HEntry["cat",    String]  ::
+      TNil.type
+
+    // Since
+    val map1: MapType1 =
```
DarkDimius: waait, you wrote the type yourself, dotty didn't infer it.
It's fine than)

DarkDimius: Do I get it right that if you don't provide expected type, compilation will fail?

OlivierBlanvillain: It does not compile, but for other reasons. I've updated the test to show that these types can be inferred.

PR #2012, opened 2017-02-20
Change 'is abstract; cannot be instantiated' to Message

ennru:
@felixmulder 
Next error to the case class scheme. Called its kind "Usage Error" - maybe the error kinds should be defined.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1127,4 +1127,12 @@ object messages {
           |""".stripMargin
   }
 
+  case class CantInstantiateAbstractClassOrTrait(cls: Symbol, isTrait: Boolean)(implicit ctx: Context)
+  extends Message(CantInstantiateAbstractClassOrTraitID) {
+    val kind = "Usage"
+    private val traitOrAbstract = if (isTrait) hl"a trait" else hl"abstract"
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1127,4 +1127,12 @@ object messages {
           |""".stripMargin
   }
 
+  case class CantInstantiateAbstractClassOrTrait(cls: Symbol, isTrait: Boolean)(implicit ctx: Context)
+  extends Message(CantInstantiateAbstractClassOrTraitID) {
+    val kind = "Usage"
+    private val traitOrAbstract = if (isTrait) hl"a trait" else hl"abstract"
+    val msg = hl"""${cls.name} is ${traitOrAbstract}; it cannot be instantiated"""
+    val explanation = ""
```
felixmulder: Perhaps an explanation here to say you can use:

```scala
new TraitName { }
```
To instantiate it?

felixmulder: But of course using the actual symbol name..

ennru: Yes, could be good. 
I thought of mentioning the `object TraitName extends TraitName` pattern which is quite common.

felixmulder:
Great! Thanks @ennru :tada:

Issue #2002, opened 2017-02-20
Overloading method with implicit methods.


nicolasstucki:
This is not overloading issue, but I have no better word for it. If one defines two methods with the same arguments but one with implicit and the other one without, the compiler will crash.

```scala
class A {
  def foo(i: Int): Int = i
  def foo(implicit i: Int): Int = i
}
```

```
java.lang.AssertionError: assertion failed: data race? overwriting symbol of type ((i: Int)Int)(A.this.foo),
long form = TermRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),A)),foo)/withSig(Signature(List(scala.Int),scala.Int)) of class class dotty.tools.dotc.core.Types$TermRefWithSignature,
last sym id = 3803, new sym id = 3804,
last owner = class A, new owner = class A,
period = frontend at run 2
	at scala.Predef$.assert(Predef.scala:165)
	at dotty.tools.dotc.core.Types$NamedType.checkSymAssign(Types.scala:1506)
	at dotty.tools.dotc.core.Types$NamedType.setDenot(Types.scala:1541)
	at dotty.tools.dotc.core.Types$NamedType.withDenot(Types.scala:1534)
	at dotty.tools.dotc.core.Types$TermRef$.withSigAndDenot(Types.scala:1942)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.termRefWithSig(SymDenotations.scala:1105)
	at dotty.tools.dotc.typer.TypeAssigner$$anonfun$assignType$5.apply(TypeAssigner.scala:498)
	at dotty.tools.dotc.typer.TypeAssigner$$anonfun$assignType$5.apply(TypeAssigner.scala:498)
	at dotty.tools.dotc.core.Types$Type.orElse(Types.scala:99)
	at dotty.tools.dotc.typer.TypeAssigner$class.assignType(TypeAssigner.scala:498)
	at dotty.tools.dotc.typer.Typer.assignType(Typer.scala:64)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedDefDef$1.apply(Typer.scala:1243)
	at dotty.tools.dotc.typer.Typer$$anonfun$typedDefDef$1.apply(Typer.scala:1214)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.typer.Typer.typedDefDef(Typer.scala:1214)
[...]

PR #2014, opened 2017-02-21
Fix #2000: Make implicit and non-implicit functions incomparable

### LINKED ISSUE #2000: Override method with implicit methods.

odersky:
Implicit and non-implicit functions are incomparable with <:<, but are
treated as equivalent with `matches`. This means implicit and non-implicit
functions of the same types override each other, but RefChecks will
give an error because their types are not subtypes.

Also contains a test for #2002.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -489,7 +489,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         case tp1 @ MethodType(_, formals1) =>
           (tp1.signature consistentParams tp2.signature) &&
             matchingParams(formals1, formals2, tp1.isJava, tp2.isJava) &&
-            (!tp1.isImplicit || tp2.isImplicit) &&  // non-implicit functions shadow implicit ones
+            (tp1.isImplicit == tp2.isImplicit) &&
             isSubType(tp1.resultType, tp2.resultType.subst(tp2, tp1))
         case _ =>
           false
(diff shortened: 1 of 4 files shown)


nicolasstucki:
LGTM

odersky:
The test failures are that i2002.scala still runs with -Yno-double-defs even though it should not, and locally it does not. Something seems to be wrong with the argument passing scheme under CI.

PR #2019, opened 2017-02-22
Fix #2009: Fix placeholder params logic for lambdas

### LINKED ISSUE #2009: Eta expansion difference to scalac
Scalac is happy to compile the code below:

```scala
object Test {

  trait Gen[T] {
    def map[U](f: T => U): Gen[U] = ???
  }

  def f[T](implicit g: Gen[T]): Gen[() => T] =
    g map ( () => _ )
}
```

whereas dotty fails with:

```scala
-- Error: local/func0.scala ----------------------------------------------------
8 |    g map ( () => _ )
  |                  ^
  |    missing parameter type for parameter _$1, expected = Test.Gen[() => T]

```

odersky:
Logic was missing placeholders in rhs of lambdas.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1001,7 +1001,7 @@ object Parsers {
 
         val t = expr1(location)
         if (in.token == ARROW) {
-          placeholderParams = Nil // don't interprete `_' to the left of `=>` as a placeholder
+          placeholderParams = Nil // don't interpret `_' to the left of `=>` as placeholder
           wrapPlaceholders(closureRest(start, location, convertToParams(t)))
         }
         else if (isWildcard(t)) {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1001,7 +1001,7 @@ object Parsers {
 
         val t = expr1(location)
         if (in.token == ARROW) {
-          placeholderParams = Nil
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -993,20 +993,22 @@ object Parsers {
...
+        def wrapPlaceholders(t: Tree) = try
+          if (placeholderParams.isEmpty) t
+          else new WildcardFunction(placeholderParams.reverse, t)
+        finally placeholderParams = saved
+
         val t = expr1(location)
         if (in.token == ARROW) {
-          placeholderParams = saved
-          closureRest(start, location, convertToParams(t))
+          placeholderParams = Nil
```
smarter: I don't understand why it's okay to forget the value of `placeholderParams` here, could you maybe add a comment clarifying this?

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -993,20 +993,22 @@ object Parsers {
...
+        def wrapPlaceholders(t: Tree) = try
+          if (placeholderParams.isEmpty) t
+          else new WildcardFunction(placeholderParams.reverse, t)
+        finally placeholderParams = saved
+
         val t = expr1(location)
         if (in.token == ARROW) {
-          placeholderParams = saved
-          closureRest(start, location, convertToParams(t))
+          placeholderParams = Nil // don't interprete `_' to the left of `=>` as a placeholder
```
smarter: typo: interprete -> interpret

PR #2022, opened 2017-02-22
Fix warnings in sbt build

smarter:


felixmulder:
@smarter - seems like the partest command doesn't understand the parameters after these changes. Might need to make it into a key like the `repl` command?

project/Build.scala
```diff
@@ -276,22 +276,20 @@ object DottyBuild extends Build {
...
-              .map(_.data.getAbsolutePath)
-              .find(_.contains("scala-library"))
-              .toList.mkString(":")
-          }.value
+        val java: String = Process("which" :: "java" :: Nil).!!
+        val attList = (dependencyClasspath in Runtime).value
+        val scalaLib = attList
+          .map(_.data.getAbsolutePath)
+          .find(_.contains("scala-library"))
+          .toList.mkString(":")
```
sjrd: You are dropping the dependency of this task on `packageAll` here. I guess this was not intentional. You should add
```scala
val _ = packageAll.value
```

project/Build.scala
```diff
@@ -327,7 +325,7 @@ object DottyBuild extends Build {
         partestLockFile.createNewFile
         partestLockFile.deleteOnExit
       },
-      runPartestRunner <<= Def.inputTaskDyn {
+      runPartestRunner := Def.inputTaskDyn {
```
sjrd: If you change this, you also need to add `.value` after the closing `}` corresponding to the `{` on this line. This could explain the problem that you observe.

felixmulder:
haha @smarter - you beat me to it 😆 thanks @sjrd!

felixmulder:
now we just need to migrate to `build.sbt`...

smarter:
Thanks @sjrd, @felixmulder , should be good know.

PR #2026, opened 2017-02-23
Better error messages for missing type of recursive definitions

### LINKED ISSUE #2001: Unclear error message with mutually recursive methods with inferred return types
```scala
class A {
  def odd(x: Int) = if (x == 0) false else !even(x-1)
  def even(x: Int) = if (x == 0) true else !odd(x-1)
}
```
dotty:
```scala
3 |  def even(x: Int) = if (x == 0) true else !odd(x-1)
  |                                            ^^^
  |                                     cyclic reference involving method odd
```
Contrast with scalac:
```scala
mutrec.scala:3: error: recursive method odd needs result type
  def even(x: Int) = if (x == 0) true else !odd(x-1)
                                            ^
```

odersky:
Fixes #2001. 

@ennru Do you want to review this one?

Btw, thanks for all the good work on error messages!

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
@@ -40,12 +40,12 @@ object ErrorReporting {
         }
       } else msg
 
-      if (cycleSym.is(Implicit, butNot = Method) && cycleSym.owner.isTerm)
-        em"""cyclic reference involving implicit $cycleSym
+    if (cycleSym.is(Implicit, butNot = Method) && cycleSym.owner.isTerm)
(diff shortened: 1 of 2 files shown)


odersky:
Fixed the indentation. Regarding @tailrec, I think it's better to leave it out, even though the method happens to be tail recursive. But it is neither performance sensitive nor likely to traverse very deeply, so there's no need to require tail recursion here.

ennru:
@odersky I'm converting the error messages for recursive things to the case class scheme and want to add a unit test for "cyclic reference involving implicit". Can you help me to come up with a smallish example?

odersky:
Hi Enno, there's one in neg/implicitDefs.

PR #2028, opened 2017-02-23
Change 'private, protected, or this expected' to Message

ennru:
@felixmulder 
Another error message with explanation. It captures the class name Symbol, but does not use it for now. 
I tried to get the error position indicator to move to the position where the keyword is missing, but did not manage to do so. Ideas?

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1607,6 +1607,7 @@ object Parsers {
      *  LocalModifier  ::= abstract | final | sealed | implicit | lazy
      */
     def modifiers(allowed: BitSet = modifierTokens, start: Modifiers = Modifiers()): Modifiers = {
+      @tailrec
       def loop(mods: Modifiers): Modifiers = {
         if (allowed contains in.token) {
           val isAccessMod = accessModifierTokens contains in.token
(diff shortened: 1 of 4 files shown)


felixmulder:
Does `pos.focus` help with that?

ennru:
No. My problem is that the parser has passed the constructor parameters:
`-- [E043] Syntax Error: tests/pos/HelloWorld.scala -----------------------------`
`5 |class Annotated @deprecated (s: List[String])`
`..|.........,.........,.........,.........,....^`
`..|    private, protected, or this expected for annotated primary constructor`

I did not find a good spot to keep `in.offset`. If captured before parsing the annotations it ends up like this:
`-- [E043] Syntax Error: tests/pos/HelloWorld.scala -----------------------------`
`5 |class Annotated @deprecated (s: List[String])`
`..|.........,......^^^^^^^^^^`
`..|....private, protected, or this expected for annotated primary constructor`

felixmulder:
So a couple of alternatives without digging too deep:

1. Take `in.lastOffset` from `Parsers.scala:2058`
2. Take positions from some other part of the constructor in `Parsers.scala:2058-2068`
3. Take positions from `Modifiers` they are positioned, so you could look at that and then add the size of the name - might be a bit dirty...

ennru:
I think it is best to settle on the last annotation's position, which includes the parameter list:
`4 |class Annotaded @inline @deprecated (s: String, i: Int)`
`..|........................^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^`

felixmulder:
LGTM, thanks @ennru!

PR #2031, opened 2017-02-25
Fix #2030: Don't chain implicit conversions

### LINKED ISSUE #2030: Type mismatch for union types is slow
To reproduce, running from commit db24246e90f4b8afcaf1fc66df99d38d9a2cb736

```scala
$ cat target/union.scala
object a {
   val x: String | Int = 'a
}
$ time ./bin/dotc target/union.scala
-- [E007] Type Mismatch Error: target/union.scala ------------------------------
2 |   val x: String | Int = 'a
  |                         ^^
  |                         found:    Symbol
  |                         required: String | Int
  |

one error found
./bin/dotc target/union.scala  30.37s user 1.44s system 262% cpu 12.094 total
```

I ran jprofiler on a repl session to see what was happening and got these results 
[union-typeerror.zip](https://github.com/lampepfl/dotty/files/799287/union-typeerror.zip)
 I don't know much about Dotty's typer to make sense of the report.

My understanding from @DarkDimius's comment on Gitter is that union types put a strain on the type-checker for several fundamental reasons. However, 12s to report a type-error on a minimal example sounds like a bug to me.

odersky:
When inferring a view, we are not allowed to use another
implicit conversion to adapt its result. Fixing this revealed
another problem where we have to re-enable implicit conversions
when pre-typing arguments in overloading resolution.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -1375,7 +1375,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
                 else WildcardType)
               val commonFormal = defn.FunctionOf(commonParamTypes, WildcardType)
               overload.println(i"pretype arg $arg with expected type $commonFormal")
-              pt.typedArg(arg, commonFormal)
+              pt.typedArg(arg, commonFormal)(ctx.addMode(Mode.ImplicitsEnabled))
             }
           }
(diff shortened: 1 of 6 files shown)


SethTisue:
in Dotty 0.16.0-RC3, I'm able to chain implicit `Conversion`s 😈 https://gist.github.com/SethTisue/5ab846d1639af0e75e86ca19002689dc

PR #2036, opened 2017-02-27
Remove wrapArrayMethodName from defn.

nicolasstucki:
Note that it is never used and the same method exists in transform.TreeGen


odersky:
LGTM, thanks!

Issue #2037, opened 2017-02-27
Stack overflow on `member` of some refined types.


nicolasstucki:
The following type will generate a stack overflow when we call `member` on it.
```scala
   val badType =
      RefinedType(
        ctx.requiredClassRef("scala.collection.AbstractIterator"),
        "GroupedIterator".toTypeName,
        TypeRef(ctx.requiredClassRef("scala.collection.AbstractIterator"), "GroupedIterator".toTypeName))

    badType.member("GroupedIterator".toTypeName)
```

This a blocker for #1840 and only started happening after rebasing on the new bootstrap. 

I have set up a simple test in [this branch](https://github.com/dotty-staging/dotty/tree/bug-with-refined-type-member) to reproduce the bug with `sbt "testOnly dotc.tests"`.

Looking at the stack it is possible to see that it loops over `glb` making the types larger each time.

nicolasstucki:
@odersky, could you have a look at this one, it is a minimized version of the blocker for the call graph that we talked about last week.

nicolasstucki:
I also noticed that this only happens on inner classes that have a type parameter.

odersky:
@nicolasstucki Can you provide more context? I tried to reconstruct your description in this test case:

    import dotty.tools.dotc._
    import core._
    import Types._
    import Contexts._
    import Symbols._
    import Decorators._

    object Test {
      def f(implicit ctx: Context) = {
        val badType =
          RefinedType(
            ctx.requiredClassRef("scala.collection.AbstractIterator"),
            "GroupedIterator".toTypeName,
            TypeRef(ctx.requiredClassRef("scala.collection.AbstractIterator"), "GroupedIterator".toTypeName))
        badType.member("bar".toTermName)
      }
      def main(args: Array[String]): Unit = {
        implicit val ctx = (new ContextBase).initialCtx
        ctx.base.initialize()
        f
      }
    }

But that one runs OK, no StackOverflow. It would be good to have a self-contained test case that exhibits the problem.

nicolasstucki:
Try to run the tests in the branch that is in the link above.

odersky:
No, I need more control over the compiler. So it would be great to have a self-contained test.

nicolasstucki:
I will try to make it self contained.

odersky:
I found it. You get the stackoverflow if you do a member with the type name "GroupedIterator".

odersky:
[...]

PR #2039, opened 2017-02-27
Add assert that refine infos are legal wrt refined names

odersky:
Explains root cause of #2037.


nicolasstucki:
I would suggest to remove the test that can't be executed in partest and merging it.

nicolasstucki:
LGTM

Issue #1221, opened 2016-04-16
Consider implementing per-tailcall-position tailrec annotation


DarkDimius:
Consider a method of form 

```scala

final def foo(a: Int): Int = {
  if (foo(a - 1) > 0) foo(a - 1) 
  else foo(a - 2)
}
```

During discussion with @sirthias I understood that it's now commonly known `tailrec` would actually transform such a method, though it cannot be annotated with `@tailrec` as it contains recursive calls in non-tail position.

It would be great if it was possible to annotate specific calls, eg: 

```scala

final def foo(a: Int): Int = {
  if (foo(a - 1) > 0) 
    foo(a - 1): @tailrec 
  else 
    foo(a - 2): @tailrec
}
```

DarkDimius:
Implemented in #1227.

PR #2043, opened 2017-03-01
Tailrec for derivesFrom/lookupRefined/classSymbol/classSymbols

nicolasstucki:
Fist two commits make changes to loops, the third only adds `@tailrec` annotations.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -105,7 +105,7 @@ object Types {
     final def isValueTypeOrLambda: Boolean = isValueType || this.isInstanceOf[PolyType]
 
     /** Does this type denote a stable reference (i.e. singleton type)? */
-    final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
+    @tailrec final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
       case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable
       case _: SingletonType | NoPrefix => true
       case tp: RefinedOrRecType => tp.parent.isStable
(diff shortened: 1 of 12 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -996,15 +996,15 @@ object Types {
...
+      @tailrec def loop(pre: Type): Type = pre.stripTypeVar match {
         case pre: RefinedType =>
           pre.refinedInfo match {
             case TypeAlias(alias) =>
               if (pre.refinedName ne name) loop(pre.parent) else alias
             case _ => loop(pre.parent)
           }
         case pre: RecType =>
-          val candidate = loop(pre.parent)
+          val candidate = pre.parent.lookupRefined(name)
```
DarkDimius: this change is not necessary. It will still get tail-rec optimized even without it.
You can use https://github.com/lampepfl/dotty/issues/1221 to mark per-callsite `@tailrec`-s

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -290,50 +290,56 @@ object Types {
      *  instance, or NoSymbol if none exists (either because this type is not a
      *  value type, or because superclasses are ambiguous).
      */
-    final def classSymbol(implicit ctx: Context): Symbol = this match {
-      case ConstantType(constant) =>
-        constant.tpe.classSymbol
```
DarkDimius: wasn't this already `tailrec` optimized?
Instead of rewriting by hand I suggest using https://github.com/lampepfl/dotty/issues/1221

compiler/src/dotty/tools/dotc/core/Types.scala
nicolasstucki: Thanks, I used it in several places. In this instance it was easier to use the old notation.

DarkDimius:
LGTM

PR #2044, opened 2017-03-01
Allow abstract type members in objects.

nicolasstucki:


DarkDimius:
why?

nicolasstucki:
Because `scalac` allows it. 

Furthermore it is already possible to have abstract types in objects by defining it in a trait
```scala
trait A {
 type T <: Any
}

object B extends A // B has the abstract type T
```

In this case it would be simpler to define it as 

```scala
object B {
  type T  <: Any
}
```

odersky:
Do we have a test that checks that bounds of abstract type members are non-empty, i.e. lowre bound <: upper bound? In that case, LGTM

tests/neg/i1050c.scala
```diff
@@ -27,7 +27,8 @@ object Tiark4 {
     v.brand("boom!")
 }
 object V { // error: cannot be instantiated
```
nicolasstucki: The error of the bound `>: Any <: Nothing` is the reason for the `cannot be instantiated` error, the full message is quite clear on the source of the issue.

sjrd:
See here for a concrete use case: https://github.com/sjrd/scala-unboxed-option/blob/4222ec1d95982ac7a5912c1d06eacb8f7b09554f/src/main/scala/uoption/package.scala#L23
Also in this case it's even in a `package object`

PR #2045, opened 2017-03-02
Fix type inference for HLists and HMaps

odersky:
Two main fixes and some cleanups:

 - Systematically support dependent implicit methods (the test uses them)
 - Make &-simplification not forget constraints


retronym:
You might like to include some tests from SI-5070.

Scalac is too pessimistic it determining if a method is dependent, dotty's implementation looks better.

```scala
// should compile
trait Web {
  type LocalName
}
trait Companion1[A]
trait WebDSL[W <: Web] {
  trait LocalNameCompanion extends Companion1[W#LocalName] {
    type A = String
  }
  implicit val LocalName: LocalNameCompanion
}
object Test {
  def t[W <: Web](implicit webDSL: WebDSL[W]): Unit = {
    import webDSL._
    implicitly[LocalNameCompanion] // succeeds
    implicitly[Companion1[W#LocalName]] // fails
  }
}
```

I see from trying to compile the first example that this is not allowed in dotty for other reasons: "W#LocalName: W is not a legal path".

```scala
trait A {
  type T
}
 
object O {
  implicit def b(implicit x: A): x.T = error("")
}
 
class Test {
  import O._
  implicit val a: A = new A {}
  implicitly[a.T]
  implicitly[a.T](b(a))
}
```

```scala
// variation that typechecks in dotty but fails in scalac. Due to scalac's approximation of `isImmediatelyDependent`
class Test {
  def foo(implicit x: A) = {
    object O {
      implicit def b(implicit y: x.type): y.T = ???
    }

    import O._
    b(x): x.T
    implicitly[x.T]
  }
}
trait A {
  type T
}
```

FYI Here's an attempted fix for SI-5070 in scalac https://github.com/scala/scala/compare/2.12.x...retronym:ticket/5070

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -1312,23 +1312,28 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
         case tp2: RefinedType if tp1.refinedName == tp2.refinedName =>
-          // Given two refinements `T1 { X = S1 }` and `T2 { X = S2 }`, if `S1 =:= S2`
-          // (possibly by instantiating type parameters), rewrite to `T1 & T2 { X = S1 }`.
-          // Otherwise rewrite to `T1 & T2 { X B }` where `B` is the conjunction of
-          // the bounds of `X` in `T1` and `T2`.
-          // The first rule above is contentious because it cuts the constraint set.
-          // But without it we would replace the two aliases by
-          // `T { X >: S1 | S2 <: S1 & S2 }`, which looks weird and is probably
-          // not what's intended.
+          // Given two refinements `T1 { X = S1 }` and `T2 { X = S2 }` rwrite to
```
senia-psm: Typo: "rwrite"

odersky:
@retronym I added the original SI-5070 as well as an adapted version of SI-5643 as new tests.

tests/pending/pos/depmet_implicit_oopsla_session_2.scala
```diff
@@ -18,7 +17,7 @@ object Sessions {
 
   // friendly interface to the theory
   def runSession[S, D: Session[S]#HasDual](session: S, dual: D) =
-    ?[Session[S]#HasDual[D]].run(session, dual)
+    implicitly[Session[S]#HasDual[D]].run(session, dual)
```
OlivierBlanvillain: Was keeping this test pending intentional? Does this PR change how `implicitly` deals with type member?

I remember seeing something similar to this `def ?` being use to get more precise types:

```scala
trait A { type Out }

implicit val a = new A { type Out = String }

val i1: A { type Out = String } = implicitly[A]
// error: type mismatch;
// found   : A
// required: A{type Out = String}

def ?[T <: AnyRef](implicit w: T): w.type = w

val i2: A { type Out = String } = ?[A]  // works fine!
```

compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
```diff
@@ -64,14 +71,29 @@ trait ConstraintHandling {
...
+   *  that `param >: bound`.
+   */
+  def narrowedBound(param: PolyParam, bound: Type, isUpper: Boolean)(implicit ctx: Context): TypeBounds = {
+    val oldBounds @ TypeBounds(lo, hi) = constraint.nonParamBounds(param)
+    val saved = homogenizeArgs
+    homogenizeArgs = Config.alignArgsInAnd
+    try
+      if (isUpper) oldBounds.derivedTypeBounds(lo, hi & bound)
+      else oldBounds.derivedTypeBounds(lo | bound, hi)
+    finally homogenizeArgs = saved
```
odersky: I think this would make the logic more complicated.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -1312,23 +1312,28 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
-              rinfo1 & rinfo2
-          tp1.derivedRefinedType(parent, tp1.refinedName, rinfo)
+
+          def isNonvariantAlias(tp: Type) = tp match {
+            case tp: TypeAlias => tp.variance == 0
+            case _ => false
+          }
+          if (homogenizeArgs &&
+              isNonvariantAlias(rinfo1) && isNonvariantAlias(rinfo2))
+            isSameType(rinfo1, rinfo2)
```
odersky: No, we perform sameType only for its effect on the constraint

tests/pending/pos/depmet_implicit_oopsla_session_2.scala
odersky: No, that's an oversight. We should keep the test as it was in pending. It worked after the first commits but then stopped working again after the final commits. Given the error message it could well be something related to existential types which dotty no longer supports, so I did not investigate further. I'll change the test to what it was an add a comm
ent.

PR #2046, opened 2017-03-02
Coherent type classes

odersky:
A simple implementation of a Coherent trait which avoids
ambiguity errors for goals that derive from it.

It implements #2047.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -516,6 +516,8 @@ class Definitions {
     def Product_productArity(implicit ctx: Context) = Product_productArityR.symbol
     lazy val Product_productPrefixR = ProductClass.requiredMethodRef(nme.productPrefix)
     def Product_productPrefix(implicit ctx: Context) = Product_productPrefixR.symbol
+  lazy val CoherentType: TypeRef                = ctx.requiredClassRef("scala.typeclass.Coherent")
+  def CoherentClass(implicit ctx: Context) = CoherentType.symbol.asClass
   lazy val LanguageModuleRef = ctx.requiredModule("scala.language")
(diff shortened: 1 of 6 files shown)


retronym:
A variation on this theme might be to allow ambiguity if all the candidates extends `Coherent`, even if the implicit search itself doesn't. That would allow code requiring typeclasses to be more general (it doesn't have to limit itself to implicitly accepting coherent instances).

I can also imagine it might be necessary from time to time to cast an incoherent instance to a coherent one. Having `Coherent` as a marker interface might make this impossible. With some boilerplate you could achieve the same through a delegating wrapper instance, I guess. Or we could find a way to make the coherent marker type completely erase away to allow casting.

odersky:
> A variation on this theme might be to allow ambiguity if all the candidates extends Coherent, even if the implicit search itself doesn't. That would allow code requiring typeclasses to be more general (it doesn't have to limit itself to implicitly accepting coherent instances).

I think that's closer in spirit to @djspiewak's proposal https://gist.github.com/djspiewak/9f6feadab02b16829c41484b394d16e4. But applied to the present proposal #2047 it would not work, because the alternatives might come from different coherence domains. I.e. the fact that they all implement a coherent type does not guarantee that they agree on the implementation of the expected type.

odersky:
@retronym Regarding casting: It's of course always good to have an escape hatch. But in the case of implicit resolution we already have one: Just pass parameters explicitly. So, do you think we need more than that?

I like the formulation of `Coherent` as a marker trait because that neatly captures the rule that all subtypes of a coherent type must themselves be coherent.

puffnfresh:
I like it.

Could this be an annotation rather than a trait? This would be consistent with things like `scala.annotation.implicitNotFound` and `scala.annoation.implicitAmbiguous`.

Can we call it something like "implicitIgnoreAmbiguous" instead of Coherent? Coherency implies a lot more than what this functionality provides.

odersky:
I am closing this for now, since we agreed we wanted to get coherency checking in first.

PR #2049, opened 2017-03-04
Change "recursive/cyclic definitions needs type" errors to Message

ennru:
@odersky
Here are my changes to move the recursive/cyclic definition type required errors to Message.
I hope the explanations make sense.


compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -52,6 +52,9 @@
     MixedLeftAndRightAssociativeOpsID,
     CantInstantiateAbstractClassOrTraitID,
     AnnotatedPrimaryConstructorRequiresModifierOrThisID,
+    OverloadedOrRecursiveMethodNeedsResultTypeID,
+    RecursiveValueNeedsResultTypeID,
+    CyclicReferenceInvolvingImplicitID,
```
odersky: I have a more general question: Why are we using a Java enumeration for this? Would it not be easier to use a `scala.Enumeration`?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1160,4 +1160,34 @@ object messages {
...
   }
+
+  case class OverloadedOrRecursiveMethodNeedsResultType(tree: Names.TermName)(implicit ctx: Context)
+  extends Message(OverloadedOrRecursiveMethodNeedsResultTypeID) {
+    val kind = "Syntax"
+    val msg = hl"""overloaded or recursive method ${tree} needs result type"""
+    val explanation =
+      hl"""|${tree} is overloaded and at least one definition of it calls another.
+           |You need to specify the calling method's return type.
+         """.stripMargin
```
odersky: In this case the method *might* be overloaded but it could also be simply recursive. So claiming that it is overloaded is false. We might try to find out whether it is overloaded to improve the error message. Essentially, it's overloaded if its owner is a class and taking the member of the name in the owner gives several alternatives.

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
ennru: @felixmulder recently changed from directly specified int constants to this Java enum.

felixmulder:
And I in turn blame @NicolasStucki ;)

In all seriousness, we wanted something lightweight (read: contributor
friendly) that could then be forgotten as soon as we're bootstrapped good
and proper (hello new scala enums!). I agreed with Nicolas to change it
since he was having trouble rebasing the call-graph when new errors were
implemented. It was a mutual decision.

ennru:
@odersky 
I made it two distinct errors "Overloaded method needs result type" and "Recursive method needs result type". But I did not find out how to dig into the owning class' members.

The "cyclic reference involving" from the `CyclicReference` type error is now a `Message`.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -28,11 +28,14 @@ object ErrorReporting {
 
   def cyclicErrorMsg(ex: CyclicReference)(implicit ctx: Context) = {
     val cycleSym = ex.denot.symbol
-    def errorMsg(msg: String, cx: Context): Message =
+    def errorMsg(msg: Message, cx: Context): Message =
       if (cx.mode is Mode.InferringReturnType) {
         cx.tree match {
           case tree: untpd.DefDef if !tree.tpt.typeOpt.exists =>
-            OverloadedOrRecursiveMethodNeedsResultType(tree.name)
+            // TODO: analysis if tree is an overloaded method (or directly recursive)
```
odersky: I think this will do the trick:

    cycleSym.owner.isClass && cycleSym.owner.member(cycleSym.name).isOverloaded

ennru: In that direction was what I tried, but `member` fails in `Tree.withType` as it encounters an `ErrorType`.

odersky: Why do you need `Tree.withtype`?

ennru: That is what I get upon calling `member(cycleSym.name)` - not sure how the code gets there.

[error] Test dotty.tools.dotc.reporting.ErrorMessagesTests.recursiveMethodNeedsReturnType failed: java.lang.AssertionError: assertion failed, took 0.01 sec
[error]     at scala.Predef$.assert(Predef.scala:151)
[error]     at dotty.tools.dotc.ast.Trees$Tree.withType(Trees.scala:116)
[error]     at dotty.tools.dotc.typer.ErrorReporting$.errorTree(ErrorReporting.scala:22)
[error]     at dotty.tools.dotc.typer.Typer$$anonfun$typed$2.apply(Typer.scala:1584)
[error]     at dotty.tools.dotc.typer.Typer$$anonfun$typed$2.apply(Typer.scala:1582)
[error]     at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:136)
[error]     at dotty.tools.dotc.core.Contexts$Context.traceIndented(Contexts.scala:57)
[error]     at dotty.tools.dotc.typer.Typer.typed(Typer.scala:1582)
...

odersky: It complains that we assign an error type into a tree, yet not errors were reported. Is that stacktrace reached from the call
 
    cycleSym.owner.info.member(cycleSym.name).isOverloaded

? If yes, can you show me the full stacktrace up to the call?

ennru: Here you go: [stacktrace.txt](https://github.com/lampepfl/dotty/files/824488/stack.txt)

odersky: According to that stacktrace the error is _not_ caused by the call 

    cycleSym.owner.info.member(cycleSym.name).isOverloaded

That one shows up nowhere. So it most likely happens later. Could it be that an error message is not reported?

ennru: The error is triggered by using `member`, even though it is not on the stack. The error is reported correctly if I don't call it and it fails when I try to, at the time of calling the error is not reported yet.

odersky: I think the `member` call causes another CyclicReference exception. That's not easy to avoid. So I think we should give up. Don't try to distinguish the error message and just say "recursive or overloaded".

odersky:
Yes, that's indeed harder than expected. Thanks for hanging in there!

PR #2050, opened 2017-03-04
Drop named type parameters in classes

odersky:
Drop the [type T] syntax, and what's associated to make it work.

Motivation: It's an alternative way of doing things for which there seems
to be little need. The implementation was provisional and bitrotted during
the various iterations to introduce higher-kinded types. So in the end the
complxity-cost for language and compiler was not worth the added benefit
that [type T] parameters provide.

Note that we still accept _named arguments_ [A = T] in expressions; these are useful
for specifying some parameters and letting others be inferred.

Review by @smarter

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -805,7 +805,7 @@ object Parsers {
     private def simpleTypeRest(t: Tree): Tree = in.token match {
       case HASH => simpleTypeRest(typeProjection(t))
       case LBRACKET => simpleTypeRest(atPos(startOffset(t)) {
-        AppliedTypeTree(t, typeArgs(namedOK = true, wildOK = true)) })
+        AppliedTypeTree(t, typeArgs(namedOK = false, wildOK = true)) })
       case _ => t
     }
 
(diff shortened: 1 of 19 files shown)


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1680,25 +1680,17 @@ object Parsers {
...
-        var mods = annotsAsMods()
-        if (ownerKind == ParamOwner.Class) {
-          mods = modifiers(start = mods)
-          mods =
-            atPos(start, in.offset) {
-              if (in.token == TYPE) {
-                val mod = atPos(in.skipToken()) { Mod.Type() }
-                (mods | Param | ParamAccessor).withAddedMod(mod)
-              } else {
-                if (mods.hasFlags) syntaxError(TypeParamsTypeExpected(mods, ident()))
```
smarter: The message `TypeParamsTypeExpected` is now unused and could be removed.

smarter:
https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/dotc/parsing/ModifiersParsingTest.scala#L99 needs to be updated to remove the use of named types.

Issue #1945, opened 2017-02-06
Dottydoc needs to be reactive


felixmulder:
The doc tool currently pretty much only supports desktop, since there are no rules for mobile devices.

Since the documentation is using bootstrap 4, we should adapt to using its preferred method of handling reactive sizing.

hhandoko:
I might be able to help 😄

felixmulder:
Let me know if you need help getting started.

hhandoko:
Yup, actually I do 😁  How do I (re)generate the existing Dottydocs to preview the changes?

I've tried to running `dottydoc` on the root project but it's throwing up some `[doc error]` messages (not being able to find the HTML pages). Is there a specific sub-project / module I need to switch to?

So here's what I've accomplished so far:

1. I've cloned the repo, compiled and managed to get to the REPL successfully
1. I've located the relevant files under `doc-tool/resources`

felixmulder:
Run `genDocs` and you'll get a `_site` directory in the docs dir :)

julienrf:
@felixmulder I think you mean “responsive” rather than “reactive” ;)

felixmulder:
Yes random ahoy buzzword, I mean responsive

hhandoko:
@felixmulder I've submitted the first PR for the responsive feature [1]. I'll be looking into the landing page next. Are there specific areas of the site / layout you want me to look at in particular?

Notes:
[1] - I prefer a few smaller PRs rather than a single large one as it is easier to manage. Hope that's alright.

felixmulder:
So the important part for getting dottydoc responsive is not the landing page - that page is going to be different for every project. The import part is getting the pages under: http://dotty.epfl.ch/docs/ to be responsive :)

hhandoko:
Cool, got it 👌

hhandoko:
@felixmulder any particular area you want me to look next? Otherwise I'm going to update the docs (`api-page.html`) template styling, e.g. clean up extra padding, improve contrast, font sizes, etc.

PR #2052, opened 2017-03-06
Feature - Responsive Dottydoc (Part 1) - Main layout

hhandoko:
Related ticket: #1945 

Added better responsive support to sidebar by implementing off-canvas behaviour (i.e. content body slides to the right instead of becoming narrower).

Tested on:
- Chrome 56.0.2924.87
- Firefox 47.0.1
- Safari 10.0.3


felixmulder:
As a side note - we're using the imperative mood for our commit messages. This means "Add X" instead of "Added X". Here's a quick primer:

> 1. Separate subject from body with a blank line
> 1. Limit the subject line to 50 characters
> 1. Capitalize the subject line
> 1. Do not end the subject line with a period
> 1. Use the imperative mood in the subject line
> 1. Wrap the body at 72 characters
> 1. Use the body to explain what and why vs. how
>
> taken from [Chris Beams](https://chris.beams.io/posts/git-commit/)

hhandoko:
Hi @felixmulder,

> The position of the hamburger icon has changed, we'd like to keep it where it was. Also in moving it, it looks like it lost its CSS styling.

By CSS styling do you mean `:hover` styles? I missed it but I can add it back again. I changed the positions since a fixed hamburger icon positioning overlaps with the content body text (it looks less intrusive in the sidebar area). Also, I am taking these cues from Elixir's ExDocs convention [1].

> The width of the sidebar now feels a bit too wide, what was your reasoning here?

I felt 250px was a bit too narrow as some package names got cut-off and most class names barely fit, so I added an extra 50px... But you have a good point, probably on narrower viewport width I can reduce it back to 250px.

> The animation from ☰ to ˿ while nice-looking, doesn't really add anything in its context, I assume it's for mobile devices / something thinner than 767px - maybe it'd look ok with the old styling restored? Let's leave it in for the next iteration of the PR.

I was looking to give more context with the left-pointing arrow, I've seen it implemented on the old Google Play Store mobile sidebar. Yes, the animation has no functional purpose, I just thought it looks nice 😁 

> In the <767px view the size of the sidebar exposed is huge, I'm not sure if that was intentional or if it looks better on mobile. But we should probably differentiate between rules for mobile devices and smaller viewports on desktops, just a thought

Yes, it is intentional. As per earlier comments I'll reduce it back to 250px before minimising the sidebar at 576px.

Let me know if you have any further thoughts.

Notes:
[1] - Example of Elixir's ExDocs output: https://hhandoko.github.io/diskusi/readme.html

felixmulder:
> By CSS styling do you mean `:hover` styles? 

I mean:

screen shot 2017-03-06 at 15 45 01
vs before:
screen shot 2017-03-06 at 15 45 08

> a fixed hamburger icon positioning overlaps with the content body text

Not sure what you mean by "overlaps" in this sentence here is what it looks like on rMBP all  up to date browsers:

screen shot 2017-03-06 at 15 53 34

> Also, I am taking these cues from Elixir's ExDocs

I see, let's try to do our own thing 🙂 

> I felt 250px was a bit too narrow as some package names got cut-off and most class names barely fit

You're right, but then again, 300px is not enough for some projects, so this will be arbitrary - if we keep it at 250px - people can override the CSS on their own docs if need be.

hhandoko:
> Not sure what you mean by "overlaps" in this sentence here is what it looks like on rMBP all up to date browsers:

WRT the screenshot, it looks OK when the content body fits entirely in one page, but it becomes an issue on lengthy pages.

- Without fixed hamburger menu positioning, the menu gets hidden when the user scrolls towards the bottom of the page. It is especially problematic on mobile devices with lengthy content body, as the user won't be able to minimise the sidebar again without scrolling to the top of the page.

- With fixed hamburger menu positioning (i.e. fixed position on top-left of content body), some left-padding must be added to the content body otherwise the menu will appear on top of the content body text / whitespace when the user scrolls towards the bottom of the page.

hamb-menu-overlap

> I see, let's try to do our own thing 🙂

Yes, I'm not saying Dottydoc should copy ExDocs, just that I found that their UI pattern is useful and solves an existing problem given that the primary layout is very similar (as mentioned above).

> You're right, but then again, 300px is not enough for some projects, so this will be arbitrary - if we keep it at 250px - people can override the CSS on their own docs if need be.

Good point 👌 

Actions:
[ ] - Add `:hover` styles to hamburger menu
[ ] - Reduce sidebar width back to 250px
[ ] - Switch to minimise sidebar at x < 576px from x < 768px

hhandoko:
@felixmulder, any thoughts? ☝️

hhandoko:
The last batch of commits completes the following:

[X] - Add `:hover` styles to hamburger menu
[X] - Reduce sidebar width back to 250px
[ ] - Switch to minimise sidebar at x < 576px from x < 768px

I've left sidebar minimisation / collapse at 768px as the layout looks really awkward between Mobile and Tablet resolution (sidebar and content body is split almost equally down the middle).

felixmulder:
Thanks for the hard work @hhandoko! :tada:

PR #2053, opened 2017-03-06
fixed PatternSyntaxException on Windows

niktrop:
`String.split(String)` expects pattern and doesn't work with `\` file separator on windows. 
`StringLike.split(Char)`  works for any character.


felixmulder:
LGTM! Thanks :)

PR #2057, opened 2017-03-07
support merging companion objects in expanded trees

liufengyun:
Support merging companion objects in expanded trees

Previous implementation is problematic when there are multiple transforms before typer:

1. There might be objects needing merging that only exist in the expanded trees, which cannot be handled by the previous algorithm.

2. There may be companion object and class defined only in the expanded trees, the previous algorithm doesn't create links for them.

This PR simplifies the companion object merging logic and fixes the problems above.

In fact, this PR supports multiple pre-typer transform during expansion. The protocol is that the expansion of a statement is either a flattened thicket or a non-thicket tree.


compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -484,46 +484,86 @@ class Namer { typer: Typer =>
...
-          case Some(cdef) =>
-            cdef.attachmentOrElse(ExpandedTree, cdef) match {
-              case Thicket(cls :: mval :: TypeDef(mname, compimpl: Template) :: crest)
-              if name.moduleClassName == mname =>
-                val mcls1 = cpy.TypeDef(mcls)(
-                    rhs = cpy.Template(impl)(body = compimpl.body ++ impl.body))
-                mdef.putAttachment(ExpandedTree, Thicket(vdef :: mcls1 :: rest))
-                moduleDef(typName) = mcls1
-                cdef.putAttachment(ExpandedTree, Thicket(cls :: crest))
+      def valid(mdef: MemberDef): Boolean = !mdef.mods.is(Package) && mdef.mods.is(Module)
```
odersky: Shorter and mode efficient:

    mdef.mods.is(Module, butNot = Package)

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -484,46 +484,86 @@ class Namer { typer: Typer =>
...
+
+      for (stat <- stats)
+        expanded(stat) match {
+          case Thicket(trees) => // companion object always expands to thickets
+            trees.map {
+              case mcls @ TypeDef(name, impl: Template) if valid(mcls) =>
+                if (moduleClsDef.contains(name)) {
```
odersky: I'd use a `get` instead of `contains` + `apply`:

    moduleClsDef.get(name) match {
      case Some((stat1, mcls1 @ TypeDef(_, impl1: Template)) => ...
      case None => ...

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -484,46 +484,86 @@ class Namer { typer: Typer =>
...
+                  val (stat1, mcls1 @ TypeDef(_, impl1: Template)) = moduleClsDef(name)
+                  if (mcls.mods.is(Synthetic) && !mcls1.mods.is(Synthetic)) { // merge synthetic into user-defined module
+                    removeInExpanded(stat, mcls)
```
odersky: Factor the two preceding  and the next two lines into a separate local method `mergeIfSynthetic(synthCls,  userCls`. Then the whole 12 line if-else can be expressed as

    mergeIfSynthetic(mcls, mcls1)
    mergeIfSynthetic(mcls1, mcls)

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -484,46 +484,86 @@ class Namer { typer: Typer =>
...
+                    val mcls2 = mergeModuleClass(stat1, mcls1, impl.body)
+                    moduleClsDef(name) = (stat1, mcls2)
+                  }
+                  else if (!mcls.mods.is(Synthetic) && mcls1.mods.is(Synthetic)) {
+                    removeInExpanded(stat1, mcls1)
+                    val mcls2 = mergeModuleClass(stat, mcls, impl1.body)
+                    moduleClsDef(name) = (stat, mcls2)
+                  }
+                  else {
+                    // redefinition of objects or case classes, handled elsewhere
```
odersky: You mean, we have a double definition? (Just asking for clarification)

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -484,46 +484,86 @@ class Namer { typer: Typer =>
...
+                  }
+                }
+                else moduleClsDef(name) = (stat, mcls)
+
+              case vdef @ ValDef(name, _, _) if valid(vdef) =>
+                if (moduleValDef.contains(name)) {
+                  val (stat1, vdef1) = moduleValDef(name)
+                  if (vdef.mods.is(Synthetic) && !vdef1.mods.is(Synthetic))  // merge synthetic into user-defined module
```
odersky: Similar remarks as for module defs apply here. Probably better to factor out in a method `mergeIfSynthetic`.

compiler/src/dotty/tools/dotc/typer/Namer.scala
liufengyun: Yes, I mean double definition. I'll refine the comment for clarity.

compiler/src/dotty/tools/dotc/typer/Namer.scala
liufengyun: The comments are addressed in the latest commit.

For the last one, I think it's probably good to keep it as it is, as the processing is not symmetric as the case above, and the code is simple and short. Refactoring into a method seems to make readability worse.

PR #2058, opened 2017-03-07
Fix #2054

### LINKED ISSUE #2054: Nested methods inside static methods doesn't work
The following code compiles fine, but it generates run-time exception.

```Scala
import scala.annotation.static

class Test

object Test {
  @static def test(n: Int): Int =  {
    List(3).map(_ + 2)
    n
  }

  def main(args: Array[String]): Unit = test(3)
}
```

Error logs:

```
Exception in thread "main" java.lang.BootstrapMethodError: java.lang.IllegalAccessError: tried to access method Test$.$anonfun$test$1(I)I from class Test
	at Test.test(static.scala:7)
	at Test$.main(static.scala:11)
	at Test.main(static.scala)
Caused by: java.lang.IllegalAccessError: tried to access method Test$.$anonfun$test$1(I)I from class Test
	at java.lang.invoke.MethodHandleNatives.resolve(Native Method)
	at java.lang.invoke.MemberName$Factory.resolve(MemberName.java:975)
	at java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:1000)
	at java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:1392)
	at java.lang.invoke.MethodHandles$Lookup.linkMethodHandleConstant(MethodHandles.java:1748)
	at java.lang.invoke.MethodHandleNatives.linkMethodHandleConstant(MethodHandleNatives.java:477)
	... 3 more
```

DarkDimius:


liufengyun:
@DarkDimius the  test `dotty.tools.dotc.InterfaceEntryPointTest.runCompilerFromInterface` failed.

compiler/src/dotty/tools/dotc/transform/MoveStatics.scala
```diff
@@ -19,6 +20,8 @@ class MoveStatics extends MiniPhaseTransform with SymTransformer { thisTransform
   import tpd._
   override def phaseName = "moveStatic"
 
+  // This phase moves methods around(in infortransform) so it may need to make other methods public
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[ExpandPrivate])
```
smarter: You moved `MoveStatic` _before_ `ExpandPrivate` but you added a requirement that it should be run `after`, so every test fails.

compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
```diff
@@ -38,6 +41,9 @@ class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { t
 
   override def phaseName: String = "expandPrivate"
 
+  // This phase moves methods around(in infortransform) so it may need to make other methods public
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[MoveStatics])
```
liufengyun: As pointed out by @smarter , there is a typo here: `infortransform`. Otherwise, LGTM.

PR #2062, opened 2017-03-07
Remove most uses of scala-reflect jar

smarter:
The remaining uses of scala-reflect (`scala.reflect.io.*`, `scala.reflect.internal.util.WeakHashSet`) cannot be removed until we integrate the backend in the dotty repo.


smarter:
Interestingly this revealed a bug in how we set `PureInterface` which is now fixed by the latest commit.

smarter:
I think there might be more issues in `defKind` but this is off-topic, I'll open an issue after this PR is merged.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -33,7 +33,11 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
   def defKind(tree: Tree): FlagSet = unsplice(tree) match {
     case EmptyTree | _: Import => NoInitsInterface
     case tree: TypeDef => if (tree.isClassDef) NoInits else NoInitsInterface
-    case tree: DefDef => if (tree.unforcedRhs == EmptyTree) NoInitsInterface else NoInits
+    case tree: DefDef =>
+      if (tree.unforcedRhs == EmptyTree)
+        (NoInitsInterface /: tree.vparamss.flatten)((fs, vparam) => fs & defKind(vparam))
```
odersky: I did not understand why the change. I assume it has to do with default arguments? But, first, if a parameter has a default value the enclosing trait it's still a NoInits, and second, won't we then see the default method which gives us the right DefKind anyway?

compiler/src/dotty/tools/dotc/util/Collections.scala
```diff
@@ -0,0 +1,39 @@
+/* NSC -- new Scala compiler
```
odersky: Does this file not duplicate standard collection functionality? Why keep it?

compiler/src/dotty/tools/dotc/util/StringOps.scala
```diff
@@ -0,0 +1,22 @@
+/* NSC -- new Scala compiler
+ * Copyright 2002-2013 LAMP/EPFL
+ * @author  Martin Odersky
+ */
+
+package dotty.tools
+package dotc
+package util
+
+/** This object provides utility methods to extract elements
```
odersky: I think this should go in StringDecorator in file Decorators.scala

compiler/src/dotty/tools/dotc/util/Collections.scala
smarter: I'm fine either way, apparently the original motivation for these methods is that they're faster than doing the equivalent using chains of methods from the standard library

compiler/src/dotty/tools/dotc/util/StringOps.scala
smarter: Sure.

compiler/src/dotty/tools/dotc/util/Collections.scala
odersky: I doubt it. If we do want to keep these methods, we should also move them to Decorators.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
smarter: Yes, it's for default arguments, the corresponding commit message gives more information: https://github.com/lampepfl/dotty/pull/2062/commits/618fe9645b064b2a28be1505fe52cb3b94f7a25b

>  if a parameter has a default value the enclosing trait is still a NoInits
Ah indeed, I missed that.


> won't we then see the default method which gives us the right DefKind anyway?
No, because `defKind` is called on each element of the body of the class before they are desugared.

It would be much simpler if `defKind` was called after desugaring (and would also avoid other issues with default parameters of class constructors for example) but I don't know what that would involve.

compiler/src/dotty/tools/dotc/util/Collections.scala
smarter: What would `map2` be decorated on? Anyway, I'll just remove those methods in favor of the standard library, it shouldn't make much of a difference.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
odersky: Right. I missed the context before. defKind is called before the trees are desugared. Yes, then  we need to do it as you describe except for the NoInits issue. Also since we traverse all methods of a trait here I'd avoid creating too much garbage via flatten. I'd just do a double forall:

    if (tree.unforcedRhs == EmptyTree) && tree.vparamss.forall(_.forall(_.rhs.isEmpty)) NoInitsInterface else NoInits

smarter:
@odersky Comments addressed, thanks! The existing `StringDecorators` was an instance of `PreName` so it felt wrong to add random methods to it, instead I renamed it to `PreNamedString` and added a new `StringDecorators` implicit class. I updated individual commits in this PR to have a cleaner history but that makes it harder to review the changes, you can see the diff from the previous set of commits at https://github.com/dotty-staging/dotty/commit/fa6f9882c6db4dde88ec0247a262c0032222d07b

smarter:
Interesting, adding the new `StringDecorator` broke the bootstrap in a weird way:
```scala
[error] -- [E008] Member Not Found Error: /home/smarter/opt/dotty/compiler/src/dotty/tools/dotc/core/Comments.scala 
[error] 405 |      defs(sym) ++= defines(raw).map {
[error]     |      ^^^^^^^^^^^^^
[error]     |value `++=` is not a member of scala.collection.immutable.Map[String, String] - did you mean `scala$collection$MapLike$$B.+`?
[error] -- [E008] Member Not Found Error: /home/smarter/opt/dotty/compiler/src/dotty/tools/dotc/typer/Docstrings.scala 
[error] 41 |          tplExp.defineVariables(sym)
[error]    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |value `++=` is not a member of scala.collection.immutable.Map[String, String] - did you mean `scala$collection$MapLike$$B.+`?
```
The errors go away if I unimport `StringDecorator`, I'll investigate.

smarter:
The issue was due to dotty trying to use the private `splitAt` in `StringDecorator`, I've fixed that by hiding `splitAt` better but I'll open an issue for that.

PR #2070, opened 2017-03-08
Fix bug in erasedLub leading to incorrect signatures

smarter:
Before this commit, the added testcase failed in a strange way:
```scala
14 |  def bla(foo: Foo) = orElse2(identity).apply(foo)
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |value of type <nonsensical><notype></nonsensical> does not take parameters
```
This happened because the TermRef for the apply method had an incorrect
signature (with a result type of `Object` instead of `Foo`), therefore its underlying type was NoType.

According to the documentation of `erasedLub`, the erasure should be:
> a common superclass or trait S of the argument classes, with the
    following two properties:
> - S is minimal: no other common superclass or trait derives from S]
> - S is last   : in the linearization of the first argument type `tp1`
                    there are no minimal common superclasses or traits that
                    come after S.
  (the reason to pick last is that we prefer classes over traits that way).

I'm not convinced that the implementation satisfies either of these two
properties, but this commit at least makes S closer to being minimal by
making sure that it does not derive from the last seen best candidate never derives from it.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -246,10 +246,14 @@ object TypeErasure {
           val cls2 = tp2.classSymbol
           def loop(bcs: List[ClassSymbol], bestSoFar: ClassSymbol): ClassSymbol = bcs match {
             case bc :: bcs1 =>
-              if (cls2.derivesFrom(bc))
-                if (!bc.is(Trait) && bc != defn.AnyClass) bc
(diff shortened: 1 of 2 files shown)


odersky:
The change LGTM. In the comment leading up to it, there's a stray `]'. Also, we should change `lub` to `erased lub` in that comment for clarity.

odersky:
I'll merge and piggyback the comment on some future PR.

smarter:
Ok, I think we should also take a closer look to make sure the comment
matches the implementation.

Issue #1950, opened 2017-02-06
Consolidate entity serialization into one implicit class


felixmulder:
The templating engine only works on `java.util.Map[String, AnyRef]` (buuh). This means that for each entity, we must convert it to this format. This is done in: [dottydoc.model.JavaConverters](https://github.com/lampepfl/dotty/blob/master/doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala).

This file is unnecessarily large, since it contains an implicit class for each subtype of `Entity`. But! It could be made to just contain one implicit class for all subtypes of `Entity`. Namely:

```scala
implicit class JavaEntity(val ent: Entity) extends AnyVal {
  def asJava: JMap[String, _] = ...
}
```

The `asJava` function would return a map with all possible fields of subtypes of `Entity` with `null` for value if the field does not exist in the current subtype.

Yes, we hate `null` in Scala land - but here we are on the border between Java and Scala and as such we have to make some concessions to please the great Java Templating engine gods.

**Added bonus points for:** getting rid of the `_root_` imports, no longer needed

wutongy:
Hi, @felixmulder. I would like to solve this issue. It would be a good chance for me to familiar with Scala code since I would like to apply for GSOC this year and contribute to Scala community.

wutongy:
Hi @felixmulder. 

Just want to confirm with you if what I try to do is right. Does the one implicit class you suggested look like that? :
```scala
implicit class JavaEntity(val ent: Entity) extends AnyVal {
  def asJava(extras: Map[String, _] = Map.empty): JMap[String, _] = ent match {
    case link: MaterializableLink => link match ...
    case ref: Reference =>  ref match ...
    ...
    case _ => null
  }
}
```

Also, may I ask how to test the code so that I can know my modification is correct and does not break anything by accident? Thanks!

felixmulder:
This is what I'm suggesting, and the easiest way to test it would be to create an anonymous instance of each entity type. E.g:

```scala
val df = new Def { ... }
val vl = new Val { ... }

// all of the testsssss
```

A test like this would be unable to compile if the interface changed.

PR #2074, opened 2017-03-10
[WIP] consolidate entity serialization

vlthr:
Related issue: #1950

Refactored the various implicit classes that handled serialization of entities into a single class. Entities are now serialized by unioning together the result of serializing their individual subtypes/inherited traits.

I also added a test to statically ensure that the implicit conversions work, and that the serializations are successful (but not whether the result is correct).

Review by @felixmulder

### DIFF
--- a/doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
+++ b/doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
@@ -4,12 +4,11 @@ package model
 import comment._
 import references._
 
-import _root_.java.util.HashMap
 import _root_.java.util.LinkedList
 
 object JavaConverters {
   import scala.collection.JavaConverters._
-  import _root_.java.util.{ Optional => JOptional, Map => JMap }
+  import _root_.java.util.{ Map => JMap }
 
   implicit class OptStr(val opt: Option[String]) extends AnyVal {
     def asJava = opt.getOrElse(null)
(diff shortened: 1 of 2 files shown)


doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
```diff
@@ -298,16 +143,88 @@ object JavaConverters {
...
-    case ent: Package   => ent.asJava(extras)
-    case ent: CaseClass => ent.asJava(extras)
-    case ent: Class     => ent.asJava(extras)
-    case ent: Trait     => ent.asJava(extras)
-    case ent: Object    => ent.asJava(extras)
-    case ent: Def       => ent.asJava
-    case ent: Val       => ent.asJava
-    case ent: TypeAlias => ent.asJava
-    case _              => Map.empty.asJava
+  private def parseEntity(ent: Entity, extras: Map[String, _]): JMap[String, _] = {
```
felixmulder: Are we still using the `extras` or can we kill that parameter?

doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
```diff
@@ -298,16 +143,88 @@ object JavaConverters {
...
+    val entity = Map(
+      "kind"     -> ent.kind,
```
felixmulder: let's kill vertical alignment now that we're writing code that may change:

```diff
+    val entity = Map(
+      "kind" -> ent.kind,
```

doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
```diff
@@ -298,16 +143,88 @@ object JavaConverters {
...
+      )
+      case _ => Map.empty
+    }
+    val df = ent match {
+      case ent: Def => Map(
+        "paramLists"          -> ent.paramLists.map(_.asJava).asJava
+      )
+      case _ => Map.empty
+    }
+      (entity ++ members ++ superTypes ++ modifiers ++ typeParams ++ constructors ++ companion ++ returnValue ++ implicitlyAddedEntity ++ typeAlias ++ trt ++ df).asJava
```
felixmulder: let's make this a:

```scala
{
  entity ++
  members ++
  ...
  df
}.asJava
```

doc-tool/test/JavaConverterTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  ImplicitlyAddedEntity,
+  TypeAlias,
+  Trait,
+  Def,
+  NonEntity
+}
+import model.references.{ConstantReference, TypeReference, NoLink}
+import model.comment.Comment
+import dotty.tools.dotc.core.Symbols.NoSymbol
+import _root_.java.util.{Optional => JOptional, Map => JMap}
```
felixmulder: we can get rid of `_root_` now, it was needed previously for historic (i.e. dumb) reasons.

doc-tool/test/JavaConverterTest.scala
```diff
@@ -0,0 +1,111 @@
...
+
+class JavaConverterTest {
+  import model.JavaConverters._
+
+  @Test def entityConversions = {
+    trait TestEntity extends Entity {
```
felixmulder: No need to create all these intermediate subclasses of `entities.scala` - see below

doc-tool/test/JavaConverterTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      override val kind = "type"
+      def alias = None
+    }
+    trait TestDef extends TestModifiers with TestTypeParams with TestImplicitlyAddedEntity {
+      def paramLists = List()
+    }
+    trait TestTrait extends TestModifiers with TestTypeParams with TestSuperTypes with TestMembers with TestCompanion {
+      def traitParams = List()
+    }
+    val ent = new TestEntity {}
```
felixmulder: Instead of extending and instantiating, just do:

```scala
val ent = new Entity {
  val symbol = ...
  val name = ...
}
...
```

doc-tool/test/JavaConverterTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    val superTypes_serialized = superTypes.asJava()
+    val modifiers_serialized = modifiers.asJava()
+    val typeParams_serialized = typeParams.asJava()
+    val constructors_serialized = constructors.asJava()
+    val companion_serialized = companion.asJava()
+    val returnValue_serialized = returnValue.asJava()
+    val implicitlyAddedEntity_serialized = implicitlyAddedEntity.asJava()
+    val typeAlias_serialized = typeAlias.asJava()
+    val def_serialized = df.asJava()
+    val trait_serialized = trt.asJava()
```
felixmulder: here we're not really testing the result of the conversion, just that the conversions apply to all entities, we could be doing:

```scala
val df = new Def { val name = "foo"; ... }.asJava
assertEquals("foo", df.get("name"))
```

100% test coverage here might be overkill, but perhaps testing a that (pseudo code):

```scala
(Foo extends Bar).asJava.get("superTypes").asScala.toList == List(Foo)
```

felixmulder:
Feel free to include feedback as a second commit, no probs

doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala
vlthr: Nope, looks like none of the current callers actually pass any extras. I'll remove it!

doc-tool/test/JavaConverterTest.scala
vlthr: Is there a way to do this without repeating the fields of Entity (and some others) in each anonymous trait instantiation? I ended up making a bunch of TestX traits implementing each entity trait just to avoid repeating common fields, but I agree that it does feel a little verbose.

doc-tool/test/JavaConverterTest.scala
vlthr: I agree, the tests are pretty weak as they are. I'll add some checking of the actual results - especially the low-hanging fruit.

doc-tool/test/JavaConverterTest.scala
felixmulder: Maybe, but there are only a few concrete instances of the entities, eg:

- Package
- CaseClass
- Class
- Object
- Trait
- Def
- Val
- TypeAlias

See `internal.scala`. No need to instantiate `Entity with Members` et al

vlthr:
Contents of the second commit:

- Added tests to ensure that the results of serialization match their
  pre-serialization values.
- Removed unused parameter `extras` from Entity.asJava() implicit methods.
- Removed _root_ imports
- Fixed several code style issues

The tests are fairly comprehensive (with the most notable omission being no testing of Comment serialization), but at the same time they are fairly repetitive, which I'm worried could lead to them being out of sync with the serialization code in the future. What are your thoughts on using reflection or macros to DRY up the test code. Too much unnecessary complexity?

doc-tool/test/JavaConverterTest.scala
```diff
@@ -3,109 +3,317 @@ package dottydoc
...
+      def comment = None
+      def annotations = List("test")
+      def parent = NonEntity
+      def members = trt :: typeAlias :: Nil
+      def superTypes = new NoLink("title", "query") :: Nil
+    }
+    assertSerializedCorrectly(pkg, pkg.asJava())
+  }
+
+  def assertEach[E, A, C[E] <: Seq[E]](expected: C[E], serialized: Any, pairwiseAssertion: (E, A) => Unit) {
```
felixmulder: If we separate this into two parameter lists, do we still need the type annotation at call site?

felixmulder: e.g:

```scala
def map[A,B](a: A, f: A => B) = f(a)
map(1, x => x * 2) // does not compile, needs type annotation

def map2[A,B](a: A)(f: A => B) = f(a)
map(1)(x => x * 2) // compiles
```

felixmulder:
Well it looks like you had fun with the testing ^^

I'll have a look in the morning, for now just one feedback point already posted.

Cheers,
Felix

doc-tool/test/JavaConverterTest.scala
vlthr: Nope, looks like we don't! Removing the type annotations significantly improves the readability of the calls.

Can I amend the commit, or would that confuse the system?

felixmulder:
just rebase and force push, that's fine :)

doc-tool/test/JavaConverterTest.scala
```diff
@@ -0,0 +1,320 @@
...
+  }
+  def assertSerializedCorrectly(expected: Entity, serialized: Any) {
+    val actual = serialized.asInstanceOf[JMap[String, _]]
+    assertEquals(expected.name, actual.get("name"))
+    assertEquals(expected.kind, actual.get("kind"))
+    assertSameSeq(expected.annotations, actual.get("annotations"))
+    assertSameSeq(expected.path, actual.get("path"))
+    assertEquals(expected.hasShortenedDocstring, actual.get("hasShortenedDocstring"))
+    // Only test if a comment is present
+    expected.comment match {
```
felixmulder: this is fine, but the idiomatic way to do this would be:

```scala
expected.comment.foreach(assertNotEquals(null, actual.get("comment")))
```

felixmulder:
Thanks @vlthr! Happy to get this merged :tada:

PR #2076, opened 2017-03-11
Fix overriding a Java method with varargs

smarter:
If A method like:
```scala
  override def foo(x: Object*)
```
overrides a Java method, it needs to be rewritten as:
```scala
  def foo(x: Seq[Object])
  override def foo(x: Array[Object]): Object = foo(Predef.wrapRefArray(x))
```

This should be handled by ElimRepeated but there were two bugs:
- `addVarArgsBridge` was called at phase `thisTransformer.next`, this is
too late to create the bridge since `T*` has already been rewritten as
`Seq[T]`
- The original method symbol needs to have the `override` flag dropped,
since it doesn't override anything.

Furthermore, RefChecks had to be moved after ElimRepeated, otherwise
the testcase would fail the overriding checks.

<hr>

@odersky I'm not sure if moving RefChecks is the best solution here. Alternatively we could teach TypeComparer about repeated params when we compare MethodTypes in a phase before ElimRepeated.

### DIFF
--- a/compiler/src/dotty/tools/dotc/Compiler.scala
+++ b/compiler/src/dotty/tools/dotc/Compiler.scala
@@ -48,9 +48,9 @@ class Compiler {
       List(new Pickler),            // Generate TASTY info
       List(new FirstTransform,      // Some transformations to put trees into a canonical form
            new CheckReentrant),     // Internal use only: Check that compiled program has no data races involving global vars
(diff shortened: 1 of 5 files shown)


smarter:
Oops, I forgot to add the testcase, I'll amend the commit.

PR #2078, opened 2017-03-12
Fix #360: Improve avoidance algorithm

### LINKED ISSUE #360: The rules for block typing are non-intuitive
This example  from `tests/pos/t1569a.scala` works:

``` scala
object Bug2 {
  class C { type T }
  class D extends C { type T = String }
  def foo(x: Int)(y: C)(z: y.T): Unit = {}
  foo(3)(new D)("hello")
}
```

But replacing `new D` by `new D {}` fails with:

```
tests/pos/t1569a.scala:11: error: type mismatch:
 found   : String("hello")
 required: Bug2.C#T
  foo(3)(new D { })("hello")
                    ^
one error found
```

Whether the expected type of the parameter is used or not is determined by `Typer#ensureNoLocalRefs`

### LINKED ISSUE #1569: Inline def with `inline x: => X` parameters do not respect semantics.

odersky:
The essential change is that we do not throw away more
precise info of the avoided type if the expected type
is fully defined.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -625,14 +625,17 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     block.tpe namedPartsWith (tp => locals.contains(tp.symbol))
   }
 
-  /** Check that expression's type can be expressed without references to locally defined
-   *  symbols. The following two remedies are tried before giving up:
-   *  1. If the expected type of the expression is fully defined, pick it as the
-   *     type of the result expressed by adding a type ascription.
-   *  2. If (1) fails, force all type variables so that the block's type is
-   *     fully defined and try again.
+  /** Ensure that an expression's type can be expressed without references to locally defined
+   *  symbols. This is done by adding a type ascription of a widened type that does
+   *  not refer to the locally defined symbols. The widened type is computed using
+   *  `TyperAssigner#avoid`. However, if the expected type is fully defined and not
+   *  a supertype of the widened type, we ascribe with the expected type instead.
+   *
+   *  There's a special case having to do with anonymous classes. Sometimes the
+   *  expected type of a block is the anonymous class defined inside it. In that
+   *  case there's technically a leak which is not removed by the ascription.
    */
-  protected def ensureNoLocalRefs(tree: Tree, pt: Type, localSyms: => List[Symbol], forcedDefined: Boolean = false)(implicit ctx: Context): Tree = {
+  protected def ensureNoLocalRefs(tree: Tree, pt: Type, localSyms: => List[Symbol])(implicit ctx: Context): Tree = {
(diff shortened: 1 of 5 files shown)


smarter:
I don't see how this is related to #1569, did you mean some other issue number? I can suggest https://github.com/lampepfl/dotty/issues/360 which I'm happy to see is fixed by this PR!

odersky:
Oh, it was SI-1659, not our issues.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -642,12 +642,18 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     }
     val leaks = escapingRefs(tree, localSyms)
     if (leaks.isEmpty) tree
-    else if (isFullyDefined(pt, ForceDegree.none)) ascribeType(tree, pt)
     else if (!forcedDefined) {
       fullyDefinedType(tree.tpe, "block", tree.pos)
-      val tree1 = ascribeType(tree, avoid(tree.tpe, localSyms))
-      ensureNoLocalRefs(tree1, pt, localSyms, forcedDefined = true)
-    } else
+      val avoidingType = avoid(tree.tpe, localSyms)
```
smarter: The comment on top of this method needs to be updated now that the order of operations changed.

odersky:
@smarter I investigated whether we could base TypeAssigner#avoid on ApproximatingTypeMap. But it seems to be pretty tricky to do so, so for the moment I abandoned the effort. See branch change-avoid-via-approx on staging for the current state of the effort. We still get discrepancies where the new avoid is not as good as the old one. I am going to let it be for now, but if you want to take a look please go ahead.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -625,37 +625,36 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-  /** Check that expression's type can be expressed without references to locally defined
-   *  symbols. The following two remedies are tried before giving up:
-   *  1. If the expected type of the expression is fully defined, pick it as the
-   *     type of the result expressed by adding a type ascription.
-   *  2. If (1) fails, force all type variables so that the block's type is
-   *     fully defined and try again.
+  /** Ensure that an expression's type can be expressed without references to locally defined
+   *  symbols. This is done by adding a type ascription of a widened type that does
+   *  not refer to the locally defined symbols. The widened type is computed using
+   *  `TyperAssigner#avoid`. However, if the expected type is fully defined and not
```
smarter: typo: TyperAssigner -> TypeAssigner

smarter:
> We still get discrepancies where the new avoid is not as good as the old one. I am going to let it be for now, but if you want to take a look please go ahead.

Thanks! I'll see if I have time to have a look.

PR #2079, opened 2017-03-12
Allow inter-parameter dependencies

odersky:
Allow a later parameter's type to refer to an earlier parameter in the same parameter section. Previously, we allowed only references to parameters in preceding sections. This is simpler, but it
can force unnatural currying. The problem becomes urgent only if we want to add dependent types more broadly. It feels natural to write 

    def f(n: Int, v: Vec(n))

and being forced to curry this as `def fun(n: Int)(v: Vec(n))` is annoying. 

The changes in this PR have given me an idea for a much more sweeping change. We might be able to unify MethodTypes and PolyTypes, which would improve regularity of the language quite significantly and which would open up lots of new possibilities. But one thing after another.

Review by @gsps?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -278,7 +278,8 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
               result
             case METHODtype =>
               val (names, paramReader) = readNamesSkipParams
-              val result = MethodType(names.map(_.toTermName), paramReader.readParamTypes[Type](end))(
+              val result = MethodType(names.map(_.toTermName))(
+                mt => paramReader.readParamTypes[Type](end), // !!!
                 mt => registeringType(mt, readType()))
               goto(end)
               result
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -279,8 +279,8 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
             case METHODtype =>
               val (names, paramReader) = readNamesSkipParams
               val result = MethodType(names.map(_.toTermName))(
-                mt => paramReader.readParamTypes[Type](end), // !!!
-                mt => registeringType(mt, readType()))
+                mt => registeringType(mt, paramReader.readParamTypes[Type](end)),
+                mt => readType())
               goto(end)
               result
             case PARAMtype =>
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -1263,7 +1263,8 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
 
         def sizeFits(alt: TermRef, tp: Type): Boolean = tp match {
           case tp: PolyType => sizeFits(alt, tp.resultType)
-          case MethodType(_, ptypes) =>
+          case tp: MethodType =>
+            val ptypes = tp.paramTypes
             val numParams = ptypes.length
             def isVarArgs = ptypes.nonEmpty && ptypes.last.isRepeatedParam
             def hasDefault = alt.symbol.hasDefaultParams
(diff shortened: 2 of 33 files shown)


odersky:
This PR contains in its first and last commit the commits of #2078.

smarter:
#2078 merged, you should be able to rebase.

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -233,6 +233,12 @@ object ProtoTypes {
       typer.adapt(targ, formal, arg)
     }
 
+    /** The type of the argument `arg`.
+     *  @pre `arg` ahs been typed before
```
smarter: typo: ahs -> has

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -278,7 +278,8 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
               result
             case METHODtype =>
               val (names, paramReader) = readNamesSkipParams
-              val result = MethodType(names.map(_.toTermName), paramReader.readParamTypes[Type](end))(
+              val result = MethodType(names.map(_.toTermName))(
+                mt => paramReader.readParamTypes[Type](end), // !!!
```
gsps: That "!!!" is slightly alarming :-)

I suppose you are referring to the fact that we now rely on the constructor calling the two closures in the right order, which is somewhat risky indeed. Why not read them explicitly and construct a constant function?

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -625,6 +624,24 @@ trait Checking {
       case _ =>
     }
   }
+
+  /** Check that method parameter types do not reference their own parameter
+   *  or later parameters in the same parameter section.
+   */
+  def checkNoForwardDependencies(vparams: List[ValDef])(implicit ctx: Context): Unit = vparams match {
```
gsps: I'm sure there's a good reason, but it seems that here we are duplicating the checking provided by Config.checkMethodTypes. Why do we need both?

tests/pos/param-depmeth.scala
```diff
@@ -0,0 +1,15 @@
+object Test {
+
+  class C { type T }
+
+  def f(x: C, y: x.T): x.T = y // ok
+
+  val c = new C { type T = String }
+  val c2 = c
```
gsps: I think we only added `val c2` to see what type is inferred, so we can now remove it, no? A similar thing goes for the `val d` below, I think.

gsps:
(I unfortunately put one of my comments in the outdated version of TreeUnpickler.scala, so the comment is collapsed by default. I think the point still stands, please have a look.)

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -380,8 +382,16 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
         case formal :: formals1 =>
 
-          def addTyped(arg: Arg, formal: Type) =
+          /** Add result of typing argument `arg` against parameter type `formal`.
+           *  @return  A type transformation to apply to all arguments following this one.
+           */
+          def addTyped(arg: Arg, formal: Type): Type => Type = {
             addArg(typedArg(arg, formal), formal)
+            if (methodType.isParamDependent)
+              _.substParam(MethodParam(methodType, n), typeOfArg(arg))
```
odersky: I am not convinced it will be a problem. First, how many methods will be `paramDependent`? Then, how long is the typical parameter list? If it is 2, then the code given is the best (just one substitution. If it is 3, or 4, not much worse. Also, note that single parameter substitutions are faster than multiple parameter ones which require a linear search anyway. So, in summary, let's wait until the profiler indicates this is a hotspot.

compiler/src/dotty/tools/dotc/typer/Checking.scala
odersky: One is a source check, the other an internal assertion that results in an exception. Not all method types are created directly from source, so the assertion is useful.

tests/pos/param-depmeth.scala
odersky: Yes, but it's a test case. Does not really matter either way.

PR #2080, opened 2017-03-12
Fix #2066: Don't qualify private members in SelectionProto's...

### LINKED ISSUE #2066: methods available via implicit conversions should not include private methods
```scala
class Foo

object Test {
  implicit class One(x: Foo) {
    def meth: Unit = {}
  }

  implicit class Two(x: Foo) {
    private def meth: Unit = {}
  }

  def test(foo: Foo): Unit = {
    foo.meth
  }
}
```
```scala
-- [E007] Type Mismatch Error: try/privimpl.scala ------------------------------
13 |    foo.meth
   |    ^^^
   |found:    Foo(foo)
   |required: ?{ meth: ? }
   |
   |
   |Note that implicit conversions cannot be applied because they are ambiguous;
   | both method Two in object Test and method One in object Test convert from Foo(foo) to ?{ meth: Unit }
```

odersky:
... unless they would be accessible in the given context.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -957,6 +957,10 @@ object SymDenotations {
       else
         companionNamed(name)(ctx.outersIterator.dropWhile(_.scope eq ctx.scope).next)
 
+    /** Is this symbol the same or a linked class of `sym`? */
+    final def isLinkedWith(sym: Symbol)(implicit ctx: Context): Boolean =
+      (symbol eq sym) || (linkedClass eq sym)
+
     /** If this is a class, the module class of its companion object.
      *  If this is a module class, its companion class.
      *  NoSymbol otherwise.
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -215,13 +215,13 @@ object TypeErasure {
   }
 
   /** The erased least upper bound is computed as follows
-   *  - if both argument are arrays of objects, an array of the lub of the element types
+   *  - if both argument are arrays of objects, an array of the erased lub of the element types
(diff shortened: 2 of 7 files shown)


liufengyun:
The fix fails the following case if we change `private` to `protected`:

```Scala
class Foo

object Test {
  implicit class One(x: Foo) {
    def meth: Unit = {}
  }

  implicit class Two(x: Foo) {
    protected def meth: Unit = {}    //  <==== change private to protected
  }

  def test(foo: Foo): Unit = {
    foo.meth
  }
}
```

odersky:
> The fix fails the following case if we change private to protected:

That's expected. `private` behaves differently from the other modifiers. For private/non-private we have different ways to search. For the other modifiers we search first, and then check that the result is accessible. But that check plays no role in (say) overloading resolution or implicit search. This has been always like this, in Java as well as in Scala, and it would be really hard to change.

smarter:
Is this as expected? This doesn't compile:
```scala
class Foo

object Test {
  implicit def two(x: Foo): Two = new Two(x)

  class Two(x: Foo) {
    private def meth: Unit = {}

    def test2(foo: Foo): Unit = {
      foo.meth
    }
  }
}
```
But this does:
```scala
class Foo

object Test {

  class Two(x: Foo) {
    implicit def two(x: Foo): Two = new Two(x)

    private def meth: Unit = {}

    def test2(foo: Foo): Unit = {
      foo.meth
    }
  }
}
```

(Both fail with scalac)

odersky:
> Is this as expected? This doesn't compile:

No, this is weird. It looks like we can't do a good job distinguishing here without losing a lot of implicit cachability. So I'll change it to scalac's behavior to only look for non-private members.

smarter:
Okay, this will require a spec change I think since the spec only mentions accessibility: http://www.scala-lang.org/files/archive/spec/2.12/07-implicits.html#views

odersky:
So, does that means scalac also is in disagreement with the spec here?

smarter:
I think so yes, but I haven't looked closely.

PR #2082, opened 2017-03-12
Super Bootstrap: Make dottydoc part of the bootstrap

smarter:
This wasn't done before because dotty could not compile dottydoc, this
got fixed by PRs #2070 and #2076.


smarter:
Heh, this found a bug in dotty! See https://github.com/lampepfl/dotty/issues/2084

smarter:
@felixmulder Looks like this is finally green!

felixmulder:
Indeed, congrats!

smarter:
@felixmulder Did you rebase instead of making a merge commit? So far we've always used merge commits and mixing the two styles would be really confusing. Plus, currently we only require the latest commit to pass the tests, as long as we use merge commits this isn't too much of a problem since we can still do bisection based on merge commits, but if we don't have merge commits it becomes impossible to do bisection.

PR #2083, opened 2017-03-12
-Yplain-printer: do not print positions by default

smarter:
You can still have positions printed by explictly passing -Yprintpos

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -483,7 +483,7 @@ class PlainPrinter(_ctx: Context) extends Printer {
           else
             Text()
 
-        nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ node.pos.toString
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -483,7 +483,11 @@ class PlainPrinter(_ctx: Context) extends Printer {
           else
             Text()
 
-        nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ node.pos.toString
+        val text = nodeName ~ "(" ~ elems ~ tpSuffix ~ ")"
+        if (ctx.settings.Yprintpos.value)
+          text ~ node.pos.toString
+        else
+          text
       case _ =>
```
odersky: Just FYI, there's a more fluid way to do this:

    nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ (node.pos.toString provided ctx.settings.Yprintpos.value)

smarter: Thanks, updated!

PR #2085, opened 2017-03-13
Fix #2072: erasure and refchecks

### LINKED ISSUE #2072: Concrete trait member should take precedence over deferred method
```scala
trait T { lazy val x: String = "foo" }
abstract class AC extends T { def x: Any }
```

compiles to:
```java
public abstract class AC implements T {
    public AC() { super.x();   }

    public abstract Object x(); // where did my lazy val go?
}

public interface T {
    default public void $init$() { }

    default public String x() {
        return "foo";
    }
}
```

See https://issues.scala-lang.org/browse/SI-10223

DarkDimius:


compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
```diff
@@ -96,23 +96,30 @@ trait TypeTestsCasts {
         /** Transform isInstanceOf OrType
```
smarter: Not specific to OrType anymore

compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
```diff
@@ -96,23 +96,30 @@ trait TypeTestsCasts {
...
          *
          *    expr.isInstanceOf[A | B]  ~~>  expr.isInstanceOf[A] | expr.isInstanceOf[B]
+         *    expr.isInstanceOf[A & B]  ~~>  expr.isInstanceOf[A] & expr.isInstanceOf[B]
```
smarter: Would you mind moving this commit to a separate PR? It's not related to fixing #2072 and I'd like to get it in soon because https://github.com/lampepfl/dotty/pull/2082 depends on it.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -642,7 +647,22 @@ object Erasure extends TypeTestsCasts{
...
-                    sym => makeBridgeDef(member, sym)(ctx)
+                    sym =>
+                      if (member.symbol.is(Flags.Deferred)) {
+                        if (!sym.is(Flags.Deferred)) {
+                          // sometimes we need to forward an abstract method in a subclass to a non-abstract method in parent.
+                          // see https://github.com/lampepfl/dotty/issues/2072 for illustrations
+
+                          val d = member.symbol.copySymDenotation(initFlags = member.symbol.flags.&~(Flags.Deferred))
+                          d.installAfter(ctx.erasurePhase.asInstanceOf[Erasure])
+                          // Dark magic. Don't look here.
```
smarter: This is not a helpful comment. Someone might need to look at this eventually and will be very confused. Also do you need both `installAfter` and `transformAfter` ?

DarkDimius: > Also do you need both installAfter and transformAfter ?

Yes. The comment here would entirely repeat all the history of transformAfter. I don't think it's worth simplifying it here. To understand why it's needed here one need to read the entire description of 979ee0ff41c2a957b22e886ea43ad4f070a9777d which introduces transformAfter

smarter: Then the comment can say "See the commit message of 979ee0f for more information."

compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
smarter: An identical commit has now been merged as part of https://github.com/lampepfl/dotty/pull/2082

PR #2094, opened 2017-03-14
Fix #2071: handle HKApply in SAMType

### LINKED ISSUE #2071: Missing type parameter in eta expansion for case blocks
Discovered by @rinfield:

```scala
object Test {
  type PF[A, B] = PartialFunction[A, B]

  val f: PF[Int, String] = {
    case i => "bar"
  }
}
```

Fails with:

```scala
-- Error: local/pf.scala -------------------------------------------------------
5 |    case i: Int => "bar"
  |    ^
  |    missing parameter type for parameter x$1 of expanded function x$1 =>
  |      x$1 @unchecked match
  |        {
  |          case i: Int =>
  |            "bar"
  |        }, expected = Test.PF[Int, String]
```

Works fine when the proper variance added.

liufengyun:
Fix #2071: handle HKApply in SAMType

Previously, the logic in `SAMType` does not take into consideration type constructors.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3328,6 +3328,8 @@ object Types {
         zeroParamClass(tp.underlying)
       case tp: TypeVar =>
         zeroParamClass(tp.underlying)
+      case tp: HKApply =>
+        zeroParamClass(tp.superType)
       case _ =>
         NoType
     }
--- a/tests/pos/i2071.scala
+++ b/tests/pos/i2071.scala
@@ -1,0 +1,7 @@
+object Test {
+  type PF[A, B] = PartialFunction[A, B]
+
+  val f: PF[Int, String] = {
+    case i => "bar"
+  }
+}


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3358,6 +3358,8 @@ object Types {
           absMems.find(_.symbol.name == nme.apply)
         else None
       }
+      else if (tp.isInstanceOf[HKApply])
+        unapply(tp.asInstanceOf[HKApply].superType)
       else None
```
odersky: Good diagnosis to fix the bug! For the fix, can you try instead to add a case to `zeroParamClass` above?

    case tp: HKApply => zeroParamClass(tp.superType)

That would be more regular.

liufengyun: Thanks @odersky , indeed the code should go to `zeroParamClass`, fixed in the new commit.

odersky:
LGTM

PR #2095, opened 2017-03-14
Fix #1488: position fix for lazy vals

### LINKED ISSUE #1488: [debug] unfriendly jumping in evaluating lazy val block
When stepping through the evaluation of the lazy block, it will jump back and forth between line 1 and `LazyHolders.scala` several times, that's acceptable, given how lazy vals are implemented.

However, it also jumps to the end of the block **before** evaluation of the body, which is not friendly.

``` Scala
0  def test() = {
1     lazy val a = {
2       val b = 8 * 9
3       val c = b + 5
4       c + b
5     }
6     println(a)
7  }
```

liufengyun:
Fixes #1488 related to lazy val positions.


odersky:
LGTM

PR #2096, opened 2017-03-14
Fix #2051: allow override T with => T or ()T

### LINKED ISSUE #2051: Cannot override a val from the parent class
This works in Scalac, but not in Dotty:

```scala
class A[T](val x:T)
class B[T](override val x:T) extends A[T](x)
```

Dotty complains:

```scala
Welcome to Scala.next (pre-alpha, git-hash: 06d3f7a)  (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_65).
Type in expressions to have them evaluated.
Type :help for more information.
scala> class A[T](val x:T)
defined class A
scala> class B[T](override val x:T) extends A[T](x)
-- Error: <console> ------------------------------------------------------------
5 |class B[T](override val x:T) extends A[T](x)
  |                        ^
  |                        overriding value x in class A of type T;
  |                         method x of type => T has incompatible type
```

Removing the `override val` keywords in `B` makes it compile, but it's not always possible (e.g. for case classes).

liufengyun:
Fix #2051: allow override `T` with `=> T` or `()T`.

I'm not sure if this is the right fix. Could you please advise @DarkDimius ?

### DIFF
--- a/tests/pos/i2051.scala
+++ b/tests/pos/i2051.scala
@@ -1,2 +1,9 @@
 class A[T](val x:T)
 class B[T](override val x:T) extends A[T](x)
+
+class C[T](val x:T, val y: Int, val z: Boolean)
+class D[T](override val x:T, y: Int, z: Boolean) extends C[T](x, y, z)
+
+trait X(val x: Int, y: Int, z: Int)
+trait Y(override val x: Int, y: Int, z: Int) extends X
+class Z(override val x: Int, y: Int, z: Int) extends Y(x, y, z) with X(x, y, z)
--- a/tests/pos/i2051.scala
+++ b/tests/pos/i2051.scala
@@ -1,0 +1,2 @@
+class A[T](val x:T)
+class B[T](override val x:T) extends A[T](x)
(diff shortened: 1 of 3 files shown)


odersky:
What about the following test:

    abstract class C { val x: Int }
    class D extends C { def x = 1 }

We are good if that one still gives the error

              overriding value x in class C of type Int;
              method x of type => Int needs to be a stable, immutable value

liufengyun:
The case above is rejected by some other check, we still have the same error as before. I just added it as a neg test.

odersky:
OK, I think we are good then. @smarter since you are currently playing with override val's do you see a reason why we should not do this?

smarter:
Not that I can think of but this change still seems weird to me, what is special about constructor parameters here?

tests/pos/i2051.scala
```diff
@@ -0,0 +1,2 @@
+class A[T](val x:T)
```
DarkDimius: it would be nice to add several fields to the body of the class that are overriden in a subclass.

It would also be nice to include tests with trait vals and trait constructors.

liufengyun:
@smarter the reason is that `x: T` is already present in the parent class, in post-typer, we just create a forwarder `override def x: T = super.x`.

tests/pos/i2051.scala
liufengyun: Thanks @DarkDimius , I just added tests in the latest commit.

PR #2100, opened 2017-03-14
Various fixes to the sbt reporter

smarter:
This doesn't fix the issue that the reporter is duplicating information returned by dotty itself, if no one can come up for a fix for that soon (it might require changes to sbt) I'll just disable position reporting for the sbt reporter again.

/cc @MasseGuillaume

### DIFF
--- a/sbt-bridge/src/xsbt/DelegatingReporter.scala
+++ b/sbt-bridge/src/xsbt/DelegatingReporter.scala
@@ -28,17 +28,18 @@ final class DelegatingReporter(delegate: xsbti.Reporter) extends Reporter
       }
 
     val position =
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
@smarter 
https://github.com/dotty-staging/dotty/blob/330de0410ce5be5b54482bf00578771cf821af24/sbt-bridge/src/xsbt/DelegatingReporter.scala#L20

have you tried `override def printSummary(implicit ctx: Context): Unit = ()` ?

smarter:
@MasseGuillaume No, look at the definition of `printSummary` in `Reporter`, it only displays a message like "X errors found" at the end of the compilation run.

PR #2101, opened 2017-03-14
Upgrade to sbt 0.13.14-RC2, switch to build.sbt

smarter:
Let's wait until https://github.com/sbt/sbt/issues/3012 is fixed first.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,0 +1,19 @@
+val dotty = Build.dotty
+val `dotty-bootstrapped` = Build.`dotty-bootstrapped`
+val `dotty-interfaces` = Build.`dotty-interfaces`
+val `dotty-doc` = Build.`dotty-doc`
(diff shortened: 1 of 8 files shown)


felixmulder:
:+1: for removing the unused scalastyle plugin

smarter:
@felixmulder I think this is ready to go in (even though sbt 0.13.14 is not released yet, this way we may be able to get potential issues fixed before the release)

felixmulder:
What about https://github.com/sbt/sbt/issues/3012?

felixmulder:
Otherwise LGTM

smarter:
Using org.scala-sbt:interface is fine even with coursier as long as we add resolvers to all subprojects as explained in https://github.com/coursier/coursier/issues/284#issuecomment-230317311

PR #2102, opened 2017-03-14
Fix #2099: Make param forwarders override the param they forward to

### LINKED ISSUE #2099: Separate compilation issue because PrivateParamAccessor are only made non-private in TreeTransformer, not DenotTransformer
`A.scala`:
```scala
class A(val member: Int)

class SubA(member: Int) extends A(member)
```
`B.scala`:
```scala
class B(member: Int) extends SubA(member)
```

If you try to compile these two files together everything works fine, but if you first compile `A.scala` and then `B.scala` you get:
```scala
Exception in thread "main" java.lang.AssertionError: assertion failed: private method member in class SubA in try/exp/A.scala accessed from method member in class B in try/exp/B.scala
        at dotty.tools.dotc.transform.ExpandPrivate.ensurePrivateAccessible(ExpandPrivate.scala:84)
```

The issue is that `ExpandPrivate` only transforms `PrivateParamAccesor` in trees:
```scala
  override def transformDefDef(tree: DefDef)(implicit ctx: Context, info: TransformerInfo) = {
    val sym = tree.symbol
    tree.rhs match {
      case Apply(sel @ Select(_: Super, _), _)
      if sym.is(PrivateParamAccessor) && sel.symbol.is(ParamAccessor) && sym.name == sel.symbol.name =>
        sym.ensureNotPrivate.installAfter(thisTransform)
      case _ =>
        if (isVCPrivateParamAccessor(sym))
          sym.ensureNotPrivate.installAfter(thisTransform)
    }
    tree
  }
```

So when we compile `A.scala` and `B.scala` together, the denotation for `SubA#member` gets transformed as expected, but when we do separate compilation, we see `SubA#member` as private, so `ExpandPrivate` explodes.
Note: see the comment at the top of `ExpandPrivate` if you're wondering why private param forwarders need to be made non-private.

I'm unsure how to fix this since it seems that we need access to the tree to decide whether or not the param forwarder should be made non-private. Maybe we need an extra flag to signal that a param forwarder forwards to an inherited param accessor with the same name?

This issue is blocking the [Super Bootstrap](https://github.com/lampepfl/dotty/pull/2082)

smarter:
Previously, in the following code:
```scala
class A(val member: Int)
class SubA(member: Int) extends A(member)
```
We created a forwarder in SubA like this:
```scala
  private[this] def member: Int = super.member
```
Since a private method cannot have the same name as an inherited
non-private one, we subsequently name-mangled the forwarder in
ExpandPrivate. However, this was only done in the TreeTransformer, not
in the DenotTransformer, this lead to separate compilation issues (see
added testcase).

Since we cannot detect that a ParamAccessor is a forwarder in a
DenotTransformer, we cannot fix the problem locally in ExpandPrivate.
Instead, we change ParamForwarding so that the forwarder now looks like:
```scala
  override def member: Int = super.member
```
To make this work we restrict ParamForwarding to only create forwarders
for non-final fields.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
@@ -3,12 +3,13 @@ package transform
 
 import core._
 import dotty.tools.dotc.ast.tpd
-import dotty.tools.dotc.core.DenotTransformers.{SymTransformer, IdentityDenotTransformer}
+import dotty.tools.dotc.core.DenotTransformers.{SymTransformer, DenotTransformer}
 import Contexts.Context
 import Symbols._
 import Scopes._
 import Flags._
 import StdNames._
+import Denotations._
 import SymDenotations._
 import Types._
 import collection.mutable
@@ -25,15 +26,10 @@ import ValueClasses._
  *  Make private accessor in value class not-private. Ihis is necessary to unbox
  *  the value class when accessing it from separate compilation units
  *
- *  Also, make non-private any private parameter forwarders that forward to an inherited
- *  public or protected parameter accessor with the same name as the forwarder.
- *  This is necessary since private methods are not allowed to have the same name
- *  as inherited public ones.
- *
  *  See discussion in https://github.com/lampepfl/dotty/pull/784
  *  and https://github.com/lampepfl/dotty/issues/783
  */
-class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { thisTransform =>
+class ExpandPrivate extends MiniPhaseTransform with DenotTransformer { thisTransform =>
   import ast.tpd._
 
   override def phaseName: String = "expandPrivate"
@@ -56,6 +52,15 @@ class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { t
     }
   }
 
+  override def transform(ref: SingleDenotation)(implicit ctx: Context): SingleDenotation =
+    ref match {
+      case ref: SymDenotation if isVCPrivateParamAccessor(ref) =>
+        ref.ensureNotPrivate
+      case _ =>
+        ref
+    }
+
+
   private def isVCPrivateParamAccessor(d: SymDenotation)(implicit ctx: Context) =
     d.isTerm && d.is(PrivateParamAccessor) && isDerivedValueClass(d.owner)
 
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
```diff
@@ -95,17 +98,4 @@ class ExpandPrivate extends MiniPhaseTransform with IdentityDenotTransformer { t
     ensurePrivateAccessible(tree.symbol)
     tree
   }
-
-  override def transformDefDef(tree: DefDef)(implicit ctx: Context, info: TransformerInfo) = {
```
DarkDimius: This code is still needed.
We still have methods in the same compilation unit that access private stuff.
See eg. https://github.com/lampepfl/dotty/issues/2054

smarter: This method only ever changed denotations for param accessors. The rest is handled by `transformIdent` and `transformSelect` above which I did not touch. If you merge your fix for #2054 I'll rebase to make sure everything still works.

DarkDimius: You're right.

smarter:
Hah! I am smitten by the own diabolical tests I wrote. Here's a minimization:
```scala
class A(val theValue: Int)
class NonVal(theValue: Int) extends A(theValue) {
  def getTheValueInNonVal = theValue
}
class X(override val theValue: Int) extends NonVal(0)


object Test {
    val x14 = new X(14)

    assert(x14.getTheValueInNonVal == 0)
  }
}
```
In both scalac and dotty before this PR, the assertion succeeds. But it now fails because calling `theValue` in `NonVal` is now a virtual call that is overriden in `X`. @odersky : What do you think of this change?

odersky:
Yes, I guess 0 is right.

smarter:
@odersky Then I don't think the proposed approach of making forwarders override the param they forward to can work.

DarkDimius:
@smarter, I've proposed an approach to name-mangle forwarders while keeping them public.

smarter:
@DarkDimius This makes things more complicated because we look for an inherited accessor by name in ParamForwarding (see `inheritedAccessor`), if we name-mangle forwarders we need to look for two names: the name-mangled one and the regular one. It does avoid other problems so this might be the best solution we can have.

smarter:
@DarkDimius I give up, I've tried various things but I can't get .expandedPrivate to work correctly before pickling, too many things break. Could you give it a try?

DarkDimius:
@smarter, I'll have a look.

smarter:
Closed in favor of https://github.com/lampepfl/dotty/pull/2106

PR #2103, opened 2017-03-14
Encode variances in parameter names

odersky:
This leads to a slight overall simplification, harmonizes pickle
format with internal representation, and makes MethodTypes and
PolyTypes more similar to each other.

I believe the change is useful as it is, but in particular it is
a useful step for an eventual unification of MethodTypes and
PolyTypes.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -241,20 +241,20 @@ object desugar {
    *     class C { type v C$T; type v T = C$T }
    */
   def typeDef(tdef: TypeDef)(implicit ctx: Context): Tree = {
-    val name =
+    val checkedName =
       if (tdef.name.hasVariance && tdef.mods.is(Param)) {
         ctx.error(em"type parameter name may not start with `+' or `-'", tdef.pos)
         ("$" + tdef.name).toTypeName
       }
       else tdef.name
     if (tdef.mods is PrivateLocalParam) {
-      val tparam = cpy.TypeDef(tdef)(name = tdef.name.expandedName(ctx.owner))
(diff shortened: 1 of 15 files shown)


DarkDimius:
A very nice simplification!

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -241,6 +241,12 @@ object desugar {
    *     class C { type v C$T; type v T = C$T }
    */
   def typeDef(tdef: TypeDef)(implicit ctx: Context): Tree = {
+    val name =
+      if (tdef.name.hasVariance && tdef.mods.is(Param)) {
+        ctx.error(em"type parameter name may not start with `+' or `-'", tdef.pos)
```
smarter: Is this even allowed by the parser?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2567,8 +2567,20 @@ object Types {
     }
   }
 
-  /** A type lambda of the form `[v_0 X_0, ..., v_n X_n] => T` */
-  class PolyType(val paramNames: List[TypeName], val variances: List[Int])(
+  /** A type lambda of the form `[X_0 B_0, ..., X_n B_n] => T`
+   *  This is used both as a type of a polymorphic method and as a type of
+   *  a higher-kidned type parameter. Variances are encoded in parameter
```
smarter: typo: kidned -> kinded

compiler/src/dotty/tools/dotc/ast/Desugar.scala
odersky: Yes, ++, -- are legal parameter names.

odersky:
Superseded by #2121

PR #2106, opened 2017-03-15
Alternative fix for #2099: avoid loading a private member when recomputing a NamedType denot

smarter:
ParamForwarding creates the following forwarder in B:
```scala
  private[this] def member: Int = super.member
```
Where the type for `super.member` is `TermRef(SubA, member)` and the
  symbol is the `val member` in `A`.
So far this is correct, but in later phases we might call `loadDenot` on
this `TermRef` which will end up calling `asMemberOf`, which before this
commit just did:
```scala
  prefix.member(name)
```
This is incorrect in our case because `SubA` also happens to have a
private `def member`, which means that our forwarder in B now forwards
to a private method in a superclass, this subsequently crashes in
`ExpandPrivate`.
(Note: in the bytecode, a private method cannot have the same name as an
overriden method, but this is already worked around in EnsurePrivate.)

The fix is simple: when we recompute a member, we should only look at
private members if the previous denotation was private.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1482,7 +1482,7 @@ object Types {
 
     /** A member of `prefix` (disambiguated by `d.signature`) or, if none was found, `d.current`. */
     private def recomputeMember(d: SymDenotation)(implicit ctx: Context): Denotation =
-      asMemberOf(prefix) match {
+      asMemberOf(prefix, allowPrivate = d.is(Private)) match {
         case NoDenotation => d.current
         case newd: SingleDenotation => newd
         case newd =>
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1585,11 +1585,12 @@ object Types {
...
-        val d = lastDenotation
-        // Never go from a non-private denotation to a private one
-        if (d == null || d.symbol.is(Private)) prefix.member(name)
-        else prefix.nonPrivateMember(name)
+      else lastDenotation match {
+        case d: SymDenotation if !d.is(Private) =>
+          // We shouldn't go from a non-private denotation to a private one
+          prefix.nonPrivateMember(name)
+        case _ =>
+          prefix.member(name)
```
odersky: On second thought, I think we should revert to the previous scheme where `allowPrivate` was passed as a boolean to `asMemberOf`. There are 3 calls to `asMemberOf`. The ones in `recomputeDenot` and `newLikeThis` can get the parameter from the symbol. The third one, in `loadDenot` should pass `false` for
`allowPrivate`. After all, it could be that a definition was public in run 1, but private in run 2. We need to be able to pick up the new definition in `loadDenot`.

smarter:
Review addressed. However I'm wondering if it wouldn't make sense to take advantage of the existing `shadowed` mechanism. In `ParamForwarder` we manually look for an accessible denotation using `isAccessibleFrom`, but this is sort of `TypeAssigner#ensureAccessible` already does.

PR #2107, opened 2017-03-15
sbt-bridge: Fix #2092: Disable position information in the reporter

### LINKED ISSUE #2092: sbt-bridge: Broken display of reported messages
If you run `sbt dotty-compiler-bootstrapped/compile` you'll see things like:
2017-03-13-235341_3840x2160_scrot

Two things are wrong:
- The filename and line number is displayed twice
- The caret is displayed twice, the second time in an extremely long line

/cc @felixmulder @MasseGuillaume

smarter:
This is temporary until someone figures out how to get sbt to not print
duplicated information when this is turned on. This might require
changes to sbt itself.

/cc @MasseGuillaume


felixmulder:
LGTM

PR #2108, opened 2017-03-16
Make pattern match exhaustive for mergeCompanionDefs.

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -538,7 +538,8 @@ class Namer { typer: Typer =>
             trees.map {
               case mcls @ TypeDef(name, impl: Template) if valid(mcls) =>
                 moduleClsDef.get(name) match {
-                  case Some((stat1, mcls1@TypeDef(_, impl1: Template))) =>
+                  case Some((stat1, mcls1)) =>
+                    assert(mcls1.rhs.isInstanceOf[Template])
```
DarkDimius: It will still throw an exception, the absent warning is misleading.
What did you intend to achive?
If you wanted to remove warning, you should use `(moduleClsDef.get(name): @unchecked)` instead

nicolasstucki: The fact that currently if `mcls1.rhs` is not a `Template` throws match exception is even more misleading. The only difference is that now this is explicitly stated in the code and that it will stop printing a warning for it. Using `@unchecked` would just hide even more the expected precondition on `(moduleClsDef.get(name)`.

odersky: I think I prefer the old version. It's shorter and cleaner. You get the value that did not match so that should be enough to track it down.

nicolasstucki: Then I'll add the @unchecked instead.

odersky: OK, that sounds good.

PR #2111, opened 2017-03-16
Better type inference in harmonizeUnion

smarter:
NOTE: This PR depends on https://github.com/lampepfl/dotty/pull/2110 because the improved type inference revealed the bug fixed by https://github.com/lampepfl/dotty/pull/2110, only the last commit is new.

<hr>

Before this commit, the added testcase failed because the type of `inv`
was inferred to be `Inv[Any]` instead of `Inv[Int]`. The situation looks
like this:
```scala
def inv(cond: Boolean) =
    if (cond)
      new Inv(1) // : Inv[A] where A >: Int
    else
      Inv.empty // : Inv[A'] where A' unconstrained
    // : Inv[A] | Inv[A']
```
To get the type of `inv`, we call `harmonizeUnion` which will take the
lub of `Inv[A]` and `Inv[A']`, eventually this mean that we do:
```
  A' <:< A
```
But since `harmonizeUnion` uses `fluidly`, this does not result in `A'`
getting constrained to be a subtype of `A`, instead we constrain `A` to
the upper bound of `A'`:
```scala
  Any <:< A
```
We use `fluidly` to avoid creating OrTypes in `lub`, but it turns out
that there is a less aggressive solution: `lub` calls `mergeIfSuper`
which then calls `isSubTypeWhenFrozen`, if we just make these subtype
calls non-frozen, we can achieve what we want. This is what the new
method `fluidLub` does.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
+++ b/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
@@ -35,15 +35,6 @@ trait ConstraintHandling {
   /** If the constraint is frozen we cannot add new bounds to the constraint. */
   protected var frozenConstraint = false
 
-  protected var alwaysFluid = false
-
-  /** Perform `op` in a mode where all attempts to set `frozen` to true are ignored */
-  def fluidly[T](op: => T): T = {
-    val saved = alwaysFluid
-    alwaysFluid = true
-    try op finally alwaysFluid = saved
-  }
-
   /** If set, align arguments `S1`, `S2`when taking the glb
    *  `T1 { X = S1 } & T2 { X = S2 }` of a constraint upper bound for some type parameter.
    *  Aligning means computing `S1 =:= S2` which may change the current constraint.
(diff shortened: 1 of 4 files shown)


smarter:
There is a single test that fails with a deep subtype error, it reduces to:
```scala
object Test {
  def test: Unit = {
    Seq(Set(1),Set(2),Set(3))
  }
}
```
By adding more sets to the Seq, you can get it to stackoverflow with deep subtype checking turned off. It doesn't happen on master but type inference is broken there (it is inferred to have type `Seq[Set[Any]]` whereas in this branch it has type `Seq[Set[Int]]` as expected).

@odersky Any idea why this happens? Here's the trace: https://gist.github.com/smarter/9dc1c8952c28886c0758acdff48c7e79

smarter:
I've replaced `fluidLub` by an even more constrained mechanism: a parameter `canConstrain` to `lub`. This is nicer because it doesn't require a global variable and it fixes the deep recursion seen above.

odersky:
Nice work! LGTM

PR #2113, opened 2017-03-16
Move 'invalid super qualifier' error to new error format.

abeln:
As part of https://github.com/lampepfl/dotty/issues/1589, use the new error message for static super references where the qualifier isn't a parent of the class.

Tested:
  Added unit test.

### DIFF
--- a/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
+++ b/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
@@ -304,5 +304,29 @@ class ErrorMessagesTests extends ErrorMessagesTest {
       assertEquals("x", tree.name.show)
     }
 
+  @Test def superQualMustBeParent =
+    checkMessagesAfter("frontend") {
+      """
+        |class A {
+        |  def foo(): Unit = ()
+        |}
+        |
+        |class B {
+        |}
+        |
+        |class C extends A {
+        |  super[B].foo
+        |}
+      """.stripMargin
+    }
+      .expect { (ictx, messages) =>
+        implicit val ctx: Context = ictx
+        val defn = ictx.definitions
 
+        assertMessageCount(1, messages)
+        val SuperQualMustBeParent(qual, cls) :: Nil = messages
+
+        assertEquals("B", qual.show)
+        assertEquals("class C", cls.show)
+      }
 }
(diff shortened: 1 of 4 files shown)


compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -304,5 +304,29 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+      """.stripMargin
+    }
+      .expect { (ictx, messages) =>
+        implicit val ctx: Context = ictx
+        val defn = ictx.definitions
 
+        assertMessageCount(1, messages)
+        val SuperQualMustBeParent(qual, cls) :: Nil = messages
+
+        assert(qual.name.toString == "B", s"qualifier was: ${qual.name}")
```
abeln: I feel like this is a watered down version of what I should be testing (string comparison vs comparing the actual type object), but I'm not familiar enough with the compiler's type hierarchy. Suggestions welcome.

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -304,5 +304,29 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+        assert(cls.denot.name.toString == "C", s"class was: ${cls.denot.name.toString}")
```
abeln: I feel like this is a watered down version of what I should be testing (string comparison vs comparing the actual type object), but I'm not familiar enough with the compiler's type hierarchy. Suggestions welcome.

felixmulder: You'd have to synthesize or request it from the symbol table anyway. I wouldn't worry about it too much. That being said, as aforementioned, using `.show` is preferable here:

```scala
assertEquals("class B", qual.show)
assertEquals("class C", cls.show)
```

Going down a bit further into the rabbit hole - if you feel that doing string comparisons is watered down: you could also request the symbols from the context by doing a lookup à la:

```scala
val classFromCtx = ctx.getClassIfDefined("full.path.to.ClassName".toTypeName)
```

The problem with doing this is that you're depending on the semantics of `getClassIfDefined` to always give you what you want. So first you'd have to do `assert(classFromCtx.exists, "couldn't get class from context")`. Then you'd have to be sure that the class you got from the context actually is the `class C` that you've defined, so you'd have to test that too. Then you'd do your comparison. For simplicity I think it's easier just to compare the shown strings :)

abeln:
@felixmulder PTAL

felixmulder:
LGTM, will merge once the tests pass :tada:

abeln:
Thanks a lot for all the suggestions! Very positive impression of the community (here and in the Gitter channel) :)

PR #2116, opened 2017-03-17
Fix ParamForwarder under non-separate compilation 

smarter:
`inheritedAccessor` looks for a method in a superclass whose name
matches the param accessor name, but that method is created by
ParamForwarder itself, so this only succeeded under separate
compilation (as demonstrated by the existing `paramForwarding_separate`
test). We fix this by simply running `inheritedAccessor` at a phase
where `ParamForwarder` has already been run.

### DIFF
--- a/tests/run/paramForwarding_together.scala
+++ b/tests/run/paramForwarding_together.scala
@@ -1,0 +1,29 @@
+class A(val member: Int) {
+  def getAMember = member
+}
+
+class SubA(member: Int) extends A(member) {
+  def getSubAMember = member
+}
+
+class B(member: Int) extends SubA(member) {
+  def getBMember = member
+}
+
+object Test {
+  def printFields(cls: Class[_]) =
+    println(cls.getDeclaredFields.map(_.toString).sorted.deep.mkString("\n"))
+
+  def main(args: Array[String]): Unit = {
+    val a = new A(10)
+    val subA = new SubA(11)
+    val b = new B(12)
+
+    println("# Fields in A:")
+    printFields(classOf[A])
+    println("# Fields in SubA:")
+    printFields(classOf[SubA])
+    println("# Fields in B:")
+    printFields(classOf[B])
+  }
+}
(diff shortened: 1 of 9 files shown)


tests/run/paramForwarding_together.scala
```diff
@@ -0,0 +1,29 @@
...
+}
+
+class SubA(member: Int) extends A(member) {
+  def getSubAMember = member
+}
+
+class B(member: Int) extends SubA(member) {
+  def getBMember = member
+}
+
```
odersky: I am not yet convinced this fixes it. Can you also add test cases where A, SubA, and B appear in different orders? Like: B, SubA, A, for instance?

smarter: You mean, order of declaration in the class?

smarter:
@odersky Ah, you're right, changing the order of declaration breaks things, I guess that makes sense since `ParamForwarder` is not a mini-phase.

smarter:
We're getting pickling differences like this:
```diff
--- compiler/before-pickling.txt    2017-03-17 17:42:45.365222146 +0100
+++ compiler/after-pickling.txt 2017-03-17 17:42:45.369222105 +0100
@@ -5985,7 +5985,7 @@
             private[this] def statement: dotty.tools.dotc.ast.untpd.ValDef = 
               <
                 <
-                  <<super:Request.this.ValOrPatHandler(ValHandler.super)>@<27630..27630>.statement:
+                  <<super:Request.this.ValOrPatHandler(ValHandler.super)>@<27630..27630>.statement:=> 
                     dotty.tools.dotc.ast.untpd.Tree
                   >@<27630..27630>
                 .asInstanceOf:([X0] => X0)>@<27630..27630>
```
I suspect that what's happening is that we're selecting the `private[this] def statement` in the superclass `ValOrPatHandler` instead of the `val statement` in the super-superclass `StatementHandler`. This is because the fix in https://github.com/lampepfl/dotty/pull/2106 intentionally allows selecting a private member when loading from a new run. @odersky How would you fix this? I think that using shadowed names in `ParamForwarder` would be cleaner than adding yet more custom logic when loading denotations.

smarter:
Superceded by #2124

PR #2118, opened 2017-03-17
Fix #2117: bug in typechecking super prefix with invalid enclosing class

### LINKED ISSUE #2117: Dotty crashes while typechecking invalid super prefix
When typechecking

`
class A {
  C.super.foo()
}
`

the typer throws with the following stacktrace: http://pastebin.com/arjFaB0Z
The issue is an unguarded pattern matching in TypeAssigner: https://github.com/lampepfl/dotty/blob/d0621108bad55f9fc66c1c5ade9a0b7edb3117e7/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala#L294

The fix (I think) is to simply handle the case where the qualifier has type ErrorType and bail out early.

abeln:
When typechecking

`class A {
  C.super.foo()
}`

If C isn't an enclosing class, the compiler was throwing because of an
unguarded pattern match.

Fix the issue by checking for ErrorType.

Tested:
Verified that the example above no longer throws.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
@@ -291,26 +291,28 @@ trait TypeAssigner {
 
   def assignType(tree: untpd.Super, qual: Tree, inConstrCall: Boolean, mixinClass: Symbol = NoSymbol)(implicit ctx: Context) = {
     val mix = tree.mix
-    val qtype @ ThisType(_) = qual.tpe
-    val cls = qtype.cls
-
-    def findMixinSuper(site: Type): Type = site.parents filter (_.name == mix.name) match {
-      case p :: Nil =>
-        p
-      case Nil =>
-        errorType(SuperQualMustBeParent(mix, cls), tree.pos)
-      case p :: q :: _ =>
-        errorType("ambiguous parent class qualifier", tree.pos)
+    qual.tpe match {
+      case err: ErrorType => untpd.cpy.Super(tree)(qual, mix).withType(err)
+      case qtype @ ThisType(_) =>
+        val cls = qtype.cls
+        def findMixinSuper(site: Type): Type = site.parents filter (_.name == mix.name) match {
+          case p :: Nil =>
+            p
+          case Nil =>
+            errorType(SuperQualMustBeParent(mix, cls), tree.pos)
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -291,26 +291,28 @@ trait TypeAssigner {
...
-
-    def findMixinSuper(site: Type): Type = site.parents filter (_.name == mix.name) match {
-      case p :: Nil =>
-        p
-      case Nil =>
-        errorType(SuperQualMustBeParent(mix, cls), tree.pos)
-      case p :: q :: _ =>
-        errorType("ambiguous parent class qualifier", tree.pos)
+    qual.tpe match {
+      case err: ErrorType => untpd.cpy.Super(tree)(qual, mix).withType(err)
```
abeln: Is this necessary, or can I just do `tree.withType(err)`?
In general, when I should I be using the tree copier?

DarkDimius: Tree copier provides structural sharing and is used when you think that the tree might stay the same after your changes.

DarkDimius:
Thanks for the contribution!
> What would be a good place to unit tests for this?

tests/neg/i2117.scala.

Would you also kindly update commit message to "Fix #2117: ..."

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
abeln: Thanks. In this case, would it be ok/better to do `tree.withType(err)`?

abeln:
Added the test. PTAL.
What's the significance of "i" in "i2117"? (i.e. what's the naming convention, given that other files in that directory don't follow the same style).

smarter:
"i" means issue and refers to an issue number on https://github.com/lampepfl/dotty/issues. "t" means ticket and refers to an issue number on scalac bug tracker (because the test was imported from scalac): https://issues.scala-lang.org/secure/Dashboard.jspa

abeln:
Friendly ping @DarkDimius @smarter @felixmulder 
Does this look good or would you like something changed? Thanks!

DarkDimius:
@abeln, thank you for pinging. It's good as is.
LGTM

PR #2120, opened 2017-03-19
Fix dottydoc exception caused by attempting to serialize NonEntity values

vlthr:
Fixes an bug in dottydoc where NonEntity values were accidentally being serialized.

The changes in e7fe9df3c14df0e5ccd88440dbd49a9dce65d48f introduced an IllegalStateException during execution due to an attempt to serialize `NonEntity` values. Previously, NonEntities would incorrectly be serialized (without throwing an exception) as instances of Package (due to the pattern match at https://github.com/lampepfl/dotty/blob/5bb7c976712e8c86898b94d1abea1acedeed3562/doc-tool/src/dotty/tools/dottydoc/model/JavaConverters.scala#L302).

The source of the NonEntity values was the default value of `entity` in DefaultParams.scala (https://github.com/lampepfl/dotty/blob/3e65cff0caf4c671d3cf98da347792c85a4ac2de/doc-tool/src/dotty/tools/dottydoc/staticsite/DefaultParams.scala#L39)

This PR fixes this issue by explicitly checking whether `DefaultParams.entity` contains a `NonEntity`, and if so, omitting it from serialization. Also gets rid of procedure syntax used in JavaConverterTest.scala.

Review by @felixmulder

Style question: do you prefer both of these commits to be squashed into a single one for brevity, or keep them separate since they're independent changes?

### DIFF
--- a/doc-tool/src/dotty/tools/dottydoc/staticsite/DefaultParams.scala
+++ b/doc-tool/src/dotty/tools/dottydoc/staticsite/DefaultParams.scala
@@ -17,27 +17,34 @@ case class DefaultParams(
 ) {
   import model.JavaConverters._
 
-  def toMap: Map[String, AnyRef] = Map(
-    "docs" -> docs,
+  def toMap: Map[String, AnyRef] = {
+    val base = Map(
+      "docs" -> docs,
 
-    "originalDocs" -> originalDocs,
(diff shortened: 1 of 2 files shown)


felixmulder:
Two commits are fine when they solve different things :)

PR #2121, opened 2017-03-19
Refactor lambda types

odersky:
This is a huge refactoring which tries to exploit commonalities between PolyTypes and MethodTypes on the one hand and PolyParams (now called TypeParamRef) and MethodParams (now called TermParamRef) on the other hand. It creates a 2x2 matrix of lambda types where one dimension is its kind (* vs higher kinded) and the other is whether the abstraction is over terms or types. 

One square of the matrix is currently unpopulated: higher-kinded types over term parameters do not exit yet. But given the refactorings they would be easy to add.

There are several benefits achieved by exploiting symmetries:

 - Reduce code duplication
 - Detect and fix subtle bugs about parameter-dependent methods which broke the symmetry before
 - Prepare the way for term-dependent lambdas on the type level
 - Better distinguish PolyTypes (on the * level) from higher kinded type lambdas. In particular, hk type lambdas are naturally type proxies but it's dangerous to treat polytypes as type proxies. 

Based on #2103


odersky:
Here's a high-level summary of what was done:

Splits: `PolyType` becomes `HKTypeLambda` or `PolyType`.

Common base traits:
    
    LambdaType   |   TermLambda      |   TypeLambda
    -------------+-------------------+------------------
    HKLambda     |  (HKTermLambda)   |   HKTypeLambda
    MethodOrPoly |   MethodType	 |   PolyType

Renamings:

    from             to
    --------------------------------
    PolyTypeTree     LambdaTypeTree
    PolyParam        TypeParamRef
    MethodParam      TermParamRef
    TypeParamInfo    ParamInfo
    paramBounds      paramInfo
    paramType        paramInfo

odersky:
Rebased to master

PR #2122, opened 2017-03-20
Fixed #2086: Add tests for issue that has already been fixed.

### LINKED ISSUE #2086: Unsoundness due to missing bound-checking of higher-kinded type parameters
```scala
object Test {
  type Bound[Left, Right] >: Left <: Right
  type Foo = Bound[Any, Nothing]

  val x: Foo = 1
  val y: String = x
}
```
This compiles with dotty but fails in scalac as expected with:
```scala
error: lower bound Left does not conform to upper bound Right
  type Bound[Left,Right] >: Left <: Right
       ^
```

nicolasstucki:


smarter:
When was this fixed?

nicolasstucki:
I'm not sure when it was fixed. I just had a look at this issue and remembered that a test case with `>: Any <: Nothing` was working properly on #2044.

smarter:
Indeed, looks like this has been fixed for at least a year, I don' t understand why it didn't fail when I tried it last week, and I can't reproduce the problem anymore. Weird.

PR #2124, opened 2017-03-20
Fix parameter accessor forwarding

odersky:
Fixes remaining problems of #2116, from which it takes one change and all tests.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1719,7 +1719,7 @@ object Types {
      *  under both private and public names, so it could still be found by looking up
      *  the public name.
      */
-    final def shadowed(implicit ctx: Context): NamedType =
+    def shadowed(implicit ctx: Context): NamedType =
       NamedType(prefix, name.shadowedName)
 
     override def equals(that: Any) = that match {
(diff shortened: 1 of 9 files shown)


compiler/src/dotty/tools/dotc/transform/ParamForwarding.scala
```diff
@@ -66,8 +68,12 @@ class ParamForwarding(thisTransformer: DenotTransformer) {
                     sym.copySymDenotation(initFlags = sym.flags | Method | Stable, info = sym.info.ensureMethodic)
                       .installAfter(thisTransformer)
                     val superAcc =
-                      Super(This(currentClass), tpnme.EMPTY, inConstrCall = false).select(alias)
-                    DefDef(sym, superAcc.ensureConforms(sym.info.widen))
+                      Super(This(currentClass), tpnme.EMPTY, inConstrCall = false)
+                        .select(alias)
+                    val stpe @ TermRef(_, _) = superAcc.tpe
+                    val superAccShadowed = superAcc.withType(stpe.shadowed)
```
smarter: Technically, this is only necessary when the super is not the base class where the paramaccessor is defined, but I guess there's no harm in always using `.shadowed`. However it would be nice to add a comment explaining when the `.shadowed` is useful.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1782,15 +1782,19 @@ object Types {
       else d.atSignature(sig).checkUnique
     }
 
-    override def newLikeThis(prefix: Type)(implicit ctx: Context): TermRef = {
-      val candidate = TermRef.withSig(prefix, name, sig)
+    private def fixDenot(candidate: TermRef, prefix: Type)(implicit ctx: Context): TermRef =
       if (symbol.exists && !candidate.symbol.exists) { // recompute from previous symbol
         val ownSym = symbol
         val newd = asMemberOf(prefix, allowPrivate = ownSym.is(Private))
```
smarter: Should we revert the addition of the `allowPrivate` parameter to `asMemberOf` since it is no longer useful for ParamForwarder and could mask other bugs that only show up under `-Ytest-pickler` ?

odersky: I'd leave it in, because the logic makes sense anyway.

odersky:
@smarter We get a difference with actual output and the check file for paramForwarding. But I don't know which one is correct, or whether it is important. Can you check?

odersky:
Also, I am not sure whether we want to avoid the $$local field for "override val" accessors. What's the point? Specifically, should we add forwarders only to PrivateParamAccessors?

smarter:
Yes, this difference also happened in my PR. I think updating the check file makes sense, the fact that Y does not get a field is coherent with the fact that C does not get a field, which was already the case before this test.

smarter:
> Also, I am not sure whether we want to avoid the $$local field for "override val" accessors. What's the point? 

Isn't it always good to avoid adding fields if possible?

odersky:
> Isn't it always good to avoid adding fields if possible?

The point is that "override val" seems non-sensical since the val is "overridden" with the same value it has already. But I agree it doe snot matter either way.

smarter:
Indeed, but override val also allows you to refine the type of the val

PR #2125, opened 2017-03-20
Operation Kill Partest (part 1)

felixmulder:
Partest is dead, long live [name for new testing "framework"]!


smarter:
Awesome! I think this is missing one great feature from partest: A good summary report. When I scroll to the end of the logs I should be able to see at a glance:
- What test methods failed
- In each failed test method, what files failed to compile/run
- The exact command-line invocation I should use to run this test manually

A related feature is log files, which partest also outputs (but delete for non-failed tests), when the compiler output is really huge, it's very helpful to be able to navigate a log file instead of scrolling for hours.  Also related: partest keeps output directories for failed run tests, this makes it easy to inspect them with javap or a decompiler to figure out what's going on.

smarter:
Also, I know this was already the case but it's really bugging me now: can we fix `dotty.tools.dottydoc.TestWhitelistedCollections` so that its output does not fill most of the log? I'm talking about all the:
```scala
Compiling (1/546): Platform.scala
Compiling (2/546): ScalaBeanInfo.scala
Compiling (3/546): BeanInfoSkip.scala
...
```

smarter:
Basically all tests that instantiate a compiler should use a custom reporter that writes things to a logfile, that logfile should be kept only if specified by the user or if the test failed.

felixmulder:
> Awesome! I think this is missing one great feature from partest: A good summary report. When I scroll to the end of the logs I should be able to see at a glance:

Sure

> Also, I know this was already the case but it's really bugging me now: can we fix `dotty.tools.dottydoc.TestWhitelistedCollections` so that its output does not fill most of the log? I'm talking about all the:

Ah huh

> Basically all tests that instantiate a compiler should use a custom reporter that writes things to a logfile, that logfile should be kept only if specified by the user or if the test failed.

Mhm

All good ideas, I'll get them in tomorrow.

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -0,0 +1,302 @@
...
+    ) +
+    compileList(
+      "testIssue34",
+      List(
+        "../compiler/src/dotty/tools/dotc/config/Properties.scala",
+        "../compiler/src/dotty/tools/dotc/config/PathResolver.scala"
+      ),
+      defaultOptions.and("-Xprompt")
+    )
+  }.times(2).pos()
```
smarter: Did you try making a `-Ytimes` option instead, as we discussed recently?

felixmulder: No - let's discuss this PM

DarkDimius:
@felixmulder, is this PR ready for review?

felixmulder:
@DarkDimius - yes mostly, I have two things left to add before I want this merged:

- [ ] being able to run a single test / or set of files
- [ ] testing output of meta tests

maybe the first one will affect the implementation somewhat, but the second one most likely won't. If you want to review it now that would be great, but you can also wait until I've fixed the first point. Up to you 🙂

felixmulder:
@DarkDimius - done with point 1

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+          s"[=======================================] compiled ($totalTargets/$totalTargets, " +
+          s"${(System.currentTimeMillis - start) / 1000}s)  "
+        )
+      }
+    }
+
+    /** Wrapper function to make sure that the compiler itself did not crash -
+     *  if it did, the test should automatically fail.
+     */
+    protected def compileTry(op: => Unit): Unit =
```
DarkDimius: tryCompile

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+  private final case class SeparateCompilationTarget(
+    dir: JFile,
+    flags: Array[String],
+    outDir: JFile
+  ) extends Target {
+
+    /** Get the files grouped by `_X` as a list of groups, files missing this
+     *  suffix will be put into the same group
+     */
+    def compilationUnits: List[Array[JFile]] =
```
DarkDimius: Given the documentaion, I'll call this compilationGroups.
Otherwise you're re-defining a term already defined in Dotty. Every JFile here is a already a single compilationUnit.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+  /** Each `Test` takes the `targets` and performs the compilation and assertions
+   *  according to the implementing class "neg", "run" or "pos".
+   */
+  private abstract class Test(targets: List[Target], times: Int, threadLimit: Option[Int], suppressAllOutput: Boolean) {
+
+    /** Actual compilation run logic, the test behaviour is defined here */
+    protected def compilationRunnable(target: Target): Runnable
+
+    /** All targets left after filtering out */
+    private val allTargets =
```
DarkDimius: filteredTargets, not allTargets

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      if (!filter.isDefined) targets
+      else targets.filter {
+        case ConcurrentCompilationTarget(files, _, _) =>
+          files.exists(file => filter.get.findFirstIn(file.getAbsolutePath).isDefined)
+        case SeparateCompilationTarget(dir, _, _) =>
+          filter.get.findFirstIn(dir.getAbsolutePath).isDefined
+      }
+
+    /** Total amount of targets being compiled by this test */
+    val totalTargets = allTargets.length
```
DarkDimius: targetsCount

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+
+    private[this] var _errors =  0
+    def errors: Int = synchronized { _errors }
```
DarkDimius: errorCount

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+
+    private[this] var _targetsCompiled = 0
+    private def targetsCompiled: Int = synchronized { _targetsCompiled }
+
+    /** Complete the current compilation with the amount of errors encountered */
+    protected final def completeCompilation(errors: Int) = synchronized {
```
DarkDimius: registerCompilation

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      _errors += errors
+    }
+
+    private[this] var _failed = false
+    /** Fail the current test */
+    protected[this] final def fail(): Unit = synchronized { _failed = true }
+    def didFail: Boolean = _failed
+
+    /** Instructions on how to reproduce failed target compilations */
+    private[this] val failureInstructions = mutable.ArrayBuffer.empty[String]
```
DarkDimius: reproduceInstructions

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+    protected final def addFailureInstruction(ins: String): Unit =
+      synchronized { failureInstructions.append(ins) }
+
+    /** The targets that failed according to the implementing subclass */
+    private[this] val failedCompilationTargets = mutable.ArrayBuffer.empty[String]
```
DarkDimius: mis-usage of the term "target" makes it really hard to understand what do those methods do

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+          addFailedCompilationTarget(dir.getAbsolutePath)
+          fail()
+      }
+
+    /** Prints to `System.err` if we're not suppressing all output */
+    protected def echo(msg: String): Unit =
+      if (!suppressAllOutput) System.err.println(msg)
+
+    /** A single `Runnable` that prints a progress bar for the curent `Test` */
+    private def statusRunner: Runnable = new Runnable {
```
DarkDimius: spawnArbiterRunnable / spawnMonitoringRunnable

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      val javaFiles = files.filter(_.getName.endsWith(".java")).map(_.getAbsolutePath)
+      val javaCompiledBefore = compileWithJavac(javaFiles)
+
+      // Then we compile the scala files:
+      val reporter = TestReporter.parallelReporter(this, logLevel =
+        if (suppressErrors || suppressAllOutput) ERROR + 1 else ERROR)
+      val driver =
+        if (times == 1) new Driver { def newCompiler(implicit ctx: Context) = new Compiler }
+        else new Driver {
+          def newCompiler(implicit ctx: Context) = new Compiler
```
DarkDimius: Can't you use the "bench" runner here instead?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+              val rep = super.doCompile(comp, files)
+              ctx.echo(s"\ntime run $run: ${(System.nanoTime - start) / 1000000}ms")
+              rep
+            }
+        }
+
+      val allArgs = addOutDir(flags)
+      driver.process(allArgs ++ files.map(_.getAbsolutePath), reporter = reporter)
+
+      // If the java files failed compilation before, we try again after:
```
DarkDimius: >we try again after

I didn't get what does this comment mean and where is "try again" happening.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      if (!javaCompiledBefore)
+        assert(compileWithJavac(javaFiles), s"java compilation failed for ${javaFiles.mkString(", ")}")
+
+      reporter
+    }
+
+    private[ParallelTesting] def execute(): this.type = {
```
DarkDimius: executeTestSuite

DarkDimius: otherwise it may look like running compiled artifacts of run-test.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        regex
+          .map(r => s"""No files matched regex "$r" in test""")
+          .getOrElse("No tests available under target - erroneous test?")
+      }
+
+      this
+    }
+  }
+
+  @inline private final def isCompilable(f: JFile): Boolean = {
```
DarkDimius: Don't use `@inline`

DarkDimius: isSourceFile

DarkDimius: would be also nice to move it companion object. It doesn't need any of the state of this class.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      def run(): Unit = compileTry {
+        target match {
+          case ConcurrentCompilationTarget(files, flags, outDir) => {
+            val sourceFiles = files.filter(isCompilable)
+            val reporter = compile(sourceFiles, flags, false, outDir)
+            completeCompilation(reporter.errorCount)
+
+            if (reporter.errorCount > 0) {
+              fail()
+              val errorMsg = target.buildInstructions(reporter.errorCount, reporter.warningCount)
```
DarkDimius: Those lines are duplicated 6 times in various places. Extract to a helper method?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+          val buildInstr = target.buildInstructions(0, warnings)
+          addFailureInstruction(buildInstr)
+
+          // Fail target:
+          failTarget(target)
+        }
+    }
+    catch {
+      case _: NoSuchMethodException =>
+        echo(s"\ntest in '$dir' did not contain a main method")
```
DarkDimius: what if it didn't contain another method because of compiler emitting wrong code?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        fail()
+
+      case _: ClassNotFoundException =>
+        echo(s"\ntest in '$dir' did was not contained within a `Test` object")
```
DarkDimius: Same here.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        fail()
+
+      case _: InvocationTargetException =>
+        echo(s"\nTest in '$dir' might be using args(X) where X > 0")
+        fail()
+    }
```
DarkDimius: What about other exceptions that VM may throw when compiler generates invalid code?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+
+    protected def compilationRunnable(target: Target): Runnable = new Runnable {
+      def run(): Unit = compileTry {
+        val (errorCount, warningCount, hasCheckFile, doVerify) = target match {
+          case ConcurrentCompilationTarget(files, flags, outDir) => {
+            val sourceFiles = files.filter(isCompilable)
+            val checkFile = files.flatMap { file =>
+              if (file.isDirectory) Nil
```
DarkDimius: why can't we have check-files for directories that contain tests?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      case _: InvocationTargetException =>
+        echo(s"\nTest in '$dir' might be using args(X) where X > 0")
+        fail()
+    }
+
+    protected def compilationRunnable(target: Target): Runnable = new Runnable {
+      def run(): Unit = compileTry {
+        val (errorCount, warningCount, hasCheckFile, doVerify) = target match {
+          case ConcurrentCompilationTarget(files, flags, outDir) => {
+            val sourceFiles = files.filter(isCompilable)
```
DarkDimius: `files.filter(isCompilable)` is repeated a lot of times. Create a helper method for it in `Target`?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        fail()
+
+      case _: InvocationTargetException =>
+        echo(s"\nTest in '$dir' might be using args(X) where X > 0")
+        fail()
+    }
+
+    protected def compilationRunnable(target: Target): Runnable = new Runnable {
+      def run(): Unit = compileTry {
+        val (errorCount, warningCount, hasCheckFile, doVerify) = target match {
```
DarkDimius: doVerify -> verifier.
And also please make it have an explicit type to indicate that that's a function. Based on a name now I would expect it to be a boolean

compiler/test/dotty/tools/dotc/reporting/TestReporter.scala
```diff
@@ -41,11 +60,56 @@ with UniqueMessagePositions with HideNonSensicalMessages {
...
 }
+
+object TestReporter {
+  private[this] val logWriter = {
+    val df = new SimpleDateFormat("yyyy-MM-dd-HH:mm")
+    val timestamp = df.format(new Date)
+    new PrintWriter(new FileOutputStream(new JFile(s"../tests-$timestamp.log"), true))
+  }
+
+  def parallelReporter(caller: AnyRef, logLevel: Int): TestReporter = new TestReporter(
```
DarkDimius: caller -> monitor or caller -> lock

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+              reporter.echo(errorMsg)
+              reporter.flushToFile()
+            }
+
+            completeCompilation(reporter.errorCount)
+            (reporter.errorCount, reporter.warningCount, checkFile.isDefined, () => verifyOutput(checkFile.get, outDir, target, reporter.warningCount))
+          }
+
+          case target @ SeparateCompilationTarget(dir, flags, outDir) => {
+            val checkFile = new JFile(dir.getAbsolutePath.reverse.dropWhile(_ == '/').reverse + ".check")
```
DarkDimius: For separate compilation folders we can have ".check" files, but we can't have them for joint compilation. Why?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+  extends Test(targets, times, threadLimit, suppressAllOutput) {
+    protected def compilationRunnable(target: Target): Runnable = new Runnable {
+      def run(): Unit = compileTry {
+        // In neg-tests we allow two types of error annotations,
+        // "nopos-error" which doesn't care about position and "error" which
+        // has to be annotated on the correct line number.
+        //
+        // We collect these in a map `"file:row" -> numberOfErrors`, for
+        // nopos errors we save them in `"file" -> numberOfNoPosErrors`
+        def errorMapAndExpected(files: Array[JFile]): (HashMap[String, Integer], Int) = {
```
DarkDimius: collectErrorMapAndExpectedCount.
Also, using strings as keys in map isn't the best idea. VM isn't prohibited from re-using strings unless you explicitly call `new`.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+              }
+
+              expectedErrors += noposErrors + errors
+            }
+          }
+
+          (errorMap, expectedErrors)
+        }
+
+        def getMissingAnnotations(errorMap: HashMap[String, Integer], reporterErrors: Iterator[MessageContainer]) = !reporterErrors.forall { error =>
```
DarkDimius: Those aren't annotation. getMissingErrorIndicators?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+          val getter = if (error.pos.exists) {
```
DarkDimius: why is it called getter?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        }
+        else if (hasMissingAnnotations()) {
+          echo {
+            s"\nErrors found on incorrect row numbers when compiling $target"
+          }
+          failTarget(target)
+        }
+        else if (!errorMap.isEmpty) {
+          echo {
+            s"\nError annotation(s) have {<error position>=<unreported error>}: $errorMap"
```
DarkDimius: annotation - mis-use of term

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      require(other.shouldDelete == shouldDelete, "can't combine tests that differ on deleting output")
+      require(other.shouldFail == shouldFail, "can't combine tests that have different expectations on outcome")
+      new CompilationTest(targets ++ other.targets, times, shouldDelete, threadLimit, shouldFail)
+    }
+
+    /** Creates a "pos" test run, which makes sure that all tests pass
+     *  compilation without generating errors and that they do not crash the
+     *  compiler
+     */
+    def pos(): this.type = {
```
DarkDimius: pos is a very bad name for such an important method.
I'd call it "checkCompiles"

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+
+      cleanup()
+    }
+
+    /** Creates a "run" test run, which is a superset of "pos". In addition to
+     *  making sure that all tests pass compilation and that they do not crash
+     *  the compiler; it also makes sure that all tests can run with the
+     *  expected output
+     */
+    def run(): this.type = {
```
DarkDimius: "checkRuns"

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      }
+
+      cleanup()
+    }
+
+    /** Creates a "neg" test run, which makes sure that each test generates the
+     *  correct amount of errors at the correct positions. It also makes sure
+     *  that none of these tests crash the compiler
+     */
+    def neg(): this.type = {
```
DarkDimius: checkFails

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+    /** Builds a `CompilationTest` which performs the compilation `i` times on
+     *  each target
+     */
+    def times(i: Int): CompilationTest =
+      new CompilationTest(targets, i, shouldDelete, threadLimit, shouldFail)
+
+    /** Builds a `Compilationtest` which passes the verbose flag and logs the
+     *  classpath
+     */
+    def verbose: CompilationTest = new CompilationTest(
```
DarkDimius: enableVerbose

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -0,0 +1,308 @@
...
+    compileDir("../compiler/src/dotty/tools/dotc/parsing", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/printing", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/repl", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/rewrite", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/transform", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/typer", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/dotc/util", picklingOptions) +
+    compileDir("../compiler/src/dotty/tools/io", picklingOptions) +
+    compileFile("../tests/pos/pickleinf.scala", picklingOptions)
+  }.limitThreads(4).pos()
```
DarkDimius: Why are thread-limits hardcoded in those tests?
Needs an explaining comment

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+
+    /** Builds a `CompilationTest` where the executed test is expected to fail
+     *
+     *  This behaviour is mainly needed for the tests that test the test suite.
+     */
+    def expectFailure: CompilationTest =
+      new CompilationTest(targets, times, shouldDelete, threadLimit, true)
+
+    /** Delete all output files generated by this `CompilationTest` */
+    def delete(): Unit = targets.foreach(t => delete(t.outDir))
```
DarkDimius: duplicates `cleanup`.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+      if (file.isDirectory) file.listFiles.foreach(delete)
+      try Files.delete(file.toPath)
+      catch {
+        case _: NoSuchFileException => // already deleted, everything's fine
+      }
+    }
+  }
+
+  /** Create out directory for directory `d` */
+  private def toCompilerDirFromDir(d: JFile, sourceDir: JFile, outDir: String): JFile = {
```
DarkDimius: what do you mean by CompilerDir? SourceDir?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+    val uniqueSubdir = file.getName.substring(0, file.getName.lastIndexOf('.'))
+    val targetDir = new JFile(outDir + s"${sourceDir.getName}/$uniqueSubdir")
+    targetDir.mkdirs()
+    targetDir
+  }
+
+  /** Make sure that directory string is as expected */
+  private def checkRequirements(f: String, sourceDir: JFile, outDir: String): Unit = {
+    require(sourceDir.isDirectory && sourceDir.exists, "passed non-directory to `compileFilesInDir`")
+    require(outDir.last == '/', "please specify an `outDir` with a trailing slash")
```
DarkDimius: windows? :-)? Why not take a JFile and test if it's an already existing directory using the API and create?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+   *  (bridges) - and then find the `@Test` method in our extending class
+   */
+  private def getCallingMethod(): String = {
+    val seen = mutable.Set.empty[String]
+    Thread.currentThread.getStackTrace
+      .filter { elem =>
+        if (seen.contains(elem.getMethodName)) false
+        else { seen += elem.getMethodName; true }
+      }
+      .take(6).find { elem =>
```
DarkDimius: why 6?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+        val callingClass = Class.forName(elem.getClassName)
+        classOf[ParallelTesting].isAssignableFrom(callingClass) &&
+        elem.getFileName != "ParallelTesting.scala"
+      }
+      .map(_.getMethodName)
+      .getOrElse {
+        throw new IllegalStateException("Unable to reflectively find calling method")
```
DarkDimius: why should this stop tests from being run? 
It would be nice to report an error but continue test execution

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+    val targets =
+      files.map(f => ConcurrentCompilationTarget(Array(f), flags, toCompilerDirFromFile(f, sourceDir, outDir))) ++
+      dirs.map(dir => SeparateCompilationTarget(dir, flags, toCompilerDirFromDir(dir, sourceDir, outDir)))
+
+    // Create a CompilationTest and let the user decide whether to execute a pos or a neg test
+    new CompilationTest(targets)
+  }
+
+  /** This function behaves similar to `compileFilesInDir` but it ignores
+   *  sub-directories and as such, does **not** perform separate compilation
```
DarkDimius: >does **not** perform separate compilation

Why?

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -0,0 +1,308 @@
...
+  def interactive: Boolean = !sys.env.contains("DRONE")
+
+  def regex: Option[String] = sys.props.get("dotty.partest.filter")
+
+  // Positive tests ------------------------------------------------------------
+
+  @Test def compilePos: Unit = {
+    compileList("compileStdLib", StdLibSources.whitelisted, scala2Mode.and("-migration", "-Yno-inline")) +
+    compileFilesInDir("../tests/pos", defaultOptions)
+  }.pos()
```
DarkDimius: >.pos

It's a good style to name methods as actions.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
smarter: Magic number?

felixmulder: sort of magic, we shouldn't go through more than a few levels of indirection... but maybe I should take out the magic number

DarkDimius:
It would also be nice to give a test run for this infrastucture at least once on an architecture without all-or-nothing memory publishing approach. It may unveil concurrency bugs. We used to have a SPARC Niagara machine at lab.

.drone.yml
```diff
@@ -6,7 +6,7 @@ pipeline:
       - ln -s /var/cache/drone/scala-scala scala-scala
       - ln -s /var/cache/drone/ivy2 "$HOME/.ivy2"
       - ./scripts/update-scala-library
-      - sbt -J-Xmx4096m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m -Ddotty.drone.mem=4096m "${TEST}"
+      - sbt -J-Xmx24G -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=4G -Ddotty.drone.mem=4096m "${TEST}"
```
felixmulder: Regarding metaspace - no idea, here I'm just playing with different settings to see if I can get the CI to run faster. Let's discuss this on Thursday at the meeting after I've experimented (=

`dotty.drone.mem` is something that partest (the old one) uses IIRC.

.drone.yml
```diff
@@ -34,7 +34,9 @@ pipeline:
 
 matrix:
   TEST:
-    - ;test;dotty-bin-tests/test
-    - ;publishLocal;dotty-bootstrapped/test
-    - partest-only-no-bootstrap --show-diff --verbose
-    - partest-only --show-diff --verbose
+    - ;set testOptions in LocalProject("dotty-compiler") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.ParallelTesting") ;test ;dotty-bin-tests/test
```
felixmulder: True, also if it wasn't clear - they will be removed once we remove partest itself.

compiler/test/dotty/tools/dotc/CompilationTests.scala
felixmulder: I agree - but was told that `.pos()` should be used since it performs side-effects - WDYT?

.drone.yml
smarter: `dotty.drone.mem` is used in `Build.scala`, it's the amount of memory we give to each jvm started from sbt.

compiler/test/dotty/tools/dotc/CompilationTests.scala
DarkDimius: I didn't mean to drop `()`, I've meant to give it another name that indicate that it's going to actually run the test. It would be also nice to make it not clash with `pos`ition.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -0,0 +1,1076 @@
...
+   *  E.g: `*`is not a regular expression, but `.*` is.
+   */
+  def regex: Option[String]
+
+  /** The `regex` string gets turned into an actual regular expression used for
+   *  filtering tests
+   */
+  private lazy val filter: Option[Regex] = regex.map(str => new Regex(str))
+
+  /** A compilation target whose files or directory of files is to be compiled
```
felixmulder: It's the target of the test - i.e. the files or directory currently being compiled. Do you have a constructive suggestion?

`TestTarget`?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: remnant from containing a list of errors - will change

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: TestSources?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Imagine this:

```java
// Foo.java
pacakge foo;
public class Foo extends Bar {}
```

```scala
// Bar.scala
package foo
abstract class Bar
```

In this case, first compiling the java files and then the Scala files will fail. So we need to retry javac once if the initial compilation failed. If the reference is the other way around we first need to compile the Java files and then the Scala files anyway.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Two-fold: the `Bench` uses `println`, and previously @smarter suggested that we implement `-Ytimes:X` as a real flag.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: AFAIK all the tooling first compiles Scala files and passes java files as sources to our compiler(this is why we have java parser in first place).

The proper way to test it would be in 3 runs without retries:
 - pass .java & .scala source files to Dotty
 - pass .java files to javac, with results of .scala compilation on classpath
 - pass .scala files to Dotty, with results of .java compilation on classpath.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Why not? The heuristics of 2.11 (and 2.12) are satisfied AFAICT and in dotty it will be enforced.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
smarter: No, we should only need to do `dotc *.java *.scala` then `javac *.java` just likes partest now does: https://github.com/scala/scala-partest/commit/1c263205b8e1b16458073814dc61888118905acb

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Well then `test in '$dir' did not contain a main method` is still the truth right?

Not sure how you're proposing it to be checked otherwise.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Same here.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
smarter: Although to cover even more ground than partest, we could have a third round of `dotc *.scala` and check that the output matches the output of the first round (otherwise there's a difference between reading Java from source files and reading Java from classfiles).

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Not sure what exceptions are thrown in this case - or if it just segfaults...

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Because by convention, directories within directories compiled with `compileFilesInDir` are treated as `SeparateCompilationTarget`s. So it is taken care of by the other `case`

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: See comment above

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: because it gets from the map :joy:

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: 0. This is not a hot-spot.
1. it rarely helps, as you're trying to out-smart the VM. In many cases you end up with more bytecode and lesser binary compatibility guarantees, but the same performance. Sometimes performance can be even worse.
http://stackoverflow.com/questions/2709095/does-the-inline-annotation-in-scala-really-help-performance
In your case it won't as the method that calls this one will be a simple forwarder and you're not making it faster as VM will inline it if it's worth it.
2. My opinion on `@inline` is that it should be only used for performance critical stuff after you've looked at the bytecode to see that it helps. In this case it doesn't.
3. `@inline private final def` is too much noise. E.g. Private already means final. `@inline here wouldn't make it faster, but may make it bigger`

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Indicators is also the wrong term - annotations is the generic term, `@-annotations` are specific to what you're referring to.

Hmm, I wonder what synonym to use for this - perhaps `getMissingExpectedErrors`?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: => `"Expected error(s)..."

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: It shouldn't segfault and in most tests that I suggest to add it will not.
Test suite should be able to survive all of it.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: our convention has thus far been to call these tests "pos tests" - do we want to let go of this nomenclature?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: `checkFails` would be misleading, since we have `expectFailure` to test if any kind of test run fails.

Perhaps `checkErrors`?

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: No - cleanup is conditional, delete is explicit.

See `tastyBootstrap` test in `CompilationTests.scala`

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: The test is still a pos-test in our nomenclature. And the class is still called PosTest.
But the main step the pos-test does is `checkCompiles`.

`Thread` isn't called runner, thought the main think it does is calling `run`.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: This method simply creates the out directory - I'll rename both of these

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: I retract this comment. It not only should fail, but fail in a specific way...
I don't have a better name here now.

DarkDimius: checkErrors?... naaah

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: That's what I'm saying: the conditional should check condition and call unconditional one instead of re-implementing it.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: `java.io.File` internally handles forward slash translation if the platform is Windows.

I could provide more of an API for this, but felt it was approaching over-engineering...

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: for the lulz - no but will remove

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Most likely it will screw up tests that keep output for compilation that depends on previous compilation - e.g. `tastyBootstrap`.

But yeah - sure, I could make it continue with some randomized value instead.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: Because we have tests that compile things on a single depth to weed out where an error ocurred - see `tastyBootstrap`

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: > checkErrors?... naaah

This was written before I saw your suggestion.
As we've both suggested the same name, I propose we go with it.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: >check that the output matches the output of the first round

While it would be nice to do this, there are two reasons why I propose not to add this now:
 - first we need to check that we generate same bytecode even in absence of java-files. Currently we don't check it
 - the check will fail. Dotty uses random number generator to renerate UUID of Tasty. We should use some stable-hash of tasty instead.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: It may have been thrown because some other method is missing, not `main`.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
DarkDimius: Some other class needed for execution of the test may be missing, though `Test` object exists.

compiler/test/dotty/tools/dotc/ParallelTesting.scala
felixmulder: `checkExpectedErrors` it is!

felixmulder:
Regarding the proposed tests in your summary:

When we implement separate VMs for test running, then we should be solving most of these issues. So as discussed during the meeting that will be part two of "Operation kill partest" (I'm renaming this PR).

DarkDimius:
> then we should be solving most of these issues.

I don't really think so. Only 4 out of >10 tests I suggest need a separate VM as a proper fix. Others don't.

felixmulder:
@DarkDimius, below is my naive characterisation. The upper list won't be a part of this PR, the bottom one I'd like to get to, the latter should have proposed solutions. Currently, the last I'm not sure how to solve.

Please edit the list in this post.

Separate VM
-------------
* test that allocates memory forever;
* test that has infinite (tail) recursion;
* test that has infinite non-tail recursion;
* test that deadlocks;
* test that contains a fork-bomb

Can be solved in current implementation
-----------------------------------------
* test that throws verification error when classloading a broken class;
    * [x] Throws and prints stack trace
* tests that contains classes that will throw AbstractMethodError, NoClassDefFoundError, ClassFormatError, IncompatibleClassChangeError, IllegalAccessError(not by throwing it explicitly, but actually having a broken class)
    * [x] Throws and prints stack trace
* test that re-defines classes in stdlib that this CI relies on with broken versions;
    * [ ] ???

DarkDimius:
> test that has infinite non-tail recursion;

this isn't something that you need a separate VM for. It's just a non-fatal exception thrown from the test.

>test that re-defines classes in stdlib that this CI relies on with broken versions; 

This checks that you correctly set-up classloaders. It should be working already but it would be nice to check it.

felixmulder:
> > test that has infinite non-tail recursion;

> this isn't something that you need a separate VM for. It's just a non-fatal exception thrown from the test

wat? This isn't something that throws: `while (true) println("screw you compiler devs")`

> > test that re-defines classes in stdlib that this CI relies on with broken versions;

> This checks that you correctly set-up classloaders. It should be working already but it would be nice to check it.

How do you propose to do so?

DarkDimius:
> wat? This isn't something that throws: while (true) println("screw you compiler devs")

this isn't non-tail recursion either :-).
non-tail recursion will be here:
```
def bar: Int = foo
def foo: Int = bar
```

>How do you propose to do so?

create a test that re-defines `scala.collection.immutable.List` with different behaviour and tests that it is actually different

felixmulder:
ah, now I get how to read that sentence :joy:

I should get checked for dysrexiah.

felixmulder:
@DarkDimius - the redefinition thing is very hard to solve with the current strategy for running programs. Since we load something that already exists in the hierarchy, it doesn't get overridden and as such we get a `NoSuchMethodError`.

I've amended the rest of your proposed changes and added a test for non-tail-recursive recursive code (as you already noted).

compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -406,56 +406,63 @@ trait ParallelTesting {
...
         val meth = cls.getMethod("main", classOf[Array[String]])
 
-        val printStream = new ByteArrayOutputStream
         Console.withOut(printStream) {
           meth.invoke(null, Array("jvm")) // partest passes at least "jvm" as an arg
         }
+      }
+      catch {
+        case ex: NoSuchMethodException =>
+          echo(s"test in '$dir' did not contain method: ${ex.getMessage}      ")
```
DarkDimius: It would be also nice to keep the stack trace. 
In case method was expected to be a forwarder it's good to know from where is it called.
Similar in other cases.

felixmulder: Fixed in f2adf21

DarkDimius:
>Since we (class) load something that already exists in the hierarchy, it doesn't get overridden and as such we get a NoSuchMethodError.

If it's true, sounds like a bug. We should always prefer source path to classpath.

smarter:
> >Since we (class) load something that already exists in the hierarchy, it doesn't get overridden and as such we get a NoSuchMethodError.
> 
> If it's true, sounds like a bug. We should always prefer source path to classpath.

If the class `List` has already been loaded there is no reason to ask the ClassLoader for anything, so even if the current ClassLoader starts with the source path, it won't get used.

DarkDimius:
>If the class List has already been loaded there is no reason to ask the ClassLoader for anything, so even if the current ClassLoader starts with the source path, it won't get used.

@smarter, I do not understand what you mean. Classloaders don't have source paths. Compiler does. If you currently compile a class, compiler should prefer it to an (old) copy available on the classpath.

odersky:
Currently there is a check of creation dates. The compiler always picks the file that's newer. It has been done like this since the days of javac and I think it's a good strategy to keep.

smarter:
@DarkDimius Compiling is not the problem, the problem is that we use the same JVM to classload the compiler output for run tests, so if you classload `Test` and `Test` tries to use your custom implementation of `List`, it will fail because an existing `List` has already been classloaded.

DarkDimius:
@odersky, yes, that what I've meant by an old copy. Didn't know javac did the same.

DarkDimius:
@smarter, JVM doesn't have global notion of classloading. It has class loading - per class loader.
The fact that another classloader in the same VM has already classloaded a class is unrelated.

By setting up classloaders correctly you should be able to get behavior that custom implementation of the List will be preferred.
Here's an example of how this is done. http://stackoverflow.com/questions/5445511/how-do-i-create-a-parent-last-child-first-classloader-in-java-or-how-to-overr

smarter:
@DarkDimius Yes you can do really crazy things if you try hard enough, but I don't think it's worth it since we'll go to multiple JVM in the future. Also consider that currently the standard collection is on the bootclasspath, so I'm not even sure if you can override that.

DarkDimius:
@smarter, I wouldn't consider 16 lines of implementation code of java as "crazy things if you try hard enough". This is a lot simpler than things we do everyday here.

I don't think this is going to be fixed by going multiple JVM approach if you'll reuse the VMs.
For the same reason: you'll need to have separate classpath for every test. All of those tests share stdlib, but have classes that may take priority over stdlib.

.drone.yml
DarkDimius: Is there an intention to remove partest before we merge this?

smarter:
I don't think we'll need a custom ClassLoader implementation if we use multiple JVMs even if they're persistent. We can just launch the JVM with nothing from scala in the classpath, and everytime we want to run a test we ask the JVM to create a new URLClassLoader with the complete classpath we want to use and classload `Test` using it.

felixmulder:
Can we agree to put this classloading business into part 2 of killing partest? 🙂

DarkDimius:
Given that it became such a contentious topic - yes.
There's one more question left: https://github.com/lampepfl/dotty/pull/2125#discussion_r108652280

.drone.yml
felixmulder: No

felixmulder:
Let's remove partest after having this run for about a week or so alongside the new suite.

DarkDimius:
I agree, but doesn't this pr remove partest from `drone.yml`?

.drone.yml
```diff
@@ -34,7 +34,9 @@ pipeline:
...
+    - ;set testOptions in LocalProject("dotty-compiler-bootstrapped") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.ParallelTesting") ;publishLocal ;dotty-bootstrapped/test
+    - ;set testOptions in LocalProject("dotty-compiler") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.ParallelTesting") ;partest-only-no-bootstrap --show-diff --verbose
+    - ;set testOptions in LocalProject("dotty-compiler-bootstrapped") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.ParallelTesting") ;partest-only --show-diff --verbose
```
felixmulder: @DarkDimius - here you can see that partest is still included, with the new unit tests filtered out.

Once we want to remove partest from the CI these lines dissappear and the only two left will be:

```yaml
- ;test ;dotty-bin-tests/test
- ;publishLocal ;dotty-bootstrapped/test
```

DarkDimius: @felixmulder, thank you for clarification.

felixmulder:
Alright - before we merge this. Can we decide upon a strategy for the memory settings?

I have not been able to see much of a difference when altering them from the current settings. Especially now that partest is kept.

DarkDimius:
For metaspace, the current change I believe is unnecessary.
@felixmulder, I propose we postpone this decision till we decide what to do with classloaders.
The approach proposed by @smarter requires substantially more metaspace & code cache memory for JIT. I won't be surprised if it will simply disable some parts of JIT due to too aggressive classloading.  

I would expect my proposed approach to be able to fit in <1gb of metaspace and <256mb of code-cache with a lot of spare room.

PR #2126, opened 2017-03-20
Remove the collection strawmans from the compiler sources

smarter:
The development of the strawman has now moved to
https://github.com/scala/collection-strawman .


odersky:
The strawman is used in some of the tests, and it would be good to keep these for monitoring possible regression. We could move it into the test directory, however.

smarter:
Indeed, I thought all the tests contained a copy of the strawmans, but not CollectionTests.scala, I'll move what's need to the test directory.

nicolasstucki:
We should probably have a git submodule with the current collection strawman. We can the use this one for our tests. While trying this out  for the callgraph I found #2188.

felixmulder:
Superseded by #2427

PR #2128, opened 2017-03-21
Delay name mangling

odersky:
This PR aims to delay name mangling until the point where we emit code. Before, names should be "semantic", i.e. the kind of a name is prepresented by its structure instead of its characters.


compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -167,35 +153,107 @@ object Names {
...
 
     override protected[this] def newBuilder: Builder[Char, Name] =
       termNameBuilder.mapResult(_.toTypeName)
+  }
 
-    def fromChars(cs: Array[Char], offset: Int, len: Int): TypeName = typeName(cs, offset, len)
+  /** A term name that's derived from an `underlying` name and that
+   *  adds `info` to it.
+   */
+  class DerivedTermName(override val underlying: TermName, override val info: NameInfo)
```
smarter: Instead of having a `NameInfo` object as parameter, what about a `NameFlagSet` ?

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -163,7 +163,7 @@ object Names {
     def info = NameInfo.TermName
     def underlying: TermName = unsupported("underlying")
 
-    private var derivedNames: AnyRef /* SimpleMap | j.u.HashMap */ =
+    @sharable private var derivedNames: AnyRef /* SimpleMap | j.u.HashMap */ =
```
smarter: Why is this safe? `synchronized` is not used when modifying `derivedNames` currently.

odersky:
The build failures are due to ContextEscapeDetection not compiling. I see:

/drone/src/github.com/lampepfl/dotty/compiler/test/dotty/tools/ContextEscapeDetection.java:23: cannot find symbol
232s
865
[error]   symbol:   class Context
232s
866
[error]   location: class dotty.tools.dotc.core.Contexts
232s
867
[error]     public abstract Contexts.Context getCtx();

Locally `sbt test` succeeds for me. Who can help debug this?

odersky:
Rebased to master

smarter:
> Locally sbt test succeeds for me. Who can help debug this?

The issue only happen with the bootstrapped tests, to run them you need to first publish the non-bootstrapped compiler:
```scala
publishLocal
```
Then run the tests in the `dotty-compiler-bootstrapped` project:
```scala
dotty-compiler-bootstrapped/test
```

The test failure is very reminiscent of a problem I fixed in the backend some times ago: https://github.com/DarkDimius/scala/pull/4

The problem was that the inner class table in the classfile had incorrect outer class names, and indeed if we do `javap -v` on the compiled-by-dotty `Contexts.scala` we see:
```scala
public static abstract #10= #7 of #9; //Context=class dotty/tools/dotc/core/Contexts$Context of class dotty/tools/dotc/core/Contexts$
```
Instead of (from the compiled-by-scalac `Contexts.scala`):
```scala
public static abstract #25= #24 of #2; //Context=class dotty/tools/dotc/core/Contexts$Context of class dotty/tools/dotc/core/Contexts
```
The crucial difference is the last thing in the line: `Contexts$` instead of `Contexts`. The problem comes from `DottyBackendInterface#nameHelper#dropModule`:
```scala
def dropModule: Name = n.stripModuleClassSuffix
```
If I print `n.stripModuleClassSuffix` I get a name which ends with a `$` which is clearly wrong. Further debugging shows that the names are `TypeName` which wrap `SimpleTermName` so they have no semantic information, I have no idea why.

smarter:
Ah, I see now, the full relevant code in the backend is:
```scala
          val outerName = innerClassSym.rawowner.javaBinaryName
          // Java compatibility. See the big comment in BTypes that summarizes the InnerClass spec.
          val outerNameModule = if (innerClassSym.rawowner.isTopLevelModuleClass) outerName.dropModule
```

Where `javaBinaryName` is defined in the backend interface:
```scala
def javaBinaryName: Name = toDenot(sym).fullNameSeparated("/")
```
And `fullNameSeparated` is defined to return a non-semantic name. This can be fixed using `.unmangleClassName`, I've pushed a commit to do so in `javaBinaryName`, though I'm not sure if this is the best place for it.

smarter:
The current failures are due to:
```scala
assert(isValidJVMName(sym.name), s"${sym.fullName} name is invalid on jvm")
```
Because in some cases we get `sym.name.toString` like: `scala.collection$TraversableOnce$collectFirst$$B` which is technically incorrect since it contains a `.`. I don't know why this happens or why it's only happening in the bootstrapped tests.

smarter:
I have two features request to aid debugging, let me know what you think:
- A `rawString` or equivalent method to get the full structure of the name, like printing a raw type or a raw tree, `debugString` is not enough since it doesn't tell if you're dealing with a `PrefixNamedKind`, etc.
- An (optional?) id for names to figure out where they were constructed by doing `assert(id == 42)`, like with symbols. (Names don't have stable hashcodes so that can't be used).

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -339,7 +339,12 @@ object Names {
 
     def isSimple = false
     def asSimpleName = throw new UnsupportedOperationException(s"$debugString is not a simple name")
-    def toSimpleName = termName(toString)
+
+    private var simpleName: SimpleTermName = null
```
smarter: I think you want `private[this] var` here to avoid creating a setter

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -125,11 +125,18 @@ object SymDenotations {
...
     private[this] var myFlags: FlagSet = adaptFlags(initFlags)
     private[this] var myInfo: Type = initInfo
     private[this] var myPrivateWithin: Symbol = initPrivateWithin
     private[this] var myAnnotations: List[Annotation] = Nil
 
+    /** The name of the symbol */
+    def name = myName
+
+    /** Update the name; only called when unpickling top-level classes */
+    def name_=(n: Name) = myName = n
```
smarter: Should this setter be `protected[dotc]` like `def info_=` ?

compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
```diff
@@ -148,46 +149,79 @@ class SymbolLoaders {
...
 
+      override def ensureComplete()(implicit ctx: Context) =
+        for (enter <- enterFlatClasses) enter(ctx)
+
       override def newScopeLikeThis() = new PackageScope
     }
 
     private[core] val currentDecls: MutableScope = new PackageScope()
 
+    def isFlatName(name: SimpleTermName) = name.lastIndexOf('$', name.length - 2) >= 0
```
smarter: Could you add a comment explaining the `name.length - 2` here? Also why is this logic different from the one for the other overload for `isFlatName` ?

compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
```diff
@@ -148,46 +149,79 @@ class SymbolLoaders {
...
-            name.isTypeName && name.isSyntheticFunction)
+        val mangled = name.mangled
+        val e = super.lookupEntry(mangled)
+        if (e != null) e
+        else if (_sourceModule.initialDenot.name == nme.scala_ && _sourceModule == defn.ScalaPackageVal &&
+                 name.isTypeName && name.isSyntheticFunction)
           newScopeEntry(defn.newFunctionNTrait(name.asTypeName))
+        else if (isFlatName(mangled.toSimpleName) && enterFlatClasses.isDefined) {
+          Stats.record("package scopes with flatnames entered")
+          enterFlatClasses.get(ctx)
```
smarter: Why not use `ensureComplete` here?

odersky:
Rebased to master

odersky:
This commit is too big to leave in the queue. There's constant breakage wrt other commits. So I want to get this in by the end of today at the latest.

smarter:
This PR LGTM, except that I don't see how `derivedNames` can be made `@sharable` without using `synchronized` blocks (also there's a test failure but I assume you noticed that).

smarter:
The use of `@sharable` for:
```scala
  @sharable private val simpleNameKinds = new mutable.HashMap[Int, ClassifiedNameKind]
  @sharable private val qualifiedNameKinds = new mutable.HashMap[Int, QualifiedNameKind]
  @sharable private val uniqueNameKinds = new mutable.HashMap[String, UniqueNameKind]
```
is also suspicious. Right now it should be fine since all the NameKinds are instantiated in `object NameKinds`, but nothing enforces that, it would help if the constructors for NameKinds were all private.

odersky:
One more rebase, hopefully the last one 😸

odersky:
I'll roll the private constructor changes into the next PR.

Issue #1625, opened 2016-10-25
Fix varargs in methods


felixmulder:
```scala
trait T1 { def foo(x: String*, y: String*) }
```
slips through the cracks as well as:

```scala
trait T2 { def foo(x: String*, y: String) }
```

This should be fixed, @sebastianharko - were you working on this?

sebastianharko:
@felixmulder Yes, I have a fix for it. To be PR-ed soon. Cheers!

PR #2135, opened 2017-03-22
Fix varargs in methods 

mielientiev:
Fix for issue: #1625 
- added varargs check after method's parameter parsing

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -4,7 +4,7 @@ package parsing
 
 import scala.collection.mutable.ListBuffer
 import scala.collection.immutable.BitSet
-import util.{SourceFile, SourcePosition}
+import util.{ SourceFile, SourcePosition }
 import Tokens._
 import Scanners._
 import MarkupParsers._
@@ -21,7 +21,7 @@ import Constants._
 import ScriptParsers._
 import Comments._
 
-import scala.annotation.{switch, tailrec}
+import scala.annotation.{tailrec, switch}
 import util.DotClass
 import rewrite.Rewrites.patch
 
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1801,6 +1801,10 @@ object Parsers {
           case _          => syntaxError(AuxConstructorNeedsNonImplicitParameter(), start)
         }
       }
+      val listOfErrors = checkVarArgsRules(result)
+      listOfErrors.foreach { vparam =>
+        syntaxError(VarArgsParamMustComeLast(), vparam.tpt.pos)
+      }
       result
     }
 
@@ -1966,10 +1970,6 @@ object Parsers {
         val name = ident()
         val tparams = typeParamClauseOpt(ParamOwner.Def)
         val vparamss = paramClauses(name)
-        val listOfErrors = checkVarArgsRules(vparamss)
-        listOfErrors.foreach { vparam =>
-          syntaxError(VarArgsParamMustComeLast(), vparam.tpt.pos)
-        }
         var tpt = fromWithinReturnType(typedOpt())
         if (in.isScala2Mode) newLineOptWhenFollowedBy(LBRACE)
         val rhs =
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -4,7 +4,7 @@ package parsing
 
 import scala.collection.mutable.ListBuffer
 import scala.collection.immutable.BitSet
-import util.{ SourceFile, SourcePosition }
+import util.{SourceFile, SourcePosition}
 import Tokens._
 import Scanners._
 import MarkupParsers._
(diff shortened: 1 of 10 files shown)


compiler/test/dotc/tests.scala
```diff
@@ -153,6 +153,7 @@ class tests extends CompilerTest {
   @Test def pos_anonClassSubtyping = compileFile(posDir, "anonClassSubtyping", twice)
   @Test def pos_extmethods = compileFile(posDir, "extmethods", twice)
   @Test def pos_companions = compileFile(posDir, "companions", twice)
+  @Test def posVarargsInMethodsT1625 = compileFiles(posDir + "varargsInMethodsT1625/")
```
felixmulder: This should also apply to constructors - I think your implementation supports this, I guess it's just the name that is misleading.

compiler/test/dotc/tests.scala
```diff
@@ -177,6 +178,8 @@ class tests extends CompilerTest {
   @Test def neg_all = compileFiles(negDir, verbose = true, compileSubDirs = false)
   @Test def neg_typedIdents() = compileDir(negDir, "typedIdents")
 
+  @Test def negVarargsInMethodsT1625 = compileFiles(negDir + "varargsInMethodsT1625/")
```
felixmulder: same here

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1209,4 +1209,16 @@ object messages {
            |${parents.mkString("  - ", "\n  - ", "")}
            |""".stripMargin
   }
+
+  case class VarArgsParamMustComeLast()(implicit ctx: Context)
+    extends Message(IncorrectRepeatedParameterSyntaxID) {
+    override def msg: String = "*-parameter must come last"
```
felixmulder: let's say `"varargs parameter must come last"`, mixing symbols and text like this feels kind of weird IMO

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -4,23 +4,24 @@ package parsing
 
 import scala.collection.mutable.ListBuffer
 import scala.collection.immutable.BitSet
-import util.{ SourceFile, SourcePosition }
+import util.{SourceFile, SourcePosition}
```
felixmulder: also - these changes here - are they done by your editor or by you? I'd leave them out of the diff

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -4,23 +4,24 @@ package parsing
...
 import ast.Trees._
 import Decorators._
 import StdNames._
 import util.Positions._
 import Constants._
 import ScriptParsers._
 import Comments._
-import scala.annotation.{tailrec, switch}
+
+import scala.annotation.{switch, tailrec}
```
felixmulder: same with this, if you intended to move this so that `scala` imports are separate from internal imports, then this should be at the top with the other one. Otherwise let's get rid off this change.

tests/neg/varargsInMethodsT1625/firstParamIsVarArgs.scala
```diff
@@ -0,0 +1,3 @@
+trait T1 {
+  def foo(x: String*, y: String): Int // error: *-parameter must come last
```
felixmulder: also - just an FYI, the ": error text here" part of the comment is not parsed in any way (currently)

no need to change or anything

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
mielientiev: I just copied error message from Scala Compiler :)

felixmulder:
Thanks @mielientiev! :tada:

compiler/test/dotc/tests.scala
```diff
@@ -153,6 +153,7 @@ class tests extends CompilerTest {
   @Test def pos_anonClassSubtyping = compileFile(posDir, "anonClassSubtyping", twice)
   @Test def pos_extmethods = compileFile(posDir, "extmethods", twice)
   @Test def pos_companions = compileFile(posDir, "companions", twice)
+  @Test def posVarargsT1625 = compileFiles(posDir + "varargsInMethodsT1625/")
```
nicolasstucki: This is an unnecessary complication, just put your test sources in `tests/pos/` and remove this `@Test def ...`. Also use the prefix `i1625` for the test sources.

compiler/test/dotc/tests.scala
```diff
@@ -177,6 +178,8 @@ class tests extends CompilerTest {
   @Test def neg_all = compileFiles(negDir, verbose = true, compileSubDirs = false)
   @Test def neg_typedIdents() = compileDir(negDir, "typedIdents")
 
+  @Test def negVarargsT1625 = compileFiles(negDir + "varargsInMethodsT1625/")
```
nicolasstucki: This is an unnecessary complication, just put your test sources in `tests/neg/` and remove this `@Test def ...`. Also use the prefix `i1625` for the test sources.

felixmulder: this is going away once we merge #2125 anyway, I'll deal with it.

Issue #2132, opened 2017-03-21
make it clear in documentation that  ./scala-scala/ is required


DarkDimius:
"Getting started" guide doesn't say anything about it.
Idea: integrate "getting started" with something similar to "tut"?

felixmulder:
Dotty doc is getting typechecking similar to tut - as soon as I've cleared away other more pressing tasks.

We could make the tests depend on running a script to make sure scala-scala is cloned and updated, I did something similar when dottydoc had a javascript client.

OlivierBlanvillain:
What's the argument against dumping everything into /tests/scala-library?

felixmulder:
I've argued in the past that this is what we eventually should do - I don't think the changes we're applying really have anything to do with improving, or making the old collections compatible with Dotty. Especially now that we are including the new collections in 2.13.

They're just a good set of integration tests IMO.

sjrd:
It can also [be automated](https://github.com/scala-js/scala-js/blob/v0.6.15/project/Build.scala#L1628-L1658), you know ;)

felixmulder:
Omg, no more sbt magic please :joy:

DarkDimius:
> What's the argument against dumping everything into /tests/scala-library?

 - not being able to upstream changes;
 - not being able to merge changes from upstream to us;
 - loosing commit history. I ___strongly___ believe that changing dotty directory structure wasn't worth it. I've lost around ___twice___ in productivity in everyday tasks;
 - there's a huge amount of work involved in maintaining the library, for as long as we can I'd prefer to use one provided by Lightbend not to spread dotty efforts thin.

That's a rare case when I'd prefer sbt magic to the alternative.

smarter:
> loosing commit history. I strongly believe that changing dotty directory structure wasn't worth it. I've lost around twice in productivity in everyday tasks;

Did the reorg break one of your tool in particular? The only change for me is that I have to remember to pass `--follow` to `git log` to get the full log.

felixmulder:
I don't think we'll be upstreaming changes anyway as the new library is on its way in, and if so - it feels like the other arguments fall away.

We could still filter in the changes from the forked library, that would solve the problem of relevant history.

DarkDimius:
> Did the reorg break one of your tool in particular? The only change for me is that I have to remember to pass --follow to git log to get the full log.

It broke both IntelliJ, github and SourceTree. 

>I don't think we'll be upstreaming changes anyway as the new library is on its way in

[...]

PR #2138, opened 2017-03-23
Add a warning and instructions for new users.

nicolasstucki:
This is a small check to make sure users haven't forgotten to clone the `scala-scala` library. It contributes to solving #2132

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -12,6 +12,8 @@ import sbt.Package.ManifestAttributes
 
 object DottyBuild extends Build {
 
+  projectChecks()
+
   val scalacVersion = "2.11.5" // Do not rename, this is grepped in bin/common.
 
   val dottyOrganization = "ch.epfl.lamp"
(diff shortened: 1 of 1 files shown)


project/Build.scala
```diff
@@ -903,4 +905,18 @@ object DottyInjectedPlugin extends AutoPlugin {
...
     ))
   }
+
+  private def projectChecks(): Unit = {
+    val scalaScala = new File("scala-scala")
+    if (!scalaScala.exists()) {
+      println(
+        s"""[WARNING] Missing `dotty/scala-scala` library
+           |You can clone the library with:
+           |  > git clone https://github.com/DarkDimius/scala.git ${scalaScala.getAbsolutePath}
```
smarter: You can do it in one line with git clone -b, see http://dotty.epfl.ch/docs/contributing/workflow.html

olhotak:
Is cloning scala-scala something that is going to be a requirement for a long time?

If so, it should also be added to docs/docs/contributing/getting-started.md.

nicolasstucki:
@olhotak issue #2132 is about adding it to the documentation. This is just a small safeguard.

DarkDimius:
Merging as this is clearly better then current state.
I disagree with putting entire stdlib in ./tests/
I believe we should have it as a git submodule.

smarter:
I also think it would make sense to have it as a submodule if we get the backend as a submodule anyway.

PR #2139, opened 2017-03-23
Fix #2137: Create dummy companions for top-level objects without a real one

### LINKED ISSUE #2137: Dotty sometimes loads denotation from classpath, even when a more recent version is being typechecked
Suppose that some file defines an object:
```scala
package dummy
object Foo
```
While typechecking this file, a term for `Foo` is entered in the scope of `dummy`, but no corresponding type is entered because there is no `class Foo`. However, once compilation is done, we output:
```scala
./dummy/Foo.class
./dummy/Foo$.class
```
If the current directory is on the classpath, this means that any subsequent compiler run will enter a type `Foo` in the scope of `dummy` because of how `PackageLoader` works. Of course, `dummy.Foo` is not a valid class, `TypeAssigner` contains a method `reallyExists` to determine this:
```scala
  /** A denotation exists really if it exists and does not point to a stale symbol. */
  final def reallyExists(denot: Denotation)(implicit ctx: Context): Boolean = try
    denot match {
      case denot: SymDenotation =>
        denot.exists && {
          denot.ensureCompleted
          !denot.isAbsent
        }
      case denot: SingleDenotation =>
        val sym = denot.symbol
        (sym eq NoSymbol) || reallyExists(sym.denot)
      case _ =>
        true
    }
  catch {
    case ex: StaleSymbol => false
  }
```

I think there is a fatal flaw in `reallyExists`: it requires forcing the info of the potentially-stale/absent symbol, this means we might unpickle trees from the classpath and pollute the scope with things that no longer exist. Here's a demonstration of how this might lead the compiler to crash:

`outerFoo_1.scala`:
```scala
package outer
class Foo
```
`outerinnerFoo_1.scala`:
```scala
package outer
package inner
object Foo {
  val a: Int = 1
}
```

`outerinnerFoo_2.scala`:
```scala
package outer
package inner
object Foo {
  // val a: Int = 1
}
```
`outerinnerTest_2.scala`
```scala
package outer
package inner
object Test {
  val x: Foo = new Foo
}
```

```
$ dotc outerFoo_1.scala outerinnerFoo_1.scala
$ dotc outerinnerTest_2.scala outerinnerFoo_2.scala
exception occurred while typechecking outerinnerFoo_2.scala

exception occurred while compiling outerinnerTest_2.scala, outerinnerFoo_2.scala
Exception in thread "main" java.util.NoSuchElementException: None.get
        at scala.None$.get(Option.scala:347)
        at scala.None$.get(Option.scala:345)
        at dotty.tools.dotc.typer.Typer.localTyper(Typer.scala:1500)
        at dotty.tools.dotc.typer.Typer.typedNamed$1(Typer.scala:1519)
        at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:1577)
```
Full stacktrace: https://gist.github.com/smarter/c854842847e1e4d12794adabee5ad7c0

I don't know exactly why this is crashing but the root cause here is that when I do `val x: Foo` in `Test`, `typedIdent` will first try `outer.inner.Foo`, which means calling `reallyExists` on it and thus unpickling `outer.inner.Foo` and `outer.inner.Foo$` from `outer/inner/Foo.class` even though we are also compiling `outerinnerFoo_2.scala` which contains a more recent version of the object `outer.inner.Foo$`.

The only solution I can think of is that whenever we enter the symbol for a class or object without companion, we should mark the companion as absent to make very sure that we never try to load it from the classpath.

smarter:
Previously, we sometimes ended up forcing a companion class symbol from
a previous run or from the classpath which lead to weird issues like in
`tests/pos-special/false-companion`. Even if we end up not forcing such
a symbol, its presence can still lead to issue: before this commit
incremental compilation of `dotty-compiler-bootstrapped` was broken
because we recorded a false dependency on the non-bootstrapped
`dotty-compiler` jar.

### DIFF
--- a/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
+++ b/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
@@ -175,7 +175,7 @@ private class ExtractDependenciesCollector(implicit val ctx: Context) extends tp
   override def traverse(tree: Tree)(implicit ctx: Context): Unit = {
     tree match {
       case Import(expr, selectors) =>
-        def lookupImported(name: Name) = expr.tpe.select(name).typeSymbol
+        def lookupImported(name: Name) = expr.tpe.member(name).symbol
         def addImported(name: Name) = {
           // importing a name means importing both a term and a type (if they exist)
           addDependency(lookupImported(name.toTermName))
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -605,6 +605,25 @@ class Namer { typer: Typer =>
           case EmptyTree =>
         }
       }
+
+      // If a top-level object has no companion class in the current run, we
+      // enter a dummy companion class symbol (`denot.isAbsent` returns true) in
+      // scope. This ensures that we never use a companion from a previous run
+      // or from the classpath. See tests/pos/false-companion for an
+      // example where this matters.
+      if (ctx.owner.is(PackageClass)) {
+        for (cdef @ TypeDef(moduleName, _) <- moduleDef.values) {
+          val moduleSym = ctx.denotNamed(moduleName.encode).symbol
+          if (moduleSym.isDefinedInCurrentRun) {
+            val className = moduleName.stripModuleClassSuffix.toTypeName
+            val classSym = ctx.denotNamed(className.encode).symbol
+            if (!classSym.isDefinedInCurrentRun) {
+              val absentClassSymbol = ctx.newClassSymbol(ctx.owner, className, EmptyFlags, _ => NoType)
+              enterSymbol(absentClassSymbol)
+            }
+          }
+        }
+      }
     }
 
     stats.foreach(expand)
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -570,8 +570,8 @@ class Namer { typer: Typer =>
 
     /** Create links between companion object and companion class */
     def createLinks(classTree: TypeDef, moduleTree: TypeDef)(implicit ctx: Context) = {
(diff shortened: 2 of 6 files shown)


smarter:
@felixmulder Actually scratch, I don't need to do anything special to get the test to run correctly with partest.

smarter:
While the test works fine with partest,I couldn't get it to even compile with JUnit, and making it do so seems complicated, so I'll just disable it until https://github.com/lampepfl/dotty/pull/2125 is in.

odersky:
What's the problem with junit?

smarter:
it will try to compile the four files together, but two of them define the same object, so it fails. I tried using the trick you used for `pos_t5545` but that fails too, because the second run needs the output of the first run to work. I tried changing the compiler options so that the output of the first run is in the classpath of the second run and that failed again and at that point I gave up.

smarter:
@felixmulder  http://dotty-ci.epfl.ch/lampepfl/dotty/1418/4 has been marked in progress for 21 hours, even though all tests finished successfully, any idea what's going on? This also seems to be happenning in http://dotty-ci.epfl.ch/lampepfl/dotty/1423/3 which succeeded a while ago but has been marked in progress for 2 hours now. Meanwhile, I'll push-force commits with a different timestamp to force the CI to rerun.

felixmulder:
We should look over upgrading the CI to the latest version, but in the mean
time - the best solution is to ask Fabien to restart it if you can't fix it.

Cheers,
Felix

PS - you have SSH access to both machines, so you should also be able to
manually restart it :)

smarter:
I was hoping there was some debug information we could recover to understand what's going on before restarting stuff

felixmulder:
You could read the docker logs for the image?

felixmulder:
On vacation so: ¯\_(ツ)_/¯

smarter:
Ping for review.

smarter:
It doesn't conflict currently so I think it's fine.

PR #2141, opened 2017-03-25
Enable incremental compilation scripted tests on CI

smarter:
This wasn't enabled before because these tests take a long time, but
thanks to https://github.com/sbt/sbt/pull/3015 scripted tests can now
manually be splitted in M groups of equal size using the "/*NofM" syntax.


smarter:
@felixmulder The scripted test runs each take ~20 minutes when the JUnit tests take ~24 minutes, so we should be able to turn them on without making the CI slower. If the CI has a configuration option for maximum number of jobs run in parallel, it should be increased.

felixmulder:
The new parallel test suite runs in ~12 minutes - but I think by tweaking the memory given to each sbt instance we can decrease that time further.

So perhaps it would be a good idea to split these further - or add them as conditional tests that are only run when the PR is merged to master?

smarter:
Okay, I guess we wait until your stuff is merged and re-evaluate then.

felixmulder:
We've decided to put this into nightly tests.

PR #2148, opened 2017-03-29
Fix #2147: redirect both java and scala std out/err

### LINKED ISSUE #2147: System.out.println not printing in new test environment.
The following tests fails a a `run` test.
```scala
object Test {
  def main(args: Array[String]): Unit = System.out.println("42")
}
```

felixmulder:


compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -399,18 +399,25 @@ trait ParallelTesting {
...
+      val oldOut = System.out
+      val oldErr = System.out
       try {
         // Do classloading magic and running here:
         val ucl = new URLClassLoader(Array(dir.toURI.toURL))
         val cls = ucl.loadClass("Test")
         val meth = cls.getMethod("main", classOf[Array[String]])
 
+        val ps = new PrintStream(printStream)
+        System.setOut(ps)
```
DarkDimius: this is thread-unsafe

How did this work in the CI before? how did other tests pass?

felixmulder: Probably because all run tests execute without being interrupted.

felixmulder: Also, CI didn't fail because of tests - it failed because we restarted the agent (aka slave). New commit makes it safe.

nicolasstucki:
LGTM

PR #2149, opened 2017-03-29
Create dummy companions for classes without a real one

smarter:
In #2139 I added code to create dummy companions for companion-less
objects, but not for companion-less classes because I thought it wasn't
needed. But it turns out that even if the classpath only has `Foo.class`
and not `Foo$.class`, a module for `Foo` is entered by
`initializeFromClassPath` when it calls `enterClassAndModule`, so we
have to add dummy companions to classes. I don't have a test to
illustrate this issue, but note that it fixes the incremental
compilation bug demonstrated by
https://github.com/dotty-staging/dotty/commits/incremental-compilation-bug.

Note: I verified that adding a dummy companion to "class Foo" did not
cause the backend to emit a Foo$.class, so this should have no visible
impact.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -606,11 +606,11 @@ class Namer { typer: Typer =>
         }
       }
 
-      // If a top-level object has no companion class in the current run, we
-      // enter a dummy companion class symbol (`denot.isAbsent` returns true) in
-      // scope. This ensures that we never use a companion from a previous run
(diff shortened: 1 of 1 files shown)


sjrd:
This could be an issue for code like this:
```scala
object Foo {
  class Bar
  val Bar = "babar"
}
```

smarter:
@sjrd The dummy companions are only added for top-level classes and objects so it should be fine. Even if it wasn't the case, no dummy companion for `class Bar` would be added if a term named `Bar` was declared in the same scope.

smarter:
Ping for review.

DarkDimius:
LGTM

PR #2154, opened 2017-03-30
Fix #2151: don't die for wrong number of typeargs applied

### LINKED ISSUE #2151: Compiler dies somewhere in the middle of typechecking
https://gist.github.com/alexknvl/0a4af0baeb6ff49a850fe6e9a9765c8a

felixmulder:


smarter:
> ERROR: terminal inactive for 10m0s, build cancelled

Looks like 10 minutes for the run tests is not enough.

felixmulder:
Looks like we have a spurious error of some kind, will investigate tomorrow.

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -12,7 +12,7 @@ import scala.util.matching.Regex
 class CompilationTests extends ParallelSummaryReport with ParallelTesting {
   import CompilationTests._
 
-  def isInteractive: Boolean = ParallelSummaryReport.isInteractive
+  def isInteractive: Boolean = false//ParallelSummaryReport.isInteractive
```
smarter: what

compiler/test/dotty/tools/dotc/reporting/TestReporter.scala
```diff
@@ -93,35 +93,25 @@ object TestReporter {
...
+        writer.println(msg)
+        _messageBuf.append(msg)
+
+        if (extraInfo.nonEmpty) {
+          writer.println(extraInfo)
+          _messageBuf.append(extraInfo)
+        }
       }
     }
+    rep
```
smarter: Would be good to add a link to the corresponding bug report if you open one

PR #2157, opened 2017-03-30
Remove dead code in pattern matching

OlivierBlanvillain:
[This entry point](https://github.com/lampepfl/dotty/blob/f75caad00256db52bbd3310f245e51d23c2a76cc/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala#L1408) to pattern matching is apparently unreachable.

- The first commit removes that case and the parts that become unreachable.
- The second commit removes `tupleExtractor` which in also unreachable in dotty.
- Third commit squashes two class together, with the dead code remove it becomes useless inheritance

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -1813,11 +1813,6 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
     }
 
     object alignPatterns extends ScalacPatternExpander {
-      /** Converts a T => (A, B, C) extractor to a T => ((A, B, CC)) extractor.
-        */
-      def tupleExtractor(extractor: Extractor): Extractor =
-        extractor.copy(fixed = defn.tupleType(extractor.fixed) :: Nil)
-
       private def validateAligned(tree: Tree, aligned: Aligned): Aligned = {
         import aligned._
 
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1348,70 +1343,46 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
...
 
       private def productElemsToN(binder: Symbol, n: Int): List[Tree] = 1 to n map tupleSel(binder) toList
       private def genTake(binder: Symbol, n: Int): List[Tree]         = (0 until n).toList map (codegen index seqTree(binder))
       private def genDrop(binder: Symbol, n: Int): List[Tree]         = codegen.drop(seqTree(binder))(expectedLength) :: Nil
 
-      // codegen.drop(seqTree(binder))(nbIndexingIndices)))).toList
-      protected def seqTree(binder: Symbol)                = tupleSel(binder)(firstIndexingBinder + 1)
-      protected def tupleSel(binder: Symbol)(i: Int): Tree = {
+      private def seqTree(binder: Symbol): Tree =
+        if (firstIndexingBinder == 0) ref(binder)
```
DarkDimius: this isn't a refactoring, why did you need this change?

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1348,70 +1343,46 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
...
-
       def resultInMonad  =
         if (aligner.isBool) defn.UnitType
         else if (isProductMatch(resultType, aligner.prodArity)) resultType
         else if (isGetMatch(resultType)) extractorMemberType(resultType, nme.get)
         else resultType
 
-      def resultType: Type
-
-      /** Create the TreeMaker that embodies this extractor call
```
DarkDimius: it would be nice to preserve documentation

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
OlivierBlanvillain: It's overriden in the only trait instance https://github.com/lampepfl/dotty/pull/2157/files#diff-b72950e41046b80e7386c4936d560cf9L1611

OlivierBlanvillain:
Alright, removed the 3rd commit.

PR #2158, opened 2017-03-30
Fix #2156

### LINKED ISSUE #2156: Pattern matcher needs to check for outer
When testing against an inner class, the pattern matcher needs to test outer pointers. Consider the following program:

    class Outer {

      case class Inner()

      val inner: Inner = new Inner

      def checkInstance(o: Outer) =
        o.inner.isInstanceOf[this.Inner]

      def checkPattern1(i: Any) =
        i match {
          case _: Inner => true
          case _ => false
        }

      def checkPattern2(i: Any) =
        i match {
          case Inner() => true
          case _ => false
        }

      def checkEquals(o: Outer) =
        o.inner == inner
    }

    object Test {

        val o1 = new Outer
        val o2 = new Outer
        assert(o1.checkInstance(o2)) // ok
        assert(!o1.checkPattern1(o2.inner)) // ok under scalac, fails for dotc-compiled code
        assert(!o1.checkPattern2(o2.inner))  // ok under scalac, fails for dotc-compiled code
        assert(!o1.checkEquals(o2))  // ok under scalac, fails for dotc-compiled code
      }
    }

We expect the `isInstanceOf` test in `checkInstance` to return `true` but the other three tests should return `false`. Yes, I know it's contorted, but these Scala rules were a compromise between
not losing performance and being precise. Essentially we accept that we lose precision in `isInstanceOf` because otherwise there would be no way to avoid the "outer test tax". But we demand that pattern matchers test outer pointers. Indeed you will find specific code doing that after patmat when you compile the code with `scalac`:

        if (x1.isInstanceOf[Outer.this.Inner].&&
        ((x1.asInstanceOf[Outer.this.Inner]: Outer.this.Inner).<outer>.eq(Outer.this)))

But such code is currently not emitted by `dotc` which is why the last three asserts fail.

The discrepancy is the reason why the semantic names PR fails the bootstrap.

DarkDimius:
New testing infrastructure seems to deadlock on my machine so I can't test stuff.
Using CI instead.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -852,9 +852,10 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
         // See the test for SI-7214 for motivation for dealias. Later `treeCondStrategy#outerTest`
         // generates an outer test based on `patType.prefix` with automatically dealises.
         expectedTp.dealias match {
-          case tref @ TypeRef(pre, name) =>
-            (pre ne NoPrefix) && tref.symbol.isClass &&
-            ExplicitOuter.needsOuterIfReferenced(tref.symbol.asClass)
+          case tref @ TypeRef(pre: SingletonType, name) =>
+            val s = tref
+            s.symbol.isClass &&
+            ExplicitOuter.needsOuterIfReferenced(s.symbol.asClass)
           case _ =>
             false
         }
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -848,10 +848,10 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
 
       val nextBinder = afterTest.asTerm
 
-      def needsOuterTest(patType: Type, selType: Type, currentOwner: Symbol): Boolean = {
+      def outerTestNeeded(implicit ctx: Context): Boolean = {
         // See the test for SI-7214 for motivation for dealias. Later `treeCondStrategy#outerTest`
         // generates an outer test based on `patType.prefix` with automatically dealises.
-        patType.dealias match {
+        expectedTp.dealias match {
           case tref @ TypeRef(pre, name) =>
             (pre ne NoPrefix) && tref.symbol.isClass &&
             ExplicitOuter.needsOuterIfReferenced(tref.symbol.asClass)
@@ -861,12 +861,6 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
       }
 
       override lazy val introducedRebindings = NoRebindings
-
-      def outerTestNeeded = {
-        /*val np = expectedTp.normalizedPrefix
-        val ts = np.termSymbol
-        (ts ne NoSymbol) && */needsOuterTest(expectedTp, testedBinder.info, ctx.owner)
-      }
 
       // the logic to generate the run-time test that follows from the fact that
       // a `prevBinder` is expected to have type `expectedTp`
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -863,9 +863,9 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
       override lazy val introducedRebindings = NoRebindings
 
       def outerTestNeeded = {
-        val np = expectedTp.normalizedPrefix
+        /*val np = expectedTp.normalizedPrefix
         val ts = np.termSymbol
-        (ts ne NoSymbol) && needsOuterTest(expectedTp, testedBinder.info, ctx.owner)
+        (ts ne NoSymbol) && */needsOuterTest(expectedTp, testedBinder.info, ctx.owner)
       }
 
       // the logic to generate the run-time test that follows from the fact that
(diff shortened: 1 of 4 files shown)


DarkDimius:
>exception caught when loading class Class: dotty.tools.FatalError: Unexpected ClassRep 'ClassRep(Some(/usr/lib/jvm/java-1.8-openjdk/jre/lib/rt.jar(java/lang/Class$ReflectionData.class)),None)' found searching for name 'lang.Class$ReflectionData'

Does our CI work?

DarkDimius:
ping @felixmulder

felixmulder:
deadlock in `runAll`?

You can get the old behaviour back by doing:

```
> set testOptions in LocalProject("dotty-compiler")
> set testOptions in LocalProject("dotty-compiler-bootstrapped")
```

felixmulder:
CI should be fine, at least the first 4 tests - they are unchanged.

Also, regarding your error - the stacktrace says that it it's in:

`dotty.tools.dotc.core.classfile.ClassfileParser.enterOwnInnerClasses(ClassfileParser.scala:669)`

DarkDimius:
@felixmulder, I cant reproduce the error locally and it fails in tests which don't have pattern matching at all, e.g. `dotty.tools.ShowClassTests.loadDotty` has nothing to do with with it, it doesn't even run compilation pipeline.

DarkDimius:
Or, `dotty.tools.dotc.parsing.DocstringTests.multipleDocStringsBeforeAndAfter`, which as you know, only runs parser.

felixmulder:
But it's failing in the bootstrapped version of Dotty.

I'll try to reproduce, 2 secs

DarkDimius:
@felixmulder, I've found the issue. No need for you to investigate anymore.

DarkDimius:
Bringing back #2156 into discussion.

felixmulder:
Alright, then I'm going to bed 👍

compiler/src/dotty/tools/io/ClassPath.scala
```diff
@@ -242,8 +242,8 @@ abstract class ClassPath {
       case Some((pkg, rest)) =>
         val rep = packages find (_.name == pkg) flatMap (_ findClass rest)
         rep map {
-          case x: ClassRep  => x
-          case x            => throw new FatalError("Unexpected ClassRep '%s' found searching for name '%s'".format(x, name))
+          case x: AnyClassRep => x
+          case x              => throw new FatalError("Unexpected ClassRep '%s' found searching for name '%s'".format(x, name))
```
smarter: `rep` is an `AnyClassRep` so this case will never happen. I believe the original code intended to check the outer pointer, though I have no idea why, here's the commit that introduced the check: https://github.com/scala/scala/commit/a1a6ab90ac5fe7d5db2f6fba788ce14c1613d9c6

DarkDimius: Code compiled by `scalac` doesn't contain the check:

```
option = rep.map((Function1)new Serializable(this, name){
                public static final long serialVersionUID = 0;
                private final String name$2;

                public final ClassRep apply(ClassRep x0$1) {
                    ClassRep classRep = x0$1;
                    if (classRep instanceof ClassRep) {
                        ClassRep classRep2;
                        ClassRep classRep3 = classRep2 = classRep;
                        return classRep3;
                    }
                    throw new dotty.tools.FatalError(new scala.collection.immutable.StringOps(Predef..MODULE$.augmentString("Unexpected ClassRep '%s' found searching for name '%s'")).format((Seq)Predef..MODULE$.genericWrapArray((Object)new Object[]{classRep, this.name$2})));
                }

                public final /* bridge */ /* synthetic */ Object apply(Object v1) {
                    return this.apply((ClassRep)v1);
                }
            });
```

smarter: I'm not saying that the check was actually performed, I'm saying that the code was written assuming that the check would be performed, which apparently it wasn't.

odersky: See comment on the ticket #2156. I think it's a bug in scalac. By guess would be it checks whether the test would redundant (which is reasonable) but gets that logic messed up. For instance, here is a situation where the test should not be omitted:

    class Outer {
      trait Base
      case class Inner() extends Base

      val x: Base = ???
      x match {
        case x: Inner => ???
      }
    }

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -848,26 +848,21 @@ class PatternMatcher extends MiniPhaseTransform with DenotTransformer {
...
+      def outerTestNeeded(implicit ctx: Context): Boolean = {
         // See the test for SI-7214 for motivation for dealias. Later `treeCondStrategy#outerTest`
         // generates an outer test based on `patType.prefix` with automatically dealises.
-        patType.dealias match {
-          case tref @ TypeRef(pre, name) =>
-            (pre ne NoPrefix) && tref.symbol.isClass &&
-            ExplicitOuter.needsOuterIfReferenced(tref.symbol.asClass)
+        expectedTp.dealias match {
+          case tref @ TypeRef(pre: SingletonType, name) =>
+            val s = tref
```
odersky: Why the added definition of `s`. Just use `tref`!

odersky:
Otherwise LGTM

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
DarkDimius: variables defined in pattern matches are rarely debuggable when compiled by scalac.

odersky: OK, if you prefer.

PR #2159, opened 2017-03-30
Make case class hashCode take class into account

odersky:
Previously, elements of the following classes had the same hash code:

    case class A()
    case class B()

Now they are distinguished.


smarter:
What's the usecase for that?

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -149,15 +149,15 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
...
+     *      <synthetic> var acc: Int = productPrefix.hashCode()
      *      acc = Statics.mix(acc, x);
      *      acc = Statics.mix(acc, Statics.this.anyHash(y));
      *      Statics.finalizeHash(acc, 2)
      *    }
      */
     def caseHashCodeBody(implicit ctx: Context): Tree = {
       val acc = ctx.newSymbol(ctx.owner, "acc".toTermName, Mutable | Synthetic, defn.IntType, coord = ctx.owner.pos)
-      val accDef = ValDef(acc, Literal(Constant(0xcafebabe)))
+      val accDef = ValDef(acc, This(clazz).select(defn.Product_productPrefix).select(defn.Any_hashCode).ensureApplied)
```
DarkDimius: Given that this code generates also `productPrefix` itself, why not inline constant `ownName` here instead?

DarkDimius: Also, given that hashcodes on strings are fixed by the spec, we can actually compute the runtime value and inline it: `Literal(ownName.toString.hashcode)`.

DarkDimius:
> What's the usecase for that?

Less hash collisions for unrelated things.

```
case class A(val a: Int)
case class B(val b: Int)
case class C(val c: Int)
...
...
case class Z(val z: Int)

new HashSet(new A(0), new B(0), new C(0), ..., new Z(0)) <-- I love O(n^2) hashmaps :-)
```

odersky:
That failure looks like a deadlock. /cc @felixmulder

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -144,20 +144,21 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
...
+     *      <synthetic> var acc: Int = "p.C".hashCode // constant folded
      *      acc = Statics.mix(acc, x);
      *      acc = Statics.mix(acc, Statics.this.anyHash(y));
      *      Statics.finalizeHash(acc, 2)
      *    }
      */
     def caseHashCodeBody(implicit ctx: Context): Tree = {
       val acc = ctx.newSymbol(ctx.owner, "acc".toTermName, Mutable | Synthetic, defn.IntType, coord = ctx.owner.pos)
-      val accDef = ValDef(acc, Literal(Constant(0xcafebabe)))
+      val accDef = ValDef(acc, Literal(Constant(clazz.fullName.toString.hashCode)))
```
retronym: `scalac` uses `productPrefix` (which is user definable), rather than the class name. 

What is the definition of `fullName` for non-member classes?

DarkDimius: > scalac uses productPrefix (which is user definable), rather than the class name.

What's the motivation behind this?
Product prefix isn't required to even return the same value, so one can break hash-code by returning different values there.

retronym: Neither are the case accessors (they could be `var`s).

I guess a motivation would be to have more stable hash codes for a local case class so that it wasn't volatile wrt fresh name suffixes.

DarkDimius: AFAIK, The full name here won't include any fresh name suffices. They will be only added after Flatten. Two local classes with the same name will get the same constant.

retronym: > Neither are the case accessors (they could be vars).

I see now, though, that a mutable `productPrefix` would be worse that a var accessor, as `equals` does not check that `this.productPrefix == that.productPrefix`.

PR #2160, opened 2017-03-31
Make testing more robust

odersky:
Emacs often produces temporary files in directories. These used to
cause NPEs in the new testing framework. We now fix this by only
compiling file names that designate source files.

Based on #2158. Only last commit is new.


smarter:
> Emacs often produces temporary files in directories

Off-topic, but if this behavior annoys you, note that you can easily get these files in a separate directory  by putting something like this in your `.emacs`:
```lisp
(setq backup-directory-alist
      `((".*" . "~/.emacs.d/backups/")))
(setq auto-save-file-name-transforms
      `((".*" "~/.emacs.d/autosaves/\\1" t)))
```

PR #2161, opened 2017-03-31
Test that SI-4400 stays fixed.

DarkDimius:


DarkDimius:
seems like a spurious CI failure

PR #2162, opened 2017-03-31
Fix #1959: infix type operators in the REPL

### LINKED ISSUE #1959: Infix types do not work in REPL
```scala
scala> trait +[N1, N2] 
defined trait +
scala> type MyAlias = Int + Int 
<console>:4: error: not found: type +
type MyAlias = Int + Int
```

abeln:
Infix type operators were broken in the REPL.
The REPL uses fold methods from the untpd package,
but those were skipping the operator subtree when folding
over an InfixOp.

Fix the issue by taking the operator into account.

Tested:
1) Verified that only the REPL code uses the modified
fold method.
2) The following interaction now works in the REPL
(used to fail with a 'can't find +' error):

scala> class +[A, B](a: A, B: B)
defined class +
scala> type IntAndString = Int + String
defined type alias IntAndString
scala> new IntAndString(42, "hello world")
val res0: +[Int, String] = $plus@6a472554

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/untpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/untpd.scala
@@ -526,11 +526,11 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
       case Function(args, body) =>
         this(this(x, args), body)
       case InfixOp(left, op, right) =>
-        this(this(x, left), right)
+        this(this(this(x, left), op), right)
       case PostfixOp(od, op) =>
-        this(x, od)
+        this(this(x, od), op)
       case PrefixOp(op, od) =>
-        this(x, od)
+        this(this(x, op), od)
       case Parens(t) =>
         this(x, t)
       case Tuple(trees) =>
--- a/tests/repl/infixTypeOp.check
+++ b/tests/repl/infixTypeOp.check
@@ -1,0 +1,5 @@
+scala> trait +[A, B]
+defined trait +
+scala> type IntAndString = Int + String
+defined type alias IntAndString
+scala> :quit


abeln:
This fixes #1959.

DarkDimius:
@abeln, thank you for your contribution, would you please edit commit message to start with 
"Fix #1959: "? It helps a lot when navigating commit history in github.

DarkDimius:
Would you also kindly include a test to repl tests that would track that the issue stays fixed?

smarter:
@abeln Nice catch! Would you mind adding a REPL test? It's pretty easy, you just need to add a file in https://github.com/lampepfl/dotty/tree/master/tests/repl which contains the output of a REPL session.

abeln:
PTAL -- added repl test
thanks for the prompt review!

DarkDimius:
>@abeln, thank you for your contribution, would you please edit commit message to start with
"Fix #1959: "? It helps a lot when navigating commit history in github.

@abeln, could you please update commit message?

abeln:
@DarkDimius oops, sorry -- I'd changed the PR title, but forgot the commit title. PTAL.

compiler/src/dotty/tools/dotc/ast/untpd.scala
```diff
@@ -526,7 +526,7 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
       case Function(args, body) =>
         this(this(x, args), body)
       case InfixOp(left, op, right) =>
-        this(this(x, left), right)
+        this(this(this(x, left), op), right)
       case PostfixOp(od, op) =>
```
smarter: I haven't checked how `UntypedTreeAccumulator` is used exactly, but if InfixOp folds over the `op`, then surely `PostfixOp` and `PrefixOp` should too?

abeln: `UntypedTreeAccumulator` (and its cousing `UntypedDeepFolder`) is used only in the repl code (`CompilingInterpreter.scala`).
Changed `{Post, Pre}fixOp` as well.
PTAL.

abeln:
@DarkDimius friendly ping :)

compiler/src/dotty/tools/dotc/ast/untpd.scala
```diff
@@ -526,11 +526,11 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
...
-        this(x, od)
+        this(this(x, od), op)
       case PrefixOp(op, od) =>
-        this(x, od)
+        this(this(x, op), od)
```
DarkDimius: Would it be possible to expand the test to also test this change?

abeln: Does Dotty have prefix type operators? If so, how are they used?
"Regular" prefix operators already worked in the REPL *before* this change: it's a different ast that doesn't end up using the code above.

This is the hypothetical test I'd add, but like I said it's not really exercising this PR.

> scala> class B(val n: Int) {
>         def unary_! = new B(-n)
>        } 
> defined class B
> scala> val x = new B(42) 
> val x: B = B@10e92f8f
> scala> (!x).n 
> val res4: Int = -42

Thoughts?

felixmulder:
Thanks for the contribution! :tada:

compiler/src/dotty/tools/dotc/ast/untpd.scala
DarkDimius: Yes, I agree. Let's get this in.

PR #2168, opened 2017-04-03
Fixes `@unchecked` warnings

allanrenucci:
- Fixe the few `@unchecked` warnings emitted when compiling dotty.
- Replace a `map` by `mapConserve` in `UntypedTreeMap`

dotty fails to compile the files I modified. Not sure if this reveals a bug or if it is a known incompatibility between scalac and dotty
```scala
-- Error: dotty/compiler/../compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala 
542 |    def unapply(tree: Tree @unchecked): Option[Type] = tree match {
    |                           ^^^^^^^^^^
    |                      annotation.unchecked does not have a constructor
-- [E007] Type Mismatch Error: dotty/compiler/../compiler/src/dotty/tools/dotc/transform/PostTyper.scala 
267 |            case ident @ Ident(_)                 => checkIdent(ident)
    |                                                                ^^^^^
    |found:    dotty.tools.dotc.ast.Trees.Ident[dotty.tools.dotc.ast.Trees.Untyped](ident)
    |required: dotty.tools.dotc.ast.tpd.Ident
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeBuffer.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeBuffer.scala
@@ -159,7 +159,7 @@ class TreeBuffer extends TastyBuffer(50000) {
       val tree = it.next
       treeAddrs.get(tree) match {
         case addr: Addr => treeAddrs.put(tree, adjusted(addr))
-        case addrs: List[Addr] => treeAddrs.put(tree, addrs.map(adjusted))
+        case addrs: List[Addr @unchecked] => treeAddrs.put(tree, addrs.map(adjusted))
       }
     }
   }
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeBuffer.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeBuffer.scala
@@ -159,7 +159,6 @@ class TreeBuffer extends TastyBuffer(50000) {
       val tree = it.next
       treeAddrs.get(tree) match {
         case addr: Addr => treeAddrs.put(tree, adjusted(addr))
-        case addrs: List[Addr @unchecked] => treeAddrs.put(tree, addrs.map(adjusted))
       }
     }
   }
--- a/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -139,13 +139,13 @@ class TreeChecker extends Phase with SymTransformer {
   class Checker(phasesToCheck: Seq[Phase]) extends ReTyper with Checking {
 
     val nowDefinedSyms = new mutable.HashSet[Symbol]
-    val everDefinedSyms = new mutable.HashMap[Symbol, Tree]
+    val everDefinedSyms = new mutable.HashMap[Symbol, untpd.Tree]
 
     // don't check value classes after typer, as the constraint about constructors doesn't hold after transform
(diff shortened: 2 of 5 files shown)


smarter:
> -- Error: dotty/compiler/../compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala 
542 |    def unapply(tree: Tree @unchecked): Option[Type] = tree match {
    |                           ^^^^^^^^^^
    |                      annotation.unchecked does not have a constructor
-

This looks like a bug but I have no idea why. Does `(tree: @unchecked) match` ? In any case we should open an issue for this bug and try to minimize it.


> -- [E007] Type Mismatch Error: dotty/compiler/../compiler/src/dotty/tools/dotc/transform/PostTyper.scala 
267 |            case ident @ Ident(_)                 => checkIdent(ident)
    |                                                                ^^^^^
 

Dotty is correct here, `ident` refer to a member of `Import#selectors` which is untyped, `checkIdent` needs to be changed to take an untpd `Ident` instead of a typed `Ident`:
```scala
def checkIdent(ident: untpd.Ident): Unit = {
```

compiler/src/dotty/tools/dotc/core/tasty/TreeBuffer.scala
```diff
@@ -159,7 +159,7 @@ class TreeBuffer extends TastyBuffer(50000) {
       val tree = it.next
       treeAddrs.get(tree) match {
         case addr: Addr => treeAddrs.put(tree, adjusted(addr))
-        case addrs: List[Addr] => treeAddrs.put(tree, addrs.map(adjusted))
+        case addrs: List[Addr @unchecked] => treeAddrs.put(tree, addrs.map(adjusted))
```
smarter: Have you tried just removing this line? Looks like the definition of `treeAddrs` has changed and it canno longer contains lists.

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -145,7 +145,7 @@ class TreeChecker extends Phase with SymTransformer {
     override def checkDerivedValueClass(clazz: Symbol, stats: List[Tree])(implicit ctx: Context) = ()
 
     def withDefinedSym[T](tree: untpd.Tree)(op: => T)(implicit ctx: Context): T = tree match {
-      case tree: DefTree =>
+      case tree: DefTree @unchecked =>
```
smarter: If possible, it's nice to add a comment explaining why `@unchecked` is correct here.

allanrenucci: @smarter Not sure why this is correct here. `tree` is an `untpd.Tree` but the type test is on `tpd.DefTree`

smarter: Yeah looking at it, it should be an `untpd.DefTree`. Same thing for `case t: MemberDef` below.

smarter: This also means changing `everDefinedSyms` to be a `mutable.HashMap[Symbol, untpd.Tree]`

allanrenucci:
@smarter I was able to minimise the bug with `@unchecked`. A quick fix is to remove the import
```scala 
import annotation.unchecked

@unchecked
class Test
```

smarter:
@allanrenucci Cool! Could you open an issue?

allanrenucci:
@smarter #2170

smarter:
LGTM!

PR #2169, opened 2017-04-04
Use symbolic names internally

odersky:
Change encoding to that symbolic operators are kept internally and are only emitted during the backend. The PR is based on #2128 of which it is the logical conclusion.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/NameKinds.scala
+++ b/compiler/src/dotty/tools/dotc/core/NameKinds.scala
@@ -62,7 +62,7 @@ object NameKinds {
   }
 
   /** The kind of SimpleNames */
-  object SimpleNameKind extends NameKind(UTF8) { self =>
+  val SimpleNameKind = new NameKind(UTF8) { self =>
     type ThisInfo = Info
     val info = new Info
     def mkString(underlying: TermName, info: ThisInfo) = unsupported("mkString")
@@ -311,7 +311,7 @@ object NameKinds {
   val PatMatSelectorName      = new UniqueNameKind("selector")
 
   /** The kind of names of default argument getters */
-  object DefaultGetterName extends NumberedNameKind(DEFAULTGETTER, "DefaultGetter") {
+  val DefaultGetterName = new NumberedNameKind(DEFAULTGETTER, "DefaultGetter") {
     def mkString(underlying: TermName, info: ThisInfo) = {
       val prefix = if (underlying.isConstructorName) nme.DEFAULT_GETTER_INIT else underlying
       prefix.toString + str.DEFAULT_GETTER + (info.num + 1)
@@ -330,14 +330,14 @@ object NameKinds {
   }
 
   /** The kind of names that also encode a variance: 0 for contravariance, 1 for covariance. */
-  object VariantName extends NumberedNameKind(VARIANT, "Variant") {
+  val VariantName = new NumberedNameKind(VARIANT, "Variant") {
     def mkString(underlying: TermName, info: ThisInfo) = "-+"(info.num).toString + underlying
   }
 
   /** Names of the form N_<outer>. Emitted by inliner, replaced by outer path
    *  in ExplicitOuter.
    */
-  object OuterSelectName extends NumberedNameKind(OUTERSELECT, "OuterSelect") {
+  val OuterSelectName = new NumberedNameKind(OUTERSELECT, "OuterSelect") {
     def mkString(underlying: TermName, info: ThisInfo) = {
       assert(underlying.isEmpty)
       info.num + "_<outer>"
@@ -359,7 +359,7 @@ object NameKinds {
   val ModuleClassName = new SuffixNameKind(OBJECTCLASS, "$", optInfoString = "ModuleClass")
 
   /** A name together with a signature. Used in Tasty trees. */
-  object SignedName extends NameKind(63) {
+  val SignedName = new NameKind(63) {
 
     case class SignedInfo(sig: Signature) extends Info {
       override def toString = s"$infoString $sig"
--- a/compiler/src/dotty/tools/dotc/core/NameKinds.scala
+++ b/compiler/src/dotty/tools/dotc/core/NameKinds.scala
@@ -61,8 +61,7 @@ object NameKinds {
     def infoString: String
   }
 
-  /** The kind of SimpleNames */
-  val SimpleNameKind = new NameKind(UTF8) { self =>
+  object SimpleNameKind extends NameKind(UTF8) { self =>
     type ThisInfo = Info
     val info = new Info
     def mkString(underlying: TermName, info: ThisInfo) = unsupported("mkString")
@@ -359,7 +358,7 @@ object NameKinds {
   val ModuleClassName = new SuffixNameKind(OBJECTCLASS, "$", optInfoString = "ModuleClass")
 
   /** A name together with a signature. Used in Tasty trees. */
-  val SignedName = new NameKind(63) {
+  object SignedName extends NameKind(63) {
 
     case class SignedInfo(sig: Signature) extends Info {
       override def toString = s"$infoString $sig"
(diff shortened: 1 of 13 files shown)


compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -571,8 +571,8 @@ class Namer { typer: Typer =>
 
     /** Create links between companion object and companion class */
     def createLinks(classTree: TypeDef, moduleTree: TypeDef)(implicit ctx: Context) = {
-      val claz = ctx.effectiveScope.lookup(classTree.name.encode)
-      val modl = ctx.effectiveScope.lookup(moduleTree.name.encode)
+      val claz = ctx.denotNamed(classTree.name).symbol
```
smarter: Why does `denotNamed` makes sense here? I changed it to `effectiveScope.lookup` in https://github.com/lampepfl/dotty/commit/b6411818b2e84c0a28c501e2814de4cfde720208 assuming it was the right thing to do. Doesn't using `denotNamed` means that we could find a companion in an outer scope which would be very wrong?

odersky: Good point. I'll revert that change.

odersky:
Status: I went over the whole PR documenting and polishing. Awaiting the backend to be made available so that we can apply the necessary changes to it.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -29,16 +29,29 @@ Macro-format:
...
-                  MANGLED        Length mangle_NameRef name_NameRef
-                  ...
+  Name          = UTF8              Length UTF8-CodePoint*
+                  QUALIFIED         Length qualified_NameRef selector_NameRef
+                  FLATTENED         Length qualified_NameRef selector_NameRef
+                  EXPANDED          Length qualified_NameRef selector_NameRef
+                  EXPANDEDPREFIX    Length qualified_NameRef selector_NameRef
+                  TRAITSETTER       Length qualified_NameRef selector_NameRef
+                  UNIQUE            Length separator_NameRef uniqid_Nat underlying_NameRef?
+									DEFAULTGETTER     Length underlying_NameRef index_Nat
```
smarter: Incorrect indentation here and below.

DarkDimius:
> Awaiting the backend to be made available so that we can apply the necessary changes to it.

PRs by @smarter are all in and the backend has been moved to https://github.com/lampepfl/scala/.

odersky:
The only remaining test failure is:

    Failed tests:
    [error] 	xsbt.ExtractUsedNamesSpecification

I have no idea what that is and can't find the string "ExtractUsedNamesSpecification" anywhere in the repo. @smarter or @felixmulder can you help?

smarter:
@odersky It's here https://github.com/lampepfl/dotty/blob/master/sbt-bridge/test/xsbt/ExtractUsedNamesSpecification.scala

smarter:
Some tests contain mangled name (`$eq`, `$qmark`, ...) which are now returned unmangled. It's safe to change the test to use the unmangled name.

odersky:
@smarter Thanks! I have tried to change to symbolic names.

smarter:
@odersky By the way, the report at the end of the CI output was not helpful here (since this is a regular JUnit test not part of the compiler test framework), but if you search for "ExtractUsedNamesSpecification" in the CI output, you'll find a much more detailed explanation of what failed.

odersky:
Seems to all work out now. Ready for review.

smarter:
Can you rebase now that https://github.com/lampepfl/dotty/pull/2322 has been merged?

odersky:
Rebased

smarter:
There's a commit called "Invalide fullName cache when name changes" (https://github.com/lampepfl/dotty/pull/2169/commits/7593cf3e9e8ee4dcc99c828ded13bb2488f2df8c) that doesn't seem to have anything do with cache invalidation. Did that change appear in another commit, was it lost, or intentionally dropped?

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -244,7 +244,8 @@ class CompilationTests extends ParallelTesting {
     }.map(_.checkCompile()).foreach(_.delete())
   }
 
-  @Test def bytecodeIdempotency: Unit = {
+  /** Add a `z` so that hey run last. TODO: Only run them selectively? */
```
smarter: typo: hey -> they

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -188,103 +233,152 @@ object Names {
...
 
-    def is(kind: NameKind): Boolean = {
+    override def is(kind: NameKind): Boolean = {
       val thisKind = this.info.kind
       thisKind == kind ||
       !thisKind.definesNewName && thisKind.tag > kind.tag && underlying.is(kind)
     }
+
+    @sharable // because it's just a cache for performance
+    private[Names] var myMangledString: String = null
```
smarter: This should be `private[this]`

compiler/src/dotty/tools/dotc/core/NameKinds.scala
```diff
@@ -243,29 +310,28 @@ object NameKinds {
...
         var original = name.take(i - str.DEFAULT_GETTER.length).asTermName
-        if (original == nme.DEFAULT_GETTER_INIT) original = Names.CONSTRUCTOR
+        if (original == nme.DEFAULT_GETTER_INIT) original = nme.CONSTRUCTOR
         apply(original, index)
       }
       else name
     }
   }
 
+  /** The kind of names that also encode a variance: 0 for contravariance, 1 for covariance. */
```
smarter: -1 for contravariance, not 0

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
smarter: Indentation still broken.

smarter:
This isn't necessary for this PR to get in, but I think it would be great if we used different types for encoded and decoded names, to make sure we don't use one instead of the other. This should be possible now that we have a clearer picture of when things should be encoded/decoded.

odersky:
> There's a commit called "Invalide fullName cache when name changes" (7593cf3) that doesn't seem to have anything do with cache invalidation. Did that change appear in another commit, was it lost, or intentionally dropped?

That was intentionally dropped because it's now redundant. We found a way to do it without changing the name so no invalidation is necessary.

compiler/src/dotty/tools/dotc/core/NameKinds.scala
odersky: No, it's 0 now.

smarter:
From http://dotty-ci.epfl.ch/lampepfl/dotty/2165/3:
> [error] Test dotty.tools.dotc.CompilationTests.compilePos failed: java.lang.AssertionError: Expected no errors when compiling, but found: 0, took 54.642 sec

That's a weird error message, /cc @felixmulder

felixmulder:
My guess is that an assertion fired, but unsure. I'd have to investigate

odersky:
Some interesting developments. Making all names strict uncovered a bug in `avoid` which is minimized in `6e3f325`. It's fixed now. I experimented with making `avoid` an approximating typemap, but did not succeed. The current state is in staging/wip-avoid-as-approx-typemap.

compiler/src/dotty/tools/dotc/core/NameKinds.scala
```diff
@@ -61,8 +61,7 @@ object NameKinds {
     def infoString: String
   }
 
-  /** The kind of SimpleNames */
-  val SimpleNameKind = new NameKind(UTF8) { self =>
+  object SimpleNameKind extends NameKind(UTF8) { self =>
```
smarter: I think ultimately, we should just make the name registration thread-safe, it doesn't cost much and it's much easier to reason about than remembering which NameKinds have side-effects, etc.

odersky: But this is another issue. The problem was not threading but the fact that the registration was  not performed at all, since objects are lazy.

smarter: Ah, I see, indeed that's different. If we had enums we could iterate over the namekinds to register them maybe :).

PR #2187, opened 2017-04-04
fix #2166: unpickling of shared CaseDef

### LINKED ISSUE #2166: Inlined pattern matching won't unpickle
```scala
object Test {
  @inline def f = "" match { case _ => false }

  def main(args: Array[String]): Unit = f
}
```

```bash
$ sbt "run -Ytest-pickler -Ystop-after:pickler /home/olivier/a.scala"
[...]
$ cat after-pickling.txt
[cannot display due to java.lang.AssertionError: assertion failed: Addr(134) Addr(138) Addr(140) MATCH, raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@37f8bb67]
```

liufengyun:
Fix #2166: the unpickling of CaseDef should take care of SHARED trees.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -1062,7 +1062,13 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
       }
 
     def readCases(end: Addr)(implicit ctx: Context): List[CaseDef] =
-      collectWhile(nextByte == CASEDEF && currentAddr != end) { readCase()(ctx.fresh.setNewScope) }
+      collectWhile((nextByte == CASEDEF || nextByte == SHARED) && currentAddr != end) {
+        if (nextByte == SHARED) {
+          readByte()
+          forkAt(readAddr()).readCase()(ctx.fresh.setNewScope)
+        }
+        else readCase()(ctx.fresh.setNewScope)
+      }
 
     def readCase()(implicit ctx: Context): CaseDef = {
       val start = currentAddr
--- a/tests/pickling/i2166.scala
+++ b/tests/pickling/i2166.scala
@@ -1,0 +1,5 @@
+object Test {
+  @inline def f = "" match { case _ => false }
+
+  def main(args: Array[String]): Unit = f
+}


smarter:
Missing test case?

liufengyun:
The pickling tests pass, but the bootstrap pickling failed. Need to investigate further.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -1062,7 +1062,13 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
       }
 
     def readCases(end: Addr)(implicit ctx: Context): List[CaseDef] =
-      collectWhile(nextByte == CASEDEF && currentAddr != end) { readCase()(ctx.fresh.setNewScope) }
+      collectWhile(nextByte == CASEDEF | nextByte == SHARED && currentAddr != end) {
```
smarter: I think you want `||`, not `|`

smarter: And you also need to put parens in the correct place because `&&` has higher precedence:
```scala
(nextByte == CASEDEF || nextByte == SHARED) && currentAddr != end)
```

liufengyun: Thanks a lot @smarter , I owe you a beer 👍  I will no longer claim I can write Scala in my CV :)

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -1062,7 +1062,13 @@ class TreeUnpickler(reader: TastyReader, tastyName: TastyName.Table, posUnpickle
       }
 
     def readCases(end: Addr)(implicit ctx: Context): List[CaseDef] =
-      collectWhile(nextByte == CASEDEF && currentAddr != end) { readCase()(ctx.fresh.setNewScope) }
+      collectWhile((nextByte == CASEDEF || nextByte == SHARED) && currentAddr != end) {
```
odersky: You can simply use `nextUnsharedTag` for this. 

    collectWhile(nextUnsharedTag == CASEDEF && currentAddr != end) { readCase()(ctx.fresh.setNewScope) }

odersky:
Sorry, that was a wrong comment. In fact your original version is the correct one.

PR #2190, opened 2017-04-04
Make leak avoidance more robust

odersky:
If class files are missing, finding an underlying class reference might give a NoType. This caused an asInstanceOf to fail. We now handle that case gracefully.

It's a small thing, and I found it too hard to reproduce in a test case. @smarter Since you are already current in `checkNoPrivateLeaks` do you want to take a quick look?


smarter:
Yes I saw that once before too, LGTM.

PR #2191, opened 2017-04-04
Fix #2186: Synchronize classpath handling with Scala 2.12

### LINKED ISSUE #2186: Incorporate scala 2.12 changes to classpath handling

smarter:
This commit is a very crude port of the classpath handling as it exists
in the 2.12.x branch of scalac (hash: 232d95a198c94da0c6c8393624e83e9b9ac84e81),
this replaces the existing Classpath code that was adapted from scalac
years ago.

I'll update this commit with more details and proper attribution later,
for now I just want to run this on the CI.

@odersky I had to add a special case to `PackageLoader` to make sure that Scala2 impl classes (ending with `$class`) were not entered in scope, do you know how we avoided impl classes being in scope so far?

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,23 +1,28 @@
-The dotty compiler frontend has been developed since November 2012 by Martin Odersky. It is expected and hoped for 
+The dotty compiler frontend has been developed since November 2012 by Martin Odersky. It is expected and hoped for
 that the list of contributors to the codebase will grow quickly. Dotty draws inspiration and code from the original
-Scala compiler "nsc", which is developed at [scala/scala](https://github.com/scala/scala). 
+Scala compiler "nsc", which is developed at [scala/scala](https://github.com/scala/scala).
 
 The majority of the dotty codebase is new code, with the exception of the components mentioned below. We have for each component tried to come up with a list of the original authors in the  [scala/scala](https://github.com/scala/scala) codebase. Apologies if some major authors were omitted by oversight.
 
 `dotty.tools.dotc.ast`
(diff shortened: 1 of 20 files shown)


smarter:
> @odersky I had to add a special case to PackageLoader to make sure that Scala2 impl classes (ending with $class) were not entered in scope, do you know how we avoided impl classes being in scope so far?

Responding to myself: this was handled by `DefaultJavaContext#isValidName` in the old classpath implementation, of course there is no equivalent in Scala 2.12 since implementation classes don't exist anymore, but we still have to deal with them as long as we maintain binary compat with Scala 2.11.

smarter:
This is pretty cool: `dotty.tools.dotc.CompilationTests.compilePos` is now twice as fast to complete, I think this is all due to the shared global ClassPath cache: https://github.com/dotty-staging/dotty/blob/6cd429d86bd88d3e7c4d49a62e0d46eef720517c/compiler/src/dotty/tools/dotc/classpath/ZipAndJarFileLookupFactory.scala#L14-L41

smarter:
This should now be ready for review, all new files are copied with some small changes. See the main commit message for more details.

@lrytz, @retronym : Does this look sane to you? Any particular thing we should be aware of?

odersky:
LGTM

DarkDimius:
Related: https://github.com/scala/bug/issues/10289

smarter:
@DarkDimius Thanks for pointing me to that, I opened https://github.com/lampepfl/dotty/issues/2317 to keep track of it.

PR #2193, opened 2017-04-05
Deterministically randomises test compilation order

OlivierBlanvillain:
The previous implementation would compile tests in different orders from machine to machine, depending on the order in which *.scala files are listed by the operating system.

### DIFF
--- a/compiler/test/dotty/tools/dotc/ParallelTesting.scala
+++ b/compiler/test/dotty/tools/dotc/ParallelTesting.scala
@@ -15,6 +15,7 @@ import scala.util.control.NonFatal
 import scala.util.Try
 import scala.collection.mutable
 import scala.util.matching.Regex
+import scala.util.Random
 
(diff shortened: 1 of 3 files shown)


compiler/test/dotty/tools/dotc/ParallelTesting.scala
```diff
@@ -1021,11 +1022,17 @@ trait ParallelTesting { self =>
       if (f.isDirectory) f.listFiles.flatMap(flatten)
       else Array(f)
 
+    val files = {
+      // Deterministically randomises compilation order
+      Random.setSeed(seed)
+      Random.shuffle(flatten(sourceDir).toList).toArray
```
sjrd: Some other thread could mess with the global instance of `Random`. You should use a new instance of `Random` in here.

felixmulder:
:tada:

PR #2194, opened 2017-04-05
Killing partest part 2 - Enter the Vulpix

felixmulder:
fixes: #2183 
fixes: #2210


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -3,18 +3,22 @@ package tools
...
+import vulpix.{ ParallelTesting, SummaryReport }
+
+class CompilationTests extends SummaryReport with ParallelTesting {
   import CompilationTests._
 
-  def isInteractive: Boolean = ParallelSummaryReport.isInteractive
+  // Test suite configuration --------------------------------------------------
 
-  def testFilter: Option[Regex] = sys.props.get("dotty.partest.filter").map(r => new Regex(r))
+  def maxDuration = 180.seconds
```
felixmulder: Note - we have some tests that run for quite a long time. In most cases these are benchmarks. But there are some which aren't and run for up to 60 seconds.

I propose to have a split between these run tests and the regular run tests in:

`./tests/run-long/``

And have an interface for them which allows for longer execution.

compiler/test/dotty/tools/dotc/vulpix/VulpixTests.scala
```diff
@@ -1,13 +1,14 @@
 package dotty
 package tools
 package dotc
+package vulpix
```
smarter: I'm wondering if it shouldn't just be `dotty.tools.vulpix` :)

compiler/test/dotty/tools/dotc/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,81 @@
...
+import java.io.{
+  File => JFile,
+  InputStream, ObjectInputStream,
+  OutputStream, ObjectOutputStream
+}
+
+import scala.concurrent.duration.Duration
+import scala.collection.mutable
+
+trait RunnerOrchestration {
```
smarter: Could you add some documentation explaining what this is and how it works here?

compiler/test/dotty/tools/dotc/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /** Destroy and respawn process after each test */
+  val safeMode: Boolean
+
+  /** Running a `Test` class's main method from the specified `dir` */
+  def runMain(dir: JFile): Status = monitor.runMain(dir)
+
+  private[this] val monitor = new RunnerMonitor
+
+  private class RunnerMonitor {
```
smarter: Docs here too plz

compiler/test/dotty/tools/dotc/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,81 @@
...
+    private def createProcess: Process = ???
+
+    private[this] val allRunners =
+      List.fill(numberOfSlaves)(new Runner(createProcess))
+
+    private[this] val freeRunners = mutable.Queue(allRunners: _*)
+    private[this] val busyRunners = mutable.Set.empty[Runner]
+
+    private def getRunner(): Runner = synchronized {
+      while (freeRunners.isEmpty) wait()
```
smarter: This looks correct, but I've forgotten everything I ever knew about Java concurrency, are you sure there's no existing higher-level concurrency thing from https://docs.oracle.com/javase/8/docs/api/?java/util/concurrent/package-summary.html you could use instead of the wait/notify dance ?

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -6,15 +6,20 @@ import org.junit.Test
...
 class CompilationTests extends SummaryReport with ParallelTesting {
   import CompilationTests._
 
-  def isInteractive: Boolean = SummaryReport.isInteractive
+  // Test suite configuration --------------------------------------------------
 
-  def testFilter: Option[Regex] = sys.props.get("dotty.partest.filter").map(r => new Regex(r))
+  def maxDuration = 3.seconds
+  def numberOfSlaves = 5
+  def safeMode = sys.env.get("SAFEMODE").isDefined
```
smarter: I would prefer `dotty.tests.safemode`

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -6,15 +6,20 @@ import org.junit.Test
...
+  def isInteractive = SummaryReport.isInteractive
+  def testFilter = sys.props.get("dotty.partest.filter").map(r => new Regex(r))
```
smarter: And here `dotty.tests.filter`, that way all the properties related to testing are in the `dotty.tests` namespace

smarter: By the way, we should have documentation for all these properties in one place, and it would be good if we limited string handling to one place too, I suggest having something like:
```scala
object DottyTestsProperties {
  // *Documentation of what safeMode is*
  def safeMode = sys.props.get("dotty.partest.filter")
  ...
}
```

compiler/test/dotty/tools/dotc/vulpix/Statuses.java
```diff
@@ -0,0 +1,20 @@
...
+        public final String output;
+        public Success(String output) { this.output = output; }
+    }
+
+    static class Failure implements Status, Serializable {
+        public final String output;
+        public Failure(String output) { this.output = output; }
+    }
+
+    static class Timeout implements Status, Serializable {}
```
smarter: Is this something that a child JVM can send? Otherwise it should not be part of `Status` assuming `Status` defines the communication protocol between the JVMs.

compiler/test/dotty/tools/dotc/vulpix/RunnerOrchestration.scala
```diff
@@ -52,37 +70,51 @@ trait RunnerOrchestration {
...
-
-        // return run status:
         status
       }
     }
 
-    private def createProcess: Process = ???
+    private def createProcess: Process = {
+      val sep = sys.props("file.separator")
+      val cp = sys.props("java.class.path")
```
smarter: This makes me sad, but I won't disallow that for now.

compiler/test/dotty/tools/dotc/vulpix/ChildMain.scala
```diff
@@ -0,0 +1,83 @@
...
+
+object ChildMain {
+  val realStdin = System.in
+  val realStderr = System.err
+  val realStdout = System.out
+
+  private def runMain(dir: JFile): Status = {
+    def renderStackTrace(ex: Throwable): String =
+      ex.getStackTrace
+        .takeWhile(_.getMethodName != "invoke0")
```
smarter: What is this for? I think I'd be happier not having my stack traces mangled in some undocumented way, what if I have a test with an invoke0 method?

.drone.yml
```diff
@@ -34,9 +34,6 @@ pipeline:
 
 matrix:
   TEST:
-    - ;set testOptions in LocalProject("dotty-compiler") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.hydra.ParallelTesting") ;test ;dotty-bin-tests/test
-    - ;set testOptions in LocalProject("dotty-compiler-bootstrapped") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.hydra.ParallelTesting") ;publishLocal ;dotty-bootstrapped/test
-    - ;set testOptions in LocalProject("dotty-compiler") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.hydra.ParallelTesting") ;partest-only-no-bootstrap --show-diff --verbose
-    - ;set testOptions in LocalProject("dotty-compiler-bootstrapped") += Tests.Argument(TestFrameworks.JUnit, "--exclude-categories=dotty.tools.dotc.hydra.ParallelTesting") ;partest-only --show-diff --verbose
-    - ;dotty-compiler/testOnly dotty.tools.dotc.CompilationTests
-    - ;publishLocal ;dotty-bootstrapped/testOnly dotty.tools.dotc.CompilationTests
+    - test
```
smarter: We should still be running the old JUnit tests with the CI.

compiler/test/dotty/tools/dotc/vulpix/VulpixTests.scala
```diff
@@ -1,15 +1,20 @@
...
 
 import org.junit.Assert._
 import org.junit.Test
 
+import scala.concurrent.duration._
 import scala.util.control.NonFatal
 
-class ParallelTestTests extends ParallelTesting {
+class VulpixTests extends ParallelTesting {
```
smarter: Please add some documentation on what this is for.

compiler/test/dotc/comptest.scala
```diff
@@ -1,9 +1,14 @@
...
-import dotty.tools.dotc.ParallelTesting
+import dotty.tools.dotc.vulpix.ParallelTesting
+
+import scala.concurrent.duration._
 
 object comptest extends ParallelTesting {
 
+  def maxDuration = 3.seconds
+  def numberOfSlaves = 5
+  def safeMode = false
```
smarter: Why is this different from the other override of safeMode?

compiler/test/dotty/tools/dotc/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,153 @@
...
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.collection.mutable
+
+import vulpix.Statuses._
+
+trait RunnerOrchestration {
+
+  /** The maximum amount of active runners, which contain a child JVM */
+  def numberOfSlaves: Int
```
smarter: This is always overriden as 5, why not make it the default?

smarter: Also, why isn't this based on the number of CPUs?

compiler/test/dotc/comptest.scala
felixmulder: This is a legacy main used by some folks in the lab - it's only running a single test. No need to restart VMs after this single test.

odersky:
For the moment I cannot yet run CompilationTests instead of the old `dotc.test` as my standard test suite from Eclipse. There are three problems:

 1. My console window looks like this:

    [======================================>] compiling (205/206, 14s)
    [======================================>] compiling (205/206, 14s)
    [======================================>] compiling (205/206, 14s)
    [======================================>] compiling (205/206, 15s)
    [======================================>] compiling (205/206, 15s)
    [======================================>] compiling (205/206, 15s)
    [======================================>] compiling (205/206, 15s)
    [======================================>] compiling (205/206, 15s)
    
  ... and so on with 10 new entries per second. This fills all available lines and crowds out useful output.

 2. When a test fails I see something like this:

    java.lang.AssertionError: Neg test shouldn't have failed, but did
	at org.junit.Assert.fail(Assert.java:88)

  It should at least tell me the name of the test that failed.

 3. It would be good to have an option to disable running tests. It takes a lot of time and if I work on the frontend I am rarely interested in that aspect.

felixmulder:
> 1. My console window looks like this:

Hopefully fixed by 4d76b10

Gotta board the plane! Will look at the rest on board

felixmulder:
@odersky:

Fixed **(1)** in 84917a3 instead. You'll now need to set property "dotty.tests.interactive" to "FALSE" to have the tests run without interactive progress bar.

Apparently eclipse had to revert a fix that handled `\r`. It's the 21st century, what's going on?

**(2)**: could you gist or otherwise add the console output here? You can use

```html
<details>paste all the things here</details>
```

Which results in:

<details>paste all the things here</details>

> 3. It would be good to have an option to disable running tests. It takes a lot of time and if I work on the frontend I am rarely interested in that aspect.

How do you want to run this, from commandline or from eclipse? An sbt alias like `testFrontend` could be added, but then you'd have to invoke it from sbt.

We could also do it via properties or env...

smarter:
> We could also do it via properties or env...

I'd say, make everything a flag and a property, so people who use IDE can set the property, and people who use sbt can use the flag which is more convenient.

DarkDimius:
>>    We could also do it via properties or env...

>I'd say, make everything a flag and a property, so people who use IDEs can set the property, and people who use sbt can use the flag which is more convenient.

I'm not sure if having two side-channels is a good thing. I'd strongly prefer to have no more than one.

smarter:
> I'm not sure if having two side-channels is a good thing. I'd strongly prefer to have no more than one.

No it's fine, the sbt flag just needs to set the property, so you only have one side-channel in the end.

compiler/test/dotty/tools/vulpix/SummaryReport.java
```diff
@@ -10,8 +10,10 @@
 import dotty.tools.dotc.reporting.TestReporter;
 import dotty.Properties;
 
-/** Note that while `ParallelTesting` runs in parallel, JUnit tests cannot with
- *  this class
+/** This class adds summary reports to `ParallelTesting`
+ *
+ *  It is written in Java because we currently cannot explicitly write static
+ *  methods in Scala without SIP-25 (`@static` fields and methods in Scala)
```
smarter: If you only need static methods, isn't using a scala-defined object sufficient? The compiler will emit a corresponding class with static methods forwarding to the object's methods for interoperability with Java.

felixmulder: Doesn't work with JUnit.

smarter: But why though.

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -387,129 +428,107 @@ trait ParallelTesting { self =>
...
-      import java.net.{ URL, URLClassLoader }
-
-      val printStream = new ByteArrayOutputStream
+    private[this] var didAddNoRunWarning = false
+    private[this] def addNoRunWarning() = if (!didAddNoRunWarning) {
+      didAddNoRunWarning = true
+      summaryReport.addStartingMessage {
+        """|WARNING
+           |-------
+           |Run tests were only compiled, not run - this is due to `dotty.tests.norun`
```
smarter: due to _the_

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -522,49 +541,64 @@ trait ParallelTesting { self =>
...
         }
 
-        if (errorCount == 0 && hasCheckFile) verifier()
-        else if (errorCount == 0) runMain(testSource.outDir, testSource)
-        else if (errorCount > 0) {
-          echo(s"\nCompilation failed for: '$testSource'")
+        if (!compilerCrashed && errorCount == 0 && hasCheckFile) verifier()
+        else if (!compilerCrashed && errorCount == 0) {
+          if (Properties.testsNoRun) addNoRunWarning()
+          else runMain(testSource.classPath) match {
```
smarter: I don't like the repetition with `verifyOutput` above here, would be nicer if we pretended that all tests had a check file (no check file would be equivalent to empty check file).

compiler/test/dotty/tools/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,195 @@
...
+
+      /** Destroys the underlying process and kills IO streams */
+      def kill(): Unit = {
+        if (process ne null) process.destroy()
+        process = null
+        childStdout = null
+        childStdin = null
+      }
+
+      private[this] var didAddCleanup = false
```
smarter: Can you document this variable, and maybe give it a nicer name?

compiler/test/dotty/tools/vulpix/RunnerOrchestration.scala
```diff
@@ -0,0 +1,195 @@
...
+        val status =
+          try Await.result(readOutput, maxDuration)
+          catch { case _: TimeoutException =>  Timeout }
+
+        // Handle failure of the VM:
+        status match {
+          case _: Success if safeMode => respawn()
+          case _: Failure => respawn()
+          case Timeout => respawn()
+          case _ => ()
```
smarter: When does that case happen?

compiler/test/dotty/tools/vulpix/Status.scala
```diff
@@ -0,0 +1,7 @@
+package dotty.tools
+package vulpix
+
+sealed trait Status
+final case class Success(output: String) extends Status
+final case class Failure(output: String) extends Status
+final case object Timeout extends Status
```
smarter: What's the difference between Timeout and Failure("Process timed out after ..."), do we really need a special case for Timeout ?

compiler/test/dotty/tools/vulpix/SummaryReport.scala
```diff
@@ -0,0 +1,94 @@
+package dotty
+package tools
+package vulpix
+
+import scala.collection.mutable
+import dotc.reporting.TestReporter
+
+trait SummaryReporting {
```
smarter: Could you add some documentation in this file, what's a SummaryReporting? What's a SummaryReport? Also, the fact that these are two different classes with very similar name is confusing, maybe use SummaryReport and DefaultSummaryReport; I'm not sure.

compiler/test/dotty/tools/vulpix/SummaryReport.scala
```diff
@@ -0,0 +1,94 @@
...
+          |
+          |$passed passed, $failed failed, ${passed + failed} total
+          |""".stripMargin
+    )
+
+    startingMessages.foreach(rep.append)
+
+    failedTests.map(x => "    " + x).foreach(rep.append)
+
+    // If we're compiling locally, we don't need reproduce instructions
```
smarter: reproduction, not reproduce

smarter: Also, I disagree, I want reproduction instructions, please make it an option at least :).

compiler/test/dotty/tools/vulpix/TestConfiguration.scala
```diff
@@ -0,0 +1,67 @@
+package dotty
+package tools
+package vulpix
+
+object TestConfiguration {
+  implicit val defaultOutputDir: String = "../out/"
+
+  implicit class RichStringArray(val xs: Array[String]) extends AnyVal {
+    def and(args: String*): Array[String] = {
```
smarter: Why?

compiler/test/dotty/tools/vulpix/VulpixTests.scala
```diff
@@ -55,4 +61,16 @@ class ParallelTestTests extends ParallelTesting {
 
   @Test def runOutRedirects: Unit =
     compileFile("../tests/partest-test/i2147.scala", defaultOptions).expectFailure.checkRuns()
+
+  @Test def infiteNonRec: Unit =
```
smarter: typo: infite -> infinite, same below

project/Build.scala
```diff
@@ -340,11 +285,11 @@ object Build {
         )
       }.evaluated,
 
-      filterTest := Def.inputTaskDyn {
+      vulpix := Def.inputTaskDyn {
         val args: Seq[String] = spaceDelimited("<arg>").parsed
-        testOptions := Seq()
+        val realArgs = if (args.nonEmpty) args else Seq(".*")
```
smarter: Please, just change the filter to contain the regexp you pass as argument instead of being equal to it :).

PR #2196, opened 2017-04-05
[WIP] avoid agressive eliding to make code re-typable

liufengyun:
Def macros require typed trees to be re-typable. Previously, the following untyped tree:

    Seq(2,1).sum

will be typed as follows:

     Seq(2, 1).sum(IntIsIntegral)

which cannot be typed again, as `IntIsIntegral` cannot be resolved. This commit
will make the tree typed as follows:

     Seq(2, 1).sum(scala.math.Numeric.IntIsIntegral)


odersky:
We have to discuss more the principle of this. Retyping in current dotty means leaving symbols and types of leaf nodes intact. It would be a huge undertaking to change that, with completely uncertain outcome.

PR #2197, opened 2017-04-05
Add enum exhaustivity checking

odersky:
This PR adds exhaustivity checking for enum values on top of #1958. The first new commit is 	b34a95a. It adds child annotations for sealed enum classes that refer to their enum values. What's new is that the symbols referenced by such annotations can be term symbols. Currently the exhaustivity checking does not deal with term symbols, so some tests fail.

@liufengyun Can you take this PR over and adapt exhaustivity checking?


liufengyun:
Thanks @odersky , I'll take over and make it work.

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -104,8 +104,14 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTran
...
       annot.derivedAnnotation(transformAnnot(annot.tree))
 
+    private def registerChild(sym: Symbol, tp: Type)(implicit ctx: Context) = {
+      val cls = tp.classSymbol
+      if (cls.is(Sealed)) cls.addAnnotation(Annotation.makeChild(sym))
+    }
+
     private def transformMemberDef(tree: MemberDef)(implicit ctx: Context): Unit = {
       val sym = tree.symbol
+      if (sym.is(CaseVal, butNot = Module | Method)) registerChild(sym, sym.info)
```
liufengyun: Nice we have this flag, it simplifies a lot of things 👍

odersky:
Very nice simplifications! LGTM

PR #2199, opened 2017-04-07
Redesign Tuples with HList-like structure

OlivierBlanvillain:
This PR redesigns Tuples to use an HList-like structure instead of flat case classes (`scala.TupleN`):

```scala
package dotty

trait Tuple
trait TupleCons[+H, +T <: Tuple] extends Tuple
object Unit extends Tuple

object TupleCons {
  def apply[H, T <: Tuple](h: H, t: T): TupleCons[H, T] = ...
  def unapply[H, T <: Tuple](t: TupleCons[H, T]): Pair[H, T] =...
}
```

These types (and values) are users facing: desugaring turns `(a, b, ...)` into `TupleCons[a, TupleCons[b, ...Unit]]` (and `TupleCons(a, TupleCons(b, ...()))`). At run-time, tuples are represented using case classes for small arities and an Array for large arities.

A new mini-phase named `TupleRewrites` takes care of optimizing `apply` and `unapply` expressions corresponding to creation and pattern matching on tuples. To preserve binary compatibility, `Scala2Unpickler` is modified to transform tuple types to the new representation, `TypeErasure` then takes care of erasing (small) tuple types back to `scala.TupleN`.

TODO:

- [ ] Restore `PatmatExhaustivityTests`. The exhaustivity checks need to be updated to properly handle the new representation.
- [ ] Restore `runCompilerFromInterface`. This now needs an explicit dotty-library dependency. It used to be inherited "automatically" from the `dependsOn(dotty-library)` in `dotty-compiler`.
- [ ] Change patdef desugaring? The current implementation uses .unapply + direct calls to `_i` which is still going to be capped at 22. We could lift that by desugaring patdef using vars and pattern matching instead.
- [ ] Restore pretty printing for tuples.


OlivierBlanvillain:
@felixmulder any idea why the CI decided not to test the latest commit?

felixmulder:
No idea - GitHub gets messy with reordered commits in rebases I've
noticed...Could be?

Otherwise - just change the hash of the latest commit and force push,
should get it going again. Also - CI runs a maximum of 20 jobs, currently
that means 3 commits concurrently.

tests/pos/t0301.scala
```diff
@@ -6,7 +6,7 @@ case class Var() extends Expr
 object Analyzer {
   def substitution(expr: Expr, cls: (Var,Var)): Expr =
     expr match {
-      case cls._2 => cls._1  // source of the error
+      case e if e == cls._2 => cls._1  // source of the error
```
julienrf: Why this change?

tests/run/double-pattern-type.scala
```diff
@@ -0,0 +1,40 @@
...
+
+    val c2 = C2("a", "b")
+    c2 match {
+      case C2(a, b) =>
+        assert(a == "a")
+        assert(b == "b")
+    }
+
+    c2 match {
+      case C2(a, b, c) =>
```
julienrf: Could you detail what this one expands to?

sjrd:
> GitHub gets messy with reordered commits in rebases I've noticed

It always does that. GitHub has this stupid notion that the commits of a PR should be displayed in timestamp order, instead of in history order. It's only a display thing, though, so I doubt it would affect your CI (it certainly never affected ours).

tests/tuples/tuple-accessors.scala
```diff
@@ -0,0 +1,204 @@
...
+    assert(v21 == 21)
+    assert(v22 == 22)
+
+    v1: Int
+    v2: Int
+    v21: Int
+    v22: Int
+
+    // We can't go above that since patDef is implemented with direct _i calls.
+    // It's should be possible to update the desugaring to use var + pattern
```
julienrf: Typo: “it should”

tests/tuples/tuple-appender.scala
```diff
@@ -0,0 +1,58 @@
...
+    import syntax.append._
+
+    val l1: (String, Boolean) =
+      ("s", true)
+
+    val l2: (Double, Double, Double) =
+      (1d, 2d, 3d)
+
+    val l3: (String, Boolean, Double, Double, Double) =
+      l1 ++ l2
```
julienrf: :heart:

tests/pos/t0301.scala
OlivierBlanvillain: Since `_2` is now added thought implicit conversion, the pattern used in this case is no longer stable. ([see commit message](https://github.com/lampepfl/dotty/pull/2199/commits/d06e3203c843e24578368a996287bea61b466a88))

tests/run/double-pattern-type.scala
OlivierBlanvillain: That's unchanged. (the first commits come from #1938 which this PR is rebased on to be able to test case class with size > 23)

DanyMariaLee:
Great news!! Will also be available everything from shapeless to operate with HLists the way we used to?

szeiger:
I'd like to point you to https://github.com/szeiger/ErasedTypes which I wrote half a decade ago with the purpose of eventually replacing Scala tuples with HArrays.  It contains abstractions over tuple arity that may be useful to add (although some will not work in the same way with Dotty making arbitrary type projections illegal). It also has a clean separation between cons-style HLists and flat HArrays (but I recommend looking at typical use cases to determine whether or not this should be preferred over the implementation here which combines the two).

tests/tuples/tuple-accessors.scala
```diff
@@ -120,7 +120,7 @@ object Test {
     v22: Int
 
     // We can't go above that since patDef is implemented with direct _i calls.
-    // It's should be possible to update the desugaring to use var + pattern
+    // Its should be possible to update the desugaring to use var + pattern
```
julienrf: It?

OlivierBlanvillain:
@DanyMariaLee It seems like the current trend is to remove things from the standard library, not add more, so I'm not really sure it shapeless style operations on tuples would have a place in the standard library.

Miles has been talking for a while about a possible shapeless redesign based on Symmetric Monoidal Category ([see this branch](https://github.com/milessabin/shapeless/tree/wip/smc)) that would allow operations like HList#length and Coproduct#length to be factored out into a single implementation over a common structure. The tuple types introduced in this PR would fit perfectly with this design!

OlivierBlanvillain:
@szeiger Thanks for pointing out your HArrays project, it's something I studied briefly before I started working on this. I think the main difference with this PR is that here nothing is intended to be exposed to users besides the `TupleCons` and `Unit` types, the fact that tuples are backed by case classes then by arrays is an implementation detail.

The question of array vs linked list representation is easily answered with a binary compatibility requirement: it wouldn't make much sense to us a flat representation up to size 22, then switch to linked afterward. However, if we lower that limit to a more reasonable size of say 6, then the question opens again :)

szeiger:
When it comes to such low-level constructs as tuples you may not want to hide too much of the details for performance reasons. I haven't done any benchmarking but exposing the TupleN classes should in theory enable better optimizations because you get monomorphic dispatch.

The motivation for having both HList and HArray in my implementation was not a mixed implementation of tuples depending on size. It doesn't do that (tuples are always flat) and for predictable performance I think it shouldn't do it. But at the type level HArrays are nested, so if you already have a proper HList type you can use that as part of the HArray type, i.e. "here's a flat implementation that corresponds to the following HList". Building up from Nat to HList and finally HArray was a natural thing to do given that my goal was not only to remove the Tuple22 limit but also to allow useful abstractions over arity.

Why would you still have special implementations up to 22 here anyway? I think you should lower that number considerably. 22 was originally chosen as large enough to represent what Scala needed, not because of performance considerations. The latter should be the driving factor for determining a good size for Dotty if you support arbitrary tuple sizes anyway. You are not bound by binary compatibility and you have Scalafix to take care of source compatibility.

DarkDimius:
> Why would you still have special implementations up to 22 here anyway? I think you should lower that number considerably. 22 was originally chosen as large enough to represent what Scala needed, not because of performance considerations. 

While originally 22 wasn't chosen due to performance considerations, It would be nice to keep performance properties of existing code. I'd prefer to keep 22 limit in place and not lower it.

OlivierBlanvillain:
> Why would you still have special implementations up to 22 here anyway?

The current very practical reason is that dotty test suite currently uses the 2.11.x standard library compiled by scalac. Methods like Map.apply and Map.head accept and return scala.Tuple2 instances, which is handled in this PR by preserving binary compatibility.

> I think you should lower that number considerably.

I definitely agree. I did a quick search on a large corpus of open-source Scala project (~600K LOC), and every tuple instantiated, pattern-matched on or accessed with _i explicitly was on tuple of size 6 or below. Above that, it's only generic operations (for example, [in Play](https://github.com/playframework/play-json/blob/master/play-functional/src/main/scala/play/api/libs/functional/Products.scala)), for which both array and linked list representation should be superior. It also seems that the majority of generic operations are based on head & tail, which obviously works better on a linked representation.

> TupleN classes should in theory enable better optimizations because you get monomorphic dispatch.

The tuple interface has no methods, so they shouldn't be dynamic dispatch happening on these tuples. Non-generic operations with statically known length use a cast and a direct call, generic operations must rely on pattern matching. If we imagine an up to size 6 case classes then linked list representation, and always put the `TupleCons` case first, we should get about the same performance than a linked list for head & tail.

I should probably plug in a few benchmarks I did just on the representations themselves: [bench.pdf](https://github.com/lampepfl/dotty/files/958383/bench.pdf). Each representation is special cased up to size 4 with case classes. `UnrolledHList` is a linked list with 3 different cons nodes, holding 1, 2 and 3 heads respectively ([actual definitions](https://github.com/OlivierBlanvillain/generic/blob/master/jmh-bench/src/main/scala/HListDefs.scala)).

odersky:
Is this ready for review? In this case it would be good to rebase first.

OlivierBlanvillain:
@odersky yes! The last missing part is to update pattern matching exhaustivity. The current implementation is special cased for tuples; to obtain the same warnings when using hlists this code would need to be either generalised for GADTs or special cased for the new structure.

mdedetrich:
> Why would you still have special implementations up to 22 here anyway? I think you should lower that number considerably. 22 was originally chosen as large enough to represent what Scala needed, not because of performance considerations. The latter should be the driving factor for determining a good size for Dotty if you support arbitrary tuple sizes anyway. You are not bound by binary compatibility and you have Scalafix to take care of source compatibility.

AFAIK, 22 was chosen actually due to the Scala runtime size. The blowup in Jar size for the various Tuple/Product/Function etc classes that get generated is polynomial, and once you start going above 22 you start adding megs onto the Scala runtime

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+                    Some((head :: args, FoldedTupleType(fstTpe.tpe, foldedTailType)))
+
+                case _ => None
+              }
+          case _ => None
+        }
+    }
+
+    tree match {
+      case TupleApplies(args, types) =>
```
odersky: `TupleApply` seems a better name.

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+   *  will go thought this transformation `N` times, thus generating `N` `TupleCons(a, opt)` where `opt`
+   *  is the optimized transformation for previous arity.
+   */
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    // Matches a tree with shape `TupleCons.apply(head, tail)` where `tail` itself a tuple
+    // with statically known lenght (Unit, DottyTuple1, DottyTuple2...). */
+    object TupleApplies {
+      def unapply(tree: Apply)(implicit ctx: Context): Option[(List[Tree], TupleType)] =
+        tree match {
+          case Apply(TypeApply(Select(ident, nme.apply), fstTpe :: _), head :: tail :: Nil)
```
odersky: `head :: tail :: Nil` should be `fst :: snd :: Nil` 

`tail` is not the right name for an element.

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+            if ident.symbol == defn.TupleConsModule =>
+              tail match {
```
odersky: It's probably better to introduce `TupleCons_apply` as a symbol in `Definitions`. Then you can test for it directly here:

    case Apply(TypeApply(fn, fstTpe :: _), head :: tail :: Nil) if fn.symbol == defn.TupleCons_apply

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+                case Literal(Constant(())) =>
+                    Some((head :: Nil, UnfoldedTupleType(fstTpe.tpe :: Nil)))
+
+                case Typed(Apply(TypeApply(Select(tailIdent, nme.apply), tailTpes), args), tt)
+                  if defn.DottyTupleNModule contains tailIdent.symbol =>
```
odersky: All tests `defn.DottyTupleNModule.contains(...)` should be replaced by a more efficient test method, which should be placed in `Definitions`.

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+                    Some((head :: args, UnfoldedTupleType(fstTpe.tpe :: tailTpes.map(_.tpe))))
+
+                case Typed(Apply(TypeApply(Select(tailIdent, nme.wrap), tailTpes), SeqLiteral(args, _) :: Nil), _)
+                  if tailIdent.symbol == defn.LargeTupleModule =>
+                    val foldedTailType = defn.TupleConsType.safeAppliedTo(tailTpes.map(_.tpe))
```
odersky: Why `safeAppliedTo`? Can't we use `appliedTo`?

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
odersky: Also: Maybe reverse `types` and `args`. After all types come first in an apply.

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+            if ident.symbol == defn.TupleConsModule =>
+              tail match {
+                case Literal(Constant(())) =>
+                    Some((head :: Nil, UnfoldedTupleType(fstTpe.tpe :: Nil)))
+
+                case Typed(Apply(TypeApply(Select(tailIdent, nme.apply), tailTpes), args), tt)
+                  if defn.DottyTupleNModule contains tailIdent.symbol =>
+                    Some((head :: args, UnfoldedTupleType(fstTpe.tpe :: tailTpes.map(_.tpe))))
+
+                case Typed(Apply(TypeApply(Select(tailIdent, nme.wrap), tailTpes), SeqLiteral(args, _) :: Nil), _)
```
odersky: These cases feel a bit clunky. How about making an extractor for dotty.TupleN applications?

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+  // the scala.tupleN one, indeed, with C <: B one could write the following:
+  //
+  // TupleCons[A, TupleCons[B, Unit]](a, TupleCons[C, Unit](c, ()))
+  //
+  // This transformation keeps the more precise type, such that the example
+  // above is rewritten to scala.Tuple2[A, C](a, b) (using C and not B).
+  private object TupleUnapplies {
+    def unapply(tree: UnApply)(implicit ctx: Context): Option[(List[Tree], TupleType)] =
+      tree match {
+        case UnApply(TypeApply(Select(selectIndent, nme.unapply), fstTpe :: _), Nil, fstPat :: sndPat :: Nil)
```
odersky: Analogous remarks as for the `TuplyApplies` case apply.

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
odersky: Also, since `transformUnApplyPatterns` is its own method, do the same here for symmetry?

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+        Typed(unapply, tpe)
+      case _ => tree
+    }
+
+  // Create an `UnApply` tree from a list of patters, used in both transformUnApply and transformTyped.
+  private def transformUnApplyPatterns(tree: Tree, patterns: List[Tree], types: TupleType)(implicit ctx: Context): UnApply = {
+    val arity = patterns.length
+    if (arity <= MaxImplementedTupleArity) {
+      val unfoldedTypes: List[Type] = types.unfolded.types
+      val refinedType  = defn.TupleNType(arity).safeAppliedTo(unfoldedTypes)
```
odersky: `appliedTo`, unless you encountered a `CyclicReference` -- but I would be surprised if you did! Note that `safeAppliedTo` is not uniformly better than `appliedTo` -- in fact it is worse because it makes up hypothetical types instead of forcing.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -758,10 +758,18 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         else errorType(i"wrong number of parameters, expected: ${protoFormals.length}", tree.pos)
 
       /** Is `formal` a product type which is elementwise compatible with `params`? */
-      def ptIsCorrectProduct(formal: Type) = {
+      def ptIsCorrectProduct(formal: Type): Boolean = {
+        val cons = defn.TupleConsType.symbol
```
odersky: `defn.TupleConsClass` (once it is introduced, see other comment)

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -758,10 +758,18 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+
+        // Flatten types nested in TupleCons as a List[Type].
+        def tupleType(t: Type, acc: List[Type] = Nil): List[Type] =
```
odersky: Rename to `tupleTypeArgs`

compiler/src/dotty/tools/dotc/transform/TupleRewrites.scala
```diff
@@ -0,0 +1,215 @@
...
+import core.Definitions.MaxImplementedTupleArity
+import core.StdNames._
+import core.Symbols._
+import core.Types._
+import transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/** Local rewrites for tuple expressions. Nested apply and unapply trees coming
+ *  from desugaring into single apply/unapply nodes on DottyTupleN/LargeTuple.
+ */
+class TupleRewrites extends MiniPhaseTransform {
```
odersky: I did not understand why we need DottyTuple in addition to scala.TupleN classes. It would be good to document this.

library/src_dotc/dotty/DottyPredef.scala
```diff
@@ -0,0 +1,471 @@
+package dotty
```
odersky: I am not sure about the `src_dotc` scheme or the other aspects of the build. Let's discuss this aspect with the dotty team as a whole.

library/src_dotc/dotty/DottyPredef.scala
```diff
@@ -0,0 +1,471 @@
...
+
+  implicit class Tuple1Assessors[A](l: TC[A, Unit]) {
+    def _1 = l match { case TC(x, _) => x }
+  }
+
+  type Tuple1[A] = TC[A, Unit]
+
+  def Tuple1[A](a: A): TC[A, Unit] = TC(a, ())
+
+  implicit class Tuple2Assessors[A, B](l: TC[A, TC[B, Unit]]) {
```
odersky: `Assessors`? Should it not be `Accessors`?

compiler/src/dotty/tools/dotc/config/Config.scala
```diff
@@ -140,7 +140,7 @@ object Config {
   final val LogPendingUnderlyingThreshold = 50
 
   /** How many recursive calls to isSubType are performed before logging starts. */
-  final val LogPendingSubTypesThreshold = 50
+  final val LogPendingSubTypesThreshold = 70
```
OlivierBlanvillain: I have some tests with size 24 tuples that triggers this limit, I'm going to revert this change and run these test without `-Yno-deep-subtypes`.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -362,14 +361,9 @@ object Erasure extends TypeTestsCasts{
       val sym = if (owner eq origSym.owner) origSym else owner.info.decl(origSym.name).symbol
       assert(sym.exists, origSym.showLocated)
 
-      def select(qual: Tree, sym: Symbol): Tree = {
-        val name = tree.typeOpt match {
```
OlivierBlanvillain: `name` is not used anywhere in this method: this is dead code.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -409,7 +403,13 @@ object Erasure extends TypeTestsCasts{
         }
       }
 
-      recur(typed(tree.qualifier, AnySelectionProto))
+      if (defn.DottyTupleNModule contains tree.qualifier.symbol) {
```
OlivierBlanvillain: What about making a `Set` version of `DottyTupleNModule`? This test is also used a few times in TupleRewrites.

julienrf:
What’s the status of this PR?

Also, this might be out of the scope of this PR but I think you should not stop in the mid-way and do the same for sealed traits ;) (using `Either` instead of tuples).

OlivierBlanvillain:
@julienrf Some of the complexity of this PR is due to have dotty as cross complied scalac/dotty project. Since going for a bootstrap only compiler is very close on the road-map I'm going to wait for that before coming back to it.

Sum types will hopefully be the follow up :)

antonkulaga:
>Why would you still have special implementations up to 22 here anyway? I think you should lower that number considerably. 22 was originally chosen as large enough to represent what Scala needed

Actually, I hitted 22 fields limit in some real-case scenarios (when I had to write large case classes to parse large cvs-s)

joan38:
Hi guys,
Any news on this feature? Is it still just paused?
Cheers

ctongfei:
I wonder if it is possible to create a typesafe `apply` method on this `HList`-like `Tuple` type.

Something like 
```scala
trait Tuple { self: A =>
  def apply[I <: Int & Singleton, X](i: I)
    (implicit ev: TupleIndexer.Aux[A, I, X]): X = ev(self)
}
```

So you could do the following at compile time:
```scala
val t: (Int, String, Double)
t(0) // type: Int
t(1) // type: String
t(2) // type: Double
t(3) // ERROR: does not compile
```

OlivierBlanvillain:
@joan38 It's postponed until scalac can emit TASTY and we can recompile everything from TASTY, given than the stdlib and thus the entire world has a binary dependency on Tuple2. Until then, the hacks needed in this PR to retain binary compatibility are a bit too much IMO.

joan38:
Thanks @OlivierBlanvillain for the update. So I guess until 2.14 millstone appears? This feature is really promising.

Glavo:
I am worried that this change will slow down the compilation speed.

felixmulder:
@Glavo - don't be :)

Also, dotty has a benchmarking suite that is being used for a lot of pull requests. So now the compiler team can actually know if something would slow down the compiler before merging.

nicolasstucki:
@OlivierBlanvillain now we have another implementation of the same idea, can we close this PR? Or is there something useful that was not covered by the current implementation?

allanrenucci:
Maybe we should extract the test cases

joan38:
@nicolasstucki Could you point out to the other implementation?

allanrenucci:
@joan38 #4938

joan38:
Thanks

nicolasstucki:
There is also #5182, where the macro implementation changed completely.

lJoublanc:
How will this interact with the standard collections `zip` operator, which is left-associative (rather than right-associative, like the implementation suggested here)?
It would be nice to be able to say e.g. `List(a) zip List(b) zip List(c) === List((a,(b,c))) === List((a,b,c))`, but with the existing associativity of `zip`, you end up with `List(((a,b),c)) <> List((a,b,c))`.

AnirudhVyas:
Is this added to Dotty already?

nicolasstucki:
Yes. Under `scala.Tuple`. `(a,b, c)` is equivalent to `a *: b *: c *: ()`

PR #2204, opened 2017-04-08
Fix #2201: Less aggressive type application reduction for better inference

### LINKED ISSUE #2201: Type alias and higher kinded type cause type mismatch
Related #2200.

I found one examples with type aliases and higher kinded types where scalac reports no type errors while dotc reports E007 Type mismatch, see https://github.com/lampepfl/dotty/commit/eae920a32ff951bcf8fab7e92d6a666eff59c0a6#commitcomment-21672052

I bumped into this error while trying to compile https://github.com/johnynek/paiges with Dotty. The commands I ran were
```
$ scalac -version
Scala compiler version 2.12.1 -- Copyright 2002-2016, LAMP/EPFL and Lightbend, Inc.
$ scalac tests/pos/packageobjecttypealias/*.scala
$ ./bin/dotc tests/pos/packageobjecttypealias/*.scala

### PR DESCRIPTION
**Pull request by smarter**
Previously we believed that reducing type applications did not affect
type inference as long as the reduced type constructor had the same
arity as the unreduced one, for example reducing `Foo[X, Y]` is fine
when `Foo` is defined as:
```scala
type Foo[A, B] = Bar[A, B]
```
but not when it's defined as:
```scala
type Foo[A] = Bar[A, A]
```
But this is not a sufficient condition: the bounds of the type
constructor arguments also matter for type inference, so we need to be
more strict and disallow reductions in cases like:
```scala
type Foo[A, B] = Bar[B, A]
```
and:
```scala
type Foo[A, B] = Bar[A, Int]
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -391,9 +391,12 @@ class TypeApplications(val self: Type) extends AnyVal {
           if (!args.exists(_.isInstanceOf[TypeBounds])) {
             val followAlias = Config.simplifyApplications && {
               dealiased.resType match {
-                case AppliedType(tyconBody, _) =>
-                  sameLength(dealiased.typeParams, tyconBody.typeParams)
-                    // Reducing is safe for type inference, as kind arity of type constructor does not change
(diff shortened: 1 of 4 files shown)


odersky:
LGTM

PR #2205, opened 2017-04-09
Fix #2220: Change handling of package objects and tweak hk type inference

### LINKED ISSUE #2220: Split benchmarks from regular run tests
Some run tests are benchmarks in disguise. These should be removed.

odersky:


compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1823,15 +1820,34 @@ object SymDenotations {
...
-      if (denots.exists) denots
-      else packageObj.moduleClass.denot match {
-        case pcls: ClassDenotation => pcls.computeNPMembersNamed(name, inherited)
-        case _ => denots
+    /** Looks in both the package object and the package for members. The precise algorithm
+     *  is as follows:
+     *
+     *  If this is the scala package or the package object exists but is currently completing,
+     *  look in the package first, and if nothing is found there, look in the package object second.
+     *  Otherwise, look in the package object first, and if nothing is found there, in
```
smarter: The spec says "The package object should not define a member with the same name as one of the top-level objects or classes defined in package p. If there is a name conflict, the behavior of the program is currently undefined.", we probably want to update that at some points with the rules here.

smarter: (http://www.scala-lang.org/files/archive/spec/2.13/09-top-level-definitions.html#package-objects)

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -726,10 +726,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
 
     tycon2 match {
       case param2: TypeParamRef =>
-        isMatchingApply(tp1) || {
-          if (canConstrain(param2)) canInstantiate(param2)
-          else compareLower(bounds(param2), tyconIsTypeRef = false)
-        }
+        isMatchingApply(tp1) ||
+        canConstrain(param2) && canInstantiate(param2) ||
```
smarter: I have an alternative fix for #2201  at https://github.com/lampepfl/dotty/pull/2204. Your fix looks like it might help in some other situations, but we would need a testcase for that.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1823,15 +1820,33 @@ object SymDenotations {
...
+     *
+     *  If this is the scala package look in the package first, and if nothing is found
+     *  there, look in the package object second. Otherwise, look in the package object
+     *  first, and if nothing is found there, in the package second.
+     *
+     *  The reason for the special treatment of the scala package is that if we
+     *  complete it too early, we freeze its superclass Any, so that no members can
+     *  be entered in it. As a consequence, there should be no entry in the scala package
+     *  object that hides a class or object in the scala package of the same name, because
+     *  the behavior would then be unintuitive for such members.
```
smarter: Would it possible to detect when a scala package object member hides a scala class/object and return an error?

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
odersky: Just merged #2204. I don't have a separate test case for this change, but would argue that the fact that we missed the case before means we cannot exclude that we will miss it in the future. So better play it safe.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
odersky: Not sure. It would be tricky since everything we do in the scala package is fraught with causing all sorts of initialization errors. Not sure it's worth it really, given that we never had a problem so far.

PR #1448, opened 2016-08-10
Fix #1447: Make X$ <:< X.type when X is an object

smarter:
This allows objects to be easily aliased.

Review by @odersky

PR #2206, opened 2017-04-09
Fix #2198: Don't widen module singletons

### LINKED ISSUE #2198: aliasing gets less precise type than Scala
In Scala, `Nil` gets the type `collection.immutable.Nil.type`, but in Dotty it gets `scala.collection.immutable.Nil$`.

```Scala
object Test {
  val Nil = scala.collection.immutable.Nil
}
```

odersky:
Since module classes are a compiler-generated construct that's not directly
visible to programmers, it seems better not to automatically widen a module
singleton to its underlying class.

Fixes #2198.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -1043,9 +1043,11 @@ class Namer { typer: Typer =>
       def isInline = sym.is(FinalOrInline, butNot = Method | Mutable)
 
       // Widen rhs type and approximate `|' but keep ConstantTypes if
-      // definition is inline (i.e. final in Scala2).
+      // definition is inline (i.e. final in Scala2) and keep module singleton types
+      // instead of widening to the underlying module class types.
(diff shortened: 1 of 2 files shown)


liufengyun:
Looks good to me, but what about case val which are not modules defined by enums?

smarter:
LGTM also, you should be able to revert the code change in https://github.com/lampepfl/dotty/pull/1448 but keep the test and it should still work.

odersky:
I'd leave #1448 in. It's still a useful subtyping relationship to have.

odersky:
@liufengyun Good question about enum vals. Here's what, e.g. None translates to:

    final case val None: Option[Nothing] = 
      {
        final class $anon() extends Option[Nothing]() { 
          def isDefined: Boolean = false
          def enumTag: Int = 1
          override def toString: String = "None"
        }
        new Option[Nothing]{...}(): Option[Nothing]
      }

I think there's nothing to do. `None` is an `Option[Nothing]`, and should be widened to that. This is analogous to the change in enum classes where `apply` also returns instances of the enum class itself instead of soemthing more specific.

PR #2209, opened 2017-04-10
Fix #2152: Instantiate dependent result type parameters

### LINKED ISSUE #2152: Orphan PolyParam
The following causes the compiler to crash when trying to pickle some trees.

```scala
class Contra[-D](task: AnyRef)
object Test {
  def narrow(task: AnyRef): Contra[task.type] = new Contra(task)
  def ident[Before](elems: Contra[Before]): Contra[Before] = elems
  val foo = null
  ident(narrow(foo))
}
```

The output from the compiler is [here](https://github.com/lampepfl/dotty/files/881600/output.txt).

odersky:
#2152 shows that dependent result type parameters can end up in
the types of terms, so we have to eliminate them. If we don't we
get orphan parameters in pickling.

Based on #2128. Only one commit is new.


propensive:
I can confirm that this fixes the issue I described (in reduced form) in #2152.

odersky:
@propensive Can you please check whether the latest commit also solves the original issue you were having? Unfortunately my original fix broke something else.

PR #2215, opened 2017-04-10
Fix #2142: Skolemize arguments of dependent methods if necessary

### LINKED ISSUE #2142: Passing new instance when singleton type is expected should result in type error
```scala
object Foo {

class A
val a1 = new A()
val a2 = new A()

def f(x: A, y: x.type) = ()
f(a1, a1)           // ok
f(a1, a2)           // error (as expected)
f(new A(), new A()) // error (*** but passes unexpectedly)
f(new A(), a1)      // error (*** but passes unexpectedly)

def g(x: A)(y: x.type) = ()
g(a1)(a1)           // ok
g(a1)(a2)           // error (as expected)
g(new A())(new A()) // error (*** but passes unexpectedly)
g(new A())(a1)      // error (*** but passes unexpectedly)

}
```

Note that while this also affects the version featuring dependencies within a single parameter group (`f()` above), the problem seems to be independent of the recent PR #2079: The problem with `g()` existed even in commits before #2079 was merged.

odersky:
This PR detects errors around unstable argument types to dependent methods that were
not reported before.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -107,10 +107,11 @@ object Types {
     final def isValueTypeOrLambda: Boolean = isValueType || this.isInstanceOf[TypeLambda]
 
     /** Does this type denote a stable reference (i.e. singleton type)? */
-    @tailrec final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
-      case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable
+    final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
+      case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable || tp.info.isStable
       case _: SingletonType | NoPrefix => true
       case tp: RefinedOrRecType => tp.parent.isStable
+      case tp: ExprType => tp.resultType.isStable
       case _ => false
     }
 
--- a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
@@ -318,7 +318,7 @@ trait TypeAssigner {
   /** Substitute argument type `argType` for parameter `pref` in type `tp`,
    *  skolemizing the argument type if it is not stable and `pref` occurs in `tp`.
    */
-  def substArgForParam(tp: Type, argType: Type, pref: ParamRef)(implicit ctx: Context) = {
+  def safeSubstParam(tp: Type, pref: ParamRef, argType: Type)(implicit ctx: Context) = {
     val tp1 = tp.substParam(pref, argType)
     if ((tp1 eq tp) || argType.isStable) tp1
     else tp.substParam(pref, SkolemType(argType.widen))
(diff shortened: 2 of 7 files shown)


compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -315,10 +315,28 @@ trait TypeAssigner {
     }
   }
 
+  /** Substitute argument type `argType` for parameter `pref` in type `tp`,
+   *  skolemizing the argument type if it is not stable and `pref` occurs in `tp`.
+   */
+  def substArgForParam(tp: Type, argType: Type, pref: ParamRef)(implicit ctx: Context) = {
```
smarter: Minor suggestion: all the `subst*` methods have the `from` argument before the `to` argument, here it's the reverse order which can be a bit confusing. If this is changed it probably makes sense to rename the method to `substParamWithArg` too.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -315,10 +315,28 @@ trait TypeAssigner {
...
+    val tp1 = tp.substParam(pref, argType)
+    if ((tp1 eq tp) || argType.isStable) tp1
+    else tp.substParam(pref, SkolemType(argType.widen))
```
smarter: I would have expected `widenSingleton` here but I don't know if that makes sense or not.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -107,10 +107,11 @@ object Types {
     final def isValueTypeOrLambda: Boolean = isValueType || this.isInstanceOf[TypeLambda]
 
     /** Does this type denote a stable reference (i.e. singleton type)? */
-    @tailrec final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
-      case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable
+    final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
+      case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable || tp.info.isStable
```
smarter: Was this change also needed to make a test pass? It seems more suspicious.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -315,10 +315,28 @@ trait TypeAssigner {
...
+  }
+
   def assignType(tree: untpd.Apply, fn: Tree, args: List[Tree])(implicit ctx: Context) = {
     val ownType = fn.tpe.widen match {
       case fntpe: MethodType =>
-        if (sameLength(fntpe.paramInfos, args) || ctx.phase.prev.relaxedTyping) fntpe.instantiate(args.tpes)
```
smarter: Would it make sense to change the definition of `instantiate` to do the skolemization?

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: Yes, pos/z1720 started failing without the fix.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
odersky: No, I think that's too low-level. By analogy, asSeenFrom does not do skolemization, but reports back that skolemization is needed, and then Typer does it.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
odersky: For SkolemTypes, it's widen also elsewhere. We don't want to get ExprTypes in there.

PR #2216, opened 2017-04-10
fix #2163: don't narrow liftedOwner if symbol is InSuperCall

### LINKED ISSUE #2163: Type uninitializedThis is not assignable
In [b1801a4f](https://github.com/dotty-staging/dotty/blob/b1801a4fd320a87ff039669ed098aa4cde1c7a33/compiler/test/dotty/tools/dotc/reporting/TestReporter.scala#L108-L126). The backend explodes with:

```
[error] Test dotty.tools.dotc.transform.PatmatExhaustivityTest.patmatExhaustivity failed: java.lang.VerifyError: Bad type on operand stack
[error] Exception Details:
[error]   Location:
[error]     dotty/tools/dotc/reporting/TestReporter$$anon$18.<init>(Ljava/io/PrintWriter;)V @7: invokedynamic
[error]   Reason:
[error]     Type uninitializedThis (current frame, stack[1]) is not assignable to 'dotty/tools/dotc/reporting/TestReporter$$anon$18'
[error]   Current Frame:
[error]     bci: @7
[error]     flags: { flagThisUninit }
[error]     locals: { uninitializedThis, 'java/io/PrintWriter' }
[error]     stack: { uninitializedThis, uninitializedThis }
[error]   Bytecode:
[error]     0x0000000: 2a2b b500 122a 2aba 0025 0000 b200 29ba
[error]     0x0000010: 0035 0000 04b7 0038 b1                 
[error] , took 0.01 sec
[error]     at dotty.tools.dotc.reporting.TestReporter$.simplifiedReporter(TestReporter.scala:126)
[error]     at dotty.tools.dotc.transform.PatmatExhaustivityTest.compileFile(PatmatExhaustivityTest.scala:20)
[error]     at dotty.tools.dotc.transform.PatmatExhaustivityTest.$anonfun$176(PatmatExhaustivityTest.scala:74)
[error]     at scala.collection.immutable.List.map(List.scala:273)
[error]     at dotty.tools.dotc.transform.PatmatExhaustivityTest.patmatExhaustivity(PatmatExhaustivityTest.scala:75)
[error]     ...
```

liufengyun:
Fix #2163: don't narrow liftedOwner if symbol is InSuperCall

Previous, the algorithm narrows the lifted owner of anonymous functions in constructor calls to the class itself, lifted as a private method. However, if the anonymous function is called in super during initialization, JVM throws a `uninitializedThis` exception.

The fix additionally checks that narrowing only happens for symbols defined outside of super calls. As a result, the anonymous functions in constructor calls will be lifted as static methods, thus avoids the problem.


odersky:
@liufengyun You correctly diagnosed the problem, but the fix is too radical. See the augmented test case, which generates illegal byte code with your fix. The correct place to put anonymous function is inside the `Outer` class, whereas with the initial patch it becomes a static function of the companion object. With the refined fix we should be good, I hope.

PR #2217, opened 2017-04-10
Upgrade to sbt 0.13.15

smarter:


felixmulder:
We should also update dotty's docker image to supply this sbt version by default.

PR #2222, opened 2017-04-11
Fix #2219: Fix type applications on WildcardType

### LINKED ISSUE #2219: Erroneous Type Mismatch Error?
```scala
type Fst[t[_[_, _]]] = t[[a, b] => a]
type Snd[t[_[_, _]]] = t[[a, b] => b]
final case class Witness[t[_[_, _]], A <: Fst[t]](value: Snd[t])
```

The compiler prints out:
```scala
-- [E007] Type Mismatch Error: tests/alex/test.scala ---------------------------
70 |  final case class Witness[t[_[_, _]], A <: Fst[t]](value: Snd[t])
   |                                                                 ^
   |              found:    Snd[t](Witness.this.value)
   |              required: Snd[(? <: [_$7 <: [_$8, _$9] => Any] => Any)]
```

smarter:
Previously we just returned the unapplied WildcardType which is
incorrect if the WildcardType is bounded. The proper thing to do is to
do the type application on the bounds of the WildcardType and wrap the
result in a WildcardType.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -429,7 +429,7 @@ class TypeApplications(val self: Type) extends AnyVal {
       case dealiased: LazyRef =>
         LazyRef(() => dealiased.ref.appliedTo(args))
       case dealiased: WildcardType =>
-        dealiased
+        WildcardType(dealiased.optBounds.appliedTo(args).bounds)
       case dealiased: TypeRef if dealiased.symbol == defn.NothingClass =>
         dealiased
       case _ if typParams.isEmpty || typParams.head.isInstanceOf[LambdaParam] =>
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -429,7 +429,7 @@ class TypeApplications(val self: Type) extends AnyVal {
       case dealiased: LazyRef =>
         LazyRef(() => dealiased.ref.appliedTo(args))
       case dealiased: WildcardType =>
-        dealiased
+        WildcardType(dealiased.optBounds.appliedTo(args).asInstanceOf[TypeBounds])
```
odersky: You can just write `.bounds` instead of `.asInstanceOf[TypeBounds])`.

smarter: Thanks! Updated.

smarter:
One of the stupid benchmark-as-test failed, merging anyway.

PR #2227, opened 2017-04-11
move t3346e from pending to run

liufengyun:
move t3346e from pending to run.


tests/run/t3346e.flags
```diff
@@ -0,0 +1 @@
+-language:Scala2
```
smarter: This won't work, flags files are ignored. For pos tests we have a special `pos-scala2` folder so we would need a `run-scala2` folder here and correct configuration in the tests, it seems easier to change the test to not use Scala2-only features. Why did you enable this test in particular?

liufengyun: I was compiling a community build library, an error message led me to the test. It actually works with current Dotty, it's good to move things from pending.

liufengyun:
Close this for now, not a priority, we'll figure out a strategy how to deal with Scala2 run tests and whether it's worth running them later.

PR #2228, opened 2017-04-11
Fix #2218: Add fullyDefinedType for class parent types

### LINKED ISSUE #2218: Problematic type inference in supercall when type parameter appears only in contravariant position
Following code compiles fine in Scalac, but fails to compile in Dotty, because the type parameter inferred is `new DefaultRule[Nothing, Out, A, X](f) with Name`.

```Scala
trait Rule[In, Out, A, X]
trait Result[Out, A, X]
trait Name

class C {
  def ruleWithName[In, Out, A, X](_name: String, f: In => Result[Out, A, X]): Rule[In, Out, A, X] with Name =
    new DefaultRule(f) with Name

  class DefaultRule[In, Out, A, X](f: In => Result[Out, A, X]) extends Rule[In, Out, A, X]
}
```

If we remove the intersection type, the following version compiles fine in Dotty:
```Scala
trait Rule[In, Out, A, X]
trait Result[Out, A, X]

class C {
  def ruleWithName[In, Out, A, X](_name: String, f: In => Result[Out, A, X]): Rule[In, Out, A, X] =
    new DefaultRule(f)

  class DefaultRule[In, Out, A, X](f: In => Result[Out, A, X]) extends Rule[In, Out, A, X]
}
```

Note that adding variance doesn't help, following version also fails:
```Scala
trait Rule[-In, +Out, +A, +X]
trait Result[+Out, +A, +X]
trait Name

class C {
  def ruleWithName[In, Out, A, X](_name: String, f: In => Result[Out, A, X]): Rule[In, Out, A, X] with Name =
    new DefaultRule(f) with Name

  class DefaultRule[In, Out, A, X](f: In => Result[Out, A, X]) extends Rule[In, Out, A, X]
}
```

odersky:
If we do not do that, any type variables in the parent type get interpolated
later, when the whole primary constructor is typed. But by then we miss the
context of what their variance was.

Fixes #2218

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -843,7 +843,7 @@ class Namer { typer: Typer =>
           val targs1 = targs map (typedAheadType(_))
           val ptype = typedAheadType(tpt).tpe appliedTo targs1.tpes
           if (ptype.typeParams.isEmpty) ptype
-          else typedAheadExpr(parent).tpe
+          else fullyDefinedType(typedAheadExpr(parent).tpe, "class parent", parent.pos)
         }
 
       /* Check parent type tree `parent` for the following well-formedness conditions:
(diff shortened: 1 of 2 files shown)


liufengyun:
LGTM

PR #2235, opened 2017-04-12
Fix #2146: Make nested implicit function types work

odersky:
Fix problems related to implicit function types.

1. Propagate expected type when typing an implicit closure.
   This is necessary to recursively apply eta expansion in the
   case of nested implicit functions.

2. Recursively transform definitions and accesses in shortcurImplicits

3. Generalize normalize to handle implicit function types

Review by @b-studios?


odersky:
@felixmulder It fails because it does not find dotty.tools.dotc.ParallelTesting.

OlivierBlanvillain:
felix says to rebase

b-studios:
@odersky Thanks alot for addressing the bug! Judging from the output of `-Xprint:shortcutImplicits` this looks really good to me. Finally we can use type aliases like:

```scala
type using[A, B] = implicit B => A
type and[A, B] = implicit B => A
def foo: Int using A and B = { ... }
```

Maybe add some tests on partially eta-expanded implicits?

### Variant 1 - Outside-In (✔)
```scala
def foo: implicit A => implicit B => Int = { implicit a: A => ... }
```
### Variant 2 - Inside-Out (✘)
```scala
def foo: implicit A => implicit B => Int = { implicit b: B => ... }
```

As I still know too little of the dotty compiler, I can't say much about how the code changes affect other existing functionality. You probably want to ask a second reviewer for that.

Blaisorblade:
Thanks for https://github.com/dotty-staging/dotty/pull/2.

> As I still know too little of the dotty compiler, I can't say much about how the code changes affect other existing functionality. You probably want to ask a second reviewer for that.

👍. But FWIW, more recursion in `normalize` after taking a step is generally sensible.

PR #2237, opened 2017-04-12
Update docs & scrips with updated links to backend.

DarkDimius:


DarkDimius:
@felixmulder, could you also please update the CI?

OlivierBlanvillain:
Yesterday you mentioned merging the `dotty-library` & `sharing-backend` branches, what do you think about merging these two now and make this the default branch to remove the need for -b in clone?

DarkDimius:
@OlivierBlanvillain, by doing this we're convoluting two questions:
 - when\if we upstream changes to backend? Given latest discussion the answer seems "never".
 - when\if we upstream modifications to stdlib? I don't see why we shouldn't upstream them before 2.13. 

I think, It will be easier to upstream changes if we don't merge those two branches. No strong opinion on this though.

felixmulder:
Needs to be rebased - scripts directory no longer exists :)

PR #2240, opened 2017-04-12
Changes necessary to make the collection strawman compile

odersky:
Several bug fix, and one language change:

We now instantiate type variables of implicits to the or-dominator of their lower bound.
This is necessary to compile code like this:

    def useOrd[T: math.Ordering](xs: T*) = ()
    useOrd(Some(1), None)

Code like this occurs in the collection strawman. It should be pretty common also elsewhere.


odersky:
Test fails in vulpix with

java.lang.NullPointerException: null, took 0.0 sec
578s
2888
[error]     at dotty.tools.vulpix.SummaryReport$$anonfun$echoSummary$5.apply(SummaryReport.scala:122)
578s
2889
[error]     at dotty.tools.vulpix.SummaryReport$$anonfun$echoSummary$5.apply(SummaryReport.scala:122)
578s
2890
[error]     at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
578s
2891
[error]     at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
578s
2892
[error]     at dotty.tools.vulpix.SummaryReport.echoSummary(SummaryReport.scala:122)
578s
2893
[error]     at dotty.tools.dotc.CompilationTests$.cleanup(CompilationTests.scala:257)
578s
2894
[error]     at dotty.tools.dotc.CompilationTests.cleanup(CompilationTests.scala)

smarter:
@odersky Try rebasing, I've made a PR recently to fix an NPE in vulpix.

smarter:
I think the real issue here is that `Ordering` is not contravariant in its type parameter. The collection strawman might be a good time to experiment with changing this, although this will have different (and hopefully better) semantics in dotty for implicit search. What do you think?

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -81,14 +84,25 @@ object Inferencing {
...
+          if (force.implicitMode) {
+            // instantiate to or-dominator of lower bound; without this tweak we'd
+            // fail to find an implicit in situations like this:
+            //
+            //   def f[T: Ordering](xs: T*) = ...
+            //   f(Some(1), None)
+            //
+            // The problem is that there is no implicit Ordering instance for the otherwise inferred
+            // lower bound of T, which is `Some[Int] | None`.
+            ctx.orDominator(ctx.typeComparer.bounds(tvar.origin).lo) <:< tvar.origin
```
smarter: Using a subtype check here to change how instantiation proceeds feels a bit too indirect. How about instead we add an `implicitMode: Boolean` parameter (or it could be called `widenUnion` maybe) to `TypeVar#instantiate` instead?

odersky: I don't want to poison more code with the `implicitMode` distinction. It's better to see this not as a check, but as establishing a constraint. Then it makes sense where it is, IMO.

smarter:
One really ugly thing we could would be to hardcode `scala.math.Ordering` as contravariant in its type parameter, I'm sure could would only work better as a result :).

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -45,9 +45,12 @@ object Inferencing {
     else throw new Error(i"internal error: type of $what $tp is not fully defined, pos = $pos") // !!! DEBUG
 
 
-  /** Instantiate selected type variables `tvars` in type `tp` */
+  /** Instantiate selected type variables `tvars` in type `tp`. Instantiation
+   *  proceeds with `implicitMode = true`, that is, type variables are approximated
+   *  from below using or-dominators.
+   */
   def instantiateSelected(tp: Type, tvars: List[Type])(implicit ctx: Context): Unit =
```
smarter: This method is now specific to implicit search but its name doesn't indicate that. I would either rename the method or move it next to its sole usage in Typer#adaptNoArgs

odersky:
I agree, the issue is that Ordering is not contravariant. Unfortunately,contravariant type classes were so far not that practical because implicit search would then prefer a more general instance over a more specific one. I.e. `Ordering[Any]` would have been chosen over `Ordering[Option[T]]`. We changed that in dotty, but it will be a while until common libraries adapt.

smarter:
Yes, I was suggesting we pretend `Ordering` is contravariant in dotty even though it isn't, if only so that we can get more experience with contravariant type classes, but as I said it would be a very ugly hack :).

julienrf:
> The collection strawman might be a good time to experiment with changing this

:+1:

odersky:
Looks like another NPE in vulpix

felixmulder:
Debugging this in #2220, I think it's to do with the cleanup logic.

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -24,7 +24,8 @@ class ScalaSettings extends Settings.SettingGroup {
   val deprecation = BooleanSetting("-deprecation", "Emit warning and location for usages of deprecated APIs.")
   val migration = BooleanSetting("-migration", "Emit warning and location for migration issues from Scala 2.")
   val encoding = StringSetting("-encoding", "encoding", "Specify character encoding used by source files.", Properties.sourceEncoding)
-  val explaintypes = BooleanSetting("-explaintypes", "Explain type errors in more detail.")
+  val explainTypes = BooleanSetting("-explain:types", "Explain type errors in more detail.")
```
smarter: If we use `:` for `-explain:types` shouldn't we have `-explain:implicits` below too?

smarter: Ah nevermind, I see it's changed in the next commit

odersky:
@felixmulder We get:

    /bin/sh: ./scripts/update-scala-library: not found

smarter:
@odersky Try rebasing

felixmulder:
Rebase! It is gone :)

odersky:
Current status if forwardParamBindings is OFF:

 - compileMixed does with a CyclicReference error
 - tools.io and compile-stdlib fail with deep subtype recursions
 - run/i2818.scala also dies with a deep subtype recursion. For this one I checked that it compiles
    OK with -Yno-deep-subtypes not set.

All other tests succeed.

odersky:
I'd like to get this in soon, because I fear that if we wait, collections will get even more code that will fail in dotty. So we need to get to a point to do CI of the strawman with dotty and scalac. For this we need:

 - this PR to go in.
 - #2314 to be fixed and merged.
 - The bridge issue discovered by @julienrf when running the strawman to be fixed.

/cc @julienrf @OlivierBlanvillain @DarkDimius

odersky:
Subsumed by #2331.

PR #2247, opened 2017-04-13
Update intructions to run a single test

OlivierBlanvillain:
I removed the `|` part, it's not possible anymore right?

### DIFF
--- a/docs/docs/contributing/testing.md
+++ b/docs/docs/contributing/testing.md
@@ -80,10 +80,9 @@ This might be aliased in the future. It is also possible to run tests filtered
 by using:
 
 ```bash
-> filterTest .*i2147.scala
+> vulpix i2147.scala
 ```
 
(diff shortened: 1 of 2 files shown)


docs/docs/contributing/testing.md
```diff
@@ -80,10 +80,9 @@ This might be aliased in the future. It is also possible to run tests filtered
...
-> filterTest .*i2147.scala
+> vulpix i2147.scala
 ```
 
 This will run both the test `./tests/pos/i2147.scala` and
-`./tests/partest-test/i2147.scala` since both of these match the given regular
-expression. This also means that you could run `filterTest .*` to run all
-integration tests.
+`./tests/partest-test/i2147.scala` since both of these match the given string.
+This also means that you could run `vulpix ""` to run all integration tests.
```
felixmulder: Just run `vulpix`, no need for the `""`

PR #2249, opened 2017-04-13
Revert <: Product requierment in pattern matching

OlivierBlanvillain:
The change in question broke the following pattern, commonly used in name based pattern matching:

```scala
object ProdEmpty {
  def _1: Int = ???
  def _2: String = ???
  def isEmpty = true
  def get = this
}
```

This type define both `_1` and `get` + `isEmpty` (but is not <: Product). After #1938, `ProdEmpty` became eligibles for both product and name based pattern. Because "in case of ambiguities, *Product Pattern* is preferred over *Name Based Pattern*", isEmpty wouldn't be used, breaking the scalac semantics.

Since this issue is blocker for cross compiling the back-end, I propose to first merge the revert, and discuss  possible alternatives later.


odersky:
LGTM

PR #2251, opened 2017-04-13
Fix #2220: disable benchmarks, set run timeout to 30 seconds

### LINKED ISSUE #2220: Split benchmarks from regular run tests

felixmulder:
Not sure if this is too aggressive. I'd like some input.


smarter:
Probably too aggressive if we want the tests to succeed under high load. I would put 30 seconds to have a good margin.

felixmulder:
I actually ran this successfully multiple times without failure on my machine with 3 seconds timeout.

Do you think we still need 10x that for the CI?

smarter:
Try running something which uses 100% of all your cores on your machine at the same time.

felixmulder:
Alright 30 it is ^^

smarter:
Some of the disabled tests are not benchmarks but stress tests for the collection, so the directory name is not correct for them, but I agree that they should be disabled. However I don't see why Course-2002-04.scala should be disabled, it looks like a legitimate test to me.

felixmulder:
@smarter - done, renamed the dir to `long-running` instead. That's perhaps more apt.

smarter:
I still think Course-2002-04.scala should not be disabled. How long does it take? Why does it take that long?

felixmulder:
@smarter - my mistake. Didn't think that I'd already moved it to disabled.

felixmulder:
@smarter - does it have your blessing now?

PR #2258, opened 2017-04-13
Fix #2246: Ensure ScalaTest compiles again

### LINKED ISSUE #2246: Semantic names not used in LambdaLift
The semantic name PR broke compilation of scalatest, see https://github.com/lampepfl/dotty/issues/2233, the stacktrace is at https://gist.github.com/smarter/2b03b8fc57f7b8e4a027affe39b4426b and fails in LambdaLift. I'm not sure why, but I've noticed that LambdaLift defines the following method:
```scala
    private def newName(sym: Symbol)(implicit ctx: Context): Name =
      if (sym.isAnonymousFunction && sym.owner.is(Method, butNot = Label))
        (sym.name ++ NJ ++ sym.owner.name).freshened
      else sym.name.freshened
```
Which seems like it should be creating a semantic qualified name instead.

odersky:
Several problems. The first had to do with name mangling. The second concerned handling of symbols in super calls, augmenting #2216. Finally, the third problem concerned interaction between
symbols in super calls and outer accessor generation.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
+++ b/compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
@@ -11,18 +11,17 @@ import core.Decorators._
 import core.StdNames.nme
 import core.Names._
 import core.NameOps._
+import core.NameKinds.ExpandPrefixName
 import core.Phases._
 import ast.Trees._
 import SymUtils._
 import ExplicitOuter.outer
 import util.Attachment
-import util.NameTransformer
 import util.Positions._
 import collection.{ mutable, immutable }
 import collection.mutable.{ HashMap, HashSet, LinkedHashMap, LinkedHashSet, TreeSet }
 
 object LambdaLift {
-  private val NJ = NameTransformer.NAME_JOIN_STRING
   private class NoPath extends Exception
 }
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -338,7 +337,9 @@ class LambdaLift extends MiniPhase with IdentityDenotTransformer { thisTransform
 
     private def newName(sym: Symbol)(implicit ctx: Context): Name =
       if (sym.isAnonymousFunction && sym.owner.is(Method, butNot = Label))
-        (sym.name ++ NJ ++ sym.owner.name).freshened
+        sym.name.rewrite {
+          case name: SimpleTermName => ExpandPrefixName(sym.owner.name.asTermName, name)
```
smarter: It's kind of weird to use `ExpandPrefixName` here since this isn't a prefix, isn't it?

odersky: Yes, but FlatName does not fit either.

odersky:
Superseded by #2263.

PR #2260, opened 2017-04-13
Fix #2184: Hyper Bootstrap! Integrate the backend as a git submodule

### LINKED ISSUE #2184: Hyper bootstrap
Depends on #2183

smarter:
The backend lives in the scalac fork at
https://github.com/lampepfl/scala/tree/sharing-backend. Before this
commit, the scala-compiler built from this fork was a dependency of
Scala just to get this backend. This made it much more cumbersome to
test changes to the backend and also forced us to depend on Scala 2.11.5
since that is the version of scalac that the forked backend is based on.

This commit changes this by adding a git submodule in the
`scala-backend` directory that points to the scalac fork. We do not
compile the whole submodule, instead we add the subset of files we need
to the dotty-compiler project in the sbt build. See backend.md for more
information (online at
http://dotty.epfl.ch/docs/contributing/backend.html) once this commit is
merged.

The most important thing to note is that whenever you clone dotty you
should do `git clone --recursive` to also clone the submodule. If you
already have a cloned dotty you'll need to do:
```shell
git submodule update --init
```

You will need to repeat this command anytime you update dotty and the backend has been updated, otherwise the content of `scala-backend` will be stale.


docs/docs/contributing/backend.md
```diff
@@ -0,0 +1,72 @@
...
+instead we add the subset of files we need to the dotty-compiler project in the
+sbt build.
+
+The most important thing to know when working with git submodules is that
+their content is not automatically updated when you do `git checkout` or `git
+pull`, instead everytime you switch branch, you need to do:
+
+``` shell
+git submodule update
+```
```
felixmulder: We could make sbt do this so that we never miss it - also helps beginners.

project/Build.scala
```diff
@@ -631,10 +648,7 @@ object DottyInjectedPlugin extends AutoPlugin {
 
       baseDirectory in (Test,run) := (baseDirectory in `dotty-compiler`).value,
 
-      libraryDependencies ++= Seq(
-        scalaCompiler % Test,
-        "com.storm-enroute" %% "scalameter" % "0.6" % Test
-      ),
+      libraryDependencies += "com.storm-enroute" %% "scalameter" % "0.6" % Test,
```
felixmulder: What is this needed for?

compiler/src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -1187,7 +1187,7 @@ object TreeTransforms {
         case NonFatal(ex) =>
           if (tree ne crashingTree) {
             crashingTree = tree
-            println(i"exception while transforming $tree of class ${tree.getClass} # ${tree.uniqueId}")
+            // println(i"exception while transforming $tree of class ${tree.getClass} # ${tree.uniqueId}")
```
felixmulder: ?

project/Build.scala
smarter: It was already here before, no idea.

compiler/src/dotty/tools/dotc/transform/TreeTransform.scala
smarter: Oops, missed that.

docs/docs/contributing/backend.md
smarter: I'm not a fan of doing more weird stuff with sbt, let's not do that for now.

project/Build.scala
felixmulder: If we can we should kill it with :fire:

smarter: Sure, but removing/updating the `bench` project is a separate issue.

smarter:
@odersky @nicolasstucki @liufengyun @OlivierBlanvillain The backend is in! Don''t forget to do:
```scala
git submodule update --init
```
After pulling these changes. See https://github.com/lampepfl/dotty/blob/master/docs/docs/contributing/backend.md (soon online at http://dotty.epfl.ch/docs/contributing/backend.html) for more information.

PR #2262, opened 2017-04-14
Make warm compilation speed with sbt 2x faster

smarter:
Previously, every call to `compile` in sbt with dotty took about the
same time because we created a new ClassLoader everytime and thus
thrased the JIT code cache, by reusing ClassLoaders we can make
`compile` about 2x faster.

You can reproduce this by running:
```shell
> dotty-compiler-bootstrapped/compile
```
This takes ~50 seconds on my machine. Then clean using:
```shell
> ;dotty-compiler-bootstrapped/clean;dotty-compiler-update
```
And run `dotty-compiler-bootstrapped/compile` again, this takes ~25
seconds for me. I get very similar timings from scalac (replacing
`dotty-compiler-bootstrapped` by `dotty-compiler`).

### DIFF
--- a/sbt-bridge/src/xsbt/CompilerClassLoader.scala
+++ b/sbt-bridge/src/xsbt/CompilerClassLoader.scala
@@ -2,6 +2,8 @@ package xsbt
 
 import java.net.{URL, URLClassLoader}
 
+import scala.collection.mutable
+
 /**  A classloader to run the compiler
  *
  *   A CompilerClassLoader is constructed from a list of `urls` that need to be on
@@ -42,6 +44,14 @@ class CompilerClassLoader(urls: Array[URL], sbtLoader: ClassLoader)
 }
 
 object CompilerClassLoader {
+  /** Cache the result of `fixBridgeLoader`.
+   *
+   *  Reusing ClassLoaders is important for warm performance since otherwise the
+   *  JIT code cache for the compiler will be discarded between every call to
+   *  the sbt `compile` task.
+   */
+  private[this] val fixedLoaderCache = new mutable.WeakHashMap[ClassLoader, ClassLoader]
+
   /** Fix the compiler bridge ClassLoader
    *
    *  Soundtrack: https://www.youtube.com/watch?v=imamcajBEJs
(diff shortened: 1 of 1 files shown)


sbt-bridge/src/xsbt/CompilerClassLoader.scala
```diff
@@ -42,6 +44,14 @@ class CompilerClassLoader(urls: Array[URL], sbtLoader: ClassLoader)
 }
 
 object CompilerClassLoader {
+  /** Cache the result of `fixBridgeLoader`.
+   *
+   *  Reusing ClassLoaders is important for warm performance since otherwise the
+   *  JIT code cache for the compiler will be discarded between every call to
+   *  the sbt `compile` task.
+   */
+  private[this] val fixedLoaderCache = new mutable.WeakHashMap[ClassLoader, ClassLoader]
```
DarkDimius: in weak hashMap only values are weak. You are keeping SBT classloaders(keys) from being GC-d.
I'd prefer to have one more level of indirection to also remove this issue.

sbt-bridge/src/xsbt/CompilerClassLoader.scala
```diff
@@ -70,7 +80,10 @@ object CompilerClassLoader {
    *  @param bridgeLoader  The classloader that sbt uses to load the compiler bridge
    *  @return A fixed classloader that works with dotty
    */
-  def fixBridgeLoader(bridgeLoader: ClassLoader) = bridgeLoader match {
+  def fixBridgeLoader(bridgeLoader: ClassLoader): ClassLoader =
+    fixedLoaderCache.getOrElseUpdate(bridgeLoader, computeFixedLoader(bridgeLoader))
```
DarkDimius: should this method be synchronized?

DarkDimius: BTW: CompilerClassLoader.loadClass should definitely be synchronized.

sbt-bridge/src/xsbt/CompilerClassLoader.scala
smarter: Are you sure? From https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html: "Hash table based implementation of the Map interface, with weak keys. An entry in a WeakHashMap will automatically be removed when its key is no longer in ordinary use."

sbt-bridge/src/xsbt/CompilerClassLoader.scala
smarter: You're right, this method should be synchronized. I don't think `loadClass` needs to be however because I'm only calling `super.loadClass` and `resolveClass` which should both be thread-safe. In fact, I think that I should be able to call https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#registerAsParallelCapable-- to make my ClassLoader parallel.

sbt-bridge/src/xsbt/CompilerClassLoader.scala
DarkDimius: You are right.

PR #2263, opened 2017-04-14
Fix #2246: Use semantic names when freshening

### LINKED ISSUE #2246: Semantic names not used in LambdaLift

odersky:
I have decided to treat the super call problems elsewhere. Both are needed to make ScalaTest compile again.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
+++ b/compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
@@ -11,18 +11,17 @@ import core.Decorators._
 import core.StdNames.nme
 import core.Names._
 import core.NameOps._
+import core.NameKinds.ExpandPrefixName
 import core.Phases._
 import ast.Trees._
 import SymUtils._
 import ExplicitOuter.outer
 import util.Attachment
-import util.NameTransformer
 import util.Positions._
 import collection.{ mutable, immutable }
 import collection.mutable.{ HashMap, HashSet, LinkedHashMap, LinkedHashSet, TreeSet }
 
 object LambdaLift {
-  private val NJ = NameTransformer.NAME_JOIN_STRING
   private class NoPath extends Exception
 }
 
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/HoistSuperArgs.scala
```diff
@@ -0,0 +1,171 @@
...
+      case tp: LambdaType =>
+        tp.derivedLambdaType(resType = replaceResult(tp.resultType, restpe))
+      case _ => restpe
+    }
+
+    /** A method representing a hoisted supercall argument */
+    def newSuperArgMethod(argType: Type) = {
+      val (staticFlag, methOwner) =
+        if (cls.owner.is(Package)) (JavaStatic, cls) else (EmptyFlags, cls.owner)
+      val argTypeWrtConstr = argType.subst(allParams, allParamRefs(constr.info))
```
liufengyun: It seems `allParamRefs(constr.info)` is called for each argument, maybe turn it into lazy val? Or, it's because `constr.info` may change between different calls?

liufengyun:
A mini test that reproduce the run-time exceptions:
```Scala
class Base[A](exp: => Option[A])

object Empty extends Base[Nothing](None)

object Test {
  def main(args: Array[String]): Unit = Empty
}
```

odersky:
The super call arguments handling had "interesting" interactions with elimByName. The new treatment has been moved to a separate PR #2282.

odersky:
@liufengyun I added your example to #2282. Thanks!

PR #2264, opened 2017-04-15
Only ignore inlined if conditions if pure.

nicolasstucki:
Idempotent ones could have a side effect.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inliner.scala
@@ -527,7 +527,7 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
       cond1.tpe.widenTermRefExpr match {
         case ConstantType(Constant(condVal: Boolean)) =>
           val selected = typed(if (condVal) tree.thenp else tree.elsep, pt)
-          if (isIdempotentExpr(cond1)) selected
+          if (isPureExpr(cond1)) selected
           else Block(cond1 :: Nil, selected)
         case _ =>
           val if1 = untpd.cpy.If(tree)(cond = untpd.TypedSplice(cond1))


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -527,7 +527,7 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
       cond1.tpe.widenTermRefExpr match {
         case ConstantType(Constant(condVal: Boolean)) =>
           val selected = typed(if (condVal) tree.thenp else tree.elsep, pt)
-          if (isIdempotentExpr(cond1)) selected
+          if (isPureExpr(cond1)) selected
```
nicolasstucki: In fact it looks like all constant type conditions that reach this code are pure. Are they all supposed to be in this particular case? If not how could I construct a condition that has a constant type and is not pure.

DarkDimius: I think code should be transformed in a different way instead(That's what is done in linker): 
```scala
if ($cond: true.type) { $themp} else {$elsep} 

==>> 
cond$; 
thenp$
```

Similar for false.type .

DarkDimius: > In fact it looks like all constant type conditions that reach this code are pure. Are they all supposed to be in this particular case?

```
lazy val s: true = { println(1); true}

@inline def bla = 
  if (s) {...} else {...}

```

nicolasstucki: That lazy val creates the idempotent condition with constant type. Just found another bug with lazy vals with constant type #2266.

felixmulder:
Deferring for now. Awaiting effects system :)

PR #2267, opened 2017-04-15
Fix #2266: Do not replace constant type lazy vals with constant.

### LINKED ISSUE #2266: Lazy val with constant type loses side effects.
```scala
lazy val x: true = { println("X"); true }
x
```
does not print `X`.

nicolasstucki:


compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -427,7 +427,7 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
   def constToLiteral(tree: Tree)(implicit ctx: Context): Tree = {
     val tree1 = ConstFold(tree)
     tree1.tpe.widenTermRefExpr match {
-      case ConstantType(value) if isIdempotentExpr(tree1) => Literal(value)
+      case ConstantType(value) if isIdempotentExpr(tree1) && !tree1.symbol.is(Lazy) => Literal(value)
```
DarkDimius: I don't think this is specific to lazy vals.
```scala
abstract class A {
  def s: Boolean
  assert(Test.r == this.s)
}

object Test extends A {
   override def s = r
   val r: true = true
   def main(args: Array[String]): Unit = {}
}

Comment (DarkDimius) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: // updated example to make it more likely to trigger

Comment (nicolasstucki) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: What is the issue you would expect with that example?

Comment (DarkDimius) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: assertion will fail, while it shouldn't.

Comment (DarkDimius) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: Sorry, forgot to update the example, though I've left the comment that I did update it.

Comment (DarkDimius) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: Now, updated.

Comment (nicolasstucki) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: Indeed, that one fails for a similar reason.

Comment (nicolasstucki) [inline compiler/src/dotty/tools/dotc/ast/TreeInfo.scala]: I added a fix for that issue.

File: compiler/src/dotty/tools/dotc/ast/TreeInfo.scala [hunk 2]
```diff
@@ -426,8 +426,17 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
    */
   def constToLiteral(tree: Tree)(implicit ctx: Context): Tree = {
     val tree1 = ConstFold(tree)
+    def canInlineConstant(value: Constant): Boolean = {
+      isIdempotentExpr(tree1) && // see note in documentation
+      // lazy value must be initialized (would not be needed with isPureExpr)
+      !tree1.symbol.is(Lazy) &&
+      // could hide initialization order issues (ex. val with constant type read before initialized)
+      (!ctx.owner.isLocalDummy || (tree1.symbol.is(Method) || value.isZero) ||
```
DarkDimius: `tree1.symbol.is(Method)`

Why is inlining the return value of a (possibly-side-effectfull) method safe?

DarkDimius:
LGTM

PR #2270, opened 2017-04-17
Make sure basic enums work with REPL

felixmulder:
These changes make sure that

- enums are correctly highlighted in the REPL
- infinite loop does not occur when parsing: `enum Option[+T] {<EOF>`


OlivierBlanvillain:
LGTM

PR #2272, opened 2017-04-18
Fix #2185: Add bytecode idempotency checks.

### LINKED ISSUE #2185: Bytecode idempotency test for compiler

nicolasstucki:
Add `bytecodeIdemporency` to compilation tests. This tests compiles any compilation test twice and compares the resulting bytecodes (currently `dotty-compiler` and `tests/pos`).


compiler/src/dotty/tools/dotc/core/tasty/TastyPickler.scala
```diff
@@ -14,7 +14,7 @@ import Decorators._
 class TastyPickler {
 
   private val sections = new mutable.ArrayBuffer[(NameRef, TastyBuffer)]
-  val uuid = UUID.randomUUID()
+  private val uuid = UUID.fromString("3cee1b79-c03a-4125-b337-d067b5cb3a94") // TODO: use a hash of the tasty tree
```
DarkDimius: I'd prefer not to merge this change as-is.

nicolasstucki: @DarkDimius, you talked about making the hash for it. Did you have something specific in mind?

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -285,6 +261,115 @@ class CompilationTests extends ParallelTesting {
...
     }.map(_.checkCompile()).foreach(_.delete())
   }
+
+  @Test def bytecodeIdemporency: Unit = {
+    var failed = 0
+    var total = 0
+    val blacklisted = Set(
+      "pos/Map/scala/collection/immutable/Map",
+      "pos/Map/scala/collection/immutable/AbstractMap",
+      "pos/t1203a/NodeSeq"
```
smarter: Why is there a blacklist? If these files fail the test, shoudn't issues be filed for them?

nicolasstucki: There will. I just haven't had time to do it yet.

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -285,6 +261,115 @@ class CompilationTests extends ParallelTesting {
...
+    }
+
+    val opt = defaultOptions.and("-YemitTasty")
+
+    def idempotency1() = {
+      compileList("dotty1", compilerSources ++ backendSources ++ backendJvmSources, opt) +
+      compileFilesInDir("../tests/pos", opt)
+    }
+    def idempotency2() = {
+      compileList("dotty1", compilerSources ++ backendSources ++ backendJvmSources, opt) +
```
smarter: It would be interesting to combine this with a bootstrap test: Compile the compiler, then compile the compiler with the compiled compiler, and check equality.

compiler/test/dotty/tools/dotc/CompilationTests.scala
nicolasstucki: Added them failures to #2274.

compiler/test/dotty/tools/dotc/CompilationTests.scala
nicolasstucki: That will be interesting, but I would add it as a second step in another PR.

compiler/src/dotty/tools/dotc/core/tasty/TastyPickler.scala
DarkDimius: This idea looks good to me.

compiler/src/dotty/tools/dotc/util/MurmurLongHash3.scala
```diff
@@ -0,0 +1,261 @@
...
+  *
+  * This is based on the earlier MurmurHash3 code by Rex Kerr, but the
+  * MurmurHash3 algorithm was since changed by its creator Austin Appleby
+  * to remedy some weaknesses and improve performance. This represents the
+  * latest and supposedly final version of the algortihm (revision 136).
+  *
+  * @see [[http://code.google.com/p/smhasher]]
+  */
+object MurmurLongHash3 extends MurmurLongHash3 {
+  final val arraySeed       = 0x3c074a61
```
DarkDimius: Why are seeds ints?

compiler/src/dotty/tools/dotc/util/MurmurLongHash3.scala
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/**
+  * An implementation of Austin Appleby's MurmurHash 3 algorithm
+  * (MurmurHash3_x86_32). This object contains methods that hash
+  * values of various types as well as means to construct `Hashing`
+  * objects.
+  *
+  * This algorithm is designed to generate well-distributed non-cryptographic
+  * hashes. It is designed to hash data in 32 bit chunks (ints).
```
smarter: 64 now, right ?

compiler/src/dotty/tools/dotc/util/MurmurLongHash3.scala
```diff
@@ -0,0 +1,261 @@
...
+**                          |/                                          **
+\*                                                                      */
+
+/* Copied and adapted from scala.util.hashing.MurmurHash3 to use Longs */
+
+package dotty.tools.dotc.util
+
+import java.lang.Long.{ rotateLeft => rotl }
+
+private[util] class MurmurLongHash3 {
```
smarter: Nice! I think this should be contributed upstream to scala-library.

nicolasstucki: Good idea. But I need to reimplement the algorithm from the source. @DarkDimius showed me some issues with my simple port. I will try to make it compatible `MurmurHash3` to be able to just add it inside in the future.

compiler/src/dotty/tools/dotc/util/MurmurLongHash3.scala
nicolasstucki: Yes

DarkDimius:
As discussed in person, using murmur3 to emmit 64-bit hashes is not evidenced to be a good hash.
Similarly, using 128bit mumur3 and taking part of is isn't good iether(murmu3 is not cryptographic hash).

I propose we don't use murmur family for this one, and instead use FarmHash / CityHash, which have been open-sourced by google and are used to hash entries in maps(our use-case).

smarter:
> As discussed in person, using murmur3 to emmit 64-bit hashes is not evidenced to be a good hash.

Interesting, why?

nicolasstucki:
Updated hashing

compiler/src/dotty/tools/dotc/util/CityHash.scala
```diff
@@ -0,0 +1,186 @@
+package dotty.tools.dotc.util
+
+/* Ported from https://github.com/google/cityhash/blob/master/src/city.cc */
```
smarter: You should preserve the copyright/license header of the original file, and you should update AUTHORS.md to note this.

compiler/src/dotty/tools/dotc/util/CityHash.scala
```diff
@@ -0,0 +1,186 @@
...
+    (a + z, b + c)
+  }
+
+  /** Return a 16-byte hash for s[0] ... s[31], a, and b.  Quick and dirty. */
+  private final def weakHashLen32WithSeeds(s: Int, a: Long, b: Long)(implicit data: Array[Byte]): (Long, Long) =
+    weakHashLen32WithSeeds(fetch64(s), fetch64(s + 8), fetch64(s + 16), fetch64(s + 24),  a, b)
+
+  private final def fetch64(idx: Int)(implicit data: Array[Byte]): Long = {
+    var x: Long = data(idx)
+    x = data(idx + 1) | (x << 8)
```
smarter: Careful with sign extension (because `Byte` is signed). If I recall correctly, you need to add`& 0xff` everywhere you convert from a `Byte` to an `Int` to avoid that.

smarter: You should add some unit tests for this class, to make sure it behaves like the reference implementation.

nicolasstucki:
@DarkDimius the hash function was updated.

DarkDimius:
> >  As discussed in person, using murmur3 to emmit 64-bit hashes is not evidenced to be a good hash.
>
> Interesting, why?

While real cryptographic hashes are expected to remain "good" if you take a subsection of it, you don't have this guarantee for non-cryptograhic hashes in general.
Murmur3 family is not cryptographic and does not include 64-bit hash. Using 128-bit hash and only taking a part of it may be a weak hash and I didn't quickly find an evidence to be proven otherwise. 

That's why I've proposed FarmHash / CityHash that have native 64-bit variants.

@nicolasstucki, I've lost context, we did we decide to use Elf hashing instead of CityHash?

nicolasstucki:
The context for using Elf hashing instead of CityHash was because the implementation is trivial and it is simpler to be sure that it is correctly implemented. But, if we want to use CityHash we can just revert 426a33ab97e853e7b57716a7846d2ea43af76522.

felixmulder:
@DarkDimius - is this good to go now?

PR #2273, opened 2017-04-19
Add missing -Yno-inline to stdlib compilation flags.

nicolasstucki:


felixmulder:
Same fix Martin had to apply to the normal stdlib. Cheers!

PR #2277, opened 2017-04-19
Join tests to increase parallelism.

nicolasstucki:
Also reorded a couple of tests to have the long tests first.


nicolasstucki:
Before this change running `sbt vulpix` on my 4 core machine was taking 210 seconds. After the change it runs in around 150 seconds.

felixmulder:
I've feared previously that tests would clash if joined too aggressively, but I've had a look at the changes and believe there will be no conflicts. Thanks for the change!

PR #2280, opened 2017-04-20
Add seconds to log timestamps to avoid log name clashes.

nicolasstucki:


compiler/test/dotty/tools/dotc/reporting/TestReporter.scala
```diff
@@ -84,10 +84,12 @@ object TestReporter {
...
 
   private[this] def initLog() = if (logWriter eq null) {
-    val df = new SimpleDateFormat("yyyy-MM-dd-HH:mm")
-    val timestamp = df.format(new Date)
-    new JFile("../testlogs").mkdirs()
-    outFile = new JFile(s"../testlogs/tests-$timestamp.log")
+    val date = new Date
+    val df0 = new SimpleDateFormat("yyyy-MM-dd")
+    val df1 = new SimpleDateFormat("yyyy-MM-dd-HH:mm:ss")
+    val folder = s"../testlogs/tests-${df0.format(date)}"
```
DarkDimius: - if tests run in parallel, you can have two tests that start at the same second.
- I would prefer to have names of the log-files that are predictable.

nicolasstucki: This is intended as a quick fix. There are many possible improvements that should be done on these logs.

felixmulder:
I agree with Nicolas, in the end this is an improvement :)

PR #2282, opened 2017-04-20
Rework ElimByName and combine with new phase HoistSuperArgs

odersky:
We split ElimByName so that we can create the necessary closures before a new phase that hoists supercall arguments out of the call. New sequence of phases:

    ByNameClosures
    HoistSuperArgs
    ElimByName


tests/neg/tailcall/t1672b.scala
```diff
@@ -46,7 +46,7 @@ object Test1772B {
     else 1 + (try {
       throw new RuntimeException
     } catch {
-      case _: Throwable => bar(i - 1) // old-error
+      case _: Throwable => bar(i - 1) // error
```
DarkDimius: why was this test affected? It doesn't seem to have anything to do with by-name changes.

DarkDimius:
> Drop InSuperCall flag

This PR is awesome!

odersky:
Current status if Config.forwardParams is turned OFF:

 - CompileMixed dies with a CyclicReference
 - tools_io and compile-stdlib fail with a deep subtype
 - run/t2818.scala also fails with a deep subtype. I verified that that test passes
    if -Yno-deep-subtypes is not set.

Overall, this is quite encouraging. CompileMixed required major wizardry to NOT produce a CyclicReference so I am not that surprised that it comes back if we change the forwarding scheme.

compiler/src/dotty/tools/dotc/transform/TransformByNameApply.scala
```diff
@@ -0,0 +1,61 @@
...
+import core.StdNames.nme
+import ast.Trees._
+
+/** Abstract base class of ByNameClosures and ElimByName, factoring out the
+ *  common functionality to transform arguments of by-name parameters.
+ */
+abstract class TransformByNameApply extends MiniPhaseTransform { thisTransformer: DenotTransformer =>
+  import ast.tpd._
+
+  /** The info of the tree's symbol at phase Nullarify (i.e. before transformation) */
```
DarkDimius: nullarify phase doesn't exist anymore

compiler/src/dotty/tools/dotc/transform/TransformByNameApply.scala
```diff
@@ -0,0 +1,61 @@
...
+  private def originalDenotation(tree: Tree)(implicit ctx: Context) =
+    tree.symbol.denot(ctx.withPhase(thisTransformer))
+
+  /** If denotation had an ExprType before, it now gets a function type */
+  protected def exprBecomesFunction(symd: SymDenotation)(implicit ctx: Context) =
```
DarkDimius: can be private

compiler/src/dotty/tools/dotc/transform/ByNameClosures.scala
```diff
@@ -0,0 +1,37 @@
...
+ *  is a synthetic method defined in Definitions. Erasure will later strip the <cbn-arg> wrappers.
+ */
+class ByNameClosures extends TransformByNameApply with IdentityDenotTransformer { thisTransformer =>
+  import ast.tpd._
+
+  override def phaseName: String = "byNameClosures"
+
+  override def mkClosure(arg: Tree, argType: Type)(implicit ctx: Context): Tree = {
+    val meth = ctx.newSymbol(
+      ctx.owner, nme.ANON_FUN, Synthetic | Method, MethodType(Nil, Nil, argType))
```
DarkDimius: I'd prefer to give it more informative name that indicates that it originated from by-name param.
Helps when reading bytecode.

compiler/src/dotty/tools/dotc/transform/HoistSuperArgs.scala
```diff
@@ -0,0 +1,209 @@
...
+          val superMeth = newSuperArgMethod(arg.tpe)
+          val superArgDef = polyDefDef(superMeth, trefs => vrefss => {
+            val paramSyms = trefs.map(_.typeSymbol) ::: vrefss.flatten.map(_.symbol)
+            val tmap = new TreeTypeMap(
+              typeMap = new TypeMap {
+                def apply(tp: Type) = tp match {
+                  case tp: NamedType
+                  if (tp.symbol.owner == cls || tp.symbol.owner == constr) &&
+                     tp.symbol.is(ParamOrAccessor) =>
+                    val mappedSym = origParams.zip(paramSyms).toMap.apply(tp.symbol)
```
DarkDimius: maybe extract `origParams.zip(paramSyms).toMap` to a lazy val inside this TTM?

compiler/src/dotty/tools/dotc/transform/TransformByNameApply.scala
odersky: No it's used in elimByName

PR #2285, opened 2017-04-21
Fix #2202: Check bridge clash in super classes.

### LINKED ISSUE #2202: Erasure-created bridge clash detection is incomplete
Same bug as [SI-9912](https://github.com/scala/bug/issues/9912), this shouldn't compile:
```scala
class A {
  def compareTo(o: Any): Int = 0
}
class B extends A with Comparable[B] {
  def compareTo(b: B): Int = 0
}
object C {
    println(new B().compareTo(new Object()))
  }
}
```
... but it does and crash at runtime:
```scala
Exception in thread "main" java.lang.ClassCastException: java.lang.Object cannot be cast to B
        at B.compareTo(t9912.scala:5)
        at C$.main(t9912.scala:9)
        at C.main(t9912.scala)
```

nicolasstucki:


tests/neg/i2202.scala
```diff
@@ -0,0 +1,12 @@
+// nopos-error
```
smarter: It should be possible to get a position for this error using Symbol#pos, no?

nicolasstucki: Yes, it is possible. I will add that to this PR.

smarter:
> @nicolasstucki nicolasstucki requested a review from smarter an hour ago

I think @DarkDimius should review this since he understands this code more than I do.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -626,17 +626,19 @@ object Erasure extends TypeTestsCasts{
...
                     if (isRequired) {
                       // check for clashes
-                      val clash: Option[Symbol] = oldSymbol.owner.info.decls.lookupAll(bridge.name).find {
-                        sym =>
-                          (sym.name eq bridge.name) && sym.info.widen =:= bridge.info.widen
-                      }.orElse(
-                        emittedBridges.find(stat => (stat.name == bridge.name) && stat.tpe.widen =:= bridge.info.widen)
+                      val clash: Option[Symbol] = oldSymbol.owner.info.memberExcluding(bridge.name, Flags.Bridge).alternatives.find {
+                        denot => {
+                          !denot.symbol.owner.is(Flags.Trait) && (denot.symbol.owner ne defn.ObjectClass) &&
```
DarkDimius: > !denot.symbol.owner.is(Flags.Trait)

why is clashing with trait methods fine?

> denot.symbol.owner ne defn.ObjectClass

Why?

> !denot.symbol.owner.derivesFrom(bridge.owner) 

Is  `denot.symbol.owner` ever a non-trivial subclass of `bridge.owner` ?
If yes, would be nice to add a test.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -626,17 +626,19 @@ object Erasure extends TypeTestsCasts{
...
+                          !denot.symbol.owner.derivesFrom(bridge.owner) && denot.symbol.info.widen =:= bridge.info.widen
+                        }
+                      }.map(_.symbol).orElse(
+                        emittedBridges.find(stat => (stat.name eq bridge.name) && stat.tpe.widen =:= bridge.info.widen)
```
DarkDimius: I'd prefer to change `eq` back to `==` to keep the diff minimal. This method has a complex history and it's important to keep it clean.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
nicolasstucki: All these conditions where added to make the compilation of the standard library work properly. I will have to explore in detail why I had to exclude traits and object. The non trivial subclass of `bridge.owner` are present in `scala.collection`. Nevertheless I will add a small test that has non-trivial subclass.

felixmulder:
Out of date, needs a second look.

nicolasstucki:
@DarkDimius this issue was fixed with the new bridge implementation.

nicolasstucki:
@DarkDimius why close it? Now I'm only adding the regression test for  #2202.

PR #2306, opened 2017-04-25
Only recreate dotty-compiler jar if it changed

smarter:
Previously `dotty-compiler/packageBin` always recreated the jar because
we overwrote the compiler.properties ressource file.


felixmulder:
Looks fine to me.

PR #2308, opened 2017-04-26
Revert "Fix #2266: Do not replace constant type lazy vals with constant."

### LINKED ISSUE #2266: Lazy val with constant type loses side effects.

felixmulder:
Reverts lampepfl/dotty#2267


odersky:
We should come back to this and try to fix it in a more principled manner. It needs to be an addition to the way we check purity, I think; the present treatment is too local.

PR #2312, opened 2017-04-26
sbt bootstrap Part 2: Revenge of the sbt bootstrap

smarter:
This is a major overhaul of how the sbt bootstrap works, the main goal is to be able to publish a usable bootstrapped compiler, this mean having artefacts with names like `dotty-*_0.1` instead of `dotty-*-bootstrapped_2.11`, this is important to be able to cross-compile with dotty like any other Scala version (I will rewrite sbt-dotty to take advantage of this soon. The other important change in this PR is that the non-bootstrapped compiler no longer needs to be published to compile the bootstrapped compiler, only `dotty-sbt-bridge` needs to be published (I haven't investigated whether this is something we could avoid too). It is still possible to bootstrap from published artefacts using the `bootstrapFromPublishedJars` setting I added, see the commit messages.


smarter:
> Has the infinite loop issue been resolved for the REPL in this case? If not, I would prefer if we fix that first - otherwise people who try out Dotty will be met with a REPL that does not work. Not very good...

I've asked @OlivierBlanvillain to look at it ;).

project/Build.scala
```diff
@@ -159,13 +183,16 @@ object Build {
...
     aggregate(`dotty-library-bootstrapped`, `dotty-compiler-bootstrapped`, `dotty-doc-bootstrapped`,
-      dottySbtBridgeBootstrappedRef).
+      dottySbtBridgeBootstrappedRef,
+      `scala-library`, `scala-compiler`, `scala-reflect`, scalap).
+    dependsOn(`dotty-compiler-bootstrapped`).
+    dependsOn(`dotty-library-bootstrapped`).
     settings(commonBootstrappedSettings).
     settings(
-      publishArtifact := false
+      moduleName := "dotty"
```
sjrd: You can automate this by having the following in `commonBootstrappedSettings`:
```scala
moduleName ~= { _.stripSuffix("-bootstrapped") }
```

sbt-bridge/test/xsbt/TestDriver.scala
```diff
@@ -0,0 +1,15 @@
+package xsbt
+
+import dotty.tools.dotc._
+import core.Contexts._
+
+class TestDriver extends Driver {
```
sjrd: Do you really need to make this public? Can't in `private` inside the companion object of `ScalaCompilerForUnitTesting`?

smarter: Sure, but it's a test so it doesn't matter much, and I'd like to avoid changing `ScalaCompilerForUnitTesting` too much since it's adapted from sbt and I synchronize it with the sbt version sometimes.

build.sbt
```diff
@@ -16,5 +16,3 @@ val `scala-library` = Build.`scala-library`
 val `scala-compiler` = Build.`scala-compiler`
 val `scala-reflect` = Build.`scala-reflect`
 val scalap = Build.scalap
-
-inThisBuild(Build.thisBuildSettings)
```
smarter: Good catch, will fix

project/Build.scala
smarter: Will do.

smarter:
I've opened https://github.com/lampepfl/dotty/issues/2335 to track the bootstrapped REPL issue.

smarter:
I managed to get the bootstrapped dotty compilable without publishing anything! See last commit. I'll merge this soon now that the REPL is fixed because this is a blocker for other things, but I'll still take into account further comments of course.

smarter:
@DarkDimius @felixmulder I've updated our Jenkins nightly build to run `dotty-bootstrapped/publishSigned` and successfully ran it once.

PR #2314, opened 2017-04-27
Remove ProductN parent on case classes

OlivierBlanvillain:
This means the compiler needs to systematically synthesise productElement and
productArity, which aligns with what scalac is doing.

This change gets rid of last reference to `scala.ProductN`! :tada:

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
+++ b/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
@@ -116,9 +116,12 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
      */
     def productElementBody(arity: Int, index: Tree)(implicit ctx: Context): Tree = {
       val ioob = defn.IndexOutOfBoundsException.typeRef
-      // That's not ioob.typeSymbol.primaryConstructor, this is the other one
-      // that takes a String argument.
-      val constructor = ioob.typeSymbol.info.decls.toList.tail.head.asTerm
+      // Second constructor of ioob that takes a String argument
+      def filterStringConstructor(s: Symbol): Boolean = s.info match {
+        case m: MethodType if s.isConstructor => m.paramInfos == List(defn.StringType)
+        case _ => false
+      }
+      val constructor = ioob.typeSymbol.info.decls.find(filterStringConstructor _).asTerm
       val stringIndex = Apply(Select(index, nme.toString_), Nil)
       val error = Throw(New(ioob, constructor, List(stringIndex)))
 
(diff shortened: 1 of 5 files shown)


OlivierBlanvillain:
I think productElement/productArity should be added in [SyntheticMethods](https://github.com/dotty-staging/dotty/blob/rip-22/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala) instead of desugaring. Drone only caught that with `legacyTests` (test` are passing), @felixmulder isn't this worrisome?

felixmulder:
Please investigate and add a meta test :)

odersky:
I agree; it would be more logical to add the methods in SyntheticMethods. Indeed we do have code there that adds productArity. productElement is missing, though.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -91,13 +92,49 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
...
+     *    case 1 => this._2
+     *    case _ => throw new IndexOutOfBoundsException(index.toString)
+     *  }
+     *  ```
+     */
+    def productElementBody(arity: Int, index: Tree)(implicit ctx: Context): Tree = {
+      val ioob = defn.IndexOutOfBoundsException.typeRef
+      // That's not ioob.typeSymbol.primaryConstructor, this is the other one
+      // that takes a String argument.
+      val constructor = ioob.typeSymbol.info.decls.toList.tail.head.asTerm
```
smarter: It would be better if you didn't rely on the order of declarations and instead filtered the declarations to only include the one that you want.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -91,25 +92,68 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
...
+     *    case _ => throw new IndexOutOfBoundsException(index.toString)
+     *  }
+     *  ```
+     */
+    def productElementBody(arity: Int, index: Tree)(implicit ctx: Context): Tree = {
+      val ioob = defn.IndexOutOfBoundsException.typeRef
+      // Second constructor of ioob that takes a String argument
+      def filterStringConstructor(s: Symbol): Boolean = s.info match {
+        case m: MethodType if s.isConstructor => m.paramInfos == List(defn.StringType)
+        case _ => false
```
odersky: I think there's an easier way to get this, but I'll play with it and do it in a different PR.

OlivierBlanvillain: `IndexOutOfBoundsException` contains two declarations, its primary constructor (which takes not argument), and this other constructor that takes a String. I first did `.decls.toList.tail.head`, then changed it to this very precise filter, do you have something in between?

PR #2318, opened 2017-04-27
[Vulpix] Abort compilation after `maxCompileTime`

felixmulder:


DarkDimius:
I think this is a very misleading change. 
It doesn't stop the compilation, it stops waiting for it and starts a new one, so that multiple will be performed concurrently.

DarkDimius:
Additionally, it blocks the compilation thread and instead uses global future threadpool, ie the number of compilation threads is not controlled using previous means anymore

felixmulder:
Non-constructive feedback. Please rephrase.

DarkDimius:
While I agree that I could have phrased it differently, I find a sneaky way of putting such changes without documenting them very misleading, hence my reaction. 

Constructive version of it was in the original https://github.com/lampepfl/dotty/pull/2125 PR.
The better way to do it would to have compilation slaves as separate VMs so that we are able to kill them.

felixmulder:
Fair enough :smiley: let's do this properly then!

PR #2322, opened 2017-04-27
Change backend name handling

odersky:
Change backend name handling to support symbolic names. The idea is to
narrow the interface to names in the backend interface and avoid direct calls
to toString on names.

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -535,11 +535,8 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     }
   }
 
-
   implicit def nameHelper(n: Name): NameHelper = new NameHelper {
-    def toTypeName: Name = n.toTypeName
     def isTypeName: Boolean = n.isTypeName
-    def toTermName: Name = n.toTermName
     def isTermName: Boolean = n.isTermName
     def startsWith(s: String): Boolean = n.startsWith(s)
   }
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -409,11 +409,10 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def newSet[K](): mutable.Set[K] = new mutable.HashSet[K]
   }
 
-
-
   val MODULE_INSTANCE_FIELD: String = nme.MODULE_INSTANCE_FIELD.toString
 
-  def internalNameString(offset: Int, length: Int): String = new String(Names.chrs, offset, length)
+  def dropModule(str: String) =
+    if (!str.isEmpty && str.last == '$') str.take(str.length - 1) else str
 
   def newTermName(prefix: String): Name = prefix.toTermName
 
@@ -423,7 +422,6 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
       Flags.Specialized | Flags.Lifted | Flags.Protected | Flags.JavaStatic |
       Flags.Bridge | Flags.VBridge | Flags.Private | Flags.Macro
   }.bits
-
 
   def isQualifierSafeToElide(qual: Tree): Boolean = tpd.isIdempotentExpr(qual)
   def desugarIdent(i: Ident): Option[tpd.Select] = {
@@ -557,7 +555,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def fullName: String = sym.showFullName
     def simpleName: Name = sym.name
     def javaSimpleName: Name = toDenot(sym).name // addModuleSuffix(simpleName.dropLocal)
-    def javaBinaryName: Name = javaClassName.replace('.', '/').toTypeName // TODO: can we make this a string? addModuleSuffix(fullNameInternal('/'))
+    def javaBinaryName: String = javaClassName.replace('.', '/') // TODO: can we make this a string? addModuleSuffix(fullNameInternal('/'))
     def javaClassName: String = toDenot(sym).fullName.toString// addModuleSuffix(fullNameInternal('.')).toString
     def name: Name = sym.name
     def rawname: Name = {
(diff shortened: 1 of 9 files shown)


odersky:
How did I verify that this works? Starting with 2c4ac2c, I added an assertion that `useMangled` is false when taking the `toString` of a name. I had to temporarily set `useMangled` in lots of front and middle end code called by the backend, so that in the end only real `toString` from the backend calls remained. I verified that all unit tests could be run without triggering the assertion. 

It's tricky to make this a CI-suitable test though, because it is inherently sequential. Name#toString doe snot know anything about the context, so we need a global variable to control the assertion, which is not thread-safe.

So for the moment I submit as is. If toString calls should be introduced in the future they would likely produce wrong code for symbolic names (once these are in). I agree it's a weak barrier against regression. Once get get parametetricity in Any, we can enforce it in the types, so then we will be in the dry.

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -537,32 +535,25 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
...
   implicit def symHelper(sym: Symbol): SymbolHelper = new SymbolHelper {
     // names
     def fullName(sep: Char): String = sym.showFullName
     def fullName: String = sym.showFullName
     def simpleName: Name = sym.name
-    def javaSimpleName: Name = toDenot(sym).name // addModuleSuffix(simpleName.dropLocal)
-    def javaBinaryName: Name = javaClassName.replace('.', '/').toTypeName // TODO: can we make this a string? addModuleSuffix(fullNameInternal('/'))
-    def javaClassName: String = toDenot(sym).fullName.toString// addModuleSuffix(fullNameInternal('.')).toString
+    def javaSimpleName: String = toDenot(sym).name.mangledString // addModuleSuffix(simpleName.dropLocal)
+    def javaBinaryName: String = javaClassName.replace('.', '/') // TODO: can we make this a string? addModuleSuffix(fullNameInternal('/'))
```
smarter: This TODO is no longer necessary.

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -67,6 +67,7 @@ object Names {
     def asSimpleName: SimpleTermName
     def toSimpleName: SimpleTermName
     def mangled: Name
```
smarter: Would be nice to add some documentation on the meaning/use of `mangled`.

smarter:
> It's tricky to make this a CI-suitable test though, because it is inherently sequential. Name#toString does not know about the context, so we need a global variable to control the assertion, which is not thread-safe.

How about doing this instead:

```scala
def toString = {
  if (Config.checkBackendNames) {
    assert(!Thread.currentThread.getStackTrace.exists(_.toString.contains("dotty.tools.dotc.backend.jvm.GenBCode"),
      "Backend should not call Name#toString, Name#mangledString should be used instead."))
  }
  // ...
}
```

smarter:
Better way to use `getStackTrace`:
```scala
Thread.currentThread.getStackTrace.exists(_.getClassName == classOf[GenBCode].getName)
```

odersky:
But then we'd have to leave in the ~15 exceptions which allow mangling. Possible but ugly.

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -293,6 +293,10 @@ object Names {
       else {
         if (Config.checkBackendNames) {
           if (!toStringOK) {
+            // We print the stacktrace instead of doing an assert directly,
+            // because asserts are caught in exception handlers which might
+            // cause other failures. In that case the first, important failure
+            // is lost.
             println("Backend should not call Name#toString, Name#mangledString should be used instead.")
```
smarter: Use `Console.err.println` (or `System.err.println` instead). `printStackTrace` will print on stderr, and if you use println right before it which prints on stdout, the output might get mixed.

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -293,6 +293,10 @@ object Names {
...
             new Error().printStackTrace()
```
smarter: Nicer way: `Thread.dumpStack()`

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -301,6 +305,9 @@ object Names {
         new String(chrs, start, length)
       }
 
+    /** It's OK to take a toString if the stacktrace does not occur a method
```
smarter: typo: "a method" -> "in a method"
And on the line below: "or it also" -> "or if it also"

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -289,7 +290,32 @@ object Names {
...
+        }
+        new String(chrs, start, length)
+      }
+
+    private def toStringOK = {
+      val trace = Thread.currentThread.getStackTrace
+      !trace.exists(_.getClassName.endsWith("GenBCode")) ||
+      trace.exists(elem =>
+          List(
+              "mangledString",
```
smarter: An alternative to defining a whitelist in this way would be to define a new method named something like `unmangledString` that does what `toString` currently do, and have `toString` forward to it. Then we can just change all the code calling `toString` to use `unmangledString` instead, but keep the stacktrace check in `toString`

odersky: I think that would affect too many things. Note that most of the whitelist does not call toString directly, sometimes there are 10-20 calls in between.

compiler/src/dotty/tools/dotc/config/Config.scala
```diff
@@ -49,6 +49,13 @@ object Config {
    */
   final val checkNoSkolemsInInfo = false
 
+  /** Check that Name#toString is not called directly from backend by analyzing
+   *  the stack trace of each toString call on names. This is very expensive,
+   *  so not suitable for continuous testing. But it can be used to find a problem
+   *  when running a specific test.
+   */
+  final val checkBackendNames = false
```
DarkDimius: I would propose to make this configurable as a flag. So that we can ask users to enable checked mode in case something looks very strange in their reports.

odersky: Maybe that's a bit paranoid? So far, errors resulting from producing symblic strings were very easy to diagnose.

PR #2323, opened 2017-04-28
Add scala scala submodule.

nicolasstucki:


DarkDimius:
LGTM after tests pass

smarter:
I would call it scala-library, not scala-scala

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -29,6 +29,7 @@ class CompilationTests extends ParallelTesting {
 
   @Test def compilePos: Unit = {
     compileList("compileStdLib", StdLibSources.whitelisted, scala2Mode.and("-migration", "-Yno-inline")) +
+    compileDir("../collection-strawman/src/main", defaultOptions) +
```
DarkDimius: maybe make it a separate test?

nicolasstucki: We would loose the parallelism.

project/Build.scala
```diff
@@ -857,12 +857,12 @@ object DottyInjectedPlugin extends AutoPlugin {
   }
 
   private def projectChecks(): Unit = {
-    val scalaScala = new File("scala-scala")
-    if (!scalaScala.exists()) {
+    val submodules = List(new File("scala-backend"), new File("scala-scala"), new File("collection-strawman"))
+    if (!submodules.forall(_.exists)) {
       println(
```
smarter: You shouldn't use println in the sbt build, see http://www.scala-sbt.org/0.13/docs/Howto-Logging.html#Log+messages+in+a+task for the proper way to do it. This should probably be logged at the error level.

smarter:
Please go through the documentation `doc/doc/` to remove references to scala-scala.

project/Build.scala
nicolasstucki: This is not a task.

smarter: But could it be one?

nicolasstucki:
I noticed that if I set the `scala-library` submodule to the `lampepfl/scala` with branch `sharing-backend` the only test that fails is `dotty.tools.dottydoc.TestWhitelistedCollections.arrayAndImmutableHasDocumentation`. If we fix this we could remove use a single library/backend submodule. @DarkDimius @smarter what do you think about that?

smarter:
@nicolasstucki You probably forgot to do `git submodule update`, there is no way that the library in sharing-backend compiles with dotty, it uses procedure syntax, etc.

nicolasstucki:
@smarter indeed at first I forgot to do `git submodule update` and noticed that it passed all tests except `TestWhitelistedCollections` on `lampepfl/scala` with branch `master`. Then I explicitly change the branch to `sharing-backend` and saw the same. 

All the tests on the `scala-library` use the scala2 flag. That is why it can handle procedure syntax, etc.

nicolasstucki:
@DarkDimius, could you check if it still LGTY. If this is merged it would simplify a bit the call graph branch tests.

nicolasstucki:
@smarter, do you agree with the changes in the last commit?

DarkDimius:
LGTM from my side, didn't check sbt though.

PR #2327, opened 2017-04-28
Fix #2324: Check contexts in right order when looking for idents

### LINKED ISSUE #2324: Incorrect ambiguous reference error when imports are shadowed by local definition
The following code emits an error in dotty but not scalac:
```scala
object A {
  def foo: Int = 1
}
object B {
  def foo: Int = 2
}
object C {
  import A._
  import B._

  def foo: Int = 3

  println(foo)
}
```
```scala
13 |  println(foo)
   |          ^^^
   |          reference to `foo` is ambiguous
   |          it is both imported by import A._
   |          and imported subsequently by import B._
```

This is of course incorrect, the reference is not ambiguous, it must be `C.foo`.

odersky:
The previous logic would look for members of a class in the outermost scope
where the class is owner. But it should be the innermost scope.

### DIFF
--- a/tests/neg/i1641.scala
+++ b/tests/neg/i1641.scala
@@ -2,7 +2,7 @@ package bar { object bippy extends (Double => String) { def apply(x: Double): St
 package object println { def bippy(x: Int, y: Int, z: Int) = "(Int, Int, Int)" }
 object Test {
   def main(args: Array[String]): Unit = {
-    println(bar.bippy(5.5))
+    println(bar.bippy(5.5)) // error
     println(bar.bippy(1, 2, 3)) // error
   }
 }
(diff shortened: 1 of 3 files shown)


tests/neg/i1641.scala
```diff
@@ -2,7 +2,7 @@ package bar { object bippy extends (Double => String) { def apply(x: Double): St
 package object println { def bippy(x: Int, y: Int, z: Int) = "(Int, Int, Int)" }
 object Test {
   def main(args: Array[String]): Unit = {
-    println(bar.bippy(5.5))
+    println(bar.bippy(5.5)) // error
     println(bar.bippy(1, 2, 3)) // error
```
smarter: Shouldn't there be two errors here now? One for the println and one for bippy. That's what happens in scalac.

odersky: Yes, but that's unrelated. Maybe they have the same position in dotty. Anyway we went from 1 to 2 errors, so it's an improvement :-)

PR #2331, opened 2017-04-30
Bug fixes to make collection strawman compile

odersky:
Replaces #2240, and is based on #2330. Compared to #2240, this PR does not need tricky language changes. The changes to union types have already been subsumed in a simpler and more general
fashion in #2330. So all that remains here is bug fixes and better diagnostic capabilties.


odersky:
Comment copy pasted from #2240:

I'd like to get this in soon, because I fear that if we wait, collections will get even more code that will fail in dotty. So we need to get to a point to do CI of the strawman with dotty and scalac. For this we need:

 - this PR to go in.
 - #2314 to be fixed and merged.
 - The bridge issue discovered by @julienrf when running the strawman to be fixed.

/cc @julienrf @OlivierBlanvillain @DarkDimius

DarkDimius:
@odersky 
>The bridge issue discovered by @julienrf when running the strawman to be fixed.

could you please point me to the issue?

julienrf:
@DarkDimius I only posted a small comment about it: https://github.com/scala/collection-strawman/pull/61#issue-222366768

odersky:
@DarkDimius The code pointed to by @julierf is what I showed you.

compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -430,16 +368,70 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
...
+   *
+   */
+  def forwardParamBindings(parentRefs: List[TypeRef],
+                           paramBindings: SimpleMap[TypeName, Type],
+                           cls: ClassSymbol, decls: Scope)(implicit ctx: Context) = {
+
+    def forwardRef(argSym: Symbol, from: TypeName, to: TypeAlias) = argSym.info match {
+      case info @ TypeAlias(TypeRef(_: ThisType, `from`)) if info.variance == to.variance =>
+        val existing = decls.lookup(argSym.name)
+        if (existing.exists) existing.info = existing.info & to
```
smarter: `enterArgBindings` sets the info using a `LazyType`, shouldn't the same thing be done here?

compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -430,16 +368,70 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
...
+   *     type param v= middle
+   *     type middle v= to
+   *
+   *  where the variances of both alias are the same, then enter a new definition
+   *
+   *     type param v= to
+   *
+   *  If multiple forwarders would be generated, join their `to` types with an `&`.
+   *
+   *  @param cls           The class for which parameter bindings should be forwarded
```
smarter: The params are not in the same order in the doc than in the declaration.

compiler/src/dotty/tools/dotc/core/TypeOps.scala
odersky: It seems we get away with it. I have a tendency to wait for an actual CyclicReference popping up before we take that measure.

PR #2332, opened 2017-04-30
Fix deprecation messages missing a position

smarter:
`pos` was accidentally passed as a fourth argument to `format` and
ignored instead of as a second argument to `deprecationWarning`.


smarter:
No test because we don't have `-Xfatal-warnings` implemented, but you can verify that when running `sbt dotty-compiler-bootstrapped/compile` the deprecation messages have a position.

odersky:
Good illustration of the hidden dangers of complex parameter lists... 

LGTM

PR #2343, opened 2017-05-01
Build.sbt: switch back to Maven-published sbt-interface

smarter:
We couldn't do this before because the sbt-interface wasn't published
but this is now fixed (https://github.com/typesafehub/sbt-republish/pull/15).

This avoids everyone using dotty having to add the typesafe-ivy releases
repo to their build.sbt.


felixmulder:
Great!

PR #2346, opened 2017-05-02
Fix #2345: Handle case of overloaded local vars in assignment

### LINKED ISSUE #2345: Reassignment to var considered mistaken for reassignment to val
```scala
package dotty.tools.util

import scala.collection.mutable

class Whatever[A] extends mutable.Set[A] {
  private[this] var count = 0
  count = 0
}
```

Fails with:

```
-- Error: local/set.scala:7:8 --------------------------------------------------
7 |  count = 0
  |  ^^^^^^^^^
  |  reassignment to val
```

odersky:
Clean up logic of typedAssign, so that it now correctly handles the case
of an overloaded left-hand-side consisting of a provate[this] var
and one or more methods.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -567,37 +567,42 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       case lhs =>
         val lhsCore = typedUnadapted(lhs, AssignProto)
         def lhs1 = typed(untpd.TypedSplice(lhsCore))
-        def canAssign(sym: Symbol) = // allow assignments from the primary constructor to class fields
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -567,37 +567,42 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       case lhs =>
         val lhsCore = typedUnadapted(lhs, AssignProto)
         def lhs1 = typed(untpd.TypedSplice(lhsCore))
-        def canAssign(sym: Symbol) = // allow assignments from the primary constructor to class fields
+        lazy val lhsVal = lhsCore.asInstanceOf[TermRef].denot.suchThat(!_.is(Method))
```
smarter: That lazy val seems unused.

PR #2348, opened 2017-05-02
Do not accidentally force diagnostic message in `reportWarning` 

felixmulder:


smarter:
One way would be to call it `contained()` and not `contained`

PR #2349, opened 2017-05-02
Fix #2292: Discard trees with errors early in Frontend

### LINKED ISSUE #2292: crash on compilable code: type of  ... is not assigned
The latest stable release of scalac (version 2.12) can compile, but dotc crashes.
```
$ cat abc.scala
package +
import tools.nsc._, reporters._, typechecker._
class CustomGlobal(currentSettings: Reporter) extends Global(currentSettings) {
  override lazy val analyzer = new { val global: this.type = this }
  with Analyzer {
    override def newTyper(context: Context) = new CustomTyper(context)
    class CustomTyper(context: Context) extends Typer(context) {
      override def typed(tree: Tree, pt: Type) = {
        if (tree.summaryString contains "Bippy") ()
        super.typed(tree, pt)
      }
    }
  }
}

```
```
$ dotc abc.scala
-- [E009] Syntax Error: abc.scala:5:2 ------------------------------------------
5 |  with Analyzer {
  |  ^^^^
  |  early definitions are not supported; use trait parameters instead

-- [E006] Unbound Identifier Error: abc.scala:1:8 ------------------------------
1 |package +
  |        ^
  |        not found: +

Exception in thread "main" dotty.tools.dotc.ast.Trees$UnAssignedTypeException: type of TypeDef(CustomGlobal,Template(DefDef(<init>,List(),List(List(ValDef(currentSettings,Ident(Reporter),EmptyTree))),TypeTree,EmptyTree),List(Apply(Select(New(Ident(Global)),<init>),List(Ident(currentSettings)))),ValDef(_,EmptyTree,EmptyTree),List(ValDef(analyzer,TypeTree,New(Template(DefDef(<init>,List(),List(),TypeTree,EmptyTree),List(),ValDef(_,EmptyTree,EmptyTree),List(ValDef(global,SingletonTypeTree(This(Ident())),This(Ident()))))))))) is not assigned
	at dotty.tools.dotc.ast.Trees$Tree.tpe(Trees.scala:97)
	at dotty.tools.dotc.ast.Trees$DenotingTree.denot(Trees.scala:236)
	at dotty.tools.dotc.ast.Trees$Tree.symbol(Trees.scala:148)
	at dotty.tools.dotc.typer.FrontEnd.firstTopLevelDef(FrontEnd.scala:73)
	at dotty.tools.dotc.typer.FrontEnd.discardAfterTyper(FrontEnd.scala:78)
<...>
```

smarter:
Previously, we called `firstTopLevelDef(unit.tpdTree)` but this can
crash with an `UnAssignedTypeException` if a top-level class is
contained in a package which got an error type. Since we will always
discard trees with errors, we can just make sure that this is done
before the call to `firstTopLevelDef`

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/FrontEnd.scala
+++ b/compiler/src/dotty/tools/dotc/typer/FrontEnd.scala
@@ -75,7 +75,7 @@ class FrontEnd extends Phase {
   }
 
   protected def discardAfterTyper(unit: CompilationUnit)(implicit ctx: Context) =
-    unit.isJava || firstTopLevelDef(unit.tpdTree :: Nil).isPrimitiveValueClass
+    ctx.reporter.hasErrors || unit.isJava || firstTopLevelDef(unit.tpdTree :: Nil).isPrimitiveValueClass
 
   override def runOn(units: List[CompilationUnit])(implicit ctx: Context): List[CompilationUnit] = {
     val unitContexts = for (unit <- units) yield {
--- a/tests/neg/i2292.scala
+++ b/tests/neg/i2292.scala
@@ -1,0 +1,3 @@
+package + // error
+
+class Foo


smarter:
Closing because this breaks the IDE: we don't want to discard units that do not typecheck because we still may use them in the IDE.

smarter:
Replaced by https://github.com/lampepfl/dotty/pull/2352

PR #2354, opened 2017-05-02
Moved Ambiguous Overload error in Typer to case class

ennru:
@felixmulder Couldn't decide on best solution how `typer.ErrorReporting.Errors` methods should be used from error case class. There are more errors to move that will need it.


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1224,26 @@ object messages {
...
+      s"""|Ambiguous overload. The ${err.overloadedAltsStr(alts)}
+          |$all match ${err.expectedTypeStr(pt)}""".stripMargin
+
+    override def kind: String = "Reference"
+
+    override def explanation: String =
+       s"""|There are ${alts.length} methods that could be referenced as the compiler knows too little about the expected type.
+           |You may specify the expected type e.g. by
+           |- assigning it to a value with a specified type, or
+           |- adding a type ascription as in (MyClass.myMethod: String => Int)
```
senia-psm: It looks like the type ascription example should be `myObject.myMethod: String => Int`

ennru: Yes, thank you. Preferrable I'd show the method in question, but as far as I know there is no easy way right now to print out the type with `=>`-syntax.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1224,26 @@ object messages {
...
+      s"""|Ambiguous overload. The ${err.overloadedAltsStr(alts)}
+          |$all match ${err.expectedTypeStr(pt)}""".stripMargin
+
+    override def kind: String = "Reference"
+
+    override def explanation: String =
+       s"""|There are ${alts.length} methods that could be referenced as the compiler knows too little about the expected type.
+           |You may specify the expected type e.g. by
+           |- assigning it to a value with a specified type, or
+           |- adding a type ascription as in `(instance.myMethod: String => Int)`
```
felixmulder: Highlight the code example?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1224,26 @@ object messages {
...
+    private def all = if (alts.length == 2) "both" else "all"
+
+    override def msg: String =
+      s"""|Ambiguous overload. The ${err.overloadedAltsStr(alts)}
+          |$all match ${err.expectedTypeStr(pt)}""".stripMargin
+
+    override def kind: String = "Reference"
+
+    override def explanation: String =
+       s"""|There are ${alts.length} methods that could be referenced as the compiler knows too little about the expected type.
```
felixmulder: This line is > 100 columns. Eventually, maybe we should have the `hl` interpolator wrap the lines.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1224,26 @@ object messages {
...
   }
+
+  case class AmbiguousOverload(tree: tpd.Tree, alts: List[SingleDenotation], pt: Type)(
+    err: typer.ErrorReporting.Errors)(
+    implicit ctx: Context)
+  extends Message(AmbiguousOverloadID) {
+
+    private def all = if (alts.length == 2) "both" else "all"
+
+    override def msg: String =
```
felixmulder: These shouldn't be `override`, I'm guessing you looked at the one above and just assumed that we write it like this now. Could you change both? 😄

felixmulder:
PS - commit messages should be in the imperative form. E.g. "Add X" over "Added X"

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: Yes. Started off from the wrong example. Will fix.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: `hl` will need to start wrapping at some stage as we put in dynamic amount of text. I will wrap this one.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: Any sight of supporting a `type.show`-variant with notation as in source code?

felixmulder: Types are `Showable` in Dotty as long as you have an implicit `Context` in scope, you can do `tpe.show`. If that's not what you meant, then please elaborate :)

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Yeah, you're completely right. There are some cases where we don't want it to wrap. For instance on lines beginning with `-, +, *,  , ` and so on... maybe we could just principle it on wrapping over 80 characters until it hits two newlines or one of the disallowed symbols. It would also need to take into account ANSI escape codes for the colors.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: `tpe.show` gives `(s: String)Int`, but I'd like to get `(String) => Int` for better cut-and-paste fidelity.

felixmulder:
LGTM, but needs a rebase

ShaneDelmore:
@felixmulder I just noticed dottybot gives the opposite advice you do over added vs add here https://github.com/lampepfl/dotty/blob/e4e481034a309d7e23c0b6b2abb87795b2d987bc/bot/src/dotty/tools/bot/PullRequestService.scala#L270

felixmulder:
@ShaneDelmore - it should be the imperative mood, e.g "add" over "added" :) So they are swapped - would you mind unswapping? :)

PR #2357, opened 2017-05-02
Moved Method Does Not Take (More) Parameters error to case class

ennru:
@felixmulder Same solution regarding `typer.ErrorReporting.Errors` here.


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1223,22 @@ object messages {
...
+      case _ => ""
+    }
+
+    val msg = hl"${err.refStr(methPartType)} does not take$more parameters"
+
+    val kind = "Reference"
+
+    val explanation =
+      hl"""|You have specified more parameter lists as defined in the method definition(s).
+           |In case ${err.refStr(methPartType)} is defined without parenthesis, you may
```
smarter: You should be able to only display this when `methPartType` is actually defined without parenthesis. I think `methPartType.isInstanceOf[ExprType]` should check for that.

ennru: Great, was looking around for an easy enough check to do just that.

ennru: @smarter I find a `TermRefWithSignature` in my samples, do I need to test for just that?

smarter: Try `methPartType.widenSingleton.isInstanceOf[ExprType]`

smarter: (A `TermRef` might not always be a reference to a method with no arguments, so we widen it to its underlying type)

felixmulder:
Also needs rebase, then LGTM

felixmulder:
@dotty-bot: could you recheck please?

PR #2358, opened 2017-05-03
Fix #2143: remove widenForMatchSelector in typing patmat

### LINKED ISSUE #2143: patmat selector gets widened too much
```scala
object Test {
  type T <: AnyRef

  def test(x: T): Unit = {
    x match {
      case foo =>
        val z: T = foo
    }
  }
}
```
This compiles with scalac but fails in dotty with:
```scala
-- [E007] Type Mismatch Error: try/collect.scala -------------------------------
7 |        val z: T = foo
  |                   ^^^
  |                   found:    Object(foo)
  |                   required: Test.T
```
This happens because we aggressively widen the selector in `typedMatch`:
```scala
        val selType = widenForMatchSelector(
            fullyDefinedType(sel1.tpe, "pattern selector", tree.pos))
```
Replacing `widenForMatchSelector` for by a call to `.widen` fixes the issue, but breaks one test:  `tests/patmat/for.scala` which is run with `-Ycheck-all-patmat` and should not report any exhaustivity warning, I'm not sure why there's a difference here. A possible solution would be to call `widenForMatchSelector` inside the exhaustivity checker, but I don't know if that gives the correct semantics.

liufengyun:
Fix #2143: remove `widenForMatchSelector` in typing pattern matches.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -354,6 +354,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     case Bind(_, pat) => project(pat)
     case UnApply(_, _, pats) =>
       if (pat.tpe.classSymbol.is(CaseClass))
+        // FIXME: why dealias is needed here?
         Kon(pat.tpe.stripAnnots.dealias, pats.map(pat => project(pat, roundUp)))
       else if (roundUp) Typ(pat.tpe.stripAnnots, false)
       else Empty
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -354,7 +354,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     case Bind(_, pat) => project(pat)
     case UnApply(_, _, pats) =>
       if (pat.tpe.classSymbol.is(CaseClass))
-        Kon(pat.tpe.stripAnnots, pats.map(pat => project(pat, roundUp)))
+        Kon(pat.tpe.stripAnnots.dealias, pats.map(pat => project(pat, roundUp)))
```
liufengyun: This change fix the exhaustivity check if we remove the `widenForMatchSelector`. But normally, shouldn't `TypeComparer` take care of dealiasing in subtype checking? Ping @odersky @smarter ?

felixmulder:
Put a FIXME then LGTM

PR #2361, opened 2017-05-03
Merge the sbt dotty plugin in this repo

smarter:
It previously lived at https://github.com/lampepfl/sbt-dotty


smarter:
I've also updated https://github.com/smarter/dotty-example-project to use the sbt-dotty plugin so things should be much less confused now. However, it is very important that someone takes the time to make a more complex examples that illustrates cross-compilation with dotty, and I don"t have time to do that now.

smarter:
Note: I think it'd make sense to make the scripted tests a separate job, but that will slow the CI down unless we increase the number of jobs

docs/_includes/getting-started.html
```diff
@@ -2,29 +2,7 @@
...
-        <h2 id="build.properties">build.properties</h2>
-        <pre class="sourceCode bordered"><code>sbt.version=0.13.11</code></pre>
-        <h2 id="plugins.sbt">plugins.sbt</h2>
-        <pre class="sourceCode bordered"><code>addSbtPlugin("com.felixmulder" % "sbt-dotty" % "0.1.7")</code></pre>
-        <p>
-            This plugin is based on the
-            <a href="https://github.com/smarter/dotty-example-project">
-                dotty-example-project
-            </a>, but let’s you skip building dotty from scratch.
+            See the <a href="https://github.com/lampepfl/dotty-example-project">example project</a>.
```
felixmulder: I don't like *removing* this, but I'm sure your reason is that we'd need to keep it in sync with the example project. I feel that it will take everyone who's interested in trying Dotty through multiple hoops. I'd like it to just be "google dotty, find webpage, minimal sbt config, run dotty". Instead of "google dotty, find webpage, find link to example project, read most of README to know which options to use, set up sbt config, run dotty".

Maybe we should restructure the intro page and put this information at the top after the logo and show a condensed version of the dotty example project README which we also refer to for more information.

docs/docs/usage/sbt-projects.md
```diff
@@ -3,10 +3,4 @@ layout: doc-page
...
 ---
 
-It is now possible to use Dotty with sbt thanks to the dotty-bridge project.
-There are two alternatives in how to create an sbt project that uses dotty:
-
-* [dotty-example-project](https://github.com/smarter/dotty-example-project)
-  for a simple sbt project that compiles code with Dotty
-* [sbt-dotty](https://github.com/felixmulder/sbt-dotty) an sbt plugin that
-  takes care of all dependencies and settings needed to get a Dotty sbt project
+See the [example project](https://github.com/lampepfl/dotty-example-project).
```
felixmulder: We could just put the config info from the front-page into an `_include` and as such we don't need to duplicate any info here.

felixmulder:
I'm fine with increasing the number of jobs running in parallel on the CI. I just haven't gotten to it because of flatMap(Oslo).

felixmulder:
Also, changes to example project README look very nice :+1:

docs/_includes/getting-started.html
smarter: > Instead of "google dotty, find webpage, find link to example project, read most of README to know which options to use, set up sbt config, run dotty".

We could make an `sbt new` template and add the command to create it on the homepage: http://www.scala-sbt.org/0.13/docs/sbt-new-and-Templates.html

smarter: > Maybe we should restructure the intro page and put this information at the top after the logo and show a condensed version of the dotty example project README which we also refer to for more information.

I really don't want to duplicate information, we're already doing a bad job at keeping one copy of our documentation up to date, and we completely fail for information in multiple places (the homepage talked about a very old version of the sbt-dotty plugin).

docs/docs/usage/sbt-projects.md
smarter: I'll let you do that in a follow-up PR ;)

smarter:
OK with merging this?

felixmulder:
I'm fine with it, but we should really clean it up. I'm not sure we should even have an example project as there **is** an example project in `sbt-dotty`.

smarter:
Having an example project as a separate repo make it easy for someone to experiment with, they can just git clone it and treat is as a starting point for their project. And I'm pretty sure we need at least two example projects: one that just compiles with dotty and one that cross-compiles, they both can be useful.

felixmulder:
Sure, but in that case we should move the documentation to this repo and have it as an include and then in dotty-example-project refer to this repo and the docs.

PR #2362, opened 2017-05-03
Change Ambiguous Import error in Typer to case class

ennru:
Another more structured error message with hopefully helpful explanations.

@odersky I tried to trigger `error(em"reference to `$name` is ambiguous; it is imported twice ...` in Typer.scala:206, but the existing tests all end up in the error detection I now changed. Do you know of an example for that path?


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1223,4 +1223,39 @@ object messages {
...
+
+    val msg =
+      s"""|reference to `${hl"$name"}` is ambiguous
+          |it is both ${bindingString(newPrec, ctx)}
+          |and ${bindingString(prevPrec, prevCtx, " subsequently")}""".stripMargin
+
+    val kind = "Reference"
+
+    val explanation =
+      hl"""|The compiler can't decide which of the possible choices you
```
ennru: I agree, the current formatters don't compose well. @felixmulder and I discussed the need automated line wrapping as well.

PR #2364, opened 2017-05-04
Fix #2360: Fix inlined module ref.

### LINKED ISSUE #2360: Compiler crash when inlining methods on imported objects.
```scala
object Test {
    import Foo._
    foo
   // but Foo.foo works fine
  }
}

object Foo extends Bar {
  inline def foo: Int = bar
}

class Bar {
  def bar = 42
}
```

fails compilation with the following error

```scala
exception while typing val Foo$_this: Foo.type(Foo) = this of class class dotty.tools.dotc.ast.Trees$ValDef # 99
exception while typing /* inlined from Foo*/ 
  {
    val Foo$_this: Foo.type(Foo) = this
    Foo$_this.bar
  } of class class dotty.tools.dotc.ast.Trees$Inlined # 143
exception while typing {
  /* inlined from Foo*/ 
    {
      val Foo$_this: Foo.type(Foo) = this
      Foo$_this.bar
    }
  ()
} of class class dotty.tools.dotc.ast.Trees$Block # 178
exception while typing {
  {
    /* inlined from Foo*/ 
      {
        val Foo$_this: Foo.type(Foo) = this
        Foo$_this.bar
      }
    ()
  }
} of class class dotty.tools.dotc.ast.Trees$Block # 179
exception while typing def main(args: Array[String]): Unit = 
  {
    {
      /* inlined from Foo*/ 
        {
          val Foo$_this: Foo.type(Foo) = this
          Foo$_this.bar
        }
      ()
    }
  } of class class dotty.tools.dotc.ast.Trees$DefDef # 180
exception while typing @scala.annotation.internal.SourceFile("/Users/nicolasstucki/GitHub/dotty/compiler/../tests/run/inline-foo.scala") 
  final module
 class Test() extends Object() { 
    {
      {
        /* inlined from Foo*/ 
          {
            val Foo$_this: Foo.type(Foo) = this
            Foo$_this.bar
          }
        ()
      }
    }
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 183
exception while typing package <empty> {
  final lazy module val Test: Test = new Test()
  @scala.annotation.internal.SourceFile("/Users/nicolasstucki/GitHub/dotty/compiler/../tests/run/inline-foo.scala") 
    final module
   class Test() extends Object() { 
      {
        {
          /* inlined from Foo*/ 
            {
              val Foo$_this: Foo.type(Foo) = this
              Foo$_this.bar
            }
          ()
        }
      }
  }
  final lazy module val Foo: Foo = new Foo()
  @scala.annotation.internal.SourceFile("/Users/nicolasstucki/GitHub/dotty/compiler/../tests/run/inline-foo.scala") 
    final module
   class Foo() extends Bar() { 
    @Foo.bar inline def foo: Int = Foo.bar
  }
  @scala.annotation.internal.SourceFile("/Users/nicolasstucki/GitHub/dotty/compiler/../tests/run/inline-foo.scala") 
    class
   Bar() extends Object() { 
    def bar: Int = 42
  }
} of class class dotty.tools.dotc.ast.Trees$PackageDef # 199
*** error while checking /Users/nicolasstucki/GitHub/dotty/compiler/../tests/run/inline-foo.scala after phase tailrec ***
```

It looks like the `this` in `val Foo$_this: Foo.type(Foo) = this` is taken as the local `this` instead of the `Foo`. I would expect `val Foo$_this: Foo.type(Foo) = Foo` or some variant of that.

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -409,11 +409,14 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
...
     for ((level, selfSym) <- sortedProxies) {
+      lazy val rhsClsSym = selfSym.info.widenDealias.classSymbol
       val rhs =
-        if (!lastSelf.exists)
-          prefix
-        else
+        if (lastSelf.exists)
           untpd.Select(ref(lastSelf), OuterSelectName(EmptyTermName, lastLevel - level)).withType(selfSym.info)
+        else if (rhsClsSym.is(Module))
+          ref(rhsClsSym.companionModule)
```
odersky: Shouldn't this be `rhsClsSym.sourceModule`?

nicolasstucki: In this case `rhsClsSym.companionModule` and `rhsClsSym. sourceModule` have the same symbol. Should I change it to `rhsClsSym. sourceModule`?

odersky: Yes, it's sourceModule for going from a module class to its value. companionModule works sort of by accident.

nicolasstucki:
Updated

compiler/src/dotty/tools/dotc/typer/Inliner.scala
nicolasstucki: Updated

odersky:
LGTM, thanks!

PR #2365, opened 2017-05-04
Add -nowarn option to silence warnings

### LINKED ISSUE #2351: Implement "-nowarn" flag
It would be beneficial to be able to silence all warnings by adding the `-nowarn` flag.

* Add a `BooleanSetting` for `-nowarn` to [ScalaSettings](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala)
* Edit [Reporter#reportWarning](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/reporting/Reporter.scala#L40-L42) to do nothing if the flag is set e.g: `if (this.settings.nowarn) ()`

markus1189:
This was too low hanging fruit to resist ;)

Fixes #2351

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -41,6 +41,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val strict = BooleanSetting("-strict", "Use strict type rules, which means some formerly legal code does not typecheck anymore.")
   val language = MultiStringSetting("-language", "feature", "Enable one or more language features.")
   val rewrite = OptionSetting[Rewrites]("-rewrite", "When used in conjunction with -language:Scala2 rewrites sources to migrate to new syntax")
+  val silentWarnings = BooleanSetting("-nowarn", "Silence all warnings.")
 
   /** -X "Advanced" settings
    */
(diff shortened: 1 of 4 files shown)


felixmulder:
Looks great to me!

Would you mind adding a test in a second commit? One way to test this would be to combine `-nowarn -Xfatal-warnings` and add a test to `./tests/pos-special/` that generates a warning.

The behaviour should be such that the warning would be fatal if issued, but since it's canceled out by `-nowarn` the test should compile without errors.

Would you mind adding that? Otherwise I'm fine with merging as is :tada:

Another way to test it would be to add a test to [ErrorMessagesTests.scala](https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala) (which should be renamed to `MessageTests`). But maybe that's overkill :)

markus1189:
Sure I'll add it, I'll ping you when I gave a good test ;)

nicolasstucki:
@markus1189 you should name the commit `Fix #2351: Add -nowarn option to silence warnings.`, this way when we merge this PR the issue will be closed with it.

markus1189:
Okay I added it to `CompilationTests` and reused the `-Xfatal-warnings` file to write an additonal tests that with `-nowarn` it does pass instead of fail.  I chose this because it is easier than propagating the settings in a way that is convenient to use in `ErrorMessagesTests`.

felixmulder:
Thanks a bunch @markus1189! :tada:

smarter:
There's an empty file ` tests/pos-special/nowarn.scala` in this PR, I don't think that's intentional.

felixmulder:
Seems to have disappeared magically...

PR #2366, opened 2017-05-04
.gitignore: add more Scala-IDE specific files

smarter:


sjrd:
IDE-specific ignores should be in their respective users' *global* gitignore file. Just sayin' ...

PR #2377, opened 2017-05-05
Fix #2375: Memoize should not generate static fields.

### LINKED ISSUE #2375: Static vars generate wrong field
If we look at the bytecode produced from run/statics.scala, we see:

```
Users/odersky/workspace/dotty/tests/run> javap -cp /classes -private Foo
Compiled from "statics.scala"
public class Foo {
  public static final long OFFSET$0;
  public long bitmap$0;
  public Foo$Bar$ Bar$lzy1;
  public static int mutable;
  public static int field;
  private static int mutable$$local;
  public Foo();
  private static {};
  public final Foo$Bar$ Bar();
  public static int method();
  public static void mutable_$eq(int);
  public static int accessor();
}
```

Note that both `mutable` and `mutable$$local` are fields. `mutable` should be an accessor method instead. This bug prevents us from dropping `$$local` suffixes in semantic names.

DarkDimius:
They are not transformed by Getters phase,
they keep the original name and never renamed.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Memoize.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Memoize.scala
@@ -69,6 +69,7 @@ import Decorators._
     val sym = tree.symbol
 
     def newField = {
+      assert(!sym.hasAnnotation(defn.ScalaStaticAnnot))
       val fieldType =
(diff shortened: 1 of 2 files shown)


odersky:
LGTM

PR #2383, opened 2017-05-05
dotty-bot: don't ask people to set a reviewer

smarter:
- They can't set one (only contributors can)
- They would have no idea who to set it to anyway.


smarter:
Oh my god, is he really going to comment on every PR? This is super annoying, please stop this.

PR #2384, opened 2017-05-06
Make bot check contributors before sending heart-warming greeting

felixmulder:


smarter:
@felixmulder That's not enough in my opinion, even external contributors shouldn't be spammed by the same message from the bot for every single PR they make, they will get annoyed very quickly. The bot should never send the same greeting message twice to a single person. Maybe you could use the github "first-time contributor" mark if it's accessible from the API.

felixmulder:
@smarter - this PR makes the bot check if the person who authored the PR is amongst the people in https://github.com/lampepfl/dotty/graphs/contributors. Which means, once they've had a PR merged, they will not be greeted by the bot unless they've made some mistake that the bot finds.

felixmulder:
If you think it should be more strict, you're welcome to take a crack at it. But it's Saturday - so I'm off duty.

smarter:
Ah I see, then that should be ok.

PR #2386, opened 2017-05-06
Fix #2378: Better error message for extractors

### LINKED ISSUE #2378: Strange type checking error for extractors
The following code type checks fine in Scalac, but failed in Dotty.

```Scala
trait Cap

trait Toolbox {
  type Tree

  val tpd: TypedTrees
  trait TypedTrees {
    type Tree
  }

  val Apply: ApplyImpl
  trait ApplyImpl {
    def unapply(tree: Tree): Option[(Tree, Seq[Tree])]
    def unapply(tree: tpd.Tree)(implicit c: Cap): Option[(tpd.Tree, Seq[tpd.Tree])]
  }
}

class Test(tb: Toolbox) {
  import tb._
  implicit val cap: Cap = null

  def foo(tree: Tree): Int = tree match {
    case Apply(fun, args) => 3
  }

  def bar(tree: tpd.Tree): Int = tree match {
    case Apply(fun, args) => 3
  }
}
```

Error message:

```
-- Error: examples/abstract.scala:22:6 -----------------------------------------
22 |  def foo(tree: Tree): Int = tree match {
   |      ^
   |      non-private method foo refers to private value tb
   |       in its type signature (tree: Test.this.tb.Tree)Int
-- [E006] Unbound Identifier Error: examples/abstract.scala:23:9 ---------------
23 |    case Apply(fun, args) => 3
   |         ^^^^^
   |         not found: type Apply

-- Error: examples/abstract.scala:26:6 -----------------------------------------
26 |  def bar(tree: tpd.Tree): Int = tree match {
   |      ^
   |      non-private method bar refers to private value tb
   |       in its type signature (tree: Test.this.tb.tpd.Tree)Int
-- [E006] Unbound Identifier Error: examples/abstract.scala:27:9 ---------------
27 |    case Apply(fun, args) => 3
   |         ^^^^^
   |         not found: type Apply

```

odersky:


odersky:
> Now it produces the following error message, which is incorrect:

I know... But it's better then what we had, no? It's surprisingly difficult to do better. I tried, but had to postpone due to lack of time.

> BTW, it seems you accidentally added scala-scala.

Ah, that seems to be the source of my problems. Git claims it's a modified file that needs to be added. But if I do that CI breaks. It would be good if someone told me what the issue here is.

liufengyun:
> Git claims it's a modified file that needs to be added. But if I do that CI breaks. It would be good if someone told me what the issue here is.

The simple rule works for me is: if there's no reason to forward a submodule to a specific new version, never add changes in the submodule folder. The changes might be some generated files that are not added to `.gitignore` of the submodule.

> It's surprisingly difficult to do better. I tried, but had to postpone due to lack of time.

I see it's difficult to do given our current `tryEither`, `typeIdent`, `typeSelect` structure. Later we can consider refactor parts of `typeIdent` and `typeSelect` to a standalone `NameResolution`, which may help. I agree we can address this later.

BTW, you need to using the following test case, as the original test case is already fixed in #2358 .

```Scala
trait Cap
trait Tree

object Apply {
  def unapply(tree: Tree): Option[(Tree, Seq[Tree])] = ???
  def unapply(tree: Tree)(implicit c: Cap): Option[(Tree, Seq[Tree])] = ???
}

class Test {
  def foo(tree: Tree): Int = tree match {
    case Apply(fun, args) => 3
  }
}
```

PR #2398, opened 2017-05-08
Generalize function symbol synthesis logic

odersky:
We will need other synthesized symbols down the road (e.g.
function types in phantom universes). With this commit one
can provide for that by installing a synthesizer in a
scope.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -108,7 +108,7 @@ class Definitions {
    *        def apply(implicit $x0: T0, ..., $x{N_1}: T{N-1}): R
    *      }
    */
-  def newFunctionNTrait(name: TypeName) = {
+  def newFunctionNTrait(name: TypeName): ClassSymbol = {
     val completer = new LazyType {
       def complete(denot: SymDenotation)(implicit ctx: Context): Unit = {
         val cls = denot.asClass.classSymbol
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -189,7 +189,14 @@ class Definitions {
 
   lazy val ScalaPackageVal = ctx.requiredPackage("scala")
   lazy val ScalaMathPackageVal = ctx.requiredPackage("scala.math")
-  lazy val ScalaPackageClass = ScalaPackageVal.moduleClass.asClass
+  lazy val ScalaPackageClass = {
+    val cls = ScalaPackageVal.moduleClass.asClass
+    cls.info.decls.openForMutations.useSynthesizer(
+      name => ctx =>
```
smarter: Where is the `ctx` parameter used here?

compiler/src/dotty/tools/dotc/core/Scopes.scala
```diff
@@ -204,6 +210,12 @@ object Scopes {
      */
     private var elemsCache: List[Symbol] = null
 
+    /** The synthesizer to be used, or `null` if no synthesis is done on this scope */
+    private var synthesize: SymbolSynthesizer = null
```
smarter: Note that with both scalac and doty, not using `private[this]` here will generate a pair of getter and setter, this is arguably something that we should fix in the compiler since `private[this]` and `private` should be considered equivalent in an object.

compiler/src/dotty/tools/dotc/core/Scopes.scala
```diff
@@ -39,6 +39,12 @@ object Scopes {
    */
   private final val MaxRecursions = 1000
 
+  /** A function that optionally produces synthesized symbols with
+   *  the given name in the given context. Returns `NoSymbol` if the
+   *  no symbol should be synthesized for the given name.
+   */
+  type SymbolSynthesizer = Name => Context => Symbol
```
smarter: For efficiency, I would consider using `(Name, Context) => Symbol` instead, but that might be a premature optimization.

compiler/src/dotty/tools/dotc/core/Definitions.scala
nicolasstucki: This might be needed for other synthetic functions like the functions that take phantom parameters.

compiler/src/dotty/tools/dotc/core/Scopes.scala
nicolasstucki: Let's leave it like this. Do we have an issue for this?

compiler/src/dotty/tools/dotc/core/Scopes.scala
nicolasstucki: I would assume that with our current simple use cases the JVM must be capable of inlineing the this additional lambda. Seem a bit premature optimization.

nicolasstucki:
LGTM

PR #2399, opened 2017-05-08
Erase arguments of phantom type

nicolasstucki:
This commit removes arguments and parameter of phantom type from all applications.
Arguments are evaluated before the application in case they have side effects.

```
def foo(i: Int, p: SomePhantom) = ???

foo(42, { println("hello"); getSomePhantom })
```

becomes

```
def foo(i: Int) = ???

val x$0 = 42
val x$1 = { println("hello"); getSomePhantom }
foo(x$0)
```

Note that now `def foo(i: Int)` and def `foo(i: Int, p: SomePhantom)` erase to the same signature.

Tests for this commit where added in `Extra tests for Phantoms 1`, they where back ported as the semantics and runtime is not affected.


compiler/src/dotty/tools/dotc/core/PhantomErasure.scala
```diff
@@ -5,10 +5,13 @@ import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.core.Symbols.defn
 import dotty.tools.dotc.core.Types.Type
 
-/** Phantom erasure erases (minimal erasure):
+/** Phantom erasure erases:
  *
- *  - Parameters/arguments are erased to ErasedPhantom. The next step will remove the parameters
- *    from the method definitions and calls (implemented in branch implement-phantom-types-part-2).
+ *  - Parameters/arguments are erased removed from the function definition/call in `PhantomArgumentEval`.
```
odersky: typo: erased removed

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -400,12 +400,15 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
     case tp: MethodType =>
       def paramErasure(tpToErase: Type) =
         erasureFn(tp.isJava, semiEraseVCs, isConstructor, wildcardOK)(tpToErase)
-      val formals = tp.paramInfos.mapConserve(paramErasure)
+      val (names, formals0) =
+        if (tp.paramInfos.forall(!_.isPhantom)) (tp.paramNames, tp.paramInfos)
```
odersky: !tp.paramInfos.exists(_.isPhantom)

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -400,12 +400,15 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
...
+        else tp.paramNames.iterator.zip(tp.paramInfos.iterator).filterNot(_._2.isPhantom).toList.unzip
```
odersky: You can probably drop `.iterator`.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -474,7 +477,7 @@ object Erasure extends TypeTestsCasts{
                   .withType(defn.ArrayOf(defn.ObjectType))
                 args0 = bunchedArgs :: Nil
               }
-              val args1 = args0.zipWithConserve(mt.paramInfos)(typedExpr)
+              val args1 = args0.filterConserve(arg => !wasPhantom(arg.typeOpt)).zipWithConserve(mt.paramInfos)(typedExpr)
               untpd.cpy.Apply(tree)(fun1, args1) withType mt.resultType
```
odersky: Add a comment why we know that argument.isPhantom if and only if parameter.isPhantomj

compiler/src/dotty/tools/dotc/transform/PhantomArgumentEval.scala
```diff
@@ -0,0 +1,105 @@
...
+    } || {
+      tree.fun match {
+        case fun: Apply => hasPhantomArgs(fun)
+        case _ => false
+      }
+    }
+  }
+
+  /** Collects all args (and possibly the function) as synthetic vals and replaces them in the tree by the reference to
+   *  the lifted its val.
```
odersky: typo?

compiler/src/dotty/tools/dotc/transform/PhantomArgumentEval.scala
```diff
@@ -0,0 +1,105 @@
...
+          (arg.isInstanceOf[Ident] && !arg.symbol.is(Method) && !arg.symbol.is(Lazy))
+        )
+      }
+    case _ =>
+  }
+
+  /* Tree transform */
+
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = tree.tpe match {
+    case _: MethodType => tree // Do the transformation higher in the tree if needed
```
odersky: tree.tpe.widen

compiler/src/dotty/tools/dotc/transform/PhantomArgumentEval.scala
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+
+  /** Collects all args (and possibly the function) as synthetic vals and replaces them in the tree by the reference to
+   *  the lifted its val.
+   */
+  private def transformApplication(tree: Tree, mkTree: (Tree, List[ValDef]) => Tree)(implicit ctx: Context): Tree = tree match {
+    case tree: Apply =>
+      def mkNewApply(newFun: Tree, accSynthVals: List[ValDef]) = {
+        val (synthVals, synthValRefs) = tree.args.map(mkSynthVal).unzip
```
odersky: insert: `.filterNot(tpd.isPureExpr)`

compiler/src/dotty/tools/dotc/transform/PhantomArgumentEval.scala
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /* Tree transform */
+
+  override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = tree.tpe match {
+    case _: MethodType => tree // Do the transformation higher in the tree if needed
+    case _ =>
+      def mkNewBlock(newApply: Tree, synthVals: List[ValDef]) = Block(synthVals, newApply)
+      if (!hasPhantomArgs(tree)) tree
+      else transformApplication(tree, mkNewBlock)
```
odersky: Change logic to:

 1. If function has impure phantom args, lift out all args using `EtaExpansion.liftApp`.

nicolasstucki:
All requested changes have been taken care of.

nicolasstucki:
Rebased

PR #2401, opened 2017-05-08
Erase inlined Phantom.assume

nicolasstucki:


nicolasstucki:
@odersky, this the next part of phantom types that should be merged. It is ready for review.

nicolasstucki:
The only method that can be safely erased is `Phantom.assume`.
Given the lack of a proper effect system to prove that methods that return phantom types have no side effects and are not recursive, the current design erases the phantom types from the functions but still performs the call.
This way, if the method has a side effect it will do it. In particular for method
```scala
object MyPhantom  extends Phantom {
  // By design, this phantom universe has no value of bottom type
  type Boo <: this.Any
  def boo: Boo = assume
}

def bot: MyPhantom.Nothing = bot
``` 
 will keep its non-termination effect and no code can ever run with this fake phantom.

Currently `inline` is used for two purposes ensure that the method is not defining a phantom value via recursion, and erase completely the code. When we will have an effect system, this will become redundant.

nicolasstucki:
Functionally subsumed by #3151

PR #2405, opened 2017-05-10
Implement simple phantom functions

nicolasstucki:


nicolasstucki:
Will not be needed with #3342

PR #2407, opened 2017-05-10
Add missing scala-library dependency

smarter:
This had been missing for a while, but was only discovered now because
dotty-compiler used to depend on scala-reflect which pulled in the
correct scala-library.


smarter:
This was missed by the tests because https://github.com/lampepfl/dotty/blob/master/sbt-dotty/sbt-test/sbt-dotty/example-project/build.sbt depends on the latest nightly instead of 0.1.1-bin-SNAPSHOT

smarter:
...and now there's a new nightly published so the test fail :).

PR #2413, opened 2017-05-11
Move 'reassignment to val' to error case class

ennru:
An error message with explanation for ´reassignment to val´. It now includes the symbol's name.

@odersky `typer.Dynamic` reports this error as well, but I don't understand how a `Dynamic` can have `val`s,  the compiler reports a missing `updateDynamic` definition. I changed it anyway, though.


felixmulder:
LGTM, thanks!

PR #2414, opened 2017-05-11
Fix 2403: Tweak places where we look for new definition

odersky:
Given a sequence of nested scope (inner to outer), typedIdent has to
decide when to look for definitions. Originally his was the last
scope before the owner changes. A recent fix changed it to the first
scope after the owner changed. But that is wrong for package scopes
because for them we first have to analyze any nested imports before
considering the package scope.

This commit fixes the problem by treating package scopes specially.
Also, the ambiguous import message was improved.

I verified that squants compiles with this PR.

### DIFF
--- a/tests/neg/i1641.scala
+++ b/tests/neg/i1641.scala
@@ -2,7 +2,7 @@ package bar { object bippy extends (Double => String) { def apply(x: Double): St
 package object println { def bippy(x: Int, y: Int, z: Int) = "(Int, Int, Int)" }
 object Test {
   def main(args: Array[String]): Unit = {
-    println(bar.bippy(5.5)) // error
+    println(bar.bippy(5.5))
     println(bar.bippy(1, 2, 3)) // error
   }
 }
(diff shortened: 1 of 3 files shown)


smarter:
>  Originally his was the lawas packagst

Looks like the commit message got a bit mangled here. Also, wow, the scoping rules are more complicated than I thought!

tests/neg/i1641.scala
```diff
@@ -2,7 +2,7 @@ package bar { object bippy extends (Double => String) { def apply(x: Double): St
 package object println { def bippy(x: Int, y: Int, z: Int) = "(Int, Int, Int)" }
 object Test {
   def main(args: Array[String]): Unit = {
-    println(bar.bippy(5.5)) // error
+    println(bar.bippy(5.5))
```
smarter: Isn't it concerning that this is now different from what scalac does? Is there some more subtlely in the way definitions are looked up that we're missing?

smarter:
I think this PR could benefit from some test cases to have some concrete examples of the rules, and also to compare with what scalac does.

olafurpg:
Any update on this PR? I'd like to revert the changes in squants and scalatest to work around this change.

PR #2415, opened 2017-05-11
Remove empty tree after erasure to pass -Ycheck:all

liufengyun:
A simple fix to make sure that empty trees are removed after erasure. This avoids exception with `-Ycheck:all`.

This problem doesn't occur in the test set because we didn't enable `-Ycheck:erasure` and the empty trees are accidentally removed in the next phase.

@felixmulder Do you think it makes sense to `-Ycheck:erasure` on all tests?

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Erasure.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Erasure.scala
@@ -600,7 +600,7 @@ object Erasure extends TypeTestsCasts{
       val stats1 =
         if (takesBridges(ctx.owner)) new Bridges(ctx.owner.asClass).add(stats)
         else stats
-      super.typedStats(stats1, exprOwner)
+      super.typedStats(stats1, exprOwner).filter(!_.isEmpty)
     }
 
     override def adapt(tree: Tree, pt: Type, original: untpd.Tree)(implicit ctx: Context): Tree =
--- a/tests/pos/erased-typedef.scala
+++ b/tests/pos/erased-typedef.scala
@@ -1,0 +1,8 @@
+trait Monadless[Monad[_]] {
+
+  type M[T] = Monad[T]
+
+  def lift[T](body: T): Monad[T] =  ???
+
+  def unlift[T](m: M[T]): T = ???
+}


felixmulder:
Hmm, I'm not sure all phases are `Ycheck`able anyway, @DarkDimius ?

DarkDimius:
There are some phases that are not ycheckable(most notably late phases) but 
I don't see why erasure will not be.

felixmulder:
Ok, then LGTM :)

liufengyun:
Thanks @felixmulder @DarkDimius . I just added `-Ycheck:erasure`, let's see if the CI is happy with the change.

felixmulder:
```
 override final def toString$extension($this: ErasedValueType(Test.Bar, ErasedValueType(Test.Foo, String))): String = 
    "".+(new Test.Foo(Test.Foo.evt2u$(new Test.Bar(Test.Foo.u2evt$($this)).f())))
  override final def hashCode$extension($this: ErasedValueType(Test.Bar, ErasedValueType(Test.Foo, String))): Int = 
    Test.Foo.hashCode$extension(new Test.Bar(Test.Foo.u2evt$($this)).f())
  override final def equals$extension($this: ErasedValueType(Test.Bar, ErasedValueType(Test.Foo, String)), x$0: Object)
    : 
  Boolean = 
    {
      case val selector2: Object = x$0
      {
        def case3(): Boolean = 
          {
            def case4(): Boolean = 
              {
                def matchFail2(): Boolean = throw new MatchError(selector2)
                if selector2.isInstanceOf[Object] then 
```

Hmm, `ErasedValueType` - instinctly looks like that's not `Ycheck`-able

liufengyun:
Thanks @felixmulder.  It's good to know. I just retracted the last commit.

DarkDimius:
Why isn't it ycheckable? It does participate in subtyping, etc

felixmulder:
"Intuitively" => "Haven't looked at it too much, still trying to help"

felixmulder:
Merging this and tracking the status via: https://github.com/lampepfl/dotty/issues/2430

PR #2423, opened 2017-05-12
Fix #2368: local variables must be initialized

### LINKED ISSUE #2368: Missing error: "local variables must be initialized"
This should be an error, like it is with scalac:
```scala
object Test {
    var foo: String = _
  }
}
```
But with dotty it isn't, until it crashes in the backend.

abeln:
Disallow var declarations initialized to a placholder, if
the var declaration is local.

`var x: Foo = _` is still allowed in fields.

Testing
=======
Added tests for both invariants above.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -490,7 +490,10 @@ object messages {
            |${"var a = _"}
            |
            |Note that this use of `_` is not placeholder syntax,
-           |but an uninitialized var definition"""
+           |but an uninitialized var definition.
+           |Only fields can be left uninitialized in this manner; local variables
+           |must be initialized.
+           |"""
   }
 
   case class IllegalStartSimpleExpr(illegalToken: String)(implicit ctx: Context)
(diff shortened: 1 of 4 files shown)


DarkDimius:
As this is not dirrectly related to typing, I'd propose to move the check to RefChecks. This is a phase that implements a lot of sanity checks that could be factored out from typer.

smarter:
Honestly, I'd rather have more checks early if possible instead of doing them in RefChecks, unless we manage to move RefChecks earlier. In the IDE I currently run with `-Ystop-after:frontend` to give faster feedback to the user, if I wanted to include RefChecks I'd have to run 3 more groups of phases.

smarter:
Another benefit of doing the check in the typer is that these checks will be done when retyping (either in Erasure or Ycheck), which could catch issues.

DarkDimius:
@smarter, I think we should just have an entirely separate pipeline for IDE and run it until erasure(this pipeline won't include sbt related phases, Pickler and other phases that don't report errors). There are important errors that only get reported by erasure.

smarter:
Yes, that might be a good idea. Though for Erasure for example, I wonder if we could avoid running the whole phase and just check signatures to avoid methods with the same erased signature, etc.

abeln:
Sounds like there's consensus for leaving the check in refchecks, then? Moved it there. PTAL.

PR #2424, opened 2017-05-12
Fix #2363: better handle extractors in exhaustivity check

### LINKED ISSUE #2363: Better handling of extractors in reachability check
Dotty reports that in the following, one case is not reachable. Extractors should be better handled to avoid such annoying warnings.

```Scala
sealed trait Expr
class IntExpr extends Expr
class BooleanExpr extends Expr

object IntExpr {
  def unapply(expr: Expr): Option[IntExpr] = ???
}

object BooleanExpr {
  def unapply(expr: Expr): Option[BooleanExpr] = ???
}


class Test {
  def foo(x: List[Expr]): Int = x match {
    case IntExpr(_) :: xs => 1
    case BooleanExpr(_) :: xs => 1                              //   <==   unreachable
    case Nil => 2
  }
}
```

liufengyun:
Fix #2363: better handle extractors in exhaustivity check

1. Better handle exactors
2. Translate the pattern `List(pat1, pat2, c:_*)` to space semantically
3. Deal with extractors that return `Some[_]` or option-less type
4. Reduce redundant counterexamples in the warning

Previously we approximate extractor either to the extracted type space or Empty,
which is too coarse, and it causes some false positives about unreachable case.
Introducing the concept extractor into the theory enables us to inspect the
patterns inside extractors and check more unreachable cases, which is not
supported by Scalac.

Scalac is silent with following code, but Dotty _correctly_ reports both unexhaustive and
unreachable warnings.

```Scala
    sealed trait Node
    case class NodeA(i: Int) extends Node
    case class NodeB(b: Boolean) extends Node
    case class NodeC(s: String) extends Node

    object Node {
      def unapply(node: Node): Option[(Node, Node)] = ???
    }

    object Test {
      def foo(x: Node): Boolean = x match { // unexhaustive
        case Node(NodeA(_), NodeB(_)) => true
        case Node(NodeA(4), NodeB(false)) => true // unreachable code
      }
    }
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -2,16 +2,17 @@ package dotty.tools.dotc
 package transform
 package patmat
 
-import core.Types._
-import core.Contexts._
-import core.Flags._
+import core._
+import Types._
+import Contexts._
+import Flags._
 import ast.Trees._
 import ast.tpd
-import core.Decorators._
-import core.Symbols._
-import core.StdNames._
-import core.NameOps._
-import core.Constants._
+import Decorators._
+import Symbols._
+import StdNames._
+import NameOps._
+import Constants._
 import reporting.diagnostic.messages._
 import config.Printers.{ exhaustivity => debug }
 
@@ -329,11 +330,11 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   import SpaceEngine._
   import tpd._
 
-  private val scalaSomeClass       = ctx.requiredClass("scala.Some".toTypeName)
-  private val scalaSeqFactoryClass = ctx.requiredClass("scala.collection.generic.SeqFactory".toTypeName)
-  private val scalaListType        = ctx.requiredClassRef("scala.collection.immutable.List".toTypeName)
-  private val scalaNilType         = ctx.requiredModuleRef("scala.collection.immutable.Nil".toTermName)
-  private val scalaConsType        = ctx.requiredClassRef("scala.collection.immutable.::".toTypeName)
+  private val scalaSomeClass       = ctx.requiredClass("scala.Some")
+  private val scalaSeqFactoryClass = ctx.requiredClass("scala.collection.generic.SeqFactory")
+  private val scalaListType        = ctx.requiredClassRef("scala.collection.immutable.List")
+  private val scalaNilType         = ctx.requiredModuleRef("scala.collection.immutable.Nil")
+  private val scalaConsType        = ctx.requiredClassRef("scala.collection.immutable.::")
 
   /** Checks if it's possible to create a trait/class which is a subtype of `tp`.
    *
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -509,14 +509,14 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
           case space => List(space)
         }
       case OrType(tp1, tp2) => List(Typ(tp1, true), Typ(tp2, true))
-      case _ if tp =:= ctx.definitions.BooleanType =>
+      case tp if tp =:= ctx.definitions.BooleanType =>
         List(
           Typ(ConstantType(Constant(true)), true),
           Typ(ConstantType(Constant(false)), true)
         )
-      case _ if tp.classSymbol.is(Enum) =>
+      case tp if tp.classSymbol.is(Enum) =>
         children.map(sym => Typ(sym.termRef, true))
-      case _ =>
+      case tp =>
         val parts = children.map { sym =>
           if (sym.is(ModuleClass))
             refine(tp, sym.sourceModule.termRef)
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -323,7 +323,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   import SpaceEngine._
   import tpd._
 
-  private val scalaOptionClass     = ctx.requiredClassRef("scala.Option".toTypeName).symbol.asClass
+  private val scalaSomeClass       = ctx.requiredClassRef("scala.Some".toTypeName).symbol.asClass
   private val scalaSeqFactoryClass = ctx.requiredClass("scala.collection.generic.SeqFactory".toTypeName)
   private val scalaListType        = ctx.requiredClassRef("scala.collection.immutable.List".toTypeName)
   private val scalaNilType         = ctx.requiredModuleRef("scala.collection.immutable.Nil".toTermName)
@@ -420,7 +420,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         Kon(pat.tpe.stripAnnots, pats.map(pat => project(pat)))
       else if (fun.symbol.owner == scalaSeqFactoryClass && fun.symbol.name == nme.unapplySeq)
         projectList(pats)
-      else if (!fun.symbol.info.finalResultType.isRef(scalaOptionClass))
+      else if (fun.symbol.info.finalResultType.isRef(scalaSomeClass))
         Kon(pat.tpe.stripAnnots, pats.map(pat => project(pat)))
       else
         Fun(pat.tpe.stripAnnots, fun.tpe, pats.map(pat => project(pat)))
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -417,14 +417,13 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     case Bind(_, pat) => project(pat)
     case UnApply(fun, _, pats) =>
       if (pat.tpe.classSymbol.is(CaseClass))
-        // FIXME: why dealias is needed here?
-        Kon(pat.tpe.stripAnnots.dealias, pats.map(pat => project(pat)))
+        Kon(pat.tpe.stripAnnots, pats.map(pat => project(pat)))
       else if (fun.symbol.owner == scalaSeqFactoryClass && fun.symbol.name == nme.unapplySeq)
         projectList(pats)
       else if (fun.symbol.info.resultType.isRef(scalaSomeClass))
-        Kon(pat.tpe.stripAnnots.dealias, pats.map(pat => project(pat)))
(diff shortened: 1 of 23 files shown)


compiler/src/dotty/tools/dotc/reporting/UniqueMessagePositions.scala
```diff
@@ -23,7 +23,7 @@ trait UniqueMessagePositions extends Reporter {
         for (pos <- m.pos.start to m.pos.end) {
           positions get (ctx.source, pos) match {
             case Some(level) if level >= m.level => shouldHide = true
-            case _ => positions((ctx.source, pos)) = m.level
+            case _ => // positions((ctx.source, pos)) = m.level
```
liufengyun: The CI fails because the change of this line. There's some bug hidden in `isHidden`, I have to comment it out to temporarily make sure all the exhaustivity/unreachability warnings are shown. Otherwise, the unreachable warning in the PR example will be incorrectly hidden.

smarter:
@liufengyun Not directly related to this PR, but while you're working on this code, could you have a look at the exhaustiveness warnings reported against dotty itself when bootstrapping? See http://dotty-ci.epfl.ch/lampepfl/dotty/2267/4 for example. You can reproduce that locally by running `dotty-compiler-bootstrapped/compile`. It would be good to see which of these warnings are real warnings we need to fix and which one are bugs in the exhaustiveness checker.

liufengyun:
@smarter Sure, I'll examine log to see if there're still false positives.

liufengyun:
Now all false positives about unreachability/exhaustiveness are fixed. True positives will be addressed in https://github.com/lampepfl/dotty/issues/2429 separately.

Could you please review @smarter ?

liufengyun:
@felixmulder The last commit 31ae57b contains some changes to the doc-tool, could you please confirm if the changes make sense?

doc-tool/src/dotty/tools/dottydoc/util/MemberLookup.scala
```diff
@@ -77,6 +77,7 @@ trait MemberLookup {
       case (x :: xs, _) =>
         if (xs.nonEmpty) globalLookup
         else lookup(entity, packages, "scala." + query)
+      case (Nil, _) => ??? // impossible as long as query is not empty
```
felixmulder: let's do something better than `???`, otherwise LGTM

doc-tool/src/dotty/tools/dottydoc/model/comment/BodyEntities.scala
```diff
@@ -64,7 +64,7 @@ final case class Text(text: String) extends Inline
 abstract class EntityLink(val title: Inline) extends Inline { def link: LinkTo }
 object EntityLink {
   def apply(title: Inline, linkTo: LinkTo) = new EntityLink(title) { def link: LinkTo = linkTo }
-  def unapply(el: EntityLink): Option[(Inline, LinkTo)] = Some((el.title, el.link))
+  def unapply(el: EntityLink): Some[(Inline, LinkTo)] = Some((el.title, el.link))
```
felixmulder: Why is this needed? I would prefer not to change the type here - unapplys should be `Option`...

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -426,14 +509,14 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
           case space => List(space)
         }
       case OrType(tp1, tp2) => List(Typ(tp1, true), Typ(tp2, true))
-      case _ if tp =:= ctx.definitions.BooleanType =>
+      case tp if tp =:= ctx.definitions.BooleanType =>
```
smarter: `tp.isRef(defn.BooleanClass)`

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -549,6 +632,8 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
         if (ctx.definitions.isTupleType(tp))
           signature(tp).map(_ => "_").mkString("(", ", ", ")")
+        else if (sym.showFullName == "scala.collection.immutable.List")
```
smarter: We should never do string comparisons to determine the type of something. Instead something like `sym == defn.ListClass` should be used, if `ListClass` is not defined in `Definitions.scala` then it should be added.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -664,31 +751,35 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      // in redundancy check, take guard as false in order to soundly approximate
       val prevs = cases.take(i).map { x =>
-        if (x.guard.isEmpty) project(x.pat, false)
+        if (x.guard.isEmpty) project(x.pat)
         else Empty
       }.reduce((a, b) => Or(List(a, b)))
 
       val curr = project(cases(i).pat)
 
+      debug.println(s"---------------reachable? ${show(curr)}")
```
smarter: I suggest adding a `Printer` for the exhaustiveness checker in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/config/Printers.scala instead of using `debug` here.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -98,19 +102,31 @@ trait SpaceLogic {
   def show(sp: Space): String
 
   /** Simplify space using the laws, there's no nested union after simplify */
-  def simplify(space: Space): Space = space match {
+  def simplify(space: Space, aggressive: Boolean = false): Space = space match {
```
smarter: Could you update the documentation of this method to explain what the `aggressive` parameter is for?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -137,26 +153,33 @@ trait SpaceLogic {
...
       case (Empty, _) => true
       case (_, Empty) => false
-      case (Or(ss), _) => ss.forall(isSubspace(_, b))
+      case (Or(ss), _) =>
+        ss.forall(isSubspace(_, b))
       case (Typ(tp1, _), Typ(tp2, _)) =>
-        isSubType(tp1, tp2) || tryDecompose1(tp1) || tryDecompose2(tp2)
-      case (Typ(tp1, _), Or(ss)) =>
+        isSubType(tp1, tp2)
+      case (Typ(tp1, _), Or(ss)) =>  // optimization
```
smarter: Maybe improve this comment to explain what this is an optimization of?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -386,7 +469,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   /** Is `tp1` a subtype of `tp2`?  */
   def isSubType(tp1: Type, tp2: Type): Boolean = {
     // check SI-9657 and tests/patmat/gadt.scala
-    val res = erase(tp1) <:< erase(tp2)
+    val res = tp1 <:< erase(tp2)
```
smarter: Could you add a comment explaining why it's correct to erase `tp2` but not `tp1` ?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -420,7 +420,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         Kon(pat.tpe.stripAnnots, pats.map(pat => project(pat)))
       else if (fun.symbol.owner == scalaSeqFactoryClass && fun.symbol.name == nme.unapplySeq)
         projectList(pats)
-      else if (fun.symbol.info.resultType.isRef(scalaSomeClass))
+      else if (!fun.symbol.info.finalResultType.isRef(scalaOptionClass))
```
smarter: I don't think this is correct in general, the return type of an option-less patmat may define a `def isEmpty: Boolean` method, if this method return `false`, then the case will not be taken.

smarter: Sorry, if `isEmpty` returns `true`, not `false`

liufengyun: Ah, I see, then it means we cannot tell from types whether an option-less is exhaustive or not. I'll retract the last commit.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: The `debug` is just a rename of the exhaustivity printer :)

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
smarter: Well, if the method is defined as `def isEmpty: false` then we know it will always return false :). But we should probably not encode this logic in the exhaustiveness checker until we have a real spec for option-less pattern-matching that explains all cases.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -469,6 +475,19 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   /** Is `tp1` a subtype of `tp2`?  */
   def isSubType(tp1: Type, tp2: Type): Boolean = {
     // check SI-9657 and tests/patmat/gadt.scala
+    //
+    // `erase` is a walkaround to make the following code pass the check:
```
smarter: typo: walkaround -> workaround

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -323,11 +329,11 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   import SpaceEngine._
   import tpd._
 
-  private val scalaSomeClass       = ctx.requiredClassRef("scala.Some".toTypeName).symbol.asClass
+  private val scalaSomeClass       = ctx.requiredClass("scala.Some".toTypeName)
   private val scalaSeqFactoryClass = ctx.requiredClass("scala.collection.generic.SeqFactory".toTypeName)
   private val scalaListType        = ctx.requiredClassRef("scala.collection.immutable.List".toTypeName)
   private val scalaNilType         = ctx.requiredModuleRef("scala.collection.immutable.Nil".toTermName)
-  private val scalaConType         = ctx.requiredClassRef("scala.collection.immutable.::".toTypeName)
+  private val scalaConsType        = ctx.requiredClassRef("scala.collection.immutable.::".toTypeName)
```
smarter: You could get rid of all the `.toTypeName` here if `Decorators.PreNamedString` is imported.

PR #2425, opened 2017-05-13
Fix caches of info that depends on inherited classes

odersky:
Change from the ground up the way we treat inherited information. We get rid of:

 - superId structures, 
 - fingerprints
 - Frozen and FullyCompleted flags, 
 - restrictions on when we can enter new symbols in scopes
 - most manual cache invalidation

We treat systematically the problem of cache dependencies by tracking reverse dependencies with weak hashmaps. We need cooperation from phases to declare what kind of inherited info they can change. 

There are two new kinds of caches: 

 - MemberNames  contains various kinds of member name sets indexed by name filter
 - BaseData contains baseClasses (i.e. base class linearization), and BaseClassSet, which
    is a sorted-array-backed set for fast subclass checking. BaseClassSets replace the 
    superId maps we had before.

The PR fixes several problems with inherited info. This manifested itself in the past in a fingerprint problem when function specialization is turned on. But fingerprints are just where the problem arises first; it is also possible to hit problems with the other caches.


compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1992,41 +1918,173 @@ object SymDenotations {
...
+  private class BaseDataImpl(createdAt: Period) extends InheritedCacheImpl(createdAt) with BaseData {
+    private[this] var cache: (List[ClassSymbol], BaseClassSet) = null
+
+    private[this] var valid = true
+    private[this] var locked = false
+    private[this] var provisional = false
+
+    final def isValid(implicit ctx: Context): Boolean = valid && isValidAt(ctx.phase)
+
+    def invalidate(): Unit =
```
smarter: Missing an `(implicit ctx: Context)`

DarkDimius:
@smarter, I'm reviewing this PR now and there are things that may affect performance. I'd propose to benchmark after review, but before merging.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1209,6 +1230,68 @@ object SymDenotations {
...
+          (myBaseTypeRefCachePeriod.runId != ctx.runId ||
+           ctx.phases(myBaseTypeRefCachePeriod.phaseId).parentsGroup != ctx.phase.parentsGroup)) {
+        myBaseTypeRefCache = new BaseTypeRefMap
+        myBaseTypeRefCachePeriod = ctx.period
+      }
+      myBaseTypeRefCache
+    }
+
+    private def invalidateBaseDataCache() = {
+      baseDataCache.invalidate()
```
DarkDimius: why do you need both call to `invalidate` and assignment to an empty cache?

In copyCaches you copy over the refernce from previous cache. It means that invalidating a future cache will also invalidate an old cache, which is still correct.

Same in other invalidation methods.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1167,15 +1168,35 @@ object SymDenotations {
       privateWithin: Symbol = null,
       annotations: List[Annotation] = null)(implicit ctx: Context) =
     { // simulate default parameters, while also passing implicit context ctx to the default values
-      val initFlags1 = (if (initFlags != UndefinedFlags) initFlags else this.flags) &~ Frozen
+      val initFlags1 = (if (initFlags != UndefinedFlags) initFlags else this.flags)
       val info1 = if (info != null) info else this.info
+      if (ctx.isAfterTyper && changedClassParents(info, info1, completersMatter = false))
+        assert(ctx.phase.changesParents, i"undeclared parent change at ${ctx.phase} for $this, was: $info, now: $info1")
```
odersky: The check is in `enteredAfter`. I think it would make little sense checking in `info_=` because typically we set the info with a clone of the previous scope and then add symbols later using `enteredAfter`.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1209,6 +1230,68 @@ object SymDenotations {
...
+    /** A cache from types T to baseTypeRef(T, C) */
+    type BaseTypeRefMap = java.util.HashMap[CachedType, Type]
+    private[this] var myBaseTypeRefCache: BaseTypeRefMap = null
+    private[this] var myBaseTypeRefCachePeriod: Period = Nowhere
+
+    private var baseDataCache: BaseData = BaseData.None
+    private var memberNamesCache: MemberNames = MemberNames.None
+
+    private def memberCache(implicit ctx: Context): LRUCache[Name, PreDenotation] = {
+      if (myMemberCachePeriod != ctx.period) {
```
odersky: But `myMemberCachePeriod` is always set to a single phase period, right?

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1249,8 +1329,11 @@ object SymDenotations {
     }
 
     override protected[dotc] final def info_=(tp: Type) = {
-      super.info_=(tp)
+      if (changedClassParents(infoOrCompleter, tp, completersMatter = true))
+        invalidateBaseDataCache()
+      invalidateMemberNamesCache()
```
odersky: member names are relatively infrequently needed, so I think we lose little to invalidate them more aggressively than strictly needed. Also, names will usually be entered later using `enteredAfter`.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
odersky: Good point. It was originally to save space, but I don't think that matters in practice.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1209,6 +1230,68 @@ object SymDenotations {
...
+
+    override def copyCaches(from: SymDenotation, phase: Phase)(implicit ctx: Context): this.type = {
+      from match {
+        case from: ClassDenotation =>
+          if (from.memberNamesCache.isValidAt(phase)) memberNamesCache = from.memberNamesCache
+          if (from.baseDataCache.isValidAt(phase)) {
+            baseDataCache = from.baseDataCache
+            myBaseTypeRefCache = from.baseTypeRefCache
+          }
+        case _ =>
```
odersky: Could happen if original info is a completer.

compiler/src/dotty/tools/dotc/core/Phases.scala
```diff
@@ -331,6 +339,8 @@ object Phases {
     final def refChecked = myRefChecked     // Phase is after RefChecks
     final def symbolicRefs = mySymbolicRefs // Phase is after ResolveSuper, newly generated TermRefs should be symbolic
     final def labelsReordered = myLabelsReordered // Phase is after LabelDefs, labels are flattened and owner chains don't mirror this
+    final def membersGroup = myMembersGroup // group id for phases where all symbols have the same non-private members
```
odersky: Good point. I'll change to phase ids, and use a more descriptive name.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
odersky: In fact no. It does matter. IsValid at only checks that the creation period is compatible with the current one. Invalidation does not affect the creation period. So we need another way to signal that the cache is nowhere valid. That's what the assignment does.

PR #2432, opened 2017-05-15
Fix #2426: Use Scala-2 syntax for annotations of class constructors

### LINKED ISSUE #2426: Impossible to annotate public primary constructor and keep Scala 2 compatibility
This doesn't compile with dotty:
```scala
class Foo @deprecated("foo", "2.11.0") (x: Int)
```
```scala
1 |class Foo @deprecated("foo", "2.11.0") (x: Int)
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |    private, protected, or this expected for annotated primary constructor
```

Adding `this` does indeed fix it, but doesn't compile with Scala 2:
```scala
class Foo @deprecated("foo", "2.11.0") this(x: Int)
```

- Is this an intentional difference?
- If yes, we should at least support the old syntax under `-language:Scala2`

odersky:
Use Scala-2 compatible syntax for annotations of primary class constructors.
In fact, we can drop Scala 2's restriction that such annotations may only
have one parameter list.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -294,6 +294,41 @@ object Parsers {
       } finally inFunReturnType = saved
     }
 
+    /** A placeholder for dummy arguments that should be re-parsed as parameters */
+    val ParamNotArg = EmptyTree
+
+    /** A flag indicating we are parsing in the annotations of a primary
+     *  class constructor
+     */
+    private var inClassConstrAnnots = false
+
+    private def fromWithinClassConstr[T](body: => T): T = {
+      val saved = inClassConstrAnnots
+      try {
+        inClassConstrAnnots = true
+        body
+      } finally {
+        inClassConstrAnnots = saved
+        if (lookaheadTokens.nonEmpty) {
+          in.insertTokens(lookaheadTokens.toList)
+          lookaheadTokens.clear()
+        }
+      }
+    }
+
+    /** Lookahead tokens for the case of annotations in class constructors.
+     *  We store tokens in lookahead as long as they can form a valid prefix
+     *  of a class parameter clause.
+     */
+    private var lookaheadTokens = new ListBuffer[TokenData]
+
+    /** Copy current token to end of lookahead */
+    private def saveLookahead() = {
+      val lookahead = new TokenData{}
+      lookahead.copyFrom(in)
+      lookaheadTokens += lookahead
+    }
+
     def migrationWarningOrError(msg: String, offset: Int = in.offset) =
       if (in.isScala2Mode)
         ctx.migrationWarning(msg, source atPos Position(offset))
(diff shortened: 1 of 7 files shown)


odersky:
This change caused a lot of complexity. More than 100 loc to deal with this issue.

compiler/src/dotty/tools/dotc/parsing/Scanners.scala
```diff
@@ -217,11 +216,42 @@ object Scanners {
...
+
+    /** If following is empty, invalidate token data `td` by setting 
+     *  `td.token` to `EMPTY`. Otherwise pop head of `following` into `td`.
+     */
+    private def popCopy(td: TokenData) = {
+      if (following.isEmpty) td.token = EMPTY
+      else {
+        td.copyFrom(following.head)
+        following = following.tail
+      }
```
felixmulder: Looks like a missing closing brace here.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2094,21 +2194,15 @@ object Parsers {
...
 
-    /** ConstrMods        ::=  AccessModifier
-     *                      |  Annotation {Annotation} (AccessModifier | `this')
+    /** ConstrMods        ::=  {Annotation} [AccessModifier]
      */
-    def constrModsOpt(owner: Name): Modifiers = {
-      val mods = modifiers(accessModifierTokens, annotsAsMods())
-      if (mods.hasAnnotations && !mods.hasFlags)
-        if (in.token == THIS) in.nextToken()
-        else syntaxError(AnnotatedPrimaryConstructorRequiresModifierOrThis(owner), mods.annotations.last.pos)
```
smarter: This means we can drop this `Message` class (but we must be careful to leave a dummy id in its place in the `ErrorMessageID` enum to avoid changing the error number of every subsequent message).

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1280,20 +1315,77 @@ object Parsers {
...
+        else exprInParens()
+      arg match {
+        case arg @ Assign(Ident(id), rhs) => cpy.NamedArg(arg)(id, rhs)
+        case arg => arg
+      }
+    }
+
+    /** Handle first argument of an argument list to an annotation of
+     *  a primary class constructor. If the current token either cannot
+     *  start an expression or is an identifier and is followed by `:`,
```
smarter: But a `:` can legitimately appear in an argument list, e.g:
```scala
class ann(x: Any) extends scala.annotation.Annotation

object Test {
  val elem: Int = 1
  class Foo @ann(elem: Float) (x: String)
}
```

Banning multiple param lists in constructor annotations still seems better to me than having complex rules to guess whether something is or isn't an argument list :).

odersky: I don't see how banning multi param solves that problem?

smarter: Indeed, it seems that scalac always expects a primary constructor annotation to have exactly one parameter list, which is a more annoying restriction. I'm still not super enthusiastic about adding so much special-casing in the parser (this will need to be replicated in scala.meta for example, /cc @xeno-by), but I'm not opposed to it. The main annoyance is that if seemingly valid code is misinterpreted, you'll get a very confusing error message, but that could be left as a future improvement.

odersky:
I tweaked it now so that the first `()` is treated as an annotation argument, not an empty 
parameter list. That way we are backwards compatible with Scala-2.

PR #2435, opened 2017-05-15
Handle case where typevar instance is fully determined

odersky:
It's possible that a type parameter in a constraint gets instantiated via
a unification operation. In that case, the type parameter does not have
a TypeBounds instance associated with it, so is not considered to be
contained in the constraint. But it must still be possible to instantiate
a type variable corresponding to this parameter.

Test case in collection strawman: `List.empty ++ List.empty`.


felixmulder:
@biboudis - add minimized test then merge :)

PR #2436, opened 2017-05-15
Fix broken link in the navigation of docs and add minor adjustments

biboudis:


docs/docs/contributing/getting-started.md
```diff
@@ -62,3 +64,7 @@ or via bash:
 ```bash
 $ dotr
 ```
+
+Before contributing in Dotty, we invite you to consult the
```
smarter: contributing _to_ Dotty

PR #2438, opened 2017-05-16
Enable value class with phantom

nicolasstucki:
This allows the implementation of extension methods on
phantoms without the cost of instanciatiation.


compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -187,7 +187,8 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
      */
     def valueHashCodeBody(implicit ctx: Context): Tree = {
       assert(accessors.length == 1)
-      ref(accessors.head).select(nme.hashCode_).ensureApplied
+      if (accessors.head.info.isPhantom) Literal(Constant(0))
```
OlivierBlanvillain: I wonder if this (and `equals`) shouldn't throw an exception instead

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -159,7 +159,7 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
       val thatAsClazz = ctx.newSymbol(ctx.owner, nme.x_0, Synthetic, clazzType, coord = ctx.owner.pos) // x$0
       def wildcardAscription(tp: Type) = Typed(Underscore(tp), TypeTree(tp))
       val pattern = Bind(thatAsClazz, wildcardAscription(clazzType)) // x$0 @ (_: C)
-      val comparisons = accessors map (accessor =>
+      val comparisons = accessors.filter(!_.info.isPhantom).map(accessor =>
```
OlivierBlanvillain: Maybe update the scaladoc to say that equals is also synthesized for `AnyVals`

nicolasstucki: The doc already mentions value classes.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
nicolasstucki: I wonder if we should actually disallow them. Still this code should not thow for case classes with phantoms.

PR #2439, opened 2017-05-16
Give to @odersky.

DarkDimius:
Ping @odersky.
Reproduction of inference problem:

```
sbt "run -optimise -Ycheck:tailrec,resolveSuper,mixin,restoreScopes,labelDef,simplify tests/pos/t7126.scala"
```

Currently it infers a Nothing as a CC[T] to be Nothing[T], and isn't able to figure out that

  Nothing[Any] =:=  Nothing

We've had a discussion that it should isntead infer CC[T] to be type lambda T => Nothing to be kind-correct.


odersky:
@DarkDimius @OlivierBlanvillain I verified that the original problem is fixed with the latest commit.

odersky:
I was trying first to avoid a kind-polymorphic Nothing but ran into several other problems. See 	9b022ab in branch wip-hk-bottom on staging. So for now this is a quick fix of the problem you were seeing.

DarkDimius:
@odersky, thanks for the fix!

PR #2440, opened 2017-05-16
Add two utility methods ensureOwner and owners on tpd.Tree

liufengyun:
Add two utility methods ensureOwner and owners on tpd.Tree

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/tpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/tpd.scala
@@ -16,6 +16,7 @@ import NameKinds.TempResultName
 
 import scala.annotation.tailrec
 import scala.io.Codec
+import scala.collection.mutable
 
 /** Some creators for typed trees */
 object tpd extends Trees.Instance[Type] with TypedTreeInfo {
@@ -603,6 +604,27 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
 
     def subst(from: List[Symbol], to: List[Symbol])(implicit ctx: Context): ThisTree =
       new TreeTypeMap(substFrom = from, substTo = to).apply(tree)
+
+    /** Ensure that top-level definitions of the current tree have the given owner */
+    def ensureOwner(owner: Symbol)(implicit ctx: Context): tpd.Tree = {
+      tree.owners.foldRight(tree) { (from, acc) =>
+        if (from eq owner) acc
+        else acc.changeOwner(from, owner)
+      }
+    }
+
+    /** Get the owners of the top-level definitions of the current tree */
+    def owners(implicit ctx: Context): List[Symbol] = {
+      val owners = mutable.Set.empty[Symbol]
+      new TreeTraverser {
+        def traverse(tree: tpd.Tree)(implicit ctx: Context): Unit = tree match {
+          case tree: tpd.DefTree => owners += tree.symbol.owner
+          case _                 => traverseChildren(tree)
+        }
+      }.traverse(tree)
+
+      owners.toList
+    }
 
     /** Change owner from `from` to `to`. If `from` is a weak owner, also change its
      *  owner to `to`, and continue until a non-weak owner is reached.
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -605,15 +605,23 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     def subst(from: List[Symbol], to: List[Symbol])(implicit ctx: Context): ThisTree =
       new TreeTypeMap(substFrom = from, substTo = to).apply(tree)
 
-    /** Ensure that top-level definitions of the current tree have the given owner */
+    /** Ensure that top-level definitions of the current tree have the given owner
+     *
+     *  @note Dependeing on the shape of the tree, this operation can be expensive.
+     *        Always prefer using `changeOwner` directly.
+     */
     def ensureOwner(owner: Symbol)(implicit ctx: Context): tpd.Tree = {
```
odersky: In fact, there's a more efficient way to do it my using TreeTypeMap directly. This can replace multiple owners in one traversal. Then you won't need the @note anymore.

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -605,15 +605,23 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     def subst(from: List[Symbol], to: List[Symbol])(implicit ctx: Context): ThisTree =
       new TreeTypeMap(substFrom = from, substTo = to).apply(tree)
 
-    /** Ensure that top-level definitions of the current tree have the given owner */
+    /** Ensure that top-level definitions of the current tree have the given owner
+     *
+     *  @note Dependeing on the shape of the tree, this operation can be expensive.
```
odersky: Typo: depending

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -605,15 +605,23 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
     }
 
-    /** Get the owners of the top-level definitions of the current tree */
+    /** Get the owners of the top-level definitions of the current tree
+     *
+     *  @note Dependeing on the shape of the tree, this operation can be expensive.
+     *        Always prefer using the known owner from the context.
+     */
     def owners(implicit ctx: Context): List[Symbol] = {
       val owners = mutable.Set.empty[Symbol]
```
odersky: Here, it's more efficient to use an immutable set.

smarter: Interesting, could you explain why it's more efficient? Would you write it as `var owners = immutable.Set.empty` ?

DarkDimius:
What is your use-case? so far all cases that needed to handle owners have used TTM directly.
As a general comment: it would be nice not to introduce unused utility method.

liufengyun:
@DarkDimius I use the two methods in the macros to automatically update owners when constructing typed definition trees and typed splices -- meta-programmers don't have to take care of them.

I agree these two utility methods don't need to be in the compiler if they are potentially problematic. I thought they might be useful for the compiler, and also to check if the hack makes sense for macros.

liufengyun:
> Also consider if you want your definition of top-level trees to handle Thicket and\or Blocks.
Also, what should happen if your top-level tree is a package\class?

The `top-level` used in the code is not top-level in the standard sense, it actually means `non-nested definitions` inside the tree.

> In your implementation order of changeOwner calls behaves differently from the one that used in TTM.

Could you please elaborate more about this?

DarkDimius:
>> In your implementation order of changeOwner calls behaves differently from the one that used in TTM.

>Could you please elaborate more about this?

Every symbol substitution create new symbol substitutions like an avalanche(see implementation in TTM that accumulates them) this will be very tricky to debug once it happens. For example,
if you substitute (a -> b), (b -> c), (c -> d) one-by-one, you succeed. If you do it all at once, you'll have infinite cycle. 

___
@liufengyun, I'd propose to not implement those methods inside compiler, and instead implement it inside your thin API:
 - they don't have use-cases yet in compiler, they won't be tested and may misbehave after minor changes to dotty internals;
 - if I got you right, you want this method to become public external API. We didn't decide on it yet, but It's better to have it clearly defined instead of making "some parts of tpd" be a public api;
 - those methods don't have clear names and even documentation for them uses terms that used inside compiler with a different meaning. In the current state it is very misleading as neither the name nor the documentation correctly explains what this method does in Dotty terminology.

liufengyun:
Thanks a lot @DarkDimius , it sounds reasonable to avoid them in the compiler.

PR #2442, opened 2017-05-16
Disallow phantom type parameter/member in classes

nicolasstucki:
This fixes potential ways to fake evidences by casting an object that has a phantom type parameter/member. See `neg/phantom-class-type-parameters.scala` as an example.


compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -506,8 +506,16 @@ trait TypeAssigner {
...
+  def assignType(tree: untpd.TypeDef, sym: Symbol)(implicit ctx: Context) = {
+    val tdef = tree.withType(symbolicIfNeeded(sym).orElse(sym.typeRef))
+    val isParam = tdef.tpe.typeSymbol.is(Param)
+    val isPhantom = tdef.tpe.isPhantom
+    if (isParam && isPhantom && tdef.symbol.owner.isPrimaryConstructor)
+      ctx.error("Classes cannot have phantom type parameters", tdef.pos)
+    else if (!isParam && isPhantom && !tdef.symbol.owner.isStaticOwner)
+      ctx.error("Non static classes cannot have phantom type members", tdef.pos)
+    tdef
+  }
```
liufengyun: A small comment: may this be done in RefChecks where a lot of checking happens?

nicolasstucki: I will try. I also felt uneasy adding all those checks in there.

nicolasstucki: Moved it to Checking

nicolasstucki:
Found a cleaner solution with #2525

PR #2443, opened 2017-05-16
Move 'Type does not take parameters' to error case class

ennru:
@felixmulder Another tiny step to explanations on compiler errors.


felixmulder:
Thanks Enno! :)

PR #2446, opened 2017-05-16
<notype> pops up after pickling with tpd changes

OlivierBlanvillain:
Another issue caused by keeping the types coming out of typer: it seems that a `<notype>` pops up after pickling. I couldn't reproduce the bug with something else than a pickling test (I guess something later in the pipeline is healing that type?), here is a minimised test along with the changes in `tpd.scala` to reproduce the issue.

```
run -classpath library/target/scala-2.11/dotty-library_2.11-0.1.1-bin-SNAPSHOT-nonbootstrapped.jar:compiler/target/scala-2.11/dotty-compiler_2.11-0.1.1-bin-SNAPSHOT-nonbootstrapped.jar:interfaces/target/dotty-interfaces-0.1.1-bin-SNAPSHOT.jar -Ytest-pickler test.scala
```

@odersky Any idea where I should look?

### DIFF
--- a/after-pickling.txt
+++ b/after-pickling.txt
@@ -1,0 +1,1 @@
+[cannot display due to class dotty.tools.dotc.reporting.diagnostic.messages$Error at ?: value of type <notype> does not take parameters, raw string = dotty.tools.dotc.printing.Formatting$StringFormatter@6e950bcf]
--- a/before-pickling.txt
+++ b/before-pickling.txt
@@ -1,0 +1,14 @@
+package <empty>.type {
+  import dotty.tools.dotc.core.Decorators._
+  import dotty.tools.dotc.core.NameOps._
+  final lazy module val Test: Test = new Test()
+  @scala.annotation.internal.SourceFile("test.scala") final module class Test()
+     extends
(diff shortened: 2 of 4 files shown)


odersky:
To see where the <NoType> comes up, change an error handler in Formatting.scala:

    class StringFormatter(protected val sc: StringContext) {

    protected def showArg(arg: Any)(implicit ctx: Context): String = arg match {
      case arg: Showable =>
        try arg.show
        catch {
          case NonFatal(ex) if false => s"[cannot display due to $ex, raw string = $toString]"
        }
      case _ => arg.toString
    }

(The `if false` needs to be added). Then you see the full error message. It would be good to make this friendlier, but it's tricky, because we also have situations where the error printing itself causes an exception, which we do not want to report in this case.

DarkDimius:
@odersky the issue comes up if TermRefWithSignature created by typer survive until pickling.
Those types have signature set which is different from signature of underlying denotation. 
I'll make a pr that showcases this issue.

felixmulder:
@OlivierBlanvillain - please turn this into an issue <3

PR #2447, opened 2017-05-16
Move 'var/val parameter may not be call-by-name' to error case class

ennru:
@felixmulder One more and I call it a day.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1342,11 +1342,11 @@ object messages {
     val msg = s"${if (mutable) "`var'" else "`val'"} parameters may not be call-by-name"
     val kind = "Syntax"
     val explanation =
-      hl"""Parameters of classes and traits may no be call-by-name. In case you
-          |want the parameter to be evaluated on demand, consider using a function
-          |and a lazy value in the class such as
-          |  ${s"class MyClass(${name}Tick: () => String) {"}
-          |  ${s"  lazy val $name = ${name}Tick"}
+      hl"""${"var"} and ${"val"} parameters of classes and traits may no be call-by-name. In case you
+          |want the parameter to be evaluated on demand, consider making it just a parameter
+          |and a ${"def"} in the class such as
+          |  ${s"class MyClass(${name}Tick: => String) {"}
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1336,4 +1336,18 @@ object messages {
           |out the parameter list when extending it.
           |"""
   }
+
+  case class VarValParametersMayNotBeCallByName(name: Names.TermName, mutable: Boolean)(implicit ctx: Context)
+    extends Message(VarValParametersMayNotBeCallByNameID) {
+    val msg = s"${if (mutable) "`var'" else "`val'"} parameters may not be call-by-name"
+    val kind = "Syntax"
+    val explanation =
+      hl"""Parameters of classes and traits may no be call-by-name. In case you
```
smarter: They can be as long as they don't have a getter e.g, this compiles:
```scala
class Foo(x: => Int)
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1336,4 +1336,18 @@ object messages {
...
+          |want the parameter to be evaluated on demand, consider using a function
+          |and a lazy value in the class such as
+          |  ${s"class MyClass(${name}Tick: () => String) {"}
+          |  ${s"  lazy val $name = ${name}Tick"}
```
smarter: But that's different from what a by-name would do, the closest thing would be to use a `def`, not a `lazy val`:
```class MyClass(fooFun: () => Int)) {
  def foo = fooFun()
```

smarter: Or even:
```scala
class MyClass(myFoo: => Int) {
  def foo() = myFoo
}
```

ennru: Ok, that's better. Will update.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: Yes, will be more specific.

PR #2449, opened 2017-05-17
Use ErasedPhantom instead of BoxedUnit as erased unit type.

nicolasstucki:


library/src/dotty/runtime/ErasedPhantom.java
```diff
@@ -0,0 +1,30 @@
...
+    public boolean equals(java.lang.Object other) {
+        return this == other;
+    }
+
+    public int hashCode() {
+        return 0;
+    }
+
+    public String toString() {
+        return "(\uD83D\uDC7B )";
```
OlivierBlanvillain: Extra space after 👻?

compiler/src/dotty/tools/dotc/transform/Memoize.scala
```diff
@@ -99,17 +99,19 @@ import Decorators._
...
 
     def erasedBottomTree(sym: Symbol) = {
       if (sym eq defn.NothingClass) Throw(Literal(Constant(null)))
       else if (sym eq defn.NullClass) Literal(Constant(null))
+      else if (sym eq defn.BoxedUnitClass) ref(defn.BoxedUnit_UNIT)
+      else if (sym eq defn.ErasedPhantomClass) ref(defn.ErasedPhantom_UNIT)
       else {
-        assert(sym eq defn.BoxedUnitClass)
-        ref(defn.BoxedUnit_UNIT)
+        assert(false) // Not in sync with isErasableBottomField
```
OlivierBlanvillain: `assert(false, "Check that isErasableBottomField before calling erasedBottomTree")`?

library/src/dotty/runtime/ErasedPhantom.java
nicolasstucki: The 👻 defines is printed on two characters but the next character is at 1 space of distance. At least on my terminal.

PR #2457, opened 2017-05-17
Make idempotency check a proper vulpix test

nicolasstucki:


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -346,13 +290,9 @@ class CompilationTests extends ParallelTesting {
     assert(new java.io.File("../out/idempotency1/").exists)
     assert(new java.io.File("../out/idempotency2/").exists)
 
-    val t0 = System.currentTimeMillis()
-    checkIdempotency()
-    println(s"checked bytecode idempotency (${(System.currentTimeMillis() - t0) / 1000.0} sec)")
+    compileFile("../tests/idempotency/IdempotencyCheck.scala", defaultOptions).checkRuns()
```
smarter: Why is this better?

felixmulder: Because of reasons

felixmulder: It improves the feedback from error messages for the venerable Stucki :)

nicolasstucki: Error messages and timing is handled by vupix. If there is a crash, vulpix will report it and continue running the test suite.

smarter: But why would there be a crash in the testcode?

PR #2458, opened 2017-05-17
Fix 2396b: Non-static synthetic case-objects should be initialized early.

DarkDimius:
Lazy val implements non-static objects.
But if object is not user-defined we don't make it `real` lazy,
we instead make it eager as no-one will notice. This actually
saves resources as otherwise you'll still need to initialize
lazy-val related stuff in constructor instead and later initialize
actual objects.

Those objests, being eager, should be initialized early,
so that constructors can't see un-initialized values.


DarkDimius:
@nicolasstucki please review.

PR #2459, opened 2017-05-17
Fix #2396: don't create refernces to packages in ElimStaticThis.

### LINKED ISSUE #2396: Bad bytecode generated for a partial function.
When running the example below, it fails with `java.lang.VerifyError` [Full stack trace](https://github.com/lampepfl/dotty/files/983955/stacktrace.txt)
```scala
class Main {
  val buzz = Some(Bees.Bee("buzz")).collect {
    case Bees.Bee(value) => value
  }
}

object Main {
    new Main
  }
}

object Bees {
  case class Bee(value: String)
}
```

Reproducible in fresh `lampepfl/dotty.g8`
OpenJDK 8 on Ubuntu
```
$ java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13)
OpenJDK 64-Bit Server VM (build 25.121-b13, mixed mode)
```

DarkDimius:
@nicolasstucki please review.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ElimStaticThis.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ElimStaticThis.scala
@@ -27,10 +27,10 @@ class ElimStaticThis extends MiniPhaseTransform {
   override def transformIdent(tree: tpd.Ident)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
     if (ctx.owner.enclosingMethod.is(JavaStatic)) {
       tree.tpe match {
-        case TermRef(thiz: ThisType, _) if thiz.cls.is(ModuleClass) =>
+        case TermRef(thiz: ThisType, _) if thiz.cls.is(ModuleClass, JavaDefined) =>
           ref(thiz.cls.sourceModule).select(tree.symbol)
         case TermRef(thiz: ThisType, _) =>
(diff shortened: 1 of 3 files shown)


nicolasstucki:
@DarkDimius do not write the `.` at the end of the first line of the commit message to avoid the @dotty-bot complaints.

compiler/src/dotty/tools/dotc/transform/ElimStaticThis.scala
```diff
@@ -27,10 +27,10 @@ class ElimStaticThis extends MiniPhaseTransform {
   override def transformIdent(tree: tpd.Ident)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
     if (ctx.owner.enclosingMethod.is(JavaStatic)) {
       tree.tpe match {
-        case TermRef(thiz: ThisType, _) if thiz.cls.is(ModuleClass) =>
+        case TermRef(thiz: ThisType, _) if thiz.cls.is(ModuleClass, JavaDefined) =>
```
smarter: please use `butNot =` for clarity.

smarter: I'm also not sure why you check JavaDefined if this is for packages, why not check `Package` ?

PR #2460, opened 2017-05-17
Change enum scheme to correspond to new description in issue #1970

odersky:
Always demand that type parameters of cases are given explicitly.


odersky:
Only first commit needs to be reviewed.

propensive:
Was the motivation that the "parameter-inference" was just too magic, otherwise?

odersky:
Too much magic, and not consistent with design principles elsewhere, yes.

sjrd:
The changes look good. I think it would be good to add tests for things that were previously impossible/very awkward. For example:
```scala
enum E[+T] {
  case A[T, U](t: T, u: U) extends E[T] // is the explicit extends necessary here?
  case B
}
```
and things like that.

odersky:
@sjrd I'll roll in the tests in a future PR. Have to merge now to write the docs.

PR #2461, opened 2017-05-17
Fix #2088: Test for simple cycle in type aliases

### LINKED ISSUE #2088: StackOverflow with cyclic structural type
This doesn't compile, as expected:
```scala
class Base {
  type A = {
    val m: A
  }
}
```
```scala
2 |  type A = {
  |       ^
  |illegal cyclic reference: alias Object{m: Base.this.A} of type A refers back to the type itself
```
But this crashes instead of giving an error:
```scala
class Base {
  type A = {
    val m: Foo.A
  }
}

object Foo extends Base
```
```scala
        at dotty.tools.dotc.core.Types$TypeAccumulator.foldOver(Types.scala:3644)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:100)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:84)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:136)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$.apply(VarianceChecker.scala:83)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:89)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:84)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:136)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$.apply(VarianceChecker.scala:83)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$.apply(VarianceChecker.scala:27)
        at dotty.tools.dotc.core.Types$TypeAccumulator.foldOver(Types.scala:3644)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:100)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:84)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:136)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$.apply(VarianceChecker.scala:83)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:89)
        at dotty.tools.dotc.typer.VarianceChecker$Validator$$anonfun$apply$2.apply(VarianceChecker.scala:84)
        at dotty.tools.dotc.reporting.Reporting$class.traceIndented(Reporter.scala:136)
...
```

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala
```diff
@@ -86,7 +86,15 @@ class VarianceChecker()(implicit ctx: Context) {
         case tp: TypeRef =>
           val sym = tp.symbol
           if (sym.variance != 0 && base.isContainedIn(sym.owner)) checkVarianceOfSymbol(sym)
-          else if (sym.isAliasType) this(status, sym.info.bounds.hi)
+          else if (sym.isAliasType) {
+            val aliasUpperBound = sym.info.bounds.hi
+            aliasUpperBound match {
+              case aliasUpperBound: RefinedType if aliasUpperBound.refinedInfo == tp =>
+                ctx.error("illegal cyclic reference involving " + sym + " and " + aliasUpperBound.refinedName, sym.pos)
```
smarter: Even if this happens to explode in VarianceChecker right now, this isn't specific to VarianceChecker in any way. I would expect this to be checked in CheckNonCyclicMap

nicolasstucki: It can be done in CheckNonCyclicMap.

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -206,7 +206,9 @@ object Checking {
           def isInteresting(prefix: Type): Boolean = prefix.stripTypeVar match {
             case NoPrefix => true
             case prefix: ThisType => sym.owner.isClass && prefix.cls.isContainedIn(sym.owner)
-            case prefix: NamedType => !prefix.symbol.isStaticOwner && isInteresting(prefix.prefix)
+            case prefix: NamedType =>
+              !prefix.symbol.isStaticOwner &&
+              (pre.derivesFrom(sym.owner) || isInteresting(prefix.prefix))
```
smarter: should this be prefix.derivesFrom ? In any case could you explain why this is the correct thing to check?

nicolasstucki: In this case we are asking if the prefix `Foo.A` is interesting. It is as `Foo` extends `Base` (where `A` is defined).

nicolasstucki: It might be able to reduce the false positives by checking if `sym` if not protected or private. I will add that.

nicolasstucki: Oh, it should be `prefix.derivesFrom` not `pre.derivesFrom`

Issue #2115, opened 2017-03-17
Add timestamp to generated docs


felixmulder:
We should add a timestamp like

```html
<!-- generated on: 2017-03-17 18:31:20 -->
```

Somewhere in the html so that we can know when the docs were last generated.

trancynn:
Looks like that information should be more accessible (it's inconvenient to look through the HTML source just to find out when docs were generated). How about adding it somewhere else (e.g. sidebar)?

felixmulder:
I'd rather not - in the 2.12 docs we opted for adding the version number to the top-left corner. When they do nightlies, it gets the git hash and date.

That's fine - but I'd rather not bloat the UI by adding a timestamp by default.

For our purposes, we can just look at the `gh-pages` branch to see that the docs were last generated a while ago.

Varunram:
@felixmulder Is it nice if we add it at the top? Or are you looking for a more specific location?

felixmulder:
Sure, this is fine - it could be done in combination with adding the project title somewhere :)

PR #2465, opened 2017-05-18
Add timestamp to generated docs

Varunram:
Intended as a solution to #2115.
Contrary to what was proposed back in the issue, appending at the end of the file seemed to be an easier alternative than reading from the file and rewriting - I'm new here, so please correct me if I'm wrong.

### DIFF
--- a/doc-tool/resources/_layouts/main.html
+++ b/doc-tool/resources/_layouts/main.html
@@ -1,3 +1,4 @@
+<!-- This page was last generated on {{ page.date | date: "%b-%d-%Y" }} -->
 <!DOCTYPE html>
 <html lang="en">
     <head>
--- a/doc-tool/src/dotty/tools/dottydoc/staticsite/DefaultParams.scala
+++ b/doc-tool/src/dotty/tools/dottydoc/staticsite/DefaultParams.scala
@@ -5,6 +5,8 @@ package staticsite
 import model.{ Entity, Package, NonEntity }
 
 import java.util.{ HashMap, List => JList, Map => JMap }
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 import scala.collection.JavaConverters._
 
 case class DefaultParams(
@@ -58,7 +60,7 @@ case class DefaultParams(
   def withDate(d: String) = copy(page = PageInfo(page.url, d))
 }
 
-case class PageInfo(url: String, date: String = "") {
+case class PageInfo(url: String, date: String = LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd")).toString ) {
   val path: Array[String] = url.split('/').reverse.drop(1)
 }
 


Varunram:
Signed the CLA!

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
```diff
@@ -311,6 +315,16 @@ case class Site(val root: JFile, val projectTitle: String, val documentation: Ma
     * defaults, the user-defined one will take precedence.
     */
   val layouts: Map[String, Layout] = {
+
+    val cal = Calendar.getInstance()
```
smarter: I would recommend against using java.util.Date/Calendar/SimpleDataFormat. These APIs have many pitfalls and are not thread-safe.  Java 8 introduced a new set of APIs to deal with date and time in the `java.time` package that should be used instead. To get the current time you should use LocalDateTime.now: https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#now--

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
```diff
@@ -311,6 +315,16 @@ case class Site(val root: JFile, val projectTitle: String, val documentation: Ma
...
+    val today = cal.getTime()
+    val currentHour = new SimpleDateFormat("hh").format(today)
+    val currentMinute = new SimpleDateFormat("mm").format(today)
+    val amOrPm = new SimpleDateFormat("a").format(today)
+    val x = "<!-- generated on " + cal.get(Calendar.DATE) + "/"+ (cal.get(Calendar.MONTH)+1)+ "/"+cal.get(Calendar.YEAR) + " " + currentHour + ":" + currentMinute + ":" + amOrPm + "-->"
+    var loc = root.getAbsolutePath.toString + "/.." + "/doc-tool/resources/_layouts/main.html"
+    scala.tools.nsc.io.File(loc).appendAll(x)
```
smarter: This is an alias for dotty.tools.io.File, use that instead.

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
Varunram: Thank you very much! Doing it right away!

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
Varunram: Thanks again and Sorry. Regarding the failing tests, would it be better if I add a check for ```loc ``` over here?

Varunram:
CLA signed already

Varunram:
Noticed this just now, sorry! Would a tacky fix along the lines of ```case class PageInfo(url: String, date: String = (LocalDateTime.now().getDayOfMonth() + "/"+ LocalDateTime.now().getMonthValue()+ "/"+ LocalDateTime.now().getYear() + " " + LocalDateTime.now().getHour() + ":" + LocalDateTime.now().getMinute() + ":" + LocalDateTime.now().getSecond()).toString) {
``` in ```DefaultParams.scala``` be acceptable?

**Comment by felixmulder** [comment]
@Varunram - we want to support people in doing something like:

```
{{ "2017-10-16" | date: '%B %d, %Y' }}
```

Therefore it might be better to fill in the correct date in `DefaultParams.scala` and use that if there wasn't a date already defined in the preamble.

Varunram:
Pushed back irrelevant commits and did the requested changes. Thanks to @felixmulder for helping me out 😄

felixmulder:
LGTM, thanks @Varunram !

PR #2476, opened 2017-05-19
Fix #2421: Add errors for inline on non supported trees

### LINKED ISSUE #2421: Missing errors for `inline var`, `inline object`, `inline class` and `inline type`.
* [x] `inline var foo`: Should not be allowed, it will break after the first assignment. Possibly for `lazy val` too.
* [x] `inline class Foo`: Should not be allowed.
* [x] `inline object Foo`: Should have a better error message. Currently _right-hand side of inline object Foo must be a constant expression or a function_.
* [x] `inline type Foo`: Should not be allowed, it makes not sense.

nicolasstucki:


felixmulder:
@nicolasstucki - needs rebase

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -550,6 +550,28 @@ trait Checking {
...
   }
 
+  /** Check that inline is only used on valid trees */
+  def checkInlineKeyword(tree: tpd.Tree)(implicit ctx: Context) = {
+    def errorOn(what: String) =
+      ctx.error("inlined keyword cannot be used on " + what, tree.pos)
+    val sym = tree.symbol
+    if (sym is Inline) {
+      tree match {
+        case _: TypeDef =>
```
DarkDimius: PlainPrinter.kindString

nicolasstucki:
Moved checks to checkFlags

nicolasstucki:
All requested changes were addressed.

nicolasstucki:
@odersky it looks like your commits lost the test file https://github.com/dotty-staging/dotty/blob/ee331a6fa255638c8db59dae269198d3a8c3dba9/tests/neg/i2421.scala

odersky:
I found a solution that requires less code and fits better. @nicolasstucki do you want to review?

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -315,7 +315,10 @@ object Checking {
         fail(AbstractMemberMayNotHaveModifier(sym, flag))
     def checkNoConflict(flag1: FlagSet, flag2: FlagSet) =
       if (sym.is(allOf(flag1, flag2)))
-        fail(i"illegal combination of modifiers: $flag1 and $flag2 for: $sym")
+        fail(i"illegal combination of modifiers: `$flag1` and `$flag2` for: $sym")
+    def checkApplicable(flag: FlagSet, ok: Boolean) =
+      if (!ok && !sym.is(Synthetic))
```
nicolasstucki: Why is the `!sym.is(Synthetic)` needed here? Is it only for the `lazy val` of the modules?

odersky: For now yes, but there might be other synthetic symbols with conflicts. In any case, if it is synthetic we should not issue an error message to the user.

nicolasstucki: Ok

nicolasstucki:
LGTM

PR #2477, opened 2017-05-19
Some enhancements for Dottydoc and ScalaDays

felixmulder:


doc-tool/src/dotty/tools/dottydoc/model/internal.scala
```diff
@@ -86,13 +86,15 @@ object internal {
...
-    modifiers: List[String],
+    private val mods: List[String],
     path: List[String],
     superTypes: List[MaterializableLink] = Nil,
     var comment: Option[Comment] = None,
     var companionPath: List[String] = Nil,
     var parent: Entity = NonEntity
-  ) extends Object
+  ) extends Object {
+    def modifiers: List[String] = mods.filterNot(_ == "private")
```
DarkDimius: should it be "final"?

felixmulder: waps, it should indeed!

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -123,5 +123,12 @@ class ScalaSettings extends Settings.SettingGroup {
     sys.props("user.dir").split(File.separatorChar).last
   )
 
+  val projectVersion = StringSetting (
+    "-project-version",
+    "project version",
+    "The current version of your project",
+    ""
```
nicolasstucki: Why is it empty?

felixmulder: If the version isn't specified, it falls back to an empty string. This renders nicely on the site.

nicolasstucki: Ok

nicolasstucki:
The new side bar looks nice :)

PR #2478, opened 2017-05-19
Fixed #1169: Add regression tests

### LINKED ISSUE #1169: Incorrect bridge gen for method that has value class with underlying Object type as result type
Incorrect bridge generation for method that has value class with the underlying Object type as result type.
The compilation of this code with -Ycheck:all:

``` scala
class A(val x: Object) extends AnyVal
abstract class VCACompanion[T] {
  def yyyy(x: Object): T
}
class XXX extends VCACompanion[A] {
  override def yyyy(x: Object): A = new A("fff")
}
```

results in:

```
 method yyyy is already defined as method yyyy: (x: Object)ErasedValueType(A, Object)
```

The problem is that there is a bridge generation during Erasure for `def yyyy(x: Object): A = ...`. 
This bridge has the same signature as original `def yyyy(x: Object): A = ...`:

```
override def yyyy(x: Object): Object = "fff"
def yyyy(x: Object): Object = new A(this.yyyy(x))
```

nicolasstucki:


tests/run/i1169b.scala
```diff
@@ -0,0 +1,18 @@
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    val xProducer = new XProducer
+    println(xProducer.produce.underlying)
```
DarkDimius: Neither of this tests is compete.

Those tests should additionally check that `Producer[T]` behaves correctly in a generic position. E.g. try defining a `toString` in X and calling it in a generic context of `Producer[T]`, where T isn't known to be `X`.
I would be suprised if it does. 

I had a look on generated bytecode. The bug is NOT fixed and the bytecode is wrong.

DarkDimius: This code should not compile as our current Value-class encoding can not handle it. Details can be found here: https://github.com/lampepfl/dotty/issues/1169#issuecomment-196784201

nicolasstucki: I misread the comment

PR #2480, opened 2017-05-19
Fix #1818: Whitelist Function1 and Function2

### LINKED ISSUE #1818: Whitelisting scala.Function1 fails to compile in partest.
When running partest with `Function1` whitlisted the following error is reported:
```scala
-- Error: /Users/nicolas.stucki/GitHub/dotty/tests/partest-generated/pos/compileStdLib/Function1.scala -----------------
35 |trait Function1[@specialized(scala.Int, scala.Long, scala.Float, scala.Double) -T1, @specialized(scala.Unit, scala.Boolean, scala.Int, scala.Float, scala.Long, scala.Double) +R] extends AnyRef { self =>
   |                ^^^^^^^^^^^^
   |                cyclic reference involving constructor specialized
```

nicolasstucki:


DarkDimius:
LGTM after the tests pass. Curious to know if the cyclic reference error was fixed(and how), or we just compile files in different order know.

nicolasstucki:
This was a bootstrap issue, we did a lot of improvements since it was blacklisted.

DarkDimius:
@nicolasstucki CI disagrees with you.

nicolasstucki:
@DarkDimius I ran all the CI tests on my machine before sending this PR. I will have to double check :(

nicolasstucki:
@DarkDimius my machine disagrees with CI. Again...

DarkDimius:
@nicolasstucki, different order of files? One of the reasons whitelist existed before is that it also provided a specific order of files.

nicolasstucki:
I will try adding an order to the files. One that does not depend on the filesystem.

DarkDimius:
@nicolasstucki I propose to order not by the file name, but instead by class name.
The reason is due to way people report bugs. They rarely post file names.

smarter:
We already have some code to order by filename, but I guess it's not used everywhere: https://github.com/lampepfl/dotty/pull/2193
I'd like to keep ordering things by filename because I have at least one test which relies on it.

nicolasstucki:
Will close this PR for now. This is not a priority.

Issue #2463, opened 2017-05-18
Top level implicit class with no arguments crashes in messages


nicolasstucki:
The source
```scala
implicit class Foo
```
crashes with
```
Exception in thread "main" java.lang.IndexOutOfBoundsException: 0
	at scala.collection.LinearSeqOptimized$class.apply(LinearSeqOptimized.scala:65)
	at scala.collection.immutable.List.apply(List.scala:84)
	at dotty.tools.dotc.reporting.diagnostic.messages$TopLevelImplicitClass.<init>(messages.scala:355)
	at dotty.tools.dotc.ast.desugar$$anonfun$17.apply(Desugar.scala:475)
	at dotty.tools.dotc.ast.desugar$$anonfun$17.apply(Desugar.scala:475)
	at dotty.tools.dotc.reporting.diagnostic.MessageContainer.contained(MessageContainer.scala:59)
	at dotty.tools.dotc.reporting.diagnostic.MessageContainer.message(MessageContainer.scala:42)
	at dotty.tools.dotc.reporting.diagnostic.MessageContainer.isNonSensical(MessageContainer.scala:71)
	at dotty.tools.dotc.reporting.HideNonSensicalMessages$class.isHidden(HideNonSensicalMessages.scala:17)
	at dotty.tools.dotc.reporting.ConsoleReporter.isHidden(ConsoleReporter.scala:13)
	at dotty.tools.dotc.reporting.Reporter.report(Reporter.scala:232)
	at dotty.tools.dotc.reporting.Reporting$class.error(Reporter.scala:89)
	at dotty.tools.dotc.core.Contexts$Context.error(Contexts.scala:57)
	at dotty.tools.dotc.ast.desugar$.classDef(Desugar.scala:475)
	at dotty.tools.dotc.ast.desugar$.defTree(Desugar.scala:655)
	at dotty.tools.dotc.typer.Namer.expand(Namer.scala:364)
	at dotty.tools.dotc.typer.Namer$$anonfun$index$1.apply(Namer.scala:636)
	at dotty.tools.dotc.typer.Namer$$anonfun$index$1.apply(Namer.scala:636)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at dotty.tools.dotc.typer.Namer.index(Namer.scala:636)
	at dotty.tools.dotc.typer.Namer.dotty$tools$dotc$typer$Namer$$recur$1(Namer.scala:427)
[...]

PR #2481, opened 2017-05-19
Fix crash for top level implicit class declaration

Varunram:
Intended as a fix for #2463

### DIFF
--- a/tests/neg/i2463.scala
+++ b/tests/neg/i2463.scala
@@ -1,0 +1,1 @@
+implicit class Foo // error
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -352,7 +352,8 @@ object messages {
     val explanation = {
       val TypeDef(name, impl @ Template(constr0, parents, self, _)) = cdef
       val exampleArgs =
-        constr0.vparamss(0).map(_.withMods(untpd.Modifiers()).show).mkString(", ")
+        if(constr0.vparamss.isEmpty) "..."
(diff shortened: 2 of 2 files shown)


nicolasstucki:
That line looks a bit too long, could you shorten it like this:
```scala
val exampleArgs = 
  if(constr0.vparamss.isEmpty) "..." 
  else constr0.vparamss(0).map(_.withMods(untpd.Modifiers()).show).mkString(", ")
```

nicolasstucki:
You should also add the test file and rename the commit to `Fix #2463: Fix crash for top level implicit class declaration` (this way github links the commit with the issue).

Thanks for the PR

Varunram:
Your wish is my command! XD

Varunram:
Sorry, just saw the second half of your message. Tests? I thought they were the same, aren't they?

nicolasstucki:
Yes, I meant the `tests/neg/i2463.scala` file.

Varunram:
Done! The reason I asked you was I couldn't find any other test file that needed changes. Thanks a lot!

tests/neg/i2463.scala
```diff
@@ -0,0 +1 @@
+implicit class Foo //error
```
nicolasstucki: You need to add a space between `//` and `error`. The framework looks for the substring `// error` in the source.

nicolasstucki: I fixed it directly from my browser, this should work now.

Varunram: Oh no, I too pushed simultaneously, correcting it now

Varunram:
I force-pushed to fix the commit history. Sorry!

tests/neg/i2463.scala
nicolasstucki: Your fast

PR #2484, opened 2017-05-19
Move to error case classes (Friday)

ennru:
Main effort was to move to error case classes in `Checking.scala`, some more explanations could be beneficial.
All changed scenarios are covered in ErrorMessagesTests.

`DoesNotConformToSelfTypeCantBeInstantiated` used to be without source position, don't know why.

@felixmulder I warned you ;-)

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -20,6 +20,7 @@ import printing.Formatting
 import ErrorMessageID._
 import Denotations.SingleDenotation
 import dotty.tools.dotc.ast.Trees
+import dotty.tools.dotc.core.Flags.{FlagSet, Mutable}
 import dotty.tools.dotc.core.SymDenotations.SymDenotation
 
 object messages {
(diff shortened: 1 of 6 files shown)


felixmulder:
😆 

Alright!

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1351,4 +1352,132 @@ object messages {
...
+          |
+          |Note: Self types are indicated with the notation
+          |  ${s"class "}$tp ${"{ this: "}$selfType${" => "}
+          |"""
+  }
+
+  case class AbstractMemberMayNotHaveModifier(sym: Symbol, flag: FlagSet)(
+    implicit ctx: Context)
+    extends Message(AbstractMemberMayNotHaveModifierID) {
+    val msg = hl"""abstract $sym may not have `$flag' modifier"""
```
felixmulder: Could've been `hl"""${"abstract"} $sym..."""`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1351,4 +1352,132 @@ object messages {
...
+    val kind = "Syntax"
+    val explanation = ""
+  }
+
+  case class TraitsMayNotBeFinal(sym: Symbol)(
+    implicit ctx: Context)
+    extends Message(TraitsMayNotBeFinalID) {
+    val msg = hl"""$sym may not be ${"final"}"""
+    val kind = "Syntax"
+    val explanation = "A trait can never be final."
```
felixmulder: ~"...because you cant instantiate it since it is abstract by nature" or something?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1351,4 +1352,132 @@ object messages {
...
+    val msg = hl"""only classes can have declared but undefined members"""
+    val kind = "Syntax"
+    val explanation = s"$varNote"
+  }
+
+  case class CannotExtendAnyVal(sym: Symbol)(implicit ctx: Context)
+    extends Message(CannotExtendAnyValID) {
+    val msg = hl"""$sym cannot extend ${"AnyVal"}"""
+    val kind = "Syntax"
+    val explanation = ""
```
felixmulder: Is there something we can say here in `explanation`? I know you wanted to just get everything over to the new format first - but can't help myself 😆

ennru: "Traits do not support extending AnyVal" ?
"AnyVal instances must be representable in a single value, which excludes usage of traits"?

felixmulder: I leave it up to your discretion :)

smarter: If someone tries to make a trait extending AnyVal, they should be told about universal traits (traits extending Any): http://docs.scala-lang.org/overviews/core/value-classes.html

ennru: This leads to the question how we'd support adding relevant links to `Message`s. Links to the lang specs will be helpful in many cases, Stackoverflow might be good, too.
But links do not always age well.

PR #2485, opened 2017-05-20
Explanation for 'traits cannot extend AnyVal'

ennru:
Found a second place with the same error.
After discussion in #2484.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1470,7 +1470,10 @@ object messages {
     extends Message(CannotExtendAnyValID) {
     val msg = hl"""$sym cannot extend ${"AnyVal"}"""
     val kind = "Syntax"
-    val explanation = ""
+    val explanation =
+      hl"""Only classes (not traits) are allowed to extend ${"AnyVal"}, but traits may extend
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1470,7 +1470,10 @@ object messages {
     extends Message(CannotExtendAnyValID) {
     val msg = hl"""$sym cannot extend ${"AnyVal"}"""
     val kind = "Syntax"
-    val explanation = ""
+    val explanation =
+      hl"""Only classes (not traits) are allowed to extend ${"AnyVal"}, but traits may extend
+          |${"Any"} to become ${Green("\"universal traits\"")} which may only have ${"def"} members.
```
nicolasstucki: I might be good to also mention that classes extending `AnyVal` can extend `universal traits`.

ennru:
I never thought of universal traits being so popular ;-) 
I hope everybody is happy with the explanation now.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1470,7 +1470,11 @@ object messages {
...
+          |Universal traits can be mixed in to classes that extend ${"AnyVal"}.
```
smarter: in to -> into

PR #2486, opened 2017-05-20
Fix #2063: Print tabs in message padding

### LINKED ISSUE #2063: Incorrect underlined position in error message when line contains tabs
```scala
$ echo "\tclass Foo extends Bar" > tab.scala
$ dotc tab.scala
-- [E006] Unbound Identifier Error: try/tab.scala ------------------------------
1 |     class Foo extends Bar
  |                  ^^^
  |                  not found: type Bar
$ scalac tab.scala
tab.scala:1: error: not found: type Bar
        class Foo extends Bar
                          ^
```

nicolasstucki:


compiler/src/dotty/tools/dotc/reporting/MessageRendering.scala
```diff
@@ -101,9 +100,12 @@ trait MessageRendering {
 
       math.min(currPad, minPad)
     }
+    val padding =
+      if (leastWhitespace - offset == pos.startColumn) pos.startColumnPadding
+      else " " * (leastWhitespace - offset)
```
smarter: I don't understand why this is correct. Why not use startColumnPadding in currPad above?

compiler/src/dotty/tools/dotc/util/SourceFile.scala
```diff
@@ -136,6 +136,17 @@ case class SourceFile(file: AbstractFile, content: Array[Char]) extends interfac
     col
   }
 
+  /** The padding of the column corresponding to `offset`, includes tabs */
+  def startColumnPadding(offset: Int): String = {
+    var idx = startOfLine(offset)
+    val pad = new StringBuffer("")
```
smarter: Use StringBuilder instead to avoid the synchronization overhead of StringBuffer, also you can use the empty constructor: `new StringBuilder`

compiler/src/dotty/tools/dotc/reporting/MessageRendering.scala
nicolasstucki: Because we do not know the size of one tab.

nicolasstucki: But indeed I need to modify the logic a bit

PR #2488, opened 2017-05-20
Consider syntax with significant indentation

odersky:
This is a trial balloon to test syntax that allows to optionally replace braces with indentation. 
It implements #2491.

For the moment this is for experimentation only, no decision has been made to merge this.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Comments.scala
+++ b/compiler/src/dotty/tools/dotc/core/Comments.scala
@@ -8,9 +8,11 @@ import util.SourceFile
 import util.Positions._
 import util.CommentParsing._
 import util.Property.Key
+import util.Chars.{isIdentifierPart, isOperatorPart}
 import parsing.Parsers.Parser
 import reporting.diagnostic.messages.ProperDefinitionNotFound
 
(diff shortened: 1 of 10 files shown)


compiler/src/dotty/tools/dotc/core/Comments.scala
```diff
@@ -50,6 +54,19 @@ object Comments {
...
+     */
+    val endCommentString: String =
+      if (raw.startsWith(endCommentPrefix)) {
+        val contents = raw.drop(endCommentPrefix.length)
+        val str = contents.takeWhile(c => isIdentifierPart(c) || isOperatorPart(c))
+        val follow = contents.drop(str.length)
+        if (follow.isEmpty || ".;,".contains(follow.head)) str else ""
+      }
+      else ""
+
```
felixmulder: Does this add semantic meaning to comments?

odersky: Not really, since you only get a warning if there's misalignment.

felixmulder:
Closing this for now, see: https://github.com/lampepfl/dotty/issues/2491#issuecomment-304212551

PR #2498, opened 2017-05-22
Reference sections for some of the new things in Dotty

odersky:


docs/docs/reference/implicit-by-name-parameters.md
```diff
@@ -0,0 +1,42 @@
...
+  val s = serializeOption[Option[Int]]
+  s.write(Some(33))
+  s.write(None)
+```
+As is the case for a normal by-name parameter, the argument for the implicit parameter `ev`
+is evaluated on demand. In the example above, if the option value `x` is `None`, it is
+not evaluated at all.
+
+The synthesized argument for an implicit parameter is backed by a lazy
+val, which means that the parameter is evaluated at most once. The
```
japgolly: If it's backed by a lazy val would it not be better to call this feature By-Need Parameters instead? Is there a reason that it's being described as call-by-name instead of call-by-need?

smarter:
How is the better type inference for new commit related to the rest? Could we get a testcase for it?

docs/docs/reference/implicit-by-name-parameters.md
odersky: No, it's complicated. The lazy val is an internal implementation detail of the synthesized argument. It is not generated at all unless the synthesis is recursive. The parameter itself is by-name, i.e. the argument computation can be invoked multiple times.

docs/docs/reference/adts.md
```diff
@@ -0,0 +1,115 @@
...
+parameterized with a type parameter `T` and a value parameter `x`. It
+is a shorthand for writing a case class that extends `Option`. Since
+`None` is not parameterized it is treated as a normal enum value.
+
+The `extends` clauses that were omitted in the example above can also
+be given explicitly:
+
+```scala
+enum Option[+T] {
+  case Some[T](x: T) extends Option[T]
```
japgolly: missing covariance

docs/docs/reference/adts.md
```diff
@@ -0,0 +1,115 @@
...
+        EnumDef ::=  id ClassConstr [`extends' [ConstrApps]]
+                     [nl] `{’ EnumCaseStat {semi EnumCaseStat} `}’
+
+2. Cases of enums are defined as follows:
+
+        EnumCaseStat  ::=  {Annotation [nl]} {Modifier} EnumCase
+        EnumCase      ::=  `case' (EnumClassDef | ObjectDef | ids)
+        EnumClassDef  ::=  id [ClsTpeParamClause | ClsParamClause]
+                           ClsParamClauses TemplateOpt
+        TemplateStat  ::=  ... | EnumCaseStat
```
japgolly: Should this be TemplateOpt or the above line be TemplateStat

docs/docs/reference/adts.md
```diff
@@ -0,0 +1,122 @@
...
+
+```scala
+enum Option[+T] {
+  case Some[+T](x: T) extends Option[T]
+  case None           extends Option[Nothing]
+}
+```
+
+Note that the parent type of `None` is inferred as
+`List[Nothing]`. Generally, all covariant type parameters of the enum
```
fthomas: That should be `Option[Nothing]`.

odersky:
I think this is ready to go in now. Would be good to have something up on the docsite so we can play with the highlighting and other appearance aspects.

felixmulder:
I'll have a look at this after dinner, amend any changes myself then merge it. Sounds good?

felixmulder:
Silence => agreement => merge! :tada:

PR #2501, opened 2017-05-22
Add note on phantom universe mismatch

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -120,6 +120,8 @@ object ErrorReporting {
       val expected1 = dropJavaMethod(expected)
       if ((found1 eq found) != (expected eq expected1) && (found1 <:< expected1))
         "\n (Note that Scala's and Java's representation of this type differs)"
+      else if (found.topType != expected.topType)
+        "\n (Note that the types are in different universes, see Phantom types)"
```
felixmulder: universes => hierarchies?

nicolasstucki: We started using universes for these difference in lattices. 

https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala#L538-L548

felixmulder: Alright, then LGTM :)

PR #2503, opened 2017-05-22
[scaladays] dottydoc improvements 2

felixmulder:
This PR is best detailed by the commit messages. Github previously panicked since the AST index was too large (>200mb) it is now around 4mb for the dotty project.


nicolasstucki:
There is a small issue, texts from `Entry Results` can overflow to to `Member Result` column. The are a couple of `bol), List())` overflowing in this example.
screen shot 2017-05-23 at 14 10 57

PR #2504, opened 2017-05-22
Fix #2502: improve warning message of exhaustivity check

### LINKED ISSUE #2502: incorrect exhaustiveness warning in dotty build
Running `sbt dotty-compiler-bootstrapped/compile` we get:
```scala
[warn] -- [E028] Pattern Match Exhaustivity Warning: compiler/../out/bootstrap/dotty-compiler-bootstrapped/scala-0.1/src_managed/main/scala/tools/nsc/backend/jvm/BTypes.scala:337:35 
[warn] 337 |    def classOrArrayType: String = this match {
[warn]     |                                   ^^^^
[warn]     |                                   match may not be exhaustive.
[warn]     |                                   
[warn]     |                                   It would fail on: ClassBType(_)
```

But the match will not fail on `ClassBType(_)`, the first case takes care of it:
```scala
    def classOrArrayType: String = this match {
      case ClassBType(internalName) => internalName
      case a: ArrayBType            => a.descriptor
    }
```

/cc @liufengyun

liufengyun:
Fix #2502: improve warning message of exhaustivity check

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -641,6 +641,14 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
   /** Display spaces */
   def show(s: Space): String = {
+
+    // does the companion object of the given symbol have custom unapply
+    def hasCustomUnapply(sym: Symbol): Boolean = {
+      val companion = sym.companionModule
+      companion.findMember(nme.unapply, NoPrefix, Flags.Synthetic).exists ||
+        companion.findMember(nme.unapplySeq, NoPrefix, Flags.Synthetic).exists
+    }
+
     def doShow(s: Space, mergeList: Boolean = false): String = s match {
       case Empty => ""
       case Typ(c: ConstantType, _) => c.value.show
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -657,7 +657,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         else if (tp.classSymbol.is(CaseClass))
         // use constructor syntax for case class
           showType(tp) + signature(tp).map(_ => "_").mkString("(", ", ", ")")
-        else if (signature(tp).nonEmpty)
+        else if (sym.is(Flags.Case) && signature(tp).nonEmpty)
```
smarter: Isn't this condition a superset of the condition of the previous if (`if (tp.classSymbol.is(CaseClass))`) and so this branch is just dead code?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -657,7 +657,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         else if (tp.classSymbol.is(CaseClass))
```
smarter: Even if a class is a case class, it might contain a custom unapply, for example:
```scala
abstract class BTypes {
  trait BType

  sealed trait RefBType extends BType {
    def classOrArrayType: String = this match {
      case ClassBType(internalName) => internalName
      case a: ArrayBType            => ""
    }
  }

  final case class ClassBType(val internalName: String) extends RefBType
  class ArrayBType extends RefBType

  object ClassBType {
    def unapply(x: RefBType): Option[String] = None
  }
}
```
Even with this patch, it warns `It would fail on: BTypes.ClassBType(_)`

liufengyun: Good catch, it's more reliable to show just `_: T` for type space.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -641,6 +641,14 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
   /** Display spaces */
   def show(s: Space): String = {
+
+    // does the companion object of the given symbol have custom unapply
+    def hasCustomUnapply(sym: Symbol): Boolean = {
+      val companion = sym.companionModule
+      companion.findMember(nme.unapply, NoPrefix, Flags.Synthetic).exists ||
```
smarter: please use `excluded = Synthetic` to avoid confusion.

PR #2505, opened 2017-05-22
Move ValueClass checking errors to case class scheme

ennru:
@felixmulder I'm a bit uncertain if the change to parameter checking in `Checking.checkDerivedValueClass` is optimal. It appeared to me the expected `List(param)` can not occur, vars always become two params (which both are marked mutable).

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -439,35 +439,39 @@ object Checking {
   def checkDerivedValueClass(clazz: Symbol, stats: List[Tree])(implicit ctx: Context) = {
     def checkValueClassMember(stat: Tree) = stat match {
       case _: TypeDef if stat.symbol.isClass =>
-        ctx.error(s"value class may not define an inner class", stat.pos)
+        ctx.error(ValueClassesMayNotDefineInner(clazz, stat.symbol), stat.pos)
       case _: ValDef if !stat.symbol.is(ParamAccessor) =>
-        ctx.error(s"value class may not define non-parameter field", stat.pos)
-      case d: DefDef if d.symbol.isConstructor =>
-        ctx.error(s"value class may not define secondary constructor", stat.pos)
+        ctx.error(ValueClassesMayNotDefineNonParameterField(clazz, stat.symbol), stat.pos)
+      case _: DefDef if stat.symbol.isConstructor =>
+        ctx.error(ValueClassesMayNotDefineASecondaryConstructor(clazz, stat.symbol), stat.pos)
       case _: MemberDef | _: Import | EmptyTree =>
       // ok
       case _ =>
-        ctx.error(s"value class may not contain initialization statements", stat.pos)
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -439,35 +439,39 @@ object Checking {
...
       if (isCyclic(clazz.asClass))
-        ctx.error("value class cannot wrap itself", clazz.pos)
+        ctx.error(ValueClassesMayNotWrapItself(clazz), clazz.pos)
       else {
         val clParamAccessors = clazz.asClass.paramAccessors.filter(_.isTerm)
         clParamAccessors match {
+          case List(acc1, acc2) if acc1.is(Mutable) =>
+              ctx.error(ValueClassParameterMayNotBeAVar(clazz, acc1), acc1.pos)
+          case List(acc1, acc2) if acc2.is(Mutable) =>
+              ctx.error(ValueClassParameterMayNotBeAVar(clazz, acc2), acc2.pos)
```
felixmulder: This looks weird to me - a value class may only have one constructor parameter - as such, these cases should never occur and the `_` case should tell you this, right?

Was there a reason I'm not seeing for which you needed these?

ennru: Yes, there is. 

The "may not be var" error would never show. For `class MyClass(var i: Int) extends AnyVal` it becomes a `List(acc1, acc2)` for set and get access, the existing code expects `List(acc)` which does not happen for `var`s. A different choice is to skip the more specific error "may not be var" and just keep the "must be exactly one val".

felixmulder: Ah, I see what you mean!

You're not looking at the class constructor here, to do that you'd call `clazz.asClass.primaryConstructor`. You're instead looking at the parameter accessors - and `var`s get a setter which is why you're seeing two items.

ennru: Are you saying, we should change to look at the constructor parameters?
Makes more sense to me. The exisiting check looks at the accessors - any connection to the check for Phantom types?

felixmulder: Not sure - @nicolasstucki

nicolasstucki: Phantoms should not affect this. But you should not assume that the constructor will have 1 parameter. There could potentially be an illegal constructor with more or less parameters. In these cases you can just disregard this error as another one will be emitted for it.

nicolasstucki: You should also probably try to get them from `clazz.asClass.primaryConstructor`. It sounds cleaner.

ennru:
I'm sorry, I didn't find out how I should access the primary constructor parameters to see it there is a `var`.
It now gets the more generic error "value class needs to have exactly one val".

felixmulder:
No worries, I'll fix it.

ennru:
Please note the `@Ignore` on `ErrorMessagesTests.valueClassParameterMayNotBeVar`.

PR #2509, opened 2017-05-23
Fix #2494: avoid infinite loop in parser

### LINKED ISSUE #2494: Infinite loop in parser when enum followed by invalid code
If you try running the following:
```scala
enum 
object
```
You'll get an infinite loop around `enumCaseStats` in the parser, this is very similar to the issue that https://github.com/lampepfl/dotty/pull/2270 fixed for the REPL.

I'm not sure what the correct fix is here, but I'll note that other while loops in the parser check an `exitOnError` flag, I assume that this is to avoid this sort of things.

odersky:


smarter:
Two error markers are needed, I'll take care of updating the PR and merging:
```scala
-- [E039] Syntax Error: ../tests/neg/i2494.scala:2:0 -------------------------------------------------------------------
2 |object // error
  |^^^^^^
  |an identifier expected, but 'object' found
-- [E039] Syntax Error: ../tests/neg/i2494.scala:2:15 ------------------------------------------------------------------
2 |object // error
  |               ^
  |               an identifier expected, but eof found
two errors found
[error] Test dotc.tests.neg_all failed: java.lang.AssertionError: assertion failed: Wrong # of errors. Expected: 1, found: 2
```

PR #2523, opened 2017-05-24
Fix #2512: Implicit keyword not allowed on return types

### LINKED ISSUE #2512: Implicit keyword should not be allowed on return types
```scala
def foo1(f: Int => implicit Int): Int = ???
def foo2(f: Int => Int): implicit Int = ???
```
Are currently allowed but do nothing, there should be an error.

biboudis:
Update Parser to accommodate the new checks

Thanks @nicolasstucki for pair programming!

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -116,9 +116,9 @@ object Parsers {
     /** Issue an error at given offset if beyond last error offset
       *  and update lastErrorOffset.
       */
-    def syntaxError(msg: => Message, offset: Int = in.offset): Unit =
+    def syntaxError(msg: => Message, offset: Int = in.offset, name: Name = in.name): Unit =
       if (offset > lastErrorOffset) {
-        val length = if (in.name != null) in.name.show.length else 0
+        val length = if (name != null) name.show.length else 0
         syntaxError(msg, Position(offset, offset + length))
         lastErrorOffset = in.offset
       }
@@ -730,11 +730,16 @@ object Parsers {
     def typ(): Tree = {
       val start = in.offset
       val isImplicit = in.token == IMPLICIT
+      var isImplicitFun = false
       if (isImplicit) in.nextToken()
       def functionRest(params: List[Tree]): Tree =
         atPos(start, accept(ARROW)) {
           val t = typ()
-          if (isImplicit) new ImplicitFunction(params, t) else Function(params, t)
+          if (isImplicit) {
+            isImplicitFun = true
+            new ImplicitFunction(params, t)
+          }
+          else Function(params, t)
         }
       val t =
         if (in.token == LPAREN) {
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -771,6 +776,9 @@ object Parsers {
         }
         else infixType()
 
+      if (isImplicit && !isImplicitFun && in.token != ARROW)
+        syntaxError("Types with implicit keyword can only be function types", start, nme.IMPLICITkw)
+
       in.token match {
         case ARROW => functionRest(t :: Nil)
         case FORSOME => syntaxError(ExistentialTypesNoLongerSupported()); t
```
odersky: I think you can avoid the `isImplicitFun` logic and simply change the next line to:

     case _ => 
        if (isImplicit) syntaxError("Types with implicit keyword can only be function types", start)
        t

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -116,9 +116,9 @@ object Parsers {
     /** Issue an error at given offset if beyond last error offset
       *  and update lastErrorOffset.
       */
-    def syntaxError(msg: => Message, offset: Int = in.offset): Unit =
+    def syntaxError(msg: => Message, offset: Int = in.offset, name: Name = in.name): Unit =
```
odersky: I think it is a bit overkill to add another parameter here.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
nicolasstucki: Even if not, it will eliminate the `in.token != ARROW`

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
nicolasstucki: Then use directly `syntaxError(msg, Position(offset, offset + length))`

odersky: Why add length at all?

biboudis: Restored it, where `length = the length of implicit keyword`, otherwise it is the length of the identifier (in this case 3, from `Int`(.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -771,6 +776,9 @@ object Parsers {
         }
         else infixType()
 
+      if (isImplicit && !isImplicitFun && in.token != ARROW)
+        syntaxError("Types with implicit keyword can only be function types", Position(start, start + nme.IMPLICITkw.asSimpleName.length))
```
nicolasstucki: Why `asSimpleName`?

biboudis: Just to get the length, no?

nicolasstucki: It is fine

PR #2524, opened 2017-05-24
Fix #2516: Support Packaging and Distribution

### LINKED ISSUE #2516: Make a package for dotty distribution
- [x] make a sbt-task to create `dotty-VERSION-.tar.gz`:  [sbt-pack](https://github.com/xerial/sbt-pack)
- [x] make a brew script to ease the install on Mac

liufengyun:
Fix #2516: This PR enable us to run `brew install lampepfl/brew/dotty` and then play with `dotc/dotr` in command line.

The packaging is based on [sbt-pack](https://github.com/xerial/sbt-pack):

- `sbt dist/pack`: generate package under `dist/target/pack/`
- `sbt dist/pack-archive`: generate `.zip` and `.tar.gz` files


felixmulder:
It seems to missing Dottydoc :)

Is this publishable to homebrew as is, and could you take care of that?

Otherwise LGTM

liufengyun:
The scripts are assumed to support `mingw` and `cygwin`, so windows users have a chance to play with Dotty 0.1 .

liufengyun:
@felixmulder I'm not sure if we need to publish the formula to Homebrew -- they have a lot of requirements on what can be accepted to brew-core, I think a shorter URL to the formula satisfy our need for now.

olafurpg:
For homebrew, I recommend creating a cask so users can install dotty via `brew install lampepfl/dotty/dotty`. I do that for scalafmt https://github.com/olafurpg/homebrew-scalafmt

Reminds me there should really be a scala cask under `scala/homebrew-scala`, only someone would need to maintain it 😢

dist/dotty.rb
```diff
@@ -0,0 +1,32 @@
+class Dotty < Formula
+  desc "Experimental Scala Compiler"
+  homepage "http://dotty.epfl.ch/"
+  url "http://dotty.epfl.ch/dotty/dotty-0.1.1-bin-SNAPSHOT.tar.gz"
```
olafurpg: Will the artifacts always be hosted on dotty.epfl.ch? Do you have scripts to automate the deployment?

I recommend against serving the files from an EPFL server. There have examples where servers hosting artifacts have been compromised http://www.techradar.com/news/popular-video-encoding-mac-app-handbrake-compromised-with-malware I think it would be best to serve these from for example Github releases or Maven Central. I have used https://github.com/aktau/github-release to automate this, but there are other ways.

Additionally, I suspect many will write installation scripts against this url scheme. For example nix, https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/scalafmt/default.nix I would recommend to try and keep this as stable as possible.

dist/dotty.rb
```diff
@@ -0,0 +1,32 @@
...
+  sha256 "4e1bda148754543844d25290a87076e6bfb0b6b0275535f97c1871e0fc5c2c4c"
```
olafurpg: If you upload the artifacts to Github releases, then the sha will be different than running sha locally. You would have to upload, download and then run sha on the dowloaded file.

dist/dotty.rb
olafurpg: Here is the script I used to upload artifacts to github releases and update a homebrew formula https://github.com/scalameta/scalafmt/blob/ad8559939c7d32ca78dfedf3ebc4eebc2fda7257/bin/publish.sh#L52-L71 I don't use it anymore, I only publish to Maven Central now

liufengyun: Thanks a lot for the tip @olafurpg , I think we should automate the publishing to Github.

dist/bin/dotd
```diff
@@ -0,0 +1,114 @@
...
+
+  cd "$saveddir"
+fi
+
+source "$PROG_HOME/bin/common"
+
+default_java_opts="-Xmx768m -Xms768m"
+bootcp=true
+
+PROG_NAME=dotty.tools.dottydoc.Main
```
liufengyun: @felixmulder could you give me a tip on how the dottydoc script works? The main difference from `dotc` is just the main method?

I run `dotd examples/hello.scala` locally, but it seems we need more libraries: `java.lang.NoClassDefFoundError: com/vladsch/flexmark/util/options/MutableDataSet`.

liufengyun:
@olafurpg I agree, we can improve the user experience for Linux/Windows as we go. Currently `sbt-pack` creates zip packages, it's more visible about what's released.

dist/bin/dotd
felixmulder: That should be fine - are you sure you've provided all dependencies? Dottydoc has some dependencies from Maven. See: https://github.com/lampepfl/dotty/blob/master/project/Build.scala#L304-L317

dist/bin/dotc
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
+
+if [ -z "$PROG_HOME" ] ; then
```
DarkDimius: There's a lot of copy paste and re-implementation for the current `bin/dotc` script.
I propose we merge the two, otherwise it will be hard to maintain and we'll be using different scripts from the ones our users do.

dist/bin/dotd
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+
+if [ -z "$PROG_HOME" ] ; then
```
DarkDimius: No need to write separate scripts for multiple entry points. See how it's done in `./bin/dotc`

dist/bin/dotr
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+# Try to autodetect real location of the script
+
+if [ -z "$PROG_HOME" ] ; then
```
DarkDimius: Same comment here.

dist/bin/dotd
liufengyun: @DarkDimius I didn't get your point. This piece of code is a reliable way to get the package home in order to run `source "$PROG_HOME/bin/common"`. `bin/dotr` and `bin/dotc` do exactly the same thing.

dist/bin/dotc
liufengyun: @DarkDimius I agree there's a potential to merge the two. But I think we can delay the merge after the Copenhagen release -- as the `bin/dotc` and `bin/dotr` have something special to the development of the compiler, such as switch between bootstrap and unbootstrap versions. We can make the decision later.

dist/bin/dotd
liufengyun: @felixmulder We've a long list of dependencies if `dottydoc` is included. It doesn't look very nice. How about we do some `assembly` for `dottydoc` and do it in a separate PR?

1. ST4-4.0.7.jar
1. antlr-3.5.1.jar
1. antlr-runtime-3.5.1.jar
1. autolink-0.6.0.jar
1. flexmark-0.11.1.jar
1. flexmark-ext-anchorlink-0.11.1.jar
1. flexmark-ext-autolink-0.11.1.jar
1. flexmark-ext-emoji-0.11.1.jar
1. flexmark-ext-gfm-strikethrough-0.11.1.jar
1. flexmark-ext-gfm-tables-0.11.1.jar
1. flexmark-ext-gfm-tasklist-0.11.1.jar
1. flexmark-ext-ins-0.11.1.jar
1. flexmark-ext-superscript-0.11.1.jar
1. flexmark-ext-tables-0.11.1.jar
1. flexmark-ext-wikilink-0.11.1.jar
1. flexmark-ext-yaml-front-matter-0.11.1.jar
1. flexmark-jira-converter-0.11.1.jar
1. flexmark-util-0.11.1.jar
1. jackson-annotations-2.2.3.jar
1. jackson-core-2.8.6.jar
1. jackson-databind-2.2.3.jar
1. jackson-dataformat-yaml-2.8.6.jar
1. jsoup-1.7.2.jar
1. liqp-0.6.7.jar
1. snakeyaml-1.17.jar

liufengyun:
@DarkDimius The new scirpts are supposed to support both `mingw` and `cygwin`, the code is from `sbt-pack`, I assume it's more widely tested than the code from minibox.

Actually, I found bugs related to `bin/dotc` about how `symlink` is resolved. The code from `sbt-pack` is more reliable.

DarkDimius:
@liufengyun, ok, agreed. 
But the point about duplication stands. We shouldn't have 3+ scripts to edit.

liufengyun:
@DarkDimius I just removed `dotd`, we'll address `dottydoc` in another PR.

Regarding `dotr` and `dotc`, I don't see any duplication. If you mean the initial code for locating `PROG_HOME`,  `bin/dotc` and `bin/dotr` also do the same for `DOTTY_ROOT` -- that piece of code cannot be shared, as sharing requires locating `PROG_HOME`/`DOTTY_ROOT` first.

dist/bin/dotd
felixmulder: :+1:

dist/bin/dotd
DarkDimius: By multiple entry points I've meant that you had similar scripts that differ only in what class they invoke in the end.

liufengyun:
I've published a test release in `lampepfl/homebrew-brew`. The standard (pre-)releases will come from @felixmulder in `lampepfl/dotty` after this PR is merged.

Now programmers can do the following:

- Mac OS: `brew install lampepfl/brew/dotty`, and play with `dotc` and `dotr`.
- Linux/Windows: download the zip [here](https://github.com/lampepfl/homebrew-brew/releases/download/v0.1-M1/dotty-0.1.1-bin-SNAPSHOT.zip), unzip and run `bin/dotc` and `bin/dotr`.

It will be great that you can try the test release and give feedback, thanks a lot.

DarkDimius:
@liufengyun, just tried it on my mac. Works good so far.

DarkDimius:
#2531 and #2529 contains a bunch of fixes that we should get in before release.
Otherwise we'll have hard time figuring what version has a bug that user is experiencing.

liufengyun:
@DarkDimius I updated the test release with your fixes, it works well 👍

dist/bin/dotc
```diff
@@ -0,0 +1,116 @@
...
+    -Oshort) addJava "-XX:+TieredCompilation -XX:TieredStopAtLevel=1" && shift ;;
+        -repl) PROG_NAME="$ReplMain" && shift ;;
+       -tasty) PROG_NAME="$FromTasty" && shift ;;
+     -compile) PROG_NAME="$CompilerMain" && shift ;;
+         -run) PROG_NAME="$ReplMain" && shift ;;
+      -bootcp) bootcp=true && shift ;;
+   -nobootcp) unset bootcp && shift ;;
+      -colors) colors=true && shift ;;
+   -no-colors) unset colors && shift ;;
+      -jrebel) jrebel=true && shift ;;
```
felixmulder: Is this option being used?

dist/bin/dotc
```diff
@@ -0,0 +1,116 @@
...
+   -no-jrebel) unset jrebel && shift ;;
```
felixmulder: This one?

liufengyun: @felixmulder I get them from the `bin/dotc` and `bin/dotr`. It seems it's fine to remove them. Doing it now.

dist/bin/dotc
```diff
@@ -89,10 +89,6 @@ case "$1" in
    -nobootcp) unset bootcp && shift ;;
       -colors) colors=true && shift ;;
    -no-colors) unset colors && shift ;;
-      -jrebel) jrebel=true && shift ;;
-   -no-jrebel) unset jrebel && shift ;;
-
-      -toolcp) require_arg classpath "$1" "$2" && toolcp="$2" && shift 2 ;;
```
liufengyun: We are currently not supporting `toolcp` and it's not properly passed in `bin/dotc`. So I remove it for now, and add it back when we have `toolcp` support.

felixmulder:
Failure is due to bintray resolution?

liufengyun:
Yes, I restarted it -- the `sbt-gpg` plugin resolution failed.

felixmulder:
Could you add the artifactory resolver to the build from Fabien so that we get the caching?

liufengyun:
@felixmulder The CI passes -- I didn't get the latest URL of the resolver, we do it separately?

olafurpg:
FWIW, here's how I enabled the artifactory resolver only for CI https://github.com/scalacenter/scalafix/pull/165/commits/1c021f8d02ce7c0445a022fd13b9e2423fa4b371#diff-fdc3abdfd754eeb24090dbd90aeec2ceR87

olafurpg:
But that does not enable the artifactory for resolving plugins, I would recommend instead to enable the artifactory globally in `~/.sbt/repositories` in the docker image.

felixmulder:
@liufengyun - yes, it'd be good to get the patch into the docker image as well, but that's of lower priority right now.

felixmulder:
Let's decide at the Monday meeting what to do. Thank for the info, looks great :)

PR #2525, opened 2017-05-24
Disallow Phantom types in casts

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/CheckPhantomCast.scala
```diff
@@ -0,0 +1,46 @@
...
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+import Types._
+import Contexts.Context
+import Symbols._
+import Decorators._
+import dotty.tools.dotc.ast.Trees._
+import dotty.tools.dotc.ast.tpd
+
+
+/** A no-op transform that checks whether the compiled sources have no Phantom types in casts */
```
OlivierBlanvillain: A no-op transform to ensure that the compiled sources have...

compiler/src/dotty/tools/dotc/transform/CheckPhantomCast.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def transformBind(tree: tpd.Bind)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
+    tree.body match {
+      case Typed(_, tpt) => checkNoPhantoms(tpt)
+      case _ =>
+    }
+    tree
+  }
+
+  private def checkNoPhantoms(tpTree: tpd.Tree)(implicit ctx: Context): Unit = {
+    val checker = new TypeMap {
```
OlivierBlanvillain: What not a `TypeTraverser`?

nicolasstucki: Em... will change it.

compiler/src/dotty/tools/dotc/transform/CheckPhantomCast.scala
```diff
@@ -0,0 +1,46 @@
...
+import Symbols._
+import Decorators._
+import dotty.tools.dotc.ast.Trees._
+import dotty.tools.dotc.ast.tpd
+
+
+/** A no-op transform that checks whether the compiled sources have no Phantom types in casts */
+class CheckPhantomCast extends MiniPhaseTransform { thisTransformer =>
+
+  override def phaseName = "checkPhantomCast"
```
DarkDimius: add a post-condition that those casts don't reappear so that Ycheck will enforce it?

compiler/src/dotty/tools/dotc/transform/CheckPhantomCast.scala
DarkDimius: +1

Issue #2468, opened 2017-05-18
`private[this] var x: String = _` crashes with "key not found: class String" in the backend


smarter:
```scala
class A {
  private[this] var x: String = _
}
```
```scala
java.util.NoSuchElementException: key not found: class String
        at scala.collection.MapLike$class.default(MapLike.scala:228)
        at scala.collection.AbstractMap.default(Map.scala:59)
        at scala.collection.mutable.HashMap.apply(HashMap.scala:65)
        at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder$locals$.load(BCodeSkelBuilder.scala:402)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:391)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:290)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genStat(BCodeBodyBuilder.scala:89)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder$$anonfun$genNormalBlock$1$1.apply(BCodeBodyBuilder.scala:860)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder$$anonfun$genNormalBlock$1$1.apply(BCodeBodyBuilder.scala:860)
        at scala.collection.immutable.List.foreach(List.scala:392)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genNormalBlock$1(BCodeBodyBuilder.scala:860)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genBlock(BCodeBodyBuilder.scala:892)
        at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:408)
        at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.emitNormalMethodBody$1(BCodeSkelBuilder.scala:606)
        at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.genDefDef(BCodeSkelBuilder.scala:640)
[...]

Issue #510, opened 2015-04-30
Memory analisys. Not a bug but a discussion and a record of data


DarkDimius:
Interesting moments that I want to share:

[Definition of retained size](https://www.yourkit.com/docs/java/help/sizes.jsp)

After compiling dotty by dotty:
Total size of input source files: 3M 
Total retained size of InitialContext: 25M
Total retained size of `ContextBase` object: 70M, amoug those:
- dotty.tools.dotc.core.Uniques$RefinedUniques: 13M
- dotty.tools.dotc.core.Contexts$ContextState$$anon$1 12M
- dotty.tools.dotc.core.Uniques$TypeAliasUniques 5M

Average size of compilation unit: 2.5M <-- will fix this in next commit.

Total size of name table: 10M

Objects allocated in phase: 
- 48% of allocated objects are allocated in TreeChecker(I ran with "-Ycheck:tailrec,resolveSuper,mixin,restoreScopes"). But they constitute only 4% of total allocated size.
- 25% by all miniphases in total. Hard to get which mini-phase ones actually did the allocation, as many are allocated by the framework itself.
- 16% in Typer;
- 3% in Erasure;
- 2% in parser;
- <1% in namer;
- <1% in GenBCode

Trees allocated by frontend(parser + typer):

| Node | % of all trees |
| --- | --- |
| dotty.tools.dotc.ast.Trees$Ident | 0.3146870081 |
| dotty.tools.dotc.ast.Trees$Apply | 0.2194938855 |
| dotty.tools.dotc.ast.Trees$Select | 0.1940912943 |
| dotty.tools.dotc.ast.Trees$TypeTree | 0.1101586148 |
| dotty.tools.dotc.ast.Trees$ValDef | 0.0763772854 |
| dotty.tools.dotc.ast.Trees$Modifiers | 0.0691124834 |
| dotty.tools.dotc.ast.Trees$This | 0.0564959438 |
| dotty.tools.dotc.ast.Trees$Block | 0.046567381 |
| dotty.tools.dotc.ast.Trees$DefDef | 0.0409008355 |
| dotty.tools.dotc.ast.Trees$Literal | 0.0294950963 |
| dotty.tools.dotc.ast.Trees$CaseDef | 0.0217459741 |
| dotty.tools.dotc.ast.Trees$Typed | 0.017798765 |
| dotty.tools.dotc.ast.Trees$If | 0.0163942366 |
[...]

PR #2528, opened 2017-05-24
Fix #2456: eliminate syntactic references to Java packages

### LINKED ISSUE #2456: Teach Ycheck to spot refernces to Java packages
Ycheck should have complained here: https://github.com/lampepfl/dotty/issues/2396#issuecomment-302088399

but it didn't.

abeln:


smarter:
> java.util.NoSuchElementException: key not found: class Int

That looks exactly like https://github.com/lampepfl/dotty/issues/2468

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
```diff
@@ -0,0 +1,24 @@
+package dotty.tools.dotc.transform
```
abeln: @DarkDimius friendly ping

DarkDimius: It's funny how now with the new github system we can have comments with wrong review dates. See reply in a comment above.

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def transformSelect(tree: Select)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    if (isJavaPackage(tree)) {
+      assert(tree.tpe.isInstanceOf[TypeRef], s"Expected tree with type TypeRef, but got ${tree.tpe.show}")
+      Ident(tree.tpe.asInstanceOf[TypeRef])
+    } else {
+      tree
+    }
+  }
+
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = {
```
abeln: @DarkDimius I added the postcondition, but is there a way to run all the tests with -Ycheck enabled? (Is that done by default in the CI?)

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
```diff
@@ -0,0 +1,46 @@
...
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Flags._
+import dotty.tools.dotc.core.Types.{Type, TypeRef}
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/**
+  * Eliminates syntactic references to Java packages, so that there's no chance
+  * they accidentally end up in the backend.
+  */
+class ElimJavaPackages extends MiniPhaseTransform {
```
abeln: Should I be extending DenotTransformer here? I'd think InfoTransformer doesn't apply, because the type of the tree is unchanged. But maybe IdentityDenotTransformer then?

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = {
+    assert(!isJavaPackage(tree), s"Unexpected reference to Java package in ${tree.show}")
+  }
+
+  /**
+    * Is the given tree a syntactic reference to a Java package?
+    */
+  private def isJavaPackage(tree: Tree)(implicit ctx: Context): Boolean = {
```
nicolasstucki: If you do `private def isJavaPackage(tree: Select)...` you will not need the first `match`.

nicolasstucki: You should do it in `checkPostCondition` where performance is not really needed.

abeln: Done. PTAL.

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
```diff
@@ -0,0 +1,46 @@
...
+  * Eliminates syntactic references to Java packages, so that there's no chance
+  * they accidentally end up in the backend.
+  */
+class ElimJavaPackages extends MiniPhaseTransform {
+
+  override def phaseName: String = "elimJavaPackages"
+
+  override def transformSelect(tree: Select)(implicit ctx: Context, info: TransformerInfo): Tree = {
+    if (isJavaPackage(tree)) {
+      assert(tree.tpe.isInstanceOf[TypeRef], s"Expected tree with type TypeRef, but got ${tree.tpe.show}")
```
abeln: How performance-sensitive is the miniphase code? Should I get rid of the assertion (and the asInstanceOf call below) in favour of *some* code duplication?

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
DarkDimius: > InfoTransformer doesn't apply, because the type of the tree

DenotTransformer are required when you change "type of a name", i.e. Denotation.
E.g. if a method/field used to return one type, but now returns another. Or if it takes more agruments. Or if a field became a method. Or if new methods\fields\types\classes are introduced.

You don't need to extend DenotTransformer here if are rebuilding trees that won't affect typing of "names". This phase doesn't.

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
DarkDimius: Selects are one the the most common nodes, so transformSelects should be fast:
https://github.com/lampepfl/dotty/issues/510

isInstanceOf & asInstanceOf checks are very fast on HotSpot, the slower part is allocation of a lambda that is the second argument of assert.

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
DarkDimius: Both CI and local test suite runs by default with Ychecks enabled. 
https://github.com/lampepfl/dotty/blob/master/compiler/test/dotc/tests.scala#L71

compiler/src/dotty/tools/dotc/transform/ElimJavaPackages.scala
abeln: Ok, so should I get rid of the assert, or are we good to go here?

PR #2529, opened 2017-05-25
Report that we are not scalac but dotc in help output

DarkDimius:


liufengyun:
LGTM, thanks for the fix.

PR #2531, opened 2017-05-25
Fixes to CLI and version reporting.

DarkDimius:
We used to report a hardcoded 0.0.1 version in one place and 2.11 in other places.
Now, we report consistently the version that was used in sbt.

Additionaly, if version countains SNAPSHOT or NIGHTLY it will also report git sha.

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
+++ b/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
@@ -115,6 +115,9 @@ object CompilerCommand extends DotClass {
     else if (settings.version.value) {
       ctx.echo(versionMsg)
       Nil
+    } else if (!Properties.isJavaAtLeast("1.8")) {
+      ctx.error("Dotty requires Java 8 or newer to run")
+      Nil
     }
     else if (shouldStopWithInfo) {
       ctx.echo(infoMessage)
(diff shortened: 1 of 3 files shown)


liufengyun:
@DarkDimius Could you check the compilation failure and run CI again?

compiler/src/dotty/tools/dotc/config/Properties.scala
```diff
@@ -135,4 +135,10 @@ trait PropertiesTrait {
     }
     okVersions exists (javaVersion startsWith _)
   }
+
+  // provide a main method so version info can be obtained by running this
+  def main(args: Array[String]): Unit = {
```
smarter: This commit is called "Remove unnecessary main" but it adds the main back.

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
```diff
@@ -115,6 +115,9 @@ object CompilerCommand extends DotClass {
     else if (settings.version.value) {
       ctx.echo(versionMsg)
       Nil
+    } else if (!Properties.isJavaAtLeast("1.8")) {
```
smarter: I don't think dotty is going to work out of the box on Java 9, so we should probably make this an equality check

liufengyun: @DarkDimius The CI shows this line doesn't work.

compiler/src/dotty/tools/dotc/config/Properties.scala
DarkDimius: Yes, something got screwed during rebase. Commit messages also got off-by-one, and the implementation of java8 check simply disappeared.

DarkDimius:
re-rebased and force-pushed. Checked locally.

liufengyun:
LGTM, thanks @DarkDimius !

PR #2543, opened 2017-05-26
Fix dotd to use java classpath

felixmulder:


dist/bin/dotd
```diff
@@ -75,4 +75,10 @@ CLASS_PATH="$CLASS_PATH$PSEP$SNAKEYAML_LIB"
 CLASS_PATH="$CLASS_PATH$PSEP$ST4_LIB"
 CLASS_PATH="$CLASS_PATH$PSEP$JSOUP_LIB"
 
-eval exec "\"$JAVACMD\"" $CLASS_PATH dotty.tools.dottydoc.Main $@
+eval exec "\"$JAVACMD\"" \
+    -Dscala.usejavacp=true \
+    $CLASS_PATH \
+    dotty.tools.dottydoc.Main \
+    $@
```
liufengyun: This change looks good, in line 67, the dot(`.`) is not actually required, dottydoc doesn't dependent on custom code to run. It's automatically added if user doesn't pass an explicit `-classpath`.

smarter:
dotd really? Can't we call it dottydoc ? :)

felixmulder:
We liked the symmetry of:

- dotc
- dotr
- dotd

:)

liufengyun:
@felixmulder What about we cut a `pre-release` today and announce it in gitter channel.  So that people can try on different platforms and we have enough time to fix problems if there are any.

felixmulder:
@liufengyun - there's a bug with dottydoc that I'm trying to fix before that can happen.

PR #2513, opened 2017-05-23
Local optimisations

OlivierBlanvillain:
This RP implements a subset of the optimisation done by linker that can be done with local knowledge. This is a first iteration and many things still need to be polished.

@DarkDimius your last 3 commits didn't completely solve the issue with `testPickling`, after cherry-picking them on top of the rebased branch still results in `<notype>` popping up in unpickling. To be investigated tomorrow :smile:

PR #2545, opened 2017-05-26
[WIP] Check if opts succeed in CI

DarkDimius:
If they do, I'll force push to PR by @OlivierBlanvillain


OlivierBlanvillain:
@DarkDimius Looks green to me, the failure comes from the bytecode tests I've added (and never tried to run bootstrapped), it's looks like a test infrastructure thing:

```scala
[error] Test dotty.tools.dotc.SimplifyPosTests.initializationError failed: java.lang.Exception: No runnable methods, took 0.002 sec
[error] Test dotty.tools.dotc.SimplifyNegTests.initializationError failed: java.lang.Exception: No runnable methods, took 0.0 sec
[error] Test dotty.tools.dotc.SimplifyEquivalences.initializationError failed: java.lang.Exception: Test class should have exactly one public constructor, took 0.0 sec
```

DarkDimius:
@OlivierBlanvillain could you please modify the bytecode tests to make sure they pass?

smarter:
`SimplifyNegTests` extends a trait with `@Test` methods, I wonder if this is related to https://github.com/scala-js/scala-2.12-junit-mixin-plugin

smarter:
@DarkDimius Does dotty emit mixin forwarders?

DarkDimius:
@smarter, it’s likely that your suggestion is the right one.
We don’t emit forwarders if default method dispatch will yield the 
same result.

smarter:
Doesn't Scala 2.12 always emit forwarders for performance reasons now?

DarkDimius:
I'm working on fixing Junit4 in Dotty

OlivierBlanvillain:
@DarkDimius I have a fix for the SimplifyTests, it it ok for to test it by force pushing this branch?

DarkDimius:
@OlivierBlanvillain, yes please force push. I'll still fix JUnit 4 in separate branch.

OlivierBlanvillain:
@DarkDimius it passed the CI! http://dotty-ci.epfl.ch/lampepfl/dotty/2614 I've pushed it to #2513.

PR #2546, opened 2017-05-26
Fix reading of jar resource files for dottydoc

felixmulder:
This was the bug that didn't work with the released version. Since the codec was not specified, it assumed ISO-8859-1 :(

Now it is correctly UTF8 (master race)

### DIFF
--- a/doc-tool/src/dotty/tools/dottydoc/staticsite/ResourceFinder.scala
+++ b/doc-tool/src/dotty/tools/dottydoc/staticsite/ResourceFinder.scala
@@ -9,8 +9,8 @@ trait ResourceFinder {
   final case class ResourceNotFoundException(message: String) extends Exception(message)
 
   protected def getResource(r: String): String =
-    Option(getClass.getResourceAsStream(r)).map(scala.io.Source.fromInputStream)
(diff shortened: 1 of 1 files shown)


doc-tool/src/dotty/tools/dottydoc/staticsite/ResourceFinder.scala
```diff
@@ -9,8 +9,8 @@ trait ResourceFinder {
   final case class ResourceNotFoundException(message: String) extends Exception(message)
 
   protected def getResource(r: String): String =
-    Option(getClass.getResourceAsStream(r)).map(scala.io.Source.fromInputStream)
+    Option(getClass.getResourceAsStream(r))
+      .map(scala.io.Source.fromInputStream(_)(scala.io.Codec.UTF8))
```
DarkDimius: don't hardcode it. Use `Properties.sourceEncoding` that all the other compiler parts use.

felixmulder: No, it needs to be hardcoded here. This is just reading of assets from the jar, it has nothing to do with compiling sources. This is guaranteed to always be UTF8.

PR #2550, opened 2017-05-27
Fix scripted empty-package test

smarter:
This test started failing after we started emitting .tasty files by
default in 1293e2f49b320873c58552ab86f49ef7999b91f8, because sbt will
not by default delete these .tasty files when the corresponding
.class file is deleted, which means that the directories containing
these files are not deleted either. This is problematic because dotty
creates package symbols based on the name of directories on the
classpath, even if the directories contain no classfile. Thankfully, I
already fixed this issue in sbt-dotty yesterday in
452bc5e9645021df496e99b824c7817af92a28ff, so all we need to do is make
the scripted tests use the sbt-dotty plugin. This commit does this, and
also move these tests to the sbt-dotty project since they depend on it now.


smarter:
Merging now so that the nightly tests start working again.

felixmulder:
Great job! Thanks for doing this on a weekend!

PR #2552, opened 2017-05-28
Add Enum Eq

odersky:
Add an Eq instance to every Enum so that they can be compared only to other values of the same type.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -72,9 +72,12 @@ object desugar {
             val defctx = ctx.outersIterator.dropWhile(_.scope eq ctx.scope).next
             var local = defctx.denotNamed(tp.name).suchThat(_ is ParamOrAccessor).symbol
             if (local.exists) (defctx.owner.thisType select local).dealias
-            else throw new java.lang.Error(
-              s"no matching symbol for ${tp.symbol.showLocated} in ${defctx.owner} / ${defctx.effectiveScope}"
-            )
+            else {
+              def msg =
+                s"no matching symbol for ${tp.symbol.showLocated} in ${defctx.owner} / ${defctx.effectiveScope}"
+              if (ctx.reporter.errorsReported) new ErrorType(msg)
+              else throw new java.lang.Error(msg)
+            }
           case _ =>
             mapOver(tp)
         }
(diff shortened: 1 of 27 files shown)


compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -411,6 +415,31 @@ object desugar {
...
+    //    enum class C[T1, ..., Tn]
+    //
+    // we generate:
+    //
+    //    implicit def eqInstance[T1$1, ..., Tn$1, T1$2, ..., Tn$2](implicit
+    //      ev1: Eq[T1$1, T1$2], ..., evn: Eq[Tn$1, Tn$2]])
+    //      : Eq[C[T1$1, ..., Tn$1], C[T1$2, ..., Tn$2]] = Eq
+    def eqInstance = {
+      def append(tdef: TypeDef, str: String) = cpy.TypeDef(tdef)(name = tdef.name ++ str)
+      val leftParams = derivedTparams.map(append(_, "$1"))
```
smarter: Any reason to not use semantic names here?

odersky: Yes, we could do that.

odersky: On the other hand, maybe it's not worth the hassle. The names literally don't matter.

docs/docs/reference/multiversal-equality.md
```diff
@@ -0,0 +1,97 @@
...
+a value `y` has type `S` instead of the expected type `T`.
+
+    val x = ... // of type T
+    val y = ... // of type S, but should be T
+    x == y      // typechecks, will always yield false
+
+If all you do with `y` is compare it to other values of type `T`, the program will
+typecheck but probably give unexpected results.
+
+Multiversal equaliy is an opt-in way to make universal equality
```
xeno-by: Typo: `equality`

docs/docs/reference/named-typeargs.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: doc-page
+title: "Named Type Arguments"
+---
```
xeno-by: It would also be great to have a page about implicit function types.

docs/docs/reference/multiversal-equality.md
```diff
@@ -0,0 +1,97 @@
...
+    implicit def eqAB: Eq[A, B] = Eq
+    implicit def eqBA: Eq[B, A] = Eq
+
+(As usual, the names of the implicit definitions don't matter, we have
+chosen `eqA`, ..., `eqBA` only for illustration).
+
+The `dotty.DottyPredef` object defines a number of `Eq`
+implicits. `dotty.DottyPredef` is a temporary `Predef`-like object.
+The contents of this object are by default imported into every
+program. Once dotty becomes standard Scala, `DottyPredef` will go away
```
odersky: Ok, let's settle on capitalizing it

docs/docs/reference/named-typeargs.md
odersky: Coming up...

odersky:
@senia-psm Fixing it, thanks!

PR #2553, opened 2017-05-28
Fix #2473: Check self for inner class redefinitions

### LINKED ISSUE #2473: Type member referencing self-typed inner class with same name causes stack overflow
Attempting to compile this code causes the stack overflow:
```scala
trait Foo { trait Inner }
trait Bar { foo: Foo => type Inner <: foo.Inner }
```
It appears that renaming one of the `Inner`s avoids the compiler crash.

Here's the [stack trace](https://github.com/lampepfl/dotty/files/1013682/failure.txt).

nicolasstucki:


tests/neg/i2473.scala
```diff
@@ -0,0 +1,7 @@
+trait Foo { trait Inner }
+trait Bar { foo: Foo =>
+  type Inner <: foo.Inner // error: type Inner cannot have the same name as trait Inner in trait Foo -- class definitions cannot be overridden
```
smarter: But it's not an override, it's a shadowing, right?

nicolasstucki: Oh, yes. I reused the old error message. I will improve it to make the distinction.

smarter: In fact, I don't think that shadowing is the right term either, consider:
```scala
trait Foo {
  val x: Int = 1
}

trait Bar { foo: Foo =>
  val x: String = ""
}
```
In both scalac and dotty, this code compiles, but if you try to make an instance of `Bar`, it fails:
```scala
object Test {
    new Foo with Bar {}
  }
}
```

```scala
-- Error: try/i2473.scala:14:4 -------------------------------------------------
14 |    new Foo with Bar {}
   |    ^
   |   class $anon inherits conflicting members:
   |     value x in trait Foo of type Int  and
   |     value x in trait Bar of type String
   |   (Note: this can be resolved by declaring an override in class $anon.)
```

nicolasstucki: I changed the message to print 
```scala
-- [E068] Syntax Error: tests/neg/i2473.scala:3:7 ------------------------------
3 |  type Inner <: foo.Inner // error
  |       ^
  |type Inner cannot have the same name as trait Inner in trait Foo -- cannot define type member with the same name as a trait member in self reference foo.
  |(Note: this can be resolved by using another name)
-- [E068] Syntax Error: tests/neg/i2473.scala:6:8 ------------------------------
6 |  class Inner // error
  |        ^
  |class Inner cannot have the same name as trait Inner in trait Foo -- cannot define class member with the same name as a trait member in self reference baz.
  |(Note: this can be resolved by using another name)
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1472,12 +1472,26 @@ object messages {
           |"""
   }
 
-  case class CannotHaveSameNameAs(sym: Symbol, cls: Symbol)(implicit ctx: Context)
+  case class CannotHaveSameNameAs(sym: Symbol, cls: Symbol, reason: CannotHaveSameNameAs.Reason)(implicit ctx: Context)
     extends Message(CannotHaveSameNameAsID) {
-    val msg = hl"""$sym cannot have the same name as ${cls.showLocated} -- class definitions cannot be overridden"""
+    import CannotHaveSameNameAs._
+    def resonMessage: String = reason match {
```
smarter: typo: reson -> reason

PR #2555, opened 2017-05-28
Fix 2551: shortcut staticRef for packages to avoid cycles

### LINKED ISSUE #2551: Spurious cyclic reference error with package objects and separate compilation
```scala
// Library.scala
package scala {
  package meta {
    package config {
      case class Version()

      trait Aliases {
        type Version = scala.meta.config.Version
        val Version = scala.meta.config.Version
      }
    }
  }

  package object meta extends scala.meta.config.Aliases
}
```

```scala
// Test.scala
import scala.meta._

object Test {
  val v = Version()
}
```

```
17:46 ~/Projects/dotty/sandbox $ ../bin/dotc Library.scala
17:46 ~/Projects/dotty/sandbox $ ../bin/dotc Test.scala
exception caught when loading class Aliases: dotty.tools.dotc.core.Types$CyclicReference: cyclic reference involving class Aliases
-- [E044] Syntax Error: Test.scala:4:10 ----------------------------------------
4 |  val v = Version()
  |          ^^^^^^^
  |          recursive value v needs type

one error found
```

Notes:
  * When Library.scala is compiled with scalac, we get the same error (so it's not a ScalaSignature vs TASTY problem). upd: I was wrong - when Library.scala is compiled with scalac, compilation succeeds.
  * When both Library.scala and Test.scala are compiled with scalac, compilation succeeds.
  * When there's just one file that merges these two files together, compilation succeeds.

liufengyun:
Attempt to fix #2551: shortcut staticRef for packages to avoid cycles

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -1186,11 +1186,11 @@ object Denotations {
      *  or a MissingRef or NoQualifyingRef instance, if it does not exist.
      *  if generateStubs is set, generates stubs for missing top-level symbols
      */
-    def staticRef(path: Name, generateStubs: Boolean = true)(implicit ctx: Context): Denotation = {
+    def staticRef(path: Name, generateStubs: Boolean = true, isPackage: Boolean = false)(implicit ctx: Context): Denotation = {
       def select(prefix: Denotation, selector: Name): Denotation = {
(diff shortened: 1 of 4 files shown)


xeno-by:
Thank you, guys! When will there be a nightly that includes the fix?

DarkDimius:
@xeno-by, it should be released this night.
We're also going to cut a RC soon. This will be in.

PR #2565, opened 2017-05-29
Pattern matching documentation

OlivierBlanvillain:
@odersky @DarkDimius This is very minimal, anything else you would like to see included?

### DIFF
--- a/docs/_includes/features.html
+++ b/docs/_includes/features.html
@@ -9,11 +9,7 @@
                 </colgroup>
                 <tbody>
                     <tr class="odd">
-                        <td>Union, intersection and <a href="http://docs.scala-lang.org/sips/pending/42.type.html">literal singleton types</a></td>
-                        <td>Implemented</td>
-                    </tr>
-                    <tr class="even">
-                        <td>Fast compilation (phase fusion)</td>
+                        <td><a href="http://dotty.epfl.ch/docs/reference/union-types.html">Union</a>, <a href="http://dotty.epfl.ch/docs/reference/intersection-types.html">intersection</a> and <a href="http://docs.scala-lang.org/sips/pending/42.type.html">literal singleton types</a></td>
                         <td>Implemented</td>
                     </tr>
                     <tr class="odd">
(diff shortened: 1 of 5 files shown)


docs/docs/reference/pattern-matching.md
```diff
@@ -0,0 +1,52 @@
...
+becomes
+
+```scala
+class CC(val i: Int, val s: String) extends Product { ... }
+
+object {
+  def apply(i: Int, s: String): CC = new CC(i, s)
+  def unapply(cc: CC): CC = cc
+}
+```
```
felixmulder: This page needs to be added to `sidebar.yml` and `index.md` to be visible to all :)

smarter:
Does this describe the current state of Dotty after https://github.com/lampepfl/dotty/commit/3e04b6f3aa4e4088220f199bd6aa5c6644c22354 ?

docs/docs/reference/pattern-matching.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: doc-page
+title: "Option-less pattern matching"
+---
+
+Dotty pattern matching supports the following [extractors](https://www.scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#extractor-patterns):
+
+Boolean Pattern
+
+- Extractor defines `def unapply(x: T): Boolean`
```
DarkDimius: I think it would be nice to give examples instead of rules. More accessible.

docs/docs/reference/pattern-matching.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Extractor defines `def unapply(x: T): U`
+- `U` has (parameterless `def` or `val`) members `isEmpty: Boolean` and `get: S`
+- If there is exactly `1` pattern, pattern-matching on `1` pattern with type `S`
+- Otherwise `N > 0` is the maximum number of consecutive (parameterless `def` or `val`) `_1: P1` ... `_N: PN` members in `U`
+- Pattern-matching on exactly `N` patterns with types `P1, P2, ..., PN`
+
+In case of ambiguities, *Product Pattern* is preferred over *Name Based Pattern*.
+
+Dotty pattern matching is not special cased to case classes. Instead, allocation of Tuples and Options is avoided by desugaring case classes to use a *Product Pattern* extractor:
```
DarkDimius: It's not clean what does this mean without context. I think we should either introduce scalac story here, or simply drop this sentence.

DarkDimius:
>  anything else you would like to see included?

I'd propose to include that Dotty-generated patterns are a LOT easier to debug, as variables will show-up in debug modes.

OlivierBlanvillain:
@DarkDimius I've incorporated your suggestions, I think the pattern matching doc needs another review.

PR #2579, opened 2017-05-29
Fix build: workaround scoping bug

smarter:


smarter:
Merging now to get the build green again, but we should come back to this scoping stuff.

PR #2580, opened 2017-05-29
Fix #2390: Use normal thisType as base for override checking

### LINKED ISSUE #2390: dotty.tools.dotc.core.Types$MissingType: cannot resolve reference to type
While upgrading the community build dotty-staging/scalatest fork to the latest dotty (0.1.1-bin-20170507-1014af3-NIGHTLY) and sbt-dotty 0.1.0-RC4 I hit an a MissingType error 
https://travis-ci.org/dotty-staging/scalatest/builds/229925482#L2104
during type erasure https://github.com/lampepfl/dotty/blob/5093126421e333b2329399736e6d7ad327e3bdfa/compiler/src/dotty/tools/dotc/core/TypeErasure.scala#L496

I made some unsuccessful attempts to minimize the error, it seems to boil down to something like this

```scala
package foo
trait TestSuite {
  protected trait NoArgTest extends (() => Int)
}
trait CancelAfterFailure { self: TestSuite => }
```

To reproduce the error:

```
git clone https://github.com/dotty-staging/scalatest.git
cd scalatest
git checkout 1b2ea9013e4f71306f6dfc5f4e1ae0db85a46629
sbt scalatest/compile "last scalatest/compile:compileIncremental"
```

odersky:
Using upwardsThisType can cause MissingType errors. It's a hairy
problem which is still not quite understood (see comment on
upwardsThisType). We currently address this by trying either
normal thisType or upwardsThisType.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -151,7 +151,8 @@ object RefChecks {
    *       before, but it looks too complicated and method bodies are far too large.
    */
   private def checkAllOverrides(clazz: Symbol)(implicit ctx: Context): Unit = {
-    val self = upwardsThisType(clazz)
+    val self = clazz.thisType
+    val upwardsSelf = upwardsThisType(clazz)
     var hasErrors = false
 
     case class MixinOverrideError(member: Symbol, msg: String)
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -245,17 +246,24 @@ object RefChecks {
           (if (otherAccess == "") "public" else "at least " + otherAccess))
       }
 
-      def compatibleTypes =
-        if (member.isType) { // intersection of bounds to refined types must be nonempty
-          member.is(BaseTypeArg) ||
-          (memberTp frozen_<:< otherTp) || {
-            val jointBounds = (memberTp.bounds & otherTp.bounds).bounds
-            jointBounds.lo frozen_<:< jointBounds.hi
+      def compatibleTypes(memberTp: Type, otherTp: Type): Boolean =
```
smarter: I suggest using other names than `memberTp` and `otherTp` here, because `memberTp` and `otherTp` are already in scope at this point and they could easily be confused when reading the code.

Issue #559, opened 2015-05-13
Rewriting tool should warn on usage on extending DelayedInit


DarkDimius:
Out of 1500 run tests that fail, 900 fail due to them extending `scala.App`, which relies on `DelayedInit`, that is not supported.

DarkDimius:
ping @samuelgruetter

densh:
@odersky is there a plan to completely remove `DelayedInit` in the future?

odersky:
Yes, we should remove delayedInit.

But given the 900 test failures I think it would be reasonable if the
rewriting tool rewrote
"extends App" instead of just issuing a warning. I.e.

    object Foo extends App { ... }

becomes

    object Foo extends App { override def main(args: Array[String]): Unit =
{ ... }}

If we have macro annotations, we could envision having syntactic sugar like

    @app object Foo { ... }

for this. App can be left as a convenience class in dotty, but it would not
have a delayedInit mechanism. More or less all it would do is make the
command line arguments available.

 - Martin




On Wed, May 13, 2015 at 1:16 PM, Denys Shabalin <notifications@github.com>
wrote:

> @odersky <https://github.com/odersky> is there a plan to completely
> remove DelayedInit in the future?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lampepfl/dotty/issues/559#issuecomment-101630704>.
>



-- 
Martin Odersky
EPFL

DarkDimius:
@odersky For tests, I've added an superclass that has a main method: https://github.com/dotty-staging/dotty/commit/28bb27672bf7e4fb2bb80df48ee8c70628de19c4 https://github.com/dotty-staging/dotty/commit/28bb27672bf7e4fb2bb80df48ee8c70628de19c4

This is almost equivalent to moving statements into the body of class, the only different is that they are actually now executed before calling main method.
This made several hundred tests pass.

densh:
`@app object Foo { ... }` looks like a really cool replacement.

DarkDimius:
Honestly, I dislike the `app` being so short. In most cases, there are have no-more than two `main` methods in their project. I'd prefer to be explicit here, and write the full `@application`, as it is easier to read.

hepin1989:
I like ```@application``` too.

odersky:
@application / @app, either is fine with me.

On Wed, May 13, 2015 at 2:32 PM, kerr <notifications@github.com> wrote:

> I like @application too.
>
> —
[...]

PR #2583, opened 2017-05-29
Add section on dropping DelayedInit

odersky:


docs/docs/reference/dropped/delayedInit.md
```diff
@@ -0,0 +1,26 @@
...
+    object HelloWorld extends App {
+      println("Hello, world!")
+    }
+
+However, the code is now run in the initializer of the object, which on
+some JVM's means that it will only be interpreted. So, better not use it
+for benchmarking! Also, if you want to access the command line arguments,
+you need to use an explicit `main` method for that.
+
+    object Hello extends App {
```
smarter: In this example, is there any point in doing `extends App`?

smarter:
See also https://github.com/lampepfl/dotty/issues/559 which lead to the creation `dotty.runtime.LegacyApp`.

docs/docs/reference/dropped/delayedInit.md
odersky: No, good catch.

odersky:
#559 does not seem to be an issue anymore, at least not for simple cases where we do not access `args`. I added a test.

olafurpg:
Thanks @smarter I opened https://github.com/scalacenter/scalafix/issues/174 to handle the cases when `args` is accessed. Currently, that case compiles but `args == null` at runtime. What do you think about removing `args` to get a compile-time error instead?

https://github.com/lampepfl/scala/blob/7ecfce1fb8d39275f082aaa3ad4dc0eee197391c/src/library/scala/App.scala#L46

smarter:
We can't really remove it because it comes from scala-library, and we don't recompile scala-library currently in dotty, we use the one from 2.11.11 as-is.

olafurpg:
I see, that makes sense @smarter

docs/docs/reference/dropped/type-projection.md
```diff
@@ -0,0 +1,15 @@
...
+title: Dropped: General Type Projection
+---
+
+Scala so far allowed general type projection `T#A` where `T` is an arbitrary type
+and `A` names a type member of `T`.
+
+Dotty allows this only if `T` is a class type. The change was made because
+unrestricted type projection is [unsound](https://github.com/lampepfl/dotty/issues/1050).
+
+The restriction rule out the [type-level encoding of compbinator
```
ShaneDelmore: /s/compbinator/combinator

olafurpg:
We just added the Scalafix `NoExtendsApp` rewrite that replaces `extends App` with an explicit `main` function, regardless if `args` is used or not https://scalacenter.github.io/scalafix/#NoExtendsApp

PR #2587, opened 2017-05-29
Fix YAML parsing, again

smarter:
Broken in https://github.com/lampepfl/dotty/pull/2583, I still don't understand why we never catch this before it's merged.


smarter:
Merging now to make the build green.

PR #2588, opened 2017-05-29
add bootstrapped distribution

liufengyun:
Now we can release bootstrapped packages by sbt:

- `dist/pack`: create non-bootstrapped release under `dist/target/pack`
- `dist/pack-archive`: create non-bootstrapped `.zip` and `.tar.gz` under `dist/target`
- `dist-boostrapped/pack`: create bootstrapped release under `dist-bootstrapped/target/pack`
- `dist-bootstrapped/pack-archive`: create bootstrapped `.zip` and `.tar.gz` under `dist-bootstrapped/target`

The scripts are shared between the two.


felixmulder:
LGTM, but @smarter is the one who is most qualified to review this since he did the bootstrap.

liufengyun:
I've created a test bootstrapped release, you can try it as follows:

```bash
brew remove dotty      # if already installed, pity I didn't change the version
brew install dotty     # use lampepfl/brew/dotty if it's the first time you install dotty
```

The [zip](https://github.com/lampepfl/homebrew-brew/releases/download/v0.1-M3/dotty-0.1.1-bin-SNAPSHOT.zip) and [tar.gz](https://github.com/lampepfl/homebrew-brew/releases/download/v0.1-M3/dotty-0.1.1-bin-SNAPSHOT.tar.gz) can be downloaded for Linux/Windows as well.

project/Build.scala
```diff
@@ -1178,4 +1178,24 @@ object Build {
...
+  lazy val `dist-bootstrapped` = project.
+    dependsOn(`dotty-interfaces`).
+    dependsOn(`dotty-library-bootstrapped`).
+    dependsOn(`dotty-compiler-bootstrapped`).
+    dependsOn(`dotty-doc-bootstrapped`).
+    settings(commonBootstrappedSettings).
+    settings(packSettings).
+    settings(
+      target := baseDirectory.value / "target",                    // override setting in commonBootstrappedSettings
+      triggeredMessage in ThisBuild := Watched.clearWhenTriggered,
```
smarter: Why is this here? If you want this, you can put it in your local sbt settings

smarter:
I'm going to say lgtm but I have no idea what this does, and I don't really want to know.

project/Build.scala
liufengyun: It's not really needed, it's copied from the setting for `dotty`, I'll remove it.

liufengyun:
Thanks @smarter , maybe you could also help check if the jars in the `tar.gz` makes sense and whether it works on your Linux. The `tar.gz` is generated from `sbt dist-bootstrapped/pack-archive`.

The particular question I have is about: whether we need `scala-library-bootstrapped` JAR in the package? I understand it's only a dummy project to automatically get dependency on `dotty-library` -- which is useful for SBT projects, but not for the command line release.

smarter:
> Thanks @smarter , maybe you could also help check if the jars in the tar.gz makes sense and whether it works on your Linux. The tar.gz is generated from sbt dist-bootstrapped/pack-archive.

Seems to be correct. At some point we need to stop using the bootclasspath for anything (it won't work with Java 9), but for now it's ok.

> The particular question I have is about: whether we need scala-library-bootstrapped JAR in the package? I understand it's only a dummy project to automatically get dependency on dotty-library -- which is useful for SBT projects, but not for the command line release.

I assume you mean the `scala-library` jar, as you correctly explained, it shouldn't be necessary for a command line release.

liufengyun:
Thanks a lot for the clarification @smarter , I merge this PR then.

PR #2590, opened 2017-05-29
Fix #2369: Should warn when ignoring command line flags

### LINKED ISSUE #2369: Should warn when ignoring command line flags.
Currently if we do `dotc -Xprint:foo -Xprint:bar -classpath pathA -classpath pathB A.scala` we ignore silently `-Xprint:foo` and `-classpath pathA`. This is true in general for all flags. We should warn or fail when we have one of these inputs.

typeness:


DarkDimius:
@typeness, your current implementation fails compilation in case flag is passed twice. And this is why the tests failed. Did you intend to use `warn`  instead?

typeness:
Any feedback?

PR #2603, opened 2017-05-30
Fix #2591: Incorrect documentation on general type projections

### LINKED ISSUE #2591: Incorrect documentation on general type projections
https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/type-projection.md says:
> Dotty allows this only if T is a class type.

Which is incorrect, T can be a type alias too, it just cannot be an abstract type. It then says:
>  It also rules out the previous encodings of type lambdas using structural types with projection as application

But the type lambda encoding is fine as long as you use type aliases, not abstract types, this works in dotty:
```scala
scala> class Foo[M[_]] 
defined class Foo
scala> val x: Foo[({type L[X] = List[X]})#L] = null 
val x: Foo[[X] => scala.collection.immutable.List[X]] = null
```

smarter:


docs/docs/reference/dropped/type-projection.md
```diff
@@ -6,10 +6,9 @@ title: Dropped: General Type Projection
 Scala so far allowed general type projection `T#A` where `T` is an arbitrary type
 and `A` names a type member of `T`.
 
-Dotty allows this only if `T` is a class type. The change was made because
-unrestricted type projection is [unsound](https://github.com/lampepfl/dotty/issues/1050).
+Dotty disallows this if `T` is an abstract type (class types and type aliases
+are fine). This change was made because unrestricted type projection
+is [unsound](https://github.com/lampepfl/dotty/issues/1050).
 
-The restriction rule out the [type-level encoding of compbinator
```
odersky: I'd leave the "restructions rule out" part in. Otherwise LGTM.

PR #2605, opened 2017-05-30
Documentation cleanup

smarter:


docs/docs/index.md
```diff
@@ -10,38 +10,3 @@ The theory behind these constructors is researched in DOT, a calculus for depend
...
-    - [Project Structure](internals/overall-structure.md): of the project
-    - [Backend](internals/backend.md): details on the bytecode backend
-    - [Contexts](internals/contexts.md): details the use of `Context` in the compiler
-    - [Dotty vs Scala 2](internals/dotc-scalac.md): explaining the differences between Scalac and Dotty
-    - [Higher Kinded Type Scheme](internals/higher-kinded-v2.md): scheme
-    - [Periods](internals/periods.md): details on the concepts of runs, phases and periods
-    - [Type System](internals/type-system.md): details on the type system of Dotty
-* Resources
-    - [Talks](resources/talks.md): presentations of the various aspects of Dotty
-
```
felixmulder: Add an optional field to `Title` named "desc" and then put these descriptions in the `sidebar.yml` so we don't lose them.

I'll fix the index generation after the flight.

Otherwise LGTM

smarter: I think most of the titles are useless anyway, and we can come up with better ones, I'll let you decide if some are worth keeping :).

felixmulder: this case class exists in "DefaultParams.scala" in the doc-tool project

felixmulder: Ok, fine - merge as is

PR #2613, opened 2017-05-30
Fix #2560: Regression in #2465: blog pages get wrong date

### LINKED ISSUE #2560: Regression in #2465: blog pages get wrong date
With the merge of timestamps, this broke something in the file parsing logic (haven't looked too close at it). But the result is:

screen shot 2017-05-29 at 11 21 41

All files now get the generated page's timestamp instead of the filename's.

cc: @Varunram

Varunram:
Intended as a fix for #2560. CC: @felixmulder @nicolasstucki 
P.S - removed some empty lines at the end of the file

### DIFF
--- a/docs/blog/_posts/2015-10-23-dotty-compiler-bootstraps.md
+++ b/docs/blog/_posts/2015-10-23-dotty-compiler-bootstraps.md
@@ -4,7 +4,7 @@ author: Martin Odersky and Dmitry Petrashko
 title: "We got liftoff!"
 subTitle: The Dotty compiler for Scala bootstraps.
 excerpt_separator: <!--more-->
-date: 2015-10-23
+date: 2015-10-23 00:00:00
 ---
 
 The [Dotty project](https://github.com/lampepfl/dotty)
--- a/doc-tool/resources/_layouts/main.html
+++ b/doc-tool/resources/_layouts/main.html
@@ -1,4 +1,4 @@
-<!-- This page was last generated on {{ page.date | date: "%b-%d-%Y" }} -->
+<!-- This page was last generated on {{ page.date | date: "%Y-%B-%d %H:%M" }} -->
 <!DOCTYPE html>
 <html lang="en">
     <head>
--- a/docs/blog/_posts/2016-01-02-new-year-resolutions.md
+++ b/docs/blog/_posts/2016-01-02-new-year-resolutions.md
@@ -3,7 +3,7 @@ layout: blog-page
 title: New Year Resolutions
 author: Martin Odersky
 authorImg: /images/martin.jpg
-date: 2016-01-02
+date: 2016-01-02 00:00:00
 ---
 
 For most of us, the change of the year is an occasion for thinking
@@ -60,7 +60,3 @@ simpler: The language, its foundations, its libraries. I hope you
 will join me in that venture.
 
 With that thought, I wish you a happy new year 2016.
-
-
-
-
--- a/docs/blog/_posts/2016-02-03-essence-of-scala.md
+++ b/docs/blog/_posts/2016-02-03-essence-of-scala.md
@@ -3,7 +3,7 @@ layout: blog-page
 title: The Essence of Scala
 author: Martin Odersky
 authorImg: /images/martin.jpg
-date: 2016-02-03
+date: 2016-02-03 00:00:00
 ---
 
 What do you get if you boil Scala on a slow flame and wait until all
@@ -143,4 +143,3 @@ project are important.
     either to increase our confidence that they are indeed sound, or
     to show that they are unsound. In my next blog I will
     present some of the issues we have discovered through that exercise.
-
--- a/docs/blog/_posts/2016-02-17-scaling-dot-soundness.md
+++ b/docs/blog/_posts/2016-02-17-scaling-dot-soundness.md
@@ -3,7 +3,7 @@ layout: blog-page
 title: Scaling DOT to Scala - Soundness
 author: Martin Odersky
 authorImg: /images/martin.jpg
-date: 2016-02-17
+date: 2016-02-17 00:00:00
 ---
 
 In my [last
@@ -150,10 +150,3 @@ should make sure not to back-slide. And if the experience of the past
 10 years has taught us one thing, it is that the meta theory of type
 systems has many more surprises in store than one might think. That's
 why mechanical proofs are essential.
-
-
-
-
-
-
-
--- a/docs/blog/_posts/2016-05-05-multiversal-equality.md
+++ b/docs/blog/_posts/2016-05-05-multiversal-equality.md
@@ -3,7 +3,7 @@ layout: blog-page
 title: Multiversal Equality for Scala
 author: Martin Odersky
 authorImg: /images/martin.jpg
-date: 2016-05-05
+date: 2016-05-05 00:00:00
 ---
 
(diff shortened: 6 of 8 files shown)


felixmulder:
This is the wrong fix :)

Varunram:
Wrong as in?

felixmulder:
See the original issue where Nicolas left a hint

docs/blog/_posts/2015-10-23-dotty-compiler-bootstraps.md
```diff
@@ -3,6 +3,7 @@ layout: blog-page
 author: Martin Odersky and Dmitry Petrashko
 title: "We got liftoff!"
 subTitle: The Dotty compiler for Scala bootstraps.
+date: 2016-10-23
```
nicolasstucki: 2015-10-23

Varunram: Done!

DarkDimius:
> How do we discern which blog post is first if they are both published on the same day?

IMHO we shouldn't publish multiple blogposts per day. We should plan it upfront.

felixmulder:
Yes, but that's a different issue altogether. It is still something the
platform has to support.

felixmulder:
Even if we don't use it!

DarkDimius:
@felixmulder, how far is this PR from getting in? 
It would be nice to have correct date on the blogpost today and have posts in the right order.

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
```diff
@@ -256,7 +256,7 @@ case class Site(
       blogposts.foreach { file =>
         val BlogPost.extract(year, month, day, name, ext) = file.getName
         val sourceFile = toSourceFile(file)
-        val date = s"$year-$month-$day 00:00:00"
+        val date = s"$month-$day-$year"
```
nicolasstucki: Why is the date in disorder now?

Varunram: Oh, that really isn't much. Its because we have `date: "%b-%d-%Y" ` in the layouts. Do I change it?

Varunram: Change in the layout i.e

nicolasstucki: This looks like an ISO date format with the `-` which is supposed to be in little-endian (and ordered). I would prefer to follow the standard.

Varunram: Done!

Varunram:
ping @felixmulder Any further changes required?

felixmulder:
@Varunram:

> How do we discern which blog post is first if they are both published on the same day? That's what putting the date explicitly in the preamble solves in Jekyll.
>
> Are we still supporting this with the current patch?

If the answer to this question is yes, then it's good to go :)

Varunram:
@felixmulder I don't think it does. But would adding the minute and second headers be of use? Does Liquid sort based on the timestamps?

felixmulder:
@Varunram: Jekyll sorts based on the timestamp. In the `_posts` dir you can see that files are named:

```
./docs/blog/_posts
├── 2015-10-23-dotty-compiler-bootstraps.md
├── 2016-01-02-new-year-resolutions.md
├── 2016-02-03-essence-of-scala.md
├── 2016-02-17-scaling-dot-soundness.md
├── 2016-05-05-multiversal-equality.md
├── 2016-12-05-implicit-function-types.md
└── 2017-05-31-first-dotty-milestone-release.md
```

This *should* give each post a timestamp of "year-month-day 00:00:00", if you want to disambiguate this, you have to put a different time in the preamble of the file. So, in a file "2016-05-05-post-name.md" we have:

```md
---
title: Post Name
date: 2016-05-05 12:00:00
---
```

See: https://jekyllrb.com/docs/frontmatter/ (ctrl-f "date")

This is what we're trying to emulate.

Varunram:
Thanks a lot for the insight! I was confused and thought something else, so this clarifies it. I'll update the PR shortly :)

felixmulder:
Any update on this @Varunram ? :)

Varunram:
Oh no, extremely sorry! Will update the PR by tonight. Sorry again!

Varunram:
@felixmulder Gentle reminder :)

felixmulder:
Do the titles now require the `00:00:00` ending? Looks like an oversight...

Othwerise LGTM

Varunram:
@felixmulder Um, not an oversight, but rather to do with how liquid parses dates. For e.g. if we put `date: 2015-10-23 12:00:00` in the header, it gets rendered as `October 23, 2015`. This would conflict with our current definition n `Site.scala` where we display dates like `2015-10-23`. This brings forward the above (dirty) proposal or we should maybe add change the formatting in `Site.scala`. Waiting for your final call on this :) (Really sorry this PR had to take so long, thanks for your patience!)

felixmulder:
@Varunram - parsing and rendering should be two different things. We should render `Month Day, Year` but be able to parse both `YYYY-MM-DD HH:MM:SS` and `YYYY-MM-DD` IMO

No worries on this taking time on my part, it's your time after all :smile:

Varunram:
@felixmulder Hm, yeah, that's true since `YYYY-MM-DD 00:00:00` looks ugly, but what I think liquid is doing is given a time date of the above format, it automatically parsed it to `Month Day, Year`. So if we do put `yyyyy-mm-dd`, it treats this as a literal string and prints the exact thing to output, without really parsing it. In order to come over this, specifying the exploit hour time might be the only option

felixmulder:
@Varunram - what happens is that during file parsing (see `Site.scala#initFiles`) we load all files as `blogposts: Array[JFile]` which are on the correct format (`Site.scala:L332`). But as you can see, these files when written in `Site.scala#generateBlog`, the date in the YAML preamble is ignored (line 259).

So to fix this issue, really all that needs to be fixed is that specific line :)

Varunram:
screen shot 2017-07-05 at 4 46 19 pm
This is what I meant by different formats if we have the extra `HH:MM:SS` for some dates and not others. My error for not being clear earlier.

felixmulder:
Yes, so if line 259 is changed to something like:

```scala
val params = defaultParams(file, 2)
val date =
  if (params.contains("date")) parseDate(defParams.page.date) // pseudo code
  else s"$year-$month-$day 00:00:00"
```

Then disaster is avoided since the formatting will be given correctly to liqp.

Varunram:
Well, after four hours of trying, here's what I got. We can't rewrite the date given in the preamble, not through `Site.scala` or similar. Seems like specifying the date is some sort of veto power :( . For e.g., you could specify the date in the preamble and try `val date="RANDOM"`, still the date in the preamble would be displayed as is. This shows that the date doesn't really "pass" through `Site.scala` and the only option to avoid this would be to explicitly give the seconds header or specify different dates. We could do without the seconds timestamp, but due to Liqp's default parsing, it changes to a different format and would look ugly. I'm really sorry that I couldn't find a proper fix 😿. On a side note, if we do decide to explicitly define the date in the preamble, the `withDate` function is useless since it does nothing (and hence, `val date` also does nothing)

nicolasstucki:
We will need to retake this issue in the future. 
@Varunram thanks for all the effort.

PR #2632, opened 2017-05-31
Fix dead links from side bar

nicolasstucki:


docs/sidebar.yml
```diff
@@ -11,8 +11,6 @@ sidebar:
           url: docs/usage/cbt-projects.html
         - title: Dottydoc
           url: docs/usage/dottydoc.html
-        - title: Migrating
-          url: docs/usage/migrating.html
```
DarkDimius: but there is a `./docs/usage/migrating.md`

nicolasstucki: Not on master

nicolasstucki: Removed in ddd48da

DarkDimius: you're right.

nicolasstucki: @DarkDimius so we should merge?

Issue #2564, opened 2017-05-29
Inlined private constructor crashes in GenBCode 


nicolasstucki:
```scala
object Foo {
  inline def bar = new Bar
  class Bar private[Foo]()
}
```

with `Xprint` crashes with
```scala
result of tests/run/Foo.scala after TreeTransform:{moveStatic, expandPrivate, selectStatic, collectEntryPoints, collectSuperCalls, dropInlined, labelDef}:

package <empty> {
  @scala.annotation.internal.SourceFile("tests/run/Foo.scala") final module 
    class
   Foo$ extends Object { 
    def <init>(): Unit = 
      {
        super()
        ()
      }
    @Foo.bar$_inlineAccessor_$1(new Foo.Foo$Bar)() inline def bar(): Foo.Foo$Bar
       = 
    Foo.bar$_inlineAccessor_$1(new Foo.Foo$Bar)
    def bar$_inlineAccessor_$1(x$0: Foo.Foo$Bar): Foo.Foo$Bar = x$0()
  }
  class Foo$Bar extends Object { 
    def <init>(): Unit = 
      {
        super()
        ()
      }
  }
  final lazy module val Foo: Foo$ = new Foo$()
}

Unexpected New(Foo.Bar/TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Foo$)),Bar)) reached GenBCode.
  Call was genLoad(New(TypeTree[TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Foo$)),Bar)]),LFoo$Bar;)
java.lang.RuntimeException: Unexpected New(Foo.Bar/TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Foo$)),Bar)) reached GenBCode.
  Call was genLoad(New(TypeTree[TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Foo$)),Bar)]),LFoo$Bar;)
	at dotty.tools.backend.jvm.DottyBackendInterface.abort(DottyBackendInterface.scala:381)
	at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:312)
[...]

PR #2639, opened 2017-05-31
Fix #2566: Ycheck that `New` node is always enclosed in a `Select`

### LINKED ISSUE #2566: Ycheck that `New`  node is always enclosed with Select around it
You can't allocate a class in Scala without calling the constructor. 
This is a nice starter task.

abeln:
Tested:
Verified that the test case in https://github.com/lampepfl/dotty/issues/2564
triggers the assert.
Verified that the check doesn't run when we ycheck a phase before frontend.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -122,6 +122,8 @@ class TreeChecker extends Phase with SymTransformer {
     val squahsedPhase = ctx.squashed(prevPhase)
     ctx.echo(s"checking ${ctx.compilationUnit} after phase ${squahsedPhase}")
 
+    assertSelectWrapsNew(ctx.compilationUnit.tpdTree)(ctx)
+
     val checkingCtx = ctx
         .fresh
         .setMode(Mode.ImplicitsEnabled)
@@ -448,6 +450,26 @@ class TreeChecker extends Phase with SymTransformer {
         })
       tree
     }
+  }
+
+  /**
+    * Checks that `New` nodes are always wrapped inside `Select` nodes.
+    */
+  def assertSelectWrapsNew(tree: tpd.Tree)(implicit ctx: Context): Unit = {
+    (new TreeAccumulator[tpd.Tree] {
+      override def apply(parent: tpd.Tree, tree: tpd.Tree)(implicit ctx: Context): tpd.Tree = {
+        tree match {
+          case tree: tpd.New if !parent.isInstanceOf[tpd.Select] =>
+            assert(assertion = false, i"`New` node must be wrapped in a `Select`:\n  parent = ${parent.show}\n  child = ${tree.show}")
+          case _: tpd.Annotated =>
+            // Don't check inside annotations, since they're allowed to contain
+            // somewhat invalid trees.
+          case _ =>
+            foldOver(tree, tree) // replace the parent when folding over the children
+        }
+        parent // return the old parent so that my siblings see it
+      }
+    })(tpd.EmptyTree, tree)
   }
 }
 


compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -122,6 +122,10 @@ class TreeChecker extends Phase with SymTransformer {
     val squahsedPhase = ctx.squashed(prevPhase)
     ctx.echo(s"checking ${ctx.compilationUnit} after phase ${squahsedPhase}")
 
+    if (ctx.isAfterTyper) {
```
abeln: @DarkDimius is this the right place to invoke the new check? Do I need to pass a fresh context?

abeln:
@DarkDimius The regular tests pass, but the legacy and bootstrap suites fail with what looks like unrelated (to this PR) errors. Are those tests flaky?

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -449,6 +453,53 @@ class TreeChecker extends Phase with SymTransformer {
...
+                    case _ => st // return the old parent so that my siblings see it
+                  }
+              }
+          }
+        }
+      }
+
+      detect(NotFound(None), tree) match {
+        case Found(Some(parent), child) =>
+          assert(assertion = false, i"`New` node must be wrapped in a `Select`:\n  parent = ${parent.show}\n  child = ${child.show}")
```
nicolasstucki: I would put the assertion inside of the `TreeAccumulator .apply` to have it fail with the accumulator stacktrace. This would allow us to attache the debugger at the moment of the assertion failure and navigate back to the trees that contained this malformed `New`.

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -449,6 +453,53 @@ class TreeChecker extends Phase with SymTransformer {
...
+
+  /**
+    * Checks that `New` nodes are always wrapped inside `Select` nodes.
+    */
+  private object WrappedNewChecker {
+
+    // Current state of the search. We return as soon as we see the first improperly wrapped `New`.
+    sealed trait State
+    case class NotFound(parent: Option[tpd.Tree]) extends State
+    case class Found(parent: Option[tpd.Tree], child: tpd.New) extends State
```
nicolasstucki: Why not just `partent: tpd.Tree`

abeln: When I invoke the accumulator on the root of the tree, there's no parent node. But I guess I could also pass an empty Thicket?

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
abeln: That's a good idea. Will do.

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
nicolasstucki: But on the root you use `NotFound`. My comment is only for `Found`.

For `NotFound` you could maybe simplify it using some trick like the `Ticket` or defining a `case object Root extends State`

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
nicolasstucki: If you do this you might not even need the parameter of `Found` and could make it a `case object`

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
abeln: Well, when we construct the Found object,we use the optional parent from the NotFound, so the optionality trickles through the hierarchy. We can get rid of it everywhere with the Thicket.

abeln: Applied your suggestion and the code is now simpler. Thanks.

abeln:
@DarkDimius @nicolasstucki 
The test failures are because sometimes annotations violate the invariant:

e.g. this test https://github.com/lampepfl/dotty/blob/master/tests/pos/functions1.scala
after tailrec ends up with this node

> Annotated(Ident(x),New(TypeTree[TypeRef(ThisType(TypeRef(NoPrefix,scala)),unchecked)]))

So should we relax the invariant? Is it ok for Annotated nodes to wrap New? (if so, which constructor gets chosen?)

Xprint:tailrec shows

val y: PartialFunction[String, String] = 
        {
          def $anonfun(x: String): String = 
            x match 
              {
                case "abc" => ""
                case _ => 
                  x
              }
          def isDefinedAt(x: String): Boolean = 
            x @unchecked match 
              {
                case "abc" => true
                case _ => true
                case _ @ _ => false
              }
          {
            class $anonfun() extends Object() with PartialFunction { 
              def apply(x: String): String = $anonfun(x)
              def isDefinedAt(x: String): Boolean = isDefinedAt(x)
            }
            new Object with PartialFunction[String, String]{...}()
          }
        }

DarkDimius:
> Is it ok for Annotated nodes to wrap New?

Annotations aren't systematically transformed in the pipeline and they also can contain some-what invalid trees. I think you shouldn't recurse into annotations at all.

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
DarkDimius: Yes, this is the right place.
TreeChecker is always after typer, no need for the if.

abeln:
@DarkDimius We no longer check inside annotations. PTAL.

abeln:
Friendly ping.

DarkDimius:
Looks good to me!
@abeln, thanks for making Ycheck better!

PR #2640, opened 2017-05-31
Fix #2636: assert in Run.scala is inefficient

### LINKED ISSUE #2636: Assert in Run.scala is inefficient
Run.scala contains:
```scala
assert(comp.phases.last.last.id <= Periods.MaxPossiblePhaseId)
```

But `phases` is a def in `Compiler` that will allocate every phase. This is pretty wasteful since we then throw away the result. We could probably move this assert inside the definition of `phases`, or at the point where we actually use `phases`.

abeln:
Move the assert after the declaration of the phases.

Tested:
Compiled one program and nothing blew up.

### DIFF
--- a/compiler/src/dotty/tools/dotc/Compiler.scala
+++ b/compiler/src/dotty/tools/dotc/Compiler.scala
@@ -40,8 +40,8 @@ class Compiler {
    *   - after erasure, asSeenFrom is the identity, so every reference has a
    *     plain SymDenotation, as opposed to a UniqueRefDenotation.
    */
-  def phases: List[List[Phase]] =
-    List(
+  def phases: List[List[Phase]] = {
+    val allPhases = List(
       List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
       List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
       List(new PostTyper),          // Additional checks and cleanups after type checking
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/Run.scala
```diff
@@ -24,7 +24,6 @@ import scala.util.control.NonFatal
 /** A compiler run. Exports various methods to compile source files */
 class Run(comp: Compiler)(implicit ctx: Context) {
 
-  assert(comp.phases.last.last.id <= Periods.MaxPossiblePhaseId)
```
DarkDimius: New runs are rarely created, this is not a performance critical place

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -109,6 +109,9 @@ class Compiler {
            new LabelDefs),          // Converts calls to labels to jumps
       List(new GenBCode)            // Generate JVM bytecode
     )
+    assert(allPhases.last.last.id <= Periods.MaxPossiblePhaseId)
```
DarkDimius: The idea is that compiler plugins may (in future) override\filter\change phases somehow. This means that this is wrong place to test the phase-plan, because it may still change. Phases have not been combine or initialized.

Additionally, Transforms require their own periods, and they will not be set here.

DarkDimius: \+ are phase ids set here at all? isn't `allPhases.last.last.id` not yet initialized here?

compiler/src/dotty/tools/dotc/Run.scala
abeln: Ok, this isn't needed then.

PR #2647, opened 2017-06-01
Local optimisations round two

OlivierBlanvillain:
Follow up PR of #2513

Objective: bootstrap Dotty with `-optimise`!


compiler/src/dotty/tools/dotc/transform/localopt/BubbleUpNothing.scala
```diff
@@ -0,0 +1,57 @@
+package dotty.tools.dotc
+package transform.localopt
+
+import core.Contexts.Context
+import core.Symbols._
+import ast.Trees._
+
+/** Every pure statement preceding a ??? can be removed.
+ *
+ *  This optimisation makes it rather tricky meaningful examples since the
```
smarter: This sentence seems to be missing some words.

compiler/src/dotty/tools/dotc/transform/localopt/Optimisation.scala
```diff
@@ -0,0 +1,20 @@
...
+import core.Contexts.Context
+import ast.tpd.Tree
+
+trait Optimisation {
+
+  /** Run first to gather information on Trees (using mutation) */
+  def visitor: Tree => Unit
+
+  /** Does the actual Tree => Tree transformation, possibly using a different
+   *  context from the one using in Optimisation.
```
smarter: Could you give more explanations on what contexts are used when? Also typo: using -> used.

compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -242,16 +242,16 @@ class GenBCodePipeline(val entryPoints: List[Symbol], val int: DottyBackendInter
     } // end of class BCodePhase.Worker1
 
     /*
-     *  Pipeline that takes ClassNodes from queue-2. The unit of work depends on the optimization level:
+     *  Pipeline that takes ClassNodes from queue-2. The unit of work depends on the optimisation level:
```
japgolly: :laughing: :+1: :100:

compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -242,16 +242,16 @@ class GenBCodePipeline(val entryPoints: List[Symbol], val int: DottyBackendInter
...
+     *    (a) no optimisation involves:
      *          - converting the plain ClassNode to byte array and placing it on queue-3
      */
     class Worker2 {
       // lazy val localOpt = new LocalOpt(new Settings())
 
-      def localOptimizations(classNode: ClassNode): Unit = {
-        // BackendStats.timed(BackendStats.methodOptTimer)(localOpt.methodOptimizations(classNode))
+      def localOptimisations(classNode: ClassNode): Unit = {
+        // BackendStats.timed(BackendStats.methodOptTimer)(localOpt.methodOptimisations(classNode))
```
DarkDimius: I propose not to rename the method here. This is how method is actually called in the backend.

DarkDimius:
@OlivierBlanvillain, If you don't mind, I propose to wait with getting this PR in.
I've done quite some changes to Simplify to enable more optimizations and add more documentation.
Would you mind if I take your changes and apply them by hand on my branch?(after my branch passes the tests)

DarkDimius:
NB: Oxford English Dictionary, prefers the `-ize`

OlivierBlanvillain:
@DarkDimius How large is your diff? I would be surprise if git manages to rebase either diffs on top of the other, and if it doesn't, it's probably easier to hand apply your changes on top of this rather than the other way around... (Happy to help if you point be to a branch)

I really don't have a preference between -ise and -ize, as long has it's not half/half :smile:

OlivierBlanvillain:
@DarkDimius Found the issue with DropNoEffects, if you want to have a look a the last commits. It's currently not tested by the CI, but this is what I use locally:

```sh
$ sbt ';project dotty-compiler-bootstrapped ;clean ;set scalacOptions ++= Seq("-optimise", "-Yopt-phases:DropNoEffects") ;test'
```

DarkDimius:
@OlivierBlanvillain, nice find! Did tests pass on your machine after the change?

OlivierBlanvillain:
The above, i.e. bootstrapped Dotty tests with `-optimise -Yopt-phases:DropNoEffects` does. That's a tiny step but I'm confident we can make it work will all optimisations enabled!

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -447,6 +448,8 @@ class Definitions {
     def Long_* = Long_mulR.symbol
     lazy val Long_divR   = LongClass.requiredMethodRef(nme.DIV, List(LongType))
     def Long_/ = Long_divR.symbol
+  val CommutativePrimitiveOperations = new PerRun[collection.Set[Symbol]](implicit ctx =>
```
DarkDimius: Don't put stuff inside Definitions, unless its frequently used by multiple phases.

Same applies for SeqFactoryType.

compiler/src/dotty/tools/dotc/transform/localopt/InlineCaseIntrinsics.scala
```diff
@@ -0,0 +1,134 @@
...
+      else a
+
+    // Seq.unapplySeq(arg) → new Some(arg)
+    // Where Seq is any companion of type <: SeqFactoryClass
+    case a: Apply if (a.symbol.name == nme.unapplySeq)           &&
+                     a.symbol.owner.derivesFrom(defn.SeqFactoryClass) &&
+                     a.symbol.extendedOverriddenSymbols.isEmpty  &&
+                     (isPureExpr(a.fun) || a.fun.symbol.is(Synthetic)) =>
+
+      def reciever(t: Tree): Type = t match {
```
smarter: typo: reciever -> receiver

compiler/src/dotty/tools/dotc/transform/localopt/InlineCaseIntrinsics.scala
```diff
@@ -0,0 +1,134 @@
...
+      case x => x
+    }
+    val recv = receiver(a)
+    if (recv.isEmpty || isPureRef(recv))
+      res
+    else
+      Block(recv :: Nil, res)
+  }
+
+  // To run this optimisation after erasure one would need to specialize it
```
smarter: Move this comment to the top of the class?

compiler/src/dotty/tools/dotc/transform/localopt/InlineLabelsCalledOnce.scala
```diff
@@ -0,0 +1,50 @@
...
+        case None => a
+        case Some(defDef) if a.symbol.is(Label) && timesUsed.getOrElse(a.symbol, 0) == 1 && a.symbol.info.paramInfoss == List(Nil) =>
+          simplify.println(s"Inlining labeldef ${defDef.name}")
+          defDef.rhs.changeOwner(defDef.symbol, localCtx.owner)
+        case Some(defDef) if defDef.rhs.isInstanceOf[Literal] =>
+          defDef.rhs
+        case Some(_) =>
+          a
+      }
+    case a: DefDef if (a.symbol.is(Label) && timesUsed.getOrElse(a.symbol, 0) == 1 && defined.contains(a.symbol)) =>
```
smarter: This same if expression is repeated below, factor it out in a def?

DarkDimius:
@OlivierBlanvillain, i've pushed a merge of our changes. I've quite substantially changed the structure of the split, because the original split introduced 70% slowdown. 
Now, there is only 20% slowdown. I'm still not sure if this split is worth this slowdown. I'd prefer to keep this phase fast, as it has to run multiple times during the pipeline and it should "pay for itself" by speeding up other phases. If this phase becomes 70% slower, it's harder for it to pay for itself.

smarter:
What changes in the split cause a slowdown?

DarkDimius:
50% that I've removed - substantially more allocations. More maps, sets & etc were allocated, unlike before. More lambdas were coined. More calls were non-local and cold. E.g. there were two contexts being used interchangeably.

In general: abstractions in Scala are anything but free. What brings the remaining 20% slowdown - I haven't checked yet and I don't think I'll have opportunity to do it in coming months.

OlivierBlanvillain:
@DarkDimius Interesting, what measure did you use?

OlivierBlanvillain:
@DarkDimius I just force pushed to restore history, I authored 3 commits under you name, you can check that nothing got lost using `git diff a386f63b f8cf56fc`.

I isolated the CI failure, it looks like `compileStdLib` compiles forever with `-optimise -Yopt-phases:Devalify` starting from commit 93e1b6da (Collapse ctx and localCtx together). I guess Devalify really need to run using the context from the overall `DefDef` tree instead of the local one. Note that before the split, the `ctx` used by every `Visitor`/`Transformer` pair is fixed in the in outer loop, [here](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/linker/Simplify.scala#L97) (which corresponds to the class parameter you instantiated), the `localCtx` was almost never used before that commit, maybe we should try getting ride of this one instead.

DarkDimius:
Let's get this in. The followup discussions can be held separately, but this PR in any case is an improvement.

compiler/src/dotty/tools/dotc/transform/localopt/Optimisation.scala
OlivierBlanvillain: Got addressed in this PR, we always use the correct, most local context!

DarkDimius:
LGTM. Glad to have it in.

PR #2652, opened 2017-06-01
Do not allow scala.Phantom self type

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1345,7 +1345,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     }
 
     def isScalaPhantomSelf(self: tpd.ValDef): Boolean = {
-      if (self.tpt.tpe.derivesFrom(defn.PhantomClass)) {
+      if (self.tpt.tpe.classSymbol eq defn.PhantomClass) {
```
odersky: Can you explain what the restriction is supposed to prevent? I did not get why it's `eq PhantonClass`, or even why having PhantomClass as a selftype is a problem. I would think it more plausible that phantom _types_, i.e. members of a subclass of `PhantomClass` should not be self types, but even there I would need to see detailed justification. It's true that classes with such self types cannot be instantiated, but why enforce that at the selftype level?

nicolasstucki: I was trying to do a trait that looked like 
```scala
trait Foo { self: Phantom => 
  def x: self.Any = self.asume
}
object Boo extends Phantom with Foo
```
But I ended with some top types with prefix `Foo` instead of `Boo` in places where it should not.

This is mostly intended as a quick fix until I have time to figure out how to re-enable them and be sure they work correctly.

nicolasstucki:
I believe the issue came from another bug that I've fixed. I will close this PR for now.

compiler/src/dotty/tools/dotc/typer/Typer.scala
nicolasstucki: FYI, e82f1ebccef851b727828095d96628dfef951a9a was the fix to the issue that I wanted to avoid.

PR #2658, opened 2017-06-02
homepage: Fix position of downward arrow to not overlap with text

smarter:
Previously, the arrow overlapepd with the "Dotty Example Project" link in the Getting Started
section, making it difficult to click.


felixmulder:
LGTM, thanks Guillaume!

Issue #2659, opened 2017-06-02
When opening a two-level deep documentation page, the sidebar is not fully expanded to show the page


smarter:
For example, in http://dotty.epfl.ch/docs/reference/adts.html the sidebar shows the "Reference" section expanded but not the "Enums" subsection.

felixmulder:
It's because the file is in the wrong directory. Is in: `./docs/reference/` but it should be in `./docs/reference/enums/`

felixmulder:
So either move it OR - make the script in `_includes/sidebar.html`:

1. look at the url of the current page
2. find the correct element in the sidebar
3. expand all parents

smarter:
Moving it seems like a good idea.

Varunram:
@smarter @felixmulder Was just about to ask whether the lazy move was a good one

felixmulder:
I think this is fine for a first fix - could you move the other pages that should be in `enums` as well? :)

Varunram:
Sure!

smarter:
@felixmulder http://dotty.epfl.ch/docs/reference/changed/lazy-vals.html seems to be under the correct subfolder and yet nothing is expanded in the sidebar.

Varunram:
The other files in changed don't seem to expand too..

Varunram:
Additional info that might be useful - So I moved stuff into their respective folders, like `enums`, `new`, `others`, etc. And every subsection except enum shows an expanded view in the sidebar. CC @felixmulder

W4anD0eR96:
This issue looks resolved. CC @smarter

PR #2664, opened 2017-06-02
Fix issue where sidebar parent tree doesn't show up properly

Varunram:
Fixes the enums case of #2659. Yet to figure out why this is happening as of now.


felixmulder:
Thanks @Varunram !

PR #2676, opened 2017-06-04
Faster base class sets

odersky:
Speed up construction and testing of base class sets.
Use a linear search with automatic bubbling up of
hits.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -1400,14 +1400,14 @@ object SymDenotations {
       baseData._2
 
     def computeBaseData(implicit onBehalf: BaseData, ctx: Context): (List[ClassSymbol], BaseClassSet) = {
-      val seen = mutable.SortedSet[Int]()
+      val seen = new BaseClassSetBuilder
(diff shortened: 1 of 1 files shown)


odersky:
@DarkDimius, since you suggested the scheme, do you want to review it?

DarkDimius:
Yes, gladly. Did you already do any measurements?

odersky:
I did not do any measurements yet. Jason did a similar change for scalac (minus the bubbling up part) and it gained them some speed there.

DarkDimius:
@biboudis @olafurpg , are the nightly benchmarks up? This PR is a good opportunity to test their sensitivity

retronym:
My hypothesis in scalac was that (part?) of the speedup was due to forcing fewer basetypeseqs, which we had to do to compute their length, part of the ordering criteria in the binary search. Here are some numbers measured before and after: https://github.com/scala/scala/pull/5829

https://github.com/scala/scala/pull/5829

olafurpg:
@DarkDimius I am helping @biboudis setup/manage the benchmark infrastructure. The benchmarks currently run an old 0.1-bin version, fix pending in https://github.com/lampepfl/dotty/pull/2690

PR #2682, opened 2017-06-04
Move only `case class` or `case object` allowed to new error format

mads-hartmann:
As part of #1589 this moves the error message 'only `case class` or
`case object` allowed' to the new error scheme.


mads-hartmann:
My bad, should be signed now 👍

felixmulder:
@dotty-bot: could you check the CLA please?

mads-hartmann:
@felixmulder Thanks for reviewing it so quickly 👍

PR #2685, opened 2017-06-05
Support Scala 2.12 Mixins

odersky:
This PR is a first step to Scala 2.12 support from Dotty. It makes it possible for Dotty to extend traits compiled with Scala 2.12.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -734,7 +734,7 @@ object Denotations {
         this
       case _ =>
         if (coveredInterval.containsPhaseId(ctx.phaseId)) {
-          if (ctx.debug) ctx.traceInvalid(this)
+          if (ctx.debug || true) ctx.traceInvalid(this)
           staleSymbolError
         }
         else NoDenotation
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -734,7 +734,7 @@ object Denotations {
         this
       case _ =>
         if (coveredInterval.containsPhaseId(ctx.phaseId)) {
-          if (ctx.debug || true) ctx.traceInvalid(this)
+          if (ctx.debug) ctx.traceInvalid(this)
           staleSymbolError
         }
         else NoDenotation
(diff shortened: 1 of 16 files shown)


smarter:
> not found: ./src/dotty/tools/dotc/transform/LinkScala2ImplClasses.scala

In compiler/test/dotc/tests.scala there is a reference to LinkScala2ImplClasses.scala that needs to be renamed.

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -399,9 +399,15 @@ object Flags {
   /** A module variable (Scala 2.x only) */
   final val Scala2ModuleVar = termFlag(57, "<modulevar>")
 
+  /** A Scala 2.12 trait that has been augmented with static members */
+  final val Scala_2_12_Augmented = typeFlag(57, "<scala12augmented")
```
senia-psm: I don't know the convention about type flag names, but this is the only one with `<`, but without `>`.
Should it be `<scala12augmented>` instead?

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -734,7 +734,7 @@ object Denotations {
         this
       case _ =>
         if (coveredInterval.containsPhaseId(ctx.phaseId)) {
-          if (ctx.debug) ctx.traceInvalid(this)
+          if (ctx.debug || true) ctx.traceInvalid(this)
```
senia-psm: This line looks strange.

odersky: Good catches, thanks!

compiler/src/dotty/tools/dotc/core/Mode.scala
```diff
@@ -70,26 +70,29 @@ object Mode {
   /** We are currently unpickling Scala2 info */
   val Scala2Unpickling = newMode(13, "Scala2Unpickling")
 
+  /** We are currently unpickling from Java 8 or higher */
+  val Java8Unpickling = newMode(14, "Java8Unpickling")
```
DarkDimius: it would be nice to leave a note that it is used as heuristic to find scala 2.12

DarkDimius:
Otherwise LGTM.
As far as I understand, those tests don't actually test anything, as we are only using scala 2.11 in our CI

PR #2686, opened 2017-06-05
Add artifactory resolver

felixmulder:
This change adds the local EPFL Artifactory for CI integration tests. Which should reduce our daily bandwidth consumption from 50gigs ot less :tada:

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,6 +2,12 @@
 //
 // e.g. addSbtPlugin("com.github.mpeltonen" % "sbt-idea" % "1.1.0")
 
+// Add EPFL Artifactory if on CI
+resolvers ++= {
+  if (sys.env.get("CI") == Some("drone")) Seq("EPFL Artifactory" at "http://scala-webapps.epfl.ch:8081")
+  else Seq()
+}
+
 // Scala IDE project file generator
 addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "5.1.0")
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -124,9 +124,10 @@ object Build {
     resourceDirectory in Compile    := baseDirectory.value / "resources",
   
     // Add EPFL resolver if on CI
-    resolvers ++=
+    resolvers ++= {
       if (sys.env.get("CI") == Some("drone")) Seq("EPFL Artifactory" at "http://scala-webapps.epfl.ch:8081")
-      else Seq(),
+      else Seq()
+    },
 
     // Prevent sbt from rewriting our dependencies
     ivyScala ~= (_ map (_ copy (overrideScalaVersion = false)))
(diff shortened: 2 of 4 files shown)


smarter:
> Which should reduce our daily bandwidth consumption from 50gigs ot less :tada:

Isn't that just because we haven't updated the ivy2 cache on the CI in a while?

project/plugins.sbt
```diff
@@ -2,6 +2,12 @@
 //
 // e.g. addSbtPlugin("com.github.mpeltonen" % "sbt-idea" % "1.1.0")
 
+// Add EPFL Artifactory if on CI
+resolvers ++= {
```
smarter: I think you're adding the same resolver 5 times, which may impact resolving speed.

felixmulder:
That's also true - but using this, even if we get out of sync - it should still be fine.

Regarding adding the resolver in multiple places - it seems to be needed for resolving plugins... T_T

project/plugins.sbt
smarter: Also you're not adding it to all the scripted tests (sbt-dotty/sbt-test/...), is that needed to?

smarter:
Yeah, but keeping in sync is not a lot of work, compared to maintaining a whole artifactory instance, I don't get the rationale here.

project/plugins.sbt
felixmulder: It might be. Maybe we should just include it in `Build.scala` and then update our cache. WDYT?

felixmulder:
What is the maintenance cost?

smarter:
What's the plan if the artifactory instance breaks? Do we all need to learn how it works? Can we disable it in some way if we need to? etc.

felixmulder:
If it's down, then it's down - it'll fall back to the other resolvers.

Fabien is running it, disabling == remove from script

smarter:
It looks like artifactory is not even open source (I looked around and couldn't find the sources at least) that's a deal breaker for me. I also don't understand why we would need to maintain a giant Java application when all we need is a 10 lines bash script to periodically check if we need to update the ivy2 cache and regenerated the docker image.

felixmulder:
The difference is that somebody still needs to do that - if you want to, that'd be great - but this fix is now.

There is also the issue of drone being tied to specific tags of the docker images it uses as its testing environment. So we'd need the script to also update `.drone.yml`, open a PR and have someone merge it (or it can commit straight to master 🤘 😈 )

We could also let the script push to the `latest` tag as well as a date tag, then we'd have the `.drone.yml` file point to latest and bisect if there's something wrong with the cache.

I'm fine with either one here really, the effect from my vantage point is the same...

DarkDimius:
An open source solution could be to set up a reverse caching proxy(e.g. squid) on EPFL infrastructure.

I'm fine with artifactory though. Sonatype & Maven central are not open-source either. They have free-ish versions, but the thing that actually runs is highly modified.

smarter:
Since sbt does not download anything in parallel, using artifactory will also be slower than having an up-to-date local cache, in adition to being much more complicated.

OlivierBlanvillain:
Counter proposal: run the tests offline to make sure that the cache is fully up to date. That would also give ultimate reproducibility of history and immunity to external failures.

PR #2687, opened 2017-06-05
Harden IDE

odersky:
Three fixes of problems I noted when using VSCode for editing the compiler


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -113,6 +113,9 @@ class DottyLanguageServer extends LanguageServer
       }
     })
 
+  private def ignoreStale[T](onError: T)(op: => T): T = 
+    try op catch { case ex: StaleSymbol => onError }
```
smarter: I'd rather not do this because StaleSymbol errors in the IDE have so far always been bug in the compiler that should be fixed, and this commit will make them much harder to track down. If you could give some reproduction steps for the case you encountered, we might be able to find and fix the root cause.

odersky:
Replaced by #2692

Issue #2570, opened 2017-05-29
Scala Wart: Weak eta-expansion


lihaoyi:
Opening this issue, as suggested by Martin, to provide a place to discuss the individual warts brought up in the blog post [Warts of the Scala Programming Language](http://www.lihaoyi.com/post/WartsoftheScalaProgrammingLanguage.html) and the possibility of mitigating/fixing them in Dotty (and perhaps later in Scala 2.x). These are based on Scala 2.x behavior, which I understand Dotty follows closely, apologies in advance if it has already been fixed

--------------------------------------------------------------------------------

Scala maintains a distinction between "functions" and "methods": in general,
methods are things you call on an object, whereas functions are objects 
themselves. However, since they're so similar ("things you can call"), it gives 
you a way to easily wrap a method in a function object called "eta expansion"

```scala
@ def repeat(s: String, i: Int) = s * i
defined function repeat

@ repeat("hello", 2)
res89: String = "hellohello"

@ val func = repeat _
func: (String, Int) => String = $sess.cmd90$$$Lambda$2796/1082786554@2a3983b9

@ func("hello", 3)
res91: String = "hellohellohello"
```

Above, we use the underscore `_` to assign `repeat _` to a value `func`, which 
is then a function object we can call. This can happen automatically, without the
`_`, based on the "expected type" of the place the method is being used. For 
example, if we expect `func` to be a `(String, Int) => String`, we can assign
`repeat` to it without the `_`:

```scala
@ val func: (String, Int) => String = repeat
func: (String, Int) => String = $sess.cmd92$$$Lambda$2803/669946146@46226d53

@ func("hello", 3)
res92: String = "hellohellohello"
```

Or by stubbing out the arguments with `_` individually:

```scala
@ val func = repeat(_, _)
func: (String, Int) => String = $sess.cmd98$$$Lambda$2832/1025548997@358b1f86
```

This works, but has a bunch of annoying limitations. Firstly, even though you 
can fully convert the method `repeat` into a `(String, Int) => String` value 
using `_`, you cannot *partially* convert it:

```scala
@ val func = repeat("hello", _)
cmd4.sc:1: missing parameter type for expanded function 
((x$1: <error>) => repeat("hello", x$1))
val func = repeat("hello", _)
                           ^
Compilation Failed
```

*Unless* you know the the "expected type" of `func`, in which case you *can*
partially convert it:

```scala
@ val func: Int => String = repeat("hello", _)
[...]

PR #2691, opened 2017-06-06
Allow partial eta expansion

odersky:
Allow partial eta expansion using the following rule:

Assume we have an application

    (x_1, ..., x_m) => f(<arg_1, ..., arg_n>)

where every `x_i` occurs exactly once as argument to `f`, and typing `f` with `(?, ..., ?) => ?`
as expected type (where there are n occurrences of `?` in the argument list)
yields a method type `(T_1, ..., T_n)R`. Let `p` be the mapping from `1..m` to `1..n`
which maps every parameter `x_i` to the position where it occurs in `<arg_1, ..., arg_n>`.

In this case, if one of the parameter types `x_i` is not given and not inferred from the
expected type, assume `T_p(i)` as the type of `x_i`, provided `T_p(i)` is fully defined
and not a repeated type `T*` or `=>T*`.

The reason for excluding repeated types `T*` is that `T*` is not a valid type for a
lambda parameter.

Fixes one part of #2570.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -724,41 +724,70 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
        */
       var fnBody = tree.body
 
+      /** A map from parameter names to unique positions where the parameter
+       *  appears in the argument list of an application.
+       */
+      var paramIndex = Map[Name, Int]()
+
+      /** If parameter `param` appears exactly once as an argument in `args`,
+       *  the singleton list consisting of its position in `args`, otherwise `Nil`.
+       */
+      def paramIndices(param: untpd.ValDef, args: List[untpd.Tree], start: Int): List[Int] = args match {
+        case arg :: args1 =>
+          if (refersTo(arg, param))
+            if (paramIndices(param, args1, start + 1).isEmpty) start :: Nil
+            else Nil
+          else paramIndices(param, args1, start + 1)
+        case _ => Nil
+      }
+
       /** If function is of the form
-       *      (x1, ..., xN) => f(x1, ..., XN)
-       *  the type of `f`, otherwise NoType. (updates `fnBody` as a side effect).
+       *      (x1, ..., xN) => f(... x1, ..., XN, ...)
+       *  where each `xi` occurs exactly once in the argument list of `f` (in
+       *  any order), the type of `f`, otherwise NoType.
+       *  Updates `fnBody` and `paramIndex` as a side effect.
+       *  @post: If result exists, `paramIndex` is defined for the name of
+       *         every parameter in `params`.
        */
       def calleeType: Type = fnBody match {
-        case Apply(expr, args) if (args corresponds params)(refersTo) =>
-          expr match {
-            case untpd.TypedSplice(expr1) =>
-              expr1.tpe
-            case _ =>
-              val protoArgs = args map (_ withType WildcardType)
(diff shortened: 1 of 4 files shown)


odersky:
Reverted back to initial commit because the rest is not directly related (and more subtle, it turns out).

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -724,41 +724,70 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-              val expr1 = typedExpr(expr, callProto)
-              fnBody = cpy.Apply(fnBody)(untpd.TypedSplice(expr1), args)
-              expr1.tpe
-          }
+        case Apply(expr, args) =>
+          paramIndex = {
+            for (param <- params; idx <- paramIndices(param, args, 0))
+            yield param.name -> idx
+          }.toMap
+          if (paramIndex.size == params.length)
```
adriaanm: Since it's a map, wouldn't this miss the (unlikely) case where the same param (one key) occurs in multiple args (multiple values, of which only the last one is retained). I don't think it's a big deal, as this could not arise from the `_` syntax, but the other comments do mention "exactly once" as a requirement.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -724,41 +724,70 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+            expr match {
+              case untpd.TypedSplice(expr1) =>
+                expr1.tpe
+              case _ =>
+                val protoArgs = args map (_ withType WildcardType)
```
adriaanm: As discussed, the [current implementation in Scala 2](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L2972) includes the known argument types / result type, to enable overload resolution. Would be good to indicate the difference in a comment.

tests/pos/i2570.scala
```diff
@@ -0,0 +1,29 @@
+object Test {
+
+  def repeat(s: String, i: Int, j: Int = 22) = s * i
```
adriaanm: When discussing this with Seth, he suggested to play around with currying by adding another argument list. The partial application should be fine: `repeat(_, 1)`, but what about `repeat(_, 1)(_)`?  (uncurrying?)

compiler/src/dotty/tools/dotc/typer/Typer.scala
odersky: The revised implementation of paramIndices makes sure that a parameter is entered in the map only if there is a unique occurrence in args.

tests/pos/i2570.scala
odersky: In dotc you get:

    1 |repeat(1, _)(_)
      |             ^
      |             unbound placeholder parameter; incorrect use of `_`

In scalac you get:

    scala> repeat(1, _)(_) 
    <console>:13: error: missing parameter type for expanded function ((x$1: <error>, x$2) => repeat(1, x$1)(x$2))
           repeat(1, _)(_)
                     ^
    <console>:13: error: missing parameter type for expanded function ((x$1: <error>, x$2: <error>) => repeat(1, x$1)(x$2))
           repeat(1, _)(_)

The fact that the types already contain <error> markers makes me suspect that scalac also issues a parsing error but somehow suppresses the error message.

odersky:
I made a note to add the missing comment about differences between dotc and scalsc in the followup PR

PR #2692, opened 2017-06-06
Harden IDE, 2nd attempt

odersky:
On stale symbols: There are legitimate scenarios how they can arise in the IDE.

For instance, we define a class, refer to it, edit the class definition and introduce errors. The class symbol temporarily no longer exists until the errors are fixed. But typed trees elsewhere will refer to it. That's why StaleSymbol errors now give warnings by default. A config option can turn them back into exceptions, which is useful if we want to find the root cause of a particular warning.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -102,11 +102,10 @@ trait SymDenotations { this: Context =>
 
   /** Configurable: Accept stale symbol with warning if in IDE */
   def acceptStale(denot: SingleDenotation): Boolean =
-    (mode.is(Mode.Interactive) && Config.ignoreStaleInIDE) && {
-      ctx.warning(denot.staleSymbolMsg)
+    mode.is(Mode.Interactive) && Config.ignoreStaleInIDE && {
+      ctx.echo(denot.staleSymbolMsg)
       true
     }
-
 }
 
 object SymDenotations {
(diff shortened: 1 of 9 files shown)


compiler/src/dotty/tools/dotc/interactive/InteractiveCompiler.scala
```diff
@@ -14,4 +15,7 @@ class InteractiveCompiler extends Compiler {
   override def phases: List[List[Phase]] = List(
     List(new FrontEnd)
   )
+
+  override def rootContext(implicit ctx: Context) =
+    super.rootContext.fresh.addMode(Mode.Interactive)
```
smarter: There's also Mode.ReadPositions that we always need in the IDE, it's currently set in InteractiveDriver but could be moved here.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -99,6 +99,14 @@ trait SymDenotations { this: Context =>
         explain("denotation is not a SymDenotation")
     }
   }
+
+  /** Configurable: Accept stale symbol with warning if in IDE */
+  def acceptStale(denot: SingleDenotation): Boolean =
+    (mode.is(Mode.Interactive) && Config.ignoreStaleInIDE) && {
+      ctx.warning(denot.staleSymbolMsg)
```
smarter: Such a warning will popup in the IDE (not currently because we ignore warnings without positions, but that's just because of an issue in the LSP (https://github.com/Microsoft/language-server-protocol/issues/249)), maybe use ctx.debug instead ?

Issue #2689, opened 2017-06-05
Apply Scala syntax on each code snippet in the reference documentation


louis-forite:
Problem
--- 
While reading the reference documentation, under `/docs/docs/reference/...` I noticed that some of the code snippets have a scala syntax applied, some of them do not, reducing readibility and consistency.

Solution
--- 
In order to improve readibility and consistency, it would be good to go through those documents and apply the scala style everywhere.

Example

    implicit val m: Map[String, Int] = Map(1 -> "abc")

    val x: String = 1  // scalac: assigns "abc" to x
                       // Dotty: type error

should become 

```scala
    implicit val m: Map[String, Int] = Map(1 -> "abc")

    val x: String = 1  // scalac: assigns "abc" to x
                       // Dotty: type error
```

louis-forite:
Comprehensive list of files to be modified
--- 
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/changed/implicit-conversions.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/changed/implicit-resolution.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/changed/lazy-vals.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/changed/structural-types.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/changed/vararg-patterns.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/class-shadowing.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/delayed-init.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/early-initializers.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/existential-types.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/procedure-syntax.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/xml.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/enums/desugarEnums.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/auto-parameter-tupling.md
 - https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/implicit-function-types.md
[...]

PR #2693, opened 2017-06-06
Apply Scala Syntax for code snippets

Varunram:
Intended as a fix for #2689. Changes primarily involve indentation changes (since using syntax highlighting with the existing indent would look a bit ugly). Do point out errors if any. Thanks!


lforite:
I think we are good to go :-)

OlivierBlanvillain:
Could you amend your commit message to make the first line < 72 characters long?

Varunram:
Done! Off by one word :)

OlivierBlanvillain:
It still conflicting :/

Varunram:
Oh no, I didn't reset it. My bad. Just a moment.

Varunram:
Done finally! Thanks a lot!

felixmulder:
This guy needs a rebase before getting merged :)

I suggest doing something like this:

```bash
$ git remote add upstream git@github.com:lampepfl/dotty.git
$ git fetch upstream && git rebase -i upstream/master
```

This will ensure that you actually rebase against the latest master. Otherwise the problem might be that you're rebasing against an old HEAD.

Varunram:
@felixmulder I think the conflicts are related to the changes. Weird that it picked out only two files though

docs/docs/reference/changed/implicit-conversions.md
```diff
@@ -6,44 +6,45 @@ title: "Restrictions to Implicit Conversions"
...
+def convert[A, B](x: A)(implicit converter: ImplicitConverter[A, B]): B =
+  converter(x)
+```
 (In reality the Dotty compiler simulates the behavior of this method directly in
 its type checking because this turns out to be more efficient).
 
 In summary, previous code using implicit conversion parameters such as
-
+```scala
     def useConversion(implicit f: A => B) = {
```
OlivierBlanvillain: Indentation is off.

docs/docs/reference/changed/implicit-conversions.md
```diff
@@ -6,44 +6,45 @@ title: "Restrictions to Implicit Conversions"
...
       val y: A = ...
       val x: B = a    // error under Dotty
     }
+```
 
 is no longer legal and has to be rewritten to
-
+```scala
     def useConversion(implicit f: ImplicitConverter[A, B]) = {
```
OlivierBlanvillain: Indentation is off.

docs/docs/reference/enums/desugarEnums.md
```diff
@@ -31,122 +31,124 @@ into case classes, case objects or vals. Finally, rule (8) expands
...
+    case C <params> <body>
+    ```
+    expands according to two alternatives, depending whether `C` has type
+    parameters or not. If `C` has type parameters, they must have the same
+    names and appear in the same order as the enum type parameters `Ts`
+    (variances may be different, however). In this case
+    ```scala
+    case C [Ts] <params> <body>
+    ```
+      expands to
```
OlivierBlanvillain: Two extra white spaces

Varunram:
@OlivierBlanvillain sorry for the late reply. I think you're right! There must t be a better way to handle this. More info over at @2689

OlivierBlanvillain:
Sounds good, thanks a lot for looking into this :smile:. I closing this PR for now, please reopen/open a new one if you have something new!

PR #2694, opened 2017-06-06
Parse Function arity without throwing exceptions

smarter:
Using NumberFormatException in the normal control flow of the program
makes it harder to debug the compiler by stopping on all instances of
RuntimeException.


odersky:
I guess that would still fail for numbers that are too large. But that should not be a problem for most debug sessions. LGTM.

Issue #2615, opened 2017-05-30
Document literal singleton types


OlivierBlanvillain:
The current best source of documentation is [SIP-23 - Literal-based singleton types](http://docs.scala-lang.org/sips/pending/42.type.html), it would be nice to have something on dotty.epfl.ch!

Varunram:
@OlivierBlanvillain Maybe a new doc file over at `docs/reference`? What are your suggestions? Thanks!

OlivierBlanvillain:
@Varunram Yes, I was thinking about a new page under Reference / New Types.

Varunram:
@OlivierBlanvillain Sorry for the delay! Was stuck up in a few other things. So I've started working on this and have the doc part almost ready. It would be of immense help if you could point out some code examples so that I may use them in the page (since this is my first time writing a doc page). Thanks for your time!

OlivierBlanvillain:
@Varunram I think the current best source of examples is SIP-23... The main use case IMO would be a shapeless style `LabeledGeneric` type class for enums / case classes / sealed trait, but we are not quite there yet.

Varunram:
Ah, makes sense. I borrowed a couple of examples from folone's talks, which were related to SIP-23. Preliminary PR for your review will be up shortly :)

OlivierBlanvillain:
Awesome! I'll do my best to quickly review it :)

SethTisue:
should this be closed, since #2699 was merged?

PR #2699, opened 2017-06-07
Add doc page for Literal Singleton Types

Varunram:
Intended as a fix for #2615.

My first time writing a doc page, so please do comment your views. I do feel that some prose can be added, which I'm searching for. CC @OlivierBlanvillain

### DIFF
--- a/docs/docs/reference/singleton-types.md
+++ b/docs/docs/reference/singleton-types.md
@@ -7,7 +7,7 @@ Literal Singleton Types allows primitive literals to be used as types. For examp
 
 ```scala
 val t: 42 = 42
-val t: "Jedi" = "Jedi"
+val x: "Jedi" = "Jedi"
 ```
 
 It is also possible to define functions returning or taking singleton types such as
--- a/docs/docs/reference/singleton-types.md
+++ b/docs/docs/reference/singleton-types.md
@@ -1,0 +1,20 @@
+---
+layout: doc-page
+title: "Literal Singleton Types"
+---
+
+Literal Singleton Types allows primitive literals to be used as types. For example:
+
+```scala
+val t: 42 = 42
+val t: "Jedi" = "Jedi"
+```
+
+It is also possible to define functions returning or taking singleton types such as
+
+```scala
+def f(t: Double): t.type = t
+val a: 1.2 = f(1.2)
+```
+
+For more details and the motivation behind the need for literal singleton types, check out [SIP-23](http://docs.scala-lang.org/sips/pending/42.type.html)
--- a/docs/sidebar.yml
+++ b/docs/sidebar.yml
@@ -27,6 +27,8 @@ sidebar:
               url: docs/reference/implicit-function-types.html
             - title: Phantom Types
               url: docs/reference/phantom-types.html
+            - title: Literal Singleton Types
+              url: docs/reference/singleton-types.html
         - title: Enums
           subsection:
             - title: Enumerations


docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
+---
+layout: doc-page
+title: "Literal Singleton Types"
+---
+
+Singleton types is one of the most popular features in Scala. Literal Singleton types takes it a step further.
```
OlivierBlanvillain: I would cut the intro down to the most important bit, something like this:

Literal Singleton Types allows primitive literals to be used as types. For example:

```scala
val fourtyTwo: 42 = 42

val hero: "Jedi".type = "Jedi"

val trivial: true = true
```

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+val t: 42.type = 42
+```
+or
+```scala
+val t: 42 = 42
+```
+
+to use literal Singleton types. Trying to do the same in Scala produces the following error:
+```
+<console>:1: error: identifier expected but string literal found.
```
OlivierBlanvillain: I don't see the point of showing this error

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+val t: "Jedi".type = "Jedi"
+```
+
+Dotty provides support for using Literal Singleton types of the last format namely:
```
OlivierBlanvillain: This is Dotty's documentation, just show the Dotty format / example :)

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+```scala
+val t: 42 =42
+val x: "Jedi" = "Jedi"
+```
+and so on.
+
+There is a possibility to use singleton types in some contexts, but only on identifiers which points to a constant that conforms to `AnyRef`. This restriction is due to `Any` not having an eq method, which is what’s used for singleton type-equality check and pattern matching.
```
OlivierBlanvillain: I'm assuming you are talking about type contexts here? I'm not aware of this restriction with `AnyRef`, do you have an example? This seems to work fine:

```scala
scala> def f[T <: Any](t: T) = t
def f[T <: Any](t: T): [T](t: T)T

scala> f[1.1](1.1)
val res3: Double = 1.1
```

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
+---
+layout: doc-page
+title: "Literal Singleton Types"
+---
+
+Singleton types is one of the most popular features in Scala. Literal Singleton types takes it a step further.
+Right now, a popular way to replicate literal singleton types is to create macros and/or use libraries such as
```
OlivierBlanvillain: > Right now

Doc should be timeless when possible.

If you want to cite shapeless here please do it more precisely by pointing to the corresponding documentation.

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+`shapeless` due to the lack of a syntax to express them.
+
+With Literal Singelton types, it is possible to do something like
```
lforite: typo on `Singelton`

Varunram: Oh, no. Really sorry, will edit that. Thanks!

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+or
+```scala
+val t: 42.type = 42
+```
+or
+```scala
+val t: 42 = 42
+```
+
+to use literal Singleton types. Trying to do the same in Scala produces the following error:
```
lforite: I think `to use literal Singleton types` is redundant with `With Literal Singelton types`. What do you think ?

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,39 @@
...
+```
```
lforite: Isn't this block missing scala syntax coloring ? I am not sure if it should be applied here as the first line is not Scala code

docs/docs/reference/singleton-types.md
Varunram: Will do :)

docs/docs/reference/singleton-types.md
Varunram: Sure, will edit the md by adding a link to shapeless. I thought of referencing this primarily because it was cited so in the SIP

docs/docs/reference/singleton-types.md
Varunram: Will remove

docs/docs/reference/singleton-types.md
Varunram: I guess I'll remove that part since its not needed :)

docs/docs/reference/singleton-types.md
Varunram: Ah, this one is on me again, I referred to some old documentation and was led to believe that. Thanks for pointing it out!

Varunram:
Done! Thanks a ton @OlivierBlanvillain and @lforite for being so patient and pointing out changes!

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,21 @@
+---
+layout: doc-page
+title: "Literal Singleton Types"
+---
+
+Literal Singleton Types allows primitive literals to be used as types. For example:
+
+```scala
+val t: "Jedi".type = "Jedi"
+val t: 42.type = 42
```
OlivierBlanvillain: This one is not valid in Dotty

Varunram: Um, that's why I had a separate section for Dotty earlier. One minute, will revert changes and push again.

Varunram:
Okay, that wasn't supposed to happen. Lemme figure out why

Varunram:
Really sorry for the delay! Some weird thing happened while resetting and that caused the PR to close. Thank you very much for putting up with me :)

docs/docs/reference/singleton-types.md
```diff
@@ -0,0 +1,20 @@
+---
+layout: doc-page
+title: "Literal Singleton Types"
+---
+
+Literal Singleton Types allows primitive literals to be used as types. For example:
+
+```scala
+val t: 42 = 42
+val t: "Jedi" = "Jedi"
```
felixmulder: Maybe a different name for the second variable? :)

Varunram: On it

Varunram:
ping @felixmulder @OlivierBlanvillain

OlivierBlanvillain:
LGTM, thanks!

PR #2701, opened 2017-06-07
Changes to eta expansion

odersky:
- allow partial eta expansion
 - always eta expand if arity is >= 1

This implements the eta expansion parts of #2570, following the recommendations in my comment there:

After discussing this with @adriaanm, we are leaning towards the following compromise proposal for handling references to unapplied methods m:

 - If m has one or more parameters, we always eta expand

 - if m is nullary and the expected type is of the form () => T, we eta expand.

Based on #2691

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2011,17 +2011,25 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
             if (nparams > 0 || pt.eq(AnyFunctionProto)) nparams
             else -1 // no eta expansion in this case
           }
+        def isExpandableApply =
+          defn.isImplicitFunctionClass(tree.symbol.maybeOwner) && defn.isFunctionType(ptNorm)
 
-        if (wtp.isImplicit)
-          err.typeMismatch(tree, pt)
-        else if (arity >= 0 &&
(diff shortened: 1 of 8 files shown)


odersky:
Seems dotty-bot got chatty again?

smarter:
@odersky This is dotty-bot way of telling you that it doesn't like commit messages where the subject line ends with a '.', like https://github.com/lampepfl/dotty/pull/2701/commits/42cbf5c94d3aea82955b2a769c713e0b17fd9db4 ;)

felixmulder:
It's turned quite passive aggressive lately! :) We'll look at remedying it!

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1981,9 +2010,48 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+         *  would go undetected.
+         *
+         *  Examples for these cases are found in run/implicitFuns.scala and neg/i2006.scala.
+         */
+        def isExpandableApply =
+          defn.isImplicitFunctionClass(tree.symbol.maybeOwner) && defn.isFunctionType(ptNorm)
+
+        // Reasons NOT to eta expand:
+        //  - we reference a constructor
+        //  - we are in a patterm
```
senia-psm: Typo. Should be "patter**n**"

PR #2705, opened 2017-06-07
support debugging for dotc/dotr

liufengyun:
Support debugging option for `dotc/dotr`:

```bash
$ dotc -debug hello.scala
Listening for transport dt_socket at address: 5005

$ dotr -d Test
Listening for transport dt_socket at address: 5005
```

With this PR, we may consider remove the `bin/dotc` scripts:

- New scripts don't depend on source code
- New scripts support all features of old scripts
- New scripts is CI tested on Linux/Mac/Windows
- `sbt dist/pack` or `sbt dist-bootstrapped/pack` can be used for compiler development

The CI on Linux/Mac/Windows for this PR is here: https://github.com/liufengyun/packtest/pull/3

Note: `-d` is already used in the compiler for output directory, thus `dotc` has to use `-debug`.

### DIFF
--- a/dist/bin/common
+++ b/dist/bin/common
@@ -112,3 +112,6 @@ SCALA_ASM=$(find_lib "*scala-asm*")
 SCALA_LIB=$(find_lib "*scala-library*")
 SCALA_XML=$(find_lib "*scala-xml*")
 SBT_INTF=$(find_lib "*sbt-interface*")
+
+# debug
+DEBUG_STR=-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005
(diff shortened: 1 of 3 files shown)


DarkDimius:
> New scripts don't depend on source code

Does this mean that they may run a "dirty" compiler that doesn't match the compiled thing?
If it's the case, I'll still keep the old scripts. I use them during local development

liufengyun:
As I said, you can run `sbt dist/pack` or `sbt dist-bootstrapped/pack` to have two different versions of dotc/dotr at the same time for development.

- `sbt dist/pack`:  generate bin under `dist/target/pack/bin`
- `sbt dist-bootstrapped/pack`: generate bin under `dist-bootstrapped/target/pack/bin`

My experience tells that it's easier and more flexible to work with the news scripts than the old ones for compiler development.

https://github.com/xerial/sbt-pack#features

Anyway, just a proposal to try the new ones. We don't remove old ones in this PR.

PR #2708, opened 2017-06-07
DLS: do not swallow exceptions in initialization

smarter:


smarter:
Merging now so that this can be part of the next nightly.

PR #2716, opened 2017-06-08
Fix #2570: Part 3 - The great () insert

### LINKED ISSUE #2570: Scala Wart: Weak eta-expansion

odersky:
Require explicit () for nullary method applications. Fix a lot of files that were in violation. 

This is the final part of addressing #2570.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
+++ b/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
@@ -53,7 +53,7 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
     def map[R](f: (Symbol, Tree) => R): List[R] = {
       val b = List.newBuilder[R]
       foreach(b += f(_, _))
-      b.result
+      b.result()
     }
   }
 
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1538,15 +1538,31 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     val pt1 = if (defn.isFunctionType(pt)) pt else AnyFunctionProto
     var res = typed(qual, pt1)
     if (pt1.eq(AnyFunctionProto) && !defn.isFunctionClass(res.tpe.classSymbol)) {
-      def msg = i"not a function: ${res.tpe}; cannot be followed by `_'"
+      ctx.errorOrMigrationWarning(i"not a function: ${res.tpe}; cannot be followed by `_'", tree.pos)
       if (ctx.scala2Mode) {
         // Under -rewrite, patch `x _` to `(() => x)`
-        ctx.migrationWarning(msg, tree.pos)
         patch(Position(tree.pos.start), "(() => ")
         patch(Position(qual.pos.end, tree.pos.end), ")")
         res = typed(untpd.Function(Nil, untpd.TypedSplice(res)))
       }
-      else ctx.error(msg, tree.pos)
+    }
+    else if (ctx.settings.strict.value) {
+      lazy val (prefix, suffix) = res match {
+        case Block(mdef @ DefDef(_, _, vparams :: Nil, _, _) :: Nil, _: Closure) =>
+          val arity = vparams.length
+          if (arity > 0) ("", "") else ("(() => ", "())")
+        case _ =>
+          ("(() => ", ")")
+      }
+      def remedy =
+        if ((prefix ++ suffix).isEmpty) "simply leave out the trailing ` _`"
+        else s"use `$prefix<function>$suffix` instead"
+      ctx.errorOrMigrationWarning(i"""The syntax `<function> _` is no longer supported;
+                                     |you can $remedy""", tree.pos)
+      if (ctx.scala2Mode) {
+        patch(Position(tree.pos.start), prefix)
+        patch(Position(qual.pos.end, tree.pos.end), suffix)
+      }
     }
     res
   }
@@ -2053,7 +2069,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
 
         // Reasons NOT to eta expand:
         //  - we reference a constructor
-        //  - we are in a patterm
+        //  - we are in a pattern
         //  - the current tree is a synthetic apply which is not expandable (eta-expasion would simply undo that)
         if (arity >= 0 &&
             !tree.symbol.isConstructor &&
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2045,16 +2045,8 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
 
         /** Is reference to this symbol `f` automatically expanded to `f()`? */
         def isAutoApplied(sym: Symbol): Boolean = {
-          def test(sym1: Symbol) =
-            sym1.is(JavaDefined) ||
-            sym1.owner == defn.AnyClass ||
-            sym1 == defn.Object_clone
           sym.isConstructor ||
-          test(sym) ||
-          sym.allOverriddenSymbols.exists(test) ||
-          sym.owner.is(Scala2x) || // need to exclude Scala-2 compiled symbols for now, since the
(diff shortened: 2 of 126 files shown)


odersky:
Migration strategy: We don't complain if the function was compiled by Scala 2 (this is because the Scala libraries need to be cleaned up first, so that they have `()` in the right places.).  

We issue a migration warning under `-language:Scala2`, and fix the problem under `-rewrite`. We 
issue an error otherwise. 

So, to migrate code to Dotty, just compile it with 

    dotc <files> -language:Scala2 -rewrite

It will insert the missing `()` arguments. ScalaFix should do the same. /cc @olafurpg.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -724,41 +724,70 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+       */
+      var paramIndex = Map[Name, Int]()
+
+      /** If parameter `param` appears exactly once as an argument in `args`,
+       *  the singleton list consisting of its position in `args`, otherwise `Nil`.
+       */
+      def paramIndices(param: untpd.ValDef, args: List[untpd.Tree], start: Int): List[Int] = args match {
+        case arg :: args1 =>
+          if (refersTo(arg, param))
+            if (paramIndices(param, args1, start + 1).isEmpty) start :: Nil
```
senia-psm: This method returns `Some` if `param` appears in `args` any odd number of times, not only "exactly one".
Please see [this code snippet](https://scalafiddle.io/sf/UrSR4Gn/0).

odersky: Good catch.

odersky:
We seem to have an error early on in the CI. The building `dotc` script fails with mysterious errors.

olafurpg:
@odersky Noted. https://github.com/scalacenter/scalafix/issues/204

smarter:
This PR needs to be rebased because scala-backend was updated in https://github.com/lampepfl/dotty/pull/2719

odersky:
We also need to tighten the rules about overriding. It should be no longer legal to override `def f(): T` with `def f: T` and vice versa, except if one of the definitions is in Java or Scala 2 code.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -470,6 +477,14 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
     case _ => false
   }
 
+  /** Is tree a compiler-generated `.apply` node that refers to the
+   *  apply of a function class? (implicit functions are excluded)
+   */
+  def isSyntheticApply(tree: Tree): Boolean = tree match {
+    case Select(qual, nme.apply) => tree.pos.end == qual.pos.end
```
odersky: Yes, but even for macro-generated apply's I would assume that we should not eta-expand (since the eta-expansion would simply undo what was generated).

docs/docs/reference/dropped/auto-apply.md
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+Methods overriding Java or Scala-2 methods are again exempted from this
+requirement.
+
+### Migrating code
+
+Existing Scala code with inconsistent parameters can still be compiled
+in Dotty under `-language:Scala2`. When paired with the `-rewrite`
+option, the code will be automatcally rewritten to conforrm to Dotty's
```
liufengyun: Typo: conforrm -> conform.

PR #2722, opened 2017-06-09
Fix positions of enum cases

smarter:
This is probably still not 100% correct, but makes enum basically usable
in the IDE (go to definition works inside an enum now)

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
+++ b/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
@@ -161,8 +161,8 @@ object DesugarEnums {
       val (tagMeth, scaffolding) = enumTagMeth(CaseKind.Object)
       val impl1 = cpy.Template(impl)(body =
         impl.body ++ List(tagMeth, toStringMeth) ++ registerCall)
-      val vdef = ValDef(name, TypeTree(), New(impl1)).withMods(mods | Final).withPos(pos)
-      flatTree(scaffolding ::: vdef :: Nil).withPos(pos.startPos)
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
```diff
@@ -161,8 +161,8 @@ object DesugarEnums {
       val (tagMeth, scaffolding) = enumTagMeth(CaseKind.Object)
       val impl1 = cpy.Template(impl)(body =
         impl.body ++ List(tagMeth, toStringMeth) ++ registerCall)
-      val vdef = ValDef(name, TypeTree(), New(impl1)).withMods(mods | Final).withPos(pos)
-      flatTree(scaffolding ::: vdef :: Nil).withPos(pos.startPos)
+      val vdef = ValDef(name, TypeTree(), New(impl1)).withMods(mods | Final)
+      flatTree(scaffolding ::: vdef :: Nil).withPos(pos)
```
odersky: Can you add an explanation how that makes a difference? I thought that the previous version would give `pos`
to the actual value definition and `startPos` to scaffolding. Isn't that what we want?

smarter: The previous version called `withPos(pos.startPos)` on the whole tree, since children positions have to be contained inside their parents positions, this means that the whole tree including `vdef` got position `pos.startPos`.

odersky: Ah I see now. There are some subtleties involved here.

PR #2731, opened 2017-06-11
Moved error message to case class scheme

gvolpe:
Moved anonymous function missing parameter type error message to case class scheme.

Corresponds to Typer.scala:770 listed on the main issue: #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -176,6 +176,34 @@ object messages {
            |Or, add an explicit `()' as a parameter list to ${cdef.name}."""
   }
 
+  case class AnonymousFunctionMissingParamType(param: untpd.ValDef,
+                                               args: List[untpd.Tree],
+                                               tree: untpd.Function,
+                                               pt: Type)
+                                              (implicit ctx: Context)
+  extends Message(AnonymousFunctionMissingParamTypeID) {
+    val kind = "Syntax"
+
+    val msg = {
+      val ofFun =
+        if (MethodType.syntheticParamNames(args.length + 1) contains param.name)
+          i" of expanded function $tree"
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -176,6 +176,34 @@ object messages {
...
+        else
+          ""
+
+      i"missing parameter type for parameter ${param.name}$ofFun, expected = $pt"
+    }
+
+    val explanation =
+      hl"""|Anonymous functions must define a type. For example, if you define a function like this one:
+           |
+           |val f = { case l@List(1,2,3) => Some(l) }
```
felixmulder: For syntax highlighting this needs to be wrapped:

```diff
+           |${"val f = { case xs @ List(1, 2, 3) => Some(xs) }"}
-           |val f = { case l@List(1,2,3) => Some(l) }
```
Some spaces would also be awesome! 👍

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -176,6 +176,34 @@ object messages {
...
+           |
+           |Make sure you give it a type of what you expect to match and help the type inference system:
+           |
+           |val f: Seq[Int] => Option[List[Int]] = { case l@List(1,2,3) => Some(l) } """
```
felixmulder: Same here with the wrapping :)

gvolpe: Thanks @felixmulder , I'll try to fix it now in my lunch time :)

gvolpe:
Hi @felixmulder , see if now it's fine whenever you can.
Cheers!

felixmulder:
Cool, thanks @gvolpe! :tada:

PR #2733, opened 2017-06-11
Moving string error message to dedicated case class

msosnicki:
Adds a case class to  PostTyper.scala:182 mentioned here: #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -176,6 +176,34 @@ object messages {
            |Or, add an explicit `()' as a parameter list to ${cdef.name}."""
   }
 
+  case class AnonymousFunctionMissingParamType(param: untpd.ValDef,
+                                               args: List[untpd.Tree],
+                                               tree: untpd.Function,
+                                               pt: Type)
+                                              (implicit ctx: Context)
+  extends Message(AnonymousFunctionMissingParamTypeID) {
+    val kind = "Syntax"
+
+    val msg = {
+      val ofFun =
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1580,4 +1580,10 @@ object messages {
     val explanation = ""
   }
 
+  case class SuperCallsNotAllowedInline(symbol: Symbol)(implicit ctx: Context)
+    extends Message(SuperCallsNotAllowedInlineID) {
+    val kind = "Syntax"
+    val msg = s"super call not allowed in inline $symbol"
+    val explanation = ""
```
felixmulder: Maybe it would be good to have an explanation here on why `inline` inherently disallows super calls.

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -803,4 +803,23 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
 
+  @Test def superCallsNotAllowedInline =
+  checkMessagesAfter("refchecks") {
+       """
+        |class A {
+        |  def foo(): Unit = ()
+        |}
+        |
+        |class B extends A {
+        |  @inline def bar(): Unit = super.foo()
```
felixmulder: FYI `inline` is now a keyword in Dotty and could be used here instead - not necessary to get in though :)

msosnicki: Cool, I'll do all the updates later on today, thanks! :)

PR #2740, opened 2017-06-12
Moving string error message to dedicated case class

msosnicki:
Error is related to super calls made from inline methods.


felixmulder:
LGTM, thanks @msosnicki!

PR #2744, opened 2017-06-13
Fix #2743: Fix hover problem

### LINKED ISSUE #2743: IDE crash on hover
Steps to reproduce: In the dotty project, open 

    collection-strawman/src/main/scala/strawman/collection/Iterable.scala

Hover on something, e.g. the `foreach` in `partition` on line 267. 

Result:

[error] Caused by: java.util.NoSuchElementException: key not found: file:///Users/odersky/workspace/dotty/collection-strawman/src/main/scala/strawman/collection/Iterable.scala
[error] 	at scala.collection.MapLike$class.default(MapLike.scala:228)
[error] 	at scala.collection.AbstractMap.default(Map.scala:59)
[error] 	at scala.collection.MapLike$class.apply(MapLike.scala:141)
[error] 	at scala.collection.AbstractMap.apply(Map.scala:59)
[error] 	at dotty.tools.languageserver.DottyLanguageServer.hover$$anonfun$1(DottyLanguageServer.scala:300)
[error] 	at dotty.tools.languageserver.DottyLanguageServer.liftedTree1$1(DottyLanguageServer.scala:108)
[error] 	at dotty.tools.languageserver.DottyLanguageServer.computeAsync$$anonfun$1(DottyLanguageServer.scala:113)

odersky:
Also add an assert in isRef to track down a stackoverflow going infinitely often
through this line. Hopefully the assert will track that down.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -125,7 +125,7 @@ object Types {
     def isRef(sym: Symbol)(implicit ctx: Context): Boolean = stripAnnots.stripTypeVar match {
       case this1: TypeRef =>
         this1.info match { // see comment in Namer#typeDefSig
-          case TypeAlias(tp) => tp.isRef(sym)
+          case TypeAlias(tp) => assert(tp ne this); tp.isRef(sym)
           case _ =>  this1.symbol eq sym
         }
       case this1: RefinedOrRecType => this1.parent.isRef(sym)
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -125,7 +125,7 @@ object Types {
     def isRef(sym: Symbol)(implicit ctx: Context): Boolean = stripAnnots.stripTypeVar match {
       case this1: TypeRef =>
         this1.info match { // see comment in Namer#typeDefSig
-          case TypeAlias(tp) => assert(tp ne this); tp.isRef(sym)
+          case TypeAlias(tp) => assert((tp ne this) && (tp ne this), tp); tp.isRef(sym)
           case _ =>  this1.symbol eq sym
         }
       case this1: RefinedOrRecType => this1.parent.isRef(sym)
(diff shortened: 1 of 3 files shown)


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -297,7 +297,7 @@ class DottyLanguageServer extends LanguageServer
     implicit val ctx = driver.currentCtx
 
     val pos = sourcePosition(driver, uri, params.getPosition)
-    val tp = Interactive.enclosingType(driver.openedTrees(uri), pos)
+    val tp = Interactive.enclosingType(driver.openedTrees.getOrElse(uri, Nil), pos)
```
smarter: There's many calls to `openedTrees(uri)` in this file, so instead of fixing it here I would change the definition of `openedTrees`.

odersky:
@smarter I followed your suggestion. What about openedFiles? I thought of having that return a NoSource as default, but got pause by this comment in InteractiveDriver:

    val source = driver.openedFiles(uri) // might throw exception

So we want it to throw that exception?

smarter:
No that comment was more here to make me remember to decide how to handle the error cases, but it looks like I dodnt

odersky:
OK, openedFiles gets a default as well.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -125,7 +125,7 @@ object Types {
     def isRef(sym: Symbol)(implicit ctx: Context): Boolean = stripAnnots.stripTypeVar match {
       case this1: TypeRef =>
         this1.info match { // see comment in Namer#typeDefSig
-          case TypeAlias(tp) => tp.isRef(sym)
+          case TypeAlias(tp) => assert((tp ne this) && (tp ne this), tp); tp.isRef(sym)
```
smarter: Same condition repeated twice?

odersky: Should be this1. I think it was a merge breakage.

PR #2747, opened 2017-06-13
Fix #1589: Reworks with explanation error:modifiers(s) not allowed for 'item'

### LINKED ISSUE #1589: Missing error messages

johnregan:
Parsers-1685 :  Sample compilation error

```
-- [E082] Syntax Error: /home/jr/projects/dotty/tests/neg/validate-parsing.scala:4:7 
4 |  lazy trait T // error: modifier(s) `lazy' not allowed for trait
  |       ^^^^^
  |       modifier(s) `lazy' not allowed for trait
```

Explanation
===========
You tried to use a modifier that is inapplicable for the type of item under modification

Consider the following example:
sealed def y: Int = 1
In this instance, the modifier 'sealed' is not applicable to the item type 'def' (method)


johnregan:
@felixmulder  can you take a look ?

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -89,8 +89,8 @@
     OnlyCaseClassOrCaseObjectAllowedID,
     ExpectedClassOrObjectDefID,
     AnonymousFunctionMissingParamTypeID,
-    SuperCallsNotAllowedInlineID
-    ;
```
felixmulder: The idea with having this on a separate line is that we won't get a `-` diff when adding an ID

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1614,4 +1615,20 @@ object messages {
...
+    val kind = "Syntax"
+    val msg = s"modifier(s) `$flags' not allowed for $sort"
+    val explanation = {
+      val code = "sealed def y: Int = 1"
+      hl"""You tried to use a modifier that is inapplicable for the type of item under modification
+         |
+         |
+         |Consider the following example:
+         |$code
+         |In this instance, the modifier 'sealed' is not applicable to the item type 'def' (method)
```
felixmulder: We could also possibly try to identify which flags have been applied to what and make a suggestion. That would be cool 😄

johnregan: @felixmulder not quite sure how to identify all modifiers for traits,methods,variables. Should this not be handled instead by the next error,
 1693         syntaxError(s"illegal modifier combination: ${mods.flags} and $flag")

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -90,7 +90,8 @@
     ExpectedClassOrObjectDefID,
     AnonymousFunctionMissingParamTypeID,
     SuperCallsNotAllowedInlineID,
-    ModifiersNotAllowedID;
+    ModifiersNotAllowedID
```
nicolasstucki: You should leave the trailing `,` at the end so the next diff wont change this line
```java
  ...
  AnonymousFunctionMissingParamTypeID,
  SuperCallsNotAllowedInlineID,
  ModifiersNotAllowedID,
  ;
```

johnregan: ok thanks nicolass

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
felixmulder: Hmm, I think these two should be harmonized. I.e. the same error message for both. You could make `sort: Option[String]` to discern between the `case _` and the rest of them.

Otherwise this LGTM now :)

johnregan: Cool, I can take that one next. Can we agree that it will be handled with the 1693 consolidation refactor and get this one through? :)

felixmulder: Sure - that's fine by me - assigning you in #1589

PR #2750, opened 2017-06-14
Add VSCode settings for the dotty project

nicolasstucki:


smarter:
I think it would break it (unless you use git update-index --assume-unchanged: http://gitready.com/intermediate/2009/02/18/temporarily-ignoring-files.html), but on the other hand, this appears to be a common things that people do, vscode itself defines settings that way: https://github.com/Microsoft/vscode/blob/master/.vscode/settings.json. So I assume that it's not common to want custom settings for one project, though someone asked for a "settings.local.json" before: https://github.com/Microsoft/vscode/issues/17634#issuecomment-277957930

smarter:
one option would be to do as the comment I linked to suggest, have a .vscode-template automatically copied to .vscode on first run (by sbt?).

nicolasstucki:
that could work

project/Build.scala
```diff
@@ -1147,6 +1147,13 @@ object Build {
            |  > git submodule update --init
         """.stripMargin)
     }
+
+    val vscodeSetting = new File(".vscode/settings.json")
+    if(!vscodeSetting.exists()) {
+      val vscodeSettingTemplate = new File(".vscode/settings-template.json")
```
smarter: I would use a .vscode-template directory instead:
- We might want to add settings for other files in .vscode
- This means that it's safe to gitignore everything in .vscode

smarter: I would also add a comment explaining what this does.

nicolasstucki: Good idea

.vscode-template/settings.json
```diff
@@ -0,0 +1,20 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "editor.tabSize": 2,
+  "editor.insertSpaces": true,
+
+  "files.trimTrailingWhitespace": true,
+
+  "search.exclude": {
+    "**/*.class": true,
```
smarter: We probably want to exclude **/target too.

.vscode-template/settings.json
```diff
@@ -0,0 +1,20 @@
...
+    "**/*.hasTasty": true,
+    "scala2-library/{doc,docs,lib,META-INF,scripts,spec,test,tools}/**": true, // only allow scala-backend/src
+    "scala2-library/src/[abcefimprs]**": true,  // only allow scala-backend/src/library
```
smarter: Do you really need two`*` here?

.vscode-template/settings.json
```diff
@@ -0,0 +1,20 @@
...
+    "scala-backend/{doc,docs,lib,META-INF,scripts,spec,test,tools}/**": true, // only allow scala-backend/src
+    "scala-backend/src/[abefilmprs]**": true, // only allow scala-backend/src/compiler
+    "scala-backend/src/scala/reflect/**": true,
```
smarter: No need for `**` at the end since we're ignoring the whole folder.

project/Build.scala
```diff
@@ -1147,6 +1147,15 @@ object Build {
            |  > git submodule update --init
         """.stripMargin)
     }
+
+    // If .vscode does not exist yet, initialize it with the contents of .vscode-template/
+    val vscodeSetting = new File(".vscode/")
+    if(!vscodeSetting.exists()) {
```
smarter: Missing space after if

project/Build.scala
```diff
@@ -1147,6 +1147,15 @@ object Build {
...
+      vscodeSetting.mkdir()
+      for (file <- new File(".vscode-template/").listFiles)
```
smarter: you could use sbt.IO.copyDirectory

.vscode-template/settings.json
nicolasstucki: Done

nicolasstucki:
Fixed all comments

felixmulder:
@smarter - can this go in now?

DarkDimius:
@smarter could you please review it today so we can get it into next milestone?

smarter:
Sure, but note that this only affects users of the dotty repo, it changes nothing for end users, so it doesn't really matter whether it gets in or not.

nicolasstucki:
As @smarter said, this is just for us. Therefore it is not a priority.

.vscode-template/settings.json
```diff
@@ -0,0 +1,21 @@
...
+{
+  "editor.tabSize": 2,
+  "editor.insertSpaces": true,
+
+  "files.trimTrailingWhitespace": true,
+
+  "search.exclude": {
+    "**/*.class": true,
+    "**/*.hasTasty": true,
+    "**/target/**": true,
```
smarter: I think this regexp should be equivalent to just `target/`

.vscode-template/settings.json
```diff
@@ -0,0 +1,21 @@
...
+    "scala2-library/{doc,docs,lib,META-INF,scripts,spec,test,tools}/*": true, // only allow scala-backend/src
```
smarter: Can you end this regexp with `/` instead of `/*` ? Same for various other regexps below.

project/Build.scala
```diff
@@ -1147,6 +1147,10 @@ object Build {
            |  > git submodule update --init
         """.stripMargin)
     }
+
+    // If contents .vscode do not exist yet, initialize them with the contents of .vscode-template/
```
smarter: I would say "Copy default configuration from .vscode-template unless configuration files already exist in .vscode"

PR #2751, opened 2017-06-14
Detect direct cycles of typerefs referring to themselves

odersky:
Have a Config option that allows to flag as assertion errors
typerefs that refer directly to themselves in a bound or alias.
I am going to use this to track down isRef stackoverflows in the IDE; I believe
it is also useful to keep around in case similar errors appear later.

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/Config.scala
+++ b/compiler/src/dotty/tools/dotc/config/Config.scala
@@ -111,7 +111,7 @@ object Config {
   /** If this flag is set, it is checked that TypeRefs don't refer directly
    *  to themselves.
    */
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/config/Config.scala
```diff
@@ -108,6 +108,11 @@ object Config {
    */
   final val checkMethodTypes = false
 
+  /** If this flag is set, it is checked that TypeRefs don't refer directly
+   *  to themselves.
+   */
+  final val checkTypeRefCycles = true
```
DarkDimius: should be false by default

PR #2755, opened 2017-06-14
Add -optimise as a separate matrix test

OlivierBlanvillain:


.drone.yml
```diff
@@ -76,3 +76,5 @@ matrix:
       CI_PUBLISH: false
     - CI_TEST: dotty-bootstrapped/test
       CI_PUBLISH: false
+    - CI_TEST: ;set scalacOptions in dotty-compiler-bootstrapped += "-optimise"; dotty-bootstrapped/test
```
smarter: Need to write ```in `dotty-compiler-bootstrapped` ```

PR #2756, opened 2017-06-14
Add idempotency check on bootstrapped dotty

nicolasstucki:


tests/idempotency/Checker.scala
```diff
@@ -0,0 +1,10 @@
+
+import java.nio.file.{Files, Path, Paths}
+import java.util.stream.{ Stream => JStream }
```
felixmulder: Why spaces in the bottom one and no spaces in the top one? :)

tests/idempotency/BootstrapChecker.scala
```diff
@@ -0,0 +1,8 @@
+import java.nio.file.{Files, Path, Paths}
+import java.util.stream.{ Stream => JStream }
```
felixmulder: See comment below

PR #2757, opened 2017-06-14
Fix #2677: Fix type of super selection

### LINKED ISSUE #2677: Super-call in intersection type not type safe
Consider:
```scala
> trait A { def x = "foo" }
> trait B { def x = 42 }
> val AB = new A with B { override def x = super.x }
> AB.x
```

I would expect the overriding method to be rejected, since it has type `String` and it is expected to be `Int & String`.

odersky:
Related fix: base classes (i.e. linearization) of AndTypes.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -384,7 +384,7 @@ object Types {
 
     /** The base classes of this type as determined by ClassDenotation
      *  in linearization order, with the class itself as first element.
-     *  For AndTypes/OrTypes, the union/intersection of the operands' baseclasses.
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -394,7 +394,10 @@ object Types {
         case tp: ClassInfo =>
           tp.cls.baseClasses
         case AndType(tp1, tp2) =>
-          tp1.baseClasses union tp2.baseClasses
```
smarter: Seems like a good reason to not have `union` be defined on `List` :)

odersky:
Replaced by #2768.

Issue #2728, opened 2017-06-10
Overlapping Edit Button on docs


Varunram:
Now that we have a "dit this page on Github" option, this seems to collide with large headings (More than 35 characters long). A fix that occurred to me was was replacing "Github" with the cat icon. What might be an ideal solution for this?

<img width="1280" alt="screen shot 2017-06-07 at 6 12 39 pm" src="https://user-images.githubusercontent.com/18598959/27001626-7f90a36a-4dec-11e7-87bb-24fc67811e2c.png">

felixmulder:
Hmm, I'd rather we changed the layout of the title. Maybe something like floating two elements right and when the text overlaps we ensure that they're split to different lines. This would also eliminate some of the media query code.

Another option is to look into flexboxing it instead.

WDYT, can you work with this restriction? :)

Varunram:
Yep, will work on it. I'll put in a PR maybe tomorrow :)

PR #2758, opened 2017-06-15
Fix Overlapping Edit Button on docs

Varunram:
Intended as a fix for #2728. I have a small concern though, do we declare a separate class or go ahead with the existing modification? CC: @felixmulder


Varunram:
Thanks!

PR #2765, opened 2017-06-15
Fix #2759: support params like @file.txt for dotc

### LINKED ISSUE #2759: Expanding cli args with `dotc @file.txt` is unsupported
```
$ cat foo.args
-help
$ scalac @foo.args
Usage: scalac <options> <source files>
where possible standard options include:
  -Dproperty=value                     Pass -Dproperty=value directly to the runtime system.
...
$ dotc @foo.args
Exception in thread "main" java.lang.UnsupportedOperationException: class dotty.tools.dotc.config.CompilerCommand$.expandArg
	at dotty.tools.dotc.util.DotClass.unsupported(DotClass.scala:10)
```

It would be nice if expanding args from a file was possible. This feature is used with scalac to run the compiler benchmark on real-world projects, see https://github.com/scala/scala-dev/issues/392

See: 
https://github.com/lampepfl/dotty/blob/8ddfcaf1667b8972e70e769feffef240ee0bbc10/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala#L40

liufengyun:
Fix #2759: support params like @file.txt for dotc

The command line parser is migrated from Scalac for handling possible quotes in params.

A test case for this feature is here, it works on Linux/Mac/Windows: https://github.com/lampepfl/packtest/pull/4

### DIFF
--- a/compiler/test/dotty/tools/dotc/CompilationTests.scala
+++ b/compiler/test/dotty/tools/dotc/CompilationTests.scala
@@ -208,7 +208,8 @@ class CompilationTests extends ParallelTesting {
       defaultOutputDir + "lib/src/:" +
       // as well as bootstrapped compiler:
       defaultOutputDir + "dotty1/dotty/:" +
-      Jars.dottyInterfaces
+      Jars.dottyInterfaces,
+      "-Ycheck-reentrant"
     )
 
     def lib =
--- a/compiler/src/dotty/tools/dotc/config/CommandLineParser.scala
+++ b/compiler/src/dotty/tools/dotc/config/CommandLineParser.scala
@@ -2,6 +2,7 @@ package dotty.tools.dotc
 package config
 
 import scala.annotation.tailrec
+import dotty.tools.sharable
 
 /** A simple (overly so) command line parser.
  *  !!! This needs a thorough test suite to make sure quoting is
@@ -30,7 +31,7 @@ object CommandLineParser {
   }
   private object DoubleQuoted extends QuotedExtractor('"')
   private object SingleQuoted extends QuotedExtractor('\'')
-  private val Word = """(\S+)(.*)""".r
+  @sharable private val Word = """(\S+)(.*)""".r
 
   // parse `in` for an argument, return it and the remainder of the input (or an error message)
   // (argument may be in single/double quotes, taking escaping into account, quotes are stripped)
--- a/compiler/src/dotty/tools/dotc/config/CommandLineParser.scala
+++ b/compiler/src/dotty/tools/dotc/config/CommandLineParser.scala
@@ -1,0 +1,67 @@
+package dotty.tools.dotc
+package config
+
+import scala.annotation.tailrec
+
+/** A simple (overly so) command line parser.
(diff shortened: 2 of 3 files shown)


liufengyun:
Hmm, the legacy CI is not happy:

```
[info] Test dotc.tests.dottyBooted started
possible data race involving globally reachable variable root in class Pattern: java.util.regex.Pattern.Node
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable matchRoot in class Pattern: java.util.regex.Pattern.Node
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable buffer in class Pattern: Array[Int]
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable namedGroups in class Pattern: java.util.Map[String, Integer]
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable groupNodes in class Pattern: Array[java.util.regex.Pattern.GroupHead]
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable capturingGroupCount in class Pattern: Int
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
possible data race involving globally reachable variable localCount in class Pattern: Int
  use -Ylog:checkReentrant+ to find out more about why the variable is reachable.
7 warnings found
7 errors found
[error] Test dotc.tests.dottyBooted failed: java.lang.AssertionError: assertion failed: Wrong # of errors. Expected: 0, found: 7
```

@smarter @felixmulder @DarkDimius  could you please advise?

smarter:
This tells me two things:
- `-Ycheck-reentrant` is not passed to the compiler when compiling dotty in the new tests, that's an issue in the tests, that's bad! /cc @felixmulder 
- `-Ycheck-reentrant` detects that there is a globally reachable value (`Word`) whose class (`Regex`) contains a field of type `Pattern` which has mutable fields. In general this might causes problem in multi-threaded code, but `Pattern` is documented as thread-safe (https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) so this should be fine. You can disable the check by adding the annotation `@sharable` in front of the declaration of  `Word`.

felixmulder:
It should indeed compile with check reentrant. Can it be enabled by this PR?

smarter:
Opened https://github.com/lampepfl/dotty/issues/2769 to track this.

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -208,7 +208,8 @@ class CompilationTests extends ParallelTesting {
       defaultOutputDir + "lib/src/:" +
       // as well as bootstrapped compiler:
       defaultOutputDir + "dotty1/dotty1/:" +
-      Jars.dottyInterfaces
+      Jars.dottyInterfaces,
+      "-Ycheck-reentrant"
```
liufengyun: @felixmulder could you please check if this is the right place to put the check?

PR #2767, opened 2017-06-15
[WIP] Fix baseClasses computation of AndTypes

odersky:
List union is clearly wrong, as it is simply the same as ++.


odersky:
Replaced by #2768.

PR #2768, opened 2017-06-15
Fix #2677: Several fixes involving super selection

### LINKED ISSUE #2677: Super-call in intersection type not type safe

odersky:
This was much harder than I thought. In the end, there are several fixes that interact in interesting ways with each other.

 1. We fix the way baseclasses are computed in AndTypes.
 2. We fix the way members of supertypes are computed.
 3. We keep the order of operands of a `&` invariant. 

The problems in (1) and (3) conspired to cancel each other out. So once I fixed (1) the problem caused by (3) became apparent, but in a very indirect way. Details are in the commit message of
0f708eb.


OlivierBlanvillain:
Failure reproduced with `sbt dotty-bootstrapped/vulpix t1751`

odersky:
The problem is we need to know in what order the files for `dotc` itself were compiled. Unfortunately it's that order that makes a difference between a compiler that works and a compiler that crashes. Can you pull that out of the logs on your machine?

tests/pos/Orderings.scala
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  def isLess[T]: T => T => implicit Ord[T] => Boolean =
+    x => y => implicitly[Ord[T]].less(x, y)
+}
+class Misc {
+  class Context
+  implicit val ctx: Context
+
+  f(2)               // OK
```
smarter: The definition of `f` appears to be missing from this test.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2307,6 +2300,37 @@ object Types {
...
+          case bc2 :: bcs2rest =>
+            if (isAnd)
+              if (bcs1set contains bc2)
+                if (bc2.is(Trait)) recur(bcs2rest)
+                else bcs1 // common class, therefore rest is the same in both sequences
+              else bc2 :: recur(bcs2rest)
+            else
+              if (bcs1set contains bc2)
+                if (bc2.is(Trait)) bc2 :: recur(bcs2rest)
+                else bcs2rest
```
smarter: Shouldn't this be `bcs2` ? Otherwise we miss `bc2` even though it's contained in the baseclasses of both operands

odersky: Good catch!

PR #2773, opened 2017-06-15
Fix "Find all references" in the IDE

smarter:


smarter:
Got an oral LGTM, merging.

PR #2776, opened 2017-06-16
Make the IDE work under Windows

smarter:
Also, when running under OS X, it should no longer be necessary to put `code` on your PATH as `open -a "Visual Studio Code"` will be used.
Also, always pass `-n` to Code to always start a new instance, this makes things more reproducible.

@felixmulder Can you check for me that everything still works OK under OS X? Just run `sbt launchIDE` in dotty and in dotty-example-project and check that they both work as expected.

### DIFF
--- a/docs/docs/usage/ide-support.md
+++ b/docs/docs/usage/ide-support.md
@@ -18,7 +18,9 @@ Usage
 =====
 1. Install [Visual Studio Code](https://code.visualstudio.com/).
 2. Make sure `code`, the binary for Visual Studio Code, is on your `$PATH`, this
-   is the case if you can start the IDE by running `code` in a terminal.
+   is the case if you can start the IDE by running `code` in a terminal. This
+   is the default on all systems except Mac where you'll need to follow these
+   instructions: https://code.visualstudio.com/docs/setup/mac#_command-line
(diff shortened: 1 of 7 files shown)


felixmulder:
@smarter - it doesn't correctly open the project on macOS, it opens a new code instance with an "untitled" document.

If I manually open the dotty dir it launches the dotty-language server, but seems to be missing a lot of features - only does "rename all occurrences" on right click, that means the servers not really running right?

smarter:
Yes, what do you see in the console?

smarter:
@felixmulder I've pushed a potential fix, let me know if it works now.

smarter:
Actually that's probably not going to help since this code is not run when using launchIDE from the dotty project itself. Can you please give me the output of your console?

smarter:
OK, thanks to @liufengyun I was able to debug the issue, I'll update the PR soon. Basically we cannot use `open` on OS X because we need to do `code --install-extension ...` and that open a new window

smarter:
@felixmulder Updated, let me know if everything works OK on OS X now :)

liufengyun:
@smarter After setting VSCode in PATH, it works well for the dotty project on my Mac 👍

PR #2779, opened 2017-06-18
Move "type argument must be fully defined" message to case class scheme

gvolpe:
Move "type argument must be fully defined" message to new error message scheme using WilcardOnTypeArgumentNotAllowedOnNew.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -91,6 +91,7 @@ public enum ErrorMessageID {
     AnonymousFunctionMissingParamTypeID,
     SuperCallsNotAllowedInlineID,
     ModifiersNotAllowedID,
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -91,6 +91,7 @@
     AnonymousFunctionMissingParamTypeID,
     SuperCallsNotAllowedInlineID,
     ModifiersNotAllowedID,
+    WildcardOnTypeArgumentNotAllowedOnNewID
```
felixmulder: This line should end in `,` so that the diff is single lines :)

(omg, java has trailing commas! :fire: :fire: 🔥 )

gvolpe: Done! :)

felixmulder:
Great, will merge once tests pass (=

PR #2781, opened 2017-06-19
Fix #2772: Special case Devalify for java.lang.System.*

### LINKED ISSUE #2772: Local optimisations inline too eagerly: System.out is final but can still be mutated by System.setOut
```scala
import java.io.OutputStream

object Test {
    val oldOut = System.out
    System.setOut(null)
    val a = () => foo(oldOut)
    a()
  }

  def foo(out: OutputStream): Unit = {
    out.write(0)
  }
}
```
Without `-optimise` this code works, with `-optimise` we get a NullPointerException.
Without `-optimise`:
```scala
result of try/inout.scala after genBCode:

package <empty> {
  @scala.annotation.internal.SourceFile("try/inout.scala") final module class 
    Test$
   extends Object { 
    def <init>(): Unit = 
      {
        super()
        ()
      }
      {
        val oldOut: java.io.PrintStream = System.out
        System.setOut(null)
        val a: Function0 = 
          closure(oldOut | Test.$anonfun$1:scala.compat.java8.JFunction0$mcV$sp)
        {
          a.apply()
          ()
        }
      }
    def foo(out: java.io.OutputStream): Unit = out.write(0)
    private def $anonfun$1(oldOut$1: java.io.PrintStream): Unit = 
      Test.foo(oldOut$1)
  }
}
```
With `-optimise`:
```scala
result of try/inout.scala after genBCode:

package <empty> {
  @scala.annotation.internal.SourceFile("try/inout.scala") final module class 
    Test$
   extends Object { 
    def <init>(): Unit = 
      {
        super()
        ()
      }
      {
        System.setOut(null)
        val a: Function0 = 
          closure(Test.$anonfun$1:scala.compat.java8.JFunction0$mcV$sp)
        {
          a.apply()
          ()
        }
      }
    def foo(out: java.io.OutputStream): Unit = out.write(0)
    private def $anonfun$1(): Unit = Test.foo(System.out)
  }
}
```
Notice that the anonymous function now calls `System.out` instead of `out`.

This happened in the Dotty Language Server (https://github.com/lampepfl/dotty/blob/master/language-server/src/dotty/tools/languageserver/Main.scala#L32-L35), the result is that outside projects cannot use the IDE with `scalaVersion := 0.2.0-bin-20170614-8ddfcaf-NIGHTLY`.

OlivierBlanvillain:
@smarter thanks for the nice minimization!

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
+++ b/compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
@@ -164,7 +164,17 @@ class Devalify extends Optimisation {
     case _ => t
   }
 
-  def readingOnlyVals(t: Tree)(implicit ctx: Context): Boolean = dropCasts(t) match {
+  def readingOnlyVals(t: Tree)(implicit ctx: Context): Boolean = {
+    def isGetterOfAImmutableField = t.symbol.isGetter && !t.symbol.is(Mutable)
+    def isCaseClassWithVar        = t.symbol.info.decls.exists(_.is(Mutable))
+    def isAccessingProductField   = t.symbol.exists                               &&
+                                    t.symbol.owner.derivesFrom(defn.ProductClass) &&
+                                    t.symbol.owner.is(CaseClass)                  &&
+                                    t.symbol.name.isSelectorName                  &&
+                                    !isCaseClassWithVar // Conservatively covers case class A(var x: Int)
+    def isImmutableCaseAccessor   = t.symbol.is(CaseAccessor) && !t.symbol.is(Mutable)
+
+    dropCasts(t) match {
     case Typed(exp, _) => readingOnlyVals(exp)
 
     case TypeApply(fun @ Select(rec, _), List(tp)) =>
@@ -173,29 +183,21 @@ class Devalify extends Optimisation {
       else false
 
     case Apply(Select(rec, _), Nil) =>
-      def isGetterOfAImmutableField = t.symbol.isGetter && !t.symbol.is(Mutable)
-      def isCaseClassWithVar        = t.symbol.info.decls.exists(_.is(Mutable))
-      def isAccessingProductField   = t.symbol.exists                               &&
-                                      t.symbol.owner.derivesFrom(defn.ProductClass) &&
-                                      t.symbol.owner.is(CaseClass)                  &&
-                                      t.symbol.name.isSelectorName                  &&
-                                      !isCaseClassWithVar // Conservative Covers case class A(var x: Int)
-      def isImmutableCaseAccessor   = t.symbol.is(CaseAccessor) && !t.symbol.is(Mutable)
       if (isGetterOfAImmutableField || isAccessingProductField || isImmutableCaseAccessor)
         readingOnlyVals(rec)
       else false
 
     case Select(rec, _) if t.symbol.is(Method) =>
-      if (t.symbol.isGetter && !t.symbol.is(Mutable))
-        readingOnlyVals(rec) // getter of a immutable field
-      else if (t.symbol.owner.derivesFrom(defn.ProductClass) && t.symbol.owner.is(CaseClass) && t.symbol.name.isSelectorName) {
+      if (isGetterOfAImmutableField)
+        readingOnlyVals(rec) // Getter of an immutable field
+      else if (isAccessingProductField) {
         def isImmutableField = {
           val fieldId = t.symbol.name.toString.drop(1).toInt - 1
           !t.symbol.owner.caseAccessors(ctx)(fieldId).is(Mutable)
         }
-        if (isImmutableField) readingOnlyVals(rec) // accessing a field of a product
+        if (isImmutableField) readingOnlyVals(rec) // Accessing a field of a product
         else false
-      } else if (t.symbol.is(CaseAccessor) && !t.symbol.is(Mutable))
+      } else if (isImmutableCaseAccessor)
         readingOnlyVals(rec)
       else false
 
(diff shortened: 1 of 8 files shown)


compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
```diff
@@ -194,13 +198,22 @@ class Devalify extends Optimisation {
...
-      readingOnlyVals(qual)
+      if (t.symbol == defn.SystemModule) {
+        // System.in is static final fields that, for legacy reasons, must be
+        // allowed to be changed by the methods System.setIn...
+        // https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.4
+        false
+      } else
+        readingOnlyVals(qual)
+
     case t: Ident if !t.symbol.is(Mutable) && !t.symbol.is(Method) && !t.symbol.info.dealias.isInstanceOf[ExprType] =>
```
smarter: Nothing says that there might not be an `Ident` corresponding to `System.in`, and in that case you would still inline it. Maybe add a `isReallyMutable` instead that you would use everywhere?

compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
```diff
@@ -182,8 +184,10 @@ class Devalify extends Optimisation {
       if (isGetterOfAImmutableField || isAccessingProductField || isImmutableCaseAccessor)
         readingOnlyVals(rec)
       else false
+
     case Select(rec, _) if t.symbol.is(Method) =>
-      if (t.symbol.isGetter && !t.symbol.is(Mutable)) readingOnlyVals(rec) // getter of a immutable field
+      if (t.symbol.isGetter && !t.symbol.is(Mutable))
+        readingOnlyVals(rec) // getter of a immutable field
```
smarter: typo: a -> an

smarter:
This also won't work if someone uses reflection to mutate a final field. I suggest creating a documentation page for `-optimize` to list all the things you're not allowed to do in your code for it to work.

compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
OlivierBlanvillain: The implementation of this case calls `desugarIdent` and recuse, I think the case where the `Ident` is `System.in` is covered.

@DarkDimius Could you explain why we even need this `desugarIdent`, why not use `Select` everywhere directly? Also where would be the proper was to move that function to factor out the version that's currently copy pasted in the backend?

OlivierBlanvillain:
I went thought all usages of `Mutable` inside the `localopt` package and I found another potential but in `DropGoodCasts`, thanks for the suggestion! (All other uses of is(Mutable) shouldn't be affect by this System.stuff because it checks on safe things such as ValDefs).

OlivierBlanvillain:
(I will later properly populate this documentation with content from the semester project report)

compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
```diff
@@ -164,7 +164,17 @@ class Devalify extends Optimisation {
     case _ => t
   }
 
-  def readingOnlyVals(t: Tree)(implicit ctx: Context): Boolean = dropCasts(t) match {
+  def readingOnlyVals(t: Tree)(implicit ctx: Context): Boolean = {
+    def isGetterOfAImmutableField = t.symbol.isGetter && !t.symbol.is(Mutable)
```
smarter: Of _An_ ImmutableField

compiler/src/dotty/tools/dotc/transform/localopt/Simplify.scala
```diff
@@ -146,4 +146,15 @@ object Simplify {
       case _ => None
     }
   }
+
+  // System.in is static final fields that, for legacy reasons, must be
```
smarter: is -> is a
fields -> field

This description is a bit confusing: it starts by talking just about System.in and then mention the rest.
The name of the method is also not great, it's too easy to confuse with `is(Mutable)`, I would call it something like `isEffectivelyMutable`, `isReallyMutable` or `isChanging`

compiler/src/dotty/tools/dotc/transform/localopt/Simplify.scala
```diff
@@ -146,4 +146,15 @@ object Simplify {
...
+  // allowed to be changed by the methods System.setIn. `isMutable` is true
+  // is the field is Mutable or if it's a field of java.lang.System.
+  // https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.4
+  def isMutable(t: Tree)(implicit ctx: Context): Boolean = t match {
+    case _ if t.symbol.is(Mutable) => true
+    case s: Symbol => (s.symbol == defn.SystemModule)
```
smarter: I think you want s.owner.symbol here.

docs/docs/reference/optimisations.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: doc-page
+title: "Local Optimisations"
+---
+
+The Dotty compiler implements several local optimisations to generate faster bytecode. These optimisations can be enabled by compiling with the `-optimise` flag. The current best source of documentation about what tree transformations are performed under `-optimise` is the Scaladoc classes in the [localopt package](https://github.com/lampepfl/dotty/tree/master/compiler/src/dotty/tools/dotc/transform/localopt).
+
+Local optimisations assumes no use of Java Reflection to mutate static final fields.
```
smarter: What about non-static final fields, are those safe to mutate?

compiler/src/dotty/tools/dotc/transform/localopt/Simplify.scala
```diff
@@ -146,4 +146,15 @@ object Simplify {
...
     }
   }
+
+  /** Is this tree mutable, or java.lang.System.{in, out, err}? These three
+   *  System members are the only static final fields that are mutable.
+   *  See https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.4
+   */
+  def isEffectivelyMutable(t: Tree)(implicit ctx: Context): Boolean = t match {
+    case _ if t.symbol.is(Mutable) => true
+    case s: Select => (s.symbol == defn.SystemModule)
```
smarter: This still looks wrong to me, `s` may be `System.in`, so you need to look at the owner

PR #2786, opened 2017-06-20
Fix #2698: Move eqXYZ instances to Eq

### LINKED ISSUE #2698: Move `Eq` instances into scala.Eq companion object?
Discussion after https://www.reddit.com/r/programming/comments/6elim6/announcing_dotty_012rc1_a_major_step_towards/ points out that the extra implicit instances of `Eq` infect the global namespace and that's undesirable — those instances are currently available for autocompletion when typing `e`.

Could one move them to `Eq`'s companion object?

That might require using a different object as default instance, and maybe patching compiler special cases—such changes are included in the scope of this issue.

odersky:


dwijnand:
I think one might want to opt-out of more than just the eqAny implicit, for instance the numeric type Eqs.

If the root issue is auto-completion pollution, then perhaps a solution that doesn't affect the ability to opt-out is to give them more contorted names. Maybe something along the lines of `$dottyEqByteNum`.

odersky:
I think we are going to move everything back to `Eq`. The problem of having `eqAny` in `Predef` is that it will always be tried first and succeed. Then there's a sub-search whether its operands support `Eq`, which will also succeed . Only after going through 3 full implicit searches will we conclude that `eqAny` is _not_ applicable, and will start to search for something more specific.

If we move `eqAny` back to `Eq` the problem goes away because we find it then through the implicit scope, together with all other choices, and we eliminate it immediately, because it is less specific. 

This means that unimporting cannot be used to disable `eqAny`. We can think about other ways to achieve it, though.

odersky:
@OlivierBlanvillain Can you take another look? I did a lot of changes after your last review.

PR #2787, opened 2017-06-20
Harden IDE some more

odersky:
I came across two more sources of crashes when doing heavy editing on Types.scala

- tpd.TreeAccumulator hit again an untyped tree. I believe this is because the tree was left in because of an error in a previous run.  Because the error was in a previous run, we cannot take
`reporter.hasErrors` as a test to avoid the crash.

 - I got a StaleSymbolError through another path as before.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -1305,8 +1305,10 @@ object Trees {
             this(this(x, arg), annot)
           case Thicket(ts) =>
             this(x, ts)
-          case _ if ctx.reporter.errorsReported =>
+          case _ if ctx.mode.is(Mode.Interactive) =>
             // in case of errors it may be that typed trees point to untyped ones.
+            // The can still traverse inside such trees, either in the run where errors
+            // are reported, or in subsequent ones.
             x
         }
       }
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -1305,8 +1305,10 @@ object Trees {
             this(this(x, arg), annot)
           case Thicket(ts) =>
             this(x, ts)
-          case _ if ctx.reporter.errorsReported =>
+          case _ if ctx.mode.is(Mode.Interactive) =>
             // in case of errors it may be that typed trees point to untyped ones.
+            // The can still traverse inside such trees, either in the run where errors
```
smarter: Missing word between "The" and "can"?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1538,7 +1538,8 @@ object Types {
               else recomputeMember(d) // symbol could have been overridden, recompute membership
             else {
               val newd = loadDenot
-              if (newd.exists) newd else d.staleSymbolError
+              if (newd.exists || ctx.mode.is(Mode.Interactive)) newd
```
smarter: Isn't this going to fail with a NoDenotation error later on? If you really want to return the old denotation this should be `d`, not `newd`

Issue #2792, opened 2017-06-22
Cyclic error involving @deprecated in Predef when compiling compileStdLib


abeln:
`vulpix tests/pos` fails on some systems while compiling the standard library:

```
sbt vulpix "tests/pos"
info] Test dotty.tools.dotc.CompilationTests.compileNeg started
[=======================================] completed (281/281, 37s)  
[info] Test dotty.tools.dotc.CompilationTests.compilePos started
-- [E045] Syntax Error: /u4/anietoro/src/dotty/compiler/../scala2-library/src/library/scala/Predef.scala:101:3 ---------
101 |  @deprecated("Use `scala.reflect.ClassTag` instead", "2.10.0")
```

@smarter mentioned a recent [PR](https://github.com/lampepfl/dotty/pull/2193/files) that normalizes the list of files in a test by sorting them. Looking at `compileList` in `ParallelTesting.scala`, it does no such sorting.

Once I sorted the files, the test now passes.

smarter:
Could you attach the log file from the test failure?

abeln:
Test run logs. The failure is on line 7383.

[tests-2017-06-22-15:17:23.txt](https://github.com/lampepfl/dotty/files/1096219/tests-2017-06-22-15.17.23.txt)

abeln:
Actually, I'm able to repro this in more than one machine, with a fresh checkout.

abeln:
To better repro the problem, sort `StdLibSources.whitelisted`, as in https://github.com/abeln/dotty/commit/ddf98890eb4fd7932bae98785e94517ecb7d4b16

smarter:
Duplicate of https://github.com/lampepfl/dotty/issues/2633

PR #2793, opened 2017-06-22
Sort stdlib sources to make bug reproducible 

abeln:
A previous PR (https://github.com/lampepfl/dotty/pull/2193/files)
added sorting of input files to tests, so that the result is deterministic.

The compileDir function was sorting, but compileList was left out.
This was exposing failures while compiling the Predef.

Fix the symptoms (but not the cause of the failure) by sorting in compileList
as well.

Tested:
  Verified that compilePos passes in the system where it was previously failing.

### DIFF
--- a/compiler/test/dotty/tools/StdLibSources.scala
+++ b/compiler/test/dotty/tools/StdLibSources.scala
@@ -17,14 +17,18 @@ object StdLibSources {
 
   def whitelisted: List[String] = {
     lazy val whitelistBasedOnBlacklist = all.diff(blacklisted)
-    if (!useExplicitWhiteList) {
-      whitelistBasedOnBlacklist
-    } else if (!new File(whitelistFile).exists()) {
-      genWhitelist(whitelistBasedOnBlacklist.map(_.replace(stdLibPath, "")))
-      whitelistBasedOnBlacklist
-    } else {
-      loadList(whitelistFile)
-    }
+    val files =
+      if (!useExplicitWhiteList) {
+        whitelistBasedOnBlacklist
+      } else if (!new File(whitelistFile).exists()) {
+        genWhitelist(whitelistBasedOnBlacklist.map(_.replace(stdLibPath, "")))
+        whitelistBasedOnBlacklist
+      } else {
+        loadList(whitelistFile)
+      }
+    // Sort the files so their compilation order is deterministic, and
+    // https://github.com/lampepfl/dotty/issues/2792 can be repro'ed.
+    files.sorted
   }
 
   def blacklisted: List[String] = loadList(blacklistFile)


abeln:
This hides the failure described in https://github.com/lampepfl/dotty/issues/2792.

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -1088,7 +1088,10 @@ trait ParallelTesting extends RunnerOrchestration { self =>
     targetDir.mkdirs()
     assert(targetDir.exists, s"couldn't create target directory: $targetDir")
 
-    val target = JointCompilationSource(s"$testName from $callingMethod", files.map(new JFile(_)).toArray, flags, targetDir)
+    // Sort the files so that the compilation order is deterministic.
+    // See https://github.com/lampepfl/dotty/issues/2792
+    val sorted = files.map(new JFile(_)).toArray.sorted
```
smarter: https://github.com/lampepfl/dotty/pull/2193/files already has some code to sort things with some specific seed, it would be nice to do some refactoring so that it is used here too.

abeln: I saw it, but it seemed a bit premature to add a seed argument to `compileList` that isn't currently used by any of the callers. If somebody needs it down the line, we can always refactor then. I don't feel strongly about this, so happy to change it if you want.

smarter: In fact, I just remembered why we don't sort in `compileList`: sometimes we want to test a particular order. The issue needs to be fixed earlier: `compileStdLib` calls `StdLibSources.whitelisted` and that needs to be sorted.

abeln:
Err... I thought this fixed things, but it doesn't.
It looks like `vulpix tests/pos` and `test-only -- --tests=compilePos` are running different tests. The former passes; the latter fails.

abeln:
More specifically, `test-only -- --tests=compileStdLib` fails.

abeln:
I don't think sorting the input will fix/hide the problem.

smarter:
Wait! I think you're fix was correct :). `test-only -- --tests=compileStdLib` does indeed run, it's this one: https://github.com/lampepfl/dotty/blob/master/compiler/test/dotc/tests.scala#L217 This file contains our old test-suite that we are still keeping around until we are more confident that our new test framework is bug-free. I suggest changing the output of `def whitelisted` in `StdlibSources` to be sorted so you'll only have to change one place.

abeln:
Right. Before closing the PR I sorted the files in both places, but both tests (via `vulpix` and `test-only`) still fail.

smarter:
Notice that it also fails on the CI now: this is actually a good thing, this means that the sorted order allows us to deterministically reproduce the failure.

abeln:
Ok, in that case PTAL. But will you merge a change that breaks the CI? Or just use the sorted order while debugging the problem?

smarter:
Just use it for debugging.

abeln:
Ok, then I'll close this :)

PR #2798, opened 2017-06-23
Fix #2795: Fix name kind testing logic

### LINKED ISSUE #2795: Missing/abstract param accessor in implicit parameter of trait
```scala
trait Foo[T: ClassTag]() {
  def foo(x: T) = Array(x)
}
class Bar extends Foo[Int]()
```
fails with a `java.lang.AbstractMethodError: Method Bar.Foo$$evidence$1()Lscala/reflect/ClassTag; is abstract`

Where the tree after `genBCode` looks like

```
<trait> trait Foo extends Object { 
    def $init$(): Unit = ()
    implicit <accessor> def Foo$$evidence$1(): scala.reflect.ClassTag
    def foo(x: Object): Object = 
      Array.apply(Predef.genericWrapArray([x : Object]), this.Foo$$evidence$1())
  }
class Bar
   extends Object with Foo { 
    def <init>(): Unit = 
      {
        super()
        super[Foo].$init$()
        ()
      }
  }
```

odersky:
The previous logic would have UniqueNames shadow everything else. So if one starts
with a UniqueName(prefix, sep, number), and then does a qualified name, one gets a
unique name over a qualified name. The resulting name should still be recognized as a
qualified name, however. The change in this commit achieves that.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Names.scala
+++ b/compiler/src/dotty/tools/dotc/core/Names.scala
@@ -233,9 +233,7 @@ object Names {
       }
     }
 
-    /** Is it impossible that trees of kind `kind` qualify as
-     *  derived trees of kind `shadowed`?
-     */
+    /** Is it impossible that names of kind `kind` also qualify as names of kind `shadowed`? */
     private def shadows(kind: NameKind, shadowed: NameKind): Boolean =
       kind.tag < shadowed.tag ||
       kind.definesQualifiedName ||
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -233,17 +233,25 @@ object Names {
       }
     }
 
+    /** Is it impossible that trees of kind `kind` qualify as
```
smarter: you mean names, not trees, right?

odersky: Oops. Of course.

Issue #2797, opened 2017-06-23
Assertion error in DottyBackendInterface: denot.name is not default getter


olafurpg:
After merging the fix for #2704, I'm hitting a new assertion error in the backend while trying to compile the scala/compiler-benchmark with Dotty.

To reproduce

```
cd dotty
git checkout 0109fef834b4f08ba4dbc0fd08fe6f05753cb9b1
export NIGHTLYBUILD=yes
sbt dotty-bootstrapped/publishLocal
cd target
git clone https://github.com/olafurpg/compiler-benchmark.git
cd compiler-benchmark
git checkout c30610306e9bc7b5ed0609f0f24ca1054a48d26f // branch dotty2
sbt ++0.2.0-bin-20170623-0109fef-NIGHTLY compilation/test
...
[error] Error while emitting ScalacBenchmark.scala
[error] assertion failed
java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:156)
	at dotty.tools.backend.jvm.DottyBackendInterface.emitArgument(DottyBackendInterface.scala:259)
	at dotty.tools.backend.jvm.DottyBackendInterface.emitAssocs$$anonfun$2(DottyBackendInterface.scala:315)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:18)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
	at dotty.tools.backend.jvm.DottyBackendInterface.emitAssocs(DottyBackendInterface.scala:315)
	at dotty.tools.backend.jvm.DottyBackendInterface.emitAnnotations$$anonfun$2(DottyBackendInterface.scala:308)
```
Compiling with -Ycheck:all gives a hint that an error occurred while typechecking this line here:
https://github.com/scala/compiler-benchmark/blob/d0f038e0a9f367e26dbe9825de9d00e9e699f592/compilation/src/main/scala/scala/tools/nsc/ScalacBenchmark.scala#L38

```
exception while typing java.nio.file.Files.walk(this.findSourceDir,
  [java.nio.file.FileVisitOption.FOLLOW_LINKS : java.nio.file.FileVisitOption]:
[...]

PR #2801, opened 2017-06-23
Add debug string to assertion error message.

olafurpg:
This commit changes the assertion error messaage in #2797 from an empty
message to:
```
found:    Array[java.nio.file.FileVisitOption]
[error] required: java.nio.file.FileVisitOption
```


odersky:
LGTM

PR #2802, opened 2017-06-24
Fix nightly build after eta-expansion changes

smarter:
The nightly build runs the sbt incremental compilation tests (that can be
run by hand using sbt "sbt-dotty/scripted source-dependencies/*" but
take a while to complete). The implicit-params test started failing
after #2701 because B.scala contains:
  val y = x(3)
which was supposed to fail when the implicit is not present but now
succeeds with `y` getting the type `E => String`. We restore the failure
by adding a type annotation to `y`

### DIFF
--- a/sbt-dotty/sbt-test/source-dependencies/implicit-params/B.scala
+++ b/sbt-dotty/sbt-test/source-dependencies/implicit-params/B.scala
@@ -1,4 +1,4 @@
 object B extends A
 {
-	val y = x(3)
+	val y: String = x(3)
 }


smarter:
This shows that we really need a warning for pure statements, like `scalac`:
```scala
class IOCapability

object Test {
  // Forgot to mark `ev` implicit!
  def doSideEffects(x: Int)(ev: IOCapability) = {
    println("x: " + x)
  }

  implicit val cap: IOCapability = new IOCapability

  // This will compile but won't actually produce any side-effect
  doSideEffects(1)
}
```
In fact, I wonder if we could make automatic eta-expansion fail when the result is a pure statement

odersky:
Good point about the purity warning. LGTM for this PR.

PR #2804, opened 2017-06-24
Emit "pure expression does nothing in statement position" warning

odersky:


lrytz:
See also https://github.com/scala/bug/issues/10317 for a limitation of the current implementation in scalac, and it seems to be the same here

som-snytt:
footnote that lrytz example warns
```
Welcome to Scala 3.4.0-RC1-bin-SNAPSHOT-git-5454110 (21, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> def f(i: Int): Int = { if (i == 42) 27; 17 }
1 warning found
-- [E190] Potential Issue Warning: -------------------------------------------------------------------------------------
1 |def f(i: Int): Int = { if (i == 42) 27; 17 }
  |                                    ^^
  |                                    Discarded non-Unit value of type Int. You may want to use `()`.
  |
  | longer explanation available when compiling with `-explain`
def f(i: Int): Int
```

PR #2807, opened 2017-06-24
Combines modifiers not allowed with type, and combinations.This also simplifies logic. Adds more detail to explanation

johnregan:


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1680,21 +1680,16 @@ object Parsers {
...
-          mods
+      else {
+        syntaxError(ModifiersNotAllowed(mods.flags, getPrintableTypeFromFlagSet(flag)))
+        Modifiers(flag)
       }
     }
 
+    private def getPrintableTypeFromFlagSet(flag: FlagSet) =
+      Map(Trait -> "trait", Method -> "method", Mutable -> "variable").get(flag)
+
```
felixmulder: Since this method is only used by `addFlag`, I would nest it in the method like:

```scala
def addFlag(mods: Modifiers, flag: FlagSet): Modifiers = {
  def getPrintableTypeFromFlagSet(flag: FlagSet) =
    Map(Trait -> "trait", Method -> "method", Mutable -> "variable").get(flag)

  if (compatible(mods.flags, flag)) mods | flag
  else {
    syntaxError(ModifiersNotAllowed(mods.flags, getPrintableTypeFromFlagSet(flag)))
    Modifiers(flag)
  }
}
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1648,18 +1648,35 @@ object messages {
...
+         |                         AccessModifier
+         |                         `override'         traits, methods
+         |  LocalModifier     ::=  `abstract'         classes, traits
+         |                         `final'            classes, traits, methods
+         |                         `sealed'           classes, traits
+         |                         `implicit'         variables, methods, classes
+         |                         `lazy'             values
+         |  AccessModifier    ::=  (`private' | `protected') [AccessQualifier] methods, classes
+         |  AccessQualifier   ::=  `[' (id | `this') `]'
+         |
```
felixmulder: Having a copy of the syntax here might be nice, but it'd be better if we could link to the spec (we could open an issue for this instead).

johnregan: That makes sense rather than duplicating maintenance effort

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
johnregan: Absolutely

johnregan:
@felixmulder can someone press merge :smile:

felixmulder:
Of course, thanks! :tada:

PR #2810, opened 2017-06-25
Fix "Find all references" in the IDE with external projects

smarter:
"Find all references" worked on dotty itself because all its
dotty-compiled dependencies are internal, but in other projects it
failed when trying to construct a SourceTree for trees in the dotty
library, because sym.sourceFile returned null, because the SourceFile
annotation pointed to a path that does not exist on the user filesystem.

For now, we just discard these trees, but we need to find some way to
match them with source jars.


smarter:
About to catch flight, will open issue later

PR #2813, opened 2017-06-26
Local optimisation: Inline local objects

OlivierBlanvillain:
@DarkDimius This is my take on re-implemented InlineLocalObjects. Why I have now is much simpler, and also does way less. In particular there is no special handling of labeldefs. Could you try coming up with unit test size examples showing up what's lost? I played around with the example from [this gist](https://gist.github.com/DarkDimius/39e8cfe4b24150ca04bd2d622ac8beac), but generate the same byte-code than linker looks impossible without global program knowledge.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/NameKinds.scala
+++ b/compiler/src/dotty/tools/dotc/core/NameKinds.scala
@@ -310,9 +310,8 @@ object NameKinds {
   val PatMatCaseName          = new UniqueNameKind("case")
   val PatMatMatchFailName     = new UniqueNameKind("matchFail")
   val PatMatSelectorName      = new UniqueNameKind("selector")
-  val LocalOptFact            = new UniqueNameKind("fact")
-  val LocalOptSelector        = new UniqueNameKind("selector")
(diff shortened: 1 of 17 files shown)


DarkDimius:
@OlivierBlanvillain, the gist you've referred to was optimized without global program knowledge.
Simplifier from https://github.com/dotty-linker/dotty/tree/opto generated the code in the gist.

DarkDimius:
Here are several examples of code that original InlineLocalObjects can handle but this reimplementation can not:

```
object Tupples{
  def foo(x: Boolean): Int = {
    val (a, b) = if (x) (1, 2) else (2, 3) // handling of IFs is missing
    a + b
  }
 def foo(x: Int): Int = {
    val (a, b) = x match {
        case i if (i%3 == 0) => (i, i + 1)
        case i if (i%2 == 0) => (i, i + 2)
        case _ => (0, 0)
      } // handling of LabelDefs is missing
    a + b
  }
}
```

While this is a nice re-implementation, it does not seem to be able to optimize even simple cases.
@OlivierBlanvillain, did you try it on any non-trivial examples that simply call constructor directly?

PR #2814, opened 2017-06-27
Fix dotc script to correctly handle -verbose.

DarkDimius:


felixmulder:
Does this also work for the released script?

PR #2811, opened 2017-06-26
Fix #2791: add reasons for failure to vulpix

felixmulder:

PR #2819, opened 2017-06-28
Fix #2797: Don't lift Java annotation arguments out of call

### LINKED ISSUE #2797: Assertion error in DottyBackendInterface: denot.name is not default getter

OlivierBlanvillain:
Here is a minimized version, the reproduction requires loading the annotation from a class file:

*2797.java:*

```java
public @interface Fork {
  int value() default -1;
  String[] jvmArgs() default { "nope" };
}
```

*2797.scala:*

```scala
@Fork(jvmArgs = Array("I'm", "hot"))
class SexyScalacBenchmark
```

```bash
rm *.class; javac -d . 2797.java; dotc 2797.scala
```


olafurpg:
Thank you for minimizing @OlivierBlanvillain !

The scala/compiler-benchmark project compiles with this workaround and the unit tests pass. The annotations still seem to be emitted weirdly since I hit a new error when running the jmh benchmarks
```
Annotation @Param is not applicable to methods.
   [org.openjdk.jmh.generators.reflection.RFMethodInfo@20affcd9]

Annotation @Param is not applicable to methods.
   [org.openjdk.jmh.generators.reflection.RFMethodInfo@133b86b1]
```
I don't know if that error is related to this issue or not, the `@Param` annotations is defined on vars here https://github.com/scala/compiler-benchmark/blob/d0f038e0a9f367e26dbe9825de9d00e9e699f592/compilation/src/main/scala/scala/tools/nsc/ScalacBenchmark.scala#L18 It seems the annotation is attached to the wrong symbol

OlivierBlanvillain:
@olafurpg Last commit should moves the annotation from getter/setter methods to the actual field, let me know if that's enough to make JMH happy!

olafurpg:
I can confirm that the JMH benchmark runs from the latest commit! Thank you @OlivierBlanvillain t

olafurpg:
Any chance to get this merged? It would help unblock https://github.com/scala/compiler-benchmark/pull/31

OlivierBlanvillain:
It needs a review, @odersky could you have a look?

DarkDimius:
@odersky  could you please review it today so we can get it into next milestone?
It would be nice to have JMH working with the released jar.

odersky:
@smarter I am not sure how Scala annotations with default arguments are handled in the backend. Maybe @DarkDimius knows more?

odersky:
I tried a translation to Scala; seems to work.

tests/pending/pos/i2797/Fork_1.java
```diff
@@ -0,0 +1,7 @@
+// Test is pending because we have no good way to test it.
```
smarter: I don't understand why this test is pending, since Fork is suffixed with 1 and Test with 2, they should be compiled as you want

odersky: They are not. I get: "Not found: Fork" when I compile Test.

smarter: I've fixed the test. The issue was that the annotation needed to be called `Fork_1` since the name of the file is `Fork_1.java`. The javac error messages are eaten by our test suite, there's a WIP PR to fix that at https://github.com/lampepfl/dotty/pull/2811

smarter:
It might not crash in the backend, but the Scala test case still generates some weird code:
```scala
@{
    <accessor> val value$1: Int = Fork.$lessinit$greater$default$1
    <accessor> val jvmArgs$1: Array{scala$Array$$T = String} = 
      Array.apply[String](["I\'m","hot" : String])(
        scala.reflect.ClassTag.apply[String](classOf[class String])
      )
    new Fork(Test.this.value$1, Test.this.jvmArgs$1)
```

It looks like scalac does something similar but with a warning:
```scala
% scalac tests/pos/i2797a/Fork.scala tests/pos/i2797a/Test.scala
tests/pos/i2797a/Test.scala:4: warning: Usage of named or default arguments transformed this annotation
constructor call into a block. The corresponding AnnotationInfo
will contain references to local values and default getters instead
of the actual argument trees
@Fork(jvmArgs = Array("I'm", "hot"))
 ^
```

smarter:
Had to disable i2797 again because of joint compilation failure, see last commit, I'll open an issue for that.

tests/pending/pos/i2797/Fork_1.java
odersky: Ah, good to know.

odersky:
In fact, we could put i2797 in pos-special and have legacyTests not try it.

smarter:
Looks like separate compilation is not used with `compileDir` unfortunately.

PR #2820, opened 2017-06-29
Merge isInstanceOfEvaluator with TypeTestCasts

odersky:
As the saying goes: "let erasure handle it... "

The reason for the change is that things were a mess before

 - isInstanceOfEvaluator did not handle &/| types. It would
   be hard to change that since it was looking at erased types.
 - isInstanceEvaluator, but not TypeTestCasts added null-checks
 - therefore, necessary null-checks were not emitted on some tests of
   the form `x.isInstanceOf[A & B]`.
 - the pattern matcher patched this up by inserting some null checks
   on its own, whereas it should just have let the translation of
   isInstanceOf do the job.


odersky:
isInstanceOfEvaluator does not work if the tested type is of the form A | B or A & B.

DarkDimius:
Aaah, I missed that those points rely on the first one.

Thanks!

odersky:
I believe we have now concentrated all logic around isInstanceOf in TypeTestCasts. Before this was distributed in various checks and rewrites in RefChecks, isInstanceOfEvaluator and TypeTestCasts. There remain some null checks in the pattern matcher which should now be redundant. There's also the outer test generated by pattern matcher, but inserting that at Erasure is too late (we'd need to do an ensureExplicitOuter before). 

Local opts before erasure should be changed so that always failing instance tests are not replaced by `false` (these need to be reported in TypeTestsCasts). If someone knows of other places where isInstanceOf is handled it would be good to know and to check whether we can also move this to TypeTestsCasts.

odersky:
Looking at line counts, IsInstanceOfEvaluator is now no longer used, even though it is left in the repo for now, in case we want to transfer some of its bits. That's a savings of 173 lines, for a total reduction of ~130 lines.

smarter:
> even though it is left in the repo for now, in case we want to transfer some of its bits

I suggest getting rid of it, otherwise it's too easy to get confused and think it's used code.

Also a big :+1: from me for this PR, this is much more sane than the previous situation.

smarter:
See also https://github.com/scala/bug/issues/10379 which also affects dotty

odersky:
@smarter We actually have quite a few miniphases which are not linked but are still in the repo. Maybe it's time for a global sweep to remove all of them?

smarter:
Sounds good to me.

compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
```diff
@@ -26,100 +25,140 @@ import ValueClasses._
...
-                    erased2 match {
-                      case Literal(Constant(true)) => erased1
-                      case _ => erased1 and erased2
-                    }
-                }
+        def derivedTree(expr1: Tree, sym: Symbol, tp: Type) =
+          cpy.TypeApply(tree)(expr1.select(sym).withPos(expr.pos), List(TypeTree(tp)))
+
+        def foundCls = expr.tpe.widen.classSymbol
+        // println(i"ta $tree, found = $foundCls")
```
felixmulder: Forgotten removal - or leave in as a nice point for debug?

PR #2822, opened 2017-06-30
Change pattern matcher

odersky:
A rewrite of the pattern matcher. It 

 - simplifies and clarifies structure (470 LOC instead of 1038) 
 - keeps generated trees small and readable
 - improves debuggability since pattern matching decision trees can be pretty-printed

Based on #2820. The hardest part was actually to get #2820 right. Once that was done
it was possible to make steady progress on the pattern matcher itself.


odersky:
Superseded by #2829.

Issue #2818, opened 2017-06-28
Styling of Blogs landing page


Varunram:
<img width="1280" alt="screen shot 2017-06-28 at 11 18 49 am" src="https://user-images.githubusercontent.com/18598959/27626106-6d3ab7e8-5c04-11e7-87ce-bda94491532f.png">
Increasing space between the date and the blog would make it look a bit better. What are your views?

felixmulder:
I don't think the blog landing page is supposed to have a date up there by the title :)

Varunram:
Closed via `e52095aa6981b055eead1b96b841ac60d5892d01`

PR #2824, opened 2017-07-01
Remove date from landing page of Blogs

Varunram:
A dirty fix for #2818. There are two possibilities - Use js to hide the element in `index.html` but that'd go against "strict" templating and the other would be to modify `index.html` like this one. Would love to hear your views on this. CC @felixmulder

### DIFF
--- a/docs/blog/index.html
+++ b/docs/blog/index.html
@@ -1,20 +1,53 @@
 ---
-layout: blog-page
-title: "Blog"
+layout: main
+extraCSS:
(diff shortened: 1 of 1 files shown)


felixmulder:
Interesting, the CI tried to publish the docs because this was taken from the down stream master...I'm sure this is a bug in drone.

LGTM, thanks @Varunram !

Varunram:
Thanks! :)

PR #2825, opened 2017-07-02
Refs #1589 Move "wrong number of parameters" error to a new errors fo…

mikla:
New schema for "wrong number of parameters" error.


felixmulder:
LGTM, thanks @mikla!

PR #2827, opened 2017-07-03
Make dotty compile with 2.12.3

felixmulder:
This PR makes dotty *compile* with 2.12.3

TODO
=====
- [x] ~Fix warnings (discussed the Eta-expansion errors, we will not be fixing them)~
- [x] ~Implement `JFunctionX` methods that were left with a `UnsupportedOperationException`~
- [x] Remove dependency on JFunctionX/JProcedureX
- [x] Fix trait encoding for 2.12 compiled traits (@odersky, but @felixmulder will have a look as well)
- [x] ~Reimplement `FunctionalInterfaces`~ kept the interfaces instead
- [x] Fix scala2 library tests
- [ ] Bump version number?


project/Build.scala
```diff
@@ -461,7 +461,7 @@ object Build {
 
       // get libraries onboard
       libraryDependencies ++= Seq("com.typesafe.sbt" % "sbt-interface" % sbtVersion.value,
-                                  ("org.scala-lang.modules" %% "scala-xml" % "1.0.1").withDottyCompat(),
+                                  "org.scala-lang.modules" %% "scala-xml" % "1.0.6",
```
smarter: Until you change withDottyCompat, you can rewrite this to:
```scala
"org.scala-lang.modules" % "scala-xml_2.12" % "1.0.6",
```

felixmulder:
@sjrd - you removed the `BoxesRuntime` hash methods in 2.12 :)

sjrd:
Apparently I also removed that very test from Scala 2.12, so it's indeed irrelevant.

odersky:
It seems the error is that it does not find `##` in ScalaRunTime.

smarter:
There's multiple errors (see https://github.com/lampepfl/dotty/pull/2827/commits/01ff24f3da5b6bc2bb14939a5f050c39d5cdad72 for some tests that fail at runtime due to some issues with the mixin encoding), but that's one of the issue yes, see https://github.com/scala/scala/commit/cce701650143d13c8b292bffd590bf7c8eb532d7 for what changed in Scalac (this commit only deprecated ScalaRunTime.hash but it got removed in a subsequent commit).

odersky:
I fixed the mixin encoding problem that showed up in unittest_collection. I could not reproduce the LazyVals encoding. @felixmulder can you take over?

felixmulder:
Thanks @odersky, we'll take it from here, enjoy your vacay! 🍹 🏖

felixmulder:
Merging as @smarter approved in person :tada:

project/Build.scala
```diff
@@ -35,7 +35,7 @@ object ExposedValues extends AutoPlugin {
 
 object Build {
 
-  val scalacVersion = "2.11.11" // Do not rename, this is grepped in bin/common.
+  val scalacVersion = "2.12.2" // Do not rename, this is grepped in bin/common.
```
rom1dep: Maybe I'm missing something, but this series isn't updating bin/common for 2.12, is it intentional?

felixmulder: It wasn't intentional,no. But, we are aware: https://github.com/lampepfl/dotty/pull/3026/

If you think you can fix the `./bin/*` scripts, then we'd be more than happy to accept a PR! :)

rom1dep: Great, until now I hadn't even noticed that there is both ./bin and ./dist/bin ;)

PR #2829, opened 2017-07-03
Add pattern matcher optimizations

odersky:
We add more extensive optimizations to pattern matching.

In particular: Eliminate redundant tests, and merge variables that have the same right-hand side.


compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -70,7 +70,6 @@ class Compiler {
            new CrossCastAnd,        // Normalize selections involving intersection types.
            new Splitter),           // Expand selections involving union types into conditionals
       List(new VCInlineMethods,     // Inlines calls to value class methods
-           new IsInstanceOfEvaluator, // Issues warnings when unreachable statements are present in match/if expressions
```
nicolasstucki: Why is `IsInstanceOfEvaluator` removed? I see that there are some modifications to it.

smarter: @nicolasstucki See https://github.com/lampepfl/dotty/pull/2820

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -530,7 +529,7 @@ object PatternMatcher {
      */
     private def inlineLabelled(plan: Plan) = {
       val refCount = referenceCount(plan)
-      def toDrop(sym: Symbol) = labelled.contains(sym) && refCount(sym) <= 1
+      def toDrop(sym: Symbol) = labelled.contains(sym) && refCount(symb) <= 1
```
smarter: This commit introduces a typo

odersky:
This is ready for review now. With the latest optimizations that do variable propagation across label boundaries the generated code is pretty good now - in fact it looks often better than what one would generate by hand.

odersky:
For instance, here's a translation of the match expression

    xs match {
      case 0 :: Nil => println("1")
      case _ :: Nil => println("2")
      case 0 :: _   => println("3")
      case x :: y   => println("4")
    }

Translation:

      case val x17: scala.collection.immutable.List[Int](Test.xs) = Test.xs
      def case16(case x30: scala.collection.immutable.::[Int]): Unit = 
        {
          case val x25: Int = x30.head
          case val x26: scala.collection.immutable.List[Int] = x30.tl$1
          if Nil.==(x26) then 
            {
              println("2")
            }
           else 
            if 0.==(x25) then 
              {
                println("3")
              }
             else 
              {
                case val x: Int = x25
                case val y: scala.collection.immutable.List[Int] = x26
                println("4")
              }
        }
      if x17.$isInstanceOf$[scala.collection.immutable.::[Int]] then 
        {
          case val x27: scala.collection.immutable.::[Int] = 
            x17.asInstanceOf[scala.collection.immutable.::[Int]]
          if 0.==(x27.head) then 
            if Nil.==(x27.tl$1) then 
              {
                println("1")
              }
             else case16(x27)
           else case16(x27)
        }
       else throw new MatchError(x17)
    }

DarkDimius:
@odersky, looks really good. Here are several more steps that Simplifier in https://github.com/dotty-linker/dotty/tree/opto is able to additionally do:

1. 
```
      if 0.==(x27.head) then 
        if Nil.==(x27.tl$1) then 
          {
            println("1")
          }
         else case16(x27)
       else case16(x27)
``` 

It sees that two else branches are the same and reorders statements so that duplicated else branches are joined:

```
      if (0.==(x27.head)) && (Nil.==(x27.tl$1)) then 
          {
            println("1")
          }
         else case16(x27) //1
``` 

This triggers one more rounds of optimizations after inlining case16 that is now called once.

2. It additionally gets rid of: 
 - x27, as an alias to immutable x17
 - x30.head && x27.head are aliased, the second one is removed and the first one is reused. x30.head is always computed on paths that computed x27.head and both are immutable. This gets rid of x25.
 

There is also an optimization in this code that both simplifier and your optimization have missed. They both compare first element with `0` twice on the path. Simplifier wasn't able to remove it due to rewriting `//1` that simplified outer test but made analyzing inner tests more complex. My intention here was to use idempotent expression elimination to remove this duplicated computation but the implementation is still not good enough.

odersky:
@DarkDimius God to see that local opts improves further on this. My guess is it's because local opts do many rounds, whereas pm does just one pass.

DarkDimius:
@odersky, I agree with your guess. Is this branch ready for review?

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1,1871 +1,931 @@
...
-
-      def chainBefore(next: Tree)(codegenAlt: Casegen): Tree = {
-        /*atPos(pos)*/{
-          // one alternative may still generate multiple trees (e.g., an extractor call + equality test)
-          // (for now,) alternatives may not bind variables (except wildcards), so we don't care about the final substitution built internally by makeTreeMakers
-          val combinedAlts = altss map (altTreeMakers =>
-            ((casegen: Casegen) => combineExtractors(altTreeMakers :+ TrivialTreeMaker(casegen.one(Literal(Constant(true)))))(casegen))
-            )
-
-          val find
```
DarkDimius: it should also check if initializer refers to vars:

```
case class Box[T](var a: T)

x match {
  case x @ Box(t) =>  // this can be filtered by checking that t is not synthetic
   x.a = null;
   t
  case Box(Box(t)) =>  // this can't, as the breakage happens through an alias to the outer
   x.asInstanceOf[Box].a = null;
   t
}

Comment (DarkDimius) [inline compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala]: why not do it even if it wasn't annotated similar to how tailrec does it?

**Comment by DarkDimius** [comment]
In private discussion we agreed to change 
  https://github.com/lampepfl/dotty/pull/2829#discussion_r126353295

to use semantic names instead. I'd also propose to extract this method and make it public so that others can call it, instead of hardcoding the current test. I'll need to modify quite a lot of code to use this test instead of checking for case flag ;-).

**Comment by DarkDimius** [comment]
Otherwise LGTM

File: compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala [hunk 4]
```diff
@@ -1,1871 +1,931 @@
...
-
-  trait CodegenCore {
-
-    // assert(owner ne null); assert(owner ne NoSymbol)
-    def freshSym(pos: Position, tp: Type = NoType, unique: UniqueNameKind = PatMatStdBinderName, owner: Symbol = ctx.owner) = {
-      ctx.newSymbol(owner, unique.fresh(), Flags.Synthetic | Flags.Case, tp, coord = pos)
+object PatternMatcher {
+  import ast.tpd._
+
+  final val selfCheck = true // debug option, if on we check that no case gets generated twice
```
odersky: No, I think we can drop it. On would typically enable it if something funky happens in the backend with double definitions.

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
DarkDimius: Github bug? This comment used to refer to different line, that implements inlining of vals.

DarkDimius: Github bug? This comment used to refer to lines deciding to use switch that currently only trigger if user asked for it.

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -1,1871 +1,931 @@
...
+        case Typed(pat, tpt) =>
+          TestPlan(TypeTest(tpt), scrutinee, tree.pos,
+            letAbstract(ref(scrutinee).asInstance(tpt.tpe)) { casted =>
+              nonNull += casted
+              patternPlan(casted, pat, onSuccess, onFailure)
+            },
+            onFailure)
+        case UnApply(extractor, implicits, args) =>
+          val mt @ MethodType(_) = extractor.tpe.widen
+          var unapp = extractor.appliedTo(ref(scrutinee).ensureConforms(mt.paramInfos.head))
```
odersky: The TypeTest is already generated by Typer.

PR #2832, opened 2017-07-05
Allow multiple calls to `checkMessagesAfter` in ErrorMessagesTests

felixmulder:
This solves the issue @bbarker was having with regards to runs.

Second commit simply cleans up formatting of ErrorMessagesTests file.

### DIFF
--- a/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
+++ b/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
@@ -11,15 +11,15 @@ import org.junit.Assert._
 
 trait ErrorMessagesTest extends DottyTest {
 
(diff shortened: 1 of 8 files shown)


compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTest.scala
```diff
@@ -11,15 +11,15 @@ import org.junit.Assert._
 
 trait ErrorMessagesTest extends DottyTest {
 
-  ctx = freshReporter(ctx)
-
-  private def freshReporter(ctx: Context) =
-    ctx.fresh.setReporter(new CapturingReporter)
+  private def freshReporter(ctx: Context) = {
```
smarter: No need for the ctx argument anymore.

PR #2836, opened 2017-07-05
Fix position of implicit conversions for IDEs

smarter:
In f24190d5c60c4e019fba97425b6f3b10167a039d the position was changed to
be zero-extent so that hover-on-type and other IDE features would work
correctly for expressions that get implicitly converted.
In 8a60a204589f16eede777c257559806f9a477d30 this got reverted without
explanation, I assume that this was an error so I'm now restoring it.


odersky:
LGTM. I believe the revert was an accident.

PR #2837, opened 2017-07-06
Refine purity check

odersky:
Refine "a pure expression does nothing in statement position" warning.
It should not be issued if the expression has type Unit. Otherwise `()`
will trigger a warning.


DarkDimius:
LGTM

Issue #2828, opened 2017-07-03
Better error message when failing to launch Visual Studio Code


ssanj:
When using **launchIDE** to launch Visual Studio Code on Macos without either installing it or having it on the path I get the following error:

> [error] ({.}/*:runCode) java.io.IOException: Cannot run program "code": error=2, No such file or directory

It might be nice to have a more friendly error message like:

> Could not find Visual Studio Code on your system. Please [download](https://code.visualstudio.com/) it and add it to your path (for [linux](https://code.visualstudio.com/docs/setup/linux), [windows](https://code.visualstudio.com/docs/setup/windows) or [mac](https://code.visualstudio.com/docs/setup/mac)) and then rerun *launchIDE*.

smarter:
For anyone wanting to try fixing this, you can do it around here: https://github.com/lampepfl/dotty/blob/34ea263a65dd56db948e94ea19a4ecd9f29e742f/sbt-dotty/src/dotty/tools/sbtplugin/DottyIDEPlugin.scala#L236-L239

smarter:
The main annoying thing here is that `IOException` can be thrown in various situations, so you can't just catch `IOException` and always print this, you would have to check the string `IOException#message` I guess.

amanjpro:
I will be willing to work on this, maybe tomorrow I will have time to push something

amanjpro:
I believe I have fixed this, but not sure how to test it. sbt-dotty/scripted runs a bunch of long tests, is there an easier way to test it?

amanjpro:
I guess this issue should be closed now, as PR #2841 is now merged

smarter:
Yes, thanks

PR #2841, opened 2017-07-07
Print nicer messages for missing vs code

amanjpro:
This will fix issue #2828


DarkDimius:
@smarter could you please review it today so we can get it into next milestone?

smarter:
@DarkDimius Sure, but note that this is a change to sbt-dotty so it won't be part of the milestone anyway, we'll need to make a new release of sbt-dotty.

PR #2843, opened 2017-07-10
Fix #2833: Fixes to harmonize

### LINKED ISSUE #2833: Unexpected type mismatch: found Int | Null, required Long
```scala
object a {
  val x: scala.Any =
    if (true)
      0L
    else if (false)
      (0: Int)
    else
      null
}
```
reports
```scala
-- [E007] Type Mismatch Error: foo.scala:5:9 -----------------------------------
5 |    else if (false)
  |         ^
  |         found:    Int | Null
  |         required: Long
  |
6 |      (0: Int)
7 |    else
8 |      null
```
I expected it to typecheck since Int, Long and Null are a subtype of Any. scalac does not report an error.

Also, note that the reported position is a bit weird. It gets more confusing inside a pattern match

```scala

object a {
  val x: scala.Any = 1 match {
    case 1 =>
      if (true) (0: Int) else null
    case 2 =>
      if (false) 0L else null
  }
}


// reports

-- [E007] Type Mismatch Error: foo2.scala:3:11 ---------------------------------
3 |    case 1 =>
  |           ^
  |           found:    Int | Null
  |           required: Long
  |
4 |      if (true) (0: Int) else null
-- [E007] Type Mismatch Error: foo2.scala:5:11 ---------------------------------
5 |    case 2 =>
  |           ^
  |           found:    Long | Null
  |           required: Long
  |
6 |      if (false) 0L else null

three errors found
```

odersky:
We try to replace harmonize with something weaker that only applies to constants. 

As a first step: make pattern matching typecheck more principled

When checking a pattern `p` against a selector `s` we should check
that `p == s` is typeable. So far we did this in a roundabout way
by calling harmonizeType and canCheckEquals directly. But that
is correct only if there are no user-defined definitions of `==`.

The new algorithm for harmonize is described a new section of the language reference.

### DIFF
--- a/docs/docs/reference/dropped/weak-conformance.md
+++ b/docs/docs/reference/dropped/weak-conformance.md
@@ -3,7 +3,7 @@ layout: doc-page
 title: Dropped: Weak Conformance
 ---
 
-In some situations, Scala used a {\em weak conformance} relation when
+In some situations, Scala used a _weak conformance_ relation when
 testing type compatibility or computing the least upper bound of a set
 of types.  The principal motivation behind weak conformance was to
 make as expression like this have type `List[Double]`:
@@ -24,12 +24,12 @@ A less obvious example is the following one, which was also typed as a
     List(n, c, d) // used to be: List[Double], now: List[AnyVal]
 
 Here, it is less clear why the type should be widened to
-`List[Double]`, a `List[AnyVal` seems to be an equally valid -- and
+`List[Double]`, a `List[AnyVal]` seems to be an equally valid -- and
 more principled -- choice.
 
 To simplify the underlying type theory, Dotty drops the notion of weak
 conformance altogether. Instead, it provides more flexibility when
-assigning a type to a constant expression. The rule is:
+assigning a type to a constant expression. The new rule is:
 
  - If a list of expressions `Es` appears as one of
 
--- a/docs/docs/reference/dropped/weak-conformance.md
+++ b/docs/docs/reference/dropped/weak-conformance.md
@@ -1,0 +1,78 @@
+---
+layout: doc-page
+title: Dropped: Weak Conformance
+---
+
+In some situations, Scala used a {\em weak conformance} relation when
+testing type compatibility or computing the least upper bound of a set
+of types.  The principal motivation behind weak conformance was to
+make as expression like this have type `List[Double]`:
+
+    List(1.0, math.sqrt(3.0), 0, -3.3) // : List[Double]
+
+It's "obvious" that this should be a `List[Double]`. However, without
+some special provision, the least upper bound of the lists's element
+types `(Double, Double, Int, Double)` would be `AnyVal`, hence the list
+expression would be given type `List[AnyVal]`.
+
+A less obvious example is the following one, which was also typed as a
+`List[Double]`, using the weak conformance relation.
+
+    val n: Int = 3
+    val c: Char = 'X'
+    val n: Double = math.sqrt(3.0)
+    List(n, c, d) // used to be: List[Double], now: List[AnyVal]
+
+Here, it is less clear why the type should be widened to
+`List[Double]`, a `List[AnyVal` seems to be an equally valid -- and
+more principled -- choice.
+
+To simplify the underlying type theory, Dotty drops the notion of weak
+conformance altogether. Instead, it provides more flexibility when
+assigning a type to a constant expression. The rule is:
+
+ - If a list of expressions `Es` appears as one of
+
+     - the elements of a vararg parameter, or
+     - the alternatives of an if-then-else or match expression, or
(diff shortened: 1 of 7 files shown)


docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: doc-page
+title: Dropped: Weak Conformance
+---
+
+In some situations, Scala used a {\em weak conformance} relation when
```
smarter: I don't think you can use latex code in our documentation. You should be able to write `*weak conformance*` instead which should be rendered as *weak conformance*.

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+                |
+              Byte
+
+    If these widenings lead to all expressions `Es` having the same type,
+    we use the transformed list of expressions instead of `Es`, otherwise
+    we use `Es` unchanged.
+
+__Examples:__
+
+    inline val b: Byte = 3
```
smarter: Would be good if there was some explanation of how the rules differ between `val` and `inline val`.

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+    inline val s: Short = 33
+    def f(): Int = b + s
+    List(b, s, 'a')     : List[Int]
+    List(b, s, 'a', f()): List[Int]
```
olafurpg: Might be good to note here that the type is `List[Int]` because `Int` is the LUB, not because of the new rule.

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+A less obvious example is the following one, which was also typed as a
+`List[Double]`, using the weak conformance relation.
+
+    val n: Int = 3
+    val c: Char = 'X'
+    val n: Double = math.sqrt(3.0)
+    List(n, c, d) // used to be: List[Double], now: List[AnyVal]
+
+Here, it is less clear why the type should be widened to
+`List[Double]`, a `List[AnyVal` seems to be an equally valid -- and
```
olafurpg: Missing closing bracket: `List[AnyVal]`

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+more principled -- choice.
+
+To simplify the underlying type theory, Dotty drops the notion of weak
+conformance altogether. Instead, it provides more flexibility when
+assigning a type to a constant expression. The rule is:
```
olafurpg: Does this rule have a name? Is it a new Dotty feature?

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+
+ - If a list of expressions `Es` appears as one of
+
+     - the elements of a vararg parameter, or
```
olafurpg: What happens when it's not a vararg parameter list?

```scala
def add[T](a: T, b: T): List[T] = List(a, b)
add(1.0f, 1L) // Double?
```

docs/docs/reference/dropped/weak-conformance.md
```diff
@@ -0,0 +1,78 @@
...
+
+__Examples:__
+
+    inline val b: Byte = 3
+    inline val s: Short = 33
+    def f(): Int = b + s
+    List(b, s, 'a')     : List[Int]
+    List(b, s, 'a', f()): List[Int]
+    List(1.0f, 'a', 0)  : List[Float]
+    List(1.0f, 1L)      : List[Double]
```
olafurpg: Add a `1: Int` here to highlight the difference between `f()` and `1: Int`?

docs/docs/reference/dropped/weak-conformance.md
olafurpg: What happens if this is `final`  instead of `inline`?

docs/docs/reference/dropped/weak-conformance.md
odersky: No, AnyVal.

docs/docs/reference/dropped/weak-conformance.md
odersky: I changed it so say that it is new. It does not have a name (yet).

docs/docs/reference/dropped/weak-conformance.md
odersky: Oops. That comes from switching context from writing a paper with LaTex...

odersky:
If I get am LGTM today we can still roll this in the release...

PR #2845, opened 2017-07-10
Fixes to TastyUnpickler: correctly handle unpicking of trait setters.

DarkDimius:
It was handled by wrong case before.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TastyUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TastyUnpickler.scala
@@ -49,7 +49,7 @@ class TastyUnpickler(reader: TastyReader) {
       case UTF8 =>
         goto(end)
         termName(bytes, start.index, length)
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/tasty/TastyUnpickler.scala
```diff
@@ -49,7 +49,7 @@ class TastyUnpickler(reader: TastyReader) {
       case UTF8 =>
         goto(end)
         termName(bytes, start.index, length)
-      case QUALIFIED | FLATTENED | EXPANDED | EXPANDPREFIX =>
+      case QUALIFIED | FLATTENED | EXPANDED | EXPANDPREFIX | TRAITSETTER =>
```
odersky: I had not realized TRAITSETTER could reach the pickler. I thought they were added after pickling. What's the scenario where this came up?

odersky: I propose we make an assert instead that triggers when TRAITSETTER is found. It would be good to find out why it arises.

PR #2855, opened 2017-07-12
Fix #2850: remove useless java opts for Mac OS

### LINKED ISSUE #2850: packaged dotc script passes funny flags to jvm
-Xdock:name="" -Xdock:icon="/Users/dark/Downloads/t/dotty-0.2.0-RC1-bin-SNAPSHOT/icon-mac.png" -Dapple.laf.useScreenMenuBar=true

icon-mac.png doesn't exist. Having an empty name isn't a good idea, but why would we put dotty in OS X Dock in the first place?

liufengyun:
Fix #2850: remove useless java opts for Mac OS


DarkDimius:
LGTM.

PR #2860, opened 2017-07-12
Fix #2856 Drop special treatment of packages in findRef

### LINKED ISSUE #2856: scala.* symbol precedes conflicting root package name
This behavior may be intentional but I could not find a reference to it in the docs.

```scala
package io.grpc {
  trait Grpc
}
package bar {
  import io.grpc.Grpc
  object a extends Grpc
}
```

Dotty reports the following error
```scala
-- [E008] Member Not Found Error: grpc.scala:6:12 ------------------------------
6 |  import io.grpc.Grpc
  |         ^^^^^^^
  |         value `grpc` is not a member of io

one error found
```
while scalac reports no error.

A workaround for error is to prefix the import: `import _root_.io.grpc`.

Given the recent popularity of `*.io` domains, a few popular Scala libraries libraries now use `io` as their root package. For example, Circe (https://circe.github.io/circe/) and Quill (http://getquill.io/).

odersky:
This aligns behavior with scalac.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -243,8 +243,27 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         if (ctx.scope == null) previous
         else {
           var result: Type = NoType
+
           val curOwner = ctx.owner
-          if ((ctx.scope ne lastCtx.scope) || (curOwner ne lastCtx.owner)) {
+
+          // Can this scope contain new definitions? This is usually the first
+          // context where either the scope or the owner changes wrt the
+          // context immediately nested in it. But for package contexts, it's
+          // the opposite: the last context before the package changes. This distinction
+          // is made so that top-level imports following a package clause are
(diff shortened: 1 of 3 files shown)


odersky:
The PR is much nicer with only the fist commit. Unfortunately CI dies with a cyclic reference error in compileStdLib. If I compile locally even with the exact sequence of files as given it succeeds. I don't have the time to track this down so for the moment we have to live with an unnecessary complication. If someone feels courageous enough to follow up on this, please do!

smarter:
> Unfortunately CI dies with a cyclic reference error in compileStdLib. If I compile locally even with the exact sequence of files as given it succeeds.

Same here, and I tried rerunning the same commit on the CI and couldn't get it to fail either. This is pretty bad, but it also means that your commit might be completely unrelated to the problem.

odersky:
Let me go back a commit and see if the problem reappears.

smarter:
Hmm, it did reappear indeed, so I'm very confused by the fact that the same commit succeeded the CI on http://dotty-ci.epfl.ch/smarter/dotty/1.

odersky:
I undid the undo. Mystery remains.

PR #2861, opened 2017-07-13
Fix #2858: Handle intersection selection when symbol only exists on one side

### LINKED ISSUE #2858: Type parameters via intersection types crash the compiler
Apparently the syntax `class Cont[type A]` has been removed. So I wrote this:

```scala
class Cont[A0](val x: A0) { type A = A0 }
val c: Cont[_] & { type A = Int } = new Cont(1)
c.x : Int
```

It fails to compile with "`found: c.A0(c.x); required: Int`".
So I tried this:

```scala
class Cont[A0](x0: A0) { type A = A0; val x: A = x0 }
```

But the compiler crashes with:

```
assertion failure for (Cont[_] & Object{A = Int})(c) <:< ?{ x: ? }, frozen = false
[info]
[info] (Cont[_] & Object{A = Int})(c) is a class dotty.tools.dotc.core.Types$TermRefWithFixedSym
[info]
[info] ?{ x: ? } is a class dotty.tools.dotc.typer.ProtoTypes$CachedSelectionProto
[info]
[info] exception occurred while typechecking src/main/scala/Main.scala
[info]
[info] exception occurred while compiling src/main/scala/Main.scala
java.lang.AssertionError: NoDenotation.owner
	at dotty.tools.dotc.core.SymDenotations$NoDenotation.owner(SymDenotations.scala:1816)
```

smarter:


compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -386,19 +386,21 @@ object Denotations {
...
          *      package as access boundary.
-         *   4. sym1 a method but sym2 is not.
+         *   5. sym1 a method but sym2 is not.
          *  The aim of these criteria is to give some disambiguation on access which
          *   - does not depend on textual order or other arbitrary choices
          *   - minimizes raising of doubleDef errors
          */
         def preferSym(sym1: Symbol, sym2: Symbol) =
           sym1.eq(sym2) ||
+            (sym1.exists && !sym2.exists) ||
```
odersky: I think you can just say `!sym2.exists`. They are different, so `sym1.exists` is redundant.

odersky:
Also, I think at least one, ad probably both tests on line 431 and 432 are now redundant and should be dropped.

smarter:
> Also, I think at least one, ad probably both tests on line 431 and 432 are now redundant and should be dropped.

I was able to drop these tests after changing `preferSym` to always check `sym1.exists` before doing other tests, this should have the same behavior.

odersky:
LGTM

PR #2865, opened 2017-07-13
Make stubs generated by Java parser more robust.

odersky:
??? as an identifier won't cut it if Predef is not imported. We now generate a TypedSplice with a symbolic reference instead. 

Review by @julienrf?


julienrf:
Thanks @odersky :)

PR #2870, opened 2017-07-14
Fix #2869: Have constructors run after group of Memoize

### LINKED ISSUE #2869: Too many class parameters get stored in fields
In this simple class

    class MT(sf: MT => String) {
       val s = sf(this)
     }

the parameter `sf` is retained in the class, even though it is only used
in the initialization of `s`. This creates memory leaks. Observed while
memory-profiling the IDE.

odersky:
As explained in the code comment, the two get too entangled otherwise
which leads to parameters being stored as class fields when they should not.

This was the most obvious source of memory leaks in the IDE. Once this is fixed
we can check whether there are others.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Constructors.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Constructors.scala
@@ -30,7 +30,24 @@ class Constructors extends MiniPhaseTransform with IdentityDenotTransformer { th
   import tpd._
 
   override def phaseName: String = "constructors"
-  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Memoize], classOf[HoistSuperArgs])
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[HoistSuperArgs])
+  override def runsAfterGroupsOf: Set[Class[_ <: Phase]] = Set(classOf[Memoize])
+    // Memoized needs to be finished because we depend on the ownerchain after Memoize
+    // when checking whether an ident is an access in a constructor or outside it.
+    // This test is done in the right-hand side of a value definition. If Memoize
+    // was in the same group as Constructors, the test on the rhs ident would be
+    // performed before the rhs undergoes the owner change. This would lead
+    // to more symbls being retained as parameters. Test case is
+    //
+    //     dotc tests/pos/captuing.scala -Xprint:constr
+    //
+    // `sf` should not be retained as a filed of class `MT`.
+
+  /** The private vals that are known to be retained as class fields */
+  private val retainedPrivateVals = mutable.Set[Symbol]()
+
+  /** The private vals whose definition comes before the current focus */
+  private val seenPrivateVals = mutable.Set[Symbol]()
 
   // Collect all private parameter accessors and value definitions that need
   // to be retained. There are several reasons why a parameter accessor or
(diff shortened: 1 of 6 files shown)


tests/pos/capturing.scala
```diff
@@ -0,0 +1,4 @@
+class MT(sf: MT => String) {
+  // `s` is retained as a field, but `sf` should not be.
```
smarter: To actually test this, I suggest making a run test and using Java reflection like https://github.com/lampepfl/dotty/blob/master/tests/run/paramForwarding.scala#L46

odersky:
@smarter Ready for review once tests have passed.

tests/run/capturing.scala
```diff
@@ -0,0 +1,9 @@
+class MT(sf: MT => String) {
+  // `s` is retained as a field, but `sf` should not be.
+  val s = sf(this)
+}
+object Test extends App {
```
smarter: I think we've deprecated/discourage the use of `App`

odersky: Yes but I am not sure why? I hate to write all these main methods.

smarter: See https://github.com/lampepfl/dotty/issues/559

compiler/src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -30,7 +30,24 @@ class Constructors extends MiniPhaseTransform with IdentityDenotTransformer { th
...
   override def phaseName: String = "constructors"
-  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[Memoize], classOf[HoistSuperArgs])
+  override def runsAfter: Set[Class[_ <: Phase]] = Set(classOf[HoistSuperArgs])
+  override def runsAfterGroupsOf: Set[Class[_ <: Phase]] = Set(classOf[Memoize])
+    // Memoized needs to be finished because we depend on the ownerchain after Memoize
+    // when checking whether an ident is an access in a constructor or outside it.
+    // This test is done in the right-hand side of a value definition. If Memoize
+    // was in the same group as Constructors, the test on the rhs ident would be
+    // performed before the rhs undergoes the owner change. This would lead
+    // to more symbls being retained as parameters. Test case is
```
smarter: typo: symbls -> symbols

compiler/src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -30,7 +30,24 @@ class Constructors extends MiniPhaseTransform with IdentityDenotTransformer { th
...
+    //
+    //     dotc tests/pos/captuing.scala -Xprint:constr
```
smarter: typo: captuing -> capturing

compiler/src/dotty/tools/dotc/transform/Constructors.scala
```diff
@@ -79,8 +93,7 @@ class Constructors extends MiniPhaseTransform with IdentityDenotTransformer { th
   }
 
   override def transformValDef(tree: tpd.ValDef)(implicit ctx: Context, info: TransformerInfo): tpd.Tree = {
-    if (mightBeDropped(tree.symbol))
-      (if (isWildcardStarArg(tree.rhs)) retainedPrivateVals else seenPrivateVals) += tree.symbol
```
smarter: Why was this condition dropped?

tests/run/capturing.scala
odersky: I am still not sure why we use LegacyApp instead of App now. They seem to be equivalent.

smarter: /cc @DarkDimius do you remember why LegacyApp had to be introduced?

compiler/src/dotty/tools/dotc/transform/Constructors.scala
odersky: It was dead code. The test looks non-sensical to me. I fail to see how a definition of a class member can end up as something like this:

     val xs = y: _*

Even if it did, why would it make a difference? I put in an assert to flag the consition, but never triggered in all our tests.

smarter:
Do we really need to split the group of phases in two? Perhaps we could move Constructors to the next group of phase instead? /cc @DarkDimius

odersky:
> Do we really need to split the group of phases in two? Perhaps we could move Constructors to the next group of phase instead?

Unfortunately no. See last commit.

smarter:
OK. I assume that this will have a non-trivial impact on performance but it should be worth it anyway.

odersky:
I think the win of not having to store all these parameters might well be stronger than the loss of having an extra phase. If we only had benchmarks that worked...

PR #2877, opened 2017-07-16
Upgrade collection strawman

nicolasstucki:


julienrf:
What’s the goal of having this submodule?

nicolasstucki:
it is there to use it in the tests. Just to check that we are not breaking any of the language features used in the strawman. It a submodule to be able to make sure that it is always the same version of the code that gets tested for any given dotty commit.

julienrf:
OK. I think it’s a bit redundant because in scala/collection-strawman our CI also compiles the code with Dotty. But I think it makes more sense to check this in Dotty rather than in the collection-strawman repo.

nicolasstucki:
We want to know if a new commit in dotty would break the strawman (compilation) before merging it to master.
We additionally check the idempotency of the generated bytecode. I have another branch where I use it to build more complex  separate compilation tests using those sources. Also currently working on some dotty benchmarks that should use a stable version of the strawman.

PR #2882, opened 2017-07-17
Fix #2771: pre-check kinds at Typer

### LINKED ISSUE #2771: Reducing open types is potentially unsafe
The following currently causes a stack overflow in dotty (it's rejected by scalac):

```Scala
trait A { type L[X] }
trait B { type L }
trait C { type M <: A }
trait D { type M >: B }

object Test {
  def test(x: C with D): Unit = {
    def f(y: x.M)(z: y.L[y.L]) = z
    f(new B { type L[F[_]] = F[F] })(1)
  }
}
```

The overflow is caused by the compiler attempting to evaluate the non-terminating type expression `y.L[y.L]` which is ill-kinded.

The type definition `type L[F[_]] = F[F] }` is clearly ill-kinded already and scalac actually reports this.

But the deeper issue is in the line above that definition: the type `y.L` has two conflicting kinds, it is both a type operator and a proper type. This is because `L` is declared as a type operator in trait `A` and as a proper type in trait `B`, which are the upper and lower bounds of `x.M`, respectively. So we can construct an instance `y` of `B` where `y.L` is a proper type and then apply it to some other type by up-casting `y` to `x.M`. Here's a variant of the example which illustrates that point more clearly:
```Scala
//...
object Test {
  def test(x: C with D): Unit = {
    def f(y: x.M)(z: y.L[Any]) = z  // <-- y.L used as a type operator
    f(new B { type L = Any })(1)    // <-- y.L defined as a proper type
  }
}
```
It is rejected with the rather absurd error message

    -- [E007] Type Mismatch Error:
     9 |		f(new B { type L = Any })(1)
       |		                          ^
       |		                          found:    Int(1)
       |		                          required: Any[Any]

Changing the type of `z` to `y.L` results instead in

    -- [E055] Syntax Error:
     8 |    def f(y: x.M)(z: y.L) = z
       |                     ^^^
       |                     missing type parameter for y.L

This is another instance of a know problem about type members with absurd bounds (see also issue #50): evaluation is unsafe in a context containing variables/values (here `x`) of a type (here `C & D`) containing type members with absurd bounds (here `x.M`). In this particular case, the unsafe evaluation happens in an open *type* (namely `y.L[Any]`) rather than an open *term*, and at compile time (rather than at runtime).

**Clarification:** by "open type" I mean a type that contains free type or term variables (as opposed to a closed type which contains no free variables).

odersky:
We normally do kind checking during PostTyper, together with bounds checking. 
However, this is not early enough, as #2771 shows. The problem is that we can 
use ill-kinded code to construct self applications of types, which leads to general
unhappiness in the typer. Prevent that by checking the outline of kinds of types
as soon as we create them.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -251,18 +251,21 @@ class TypeApplications(val self: Type) extends AnyVal {
   }
 
   /** Is self type higher-kinded (i.e. of kind != "*")? */
-  def isHK(implicit ctx: Context): Boolean = self.dealias match {
-    case self: TypeRef => self.info.isHK
-    case self: RefinedType => false
-    case self: HKTypeLambda => true
-    case self: SingletonType => false
+  def isHK(implicit ctx: Context): Boolean = hkResult.exists
+
+  /** If self type is higher-kinded, its result type, otherwise NoType */
+  def hkResult(implicit ctx: Context): Type = self.dealias match {
+    case self: TypeRef => self.info.hkResult
+    case self: RefinedType => NoType
+    case self: HKTypeLambda => self.resultType
+    case self: SingletonType => NoType
     case self: TypeVar =>
       // Using `origin` instead of `underlying`, as is done for typeParams,
       // avoids having to set ephemeral in some cases.
-      self.origin.isHK
-    case self: WildcardType => self.optBounds.isHK
-    case self: TypeProxy => self.superType.isHK
-    case _ => false
+      self.origin.hkResult
+    case self: WildcardType => self.optBounds.hkResult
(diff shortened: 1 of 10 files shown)


smarter:
While this PR helps avoid some absurd applications, I don't think it solves the fundamental issue where a type can be seen as having multiple kinds because of bad bounds, here's a slightly modified test case that causes a stackoverflow while beta-reducing:
```scala
trait A { type L[X[_]] }
trait B { type L }
trait C { type M <: A }
trait D { type M >: B }

object Test {
  def test(x: C with D): Unit = {
    def f(y: x.M)(z: y.L[y.L]) = z
    f(new B { type L[F[_[_]]] = F[F] })(1)
  }
}
```

sstucki:
Found a way to break the fixed version, though the error is different. See #2887.

PR #2884, opened 2017-07-18
Remove Bench or A generic JHM benchmark interface

liufengyun:
The old `bench/` project is no longer necessary, as it's superseded by:

- https://github.com/liufengyun/bench
- https://github.com/liufengyun/bench-web

Web UI: https://liufengyun.github.io/bench/

We can either remove the project `bench` completely, or change it in to a generic test interface, as this PR does. This PR will enable developers to run jmh benchmarks in command line:

```bash
sbt dotty-bench/jmh:compile
sbt dotty-bench/pack
bench/target/pack/bin/bench  hello.scala    # arguments are the same as compiler options
```

Note: The separation of benchmarking into a separate repo is justified by the fact that (1) we need to test history of Dotty, thus the test driver has to be independent of versions of Dotty; (2) Dotty doesn't depend on it; (3) faster development process.


DarkDimius:
Suggestions:
 - measure in milliseconds;
 - graph multiple lines:
   - average
   - median [50% percentile]
   - 5% percentile
   - 95% percentile
   - 99% percentile
   - trendline for all listed above
 - store raw data for runs, so that we don't need to rerun benchmarks.

 - add a test that is supposed to always be flat, will be used as indicator of stability of machine.

DarkDimius:
and please, make the Y axis start from 0.

DarkDimius:
One more suggestion: could you align the x axis across multiple benchmarks? so that one vertical always relates to the same PR on all benchmarks.

DarkDimius:
LGTM

PR #2885, opened 2017-07-18
More IDE work

odersky:
- Make NamedTrees survive TypeErrors
    (Same principle as for completions applies).
 - Null out closures in LazyRef and LazyBodyAnnotation once they are evaluated.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Annotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Annotations.scala
@@ -61,7 +61,7 @@ object Annotations {
     def tree(implicit ctx: Context) = body
   }
 
-  case class LazyBodyAnnotation(bodyExpr: Context => Tree) extends BodyAnnotation {
+  case class LazyBodyAnnotation(private var bodyExpr: Context => Tree) extends BodyAnnotation {
     private var evaluated = false
     private var myBody: Tree = _
     def tree(implicit ctx: Context) = {
@@ -69,6 +69,7 @@ object Annotations {
       else {
         evaluated = true
         myBody = bodyExpr(ctx)
+        bodyExpr = null
       }
       myBody
     }
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2156,16 +2156,14 @@ object Types {
     }
   }
 
-  case class LazyRef(refFn: () => Type) extends UncachedProxyType with ValueType {
-    private var myRefFn: () => Type = refFn
+  case class LazyRef(private var refFn: () => Type) extends UncachedProxyType with ValueType {
     private var myRef: Type = null
     private var computed = false
     def ref = {
       if (computed) assert(myRef != null)
       else {
         computed = true
-        myRef = myRefFn()
-        myRefFn = null // null out to avoid memory leak
+        myRef = refFn()
       }
       myRef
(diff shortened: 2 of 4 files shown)


compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -62,6 +62,9 @@ object Interactive {
       sourceSymbol(sym.owner)
     else sym
 
+  private def safely[T](op: => List[T]): List[T] =
```
smarter: Mark it `@inline` ?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -131,7 +131,7 @@ object Interactive {
    *  @param includeReferences  If true, include references and not just definitions
    */
   def namedTrees(trees: List[SourceTree], includeReferences: Boolean, treePredicate: NameTree => Boolean)
-    (implicit ctx: Context): List[SourceTree] = {
+    (implicit ctx: Context): List[SourceTree] = safely {
```
smarter: It's a bit of a shame that this means we'll discard all results when the problem might be localized. Alternatively, we could move the try/catch around `traverseChildren` I think.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2146,13 +2146,15 @@ object Types {
   }
 
   case class LazyRef(refFn: () => Type) extends UncachedProxyType with ValueType {
+    private var myRefFn: () => Type = refFn
```
smarter: I think this would be slightly nicer as `case class LazyRef(private var refFn: () => Type)`, this way you cannot accidentally capture `refFn` outside of the constructor.

smarter: Also, another way to avoid the leak would be to write `refFn: Context => Type` at the cost that `equals` could not be made to force the reference anymore. I tried it before and it seemed to work, but I haven't benchmarked it.

odersky: I am not sure it would work in general, though. Seems generally dicy to evaluate in a different context. We'd have to check every site where we create a LazyRef whether it is permissible.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: I think it's pretty rare that we would get a TypeError, so not a big deal either way.

compiler/src/dotty/tools/dotc/core/Annotations.scala
```diff
@@ -61,14 +61,15 @@ object Annotations {
     def tree(implicit ctx: Context) = body
   }
 
-  case class LazyBodyAnnotation(bodyExpr: Context => Tree) extends BodyAnnotation {
+  case class LazyBodyAnnotation(private var bodyExpr: Context => Tree) extends BodyAnnotation {
```
smarter: It's suspicious that a `Context => Tree` function would capture a Context.

smarter: Oh actually I guess this is intentional: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Inliner.scala#L193

PR #2886, opened 2017-07-18
Fix #2770: Tighten type compatibility checks for overrides

### LINKED ISSUE #2770: Missing kind checks when overriding a type
```scala
trait A { type L }
trait B extends A { type L[X] }
```
In scalac this fails with:
```scala
 error: The kind of type L does not conform to the expected kind of type L in trait A.
B.this.L's type parameters do not match type L's expected parameters:
type L (in trait B) has one type parameter, but type L (in trait A) has none
trait B extends A { type L[X] }
                         ^
```
but it succeeds in dotty, which probably leads to interesting issues.

odersky:
- We now require strict subtypes if owning classes are in
   a subclass relationship. Nonempty bounds good enough only
   if classes are unrelated.
 - We treat class types correctly by eta-expanding their types.
 - We do kind checking on types.

Previously, some illegal things slipped through because of the way Any works.
In fact a hk type like `[X] => Any` is a subtype of `Any` according to
TypeComparer. But kind checking should prevent us from making that
comparison in the first place.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -183,25 +183,27 @@ object RefChecks {
     def infoString0(sym: Symbol, showLocation: Boolean) = {
       val sym1 = sym.underlyingSymbol
       def info = self.memberInfo(sym1)
-      i"${if (showLocation) sym1.showLocated else sym1}${
+      val infoStr =
         if (sym1.isAliasType) i", which equals ${info.bounds.hi}"
         else if (sym1.isAbstractType) i" with bounds$info"
         else if (sym1.is(Module)) ""
         else if (sym1.isTerm) i" of type $info"
         else ""
-      }"
+      i"${if (showLocation) sym1.showLocated else sym1}$infoStr"
     }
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -183,25 +183,27 @@ object RefChecks {
...
     /* Check that all conditions for overriding `other` by `member`
        * of class `clazz` are met.
        */
     def checkOverride(member: Symbol, other: Symbol): Unit = {
-      def memberTp = self.memberInfo(member)
-      def otherTp = self.memberInfo(other)
+      def memberTp(self: Type) =
+        if (member.isClass) TypeAlias(member.typeRef.EtaExpand(member.typeParams))
+        else self.memberInfo(member)
+      def otherTp(self: Type) = self.memberInfo(other)
```
smarter: Why isn't `other` eta-expanded like `member`?

odersky: Other cannot be a class anyway. Classes cannot be overridden.

PR #2889, opened 2017-07-19
Fix #2888: Avoid caching supertypes of hk apps with provisional infos

### LINKED ISSUE #2888: Ignored upper bound of type parameter
In scala/collection-strawman, [this line](https://github.com/scala/collection-strawman/blob/map-ops/src/main/scala/strawman/collection/immutable/Map.scala#L37) gives the following compilation error:

~~~
def removeAll(keys: IterableOnce[K]): C = fromSpecificIterable(keys.iterator().foldLeft[CC[K, V]](coll)(_ - _))
                                                                                                        ^^^
                                 value `-` is not a member of CC[K, V]
~~~

Though `CC` is upper bounded by `MapOps` (see [here](https://github.com/scala/collection-strawman/blob/map-ops/src/main/scala/strawman/collection/immutable/Map.scala#L14)), which provides the `-` operation. Note that the `CC` upper bound is recursive.

This code compiles fine with Scala 2.12 and 2.13.

odersky:
If the type constructor of a hk application has a provisional info (because
we are starting to initialize an F-bounded definition), we cannot cache its
supertype. We get an empty bound if we do.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3012,12 +3012,16 @@ object Types {
 
     override def superType(implicit ctx: Context): Type = {
       if (ctx.period != validSuper) {
+        var canCache = true
         cachedSuper = tycon match {
           case tp: HKTypeLambda => defn.AnyType
           case tp: TypeVar if !tp.inst.exists =>
             // supertype not stable, since underlying might change
-            return tp.underlying.applyIfParameterized(args)
-          case tp: TypeProxy => tp.superType.applyIfParameterized(args)
+            canCache = false
+            tp.underlying.applyIfParameterized(args)
+          case tp: TypeProxy =>
+            canCache = !tp.typeSymbol.is(Provisional)
+            tp.superType.applyIfParameterized(args)
           case _ => defn.AnyType
         }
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3012,12 +3012,16 @@ object Types {
 
     override def superType(implicit ctx: Context): Type = {
       if (ctx.period != validSuper) {
+        var canCache = true
```
smarter: This variable appears to be set but never read.

smarter:
Also this is missing the testcase from #2888

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3012,15 +3012,19 @@ object Types {
...
         cachedSuper = tycon match {
```
smarter: Modifying `cachedSuper` when we cannot cache the value seems dangerous. If `validSuper` refers to a non-empty period, then a subsequent call to `superType` might be done with `ctx.period` equal to `validSuper` and would return the new, incorrect value in `cachedSuper`.

odersky: Fair enough

PR #2890, opened 2017-07-19
Fix #2140: Avoid exponential subtype complexity

### LINKED ISSUE #2140: Exponential compilation time in pattern matching
This compiles in 9 seconds, with `case 12` uncommented it goes to 21 seconds and so on. Adding `: Any` ascriptions on every tuple brings things back to normal.

```scala
object Test {
    val s = 1
    val t: Any = 23 match {
      case 0  => (s)
      case 1  => (s, s)
      case 2  => (s, s, s)
      case 3  => (s, s, s, s)
      case 4  => (s, s, s, s, s)
      case 5  => (s, s, s, s, s, s)
      case 6  => (s, s, s, s, s, s, s)
      case 7  => (s, s, s, s, s, s, s, s)
      case 8  => (s, s, s, s, s, s, s, s, s)
      case 9  => (s, s, s, s, s, s, s, s, s, s)
      case 10 => (s, s, s, s, s, s, s, s, s, s, s)
      case 11 => (s, s, s, s, s, s, s, s, s, s, s, s) // 9s
      // case 12 => (s, s, s, s, s, s, s, s, s, s, s, s, s) // 21s
      // case 13 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s) // 67s
      // case 14 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 15 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 16 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 17 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 18 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 19 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 20 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 21 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
      // case 22 => (s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s, s)
    }
  }
}
```

It looks like most of the time is spent in `isSubType`, probably when computing the least upper bound?

odersky:
Avoid exponential subtype complexity arising from revisiting |-types
unnecessarily.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -993,7 +993,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     case tp: ProtoType => false
     case tp: RefinedOrRecType => isCovered(tp.parent)
     case tp: AnnotatedType => isCovered(tp.underlying)
-    case AndType(tp1, tp2) => isCovered(tp1) && isCovered(tp2)
+    case tp: AndOrType => isCovered(tp.tp1) && isCovered(tp.tp2)
     case _ => false
   }
 
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -983,14 +983,10 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
 
   /** A type has been covered previously in subtype checking if it
    *  is some combination of TypeRefs that point to classes, where the
-   *  combiners are RefinedTypes, RecTypes, AndTypes or AnnotatedTypes.
-   *  One exception: Refinements referring to basetype args are never considered
-   *  to be already covered. This is necessary because such refined types might
-   *  still need to be compared with a compareAliasRefined.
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -993,7 +993,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     case tp: ProtoType => false
     case tp: RefinedOrRecType => isCovered(tp.parent)
     case tp: AnnotatedType => isCovered(tp.underlying)
-    case AndType(tp1, tp2) => isCovered(tp1) && isCovered(tp2)
+    case tp: AndOrType => isCovered(tp.tp1) && isCovered(tp.tp2)
```
smarter: The documentation of the method needs to be updated to refer to AndOrTypes instead of AndTypes

PR #2894, opened 2017-07-20
Fix #2367: Allow SAM conversion for overloaded functions

### LINKED ISSUE #2367: lambda not SAM-converted in some cases related to overloading
This compiles with scala 2.12 but not dotty:
```scala
new Thread(() => println("hi"))
```
```scala
5 |new Thread(() => println("hi"))
  |    ^^^^^^
  |none of the overloaded alternatives of constructor Thread in class Thread with types
  | (x$0: ThreadGroup, x$1: Runnable, x$2: String, x$3: Long)Thread
  | (x$0: ThreadGroup, x$1: Runnable, x$2: String)Thread
  | (x$0: Runnable, x$1: String)Thread
  | (x$0: ThreadGroup, x$1: String)Thread
  | (x$0: String)Thread
  | (x$0: ThreadGroup, x$1: Runnable)Thread
  | (x$0: Runnable)Thread
  | ()Thread
  |match arguments (() => Unit)
```

The following works as expected:
```scala
new Thread((() => println("hi")): Runnable)
```

odersky:
If just one of a number of alternatives takes a SAM, allow a lambda argument,
just like we do for non-overloaded methods with SAM type parameters.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -474,21 +474,24 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
     type TypedArg = Arg
     type Result = Unit
 
+    protected def argOK(arg: TypedArg, formal: Type) = argType(arg, formal) match {
+      case ref: TermRef if ref.denot.isOverloaded =>
+        // in this case we could not resolve overloading because no alternative
+        // matches expected type
+        false
+      case argtpe =>
+        def SAMargOK = formal match {
+          case SAMType(meth) => argtpe <:< meth.info.toFunctionType()
+          case _ => false
+        }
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -474,21 +474,24 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+      case ref: TermRef if ref.denot.isOverloaded =>
+        // in this case we could not resolve overloading because no alternative
+        // matches expected type
+        false
+      case argtpe =>
+        def SAMargOK = formal match {
+          case SAMType(meth) => argtpe <:< meth.info.toFunctionType()
+          case _ => false
+        }
+        isCompatible(argtpe, formal) || ctx.mode.is(Mode.ImplicitsEnabled) && SAMargOK
```
smarter: Maybe add some note that implicit search and SAM conversion are tried at the same time to justify checking ImplicitsEnabled here?

PR #2896, opened 2017-07-20
Fix #2300: Try to heal subtype failures by instantiating type variables

### LINKED ISSUE #2300: Type level computation has different result than in Scalac
```scala
object hlists {
  sealed abstract class HList {
    type Merge[L <: HList] <: HList
    
    def merge[L <: HList](l: L): Merge[L]
  }
  final case class HCons[H, T <: HList](head : H, tail : T) extends HList {
    type Merge[L <: HList] = HCons[H, tail.Merge[L]]
    
    def merge[L <: HList](l: L): Merge[L] = HCons(head, tail.merge(l))
  }
  sealed trait HNil extends HList {
    type Merge[L <: HList] = L
    
    def merge[L <: HList](l: L): Merge[L] = l
  }
  final val HNil: HNil = { case object HNil extends HNil; HNil }
}

import hlists._

val merged: HCons[Int,HCons[String,HNil]] = HCons(42, HNil) merge HCons("foo", HNil)
```

Pasted in the Scala 2.12.2 REPL this gives:
```scala
merged: hlists.HCons[Int,hlists.HCons[String,hlists.HNil]] = HCons(42,HCons(foo,HNil))
```

Pasted in the Dotty REPL:
```scala
-- [E007] Type Mismatch Error: <console>:8:60 ----------------------------------
8 |val merged: HCons[Int,HCons[String,HNil]] = HCons(42, HNil) merge HCons("foo", HNil)
  |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |found:    hlists.HCons[H, T]#Merge[L]
  |required: hlists.HCons[Int, hlists.HCons[String, hlists.HNil]]
  |
  |where:    H is a type variable with constraint >: Int(42)
  |          L is a type variable with constraint >: hlists.HCons[H, T] and <: hlists.HList
  |          T is a type variable with constraint >: hlists.HNil(hlists.HNil) and <: hlists.HList
```

Without the explicit type annotation in Scala:
```scala
scala> HCons(42, HNil) merge HCons("foo", HNil)
res0: hlists.HCons[Int,hlists.HCons[String,hlists.HNil]] = HCons(42,HCons(foo,HNil))
```

Without the explicit type annotation in Dotty:
```scala
scala> HCons(42, HNil) merge HCons("foo", HNil) 
val res0: hlists.HCons[Int, hlists.HNil]#Merge[hlists.HCons[String, hlists.HNil]] = HCons(42,HCons(foo,HNil))
```

The strange part (to me at least) is that Dotty does seem to know that those types are equal:
```scala
scala> implicitly[HCons[Int,HNil]#Merge[HCons[String,HNil]] =:= HCons[Int,HCons[String,HNil]]] 
val res2: =:=[hlists.HCons[Int, hlists.HNil]#Merge[hlists.HCons[String, hlists.HNil]], 
  hlists.HCons[Int, hlists.HCons[String, hlists.HNil]]
] = <function1>
```

odersky:
The idea is that if an adapt fails because an actual type A is not a
subtype of an expected type B, we try to instantiate all type variables
in A and, if that changes something, try again.

This mitigates the discrepancy that dotc is more lazy in instantiating type
variables than scalac.

Fixes #2300.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2191,8 +2191,11 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
         val prevConstraint = localCtx.typerState.constraint
         if (isFullyDefined(wtp, force = ForceDegree.all)(localCtx) &&
             localCtx.typerState.constraint.ne(prevConstraint)) {
-          localCtx.typerState.commit()
-          return adapt(tree, pt, original)
+          val tree1 = adapt(tree, pt, original)(localCtx)
+          if (!localCtx.reporter.hasErrors) {
+            localCtx.typerState.commit()
+            return tree1
+          }
         }
       }
       // try an implicit conversion
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2186,25 +2186,23 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
           }
         case _ =>
       }
-      // try to fully instantiate type
-      { val localCtx = ctx.fresh.setNewTyperState
-        val prevConstraint = localCtx.typerState.constraint
-        if (isFullyDefined(wtp, force = ForceDegree.all)(localCtx) &&
-            localCtx.typerState.constraint.ne(prevConstraint)) {
-          val tree1 = adapt(tree, pt, original)(localCtx)
-          if (!localCtx.reporter.hasErrors) {
-            localCtx.typerState.commit()
-            return tree1
-          }
-        }
-      }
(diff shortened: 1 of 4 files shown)


liufengyun:
The latest speedup most likely comes from this PR -- due to some reordering of the commits, the graph doesn't point to the correct PR.

smarter:
>  due to some reordering of the commits, the graph doesn't point to the correct PR.

Can you explain that? What reordering of commits? It'd be good if the graph was accurate.

liufengyun:
Now it's clear -- the drop is due to my restarting of the testing server this morning. I've fixed the CPU to a determined frequency, the restart lost the setting.

Now I restored the setting (permanent), re-ran all tests for this morning and the graphs are explainable. The graph does point to the correct PRs.

PR #2897, opened 2017-07-21
#1589 Move duplicate qualified modifier error to new format

lorandszakacs:
This was a rather niched error message, since most of the time it is superseded by other errors.

This corresponds to the `Parsers:1703` index from the issue description.

P.S. I now realized that I will be leaving to areas with no internet soon... so hopefully all is well with the PR, otherwise I'll have to address it in a few days 😟

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -93,7 +93,8 @@ public enum ErrorMessageID {
     ModifiersNotAllowedID,
     WildcardOnTypeArgumentNotAllowedOnNewID,
     ImplicitFunctionTypeNeedsNonEmptyParameterListID,
-    WrongNumberOfParametersID
+    WrongNumberOfParametersID,
+    DuplicatePrivateProtectedQualifierID,
     ;
 
     public int errorNumber() {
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1694,4 +1694,12 @@ object messages {
     val explanation = ""
   }
 
+  case class DuplicatePrivateProtectedQualifier()(implicit ctx: Context)
+    extends Message(DuplicatePrivateProtectedQualifierID) {
+    val kind = "Syntax"
+    val msg = s"duplicate private/protected qualifier"
+    val explanation =
+      hl"""It is not allowed to combine `private` and `protected` modifiers even if they are qualified to different scopes""".stripMargin
```
nicolasstucki: You do not need `.stripMargin` here.

nicolasstucki: You also do not need `"""` for the string, `"` is better.

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -93,7 +93,8 @@
     ModifiersNotAllowedID,
     WildcardOnTypeArgumentNotAllowedOnNewID,
     ImplicitFunctionTypeNeedsNonEmptyParameterListID,
-    WrongNumberOfParametersID
+    WrongNumberOfParametersID,
+    DuplicatePrivateProtectedQualifierID
```
nicolasstucki: Add a `,` after `DuplicatePrivateProtectedQualifierID`

lorandszakacs:
@nicolasstucki, I addressed the issues. Thanks!

nicolasstucki:
@lorandszakacs thanks a lot!

PR #2898, opened 2017-07-21
Experiment: Replace type application encoding with native AppliedType

odersky:
This is an experiment to replace the encoding of type applications in terms of type members by native AppliedTypes. 

Arguments for keeping the existing encoding:

 - It's close to DOT, giving a semantics of type applications in terms of proven foundations
 - It gives a natural and simple meaning to wildcard arguments, e.g. `List[_]`.

Arguments for changing to native applications:

 - It's probably easier to understand in the compiler, since it is closer to source syntax.
 - It's easier to give good error messages.
 - It might avoid some cyclic reference errors.
 - It might be faster, since types are smaller and there's less back-and-forth encoding.

I believe keeping the DOT scheme would be overall simpler if it was not for higher-kinded types. Higher-kinded types force us to introduce Application nodes anyway, so maybe the overall implementation effort is simpler if we just generalize this to all types. We have to figure out how to best represent wildcard arguments, though. The purpose of this experiment is to do this, and then to evaluate implementation complexity and compiler performance.

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -889,7 +889,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
           // to inner symbols of DefDef
           // todo: somehow handle.
 
-    def parents: List[Type] = tp.parents
+    def parents: List[Type] = tp.parentsNEW
   }
 
 
--- a/compiler/src/dotty/tools/dotc/ast/tpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/tpd.scala
@@ -213,7 +213,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     ta.assignType(untpd.TypeDef(sym.name, TypeTree(sym.info)), sym)
 
   def ClassDef(cls: ClassSymbol, constr: DefDef, body: List[Tree], superArgs: List[Tree] = Nil)(implicit ctx: Context): TypeDef = {
-    val firstParentRef :: otherParentRefs = cls.info.parents
+    val firstParentRef :: otherParentRefs = cls.info.parentRefs
     val firstParent = cls.typeRef.baseTypeWithArgs(firstParentRef.symbol)
     val superRef =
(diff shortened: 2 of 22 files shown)


liufengyun:
If I understand well, the difficulty with wildcard is the following, which type checks with current `dotc` but not in `scalac`:

```Scala
class Option[T](var t: T) {
  def foo = {
    val opt: Option[_] = new Option[String]("")
    val x = opt.t
    opt.t = x
  }
}
```

liufengyun:
I was thinking about how native application can achieve the same for wildcards based on existential types. Here are some tentative thoughts:

__Auto Packing__. If the expected type has underscores, instantiate the underscores with _fresh type variables_ with the bounds as check condition, type check the given term with the new expected type. Type checking succeeds only if the type variables are instantiated and the instantiation satisfies the check condition.

__Auto Unpacking__. When type checking a variable definition of a type with underscores, always define _fresh type parameters_ in the same scope with the same bounds as the underscores, then type all references to the variable as the type instantiated by the type parameters.

__Auto Instantiation__. If the synthesized type of a term has underscores, instantiate the underscores to its bound (Note: restrict bounds of underscore to be either lower bound or upper bound, as in Java).

By _type parameters_, I mean the abstract type `T` like the follows:

```Scala
 def foo: Option[_ <: Int] = {
    type T <: Int                                       // <= inserted type parameter for `x`
    val x: Option[_ <: Int] = Some(5)
    x                                                   // usage of `x` gets the type: Option[T]
 }
```

The rule _Auto Instantiation_ is obviously problematic for Scala, I've no idea how to do better there -- the ideal solution is to introduce type parameters, but we don't know what's the correct scope of those type parameters. One possible way is to do as it's mentioned in `Wild FJ`[1, Section 3.1] to make the parameters global.

There's also subtle interaction with mutable vars as in the encoding (#2928).


[1] [Wild FJ](http://www.forskningsdatabasen.dk/en/catalog/2185796650), _Torgersen et. al._, 2005

odersky:
@liufengyun Yes, this looks much like what we use skolemization for, and how scalac treats existential types. But details get very tricky. Happy to discuss over lunch/coffee.

odersky:
Closed for the reasons given in #2947

PR #2899, opened 2017-07-21
Remove calls to deprecated constructor of asm.Handle

nicolasstucki:


nicolasstucki:
Hm, it looks like we use both asm 4 and 5

smarter:
The corresponding change in scalac was a bit more complicated for some reason: https://github.com/scala/scala/commit/e619b033350a3378d650db4c3e5b1bfc83b73d81

> Hm, it looks like we use both asm 4 and 5

I don't think that's the case, that would be very weird

PR #2900, opened 2017-07-21
remove deAnonymize

liufengyun:
As @odersky pointed out, this method is in fact identity, should be removed.


DarkDimius:
LGTM, after tests pass

PR #2908, opened 2017-07-24
Generalize Approximating Type Maps

odersky:
Approximating type maps now work with bounds instead of NoTypes. This
gives more precision (in particular for type applications), and also
makes it possible to combine approximation with other mappings.

The biggest win here is that we can now express AsSeenFrom soundly in terms of ApproximatingTypeMaps. We can therefore remove the UnsafeNonVariant annotations and 
retroactive skolemizations which felt kludgey.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeOps.scala
@@ -132,18 +132,9 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
 
   /** Approximate a type `tp` with a type that does not contain skolem types. */
   object deskolemize extends ApproximatingTypeMap {
-    private var seen: Set[SkolemType] = Set()
     def apply(tp: Type) = tp match {
-      case tp: SkolemType =>
-        if (seen contains tp) NoType
-        else {
-          val saved = seen
-          seen += tp
-          try approx(hi = tp.info)
-          finally seen = saved
-        }
-      case _ =>
-        mapOver(tp)
+      case tp: SkolemType => range(hi = apply(tp.info))
+      case _ => mapOver(tp)
     }
   }
 
--- a/compiler/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeOps.scala
@@ -132,9 +132,11 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
 
   /** Approximate a type `tp` with a type that does not contain skolem types. */
   object deskolemize extends ApproximatingTypeMap {
-    def apply(tp: Type) = tp match {
-      case tp: SkolemType => range(hi = apply(tp.info))
-      case _ => mapOver(tp)
+    def apply(tp: Type) = /*ctx.traceIndented(i"deskolemize($tp) at $variance", show = true)*/ {
+      tp match {
+        case tp: SkolemType => range(hi = atVariance(1)(apply(tp.info)))
+        case _ => mapOver(tp)
+      }
     }
   }
 
--- a/compiler/src/dotty/tools/dotc/core/TypeOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeOps.scala
@@ -19,122 +19,81 @@ import ast.tpd._
 trait TypeOps { this: Context => // TODO: Make standalone object.
 
   /** The type `tp` as seen from prefix `pre` and owner `cls`. See the spec
-   *  for what this means. Called very often, so the code is optimized heavily.
-   *
-   *  A tricky aspect is what to do with unstable prefixes. E.g. say we have a class
-   *
-   *    class C { type T; def f(x: T): T }
-   *
-   *  and an expression `e` of type `C`. Then computing the type of `e.f` leads
-   *  to the query asSeenFrom(`C`, `(x: T)T`). What should its result be? The
-   *  naive answer `(x: C#T)C#T` is incorrect given that we treat `C#T` as the existential
-   *  `exists(c: C)c.T`. What we need to do instead is to skolemize the existential. So
-   *  the answer would be `(x: c.T)c.T` for some (unknown) value `c` of type `C`.
-   *  `c.T` is expressed in the compiler as a skolem type `Skolem(C)`.
-   *
-   *  Now, skolemization is messy and expensive, so we want to do it only if we absolutely
-   *  must. Also, skolemizing immediately would mean that asSeenFrom was no longer
-   *  idempotent - each call would return a type with a different skolem.
-   *  Instead we produce an annotated type that marks the prefix as unsafe:
-   *
-   *     (x: (C @ UnsafeNonvariant)#T)C#T
-   *
-   *  We also set a global state flag `unsafeNonvariant` to the current run.
-   *  When typing a Select node, typer will check that flag, and if it
-   *  points to the current run will scan the result type of the select for
-   *  @UnsafeNonvariant annotations. If it finds any, it will introduce a skolem
-   *  constant for the prefix and try again.
-   *
-   *  The scheme is efficient in particular because we expect that unsafe situations are rare;
-   *  most compiles would contain none, so no scanning would be necessary.
+   *  for what this means.
    */
   final def asSeenFrom(tp: Type, pre: Type, cls: Symbol): Type =
-    asSeenFrom(tp, pre, cls, null)
+    new AsSeenFromMap(pre, cls).apply(tp)
 
-  /** Helper method, taking a map argument which is instantiated only for more
-   *  complicated cases of asSeenFrom.
-   */
-  private def asSeenFrom(tp: Type, pre: Type, cls: Symbol, theMap: AsSeenFromMap): Type = {
+  /** The TypeMap handling the asSeenFrom */
+  class AsSeenFromMap(pre: Type, cls: Symbol) extends ApproximatingTypeMap {
 
-    /** Map a `C.this` type to the right prefix. If the prefix is unstable and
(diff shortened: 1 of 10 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3842,63 +3862,188 @@ object Types {
...
+          range(derivedAppliedType(tp, tyconLo, args), derivedAppliedType(tp, tyconHi, args))
+        case _ =>
+          if (args.exists(isRange)) {
+            if (variance > 0) tp.derivedAppliedType(tycon, args.map(rangeToBounds))
+            else {
+              val loBuf, hiBuf = new mutable.ListBuffer[Type]
+              def distributeArgs(args: List[Type], tparams: List[ParamInfo]): Boolean = args match {
+                case Range(lo, hi) :: args1 =>
+                  val v = tparams.head.paramVariance
+                  if (v == 0) false
```
liufengyun: Missing "return" before "false"? Otherwise, it seems `distributeArgs` always return `true`.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -268,7 +268,20 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+                      // would not detect types to be compatible. Note: If we replace the
+                      // 2nd condition by `defDenot.symbol.maybeOwner.isType` we get lots
+                      // of failures in the `tastyBootstrap` test. Trying to compile these
+                      // files in isolation works though.
+                      // TODO: Investigate why that happens.
+                      defDenot.symbol.owner
+                    else
+                      curOwner
+                  effectiveOwner.thisType.select(name, defDenot)
+                }
```
liufengyun: I still have no idea why this change becomes necessary despite the comments. Is it due to the change of variance for NamedTypes in TypeMaps?

compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -19,131 +19,76 @@ import ast.tpd._
...
-            asSeenFrom(tp.refinedInfo, pre, cls, theMap))
-        case tp: TypeAlias if tp.variance == 1 => // if variance != 1, need to do the variance calculation
-          tp.derivedTypeAlias(asSeenFrom(tp.alias, pre, cls, theMap))
-        case _ =>
-          (if (theMap != null) theMap else new AsSeenFromMap(pre, cls))
-            .mapOver(tp)
+          case tp: ThisType =>
+            toPrefix(pre, cls, tp.cls)
+          case _: BoundType | NoPrefix =>
+            tp
```
odersky: I believe `tp == NoType` will only happen in exceptional cases. So we don't want to slow down the general logic by testing for it early.

liufengyun: Actually during the debugging, it's quite common, triggered from the Namer: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Namer.scala#L1011

odersky: That's probably still a lot less than the other cases. To be sure either way we'd have to measure (i.e. add counters).

liufengyun: I agree, it's better to handle that special case in `Namer` instead of `AsSeenFromMap`.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3842,63 +3862,188 @@ object Types {
...
   }
 
-  /** A type map that approximates NoTypes by upper or lower known bounds depending on
+  case class Range(lo: Type, hi: Type) extends UncachedGroundType {
+    assert(!lo.isInstanceOf[Range])
+    assert(!hi.isInstanceOf[Range])
+
+    override def toText(printer: Printer): Text =
+      lo.toText(printer) ~ ".." ~ hi.toText(printer)
+  }
```
odersky: Yes, good idea.

compiler/src/dotty/tools/dotc/typer/Typer.scala
odersky: I am not sure.

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -68,6 +68,15 @@ class PlainPrinter(_ctx: Context) extends Printer {
       }
     else tp
 
+  private def sameBound(lo: Type, hi: Type): Boolean =
+    try lo =:= hi
+    catch { case ex: Throwable => false }
```
smarter: I think this is a bit sloppy :). I would like `NonFatal(ex)` more, or even better: define a base trait that all non-fatal exceptions we throw in the typechecker must extend. (Or even even better: make the high-level APIs of the typer deal with exceptions themselves, clearly there is no value in propagating the exceptions up more if we're just ignoring them).

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,63 +3860,193 @@ object Types {
...
+          range(derivedRefinedType(tp, parentLo, info), derivedRefinedType(tp, parentHi, info))
+        case _ =>
+          if (parent.isBottomType) parent
+          else info match {
+            case Range(infoLo, infoHi) =>
+              def propagate(lo: Type, hi: Type) =
+                range(derivedRefinedType(tp, parent, lo), derivedRefinedType(tp, parent, hi))
+              tp.refinedInfo match {
+                case rinfo: TypeBounds =>
+                  val v = if (rinfo.isAlias) rinfo.variance * variance else variance
```
smarter: I don't understand what's going on here, could you add some comments? I've also checked and we have zero test cases where v does not end up being 0 here, so it's hard to judge the validity of the code.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,63 +3860,193 @@ object Types {
...
+                  if (v > 0) tp.derivedRefinedType(parent, tp.refinedName, rangeToBounds(info))
+                  else if (v < 0) propagate(infoHi, infoLo)
```
smarter: I find this especially suspicious, this seems to violate the invariant that `Range#lo` is a subtype of `Range#hi`, am I missing something?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,63 +3860,193 @@ object Types {
...
-      if (tycon.exists && args.forall(_.exists)) tp.derivedAppliedType(tycon, args)
-      else approx() // This is rather coarse, but to do better is a bit complicated
+      tycon match {
+        case Range(tyconLo, tyconHi) =>
+          range(derivedAppliedType(tp, tyconLo, args), derivedAppliedType(tp, tyconHi, args))
+        case _ =>
+          if (args.exists(isRange)) {
+            if (variance > 0) tp.derivedAppliedType(tycon, args.map(rangeToBounds))
+            else {
+              val loBuf, hiBuf = new mutable.ListBuffer[Type]
```
smarter: I think this case is also complicated enough to warrant a comment :).

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,63 +3860,193 @@ object Types {
...
+      case tp: Range => tp.hi
+      case _ => tp
+    }
+
+    protected def rangeToBounds(tp: Type) = tp match {
+      case Range(lo, hi) => TypeBounds(lo, hi)
+      case _ => tp
+    }
+
+    protected def atVariance[T](v: Int)(op: => T): T = {
```
smarter: I think this is worth putting in `TypeMap` itself, considering how common the pattern `val saved = variance; variance = ...` is. Also maybe worth marking `inline` ?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,63 +3860,193 @@ object Types {
...
    *     variance < 0 : approximate by lower bound
-   *     variance = 0 : propagate NoType to next outer level
+   *     variance = 0 : propagate bounds to next outer level
    */
   abstract class ApproximatingTypeMap(implicit ctx: Context) extends TypeMap { thisMap =>
-    def approx(lo: Type = defn.NothingType, hi: Type = defn.AnyType) =
-      if (variance == 0) NoType
-      else apply(if (variance < 0) lo else hi)
 
+    protected def range(lo: Type, hi: Type) =
```
smarter: Calling this method `range` makes it a bit too easy to confuse it with `Range`. Perhaps call it `between` ?

odersky: The current situation is analogous to `range` / `Range` for collection ranges. I believe that's a good thing.

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -68,6 +68,15 @@ class PlainPrinter(_ctx: Context) extends Printer {
...
+
+  private def homogenizeArg(tp: Type) = tp match {
```
odersky: In fact, the name was correct, but the condition to do this only under `homogenizedView` was lacking.

odersky:
I'll roll the comment fixes into #2945. Can you review that one next? 😄

smarter:
Sure!

liufengyun:
test performance please

smarter:
Merged as part of https://github.com/lampepfl/dotty/pull/2970 (and re-merged as part of https://github.com/lampepfl/dotty/pull/2945)

PR #2910, opened 2017-07-24
Link from tasty

nicolasstucki:


nicolasstucki:
@smarter deduplicated `ClassSymbol#tree`

nicolasstucki:
The requested change has been made.

smarter:
Thanks, not sure if I'll have time to review this today, this isn't urgent anyway is it?

nicolasstucki:
It is not urgent.
It will help me expose some bug later on and make sure that the current state does not break.

nicolasstucki:
Will need rebase on #2958

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -558,32 +558,44 @@ object Symbols {
...
+      val t = unitTree
+      if (t.isEmpty) t
+      else findTree(tree).get
+    }
+
+    /** If this is a top-level class, and if `-Yretain-trees` or `-YlinkOptimise` is set,
+      * return the PackageDef tree for this class, otherwise EmptyTree.
+      * This will force the info of the class.
+      */
+    def unitTree(implicit ctx: Context): tpd.Tree /* tpd.PackageDef | tpd.TypeDef | tpd.EmptyTree */ = {
```
smarter: I'm unhappy about having both `unitTree` and `tree`. I think we should just have `unitTree` (but give it a different name like just `tree` because it's not the full compilation unit tree, it's only the current class tree wrapped with PackageDefs).

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -558,32 +558,44 @@ object Symbols {
...
+          Some(tree)
+        case _ =>
+          None
+      }
+      val t = unitTree
+      if (t.isEmpty) t
+      else findTree(tree).get
+    }
+
+    /** If this is a top-level class, and if `-Yretain-trees` or `-YlinkOptimise` is set,
```
smarter: Can we please decide if options should be written `-like-this` or `-likeThis` and stick to it? Having both is really annoying.

nicolasstucki: Oh, it is actually `-Xlink-optimise` and this is a typo.

nicolasstucki:
@smarter ping review

nicolasstucki:
Rebased

nicolasstucki:
Rebased

compiler/src/dotty/tools/dotc/CompilationUnit.scala
```diff
@@ -20,3 +23,19 @@ class CompilationUnit(val source: SourceFile) {
...
   var pickled: Map[ClassSymbol, Array[Byte]] = Map()
 }
+
+object CompilationUnit {
+
+  /** Make a compilation unit for top class `clsd` with the contends of the `unpickled` */
+  def mkCompilationUnit(clsd: ClassDenotation, unpickled: Tree)(implicit ctx: Context): CompilationUnit = {
+    val unit1 = new CompilationUnit(new SourceFile(clsd.symbol.sourceFile, Seq()))
+    unit1.tpdTree = unpickled
+    force.traverse(unit1.tpdTree)
```
smarter: Do you actually need to force the tree for the linker?

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -47,6 +47,7 @@ class Compiler {
       List(new PostTyper),          // Additional checks and cleanups after type checking
       List(new sbt.ExtractAPI),     // Sends a representation of the API of classes to sbt via callbacks
       List(new Pickler),            // Generate TASTY info
+      List(new LinkAll),            // Link all
```
smarter: Not a very detailed description :)

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -102,6 +102,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val YcheckAllPatmat = BooleanSetting("-Ycheck-all-patmat", "Check exhaustivity and redundancy of all pattern matching (used for testing the algorithm)")
   val YretainTrees = BooleanSetting("-Yretain-trees", "Retain trees for top-level classes, accessible from ClassSymbol#tree")
   val YshowTreeIds = BooleanSetting("-Yshow-tree-ids", "Uniquely tag all tree nodes in debugging output.")
+  val YRunClasspath = PathSetting("-YRunClasspath", "Specify where to find user class files while executing run tests.", defaultClasspath)
```
smarter: If this only for tests it shouldn't be a compiler option. You can use a property instead like the stuff in compiler/test/dotty/Properties.scala

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -113,6 +114,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val YoptPhases = PhasesSetting("-Yopt-phases", "Restrict the optimisation phases to execute under -optimise.")
   val YoptFuel = IntSetting("-Yopt-fuel", "Maximum number of optimisations performed under -optimise.", -1)
   val optimise = BooleanSetting("-optimise", "Generates faster bytecode by applying local optimisations to the .program") withAbbreviation "-optimize"
+  val XlinkOptimise = BooleanSetting("-Xlink-optimise", "Link class files.").withAbbreviation("-Xlink-optimize")
```
smarter: Longer description?

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -555,35 +555,29 @@ object Symbols {
 
     type ThisName = TypeName
 
-    /** If this is a top-level class, and if `-Yretain-trees` is set, return the TypeDef tree
-     *  for this class, otherwise EmptyTree. This will force the info of the class.
-     */
-    def tree(implicit ctx: Context): tpd.Tree /* tpd.TypeDef | tpd.EmptyTree */ = {
+    /** If this is a top-level class, and if `-Yretain-trees` or `-Xlink-optimise` is set,
```
smarter: This comment is not 100% true since if this is a top-level class and -Xlink-optimise is set, but the class comes from the sources and not tasty, it will return EmptyTree.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -555,35 +555,29 @@ object Symbols {
...
+        val body = unpickler.body(ctx.addMode(Mode.ReadPositions))
+        myTree = body.headOption.getOrElse(tpd.EmptyTree)
         unpickler = null
-
-        myTree = findTree(unpickledTree).get
       }
       myTree
     }
-    private[dotc] var myTree: tpd.Tree = tpd.EmptyTree
+    private var myTree: tpd.Tree /* tpd.PackageDef | tpd.TypeDef | tpd.EmptyTree */ = tpd.EmptyTree
```
smarter: Could be `private[this]`

compiler/src/dotty/tools/dotc/interactive/SourceTree.scala
```diff
@@ -43,13 +43,15 @@ object SourceTree {
         sym.sourceFile == null) // FIXME: We cannot deal with external projects yet
       None
     else {
-      sym.tree match {
-        case tree: tpd.TypeDef =>
+      import ast.Trees._
+      def findTree(tree: tpd.Tree): Option[SourceTree] = tree match {
```
smarter: Rename to `sourceTreeOfClass` ?

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -154,6 +154,9 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
     cpy.Template(impl)(self = EmptyValDef)
   }
 
+  override def transformPackageDef(tree: PackageDef)(implicit ctx: Context, info: TransformerInfo): Tree =
```
smarter: Add a comment explaining why this is needed?

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -154,6 +154,9 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
...
+    if (tree.stats.isEmpty) theEmptyTree else tree
```
smarter: Why `theEmptyTree` and not `EmptyTree` ?

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
```diff
@@ -0,0 +1,78 @@
...
+ *
+ *  TODO: Next step is to only load compilation units reachable in the call graph
+ */
+class LinkAll extends MiniPhaseTransform {
+  import tpd._
+  import LinkAll._
+
+  override def phaseName = "linkAll"
+
+  override def prepareForUnit(tree: tpd.Tree)(implicit ctx: Context): TreeTransform = NoTransform
```
smarter: Why?

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  override def runOn(units: List[CompilationUnit])(implicit ctx: Context): List[CompilationUnit] =
+    if (ctx.settings.XlinkOptimise.value) super.runOn(allUnits(Set.empty, units.toSet, Set.empty))
+    else super.runOn(units)
+
+  /** Loads and processes new compilation units, possibly loading more units. */
+  private def allUnits(processed: Set[CompilationUnit], unprocessed: Set[CompilationUnit], loadedClasses: Set[ClassDenotation])(implicit ctx: Context): List[CompilationUnit] = {
```
smarter: I would use an inner method for the recursive calls and not have the parameters processed and loadedClasses on the outer method.

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
```diff
@@ -0,0 +1,78 @@
...
+    override def apply(acc: Set[ClassDenotation], tree: tpd.Tree)(implicit ctx: Context): Set[ClassDenotation] = tree match {
+      case New(tpt) => accum(acc, tpt.tpe.classSymbol)
+      case AppliedTypeTree(tpt, _) if inParents => accum(acc, tpt.symbol)
+      case tree: RefTree if inParents || tree.symbol.is(Module) =>
+        foldOver(accum(acc, tree.symbol), tree)
+      case tree @ Template(constr, parents, self, _) =>
+        val acc1 = this(acc, constr)
+        inParents = true
+        val acc2 = this(acc1, parents)
+        inParents = false
```
smarter: Don't think this is correct in general, you assume that inParents will be false before you set it to true, but you can have a Template inside the parents of a Template, e.g:
```scala
class A(x: Int)
class B extends A({ class Foo; 1 })
```

compiler/src/dotty/tools/dotc/CompilationUnit.scala
nicolasstucki: When we first tried it with @DarkDimius we needed it. But it looks like we do not need it anymore.

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
nicolasstucki: see comment

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
nicolasstucki: This works, I added a test for it.

nicolasstucki:
@smarter, all requested changes have been done.

PR #2912, opened 2017-07-25
Fix #2895: Two fixes for inlining

### LINKED ISSUE #2895: Method inlined in a subclass is not found when called from the parent class
See the following snippet: https://scastie.scala-lang.org/nAtEZkgdSPuuQWXfiS4XrA

If we remove the `inline` keyword it works fine.

odersky:
Two fixes:

 1. Generate bridges for inline methods 

   There was a test before that explicitly suppressed bridge generation
   for inline methods. I am not sure why. @DarkDimius, does the one ring a bell?

 2. Sharpen condition when to elide a `this` binding.


DarkDimius:
LGTM

PR #2913, opened 2017-07-25
Run benchmarks directly from sbt

nicolasstucki:
To benchmark use 
```
dotty-bench/jmh:run XYZ.scala
```

Setting up forks, warmups, ireations can be done in `bench/src/main/scala/Benchmarks.scala`


project/Build.scala
```diff
@@ -828,7 +828,8 @@ object Build {
     dependsOn(`dotty-compiler`).
     settings(commonNonBootstrappedSettings).
     settings(
-      mainClass in (Jmh, run) := Some("dotty.tools.benchmarks.Bench")   // custom main for jmh:run
+      mainClass in (Jmh, run) := Some("dotty.tools.benchmarks.Bench"), // custom main for jmh:run
```
DarkDimius: Could you please print the entire command line that was used to launch the behcnmark vm?

bench/src/main/scala/Benchmarks.scala
```diff
@@ -18,17 +18,25 @@ object Bench {
...
-               .timeUnit(TimeUnit.MICROSECONDS)
-               .forks(5)
-               .warmupIterations(5)
-               .measurementIterations(10)
+               .timeUnit(TimeUnit.MILLISECONDS)
+               .forks(1)
+               .warmupIterations(12)
+               .measurementIterations(20)
                .resultFormat(ResultFormatType.CSV)
                .result("result.csv")
```
liufengyun: We can remove this file, and there'll be no `result.csv`, which is useless.

project/Build.scala
nicolasstucki: @liufengyun do you know if that is possible with the jmh plugin?

DarkDimius:
If it is not, this should be a strong reason to prefer a bash script.

project/Build.scala
liufengyun: When the command runs, we'll have the output in the console, which shows the VM options:

# JMH 1.18 (released 134 days ago)
# VM version: JDK 1.8.0_131, VM 25.131-b11
# VM invoker: /usr/lib/jvm/java-8-oracle/jre/bin/java
# VM options: -Xbootclasspath/a:some/jars
# Warmup: 10 iterations, 1 s each
# Measurement: 20 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: dotty.tools.benchmarks.Worker.compile

liufengyun:
We can add a script `bin/bench`, WDYT @nicolasstucki ?

```shell
sbt "dotty-bench/jmh:run $@"
```

nicolasstucki:
@DarkDimius this is only intended as way quick to benchmark changes in the compiler itself. Hence sbt is the simplest option to make sure that everything is up to date with the latest changes in the compiler code. If we want reproducible benchmarks from command line  we should use the benchmarks from https://github.com/liufengyun/bench instead.

nicolasstucki:
@liufengyun I added a couple of new commits.

liufengyun:
@nicolasstucki I thought I already merged this PR, well done :)

PR #2914, opened 2017-07-25
Fix cyclic reference errors in collection strawman

odersky:
This should make collection strawman #137 build. There were some "interesting" F-bounded types which triggered a series of problems when read back with Tasty unpicklers. First, the underlying error message ("wrong number of arguments") was hidden because it caused a CRE. Next, the wrong number of argument problems demanded that we are more careful reading applicatons. Once that was done we got Stackoverflows. To avoid these, we have to maintain any LazyRefs we have in the original types in the pickle information. 

Review by @szeiger or @julienrf?

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
@@ -996,7 +996,12 @@ class TreeUnpickler(reader: TastyReader, nameAtRef: NameRef => TermName, posUnpi
               val refinements = readStats(refineCls, end)(localContext(refineCls))
               RefinedTypeTree(parent, refinements, refineCls)
             case APPLIEDtpt =>
-              AppliedTypeTree(readTpt(), until(end)(readTpt()))
+              // If we do directly a tpd.AppliedType tree we might get a
+              // wrong number of arguments in some scenarios reading F-bounded
+              // types. This came up in #137 of collection strawman.
+              val tycon = readTpt()
+              val args = until(end)(readTpt())
+              untpd.AppliedTypeTree(tycon, args).withType(tycon.tpe.safeAppliedTo(args.tpes))
             case ANDtpt =>
               val tpt1 = readTpt()
               val tpt2 = readTpt()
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -293,6 +293,11 @@ class TreeUnpickler(reader: TastyReader, nameAtRef: NameRef => TermName, posUnpi
           RecType(rt => registeringType(rt, readType()))
         case RECthis =>
           RecThis(readTypeRef().asInstanceOf[RecType])
+        case LAZYref =>
+          val rdr = fork
+          def readUnderlying() = rdr.readType()
+          skipTree()
+          LazyRef(readUnderlying)
```
smarter: This means LazyRefs will keep the context from which they were unpickled around, if these LazyRefs are subsequently stored in some cache, that could lead to space leaks, no?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -708,10 +709,15 @@ object messages {
...
-      case symbol   => symbol.showFullName
-    }
+    private val prettyName =
+      try
+        fntpe.termSymbol match {
+          case NoSymbol => fntpe.show
+          case symbol   => symbol.showFullName
+        }
+      catch {
+        case NonFatal(ex) => fntpe.show
```
smarter: If this is only for cyclic references, could we do `case ex: CyclicReference` instead?

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
odersky: Yes, correct. Maybe the best option is to allow a context to be passed into a LazyRef. In the case of unpickling, this is fine; we don't need to hang on to more than the reader (I think, at least that's what we do for the other completers).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
odersky: I think we by default we should not throw any exception from printing. It detracts from the underlying cause.

smarter: Sure, but we shouldn't be throwing any kind of exceptions, and silently catching them all might hide issues.

DarkDimius:
@odersky, instead of pickling lazy refs, could we create them at un-pickling similarly to how we create them at typer?

odersky:
> @odersky, instead of pickling lazy refs, could we create them at un-pickling similarly to how we create them at typer?

I tried that, but it didn't help. Did not get to find out why.

odersky:
Merging now to get the strawman unstuck, but I opened an issue to revert the LazyRef addition if possible. /cc @julienrf

julienrf:
Thanks @odersky!

PR #2918, opened 2017-07-26
Fix #2853: require at least one parent after `extends`

### LINKED ISSUE #2853: Parsing: `extends` with 0 parents compiles
Following bug is discovered by @nicolasstucki during a discussion.

The compiler should produce an error message, but it accepts the code below:

```Scala
class A extends { }
object A extends { }
```

liufengyun:
Fix #2853: require at least one parent after `extend`


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2312,7 +2312,12 @@ object Parsers {
     /** TemplateOpt = [`extends' Template | TemplateBody]
      */
     def templateOpt(constr: DefDef): Template =
-      if (in.token == EXTENDS) { in.nextToken(); template(constr)._1 }
+      if (in.token == EXTENDS) {
+        val offset = in.skipToken()
+        val (tmpl, _) = template(constr)
+        if (tmpl.parents.isEmpty) syntaxError("parents expected after `extend`", offset)
```
nicolasstucki: `extend` -> `extends`

nicolasstucki:
You caught a bug in test file `tests/run/t6090.scala`

nicolasstucki:
LGTM

DarkDimius:
We've decided to keep this behavior:
 - it's consistent with the grammar
 - it's consistent with scala2 behavior.

PR #2919, opened 2017-07-26
#1589 Move Parsers:2177 err to new format

lorandszakacs:
Move error for `Parsers:2177` to new format.

Actual line numbers have started getting out of sync with the description in issue #1589. But it should be quite straightforward to see that this was also necessary.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1702,4 +1702,12 @@ object messages {
       hl"It is not allowed to combine `private` and `protected` modifiers even if they are qualified to different scopes"
   }
 
+  case class ExpectedStartOfTopLevelDefinition()(implicit ctx: Context)
+    extends Message(ExpectedStartOfTopLevelDefinitionID) {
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1702,4 +1702,12 @@ object messages {
       hl"It is not allowed to combine `private` and `protected` modifiers even if they are qualified to different scopes"
   }
 
+  case class ExpectedStartOfTopLevelDefinition()(implicit ctx: Context)
+    extends Message(ExpectedStartOfTopLevelDefinitionID) {
+    val kind = "Syntax"
+    val msg = "expected start of definition"
+    val explanation =
+      hl"you have to provide either ${"class"}, ${"object"}, or ${"enum"} definitions after qualifiers"
```
nicolasstucki: Missing `trait`.

nicolasstucki:
Otherwise LGTM

lorandszakacs:
@nicolasstucki, thanks!

I updated it.

nicolasstucki:
Thanks

PR #2922, opened 2017-07-26
Fix #2412: Local error causes spurious errors to be reported

### LINKED ISSUE #2412: Local error in block causes spurious errors in outer tree to be reported
This code should only report one error:
```scala
object Test {
  def test(foo: List[String]): Unit = {
    foo.filter(f => {
      iDontExist // error
      true
    })
  }
}
```
But with dotty we get two:
```scala
-- Error: try/err.scala:3:15 ---------------------------------------------------
3 |    foo.filter(f => {
  |               ^
  |               missing parameter type for parameter f, expected = ?
-- [E006] Unbound Identifier Error: try/err.scala:4:6 --------------------------
4 |      iDontExist
  |      ^^^^^^^^^^
  |      not found: iDontExist
```
Even though the type of `f` is correctly inferred as `String`, as can be seen using `-Xprint:frontend`. Interestingly this does not happen if we replace `filter` by `map`. This seems to be related to the way we fallback in `typedApply`: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Applications.scala#L695-L707

smarter:


smarter:
Ping for review.

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -185,15 +187,28 @@ object ProtoTypes {
     /** A map recording the typer states in which arguments stored in myTypedArg were typed */
     private var evalState: SimpleMap[untpd.Tree, TyperState] = SimpleMap.Empty
 
-    def isMatchedBy(tp: Type)(implicit ctx: Context) =
+    private[this] var thisCallArgCtxUsed: Boolean = false
+
+    /** Evalute `op` with the proper context to type the arguments. */
+    protected def withArgCtx[T](op: Context => T)(implicit ctx: Context) = {
+      if (isSelfConstrCall && !thisCallArgCtxUsed) {
+        // Context#thisCallArgContext is not idempotent so we must only use it
```
odersky: Can we use `Mode.InSuperCall` as an indicator whether we are already in the right context, instead of defining a special field for this?

smarter: Yes that should work, I'll give it a try.

smarter:
> It would be good to motivate this in more detail, because I am not convinced yet. Logically, it seems correct to store the context in the prototype and evaluate arguments in this context.

That's fine as long as there is exactly one way to type the arguments and no backtracking takes place, but there's plenty of backtracking in `typedApply` so using the stored context causes weird issues like reporting errors from branches we backtracked out of.

> You compensate for supercall arguments by passing a special flag. But how do you know architecturally that you do not need other compensations as well?

Yes, it's a bit dicey, it relies on the fact that `FunProto#typedArgs` will always be called in a Context that is an inner Context from the one where we created the `FunProto`, so the arguments should still be typeable, and they will also have the correct TyperState/Reporter/... needed for backtracking to work correctly. Ideally I think we would have a design where we don't store untyped trees in prototypes to avoid this kind of things.

smarter:
Superceded by https://github.com/lampepfl/dotty/pull/4871

PR #2926, opened 2017-07-27
Implement value classes with phantom parameters

nicolasstucki:
In this implementation the first parameter of the
value class must be a subtype of scala.Any. Every
other parameter must be phantom.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -493,7 +493,7 @@ object Checking {
             if (param.is(Mutable))
               ctx.error(ValueClassParameterMayNotBeAVar(clazz, param), param.pos)
             if (param.info.isPhantom)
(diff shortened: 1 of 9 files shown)


compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -489,13 +489,17 @@ object Checking {
...
         clParamAccessors match {
-          case List(param) =>
+          case param :: params =>
             if (param.is(Mutable))
               ctx.error(ValueClassParameterMayNotBeAVar(clazz, param), param.pos)
             if (param.info.isPhantom)
-              ctx.error("value class parameter must not be phantom", param.pos)
-          case _ =>
-            ctx.error(ValueClassNeedsExactlyOneValParam(clazz), clazz.pos)
+              ctx.error("value class first parameter must not be phantom.", param.pos)
```
OlivierBlanvillain: extra `.`

PR #2927, opened 2017-07-27
Fix #2430: Make Erasure pass Ycheck

### LINKED ISSUE #2430: Value types are not properly compared in `TypeComparer`
The following subtyping test fails `Ycheck` after erasure:

```scala
class Test1(val x: Int) extends AnyVal
class Test2(val y: Test1) extends AnyVal
```

In tree checker:

```scala
tree.tpe <:< pt // tree.tpe = ErasedValueType(Test2, ErasedValueType(Test1, Int))($this)
                //       pt = Int
```

This is what causes erasure not to be `Ycheck`able. The next phase, `ElimErasedValueType`, removes the `ErasedValueType`s and casts so that it is again, `Ycheck`able.

ping @odersky

smarter:
During Erasure, a value class A with underlying type U has its type
semi-erased to ErasedValueType(A, [semi-erasure of U]). To make Erasure
typecheck, special cast functions A.evt2u$ and A.u2evt$ are used to cast
between the underlying type and the class EVT type. Before this commit,
this conversion was not always correctly done for nested value classes,
this did not lead to problems in practice but broke -Ycheck:erasure.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Erasure.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Erasure.scala
@@ -239,26 +239,44 @@ object Erasure {
      *  Casts from and to ErasedValueType are special, see the explanation
      *  in ExtensionMethods#transform.
      */
-    def cast(tree: Tree, pt: Type)(implicit ctx: Context): Tree = {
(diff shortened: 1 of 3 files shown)


DarkDimius:
Looks like the right approach. Great to ycheck erasure again.

PR #2931, opened 2017-07-28
Fix #2916: Reorder parameters

### LINKED ISSUE #2916: Named arguments reorder effects
```scala
def foo(bar: Int, bat: Int) = 1
def p(x: Int) = { println(x); x }
```
Scala 2.12:
```scala
scala> foo(bat = p(1), bar = p(2)) 
1
2
```
Dotty:
```scala
scala> foo(bat = p(1), bar = p(2)) 
2
1
```

nicolasstucki:
This fixes the ordering issue described in #2916.
Additionally fixes another related issue where not all parameters that needed to be lifted where lifted. All these in presence of reordering done when having combination of effectful/pure arguments.


compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -653,9 +653,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
 
             typedArgs = liftArgs(liftedDefs, methType, liftable) ++ rest
 
-            val newOrder = (liftedDefs zip indices).sortBy(_._2).unzip._1
-            liftedDefs = ListBuffer.empty
-            liftedDefs ++= newOrder
+            liftedDefs = (liftedDefs zip indices).sortBy(_._2).unzip._1.to
```
biboudis: 😮

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -580,40 +584,78 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+          case nil => n
+        }
+      case nil =>
+        ys match {
+          case EmptyTree :: ys1 => firstDiffSkipInitEmptyTrees(xs, ys1, n)
+          case y :: ys1 => n
+          case nil => -1
+        }
+    }
+
```
odersky: Why not simply: 

    firstDiff(xs, ys.dropWhile(_ == EmptyTree), n)

?

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -580,40 +584,78 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
-          if (!sameSeq(args, orderedArgs) && !isJavaAnnotConstr(methRef.symbol)) {
+          if (firstDiffSkipInitEmptyTrees(args, orderedArgs) >= 0 && !isJavaAnnotConstr(methRef.symbol)) {
             // need to lift arguments to maintain evaluation order in the
             // presence of argument reorderings.
             liftFun()
-            val eqSuffixLength = firstDiff(app.args.reverse, orderedArgs.reverse)
+            val eqSuffixLength = firstDiffSkipInitEmptyTrees(app.args.reverse, orderedArgs.reverse)
             val (liftable, rest) = typedArgs splitAt (typedArgs.length - eqSuffixLength)
+
+            // Mapping of index of each `liftable` into original args ordering
```
odersky: Lots of state here; better to move this out to a separate method and describe what it does in a doc comment.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -580,40 +584,78 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+              case arg =>
+                if (args.contains(arg)) {
+                  indices += prefixShift + args.indexOf(arg)
+                } else {
+                  indices += prefixShift + nextDefaultParamIndex
+                  nextDefaultParamIndex += 1
+                }
+            }
+
             typedArgs = liftArgs(liftedDefs, methType, liftable) ++ rest
```
odersky: Why does it not work if we lift args before reordering?

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -605,13 +606,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
-            val (liftable, rest) = typedArgs splitAt (typedArgs.length - eqSuffixLength)
-            typedArgs = liftArgs(liftedDefs, methType, liftable) ++ rest
+
+            // lift arguments in the definition order
+            val argDefBuff = mutable.ListBuffer.empty[Tree]
+            typedArgs = liftArgs(argDefBuff, methType, typedArgs)
+
+            def orderedArgDefs = {
+              val impureArgs = typedArgBuf.filterNot { // pure are not lifted by liftArgs
+                case NamedArg(_, arg) => isPureExpr(arg)
```
odersky: Let's instead add a case for `NamedArg` in `TreeInfo.exprPurity`. That will avoid the special treatment here.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -605,13 +606,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
             // need to lift arguments to maintain evaluation order in the
             // presence of argument reorderings.
+
             liftFun()
-            val eqSuffixLength = firstDiff(app.args.reverse, orderedArgs.reverse)
-            val (liftable, rest) = typedArgs splitAt (typedArgs.length - eqSuffixLength)
-            typedArgs = liftArgs(liftedDefs, methType, liftable) ++ rest
+
+            // lift arguments in the definition order
+            val argDefBuff = mutable.ListBuffer.empty[Tree]
```
odersky: Usual naming convention in `dotc` is just `Buf` for buffer, i.e no double ff.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -605,13 +606,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+            typedArgs = liftArgs(argDefBuff, methType, typedArgs)
+
+            def orderedArgDefs = {
+              val impureArgs = typedArgBuf.filterNot { // pure are not lifted by liftArgs
+                case NamedArg(_, arg) => isPureExpr(arg)
+                case arg => isPureExpr(arg)
+              }
+              var defaultParamIndex = args.size
```
odersky: Should be a `val`

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -605,13 +606,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
             liftFun()
-            val eqSuffixLength = firstDiff(app.args.reverse, orderedArgs.reverse)
-            val (liftable, rest) = typedArgs splitAt (typedArgs.length - eqSuffixLength)
-            typedArgs = liftArgs(liftedDefs, methType, liftable) ++ rest
+
+            // lift arguments in the definition order
+            val argDefBuff = mutable.ListBuffer.empty[Tree]
+            typedArgs = liftArgs(argDefBuff, methType, typedArgs)
+
+            def orderedArgDefs = {
```
odersky: Can be `val` instead of `def`.

odersky: Since the logic here is somewhat tricky, it would be good to add a comment with an example, what orderedArgDefs is supposed to achieve.

nicolasstucki:
Ready for review

nicolasstucki:
test performance please

PR #2932, opened 2017-07-29
Harden IDE some more

odersky:
More fixes that prevent crashes and infinite loops observed in the IDE. 

Also, some more fundamental changes that I added here to test them out in isolation of the other work on applied types I am doing.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -115,7 +115,7 @@ object Trees {
      */
     def withType(tpe: Type)(implicit ctx: Context): ThisTree[Type] = {
       if (tpe.isInstanceOf[ErrorType])
-        assert(ctx.reporter.errorsReported)
+        assert(ctx.mode.is(Mode.Interactive) || ctx.reporter.errorsReported)
       else if (Config.checkTreesConsistent)
         checkChildrenTyped(productIterator)
       withTypeUnchecked(tpe)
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -268,7 +268,20 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
             if (qualifies(defDenot)) {
               val found =
                 if (isSelfDenot(defDenot)) curOwner.enclosingClass.thisType
-                else curOwner.thisType.select(name, defDenot)
+                else {
+                  val effectiveOwner =
+                    if (curOwner.isTerm && defDenot.symbol.isType)
+                      // Don't mix NoPrefix and thisType prefixes, since type comparer
+                      // would not detect types to be compatible. Note: If we replace the
```
smarter: What kind of issue did you see before this change? Could this be boiled down to a test case?

odersky: It's only in the applied type branch that this became an issue. We saw the same type parameter TypeRef once with a NoPrefix prefix, the other with a ThisType. They should be the same but subtype checking failed. We could make subtype checking more lenient, but I think it's better not to have variations how we construct these things. I believe the current scheme hides the difference because we follow an alias from TypeParamAccessor to TypeParam, that's why we did not see it so far.

odersky:
Rebased to master.

odersky:
I collected all IDE improvements here. Will submit the other improvements separately.

PR #2934, opened 2017-07-29
Change variances in TypeMaps and Accumulators

odersky:
Based on #2932.

Also contains the owner change for typed idents.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -268,7 +268,20 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
             if (qualifies(defDenot)) {
               val found =
                 if (isSelfDenot(defDenot)) curOwner.enclosingClass.thisType
-                else curOwner.thisType.select(name, defDenot)
+                else {
+                  val effectiveOwner =
+                    if (curOwner.isTerm && defDenot.symbol.isType)
+                      // Don't mix NoPrefix and thisType prefixes, since type comparer
+                      // would not detect types to be compatible. Note: If we replace the
+                      // 2nd condition by `defDenot.symbol.maybeOwner.isType` we get lots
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -268,7 +268,20 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
                 if (isSelfDenot(defDenot)) curOwner.enclosingClass.thisType
-                else curOwner.thisType.select(name, defDenot)
+                else {
+                  val effectiveOwner =
+                    if (curOwner.isTerm && defDenot.symbol.isType)
+                      // Don't mix NoPrefix and thisType prefixes, since type comparer
+                      // would not detect types to be compatible. Note: If we replace the
+                      // 2nd condition by `defDenot.symbol.maybeOwner.isType` we get lots
+                      // of failures in the `tastyBootstrap` test. Trying to compile these
+                      // files in isolation works though.
```
smarter: `testPickling` also failed, here's a simplified way to reproduce the failure:
```shell
./bin/dotc -Xprint-types -Ytest-pickler compiler/src/dotty/tools/io/File.scala compiler/src/dotty/tools/io/Jar.scala
```
This fails with a pickling difference, here's the diff:
```diff
--- before-pickling.txt 2017-07-30 16:10:33.955539823 +0200
+++ after-pickling.txt  2017-07-30 16:10:33.955539823 +0200
@@ -111,7 +111,7 @@
           <<new java.util.jar.JarFile:java.util.jar.JarFile>:
             ((x$0: java.io.File): java.util.jar.JarFile)
           >
-        (<<Jar.this.file:dotty.tools.io.File>.jfile:=> dotty.tools.io.JFile>):
+        (<<Jar.this.file:dotty.tools.io.File>.jfile:<notype>>):
           java.util.jar.JarFile
         >
       lazy val manifest: scala.Option[java.util.jar.Manifest] = 
```

PR #2942, opened 2017-08-02
Vulpix should catch and report all exceptions during compilation of test cases

gsps:
Vulpix currently only catches `NonFatal` exceptions in `tryCompile`, silently omitting exceptions such as those for stack overflows. Worse yet, such exceptions are then caught somewhere higher up in vulpix, but `registerCompletion(1)` is never called for the respective failing test cases, causing vulpix to hang in an idle state.

Vulpix should arguably catch all sorts of exceptions and also report the respective stack traces. I'd be grateful if someone also pointed out where such fatal exceptions were currently caught, perhaps that exception handler higher up should actually abort testing.

### DIFF
--- a/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
+++ b/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
@@ -324,7 +324,7 @@ trait ParallelTesting extends RunnerOrchestration { self =>
         if (!isInteractive) realStdout.println(testing)
         op
       } catch {
-        case NonFatal(e) => {
+        case e: Throwable => {
           // if an exception is thrown during compilation, the complete test
           // run should fail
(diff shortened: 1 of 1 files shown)


felixmulder:
From a quick glance at the java `ExecutorService` API you can gleam that the thread pool returns a future on `pool.submit` and that said future is never "gotten" just executed by side-effect. Thus, what happens - is most likely - that the exception bleeds out into stderr, but never kills the executor service.

I'm sort of feeling that this is a feature rather than a bug (with the exception (get it ? haha) of your change to catch all). But, maybe there's something I'm not seeing....

gsps:
Fair enough, I guess now that we catch all exceptions during compilation in that one place it won't be a problem anyways. Thanks for looking into it!

PR #2945, opened 2017-08-02
Implement avoid in terms of AproximatingTypeMap

odersky:
Based on #2908. Last two commits are new.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3947,7 +3947,7 @@ object Types {
               else if (!infoLo.isAlias && !infoHi.isAlias) propagate(infoLo, infoHi)
               else range(tp.bottomType, tp.topType)
                 // Using `parent` instead of `tp.topType` would be better for normal refinements,
-                // but it would also turn *-types to a hk-types, which is not what we want.
+                // but it would also turn *-types into hk-types, which is not what we want.
                 // We should revisit this point in case we represent applied types not as refinements anymore.
             case Range(infoLo, infoHi) =>
               propagate(infoLo, infoHi)
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3679,13 +3679,25 @@ object Types {
 
   // ----- TypeMaps --------------------------------------------------------------------
 
-  abstract class TypeMap(implicit protected val ctx: Context) extends (Type => Type) { thisMap =>
+  /** Common base class of TypeMap and TypeAccumulator */
+  abstract class VariantTraversal {
+    protected[core] var variance = 1
+
+    @inline protected def atVariance[T](v: Int)(op: => T): T = {
+      val saved = variance
+      variance = v
+      val res = op
+      variance = saved
+      res
+    }
+  }
+
+  abstract class TypeMap(implicit protected val ctx: Context)
+  extends VariantTraversal with (Type => Type) { thisMap =>
 
     protected def stopAtStatic = true
 
     def apply(tp: Type): Type
-
-    protected[core] var variance = 1
 
     protected def derivedSelect(tp: NamedType, pre: Type): Type =
       tp.derivedSelect(pre)
@@ -3724,16 +3736,13 @@ object Types {
         case tp: NamedType =>
           if (stopAtStatic && tp.symbol.isStatic) tp
           else {
-            val saved = variance
-            variance = variance max 0
+            val prefix1 = atVariance(variance max 0)(this(tp.prefix))
               // A prefix is never contravariant. Even if say `p.A` is used in a contravariant
               // context, we cannot assume contravariance for `p` because `p`'s lower
               // bound might not have a binding for `A` (e.g. the lower bound could be `Nothing`).
               // By contrast, covariance does translate to the prefix, since we have that
               // if `p <: q` then `p.A <: q.A`, and well-formedness requires that `A` is a member
               // of `p`'s upper bound.
-            val prefix1 = this(tp.prefix)
-            variance = saved
             derivedSelect(tp, prefix1)
           }
         case _: ThisType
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,84 +3861,222 @@ object Types {
...
+              // If infoLo and infoHi are both aliases with the same non-zero variance
+              // we can propagate to a range of the refined types. If they are both
+              // non-alias ranges we know that infoLo <:< infoHi and therefore we can
+              // propagate to refined types with infoLo and infoHi as bounds.
+              // In all other cases, Nothing..Any is the only interval that contains
+              // the range. i966.scala is a test case.
+              if (v1 > 0 && v2 > 0) propagate(infoLo, infoHi)
+              else if (v1 < 0 && v2 < 0) propagate(infoHi, infoLo)
+              else if (!infoLo.isAlias && !infoHi.isAlias) propagate(infoLo, infoHi)
+              else range(tp.bottomType, tp.topType)
```
smarter: Could `parent` be used instead of `tp.topType` ?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3851,84 +3861,222 @@ object Types {
...
+                case arg :: args1 =>
+                  loBuf += arg; hiBuf += arg
+                  distributeArgs(args1, tparams.tail)
+                case nil =>
+                  true
+              }
+              if (distributeArgs(args, tp.typeParams))
+                range(tp.derivedAppliedType(tycon, loBuf.toList),
+                      tp.derivedAppliedType(tycon, hiBuf.toList))
+              else range(tp.bottomType, tp.topType)
```
smarter: Could `tp.superType` be used instead of `tp.topType` ?

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: I added a comment explaining why not:

                // Using `parent` instead of `tp.topType` would be better for normal refinements,
                // but it would also turn *-types to a hk-types, which is not what we want.
                // We should revisit this point in case we represent applied types not as refinements anymore.

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: Not sure `tp.superType` is not necessarily a supertype of the derived application because
the arguments have changed. We could try to re-do the application on the upper bound of the constructor, but this looks murky with what we know right now. Theoretical underpinnings would be good to have here!

compiler/src/dotty/tools/dotc/core/Types.scala
smarter: > it would also turn *-types to a hk-types

Would it? `TypeApplications#isHK` only returns true for `HKTypeLambda`, and before this PR `avoid` used to return the parent for refinement types and it worked fine.

smarter:
The new `avoid` is a bit too strict I think:
```scala
object Test {
  def test = {
    object Hi {
      type A = Int
    }

    val x: Hi.A = 1

    List(x)
  }

  val hi: List[Int] = test // Fails because `test` has type `List[Any]` instead of `List[Int]`
}
```

Because when we avoid `Hi` we return `Object`. This can easily be fixed by modifying `classBound` to return `Object { type A = Int }`:
```diff
diff --git a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
index 4340cae22c..11849b870e 100644
--- a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
@@ -49,7 +49,7 @@ trait TypeAssigner {
         parentType.findMember(decl.name, info.cls.thisType, excluded = Private)
           .suchThat(decl.matches(_))
       val inheritedInfo = inherited.info
-      if (inheritedInfo.exists && decl.info <:< inheritedInfo && !(inheritedInfo <:< decl.info)) {
+      if (!inheritedInfo.exists || (decl.info <:< inheritedInfo && !(inheritedInfo <:< decl.info))) {
         val r = RefinedType(parent, decl.name, decl.info)
         typr.println(i"add ref $parent $decl --> " + r)
         r
```

Do you think that's a good way to go?

compiler/src/dotty/tools/dotc/core/Types.scala
smarter: Alternatively we could return derivedRefinedType(tp, parent, WildcardType) I think

odersky:
@smarter I developed another fix, which is more direct.

smarter:
> For a Block, TypeAssigner does an avoid just for term symbols whereas
Typer does one for all local symbols.

How come we don't avoid type symbols in TypeAssigner anyway?

odersky:
> How come we don't avoid type symbols in TypeAssigner anyway?

I think some of the later phases might allow anonymous classes to be defined in one block be used in another. As long as they are still in the same method, that's OK. But I am not sure about specifics.

smarter:
OK, opened https://github.com/lampepfl/dotty/issues/2987 to track this.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -108,18 +108,26 @@ trait TypeAssigner {
           mapOver(tp)
       }
 
-      /** Two deviations from standard derivedSelect:
-       *   1. The teh approximation result is a singleton references C#x.type, we
+      /** Three deviations from standard derivedSelect:
+       *   1. We first try a widening conversion to the type's info with
+       *      the original prefix. Since the original prefix is known to
+       *      be a subtype of the returned prefix, this can improve results.
+       *   2. IThen, if the approximation result is a singleton reference C#x.type, we
```
smarter: typo: IThen -> Then

liufengyun:
test performance please

PR #2947, opened 2017-08-02
[Do not merge] Aborted Experiment: Native AppliedType 

odersky:
A more developed version of #2898. 

In the end I did not go ahead with this. Here's a summary why:

First, there are other parts of the compiler that rely on the refinement encoding in interesting ways.  In particular, the way we compute findMember relies on the fact that the member type
of `x` in class `C` is a function of the member types of `x` in the parent classes of `C`. This is no longer true when we instantiate arguments immediately. To pick an example form collections: The type of `toCollection` has the form

     toCollection: (x: Repr)T

where `Repr` is a parameter to all collections. If we see that type from a specific 
collection `sc` we get `(x: sc.Repr)T`. The inherited types of `toCollection` are 
all of the form (x: P.this.Repr)T` where `P` is a parent class. The specific collection
type is obtained by joining the patent types with `&` after replacing `P.this` with `sc.type`.

But if we expand parameters with arguments we get `(x: P)T` for the parent types instead. instead.  And there is no safe operation which will allow us to recover `(x: SC)T` from that.

Second, there's the problem of bounds propagation (see `bounds-propagation.scala` test case).

Those two problems can be solved by having a mixed scheme where we keep type parameters as members, but make them private and have special types that refer to them instead. But that bring back part of the complexity of the refinement type scheme. 

The decisive reason to drop is elsewhere: The Linker relies on being able to refine any parameter bounds in a class or all its base classes retro-actively through refinements. If parameters are no longer instantiated through refinements, this will not work anymore. So if we want to keep the 
option of a fast and precise control-flow analysis, we need to keep parameters as refinements.

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -889,7 +889,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
           // to inner symbols of DefDef
           // todo: somehow handle.
 
-    def parents: List[Type] = tp.parents
+    def parents: List[Type] = tp.parentsNEW
   }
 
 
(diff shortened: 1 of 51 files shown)


odersky:
Closed for the reasons given in the PR.

PR #2952, opened 2017-08-03
Fix #2636: Move last phase period assertion

### LINKED ISSUE #2636: Assert in Run.scala is inefficient

nicolasstucki:


compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -108,9 +108,14 @@ class Compiler {
            new CollectSuperCalls,   // Find classes that are called with super
            new DropInlined,         // Drop Inlined nodes, since backend has no use for them
            new LabelDefs),          // Converts calls to labels to jumps
-      List(new GenBCode)            // Generate JVM bytecode
+      List(checkedIdOverflow(new GenBCode)) // Generate JVM bytecode
     )
 
+  private def checkedIdOverflow(phase: Phase): Phase = {
```
DarkDimius: Nice!

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -108,9 +108,14 @@ class Compiler {
...
+    assert(phase.id <= Periods.MaxPossiblePhaseId)
```
DarkDimius: would this ever fail? Phase.id extracts it from the  period inside the phase, i.e. it reads a field which you're supposedly testing for overflow.

DarkDimius: I propose to do the test in Phase.init instead.

DarkDimius: https://github.com/lampepfl/dotty/pull/2953

nicolasstucki: I don't know if it will ever fail, but that was the check that was there before. Probably at some point this was failing.

nicolasstucki:
@DarkDimius merged contents of both PRs

PR #2953, opened 2017-08-03
Phases: test for too many phases

DarkDimius:


nicolasstucki:
Shouldn't we also remove the previous assertion with this version?

nicolasstucki:
Change added to #2952

PR #2956, opened 2017-08-03
Fix #2943: Insert LazyRefs automatically upon Unpickling

### LINKED ISSUE #2943: Avoid LazyRefs in TASTY
#2914 introduced LazyRefs in the pickle format. We should try to avoid this, since it makes
it very hard to produce legal Tasty trees from other sources than the Dotty compiler. It would be 
better if the Unpickler could insert the LazyRefs automatically where needed.

odersky:
This relieves Tasty producers from having to insert LazyRefs
themselves. Fixes #2943.


DarkDimius:
Very glad to see this change! 👍

PR #2958, opened 2017-08-04
Fix #2944: propagate information from GADT bounds to normal info.

### LINKED ISSUE #2944: Issue with GADTs tasty types
```scala
trait Map2[K] {
  def get(k: K) = null
  def foo = {
    this match {
      case that: Map2[b] => that.get(3.asInstanceOf[b])
      case that: Map2[c] => that.get(4.asInstanceOf[K])
      case _ => get(5.asInstanceOf[K])
    }
  }
}
```

gets wrong types for binding in `3.asInstanceOf[b]` and `4.asInstanceOf[K]` involving `b` and `c`.

The full tasty pretty printed is [here](https://gist.github.com/nicolasstucki/d6918ca6b6b0a08df949d2271250f4d3).

DarkDimius:
This allows all other parts of pipeline to be less aware of existence
of the GADT bounds.

In particular, it makes us pickle them as normal type bounds.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -45,6 +45,8 @@ import reporting.diagnostic.messages.SuperCallsNotAllowedInline
  *  (11) Minimizes `call` fields of `Inline` nodes to just point to the toplevel
  *       class from which code was inlined.
  *
+ *  (12) Converts GADT bounds into normal type bounds
+ *
  *  The reason for making this a macro transform is that some functions (in particular
  *  super and protected accessors and instantiation checks) are naturally top-down and
  *  don't lend themselves to the bottom-up approach of a mini phase. The other two functions
@@ -52,9 +54,16 @@ import reporting.diagnostic.messages.SuperCallsNotAllowedInline
  *  mini-phase or subfunction of a macro phase equally well. But taken by themselves
  *  they do not warrant their own group of miniphases before pickling.
  */
-class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTransformer =>
+class PostTyper extends MacroTransform with SymTransformer  { thisTransformer =>
+
 
   import tpd._
+
+  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
+    if (sym.is(BindDefinedType) && ctx.gadt.bounds.contains(sym.symbol)) {
+      sym.copySymDenotation(info = ctx.gadt.bounds.apply(sym.symbol) & sym.info)
+    } else sym
+  }
 
   /** the following two members override abstract members in Transform */
   override def phaseName: String = "posttyper"
--- a/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -59,10 +59,10 @@ class PostTyper extends MacroTransform with SymTransformer  { thisTransformer =>
 
   import tpd._
 
-  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
-    if (sym.is(BindDefinedType) && ctx.gadt.bounds.contains(sym.symbol)) {
-      sym.copySymDenotation(info = ctx.gadt.bounds.apply(sym.symbol) & sym.info)
-    } else sym
+  def transformSym(ref: SymDenotation)(implicit ctx: Context): SymDenotation = {
+    if (ref.is(BindDefinedType) && ctx.gadt.bounds.contains(ref.symbol)) {
+      ref.copySymDenotation(info = ctx.gadt.bounds.apply(ref.symbol) & ref.info)
+    } else ref
   }
 
   /** the following two members override abstract members in Transform */
--- a/tests/pos/i2944.scala
+++ b/tests/pos/i2944.scala
@@ -3,6 +3,7 @@ trait Map2[K] {
   def foo: K = {
     this match {
       case that: Map2[b] => that.get(3.asInstanceOf[b])
+      //case that: Map2[c] => that.get(4.asInstanceOf[K])
       case _ => get(5.asInstanceOf[K])
     }
   }
--- a/tests/pos/i2944.scala
+++ b/tests/pos/i2944.scala
@@ -1,0 +1,10 @@
+trait Map2[K] {
+  def get(k: K): K = k
+  def foo: K = {
+    this match {
+      case that: Map2[b] => that.get(3.asInstanceOf[b])
+      //case that: Map2[c] => that.get(4.asInstanceOf[K])
+      case _ => get(5.asInstanceOf[K])
+    }
+  }
+}
--- a/tests/pos/i2944.scala
+++ b/tests/pos/i2944.scala
@@ -3,7 +3,6 @@ trait Map2[K] {
   def foo: K = {
     this match {
       case that: Map2[b] => that.get(3.asInstanceOf[b])
-      //case that: Map2[c] => that.get(4.asInstanceOf[K])
       case _ => get(5.asInstanceOf[K])
     }
   }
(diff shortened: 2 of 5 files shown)


DarkDimius:
All tests passed!
@nicolasstucki, could you please check if it fixes the original test-case?
@odersky please review the code.

DarkDimius:
@nicolasstucki, i've added a test file for #2944 but it just tests compilation, which produced wrong tasty before this fix.

tests/pos/i2944.scala
```diff
@@ -0,0 +1,10 @@
+trait Map2[K] {
+  def get(k: K): K = k
+  def foo: K = {
+    this match {
+      case that: Map2[b] => that.get(3.asInstanceOf[b])
```
nicolasstucki: This test case passes my the check with my tests 👍

tests/pos/i2944.scala
```diff
@@ -0,0 +1,10 @@
...
+      //case that: Map2[c] => that.get(4.asInstanceOf[K])
```
nicolasstucki: This one still crashes while trying to type `that.get(3.asInstanceOf[K].asInstanceOf[b @ b])`.

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -45,17 +45,26 @@ import reporting.diagnostic.messages.SuperCallsNotAllowedInline
...
  */
-class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTransformer =>
+class PostTyper extends MacroTransform with SymTransformer  { thisTransformer =>
+
 
   import tpd._
 
+  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
+    if (sym.is(BindDefinedType) && ctx.gadt.bounds.contains(sym.symbol)) {
+      sym.copySymDenotation(info = ctx.gadt.bounds.apply(sym.symbol) & sym.info)
```
smarter: you could probably remove `sym.symbol` from the `ctx.gadt.bounds` map at that point too.

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -45,17 +45,26 @@ import reporting.diagnostic.messages.SuperCallsNotAllowedInline
...
  *  mini-phase or subfunction of a macro phase equally well. But taken by themselves
  *  they do not warrant their own group of miniphases before pickling.
  */
-class PostTyper extends MacroTransform with IdentityDenotTransformer  { thisTransformer =>
+class PostTyper extends MacroTransform with SymTransformer  { thisTransformer =>
+
 
   import tpd._
 
+  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
```
smarter: Nitpick: `sym` is usually used for symbols, for a denotation I would use `ref` or `d`.

DarkDimius:
You can't. You'll need to remove it from top context

DarkDimius:
Name sym is used consistently in all SymTransforms. See original trait.

smarter:
> Name sym is used consistently in all SymTransforms. See original trait.
```shell
% git grep "def transformSym" compiler
compiler/src/dotty/tools/dotc/core/DenotTransformers.scala:    def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation
compiler/src/dotty/tools/dotc/transform/Flatten.scala:  def transformSym(ref: SymDenotation)(implicit ctx: Context) = {
compiler/src/dotty/tools/dotc/transform/Getters.scala:  override def transformSym(d: SymDenotation)(implicit ctx: Context): SymDenotation = {
compiler/src/dotty/tools/dotc/transform/Mixin.scala:  override def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation =
compiler/src/dotty/tools/dotc/transform/MoveStatics.scala:  def transformSym(sym: SymDenotation)(implicit ctx: Context): SymDenotation = {
compiler/src/dotty/tools/dotc/transform/NormalizeFlags.scala:  def transformSym(ref: SymDenotation)(implicit ctx: Context) = {
compiler/src/dotty/tools/dotc/transform/PrivateToStatic.scala.disabled:  override def transformSym(sd: SymDenotation)(implicit ctx: Context): SymDenotation =
compiler/src/dotty/tools/dotc/transform/TreeChecker.scala:  def transformSym(symd: SymDenotation)(implicit ctx: Context): SymDenotation = {
```

I think everyone is confused about what name to use :).

DarkDimius:
> This one still crashes while trying to type that.get(4.asInstanceOf[K].asInstanceOf[b @ b]).

Nice) will look into it! Curious why the bind node is inside of cast...

smarter:
> You can't. You'll need to remove it from top context

Why? When we create a fresh context we keep a reference to the same `GADTMap` in `_gadt`.

DarkDimius:
>     This one still crashes while trying to type that.get(4.asInstanceOf[K].asInstanceOf[b @ b]).

@nicolasstucki , though I authored this testcase, I forgot how to make this bind appear.
The tasty looks correct. The error happens on reading. How did you force reading it?

DarkDimius:
Ok, found a bug by reading code. TreeUnpickeler.readTpt is obviously wrong.

nicolasstucki:
@DarkDimius you can reproduce it easily for this branch https://github.com/dotty-staging/dotty/tree/link-from-tasty-show-bug. Just run the `reproducefailure.sh` script.

nicolasstucki:
That branch contains the current fix.

DarkDimius:
@nicolasstucki, could you please validate that the last commit fixes the issue?

nicolasstucki:
It did not fix it, but that is clearly the source of the issue. Now I get:
```
exception while typing c of class class dotty.tools.dotc.ast.Trees$Ident # 401946
exception while typing 4.asInstanceOf[K].asInstanceOf[c] of class class dotty.tools.dotc.ast.Trees$TypeApply # 401947
exception while typing that.get(4.asInstanceOf[K].asInstanceOf[c]) of class class dotty.tools.dotc.ast.Trees$Apply # 402179
```

Note that you did replace the binding by the `Ident` but it still can't pas Ycheck.

DarkDimius:
@nicolasstucki, sorry. I did a stupid mistake too :-) Fix incomming.

DarkDimius:
Oh... it is actually much worse: on unpickling, c becomes _a term_.

DarkDimius:
aaaand we create symbol for it twice on unpickling. And those two symbols are obviously incompatible between each other.

DarkDimius:
I propose we postpone fixing the remaining bug till next dotty meeting. My proposed fix is to revert 79e22287def4b5bf04e7ecd02507b9e72a4134f5 and ad6b5782e99aa946b7426fe3ceebfb2e429cc660. Those commits made bind be handled as-if it was  TypeTree in Tasty. This broke a lot of stuff because unpickler assumes that it can read TypeTrees repeatedly, which is not true for bind, which is a DefTree. We assume in multiple locations that DefTree should be only read once, as you'll get a fresh symbol every time you read it.

DarkDimius:
@nicolasstucki, is the remaining bug also a blocker?

nicolasstucki:
@DarkDimius the fix in the first commit unblocked it. The second could be left for later, lets open a separate issue for it. Thanks.

nicolasstucki:
Unfortunately the the first fix did not work in the original code from `strawman.collection.Map.equals`.

I minimised the issue again to 
```scala
trait Map2[K] {
  def get(k: K): K = k
  def foo = {
    val o = new Object 
    o match { // now match on Object and not this
      case that: Map2[b] => that.get(3.asInstanceOf[b])
      case _ => get(5.asInstanceOf[K])
    }
  }
}
```

DarkDimius:
@nicolasstucki and I have figured that this is another manifestation of https://github.com/lampepfl/dotty/pull/2958#issuecomment-320276098

nicolasstucki:
Rebased.

@smarter could you re-review this quickly before merging.

tests/pos/i2944.scala
smarter: Do we have a real test for this now?

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
smarter: I still think this is possible but  ¯\\\_(ツ)_/¯

liufengyun:
test performance please

This PR fails the compilation of https://github.com/liufengyun/scalaPB on the bench server.

PR #2970, opened 2017-08-10
Fix #2948: Use symbol's info when mapping inherited denotations

### LINKED ISSUE #2948: Missing implicit argument in ListBuffer.unzip
For some reason the `implicit asPair: A => (A1, A2)` parameter of the unzip is not found or not inserted. As a consequence `unzip` is treated as a partially applied function. Note that also happens for `ArrayBuffer` but not for `List` and `Array`.

```scala
class Foo {
  val zipped: ListBuffer[(String, Int)] = null
  val unzipped: (ListBuffer[String], ListBuffer[Int]) = zipped.unzip
}
```

```scala
5 |  val unzipped: (ListBuffer[String], ListBuffer[Int]) = zipped.unzip
  |                                                                    ^
  |found:    ((String, Int)) => (Any, Any) => (scala.collection.Traversable[Any], 
  |  scala.collection.Traversable[Any]
  |)
  |required: (collection.mutable.ListBuffer[String], collection.mutable.ListBuffer[Int])
  |
```

odersky:
The test case shows that it is inadmissible to combine the infos of inherited denotations
into a new denotation. The problem here is that a type parameter CC was instantiated in
an inherited denotation to Traversable, yet the parameter was afterwards instantiated to
ListBuffer. This shows that infos from inherited denotations are useless; instead we have
to go back to the inherited symbol's infos and map them with an asSeenFrom.

This fix also shows that one of the reasons for abandoning #2947 was wrong. We cannot form
denotations from parent denotations in any case, so instantiating early should be fine with
the change in this commit.

Based on #2945

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
+++ b/compiler/src/dotty/tools/dotc/core/ConstraintHandling.scala
@@ -362,13 +362,11 @@ trait ConstraintHandling {
       def pruneLambdaParams(tp: Type) =
         if (comparedTypeLambdas.nonEmpty) {
           val approx = new ApproximatingTypeMap {
+            if (fromBelow) variance = -1
             def apply(t: Type): Type = t match {
               case t @ TypeParamRef(tl: TypeLambda, n) if comparedTypeLambdas contains tl =>
-                val effectiveVariance = if (fromBelow) -variance else variance
                 val bounds = tl.paramInfos(n)
-                if (effectiveVariance > 0) bounds.lo
-                else if (effectiveVariance < 0) bounds.hi
-                else NoType
+                range(bounds.lo, bounds.hi)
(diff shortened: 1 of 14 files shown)


odersky:
@felixmulder CI seems to hang here. How can we unblock it?

smarter:
@odersky I restarted drone and the tests for your PR.

felixmulder:
Looks like it's only running one job concurrently. Will try to fix it now.

felixmulder:
Alright, should be unblocked now. Guillaume and I both jumped on this at the same time, so we ended up killing and starting docker images left and right :joy:

odersky:
Only last two commits need to be reviewed.

nicolasstucki:
Last two commits LGTM

PR #2975, opened 2017-08-14
Fix #2971: Soudness issue with variance and higher kinded types

### LINKED ISSUE #2971: Soundness issue with variance and higher kinded types
While playing around with dotty I encountered what seems to be a issue with the typechecker. Here's a minimal example where I was able to reproduce the issue:

```scala
case class Foo[+X[_]](will: X[Int]) {
  def foo[Y[_]](right: Foo[Y]) = Foo.doFoo(this, right)
}

class A[X] { def crash = true }
class B[X]

object Foo {
  def doFoo[X[_]](left: Foo[X], right: Foo[X]): Foo[X] = right

    val fooA = Foo(new A[Int])
    val fooB = Foo(new B[Int])
    // The type for this is inferred correctly to Foo[A|B]
    val fine = doFoo(fooA, fooB)
    // This throws a ClassCastException because fooB isn't a Foo[A]
    val oops: Foo[A] = fooA.foo(fooB)
    println(oops.will.crash)
  }
}
```

It seems like the type that dotty is inferring for `foo` is `Foo[X]`, even though it should be `Foo[X|Y]`, and because of this is allowing me to then use the result in ways I shouldn't be able to.

Also, I couldn't find a way to annotate foo to get it to typecheck and work. One variation I tried was this:

```scala
case class Foo[+X[_]](will: X[Int]) {
  def foo[Y[_]](right: Foo[Y]): Foo[X|Y] = Foo.foo(this, right)
}

object Foo {
  def foo[X[_],Y[_]](left: Foo[X], right: Foo[Y]): Foo[X|Y] = right
}
```

But I was met with

```
-- [E007] Type Mismatch Error: /tmp/scastie6848463800932418099/src/main/scala/main.scala:6:62 
6 |  def foo[X[_],Y[_]](left: Foo[X], right: Foo[Y]): Foo[X|Y] = right
  |                                                              ^^^^^
  |                                                   found:    Foo[Y](right)
  |                                                   required: Foo[X]
  |
```

I'd expect the required type should be `Foo[X|Y]` (or to be more precise `Foo[[T] => X[T] | Y[T]]`) , for which a `Foo[Y]` should be fine right? Or is there a problem using subtyping with higher kinded types like this?

Either way, the first example definitely seems like an issue of soundness.

smarter:
We shouldn't use `HKApply#superType` in `isRef`, because this means
taking the upper bound of abstract type constructors.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -413,6 +413,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
               // of an `AndType` can lead to a cascade of subtyping checks
               // This twist is needed to make collection/generic/ParFactory.scala compile
               fourthTry(tp1, tp2) || compareRefinedSlow
+            case tp1: HKTypeLambda =>
+              // HKTypeLambdas do not have members.
+              fourthTry(tp1, tp2)
             case _ =>
               compareRefinedSlow || fourthTry(tp1, tp2)
           }
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -132,7 +132,10 @@ object Types {
           case _ =>  this1.symbol eq sym
         }
       case this1: RefinedOrRecType => this1.parent.isRef(sym)
-      case this1: HKApply => this1.superType.isRef(sym)
+      case this1: HKApply =>
+        val this2 = this1.dealias
+        if (this2 ne this1) this2.isRef(sym)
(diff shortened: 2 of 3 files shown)


smarter:
There's one test failure which is extremely annoying because it only happens when compiling all of the neg tests, even though the failure is in `tests/neg/i1650.scala`:
```shell
> vulpix tests/neg
# fails (but not everytime)
> vulpix tests/neg/i1650.scala
# succeeds
```
Here's the stacktrace extracted from the log:
https://gist.github.com/smarter/79becb6da8f91254e5b92de42e24cff3

Any idea what's going on?

smarter:
Pushed a fix, though it's a bit kludgy to add special cases like that.

PR #2978, opened 2017-08-15
Avoid insertion of spurious applies when inlining

odersky:
The test case crashed before, because it encountered
a selection with an inserted apply, which should be typed
but was not. The fix prevents inserting the apply.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inliner.scala
@@ -16,6 +16,7 @@ import Contexts.Context
 import Names.{Name, TermName, EmptyTermName}
 import NameOps._
 import NameKinds.InlineAccessorName
+import ProtoTypes.selectionProto
 import SymDenotations.SymDenotation
 import Annotations._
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -524,7 +525,9 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
     }
 
     override def typedSelect(tree: untpd.Select, pt: Type)(implicit ctx: Context): Tree = {
-      val res = super.typedSelect(tree, pt)
+      assert(tree.hasType, tree)
+      val qual1 = typed(tree.qualifier, selectionProto(tree.name, pt, this))
+      val res = untpd.cpy.Select(tree)(qual1, tree.name).withType(tree.typeOpt)
```
liufengyun: I'm wondering why use `AnySelectionProto` instead of using `selectionProto(tree.name, pt, this)` as expected type in `Retyper.typedSelect`?

PR #2983, opened 2017-08-15
Fix #2250: add green test

### LINKED ISSUE #2250: Implicit conversion of `Array[_]` failed
The code compiles in Scalac, but fails in Dotty. Both candidates `Predef.genericWrapArray` and `Predef.genericArrayOps` are found, but disqualified due to type inference error.

```Scala
class T {
  def foo(any: AnyRef): Unit = {
    any.asInstanceOf[Array[_]].iterator
  }
}
```

liufengyun:
Fix #2250: add green test, it's probably fixed in #2739.


nicolasstucki:
LGTM

PR #2986, opened 2017-08-16
Harden IDE: Cleanup trees after typechecking them in IDE

odersky:
I decided in the end it is simpler and more modular to sanitize trees in the IDE, after type checking them. That way we can separate concerns and need not burden the general tree handling.

Sanitization involves removing all attachments and symbol completers.

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala
@@ -179,6 +179,14 @@ class InteractiveDriver(settings: List[String]) extends Driver {
 
   private val compiler: Compiler = new InteractiveCompiler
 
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala
```diff
@@ -179,6 +179,16 @@ class InteractiveDriver(settings: List[String]) extends Driver {
 
   private val compiler: Compiler = new InteractiveCompiler
 
+  private def cleanup(tree: tpd.Tree)(implicit ctx: Context): Unit = tree.foreachSubTree { t =>
```
smarter: Add a documentation comment explaining why cleanup is necessary?

PR #2988, opened 2017-08-16
Fix #1463: no type cast in singelton equality test

### LINKED ISSUE #1463: Bad code for pattern matches with values.
Consider the following test case adapted from pending/run/t3150:

```
object Test {

  class Bob2 {
    override def equals(other: Any) = true
  }
  val Bob2 = new Bob2

  def f0(x: Any) = x match { case Bob2 => Bob2 } // class cast exception at runtime, dotc only
  def f1(x: Any) = x match { case Bob => Bob } // class cast exception at runtime, dotc only
  def f2(x: Any): Bob.type = x match { case x @ Bob => x } // class cast exception at runtime, dotc and javac.

    assert(f0(Bob2) eq Bob2)
    assert(f0(0) eq Bob2)  // only dotty fails here
    assert(f0(Nil) eq Bob2)

    assert(f1(Bob) eq Bob)
    assert(f1(0) eq Bob)  // only dotty fails here
    assert(f1(Nil) eq Bob)

    assert(f2(Bob) eq Bob)
    assert(f2(0) eq Bob) // both dotty and scalac fail here
    assert(f2(Nil) eq Bob)
  }
}
```

Dotty causes a class cast exception in three scenarios, scalac only in the last. Each class cast exception is a bug. We should decide what the precise typing and comparison rules for pattern matches involving equals tests are.

liufengyun:
Fix #1463: no type cast in singelton equality test

### DIFF
--- a/tests/run/i1463.scala
+++ b/tests/run/i1463.scala
@@ -1,0 +1,26 @@
+object Test {
+  case object Bob { override def equals(other: Any) = true }
+
+  class Bob2 {
+    override def equals(other: Any) = true
+  }
+  val Bob2 = new Bob2
+
+  def f0(x: Any) = x match { case Bob2 => Bob2 } // class cast exception at runtime, dotc only
+  def f1(x: Any) = x match { case Bob => Bob } // class cast exception at runtime, dotc only
+  def f2(x: Any): Bob.type = x match { case x @ Bob => x } // class cast exception at runtime, dotc and javac.
+
+  def main(args: Array[String]): Unit = {
+    assert(f0(Bob2) eq Bob2)
+    assert(f0(0) eq Bob2)  // only dotty fails here
+    assert(f0(Nil) eq Bob2)
+
+    assert(f1(Bob) eq Bob)
+    assert(f1(0) eq Bob)  // only dotty fails here
+    assert(f1(Nil) eq Bob)
+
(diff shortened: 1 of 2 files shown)


tests/run/i1463.scala
```diff
@@ -0,0 +1,26 @@
...
+  case object Bob { override def equals(other: Any) = true }
+
+  class Bob2 {
+    override def equals(other: Any) = true
+  }
+  val Bob2 = new Bob2
+
+  def f0(x: Any) = x match { case Bob2 => Bob2 } // class cast exception at runtime, dotc only
+  def f1(x: Any) = x match { case Bob => Bob } // class cast exception at runtime, dotc only
+  def f2(x: Any): Bob.type = x match { case x @ Bob => x } // class cast exception at runtime, dotc and javac.
```
smarter: This should not typecheck because `x` in the right hand side should have type `Any`, not `Bob.type`, as explained in my comment: https://github.com/lampepfl/dotty/issues/1463#issuecomment-241671824
This matches the behavior of scalac with `-Xfuture`.

liufengyun: Thanks @smarter , I see both approaches may cause some confusion to programmers. We need a spec here, are  you happy with what's in `-Xfuture` in 2.11 and 2.12, @odersky ?

PR #2993, opened 2017-08-18
Put idempotency tests in their own category

nicolasstucki:
The vulpix command will not run them anymore.


compiler/test/dotty/tools/dotc/IdempotencyTests.scala
```diff
@@ -0,0 +1,56 @@
+package dotty
+package tools
+package dotc
+
+import org.junit.{ Test, BeforeClass, AfterClass }
+
+import java.nio.file._
```
biboudis: I think you don't need this and the next three lines.

PR #2999, opened 2017-08-21
Fix #2964: Refresh names for anonymous classes

### LINKED ISSUE #2964: Anonymous classes indices should not be global
Currently fresh names for anonymous classes follow this pattern 
```scala
class Foo {
  new Object {} // will be named Foo$$anon$1
}
class Bar {
  new Object {} // will be named Bar$$anon$2
  new Object {} // will be named Bar$$anon$3
}
class Baz {
  new Object {} // will be named Baz$$anon$4
}
```
which will depend on the compilation order.

We need to change it to give fresh names per owner class like:

```scala
class Foo {
  new Object {} // will be named Foo$$anon$1
}
class Bar {
  new Object {} // will be named Bar$$anon$1
  new Object {} // will be named Bar$$anon$2
}
class Baz {
  new Object {} // will be named Baz$$anon$1
}
```

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/LiftedClasses.scala
```diff
@@ -0,0 +1,48 @@
...
+import dotty.tools.dotc.core.SymDenotations.SymDenotation
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/** Renames lifted classes to local numbering scheme */
+class LiftedClasses extends MiniPhaseTransform with SymTransformer { thisTransformer =>
+
+  override def phaseName = "liftedClasses"
+
+  override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[Flatten])
```
DarkDimius: It's not really flatten, but RestoreScopes.

compiler/src/dotty/tools/dotc/transform/TransformWildcards.scala
```diff
@@ -16,6 +17,8 @@ class TransformWildcards extends MiniPhaseTransform with IdentityDenotTransforme
 
   override def phaseName = "transformWildcards"
 
+  override def runsAfter: Set[Class[_ <: Phases.Phase]] = Set(classOf[LiftedClasses])
```
DarkDimius: Why is there a dependency between those two phases?

compiler/src/dotty/tools/dotc/transform/LiftedClasses.scala
```diff
@@ -0,0 +1,48 @@
...
+  def transformSym(ref: SymDenotation)(implicit ctx: Context): SymDenotation =
+    if (needsRefresh(ref.symbol)) ref.copySymDenotation(name = refreshedName(ref.symbol))
+    else ref
+
+  override def transformTypeDef(tree: tpd.TypeDef)(implicit ctx: Context, info: TransformerInfo): tpd.Tree =
+    if (needsRefresh(tree.symbol)) cpy.TypeDef(tree)(name = tree.symbol.name.asTypeName)
+    else tree
+
+  private def needsRefresh(sym: Symbol)(implicit ctx: Context): Boolean =
+    sym.isClass && !sym.is(Package) && sym.name.is(UniqueName)
```
DarkDimius: Do we have packages with UniqueName?

compiler/src/dotty/tools/dotc/transform/LiftedClasses.scala
```diff
@@ -0,0 +1,48 @@
...
+import dotty.tools.dotc.core.Phases
+import dotty.tools.dotc.core.StdNames._
+import dotty.tools.dotc.core.SymDenotations.SymDenotation
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
+
+/** Renames lifted classes to local numbering scheme */
+class LiftedClasses extends MiniPhaseTransform with SymTransformer { thisTransformer =>
+
+  override def phaseName = "liftedClasses"
```
DarkDimius: renameLifted?

compiler/src/dotty/tools/dotc/transform/LiftedClasses.scala
DarkDimius: Why do you apply it only to classes, don't we have the same issue with methods that are lifted out?

DarkDimius:
```scala
class Foo {
  def ann = {
    def foo() = 1;
  }
}
class Bar {
  def ann = {
    def foo() = 1;
  }
}
```

```java
public class Foo {
    public void ann() {
    }

    private static int foo$1() {
        return 1;
    }
}

public class Bar {
    public void ann() {
    }

    private static int foo$2() { // <---  same issue
        return 1;
    }
}

File: compiler/src/dotty/tools/dotc/transform/TransformWildcards.scala (hunk 2 above)
Comment (nicolasstucki) [inline compiler/src/dotty/tools/dotc/transform/TransformWildcards.scala]: Only checked experimentally. It failed if I put it afterward, therefore I added the explicit `runsAfter` to make this evident to anyone that tries to reorder phases.

**Comment by nicolasstucki** [comment]
@DarkDimius the requested changes have been made.

File: compiler/src/dotty/tools/dotc/transform/RenameLifted.scala [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+  /** If the name of the symbol with a unique name needs to be refreshed
+   *    - if it is a lifted class
+   *    - if it is a lifted method
+   */
+  private def needsRefresh(sym: Symbol)(implicit ctx: Context): Boolean =
+    (sym.isClass || sym.is(Private | Method | JavaStatic)) && sym.name.is(UniqueName)
+
+  /** Refreshes the number of the name based on the full name of the symbol */
+  private def refreshedName(sym: Symbol)(implicit ctx: Context): Name = {
+    lazy val rewriteUnique: PartialFunction[Name, Name] = {
```
OlivierBlanvillain: Why lazy?

OlivierBlanvillain: Oh I see, that's a funny way to write recursion :smile:

nicolasstucki: Do you know a cleaner way to define recursive partial functions?

OlivierBlanvillain: I would do it with a `def`

nicolasstucki: Changed it. Though now it will create a new instance of the function each iteration, I guess this should not happen too often.

PR #3003, opened 2017-08-21
Print Lists consistently in exhaustivity warnings

AleksanderBG:
First off - I wasn't too sure if I should open an issue for this, since it's so trivial. I apologise if I should.

This PR adjusts how `List`s are printed in error messages. Previously, depending on the specific pattern match in question, an unmatched `::` would print differently depending on whether we ended up with a `Typ` space or `Kon` space. For the simplest example, `Typ(::)` was printed as `List(_)`, but `Kon(::, Int, List[Int])` was printed as `List(_, _*)`. The error messages were also slightly confusing, since to me `List(_)` suggests that the list has only one member, whereas a `::` can have one-or-more members. After adjustments, `List` now prints consistently as though we had ended up with the `Kon` space. Note that the code is equivalent to showing `Kon(tp, signature(tp).map(Typ(_, false)))`.

/cc @liufengyun

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -661,7 +661,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         else if (scalaListType.isRef(sym))
           if (mergeList) "_*" else "_: List"
         else if (scalaConsType.isRef(sym))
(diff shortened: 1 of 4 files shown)


smarter:
> First off - I wasn't too sure if I should open an issue for this, since it's so trivial. I apologise if I should.

It's fine to open PRs without corresponding issues :).

liufengyun:
Thanks @AleksanderBG , LGTM!

PR #3011, opened 2017-08-23
Fix #3004 and simplify exhaustivity check

### LINKED ISSUE #3004: Custom unapply methods can produce misleading error messages
As a simple example, the following code:
```scala
object O {
         sealed trait Fruit
         object Apple extends Fruit
         object Banana extends Fruit
         sealed class C(f1: Fruit, f2: Fruit)
         object C {
           def unapply(c: C): Some[Banana.type] = Some(Banana)
         }
         def m(c: C) = c match { case C(b) => b }
}
```
produces the following message:
```
-- [E028] Pattern Match Exhaustivity Warning: <console>:12:16 ------------------
12 |  def m(c: C) = c match { case C(b) => b }
   |                ^
   |                match may not be exhaustive.
   |
   |                It would fail on: C(Apple, _)
```
The pattern match is, in fact, exhaustive and the message is somewhat puzzling.

The underlying issue is the interaction between this code in the `project` method:
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala#L430-L431
and the places in rest of the code which "project" `Typ(tp)` as `Kon(tp, signature(tp).map(Typ(_))`:
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala#L267
When a `Kon` space resulting from projecting `C.unapply` is subtracted from `Typ(C)`, "inconsistent" `Kon` spaces are created, which then in turn causes this issue.

/cc @liufengyun

liufengyun:
Fix #3004 and simplify exhaustivity check


liufengyun:
Review @AleksanderBG ?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -641,6 +628,15 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
   /** Display spaces */
   def show(s: Space): String = {
+    def params(tp: Type): List[Type] = {
```
AleksanderBG: This function seems to be mostly doing unnecessary work - the actual types returned are never used, and the only thing that matters is the length of the returned list.

AleksanderBG:
@liufengyun - I've tried to review what I can, and apart from that single nitpick everything looks good to me. I'm a bit out of my depth when it comes to verifying that the new implementation of `signature` is correct, though. I must say that unifying `Kon` and `Fun` spaces seems like a very good approach - the overall algorithm now looks much cleaner.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
AleksanderBG: To clarify, I've checked that the length of `.firstParamTypes` does not actually change - all tests still pass if this function returns `ktor.fistParamTypes`.

tests/patmat/t9232.check
```diff
@@ -1 +1,3 @@
-13: Pattern Match Exhaustivity: Node2(), Node1(Foo(_))
+13: Pattern Match Exhaustivity: Node2()
+17: Pattern Match Exhaustivity: Node2(), Node1(Foo(Nil)), Node1(Foo(List(_, _*)))
```
smarter: Should `List(_, _*)` be `List(_, _: _*)` ?

liufengyun:
Thanks @AleksanderBG . The `signature` is stolen from the new patternMatcher, which follows the specification closely.

smarter:
> The signature is stolen from the new patternMatcher, which follows the specification closely.

Would be nice to add a comment pointing to the corresponding code in the pattern matcher. Even nicer if we could share some code but I don't know if that's possible here :).

liufengyun:
test performance please

PR #3019, opened 2017-08-25
Add compilation order idempotency tests

nicolasstucki:


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -324,4 +317,11 @@ class CompilationTests extends ParallelTesting {
 object CompilationTests {
   implicit val summaryReport: SummaryReporting = new SummaryReport
   @AfterClass def cleanup(): Unit = summaryReport.echoSummary()
+
+  def sources(paths: JStream[Path], excludedFiles: List[String] = Nil): List[String] =
```
biboudis: This definition is the same with the one under `tastyBootstrap` and both are non-capturing. Does it make sense to merge?

nicolasstucki: Yes, I will use the same function. Thanks.

nicolasstucki: Done

PR #3020, opened 2017-08-25
Properly escape wrapped names in REPL

felixmulder:


felixmulder:
Merging as agreed with @smarter :)

PR #3021, opened 2017-08-25
Fix #3018: use List instead of Map in InlineLocalObjects

### LINKED ISSUE #3018: Incorrect local optimisation cause run-time exceptions
Compile the following code with (do `sbt dist/pack` first if it doesn't exist):

    dist/target/pack/bin/dotc -optimise space.scala

```Scala
class Type
class Symbol

sealed trait Space
case object Empty extends Space
case class Typ(tp: Type, decomposed: Boolean) extends Space
case class Prod(tp: Type, unappTp: Type, unappSym: Symbol, params: List[Space], full: Boolean) extends Space
case class Or(spaces: List[Space]) extends Space

object Test {
  def simplify(space: Space, aggressive: Boolean = false): Space = space match {
    case Prod(tp, fun, sym, spaces, full) =>
      val sp = Prod(tp, fun, sym, spaces.map(simplify(_)), full)
      if (sp.params.contains(Empty)) Empty
      else sp
    case Or(spaces) => space
    case Typ(tp, _) => space
    case _ => space
  }

    simplify(Prod(new Type, new Type, new Symbol, Nil, false))
  }
}
```

However, running it gives the code `dist/target/pack/bin/dotr Test` generates a run-time exception:

```
Exception in thread "main" java.lang.ClassCastException: scala.collection.immutable.Nil$ cannot be cast to Type
	at Test$.simplify(Space3.scala:13)
	at Test$.main(Space3.scala:22)
	at Test.main(Space3.scala)
```

This issue is reproduced from #3011.

OlivierBlanvillain:
Using Map#values resulted in losing the original order or arguments, a List fixes that. The lookup cost will be slightly higher this way, but that should be negligible.


compiler/src/dotty/tools/dotc/transform/localopt/InlineLocalObjects.scala
```diff
@@ -99,10 +99,9 @@ class InlineLocalObjects(val simplifyPhase: Simplify) extends Optimisation {
         Thicket(newFieldsDefs :+ recreate)
 
       case t @ Select(rec, _) if isImmutableAccessor(t) =>
-        newFieldsMapping.getOrElse(rec.symbol, Map.empty).get(t.symbol) match {
-          case None         => t
-          case Some(newSym) => ref(newSym)
-        }
+        newFieldsMapping.getOrElse(rec.symbol, Nil).collect {
+          case ((oldSym, newSym)) if oldSym == t.symbol => ref(newSym)
```
nicolasstucki: This is now linear. Consider using a `LinkedHashMap` instead.

compiler/src/dotty/tools/dotc/transform/localopt/InlineLocalObjects.scala
```diff
@@ -34,11 +34,11 @@ class InlineLocalObjects(val simplifyPhase: Simplify) extends Optimisation {
   val gettersCalled = mutable.HashSet[Symbol]()
 
   // Map from class to new fields, initialised between visitor and transformer.
-  var newFieldsMapping: Map[Symbol, Map[Symbol, Symbol]] = null
-  //                   |           |       |
-  //                   |           |       New fields, replacements these getters
-  //                   |           Usages of getters of these classes
-  //                   ValDefs of the classes that are being torn apart; = candidates.intersect(gettersCalled)
+  var newFieldsMapping: Map[Symbol, List[(Symbol, Symbol)]] = null
```
nicolasstucki: Aren't these fields supposed to be private?

OlivierBlanvillain: I didn't bother annotating everything private, but yes, everything is `private[transform.localopt]` minimum.

nicolasstucki:
LGTM

tests/run/i3018.flags
```diff
@@ -0,0 +1 @@
+-optimise
```
smarter: flags files are ignored by vulpix, flags have to be put in the code in CompilationTests.scala

nicolasstucki: fixed in #3029

PR #3026, opened 2017-08-28
Fix #3024: repl alias in sbt and `dist/bin/dotr` not working

### LINKED ISSUE #3024: SBT alias `repl` references wrong `Main`
Usual ways as given in : http://dotty.epfl.ch/docs/contributing/getting-started.html
dotty.tools.dotc.repl.Main is not found anymore when using bin/dotr and repl task in sbt also no longer works
sbt dotty-repl/repl work OK, thanks @smarter for the tip

felixmulder:
We should also decide if we should ditch the old ./bin scripts in favor of the `./dist/bin` scripts. Keeping both up to date feels tedious.

### DIFF
--- a/dist/bin/dotc
+++ b/dist/bin/dotc
@@ -32,7 +32,7 @@ bootcp=true
 
 CompilerMain=dotty.tools.dotc.Main
 FromTasty=dotty.tools.dotc.FromTasty
-ReplMain=dotty.tools.dotc.repl.Main
+ReplMain=dotty.tools.repl.Main
 
 PROG_NAME=$CompilerMain
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -391,8 +391,6 @@ object Build {
         )
       }.evaluated,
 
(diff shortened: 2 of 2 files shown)


rom1dep:
I just discovered about ./dist/bin, it doesn't do anything here but printing errors about missing dotty/dist/lib/’.

If I had a vote for saving either, it would go to ./bin :)

liufengyun:
@rom1dep `dist/bin` is mainly for releasing Dotty packages. To try it, you need to first run `sbt dist/pack` then run `dist/target/pack/bin/dotr`.

For development, we usually run `dotc` from sbt, thus no script is needed for development.

rom1dep:
OK, I've been having ./bin in my path for a while and was appreciating the convenience of it triggering a rebuild as required. As my sole contribution here consists of fiddling with dotty's repl a few times a week/month, that's indeed not a strong defense for the extra maintenance cost :)

PR #3029, opened 2017-08-28
Fix flags for -optimise tests

nicolasstucki:


OlivierBlanvillain:
It would be nice to compile and run all the tests with -optimize like it was originally instead of this tiny subset, I will try to do that shortly.

LGTM

nicolasstucki:
@OlivierBlanvillain we should probably run them with and without `-optimise`.

PR #3032, opened 2017-08-29
Remove duplication from `UserFacingPrinter`

felixmulder:
@smarter: how far do we want to take this? Should I try to get rid of `UserFacingPrinter` completely?

I think there might be some things that won't print as expected if we do that. But I could try, anyway - up to you.


smarter:
I probably won't have time to review this week, you may want to ask someone else.

> how far do we want to take this? Should I try to get rid of UserFacingPrinter completely?

The less stuff we have the better it is I think, if a separate printer is needed for the REPL it needs to be well justified.

Also in https://github.com/lampepfl/dotty/pull/3032/commits/e331f87642a10e80db4cd931476bfd0ada92d9f3 you put `@smarter` in the commit message which means that everytime you re-push this commit I get an email, please avoid that :).

project/Build.scala
```diff
@@ -358,47 +359,6 @@ object Build {
...
-          case (beforeCp, Nil) => beforeCp ++ ("-classpath" :: classPath :: Nil)
-          case (beforeCp, rest) => beforeCp ++ rest
-        }
-
-        (runMain in Compile).toTask(
-          s" dotty.tools.repl.Main " + fullArgs.mkString(" ")
-        )
-      }.evaluated,
-
-      repl := run.evaluated,
```
smarter: Missing a definition `repl :=` in dotty-compiler now.

felixmulder: Fixed in #3026 :)

compiler/src/dotty/tools/dotc/printing/UserFacingPrinter.scala
```diff
@@ -0,0 +1,71 @@
+package dotty.tools.dotc
+package printing
+
+import core._
+import Constants.Constant, Contexts.Context, Denotations._, Flags._, Names._
+import NameOps._, StdNames._, Decorators._, Scopes.Scope, Types._, Texts._
+import SymDenotations.NoDenotation, Symbols.{ Symbol, ClassSymbol, defn }
+
+class UserFacingPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
```
OlivierBlanvillain: Do you use `_ctx` because `ctx` is already defined in `RefinedPrinter`?

OlivierBlanvillain:
It's great to have these things merged together, LGTM!

build.sbt
```diff
@@ -18,7 +18,6 @@ val `dotty-bench-bootstrapped` = Build.`dotty-bench-bootstrapped`
 val `scala-library` = Build.`scala-library`
 val `scala-compiler` = Build.`scala-compiler`
 val `scala-reflect` = Build.`scala-reflect`
-val `dotty-repl` = Build.`dotty-repl`
```
OlivierBlanvillain: Is it intentional to keep the `dotty-repl` project in Build.scala?

compiler/src/dotty/tools/dotc/printing/UserFacingPrinter.scala
felixmulder: Yes and that's what we do in RefinedPrinter when extending plain printer :)

build.sbt
felixmulder: Looks like an oversight when rebasing.

OlivierBlanvillain: (Remove it in #3044)

PR #3033, opened 2017-08-29
Native applied types

odersky:
This is ready for review now.


smarter:
Github seems to be having some issues with this PR and it's preventing the CI from checking it out. Could you try pushing to it again? If that doesn't work we may have to close this PR and open a new one.

smarter:
Ah, I think that what's going on is that our CI is configured to use the github branch corresponding to this PR merged on top of master. Since merging currently fails with conflicts, this branch doesn't exist. So I'm afraid you cannot test this on the CI until you rebase.

allanrenucci:
@odersky The last CI failure is not legitimate. Github picked the wrong CI (I am testing a new CI instance with a newer version of Drone). If you push a new commit, it should pick up the right CI.

repl/src/dotty/tools/repl/UserFacingPrinter.scala
```diff
@@ -21,6 +21,7 @@ import dotc.printing.{ GlobalPrec, DotPrec, Printer, PlainPrinter }
 import dotc.typer.Implicits.SearchResult
 import dotc.typer.ImportInfo
 
+// TODO: Avoid code duplication between userfacing and refined printers
```
smarter: Most of the duplication is gone in https://github.com/lampepfl/dotty/pull/3032 so you should be able to drop this commit when rebasing.

smarter:
The dotty-bin-tests CI failure should go away if you rebase (things changed in .drone.yml)

smarter:
The exhaustiveness checker is outputting lots of false warnings right now:
```scala
object Test {
  def test(i: Option[Int]) = i match {
    case Some(a) =>
    case None =>
  }
}
```
```scala
-- [E028] Pattern Match Exhaustivity Warning: try/pm.scala:2:29 ----------------
2 |  def test(i: Option[Int]) = i match {
  |                             ^
  |                             match may not be exhaustive.
  |                             
  |                             It would fail on: Some[Int](_)
```

odersky:
Seemed the CI could not find javac in two tests.

odersky:
test performance please

liufengyun:
There's an merge error for the PR which fails the compilation.  I'm investigating. Normally the test takes about 40min.

liufengyun:
I didn't a local merge of this PR with the master, but `dotty-bootstrapped/compile` generates exceptions. The same happens on the performance server.

smarter:
Make sure there's a `dotty-bootstrapped/clean` done before.

liufengyun:
Thanks for the tip @smarter -- now it works, the performance test will start immediately after #3054 finishes.

smarter:
@liufengyun Is there a way to ask the bot to benchmark a particular commit of the PR? It looks like we're regressing on scala-library on the last commit, but I've also noticed that the CI on the latest commit now takes longer to run than master, whereas it was faster this week-end before the last few commits were added. It'd be nice if we could check their impact without pushing new changes to the PR.

liufengyun:
Currently it's not supported, but it's not difficult to add this feature.
I'll make the feature work tomorrow---even to test multiple commits, so
developers can play with variants and see changes in one graph.

Le 5 sept. 2017 00:15, "Guillaume Martres" <notifications@github.com> a
écrit :

smarter:
@liufengyun By the way, are the benchmarks done with the non-bootstrapped dotty or the bootstrapped dotty?

liufengyun:
@smarter Yesterday, I've changed to bootstrapped dotty.

smarter:
@liufengyun Ah! But then all the historical data points are with the non-bootstrapped dotty? That means we cannot do any valid comparison between the result for this PR and the existing points.

liufengyun:
@smarter The date always have some noise, but the variation from non-bootstrapped to bootstrapped is very low, as can be seen from http://dotty-bench.epfl.ch.

liufengyun:
BTW, the points for yesterday's merge are already bootstrapped -- that are the points just before the last one.

smarter:
> BTW, the points for yesterday's merge are already bootstrapped -- that are the points just before the last one.

That's very surprising since these points do not look any different from the non-bootstrapped points. I would expect some difference in one direction or another. Are we completely sure the new points are bootstrapped? :) This also conflicts with reports from https://github.com/scala/compiler-benchmark/pull/31#issuecomment-324277065 which say that dotty is much slower than scalac.

smarter:
https://github.com/jvm-profiling-tools/async-profiler can be used to generate flame graphs, but it requires Linux. It'd be really cool if we could run this as part of the benchmarks.

liufengyun:
> That's very surprising since these points do not look any different from the non-bootstrapped points. I would expect some difference in one direction or another. Are we completely sure the new points are bootstrapped? :)

It's definitely the bootstrapped version.

It probably means that the generated code from Dotty is comparable to Scalac 2.12, or the optimisations in Dotty is insignificant compared to that of JVM.

smarter:
test performance please d92e3e6 fb82b09 fc83c24 3034f59 1e6572d 7204538 393d91f 9af7324 922b379 a7ec349

smarter:
@liufengyun Oops, I guess https://github.com/liufengyun/bench/pull/47 is not active yet ^ ?

liufengyun:
@smarter : a colon is required after `please`. I just remove the job from the queue. You can try issue the command again.

smarter:
test performance please: d92e3e6 fb82b09 fc83c24 3034f59 1e6572d 7204538 393d91f 9af7324 922b379 a7ec349

smarter:
@liufengyun  Hmm I think there's a bug in your script that I didn't caught while reviewing it: 0 is allowed in git hashes, but the bot replaced 3034f59 by 334f59 :smile:

liufengyun:
@smarter : Thanks, I just noticed, fixed, deployed and removed the job.

smarter:
test performance please: d92e3e6 fb82b09 fc83c24 3034f59 1e6572d 7204538 393d91f 9af7324 922b379 a7ec349

smarter:
Hmm, 0 is still missing in the hashes.

liufengyun:
The polling service just restarted.

smarter:
test performance please: d92e3e6 fb82b09 fc83c24 3034f59 1e6572d 7204538 393d91f 9af7324 922b379 a7ec349

liufengyun:
The daily benchmark is running from 12:00pm. Depending on how many PRs we merge today, each PR or commit will take ~40min. If everything goes well, we should be able to get the result in the morning.

smarter:
@liufengyun The results don't seem to contain the commits scheduled for benchmarking.

liufengyun:
@smarter I'm investigating the problem, will keep you updated once it works.

liufengyun:
test performance please: d92e3e6 fb82b09 fc83c24 3034f59 1e6572d 7204538 393d91f 9af7324 922b379 a7ec349

liufengyun:
test performance please: d92e3e6 fb82b09 fc83c24

smarter:
Looks like https://github.com/lampepfl/dotty/commit/fb82b096f18d87a80216a651e8dc819f477734ea was bad for performance, and https://github.com/lampepfl/dotty/commit/fc83c2455d47dadd37d240a36eccf2dbd1de267d made things better but still worse than before.

@liufengyun It would be nice if clicking on a point on the graph jumped to the commit diff on github instead of the corresponding PR, sorry for making another feature request :).

liufengyun:
@smarter I'm happy to improve the bench infrastructure. Let's talk more about the feature tomorrow, it seems in some cases jump to PR is more friendly.

odersky:
We already integrated the optimizations that are easily applicable, and the rest is complicated and does not seem to buy us much.

PR #3035, opened 2017-08-30
Use new docker image with updated cache after the switch to 2.12

OlivierBlanvillain:
log: https://gist.github.com/OlivierBlanvillain/de56bb13951ca87d9d86aabbbf41c450


felixmulder:
Gotta sign the file mate `drone sign lampepfl/dotty` 👍

allanrenucci:
I guess you should update all the images in the file

allanrenucci:
From my understanding we set `pull: true` in the `done.yml` file to always pull the latest image. However, to have the latest image we should write `image:lampepfl/dotty:latest` and a specific version.
Am I missing something?

OlivierBlanvillain:
Tags of docker images are mutable, with `pull: true` drone will always ask docker-hub, "hey is <local-hash> the latest version of `lampepfl/dotty:2017-08-30`". I guess it could be removed since the plan is to keep all images for better reproducibility :)

PR #3045, opened 2017-08-31
Fix #2395: Check if the scala or dotty libraries are missing

### LINKED ISSUE #2395: Provide a nice error message if `scala.` isn't found
We should check that classes from dotty-library and scala-library can be found and report a nice error.

Current error message:

```
java.lang.Exception
    at dotty.tools.dotc.core.Denotations$MissingRef.<init>(Denotations.scala:1034)
    at dotty.tools.dotc.core.Denotations$MissingRef$.apply(Denotations.scala:1033)
    at dotty.tools.dotc.core.Denotations$DenotationsBase.select$1(Denotations.scala:1183)
creating stub for scala, owner = module class <root>#2, file = null
decls = module sun#3
  module class sun#4
...
bad symbolic reference. A signature
refers to scala[ModuleClass]/T in package <root> which is not available.
It may be completely missing from the current classpath, or the version on
the classpath might be incompatible with the version used when compiling the signature.
Exception in thread "main" java.lang.UnsupportedOperationException: class dotty.tools.dotc.core.Scopes$EmptyScope$.openForMutations
    at dotty.tools.dotc.util.DotClass.unsupported(DotClass.scala:10)
    at dotty.tools.dotc.core.Scopes$Scope.openForMutations(Scopes.scala:161)
```

nicolasstucki:


compiler/test/dotty/tools/vulpix/TestConfiguration.scala
```diff
@@ -23,8 +23,10 @@ object TestConfiguration {
     "-Yforce-sbt-phases"
   )
 
-  val classPath = {
-    val paths = Jars.dottyTestDeps map { p =>
+  val classPath = mkClassPath(Jars.dottyTestDeps)
+
+  def mkClassPath(deps: List[String]) = {
```
allanrenucci: Explicit return type?

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -1199,7 +1199,10 @@ object Denotations {
             val alt =
               if (generateStubs) missingHook(owner.symbol.moduleClass, selector)
               else NoSymbol
-            if (alt.exists) alt.denot else MissingRef(owner, selector)
+            if (alt.exists) alt.denot
+            else if (owner.symbol == defn.RootClass && (selector == nme.scala_ || selector == nme.scalaShadowing))
```
allanrenucci: Maybe document what this test or extract it to an inner function.

I don't have a good understanding of this does though. @smarter?

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -1192,14 +1192,23 @@ object Denotations {
...
           val result = if (isPackage) owner.info.decl(selector) else owner.info.member(selector)
           if (result.exists) result
           else {
             val alt =
               if (generateStubs) missingHook(owner.symbol.moduleClass, selector)
               else NoSymbol
-            if (alt.exists) alt.denot else MissingRef(owner, selector)
+            if (alt.exists) alt.denot
+            else if (isPackageFromCoreLibMissing) throw new MissingCoreLibraryException(selector.toString)
+            else MissingRef(owner, selector)
```
nicolasstucki: @smarter could you review the `isPackageFromCoreLibMissing` condition in this file. Thanks.

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -1192,14 +1192,23 @@ object Denotations {
     def staticRef(path: Name, generateStubs: Boolean = true, isPackage: Boolean = false)(implicit ctx: Context): Denotation = {
       def select(prefix: Denotation, selector: Name): Denotation = {
         val owner = prefix.disambiguate(_.info.isParameterless)
+        def isPackageFromCoreLibMissing: Boolean = {
+          owner.symbol == defn.RootClass &&
+          {
```
smarter: Use `(` not `{`

compiler/src/dotty/tools/dotc/core/Denotations.scala
smarter: Looks fine to me.

nicolasstucki:
Rebased after drone update

PR #3054, opened 2017-09-04
Fix #3015: use type inference to type child classes

### LINKED ISSUE #3015: Remove magic in exhaustivity check
In the exhaustivity check, when decomposing a type `path.B[Int, String]`, we get two child classes `C1` and `C1`. The problem is how to turn the two classes into proper types related to `path`, `Int` and `String`?

Current implementation resorts to some ad-hoc hack which is not well-grounded. The correct way to solve the problem is to use type inference, as suggested by Martin.

liufengyun:
Fix #3015: use type inference to type child classes

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -7,16 +7,19 @@ import Types._
 import Contexts._
 import Flags._
 import ast.Trees._
-import ast.tpd
+import ast.{tpd, untpd}
 import Decorators._
 import Symbols._
 import StdNames._
 import NameOps._
 import Constants._
-import typer.Applications._
+import typer._
+import Applications._
+import Inferencing._
+import ProtoTypes._
 import transform.SymUtils._
 import reporting.diagnostic.messages._
-import config.Printers.{ exhaustivity => debug }
+import config.Printers.{exhaustivity => debug}
 
 /** Space logic for checking exhaustivity and unreachability of pattern matching
  *
@@ -524,20 +527,10 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       case tp =>
         val parts = children.map { sym =>
           if (sym.is(ModuleClass))
-            refine(tp, sym.sourceModule.termRef)
-          else if (sym.isTerm)
-            refine(tp, sym.termRef)
-          else if (sym.info.typeParams.length > 0 || tp.isInstanceOf[TypeRef])
-            refine(tp, sym.typeRef)
+            refine(tp, sym.sourceModule)
           else
-            sym.typeRef
-        } filter { tpe =>
-          // Child class may not always be subtype of parent:
-          // GADT & path-dependent types
-          val res = tpe <:< expose(tp)
-          if (!res) debug.println(s"unqualified child ousted: ${tpe.show} !< ${tp.show}")
-          res
-        }
+            refine(tp, sym)
+        } filter(_.exists)
 
         debug.println(s"${tp.show} decomposes to [${parts.map(_.show).mkString(", ")}]")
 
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -524,43 +527,80 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-  def refine(tp1: Type, tp2: Type): Type = (tp1, tp2) match {
-    case (tp1: RefinedType, _: TypeRef) => tp1.wrapIfMember(refine(tp1.parent, tp2))
-    case (tp1: HKApply, _) => refine(tp1.superType, tp2)
-    case (TypeRef(ref1: TypeProxy, _), tp2 @ TypeRef(ref2: TypeProxy, _)) =>
-      if (ref1.underlying <:< ref2.underlying) tp2.derivedSelect(ref1) else tp2
-    case (TypeRef(ref1: TypeProxy, _), tp2 @ TermRef(ref2: TypeProxy, _)) =>
-      if (ref1.underlying <:< ref2.underlying) tp2.derivedSelect(ref1) else tp2
-    case _ => tp2
+  def instantiate(tp1: Type, tp2: Type)(implicit ctx: Context): Type = {
+    val approx = new TypeMap {
```
odersky: What does approx do? A more telling name or a comment would help.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -524,43 +527,80 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      def apply(t: Type): Type = t match {
+        case t @ ThisType(tref) if !tref.symbol.isStaticOwner && !tref.symbol.is(Module)  =>
+          newTypeVar(TypeBounds.upper(mapOver(tref & tref.givenSelfType)))
```
odersky: I think you want 

    newTypeVar(TypeBounds.upper(mapOver(t.underlying)))

here. That achieves what was written in simpler form. But why `upper`? Without an explanation it's hard to know whether it is correct. But it feels dangerous because the ThisType might appear at negative variance positions as well. How do you arge that the bound should not be changed in this case?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -524,43 +527,80 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      }
+    }
+
+    val tvars = tp1.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
+    val protoTp1 = approx(tp1.appliedTo(tvars))
+
+    if (protoTp1 <:< tp2 && isFullyDefined(protoTp1, ForceDegree.all)) protoTp1
+    else {
+      debug.println(s"$protoTp1 <:< $tp2 = false")
+      NoType
```
odersky: Neat that this works! We might want to move it to something more generally available since it looks like the Linker will need something like this was well /cc @nicolasstucki. @DarkDimius

odersky:
Can rou rebase this PR on top of #3033? That would be the real test that exhaustiuveness checking is no longer tied to the particularities of how we encode applied types. I tried but got failures.

liufengyun:
test performance please

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: The argument here is that the child type created from `childSym.typeRef` can only be of the shape `path.Child`, thus the `ThisType`  is always covariant.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: Indeed, a better name is more readable.

PR #3059, opened 2017-09-04
fix #2996: remove useless Makefile in distributed archive

### LINKED ISSUE #2996: Purpose of the Makefile in the binary release
What's the purpose of the Makefile in the binary release? `./bin/dotr` worked fine without running `make`. When should you run make?

liufengyun:
Fix #2996: remove useless Makefile in distributed archive

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -8,7 +8,8 @@ import java.util.Calendar
 
 import scala.reflect.io.Path
 import sbtassembly.AssemblyKeys.assembly
-import xerial.sbt.Pack._
+import xerial.sbt.pack.PackPlugin._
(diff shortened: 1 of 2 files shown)


allanrenucci:
Should we rebuild the Docker image since you upgraded a dependency?

liufengyun:
@allanrenucci if the docker has network access or ivy mapping, then it seems fine.

allanrenucci:
Sure it can download the updated dependency but it will slow down the build. Rebuilding the Docker will put it in the cache. I think. @felixmulder

felixmulder:
Yes, you guys should totes rebuild ze cache :+1:

PR #3064, opened 2017-09-05
Benchmark with no patmat optimizations

smarter:


smarter:
test performance please

smarter:
test performance please

smarter:
@liufengyun It's unfortunate that when retesting a PR the same url is used, so the old benchmark results are overwritten.

liufengyun:
@smarter Before we improve that, all web site versions are on github: https://github.com/liufengyun/bench/commits/gh-pages/3064

You just need to checkout the specific version, and use a tool like [http-server](https://www.npmjs.com/package/http-server) to show the version you are interested in.

In the next version, I'll give different for different test command.

PR #3066, opened 2017-09-05
0.3.0-RC2

allanrenucci:


smarter:
Note: disabling the 2.12 support was needed because the 2.12 detection was done by looking at the Java bytecode version, this fails for code compiled with 2.11 configured to emit Java 8 bytecode, this happened in practice: https://gitter.im/lampepfl/dotty?at=59ad4117b16f26464203c4c0

PR #3061, opened 2017-09-05
Applied types, unoptimized

odersky:
This PR covers AppliedTypes before any attempt at optimizations. Recent experience has shown that attempts as optimizations often have unintuitive results, so we should test them very carefully and measure each of them individually.

PR #3074, opened 2017-09-06
Fix #3015: exhaustivity check on top of native apply

### LINKED ISSUE #3015: Remove magic in exhaustivity check

liufengyun:
Fix #3015: exhaustivity check on top of native apply


liufengyun:
test performance please

liufengyun:
The [performance graph](http://dotty-bench.epfl.ch/3074) shows that the new exhaustivity check performs the same on normal patmat tests as before, but degrades significantly on the abnormal test case [V](https://github.com/liufengyun/bench/blob/master/tests/exhaustivity-V.scala): from 2s to 3.5s.

odersky:
We see now a much bigger performance gain than with applied types alone on dotc. It's about 10% faster as opposed to maybe 2 or 3% before. Interesting, given that exhaustivity checks are now slower than before. My guess is that the reporting of the spurious exhaustivity warnings in dotc itself (there were about 300 of them) impacted overall performance. Anyway, getting 10% on dotc and 6-7 % on Vector or stdlib is very encouraging!

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -403,20 +403,31 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-    case Typed(expr, _) => Typ(expr.tpe.stripAnnots, true)
+    case Typed(expr, _) => Typ(erase(expr.tpe.stripAnnots), true)
     case _ =>
       debug.println(s"unknown pattern: $pat")
       Empty
   }
 
+  /* Erase a type binding according to erasure semantics in pattern matching */
+  def erase(tp: Type): Type = tp match {
+    case tp@AppliedType(tycon, args) => erase(tp.superType)
```
odersky: Useless erase, this is thrown away

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -403,20 +403,31 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      if (tycon.isRef(defn.ArrayClass)) tp.derivedAppliedType(tycon, args.map(erase))
+      else tp.derivedAppliedType(tycon, args.map(t => WildcardType(TypeBounds.empty)))
```
odersky: I think you can just use the WildcardType object here.

liufengyun:
@odersky The printing of exhaustivity warnings involves a lot of IO, which has a bad impact on performance. You can also test performance by disable the exhautivity check in the pattern matcher, and see how it improves the numbers.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -598,10 +579,34 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     val tvars = tp1.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
     val protoTp1 = thisTypeMap(tp1.appliedTo(tvars))
 
-    if (protoTp1 <:< tp2 && isFullyDefined(protoTp1, ForceDegree.all)) protoTp1
+    // replace type parameter references with fresh type vars or bounds
+    val typeParamMap = new TypeMap {
+      def apply(t: Type): Type = t match {
+
+        case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
+          // See tests/patmat/gadt.scala  tests/patmat/exhausting.scala
```
odersky: Add: if tp.symbol.is(TypeParam) &&

smarter:
> The performance graph shows that the new exhaustivity check performs the same on normal patmat tests as before, but degrades significantly on the abnormal test case V: from 2s to 3.5s.

Do you know why https://github.com/liufengyun/bench/blob/master/tests/exhaustivity-V.scala performs worse now?

liufengyun:
@smarter I think it's because the type inference approach is more expensive than the previous hack for generics.

smarter:
But there is nothing to infer in https://github.com/liufengyun/bench/blob/master/tests/exhaustivity-V.scala, right? Maybe we could avoid the expensive use of type inference in some cases. (Of course this shouldn't be required to get this PR in, but maybe something to investigate later)

liufengyun:
@smarter Good point, as `Product` is not decomposable, the code path should be the same as before for the test `V`, as we didn't make any changes to the space algorithm.

liufengyun:
Superseded by #3061

Issue #2924, opened 2017-07-27
tests/run/bridges.scala cannot be compiled by clean compiler


nicolasstucki:
It can be reproduced with:
`vulpix tests/run/bridges.scala`

but not with
`vulpix`

nicolasstucki:
This has been happening since it was added in 92fe2a50a91b22c3efa0f1b5746378d987966429

nicolasstucki:
There is probably some state left from previous runs in the compiler that allows it to work.

smarter:
No idea why it doesn't break usually, but `bridges.scala` contains a block with 1424 statements which causes `withDefinedSyms` to blow up with a stackoverflow because `foldRightBN` is not tail-recursive. This should be easy to fix by replacing `foldRightBN` by a while loop. But someone should investigate why this doesn't trigger when running all the tests, maybe vulpix forgets to run bridges.scala for some reason?

DarkDimius:
@smarter @nicolasstucki VM needs different amount of stack in JITed and interpreted mode.
You can have code that does not stackoverflow when compiled but does when interpreted.

PR #3076, opened 2017-09-06
Mitigate #2924

nicolasstucki:
To make the stack shorter we implement foldRightBN with
a foldLeftBN which allocates only half of the closures.
By doing so we need to eagerly reverse the list which is
in most cases small or empty.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Decorators.scala
+++ b/compiler/src/dotty/tools/dotc/core/Decorators.scala
@@ -112,9 +112,15 @@ object Decorators {
         else x1 :: xs1
       }
 
-    def foldRightBN[U](z: => U)(op: (T, => U) => U): U = xs match {
-      case Nil => z
-      case x :: xs1 => op(x, xs1.foldRightBN(z)(op))
+    def foldRightBN[U](z: => U)(op: (T, => U) => U): U =
+      xs.reverse.foldLeftBN(z)(op)
+
+    final def foldLeftBN[U](acc: => U)(op: (T, => U) => U): U = {
+      @tailrec def fold(xs: List[T], acc: => U): U = xs match {
+        case x :: xs1 => fold(xs1, op(x, acc))
+        case Nil => acc
+      }
+      fold(xs, acc)
     }
 
     final def hasSameLengthAs[U](ys: List[U]): Boolean = {
--- a/compiler/src/dotty/tools/dotc/core/Decorators.scala
+++ b/compiler/src/dotty/tools/dotc/core/Decorators.scala
@@ -112,15 +112,12 @@ object Decorators {
         else x1 :: xs1
       }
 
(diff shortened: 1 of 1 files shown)


nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/core/Decorators.scala
```diff
@@ -112,9 +112,15 @@ object Decorators {
...
+    def foldRightBN[U](z: => U)(op: (T, => U) => U): U =
+      xs.reverse.foldLeftBN(z)(op)
+
+    final def foldLeftBN[U](acc: => U)(op: (T, => U) => U): U = {
+      @tailrec def fold(xs: List[T], acc: => U): U = xs match {
+        case x :: xs1 => fold(xs1, op(x, acc))
+        case Nil => acc
+      }
+      fold(xs, acc)
     }
```
allanrenucci: Should be
```scala
final def foldLeftBN[U](acc: => U)(op: (=> U, T) => U): U = {
  ...
  case x :: xs1 => fold(xs1, op(acc, x))
```
That's what I meant by "accumulator is the first arg of the closure"

nicolasstucki: That is right, but if I do that the `op` must be put in another closure to reverse the parameter order.
I will make `fold` an inner method of `foldRightBN` and avoid this altogether.

nicolasstucki:
test performance please

nicolasstucki:
While executing `vulpix` run tests, `foldRightBN` is mostly called on lists of size 0 and 1. This does not count the recursive calls that would have been in the original implementation.
```
size->invocations
0->1111034
1->1081070
2->142603
3->44560
4->19410
5->8234
6->5164
7->2277
8->1635
9->963
10->789
11->542
12->433
13->292
14->239
15->198
16->214
17->159
18->177
19->153
20->199
21->178
22->145
23->93
24->27
25->35
26->150
27->27
28->27
29->7
30->28
31->24
32->24
33->3
35->9
36->9
37->5
39->9
40->2
43->5
45->7
46->28
49->38
51->3
52->31
53->4
54->5
56->3
57->7
58->7
61->3
62->14
65->7
67->2
69->5
70->7
73->3
79->3
86->7
95->3
109->14
111->7
164->3
453->7
523->3
1424->7
```

allanrenucci:
LGTM

biboudis:
@allanrenucci I was going to write on my review that we may need an extra test case. :-) Yesterday, we discussed with @nicolasstucki, that maybe an extra test would be needed to stress test this.

allanrenucci:
@biboudis This does not solve the core issue but mitigates it. @nicolasstucki will try to solve the core issue. We can include the test case in the next PR.

I merged this one to unblock some CI builds.

biboudis:
👍 👍

PR #3079, opened 2017-09-06
Revert phantom casts

nicolasstucki:


odersky:
I think this is the right way to go. There should be no special rules for phantom types, or as few as possible. We can discuss it internally.

odersky:
LGTM

PR #3065, opened 2017-09-05
Fix #3050: Avoid retyping synthesised code

biboudis:
This is a crude way to fix #3050's bugs with retyping inlined code involving pattern matching.

WDYT?

PR #3080, opened 2017-09-06
Fix #2924: make TreeChecker.typedBlock iterative

### LINKED ISSUE #2924: tests/run/bridges.scala cannot be compiled by clean compiler

nicolasstucki:
This PR makes it posssible to compile arbitarley large blocks without growing the stack.


nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
test performance please

biboudis:
Yeap. I think this is a blocker for #3065 and more as we saw over the last weeks. /cc @odersky

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -92,26 +92,27 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
   private def reorderAndComplete(stats: List[Tree])(implicit ctx: Context): List[Tree] = {
     val moduleClassDefs, singleClassDefs = mutable.Map[Name, Tree]()
 
-    def reorder(stats: List[Tree]): List[Tree] = stats match {
+    def reorder(stats: List[Tree], revPrefix: List[Tree] = Nil): List[Tree] = stats match {
```
odersky: This is a rather complex algorithm. Whenever I see reversed lists my head starts to hurt. Can we simplify it using ListBuffers? Also, if we keep it we need more docs to explain it.

compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -149,34 +149,42 @@ class TreeChecker extends Phase with SymTransformer {
 
     def withDefinedSym[T](tree: untpd.Tree)(op: => T)(implicit ctx: Context): T = tree match {
       case tree: untpd.DefTree =>
-        val sym = tree.symbol
```
odersky: I think the original logic was complicated - that foldRightBN is sneaky because it mixes laziness with side-effects ! The changes remove the foldRightBN complexity but make everything else more complex. Proposal: make `withDefinedSyms` take a repeated argument and remove withDefinedSym. I.e.

       def withDefinedSyms[T](trees: untpd.Tree*)(op: => T)(implicit ctx: Context)

`withDefinedSyms` can step iteratively through all the trees, record the symbols found as defined in a ListBuffer or List and then execute `op`.

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
odersky: To clarify: I believe the original reorder algorithm was already close to the limit of tolerable complexity.

odersky:
Here's a version of `reorder` using a `ListBuffer`. Downside is that `insert` on listbuffers has linear complexity. So I am not sure this is an improvement. But maybe we find something else...

    def reorder(stats: List[Tree], out: mutable.ListBuffer[Tree]): Unit = stats match {
      case (stat: TypeDef) :: stats1 if stat.symbol.isClass =>
        if (stat.symbol is Flags.Module) {
          moduleClassDefs += (stat.name -> stat)
          singleClassDefs -= stat.name.stripModuleClassSuffix
          val idx = out.size
          reorder(stats1, out)
          if (moduleClassDefs contains stat.name) out.insert(idx, stat)
        } else {
          out += stat
          val idx = out.size
          reorder(stats1, out)
          moduleClassDefs remove stat.name.moduleClassName match {
            case Some(mcdef) => out.insert(idx, mcdef)
            case None => singleClassDefs += (stat.name -> stat)
          }
        }
      case stat :: stats1 =>
        out += stat
        reorder(stats1, out)
      case Nil =>
    }

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
odersky: Thinking more about it, I believe the `revPrefix` idea is the best option. So OK to leave the code as is. Just document what `revPrefix` stands for. I.e. at any point:

    reordered statements = revPrefix.reverse ++ reorder(stats, Nil)

Also, I'd not make `revPrefix` have a default parameter. To easy to make errors that way.

nicolasstucki:
test performance please

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -92,26 +92,35 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
...
 
-    def reorder(stats: List[Tree]): List[Tree] = stats match {
+    /* Returns the result of reordering stats and prepending revPrefix in reverse order to it.
+     * The result of reorder is equivalent to reorder(stats, revPrefix) = revPrefix.reverse ::: reorder(stats, Nil).
+     * This implementation is tail recursive as long as the element is not a module TypeDef.
+     */
+    def reorder(stats: List[Tree], revPrefix: List[Tree]): List[Tree] = stats match {
       case (stat: TypeDef) :: stats1 if stat.symbol.isClass =>
         if (stat.symbol is Flags.Module) {
+          def pushOnTop(xs: List[Tree], ys: List[Tree]): List[Tree] = xs match {
```
odersky: This one is a natural fold

    (ys /: xs)((ys, x) => x :: ys)

odersky: But it's also OK to leave as it is

PR #3084, opened 2017-09-07
Simplify CI builds

allanrenucci:


project/scripts/sbt
```diff
@@ -1,26 +1,20 @@
...
-  if [ $2 = true ]; then
-    SHOULD_RUN=true
-  else
-    SHOULD_RUN=false
-  fi
-else
-  SHOULD_RUN=true
+CMD="$1"
+
+if [ -z "$CMD" ]; then
```
felixmulder: Nitpick, this should be:

```bash
if [ -z ${CMD+x} ]; then
```

otherwise you can pass empty string here and it will accept it as valid input.

project/scripts/sbtPublish
```diff
@@ -1,34 +1,26 @@
...
-SHOULD_RUN=$1
-
-# Sonatype credentials:
-SONATYPE_USER=$2
-SONATYPE_PW=$3
-
-# PGP Credentials:
-PGP_PW=$4
+# Usage:
+# SONATYPE_USER=<sonatype user> SONATYPE_PW=<sonatype pw> PGP_PW=<pgp pw> ./sbtPublish <publish cmd>
```
felixmulder: :+1:

allanrenucci:
Unfortunately there seems to be a major issue. The builds in the same group run in parallel but they share the same workspace... So they fail 😭 
Maybe there is a way to not share the same workspace.

project/scripts/sbt
allanrenucci: My understanding is that it is the other way around:
```shell
> FOO=""; if test -z ${FOO+x}; then echo "Empty"; else echo "Full"; fi
Full
> FOO=""; if test -z "$FOO"; then echo "Empty"; else echo "Full"; fi
Empty
```

felixmulder: Maybe I wasn't clear - you're testing if the variable `FOO` is empty string - instead of testing if the variable is set at all. E.g. setting `FOO=""` versus not setting it at all

:)

felixmulder:
> Maybe there is a way to not share the same workspace.

Yeah, this was our original fear with the matrix builds as well - but turns out that the matrix build actually separates things :joy:

project/scripts/sbt
felixmulder: But seriously though, nitpick 😛

allanrenucci: I want the script to exit if the variable is not set or if it is an empty string. An empty cmd should be reject IMO. With `-z "$CMD"`, it rejects both cases.

smarter:
The ivy cache does not seem to be used anymore with this PR, in http://dotty-ci.epfl.ch/lampepfl/dotty/103/1/test everything gets downloaded.

allanrenucci:
@smarter Good catch. I was trying to understand the slow down, this might be the cause

project/scripts/sbt
```diff
@@ -1,26 +1,20 @@
...
+  echo "Error: missing sbt command"
+  exit 1
 fi
 
-if [ $SHOULD_RUN = true ]; then
-  # get the ivy2 cache:
-  ln -s /var/cache/drone/ivy2 "$HOME/.ivy2" || true
```
allanrenucci: @felixmulder What was `|| true` for?

allanrenucci:
@felixmulder Can you review again. I managed to solve the issue of shared workspace by running each instance of `sbt` in its own directory

project/scripts/sbt
felixmulder: I think this was just there to allow `ln` to fail, having a cache is not the most important thing to this script. It could definitely be removed, but we then loose the ability to test without `/var/cache/drone/ivy2` being present - not a big deal I guess.

project/scripts/genDocs
```diff
@@ -1,68 +1,63 @@
...
+# make sure that BOT_PASS is set
+if [ -z "$BOT_PASS" ]; then
+    echo "Error: BOT_PASS env unset, unable to push without password" 1>&2
+    exit 1
 fi
+
+echo "Working directory: $PWD"
+
+# get the ivy2 cache
+ln -s /var/cache/drone/ivy2 "$HOME/.ivy2"
```
felixmulder: Just a thought that popped up - does it make sense to do the release builds without the cache?

cc: @smarter

project/scripts/sbt
allanrenucci: I had the `ln` command run as an initial step before the tests but it was not persisted between the steps and the CI started re-downloading artifacts.  So I added it in all the CI scripts

.drone.yml
```diff
@@ -1,17 +1,56 @@
 pipeline:
+  # We run tests in parallel. Tests run in a copy of the working directory to avoid conflict
+  test_legacy:
+    group: test
+    image: lampepfl/dotty:2017-09-05
+    commands:
+      - cp -R . /tmp/0/ && cd /tmp/0/
```
felixmulder: I see you went for the shortest possible name here 😄

project/scripts/sbt
felixmulder: We used to have a `common` script that was sourced from the other scripts - if there's too much duplication, you could consider doing something similar eventually.

gsps: Note that `ln` also fails if "$HOME/.ivy2" already exists, so I think you might get spurious error messages in the log upon running the script the second time on the same container.

felixmulder: Ah, true! That's exactly, I remember now!

allanrenucci: Does `ln` survive multiple steps? Because I tried to do it once before all the steps, but then the CI would redownload artifacts: http://dotty-ci.epfl.ch/lampepfl/dotty/102

allanrenucci: I will experiment

felixmulder: It doesn't seem so  🤔

felixmulder: It does however survive within a single `command` section, IIRC

.drone.yml
```diff
@@ -1,17 +1,56 @@
...
+    secrets: [ bot_pass ]
+    when:
+      event: push
 
   publish_nightly:
     image: lampepfl/dotty:2017-09-05
     environment:
       - NIGHTLYBUILD=yes
     commands:
-      - ./project/scripts/sbt ";clean ;publishLocal" "${CI_PUBLISH}"
```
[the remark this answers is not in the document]
felixmulder: IIRC something to do with testing the SBT integrations, but I'll defer to @smarter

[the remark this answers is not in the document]
smarter: I don't publishLocal has been necessary for anything in a long time.

project/scripts/genDocs
smarter: > does it make sense to do the release builds without the cache?

I don't see in what situation that could make a difference?

project/scripts/sbtPublish
```diff
@@ -1,40 +1,38 @@
...
+ln -s /var/cache/drone/ivy2 "$HOME/.ivy2"
+
 CMD="     ;set credentials in ThisBuild := Seq(Credentials(\"Sonatype Nexus Repository Manager\", \"oss.sonatype.org\", \"$2\", \"$3\"))"
 CMD="$CMD ;set pgpPassphrase := Some(\"\"\"$4\"\"\".toCharArray)"
 CMD="$CMD ;set pgpSecretRing := file(\"/keys/secring.asc\")"
 CMD="$CMD ;set pgpPublicRing := file(\"/keys/pubring.asc\")"
 CMD="$CMD $RELEASE_CMD"
 
+# run sbt with the supplied arg
 sbt -J-Xmx4096m \
```
smarter: Would be nice if there was less duplication between sbt and sbtPublish

smarter:
Also probably unrelated to this PR but the CI logs used to be in color, what happened to that?

allanrenucci:
>Stuff is still being downloaded

@smarter This happened also before this PR: http://dotty-ci.epfl.ch/lampepfl/dotty/116/3/test
It happens during the bootstrapped tests because when we generated the Dotty Docker image, we didn't compile with bootstrapped Dotty. I can address this in a separate PR

allanrenucci:
>Also probably unrelated to this PR but the CI logs used to be in color, what happened to that?

I see colored logs. Maybe an issue with Firefox?

project/scripts/genDocs
```diff
@@ -1,68 +1,64 @@
...
+
+# this command will generate docs in $PWD/docs/_site
+sbt -J-Xmx4096m \
+    -J-XX:ReservedCodeCacheSize=512m \
+    -J-XX:MaxMetaspaceSize=1024m \
+    -Ddotty.drone.mem=4096m \
```
allanrenucci: @felixmulder What is this option for?

project/scripts/genDocs
felixmulder: Corrupt cache causing issues with jars - that's why we chose to have it static rather than using the dynamic caching plugin. Other than that - I've got nothing :)

project/scripts/genDocs
felixmulder: See https://github.com/lampepfl/dotty/blob/master/project/Build.scala#L627-L632

allanrenucci: Is it different than setting `-J-Xmx4096m`?

allanrenucci: Got it. One is for the JVM running sbt and the other for the forked JVM for the tests

allanrenucci:
@smarter I removed the duplication in the scripts. If this looks good to you, I am merging it

project/scripts/sbt
```diff
@@ -3,20 +3,17 @@
...
 
-# get the ivy2 cache
-ln -s /var/cache/drone/ivy2 "$HOME/.ivy2"
-
 # run sbt with the supplied arg
 sbt -J-Xmx4096m \
     -J-XX:ReservedCodeCacheSize=512m \
     -J-XX:MaxMetaspaceSize=1024m \
-    -Ddotty.drone.mem=4096m "$CMD"
+    -Ddotty.drone.mem=4096m \
```
OlivierBlanvillain: So we still use that thing?

project/scripts/sbt
```diff
@@ -3,20 +3,17 @@
...
+    -ivy /var/cache/drone/ivy2 \
```
OlivierBlanvillain: I this doesn't do what you think it does. This way you still download all the dependence needed for sbt itself...

project/scripts/sbt
allanrenucci: From what I understood, `-J-Xmx4096m` is an option passed to the VM that run `sbt` and `Ddotty.drone.mem=4096m` will be passed to the forked VMs

project/scripts/sbt
allanrenucci: No sure about that. If you look at the CI test logs (except for test_bootstrapped), I don't see any download.

smarter:
Would be good to make sure there is no speed loss by comparing master CI time and this branch CI time when nothing else is running.

allanrenucci:
No thorough benchmark but this PR took 12 min to build when nothing else was running. Looking at the history of the builds, build time is rarely under 12min.

PR #3088, opened 2017-09-08
Tune bench warmups

nicolasstucki:


bench/src/main/scala/Benchmarks.scala
```diff
@@ -22,8 +22,15 @@ object Bench {
       println("Missing <args>")
       return
     }
+    val (intArgs, args1) = args.span(x => try { x.toInt; true } catch { case _: Throwable => false } )
 
-    val args2 = args.map { arg =>
+    var warmup = 30
+    var iterations = 20
+
+    if (intArgs.length > 0) warmup = intArgs(0).toInt
```
allanrenucci:
```scala
val warmup = if (intArgs.length > 0) intArgs(0).toInt else 30
```

bench/src/main/scala/Benchmarks.scala
```diff
@@ -22,8 +22,15 @@ object Bench {
...
+    if (intArgs.length > 1) iterations = intArgs(1).toInt
```
allanrenucci: same

nicolasstucki:
Yes, the CI failure was due to #2924

PR #3095, opened 2017-09-09
Simplify CI

allanrenucci:


allanrenucci:
It turns out sbt-scripted tests fail if we don't run `ln -s /var/cache/drone/ivy2 "$HOME/.ivy2"` before.

PR #3105, opened 2017-09-12
Update sbt to 0.13.16

allanrenucci:
We should also update the Docker image


smarter:
We can't upgrade yet because of https://github.com/sbt/sbt/issues/3460

PR #3109, opened 2017-09-12
Fix #3108: avoid generating useless scripts in distribution

### LINKED ISSUE #3108: distribution now generates useless files
`sbt dist-bootstrapped/pack` will generate `from-tasty`, `path-resolver`, `main` and `bench`, which are useless.

liufengyun:
Fix #3108: avoid generating useless scripts in distribution

The updated version of `sbt-pack` by default generates scripts for all main methods, thus an explicit setting is required.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -1126,6 +1126,7 @@ object Build {
   }
 
   lazy val commonDistSettings = packSettings ++ Seq(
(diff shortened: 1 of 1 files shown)


liufengyun:
Thanks @allanrenucci !

Issue #3111, opened 2017-09-13
launchIDE fails with missing vscode exports


esarbe:
`launchIDE` fails with missing vscode exports

```shell
> launchIDE
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list

> dotty@0.1.1 update-all /home/esabe/source/dotty/vscode-dotty
> npm install && node ./node_modules/vscode/bin/install && curl -L -o out/coursier https://github.com/coursier/coursier/raw/v1.0.0-RC3/coursier

Detected VS Code engine version: ^1.12.0
Found minimal version that qualifies engine range: 1.12.0
Fetching vscode.d.ts from: https://raw.githubusercontent.com/Microsoft/vscode/72672be0b7d3eef0784077b880615f91b7ec85aa/src/vs/vscode.d.ts
vscode.d.ts successfully installed!

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   137  100   137    0     0     90      0  0:00:01  0:00:01 --:--:--    90
100 11814  100 11814    0     0   7113      0  0:00:01  0:00:01 --:--:--  7113

4 import { TextDocumentChangeEvent, TextDocument, Disposable, OutputChannel, FileSystemWatcher as VFileSystemWatcher, DiagnosticCollection, ProviderResult, CancellationToken, Position as VPosition, Location as VLocation, Range as VRange, CompletionItem as VCompletionItem, CompletionList as VCompletionList, SignatureHelp as VSignatureHelp, Definition as VDefinition, DocumentHighlight as VDocumentHighlight, SymbolInformation as VSymbolInformation, CodeActionContext as VCodeActionContext, Command as VCommand, CodeLens as VCodeLens, FormattingOptions as VFormattingOptions, TextEdit as VTextEdit, WorkspaceEdit as VWorkspaceEdit, Hover as VHover, DocumentLink as VDocumentLink, TextDocumentWillSaveEvent, WorkspaceFolder as VWorkspaceFolder } from 'vscode';
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   ~~~~~~~~~~~~~~~

vscode-dotty/node_modules/vscode-languageclient/lib/client.d.ts(4,706): error TS2305: Module ''vscode'' has no exported member 'WorkspaceFolder'.


[...]

PR #3112, opened 2017-09-13
#3111: upgrade vscode engine requirement

esarbe:
upgrades vscode engine requirement so `launchIDE` will work

### DIFF
--- a/vscode-dotty/package.json
+++ b/vscode-dotty/package.json
@@ -11,7 +11,7 @@
   },
   "icon": "images/dotty-logo.svg",
   "engines": {
-    "vscode": "^1.12.0"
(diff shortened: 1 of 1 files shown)


felixmulder:
Thanks for the PR!

Let's get @smarter to review this :)

smarter:
Thanks for looking into this! I'd like to avoid upgrading the vscode required version if possible, so I created an alternative fix: https://github.com/lampepfl/dotty/pull/3131. Could you check if it works for you? (You'll probably want to first delete `vscode-dotty/node_modules` to get back to a reproducible state).

esarbe:
Nice! I'll have a look.

PR #3115, opened 2017-09-13
Upgrade backend

nicolasstucki:


nicolasstucki:
Currently testing that https://github.com/lampepfl/scala/pull/23 works

PR #3117, opened 2017-09-13
Experiment: cache symHelpers

nicolasstucki:


nicolasstucki:
test performance please

smarter:
A better approach I think would be to make SymHelper a value class.

smarter:
Ah, but I guess that won't help in the backend where we use the abstract `SymbolHelper`.

nicolasstucki:
This change reduces the allocations of `symHelper` from 14373196 to 320878 when doing `sbt ;dotty-bootstrapped/clean;dotty-bootstrapped/compile`.

nicolasstucki:
All those objects were probably GCd in the young generation.

nicolasstucki:
It did not show any improvement on performance

Issue #3071, opened 2017-09-05
IOException when output directory is missing


allanrenucci:
```scala
> dotc -d out Foo.scala
java.io.IOException: No such file or directory
	at java.io.UnixFileSystem.createFileExclusively(Native Method)
	at java.io.File.createNewFile(File.java:1012)
	at dotty.tools.io.AbstractFile.fileOrSubdirectoryNamed(AbstractFile.scala:236)
	at dotty.tools.io.AbstractFile.fileNamed(AbstractFile.scala:247)
	at scala.tools.nsc.backend.jvm.BytecodeWriters.getFile(BytecodeWriters.scala:35)
	at scala.tools.nsc.backend.jvm.BCodeHelpers.getFileForClassfile(BCodeHelpers.scala:38)
	at dotty.tools.backend.jvm.GenBCodePipeline$Worker1.visit$$anonfun$1(GenBCode.scala:217)
	at scala.Option.foreach(Option.scala:257)
	at dotty.tools.backend.jvm.GenBCodePipeline$Worker1.visit(GenBCode.scala:219)
	at dotty.tools.backend.jvm.GenBCodePipeline$Worker1.run(GenBCode.scala:152)
	at dotty.tools.backend.jvm.GenBCodePipeline.buildAndSendToDisk(GenBCode.scala:367)
	at dotty.tools.backend.jvm.GenBCodePipeline.run(GenBCode.scala:333)
	at dotty.tools.backend.jvm.GenBCode.run(GenBCode.scala:54)
	at dotty.tools.dotc.core.Phases$Phase.runOn$$anonfun$2(Phases.scala:283)
	at scala.collection.immutable.List.map(List.scala:284)
	at dotty.tools.dotc.core.Phases$Phase.runOn(Phases.scala:285)
	at dotty.tools.dotc.Run.compileUnits$$anonfun$1$$anonfun$1(Run.scala:81)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:18)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
	at dotty.tools.dotc.Run.compileUnits$$anonfun$1(Run.scala:89)
[...]

PR #3120, opened 2017-09-13
#3071: create new directory setting and use for '-d' parameter

esarbe:
also make sure that directory exists when updating setting

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -18,7 +18,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val javaextdirs = PathSetting("-javaextdirs", "Override java extdirs classpath.", Defaults.javaExtDirs)
   val sourcepath = PathSetting("-sourcepath", "Specify location(s) of source files.", "") // Defaults.scalaSourcePath
   val classpath = PathSetting("-classpath", "Specify where to find user class files.", defaultClasspath) withAbbreviation "-cp"
-  val d = DirectorySetting("-d", "directory|jar", "destination for generated classfiles.", Directory(Path(".")))
+  val outputDir = DirectorySetting("-d", "directory|jar", "destination for generated classfiles.", Directory(Path(".")))
   val priorityclasspath = PathSetting("-priorityclasspath", "class path that takes precedence over all other paths (or testing only)", "")
 
   /** Other settings */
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -158,6 +161,10 @@ object Settings {
             case Success(v) => update(v, args)
             case Failure(ex) => fail(ex.getMessage, args)
           }
+        case (DirectoryTag, arg :: args) =>
+          val path = Path(arg)
+          if (path.isDirectory) update(arg, args)
```
allanrenucci: Here, it should be `update(path.toDirectory, args)`

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -17,7 +18,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val javaextdirs = PathSetting("-javaextdirs", "Override java extdirs classpath.", Defaults.javaExtDirs)
   val sourcepath = PathSetting("-sourcepath", "Specify location(s) of source files.", "") // Defaults.scalaSourcePath
   val classpath = PathSetting("-classpath", "Specify where to find user class files.", defaultClasspath) withAbbreviation "-cp"
-  val d = StringSetting("-d", "directory|jar", "destination for generated classfiles.", ".")
+  val d = DirectorySetting("-d", "directory|jar", "destination for generated classfiles.", Directory(Path(".")))
```
allanrenucci: We should probably rename this setting to something like `outputDir`

esarbe:
I've had to touch up the tests so that the output directory actually exists.

allanrenucci:
Could you add a test under `compiler/test/dotty/tool/dotc/SettingTests.scala` and do something like:
```scala
class SettingTests {
  ...

  @Test def missingOutputDir: Unit = {
    val options = Array("-d", "not_here")
    val reporter = Main.process(options)
    assertEquals(reporter.errorCount, 1)
    assertEquals(reporter.allErrors.head.message, "'not_here' does not exist or is not a directory")
  }
}
```

esarbe:
Sure thing!

smarter:
Nitpick: To automatically close the issue when a PR is merged, you should put "Fix #nnnn" in the commit message, just putting "#nnnn" is not enough.

esarbe:
Thanks for the nitpicking! I've updated the last commit comment, I hope that suffices.

PR #3124, opened 2017-09-14
Revert #3059: which generates non-bootstrapped for bootstrapped

liufengyun:
Revert #3059: which generates non-bootstrapped for bootstrapped.

There's some bug with the latest version of the sbt-pack plugin.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -8,8 +8,7 @@ import java.util.Calendar
 
 import scala.reflect.io.Path
 import sbtassembly.AssemblyKeys.assembly
-import xerial.sbt.pack.PackPlugin._
-import autoImport._
+import xerial.sbt.Pack._
 
 import sbt.Package.ManifestAttributes
 
@@ -1128,7 +1127,6 @@ object Build {
   lazy val commonDistSettings = packSettings ++ Seq(
     packMain := Map(),
     publishArtifact := false,
-    packGenerateMakefile := false,
     packExpandedClasspath := true,
     packResourceDir += (baseDirectory.value / "bin" -> "bin"),
     packArchiveName := "dotty-" + dottyVersion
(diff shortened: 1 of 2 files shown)


allanrenucci:
What is the reason for reverting this change? We will need a new docker image for the CI because this PR change a plugin version

nicolasstucki:
The generated jar contained the non-bootstrapped version :(

liufengyun:
@allanrenucci is it possible to have some mapping of type system between the host OS and the docker, so that we don't need to build a new docker image when we change plugins or libraries.

liufengyun:
test performance please

nicolasstucki:
test performance please

liufengyun:
Now we have the setting `measure "tests/implicitNums.scala" 100 tests/implicitNums.scala`.

test performance please

smarter:
Exhaustivity T and V drop 200ms of runtime; I wonder why those tests in particular show such a big difference.

liufengyun:
T and V are quite noisy, maybe because we need more warmups. This test is intended to check if more warmups can bring down the number of mini-tests. We set 100 warmups for `implicit cache II`, but it seems to have no effect -- the result is the same as 10 warmups + 20 runs.

retronym:
JMH has a handy profiler, `-prof hs_comp`, that logs JIT compiler activity after each iteration. Observing the output after each measurement iteration to ensure that the overall time spent compiling is no longer rapidly growing helps to validate the chosen warmup duration.

You can disable your warmup altogether to see the output of this profiler periodically after each iteration.

Here's an example of 200s of execution of of a scalac benchmark: https://gist.github.com/109d3b6a3b72c9d87967f62985b78df9

Notice that in steady state, the delta of `compiler.totalTime` is low (~0.5s compile time per 10s measurement iteration), and the delta of `compiler.totalCompiles` is low (<100 methods).

In addition to selecting a warmup duration, selecting the number of forks is important. We tend to use 3, which I think is even too low and probably our biggest cause of jitter.

nicolasstucki:
Thanks @retronym I will have a look.

liufengyun:
test performance please

PR #3125, opened 2017-09-14
Do not generate equals method for case object

nicolasstucki:
Found this while inspecting the bytecode of dotty bootstrapped.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
+++ b/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
@@ -41,16 +41,19 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
 
   private var myValueSymbols: List[Symbol] = Nil
   private var myCaseSymbols: List[Symbol] = Nil
+  private var myCaseModuleSymbols: List[Symbol] = Nil
 
   private def initSymbols(implicit ctx: Context) =
     if (myValueSymbols.isEmpty) {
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -41,16 +41,19 @@ class SyntheticMethods(thisTransformer: DenotTransformer) {
...
   private var myValueSymbols: List[Symbol] = Nil
   private var myCaseSymbols: List[Symbol] = Nil
+  private var myCaseModuleSymbols: List[Symbol] = Nil
 
   private def initSymbols(implicit ctx: Context) =
     if (myValueSymbols.isEmpty) {
       myValueSymbols = List(defn.Any_hashCode, defn.Any_equals)
       myCaseSymbols = myValueSymbols ++ List(defn.Any_toString, defn.Product_canEqual,
         defn.Product_productArity, defn.Product_productPrefix, defn.Product_productElement)
+      myCaseModuleSymbols = myCaseSymbols.filter(_ ne defn.Any_equals)
```
OlivierBlanvillain: I would just do `myValueSymbols = defn.Any_hashCode :: (if (clazz.is(Module)) defn.Any_equals :: Nil else Nil)` instead

nicolasstucki: That can't be done here, this is precomputing the lists for any `clazz`. We do not have access to the `clazz` in `initSymbols`

PR #3127, opened 2017-09-15
Partially fix #2806

### LINKED ISSUE #2806: Latest version of dotty should be written down in a fixed place
sbt-dotty currently has some complicated logic to try to guess the latest nightly build of dotty based on maven xml, but it's hardcoded to `dotty_0.2` (and was previously hardcoded to `dotty_0.1`). Users that want to  know what the latest version of dotty is also have to look at maven and scroll through a bunch of versions. It would be easier for everyone if we had something like:
Fetching http://dotty.epfl.ch/last-stable returns the string "0.1.2-RC1" (until we have a real stable release).
Fetching http://dotty.epfl.ch/last-rc returns the string "0.1.2-RC1".
Fetching http://dotty.epfl.ch/last-nightly returns the string "0.2.0-bin-20170622-7381e9d-NIGHTLY".

This could be done by the CI when the site is generated.

allanrenucci:
Write latest nightly version in file when generating documentation

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -34,21 +34,20 @@ object ExposedValues extends AutoPlugin {
 }
 
 object Build {
+  val baseVersion = "0.4.0"
+  def nightlyVersion =
+    baseVersion + "-bin-" + VersionUtil.commitDate + "-" + VersionUtil.gitHash + "-NIGHTLY"
 
   val scalacVersion = "2.12.3"
 
   val dottyOrganization = "ch.epfl.lamp"
   val dottyGithubUrl = "https://github.com/lampepfl/dotty"
   val dottyVersion = {
-    val baseVersion = "0.4.0"
     val isNightly = sys.env.get("NIGHTLYBUILD") == Some("yes")
     val isRelease = sys.env.get("RELEASEBUILD") == Some("yes")
-    if (isNightly)
-      baseVersion + "-bin-" + VersionUtil.commitDate + "-" + VersionUtil.gitHash + "-NIGHTLY"
-    else if (isRelease)
-      baseVersion
-    else
-      baseVersion + "-bin-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


project/Build.scala
```diff
@@ -316,6 +315,8 @@ object Build {
     parallelExecution in Test := false,
 
     genDocs := Def.inputTaskDyn {
+      IO.write(file("./docs/_site/last-nightly"), nightlyVersion)
```
olafurpg: "latest-nightly"?

olafurpg: Are the docs only updated on nightly release? Will there always be a corresponding published version on sonatype?

allanrenucci:
Open question: Do we want a more general solution that also exposes "latest stable" and "latest RC"?

project/Build.scala
allanrenucci: Hum, you're right this is wrong. The doc is updated on every PR merged on master. So the version will not be published on sonatype yet. @smarter Should we publish doc only on nightly?

felixmulder: You guys are coming to all the right conclusions :tada: 

It indeed needs to be the latest nightly version published. Maybe maven has a query API for this?

olafurpg: The latest base version would do for my needs, we could use that to query maven-metadata.xml for the latest release here https://github.com/lampepfl/dotty/blob/58ca93df0caadd238e7450ae60f8044ecc0ad9ef/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L17

smarter:
The data we publish should reflect what is actually available on maven, so we need to do something like call `dottyLatestNightlyBuild`

> Open question: Do we want a more general solution that also exposes "latest stable" and "latest RC"?

Yes. You could get this information from git tags or more simply just have a text file somewhere that is manually updated.

PR #3133, opened 2017-09-17
add dummy fields to trees to test cache locality

liufengyun:
add dummy fields to trees to test cache locality

PR #3137, opened 2017-09-17
Empty PR to test the benchmarks again

smarter:
I'm suspicious.


smarter:
test performance please

liufengyun:
This morning I changed the forks and warmups for some noisy lines (mainly
Vector and exhaustivity check tests), that might explain some of the
changes.

Le 17 sept. 2017 5:47 PM, "Guillaume Martres" <notifications@github.com> a
écrit :

smarter:
I think the bot decided to run the benchmarks again because you quoted the command :).

liufengyun:
Thanks for the catch @smarter . The duplicate job removed and system updated. Now commands have to start from a line, so quoted commands are excluded.

smarter:
Looks like the benchmark changes made a big difference, invalidating the results of https://github.com/lampepfl/dotty/pull/3133 and https://github.com/lampepfl/dotty/pull/3136. @liufengyun I suggest reverting the changes for now. In the future I guess we'll need to rerun the benchmarks for all points when we make changes like this.

liufengyun:
@smarter I've temporarily reverted the changes in https://github.com/liufengyun/bench/commit/19cd60cdf8f7819ea418a0cc0a92eca9443569e5 on the server, which affects 5 lines. Now you can try again.

Note that re-run all benchmarks is both computationally expensive and laborious, thus re-run is only justified when there's a compelling reason.

1. Each point takes about 1 hour
1. The dependencies/structure of the compiler is changing, which requires updating the infrastructure to fit the compiler every time it changes.

smarter:
test performance please

smarter:
This looks more reasonable to me.

PR #3138, opened 2017-09-17
Java 9 support

smarter:


smarter:
sbt 0.13 cannot be used with Java 9 because it tries to compile the project build with Scala 2.10 which really does not like Java 9, so we'll probably have to wait until we switch to sbt 1 to test this properly. But it seems to work:
```scala
$ JAVA_HOME=~/opt/openjdk-9-181 ./bin/dotr
scala> java.util.List.of(1,2) 
val res0: java.util.List[Int] = [1, 2]
```

retronym:
The lastest sbt launcher script extracts rt.jar when starting on java 9
which lets scalac 2.10 work

smarter:
@retronym I gave it a try but got an assert in scala.reflect: https://gist.github.com/smarter/67ee91680bc3cc346fb7500ee023112f am I doing something wrong?

smarter:
The ExposedValues mentioned in the stacktrace is [this thing](https://github.com/lampepfl/dotty/blob/657385ba36eb138f1f78413311b7c108d3dc8ac2/project/Build.scala#L26-L34) stolen from Scala.js to avoid having to put everything into build.sbt

retronym:
Move your autoplugin out of the empty package: https://github.com/scala/scala-dev/issues/304

smarter:
Nice! Thanks :). I now have a bootstrapped dotty on Java 9. My sbt ExtractDependencies phase warned me I wasn't handling `PlainNioFile` which is also something that is missing from zinc as far as I can tell: https://github.com/sbt/zinc/blob/1.x/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala#L113-L123

retronym:
Good catch, I've added a todo item to https://github.com/scala/scala-dev/issues/139

smarter:
>  I now have a bootstrapped dotty on Java 9.

I spoke too soon. the EA 181 build from http://jdk.java.net/9/ works, but a nightly from https://builds.shipilev.net/openjdk-jdk9-dev-release/ doesn't. It crashes with:
```scala
Exception in thread "main" java.lang.BootstrapMethodError: call site initialization exception
        at java.base/java.lang.invoke.CallSite.makeSite(CallSite.java:385)
        at java.base/java.lang.invoke.MethodHandleNatives.linkCallSiteImpl(MethodHandleNatives.java:250)
        at java.base/java.lang.invoke.MethodHandleNatives.linkCallSite(MethodHandleNatives.java:240)
        at dotty.runtime.LazyVals$.<init>(LazyVals.scala:87)
        at dotty.runtime.LazyVals$.<clinit>(LazyVals.scala)
        at dotty.tools.dotc.core.Contexts$ContextBase.<clinit>(Contexts.scala:542)
        at dotty.tools.dotc.Driver.initCtx(Driver.scala:36)
        at dotty.tools.dotc.Driver.process(Driver.scala:88)
        at dotty.tools.dotc.Driver.process(Driver.scala:105)
        at dotty.tools.dotc.Driver.main(Driver.scala:132)
        at dotty.tools.dotc.Main.main(Main.scala)
Caused by: java.lang.invoke.LambdaConversionException: Type mismatch for instantiated parameter 0: int is not a subtype of class java.lang.Object
        at java.base/java.lang.invoke.AbstractValidatingLambdaMetafactory.checkDescriptor(AbstractValidatingLambdaMetafactory.java:308)
        at java.base/java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:294)
        at java.base/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:311)
        at java.base/java.lang.invoke.CallSite.makeSite(CallSite.java:332)
        ... 10 more
```
The corresponding LMF bootstrap method is:
```scala
  3: #68 invokestatic java/lang/invoke/LambdaMetafactory.metafactory:(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/Met
    Method arguments:
      #70 (Ljava/lang/Object;)Ljava/lang/Object;
      #153 invokespecial dotty/runtime/LazyVals$.$init$$$anonfun$4:(I)Ljava/lang/Object;
      #154 (I)Ljava/lang/Object;
```
which indeed looks fishy. The check was added in http://hg.openjdk.java.net/jdk9/jdk9/jdk/rev/4be3ef759ead but the referenced bug report appears to be private.

smarter:
> I spoke too soon. the EA 181 build from http://jdk.java.net/9/ works, but a nightly from https://builds.shipilev.net/openjdk-jdk9-dev-release/ doesn't.

My bad, EA 181 doesn't work either, I messed up and was accidentally running an old jdk9 build.

Blaisorblade:
FWIW we’re now on sbt 1 which might help.

Glavo:
Hmmm, what is the status of this PR? Is it frozen? Now Dotty is my only obstruction to using Java 10 by default.

project/Build.scala
```diff
@@ -578,9 +578,15 @@ object Build {
...
-          "-Ddotty.tests.classes.compiler=" + jars("dotty-compiler")
+          "-Ddotty.tests.classes.dottyInterfaces=" + jars("dotty-interfaces"),
+          "-Ddotty.tests.classes.dottyLibrary=" + jars("dotty-library"),
+          "-Ddotty.tests.classes.dottyCompiler=" + jars("dotty-compiler"),
+          "-Ddotty.tests.classes.compilerInterface=" + findLib(attList, "compiler-interface"),
+          "-Ddotty.tests.classes.scalaLibrary=" + findLib(attList, "scala-library"),
+          "-Ddotty.tests.classes.scalaAsm=" + findLib(attList, "scala-asm"),
+          "-Ddotty.tests.classes.scalaXml=" + findLib(attList, "scala-xml"),
+          "-Ddotty.tests.classes.jlineTerminal=" + findLib(attList, "jline-terminal"),
+          "-Ddotty.tests.classes.jlineReader=" + findLib(attList, "jline-reader")
```
allanrenucci: We also use `BuildInfoPlugin` to achieve the same thing. Should we pick one over the other?

smarter: No preference, but let's not change this now.

smarter:
This is finally ready for review ;)

compiler/test/dotty/tools/vulpix/TestConfiguration.scala
```diff
@@ -10,47 +10,44 @@ object TestConfiguration {
...
-            |separated list of jars.
-            |
-            |When compiling with eclipse, you need the sbt-interfaces jar, put
-            |it in extras."""
-      )
+      assert(file.exists, s"File $p couldn't be found.")
       file.getAbsolutePath
-    } mkString(":")
-  }
+    }).mkString(":")
```
allanrenucci: We should use `File.pathSeparator`

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -204,8 +204,9 @@ class CompilationTests extends ParallelTesting {
       defaultOutputDir + libGroup + "/src/:" +
       // as well as bootstrapped compiler:
       defaultOutputDir + dotty1Group + "/dotty/:" +
-      // and the other compiler dependecies:
-      Jars.dottyInterfaces + ":" + Jars.jlineTerminal + ":" + Jars.jlineReader,
+      // and the other compiler dependenies:
+      Properties.compilerInterface + ":" + Properties.scalaLibrary + ":" + Properties.scalaAsm + ":" +
+      Properties.dottyInterfaces + ":" + Properties.jlineTerminal + ":" + Properties.jlineReader,
```
allanrenucci: Let's use `TestConfiguration.mkClasspath`?

smarter:
> How do we test this? Should we use Java 9 on the CI instead of Java 8?

No because we still want to support Java 8, instead ideally we should run the tests both with Java 8 and with Java 10 (and replace 10 by 11 when it's released).

allanrenucci:
Ok. I guess I'll set this up as a nightly job then. We don't want to double CI time.

compiler/test/dotty/tools/dotc/CompilationTests.scala
smarter: mkClasspath checks that each directory exists so won't work on libGroup and dotty1Group here. I agree this code would be cleaned up more but let's keep that for another pr.

compiler/test/dotty/tools/vulpix/TestConfiguration.scala
smarter: Done.

compiler/test/dotty/tools/dotc/CompilationTests.scala
allanrenucci: Ok. Then we should use `File.pathSeparator` here as well

smarter: There's a lot of place where we use ":" right now which need to be fixed I've opened an issue for it: https://github.com/lampepfl/dotty/issues/5000

Blaisorblade:
Bootstrap started failing since the merge with `FileAlreadyExistsException` http://dotty-ci.epfl.ch/lampepfl/dotty/7060/5 @smarter @allanrenucci

PR #3143, opened 2017-09-19
Update ReTyper to Ycheck patterns & inline fixes

OlivierBlanvillain:
This PR does minor adjustments to Typer and ReTyper such that patterns pass Ycheck. This is required to be able to inline pattern matching expressions.


compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -512,8 +512,8 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
       val (meth, targs, argss) = decomposeCall(fn)
       (meth, targs, argss :+ args)
     case TypeApply(fn, targs) =>
-      val (meth, Nil, Nil) = decomposeCall(fn)
-      (meth, targs, Nil)
+      val (meth, targss, args) = decomposeCall(fn)
```
odersky: I don't understand the change. Can you give an example where this matters?

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -962,9 +962,10 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
 
         val dummyArg = dummyTreeOfType(ownType)
         val unapplyApp = typedExpr(untpd.TypedSplice(Apply(unapplyFn, dummyArg :: Nil)))
-        val unapplyImplicits = unapplyApp match {
+        def unapplyImplicits(unapp: Tree = unapplyApp): List[Tree] = unapp match {
           case Apply(Apply(unapply, `dummyArg` :: Nil), args2) => assert(args2.nonEmpty); args2
```
odersky: I think it's safer in that case to not have a default argument for `unapp`.

compiler/src/dotty/tools/dotc/typer/ReTyper.scala
```diff
@@ -48,11 +52,22 @@ class ReTyper extends Typer {
   override def typedSuper(tree: untpd.Super, pt: Type)(implicit ctx: Context): Tree =
     promote(tree)
 
+  override def typedTyped(tree: untpd.Typed, pt: Type)(implicit ctx: Context): Tree = {
+    assertTyped(tree)
+    val type1 = checkSimpleKinded(typedType(tree.tpt))
```
odersky: Rename `type1 -> tpt1`, `tree1 -> expr1` to conform to naming conventions used elsewhere. (`tpt` stands for "type tree")

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
OlivierBlanvillain: The test added in the same commit (e199fc838e8c437c6cf88b7f27bbb6f350ed3bb2) breaks with a match error before this change

OlivierBlanvillain:
Completely forgot about this PR :sweat_smile:. I've addressed the review and rebased, so it should be good to go!

tests/pos/inline-t9232b.scala
```diff
@@ -0,0 +1,23 @@
+final class Foo(val value: Int)
+
+object Foo {
+  inline def unapplySeq(foo: Foo): Some[Seq[Int]] = Some(List(foo.value))
```
allanrenucci: Should this be changed to `inline def unapply(foo: Foo)` after #3747?

OlivierBlanvillain: I would say not for now, if/when `unapplySeq` is effectively not supported in Dotty then we should turn this tests into a scala2 test.

PR #3134, opened 2017-09-17
[WIP] Change references to private fields

odersky:

PR #3087, opened 2017-09-08
Generalize RefTypes

odersky:
Based on #3061.

The aim of this PR is to eliminate the various subtypes of TermRef and TypeRef and replace them by generalizing Name Name/Signature and Symbol to a new concept of `Designator`. 

The expected benefits of doing this are:

 - Drop the kludgey way we deal with "shadowed names" by making all private references symbolic
 - Allow to eliminate TypeArgRefs

PR #3147, opened 2017-09-21
WIP: Change namedtypes

odersky:
This is a second attempt to simplify named types. The first attempt ran into lots of complications when trying to accommodate private members. I decided it's better to start from scratch. 

The new attempt tries to keep the basic layout of a NamedType as a combination of prefix and name, without fiddling with designators. Instead we want to base more functionality on the symbol in the type.


odersky:
This PR turned out to be a dead-end, and the approach in #3087 and #3134 is preferable.

The issue is that, as long as we base our tooling on Tasty we cannot define references by their symbols, since these references can go across compilation units. So we need the full distinction between references as names, names with signatures and/or namespaces, and symbols. There's no way around it. We just have to make it as streamlined as we can.

Issue #2239, opened 2017-04-12
Inferred type refers to a local TermRef


liufengyun:
The following code compiles fine with Scalac, but fails in Dotty. If the commented version of `~++` is used, everything is fine.

```Scala
trait Rule[-In, +A] extends (In => A) {
  def flatMap[In2 <: In, B](fa2ruleb: A => Rule[In2, Seq[B]]): Rule[In, Seq[B]] = ???
  def map[B](fa2b: A => B): Rule[In, B] = ???

  def ~++[In2, B >: A](next: => Rule[In2, Seq[B]]) = for (a <- this; b <- next) yield a :: b.toList
  // def ~++[In2, B >: A](next: => Rule[In2, Seq[B]]): Rule[In, Seq[B]] = for (a <- this; b <- next) yield a :: b.toList
}

class SeqRule {
  type S
  type A
  def * : Rule[S, List[A]] = ???
  def +(rule: Rule[S, A]) : Rule[S, Seq[A]] = rule ~++ *
}
```

The cause is that `~++` gets an "ill" type that refers to the local variable `b` (see below at bottom). The check is not handled by `Typer#escapingRefs`. I tried to modify `TypeOps#simplify` to simplify `TypeRef(TermRef(NoPrefix,b),scala$collection$TraversableOnce$$A)` to just `ParamRef(B)`, but didn't succeed.

```Scala
PolyType(
  List(In2, B),

  List(TypeBounds(TypeRef(ThisType(TypeRef(NoPrefix,scala)),Nothing), TypeRef(ThisType(TypeRef(NoPrefix,scala)),Any)), TypeBounds(TypeRef(TermRef(NoPrefix,rule),Rule$$A), TypeRef(ThisType(TypeRef(NoPrefix,scala)),Any))),

  MethodType(
    List(next),
    List(
      ExprType(
        RefinedType(
          RefinedType(
            TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Rule),
            Rule$$In,
            TypeAlias(ParamRef(In2), -1)
          ),

          Rule$$A,
          TypeAlias(
            RefinedType(
              TypeRef(ThisType(TypeRef(NoPrefix,collection)),Seq),
              scala$collection$Seq$$A,
              TypeAlias(ParamRef(B), 1)
            ),
          1))
      )
    ),

    RefinedType(
      RefinedType(
        TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),Rule),
[...]

PR #3148, opened 2017-09-21
Add test case for #2239

liufengyun:
Add test case for #2239, which is fixed after the native apply refactoring #3061


liufengyun:
Thanks @allanrenucci !

PR #3151, opened 2017-09-21
Erase all phantom expressions

nicolasstucki:
Erase all calls to methods returning a phantom and their definition.


odersky:
Do we still pursue this? I thought we'd switch to `unused`.

nicolasstucki:
I would say yes, because it brings the semantics of phantoms closer to the ones of  `unused`.  Which will allow to migrate the tests of phantom to unused. I do not feel this is a priority, but It might be useful nevertheless.

odersky:
I would advocate to start from scratch with `unused` instead. There are few things that are worse than layers of complexity where you don't know why they are there and then it turns out that it was because of some feature that's no longer used. That brings a general sense of insecurity to anyone tying to understand the codebase. So, clarity first! Tests can and should be re-integrated when `unused` is functional.

PR #3153, opened 2017-09-21
Make sbt scripted tests use the right ivy HOME

allanrenucci:
When running the CI, ivy HOME is not the default one but sbt scripted
tests used to use the default one anyway


allanrenucci:
scripted tests ran here: http://dotty-ci.epfl.ch/lampepfl/dotty/333

PR #3156, opened 2017-09-21
Fix #3144: emit warnings for unchecked type patterns

### LINKED ISSUE #3144: Pattern matching with type annotations allows impossible cases
This should not compile:

```scala
  sealed trait Foo[T]
  case class Bar[T](s: String)
  
  def shouldError[T](foo: Foo[T]): String = 
    foo match {
      case bar: Bar[T] => bar.s
    }
```

Not even a warning is produced, as is.

liufengyun:
Fix #3144: emit warnings for unchecked type patterns

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -880,6 +880,16 @@ object messages {
            |"""
   }
 
+  case class UncheckedTypePattern(msg: String)(implicit ctx: Context)
+    extends Message(UncheckedTypePatternID) {
+    val kind = "Unchecked Type Pattern"
+
+    val explanation =
+      hl"""|Type arguments and type refinements are erased during compile time, thus it's
+           |impossible to check them at run-time.
+           |"""
+  }
+
   case class MatchCaseUnreachable()(implicit ctx: Context)
   extends Message(MatchCaseUnreachableID) {
     val kind = s"""Match ${hl"case"} Unreachable"""
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -882,11 +882,13 @@ object messages {
 
   case class UncheckedTypePattern(msg: String)(implicit ctx: Context)
     extends Message(UncheckedTypePatternID) {
-    val kind = "Unchecked Type Pattern"
+    val kind = "Pattern Match Exhaustivity"
 
     val explanation =
       hl"""|Type arguments and type refinements are erased during compile time, thus it's
            |impossible to check them at run-time.
+           |
+           |You can either replace the type arguments by `_` or use `@unchecked`.
            |"""
   }
 
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -417,7 +417,11 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     case tp @ AppliedType(tycon, args) =>
       if (tycon.isRef(defn.ArrayClass)) tp.derivedAppliedType(tycon, args.map(erase))
       else {
-        val ignoreWarning = args.forall(p => p.typeSymbol.is(BindDefinedType) || p.isInstanceOf[TypeBounds])
+        val ignoreWarning = args.forall { p =>
+          p.typeSymbol.is(BindDefinedType) ||
+            p.hasAnnotation(defn.UncheckedAnnot) ||
+            p.isInstanceOf[TypeBounds]
+        }
         if (!ignoreWarning)
           warn("type arguments are not checked since they are eliminated by erasure")
 
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -7,7 +7,7 @@ import Types._
 import Contexts._
 import Flags._
 import ast.Trees._
-import ast.{tpd, untpd}
+import ast.tpd
 import Decorators._
 import Symbols._
 import StdNames._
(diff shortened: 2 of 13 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -403,21 +404,30 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
   /* Erase a type binding according to erasure semantics in pattern matching */
-  def erase(tp: Type): Type = tp match {
+  def erase(tp: Type)(implicit pos: Position): Type = tp match {
     case tp @ AppliedType(tycon, args) =>
       if (tycon.isRef(defn.ArrayClass)) tp.derivedAppliedType(tycon, args.map(erase))
-      else tp.derivedAppliedType(tycon, args.map(t => WildcardType))
+      else {
+        val ignoreWarning = args.forall(p => p.typeSymbol.is(BindDefinedType) || p.isInstanceOf[TypeBounds])
+        if (!ignoreWarning)
+          ctx.warning(UncheckedTypePattern("type arguments are not unchecked since they are eliminated by erasure"), pos)
```
smarter: I think you meant to write "are not checked"

allanrenucci:
I think there are two separate issues described in #3144:
- Missing unchecked warning for type pattern that are eliminated by erasure. This seems to be fixed in this PR. Do we want to support the `@unchecked` annotation like scalac?
```scala
class Test {
  def f(x: Any): Int = x match {
    case xs: List[Int] @unchecked => xs.head
  }
}
```
- Unreachable case for unrelated type when one the type is a sealed trait. Maybe I am missing something but the case seems unreachable to me (scalac doesn't emit a warning either).
```scala
sealed trait Foo
class Bar
  
class Test {
  def shouldError(foo: Foo) = foo match {
    case _: Bar => 1
  }
}
```

liufengyun:
@allanrenucci The second is not an issue, as there could be a class like the follows:

```Scala
class Ruu extends Bar with Foo
```

allanrenucci:
@liufengyun But Foo is `sealed`, so we know there is no such class `Ruu`, right?

liufengyun:
Ah I see, currently the algorithm just assumes the first case is always reachable to save some computation. We could also check if the first branch is reachable or not, but I'm not sure if it's worth it.

BTW, we've checks for hopeless matches, which are located in `IsInstanceOfEvaluator`.

allanrenucci:
>Ah I see, currently the algorithm just assumes the first case is always reachable to save some computation.

If `Foo` is a `class` dotc gives me an error for the first case though: 
```scala
-- Error: tests/allan/Test.scala:6:9 -------------------------------------------
6 |    case _: Foo => 1
  |         ^
  |      this case is unreachable since class Bar and class Foo are unrelated

one error found
```

liufengyun:
@allanrenucci That check is from `Erasure`, not from the exhaustivity check. The Erasure handles it more generally by checking if `isInstanceOf` is sensical.

allanrenucci:
You might want to add a test case for
```scala
case xs: List[Int @unchecked] =>
```

tests/patmat/3144b.scala
```diff
@@ -1,6 +1,8 @@
 class Test {
   def f(x: Any): Int = x match {
     case xs: List[Int] @unchecked => xs.head
+    case xs: List[Int @unchecked] => xs.head
```
allanrenucci: I think you should extract this case into another function. This case is unreachable here

tests/patmat/3144b.scala
```diff
@@ -0,0 +1,8 @@
+class Test {
+  def f(x: Any): Int = x match {
+    case xs: List[Int] @unchecked => xs.head
+    case xs: List[Int @unchecked] => xs.head
```
allanrenucci: You should extract this case into another function. This case is unreachable here

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -880,6 +880,16 @@ object messages {
            |"""
   }
 
+  case class UncheckedTypePattern(msg: String)(implicit ctx: Context)
+    extends Message(UncheckedTypePatternID) {
+    val kind = "Unchecked Type Pattern"
```
smarter: This is a very specific kind, kinds are supposed to apply to more than one message, I guess here it could be "Pattern Match Exhaustivity" (kinds should also be an enum, but that's a separate issue).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -880,6 +880,16 @@ object messages {
...
+
+    val explanation =
+      hl"""|Type arguments and type refinements are erased during compile time, thus it's
+           |impossible to check them at run-time.
```
smarter: The explanation could mention possible fixes: either replace the type arguments by `_` or use `@unchecked`.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -403,21 +403,37 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      def warn(msg: String): Unit = if (!unchecked) ctx.warning(UncheckedTypePattern(msg), tpt.pos)
+      Typ(erase(expr.tpe.stripAnnots)(warn), true)
     case _ =>
       debug.println(s"unknown pattern: $pat")
       Empty
   }
 
   /* Erase a type binding according to erasure semantics in pattern matching */
-  def erase(tp: Type): Type = tp match {
+  def erase(tp: Type)(implicit warn: String => Unit): Type = tp match {
```
smarter: Instead of passing an implicit around, you could have an outer function which defines `warn` and an inner recursive function which uses it.

liufengyun: @smarter thanks a lot, I've addressed other comments.  For this one, doing the refactoring would require to pass an additional `Position` to `erase`, which will make the API a little ugly.

Issue #3155, opened 2017-09-21
Documentation is broken


allanrenucci:
#3061 broke the documentation: http://dotty-ci.epfl.ch/lampepfl/dotty/317/

allanrenucci:
The newly introduced `TypeArgRef ` must be handled in dottydoc

allanrenucci:
Fixed in #3157

PR #3157, opened 2017-09-22
Add missing case in factories

odersky:


odersky:
Merging now to unbreak the build

allanrenucci:
This fixes #3155. Can we add a test case for it?

allanrenucci:
@odersky Could you give me a code snippet that produces a `TypeArgRef` and used to break the documentation. I can make a test case out of it.

odersky:
@allanrenucci `TypeArgRefs` will probably go away soon, so I think it's not worth making a test case for them now.

PR #3158, opened 2017-09-22
Setup Drone cache

allanrenucci:


.drone.yml
```diff
@@ -100,5 +108,15 @@ pipeline:
...
 
+  rebuild-cache:
+    image: drillster/drone-volume-cache
+    rebuild: true
+    mount:
+      - /drone/cache/ivy2
+    volumes:
+      - /tmp/cache:/cache
+#    when:
+#     event: push
```
nicolasstucki: What is the default?

allanrenucci: Everything: [push, pull_request, tag, deployment]. I think we should switch to push only.
I commented it out to make sure I can build a cache and reuse a cache

PR #3159, opened 2017-09-22
Add regression tests on optimise

nicolasstucki:


OlivierBlanvillain:
Maybe we should create an issue to not forget about these pattern matching warnings?

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -186,6 +189,7 @@ class CompilationTests extends ParallelTesting {
 
   @Test def runAll: Unit = {
     compileFilesInDir("../tests/run", defaultOptions) +
+    compileFilesInDir("../tests/run", defaultOptimised) +
```
allanrenucci: If you run everything in `tests/run` with `defaultOptimised`, aren't the lines below redundant?

nicolasstucki: Yes, I will remove them. Thanks

allanrenucci:
I think this PR is the cause of the spurious failures than can be observed [here](http://dotty-ci.epfl.ch/lampepfl/dotty/363/6) and [here](http://dotty-ci.epfl.ch/lampepfl/dotty/364/5). I think we should revert it

nicolasstucki:
Then revert it, I will test again next week

PR #3164, opened 2017-09-25
Revert "Revert "Add regression tests on optimise""

nicolasstucki:
This reverts commit 1943f4e4985ebd1d9bbd5f06418fef0f2b79dd21.


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -298,6 +294,13 @@ class CompilationTests extends ParallelTesting {
     tests.foreach(_.delete())
   }
 
+  @Test def testOptimised: Unit = {
+    val outputDir = defaultOutputDir + "optimised/"
```
allanrenucci: `outputDir` is never used

nicolasstucki: supposed to be used in the tree lines bellow

nicolasstucki: Fixed

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -298,6 +294,13 @@ class CompilationTests extends ParallelTesting {
...
+    compileFilesInDir("../tests/pos", defaultOptimised, outputDir).checkCompile()
+    compileFilesInDir("../tests/run", defaultOptimised, outputDir).checkRuns()
+    compileShallowFilesInDir("../tests/neg", defaultOptimised, outputDir).checkExpectedErrors()
```
allanrenucci: Should we move each test in its corresponding test group (i.e. `compilePos`, `compileNeg`, `runAll`) to run them in parallel or should we keep them together for clarity?

nicolasstucki: I would prefer to keep the separated for now because I going to start modifying stuff in the local optimizations. It will be clearer to see where the failures occurred if they happen. There are enough tests in each one to take advantage of parallelism, we can improve it later.

PR #3167, opened 2017-09-25
More robust way to refer to private definitions in types

odersky:
Changes the way we represent references to private data.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Designators.scala
+++ b/compiler/src/dotty/tools/dotc/core/Designators.scala
@@ -44,6 +44,10 @@ object Designators {
   type TermDesignator = Designator { type ThisName = TermName }
   type TypeDesignator = Designator { type ThisName = TypeName }
 
+  /** Names that come with the namespace where they are defined.
+   *  Used to give a stable reference to private names, and also to
+   *  Scala 2x inner classes.
+   */
   case class LocalName[N <: Name](name: N, nameSpace: TypeRef) extends Designator {
     type ThisName = N
 
--- a/compiler/src/dotty/tools/dotc/core/Designators.scala
+++ b/compiler/src/dotty/tools/dotc/core/Designators.scala
@@ -3,6 +3,9 @@ package dotc
 package core
 
 import Names._
+import NameKinds.SignedName
+import Types.{TypeRef, NameSpace, noNameSpace}
+import Symbols.Symbol
 import Contexts.Context
 
 /** Defines a common superclass of Name and Symbol and its Term/Type variants
@@ -10,7 +13,7 @@ import Contexts.Context
  */
 object Designators {
 
-  abstract class Designator extends util.DotClass {
+  abstract class Designator extends util.DotClass { self =>
 
     type ThisName <: Name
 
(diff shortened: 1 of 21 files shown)


odersky:
Previously, we referred to an entity in a NamedType via a prefix type, a name, and (for terms) a signature. This is not enough to distinguish multiple private entities with the same prefix and name or a private definition and a public one with the same prefix and name. We dealt with this using the concept of a "ShadowedName", but it looked more like a patch than a solid fix. 

We now deal with the problem directly, by adding a nameSpace as a potential third element of a designator. A nameSpace is a TypeRef that indicates where to find a definition. References to  private definitions come with a specific nameSpace.

compiler/src/dotty/tools/dotc/core/tasty/TastyUnpickler.scala
```diff
@@ -64,8 +64,8 @@ class TastyUnpickler(reader: TastyReader) {
         val result = readName().toTypeName
         val params = until(end)(readName().toTypeName)
         var sig = Signature(params, result)
-        if (sig == Signature.NotAMethod) sig = Signature.NotAMethod
-        SignedName(original, sig)
+        if (sig == Signature.NotAMethod) sig = Signature.NotAMethod // needed temporarily, as long as we read old tasty
```
smarter: Drop this line and bump the tasty major version instead?

compiler/src/dotty/tools/dotc/core/Designators.scala
```diff
@@ -23,8 +26,49 @@ object Designators {
...
+    def withSig(sig: Signature): Designator{ type ThisName = TermName } = {
+      val unsigned = this.asInstanceOf[TermName].exclude(SignedName)
+      if (sig eq Signature.NotAMethod) unsigned else SignedName(unsigned, sig)
+    }
   }
 
   type TermDesignator = Designator { type ThisName = TermName }
   type TypeDesignator = Designator { type ThisName = TypeName }
+
+  case class LocalName[N <: Name](name: N, nameSpace: TypeRef) extends Designator {
```
smarter: Would be nice to document the concept of a LocalName.

smarter: Also, it's a bit confusing that LocalName is not a subclass of Name, I would maybe call it LocalDesignator ?

compiler/src/dotty/tools/dotc/core/Designators.scala
```diff
@@ -23,8 +26,49 @@ object Designators {
 
     def asTerm(implicit ctx: Context): TermDesignator = unsupported("asTerm")
     def asType(implicit ctx: Context): TypeDesignator = unsupported("asType")
+
+    def withNameSpace(space: NameSpace)(implicit ctx: Context): Designator { type ThisName = self.ThisName } =
```
smarter: Shouldn't this function be defined in `Name` instead of `Designator`?

odersky: No, because it is overridden in LocalName.

compiler/src/dotty/tools/dotc/core/tasty/TastyUnpickler.scala
odersky: See #3211

odersky:
test performance please

PR #3170, opened 2017-09-25
Fix #1463: add test case

### LINKED ISSUE #1463: Bad code for pattern matches with values.

liufengyun:
Add test case for #1463 .

### DIFF
--- a/tests/run/i1463.scala
+++ b/tests/run/i1463.scala
@@ -1,0 +1,26 @@
+object Test {
+  case object Bob { override def equals(other: Any) = true }
+
+  class Bob2 {
+    override def equals(other: Any) = true
+  }
+  val Bob2 = new Bob2
+
+  def f0(x: Any) = x match { case Bob2 => Bob2 }
+  def f1(x: Any) = x match { case Bob => Bob }
+  // def f2(x: Any): Bob.type = x match { case x @ Bob => x } // should not type check
+
+  def main(args: Array[String]): Unit = {
+    assert(f0(Bob2) eq Bob2)
+    assert(f0(0) eq Bob2)
+    assert(f0(Nil) eq Bob2)
+
+    assert(f1(Bob) eq Bob)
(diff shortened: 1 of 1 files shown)


tests/run/i1463.scala
```diff
@@ -0,0 +1,26 @@
...
+  case object Bob { override def equals(other: Any) = true }
+
+  class Bob2 {
+    override def equals(other: Any) = true
+  }
+  val Bob2 = new Bob2
+
+  def f0(x: Any) = x match { case Bob2 => Bob2 }
+  def f1(x: Any) = x match { case Bob => Bob }
+  // def f2(x: Any): Bob.type = x match { case x @ Bob => x } // should not type check
```
smarter: Could you add a neg test for that?

PR #3173, opened 2017-09-25
move 'return outside method definition' to error class

ennru:
@felixmulder Long time, no commit.


nicolasstucki:
@enru Will need to be rebased. Conflicts with the merge of your other PR

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1737,4 +1737,12 @@ object messages {
       hl"you have to provide either ${"class"}, ${"trait"}, ${"object"}, or ${"enum"} definitions after qualifiers"
   }
 
+  case class ReturnOutsideMethodDefinition()(implicit ctx: Context)
+    extends Message(ReturnOutsideMethodDefinitionID) {
+    val kind = "Syntax"
+    val msg = s"return outside method definition"
```
nicolasstucki: I would turn the phrasing around and say that `retrun` must be inside a method definition.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1737,4 +1737,12 @@ object messages {
...
+    val explanation =
+      hl"${"return"} is a keyword and may only be used within methods"
```
nicolasstucki: Here I would explain where the return vas found (i.e. in `ctx.owner` if it is not `NoContext`)

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
ennru: I kept the wording from the original, which is the same in scalac.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1737,12 +1738,27 @@ object messages {
...
+          |Instead, you should rely on the last expression's value being
+          |returned from a method.
+          |"""
   }
 
+  case class ReturnOutsideMethodDefinition(checkedContext: Context)(implicit ctx: Context)
+    extends Message(ReturnOutsideMethodDefinitionID) {
+    val kind = "Syntax"
+    val msg = hl"${"return"} outside method definition"
+    private val contextInfo =
```
allanrenucci: Should probably be a `def`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
nicolasstucki: Ok

nicolasstucki: Though we can improve

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -980,7 +980,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     def enclMethInfo(cx: Context): (Tree, Type) = {
       val owner = cx.owner
       if (cx == NoContext || owner.isType) {
```
allanrenucci: @nicolasstucki I'm curious. Does it ever happen to have `NoContext`?

allanrenucci:
Can your rebase on top of master. Also you can now define the message as follow:
```scala
  case class ReturnOutsideMethodDefinition(owner: Symbol)(implicit ctx: Context)
    extends Message(ReturnOutsideMethodDefinitionID) {
    val kind = "Syntax"
    val msg = hl"${"return"} outside method definition"
    val explanation =
      hl"""You used ${"return"} in ${owner.show}".
          |${"return"} is a keyword and may only be used within method declarations.
          |"""
  }
```

ennru:
So I understand, the `NoContext` case does not exist in practice?

allanrenucci:
>So I understand, the NoContext case does not exist in practice?

No, `cx` should never be `NoContext`.  It was fixed in #3188

ennru:
Sorry, looks like git and I didn't come along that well.
I'll make a new clean PR.

PR #3180, opened 2017-09-26
Add chat widget in the docs

OlivierBlanvillain:
See https://typelevel.org/cats/datatypes/validated.html for an example of the result


smarter:
I think it'd make sense to put it on the front page too: http://dotty.epfl.ch/

OlivierBlanvillain:
Updated to be on every page, I also changed the default color to make it less intrusive:

screenshot_2017-09-26_17-36-11

PR #3194, opened 2017-09-27
Only execute optimised/idempotency test on Drone

nicolasstucki:


smarter:
What if I really want to run them locally because they failed on drone?

nicolasstucki:
You comment out the jUnit assumptions

allanrenucci:
What about we move `testOptimised` to its own class instead. So that someone can do `testOnly CompilationTests` if he does not want to run the optimised tests?

nicolasstucki:
@odersky wanted it to work form Eclipse out of the box, the `testOnly` will not work in that scenario.

odersky:
Actually, I don't care about Eclipse anymore... But I think it's worthwhile to have a subset of tests that cover the essential aspects of what likely goes wrong, and that don't take a huge amount of time. We can get a much better workflow for developing that way. For me personally, the tests that matter are in particular, pos, neg, run, testPickler, and tastyBottstrap, and the repl tests. The others don't matter because they don't take much time anyway. Except for idempotent and optimized. I think it's really rare that these go wrong in isolation. And they do both take a lot of time.

odersky:
Actually, I think I merged this too early. The fact that we have to edit the build script to test these locally is bad. 

@nicolasstucki Can we change this so that optimised and idempotency can still be run as test-only targets, but are not run under `test`?

nicolasstucki:
With #3227 we can run those tests with `testOnly` but they are ignored with the other ones.

PR #3196, opened 2017-09-27
move 'Only static object can extend Phantom' to error class

ennru:
@allanrenucci Is there a better suited wording to explain "static". 
I reckon most people using Phantom types know about this basic stuff, right?


nicolasstucki:
We are currently redesigning the core of the phantoms, the error messages related might change in the near future. 
Hence we should not start creating error messages for them as we would start reserving IDs in`ErrorMessageID.java` for nothing.

ennru:
Ok. Never mind.

PR #3197, opened 2017-09-27
Error message for enum case in non-enum class companion object

esarbe:
see #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
+++ b/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
@@ -33,7 +33,7 @@ object DesugarEnums {
   /** Is enum case `tree` situated in a companion object of an enum class? */
   def enumCaseIsLegal(tree: Tree)(implicit ctx: Context): Boolean = (
     ctx.owner.is(ModuleClass) && enumClass.derivesFrom(defn.EnumClass)
-    || { ctx.error(EnumCaseDefinitionInNonEnumOwner(), tree.pos)
+    || { ctx.error(EnumCaseDefinitionInNonEnumOwner(ctx.owner), tree.pos)
          false
        }
     )
--- a/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
+++ b/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
@@ -6,6 +6,7 @@ import core._
 import util.Positions._, Types._, Contexts._, Constants._, Names._, NameOps._, Flags._
 import SymDenotations._, Symbols._, StdNames._, Annotations._, Trees._
 import Decorators._
+import dotty.tools.dotc.reporting.diagnostic.messages.EnumCaseDefinitionInNonEnumOwner
 import collection.mutable.ListBuffer
 import util.Property
 import typer.ErrorReporting._
@@ -32,7 +33,7 @@ object DesugarEnums {
   /** Is enum case `tree` situated in a companion object of an enum class? */
   def enumCaseIsLegal(tree: Tree)(implicit ctx: Context): Boolean = (
     ctx.owner.is(ModuleClass) && enumClass.derivesFrom(defn.EnumClass)
-    || { ctx.error(em"case not allowed here, since owner ${ctx.owner} is not an `enum' object", tree.pos)
+    || { ctx.error(EnumCaseDefinitionInNonEnumOwner(), tree.pos)
          false
        }
     )
--- a/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
+++ b/compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
@@ -6,7 +6,7 @@ import core._
 import util.Positions._, Types._, Contexts._, Constants._, Names._, NameOps._, Flags._
 import SymDenotations._, Symbols._, StdNames._, Annotations._, Trees._
 import Decorators._
-import dotty.tools.dotc.reporting.diagnostic.messages.EnumCaseDefinitionInNonEnumOwner
+import reporting.diagnostic.messages.EnumCaseDefinitionInNonEnumOwner
 import collection.mutable.ListBuffer
 import util.Property
 import typer.ErrorReporting._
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
```diff
@@ -6,6 +6,7 @@ import core._
 import util.Positions._, Types._, Contexts._, Constants._, Names._, NameOps._, Flags._
 import SymDenotations._, Symbols._, StdNames._, Annotations._, Trees._
 import Decorators._
+import dotty.tools.dotc.reporting.diagnostic.messages.EnumCaseDefinitionInNonEnumOwner
```
liufengyun: Can we shorten this to just `reporting.diagnostic.messages.EnumCaseDefinitionInNonEnumOwner`?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1748,7 +1748,7 @@ object messages {
     val explanation =
       hl"you have to provide either ${"class"}, ${"trait"}, ${"object"}, or ${"enum"} definitions after qualifiers"
   }
-
+
```
liufengyun: Introduction of a trailing space?

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1770,4 +1770,21 @@ object messages {
...
   }
 
+  case class EnumCaseDefinitionInNonEnumOwner(owner: Symbol)(implicit ctx: Context)
+    extends Message(EnumCaseDefinitionInNonEnumOwnerID) {
+      val kind = "Syntax"
+      val msg = em"case not allowed here, since owner ${owner} is not an `enum' object"
+      val explanation = {
+
+        hl"""
+            | ${"case"} is only allowed at this place if the surrounding object is the companion object of an `enum class`.
```
liufengyun: Just curious, is `${"case"}` equivalent to `case`? I see similar uses above which I don't understand.

allanrenucci:
Before we get this in, I think we should discuss if it is a good idea or not to have links in the error messages (ref #3199).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
esarbe: Uh, nice catch! Thanks.

esarbe:
I can remove the link to circumvent the discussion. I'm quite willing to come back and re-introduce the link if this should be outcome of the discussion.

allanrenucci:
You can make the tests pass by rebasing on top of master

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1770,4 +1770,18 @@ object messages {
...
+      val kind = "Syntax"
+      val msg = em"case not allowed here, since owner `${owner}` is not an `${"enum"}` object"
+      val explanation = {
+
+        hl"""
+            | `${"case"}` is only allowed at this place if the surrounding object is the companion object of an `${"enum"} ${"class"}`.
+            | If you wanted to create an enumeration make sure that the corresponding class has the `${"enum"}` keyword.
+            | Otherwise you might have forgotten `${"class"}` or `${"object"}` after this `${"case"}` here.
+            |
+          """.stripMargin
```
allanrenucci: You don't need to `stripMargin` the `hl` interpolator does it for you.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1770,4 +1770,18 @@ object messages {
           |"""
   }
 
+  case class EnumCaseDefinitionInNonEnumOwner(owner: Symbol)(implicit ctx: Context)
+    extends Message(EnumCaseDefinitionInNonEnumOwnerID) {
+      val kind = "Syntax"
+      val msg = em"case not allowed here, since owner `${owner}` is not an `${"enum"}` object"
+      val explanation = {
+
+        hl"""
```
allanrenucci: Start your message here

esarbe:
Thanks for spotting these! Should be better now.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1777,5 +1777,15 @@ object messages {
     val explanation =
       hl"""A class marked with the ${"final"} keyword cannot be extended"""
   }
-
+  case class EnumCaseDefinitionInNonEnumOwner(owner: Symbol)(implicit ctx: Context)
+    extends Message(EnumCaseDefinitionInNonEnumOwnerID) {
+      val kind = "Syntax"
+      val msg = em"case not allowed here, since owner `${owner}` is not an `${"enum"}` object"
```
allanrenucci: You shouldn't backtick `owner` and `enum`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1777,5 +1777,15 @@ object messages {
...
+      val explanation =
+        hl"""
+            | `${"case"}` is only allowed at this place if the surrounding object is the companion object of an `${"enum"} ${"class"}`.
+            | If you wanted to create an enumeration make sure that the corresponding class has the `${"enum"}` keyword.
+            | Otherwise you might have forgotten `${"class"}` or `${"object"}` after this `${"case"}` here.
+          """
```
allanrenucci: I would prefer something like:
```scala
  val explanation =
    hl"""${"enum"} cases are only allowed within the companion ${"object"} of an ${"enum class"}.
        |If you want to create an ${"enum"} case, make sure the corresponding ${"enum class"} exists
        |and has the ${"enum"} keyword."""
```

esarbe: Let's take that.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
esarbe: should be fixed

PR #3198, opened 2017-09-27
Fix #3168: Don't insert apply if tree has method type

### LINKED ISSUE #3168: Bad interaction between overloading and implicit search
The following code snippet fails to compile with Dotty
```scala
object Test {
  class C {
    def foo(x: Int) = 1
    def foo(x: Double) = 2
  }

  implicit class COps(val x: C) {
    def foo(x: String) = 3
  }
  
  def test: Unit = {
    (new C).foo("Hello")
  }
}
```
The compiler is not able to find the method `foo` added via implicit conversion. If one of the overloaded version of `foo` is commented out, the code snippet compiles

odersky:
It would be simply an eta-expansion followed by an application.
As #3168 shows, overloading resolution gets confused by this.


allanrenucci:
You mean #3168 not 3169

odersky:
Yes, indeed. #3168.

allanrenucci:
Your first commit message also refers to 3169 instead of 3168

nicolasstucki:
Needs rebase

PR #3210, opened 2017-09-29
stripMargin on explanation text

esarbe:


esarbe:
not sure it's worth to open a issue on that, so here's a standalone PR

odersky:
But here in find (in Decorators.scala): 

  implicit class StringInterpolators(val sc: StringContext) extends AnyVal {

    /** General purpose string formatting */
    def i(args: Any*)(implicit ctx: Context): String =
      new StringFormatter(sc).assemble(args)

    /** Formatting for error messages: Like `i` but suppress follow-on
     *  error messages after the first one if some of their arguments are "non-sensical".
     */
    def em(args: Any*)(implicit ctx: Context): String =
      new ErrorMessageFormatter(sc).assemble(args)

    /** Formatting with added explanations: Like `em`, but add explanations to
     *  give more info about type variables and to disambiguate where needed.
     */
    def ex(args: Any*)(implicit ctx: Context): String =
      explained(implicit ctx => em(args: _*))

    /** Formatter that adds syntax highlighting to all interpolated values */
    def hl(args: Any*)(implicit ctx: Context): String =
      new SyntaxFormatter(sc).assemble(args).stripMargin
  }

So it seems `hl` already strips the margin. No need to do that explicitly.

odersky:
Btw, I find it a pity that we can't compose these formatters. I.e you can have `ex` which gives you nice explanations about type variables or `hl` which does highlighting, but not both.

felixmulder:
@allanrenucci - `stripMargin` should already be applied to `hl` as well as `i` IIRC.

I agree with Martin's point that it would make sense if these formatters were consolidated. It would also be interesting to see if we could make `hl` a bit smarter by doing line-wrapping for instance.

PR #3212, opened 2017-09-29
Remove patter matches on ImplicitMethodType and JavaMethodType

nicolasstucki:


nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -288,6 +288,18 @@ object Types {
     /** Is this an alias TypeBounds? */
     final def isAlias: Boolean = this.isInstanceOf[TypeAlias]
 
+    /** Is this an MethodType which has implicit parameters */
```
allanrenucci: wrong doc

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -288,6 +288,18 @@ object Types {
...
+    final def isJavaMethod: Boolean = this match {
+      case mt: MethodType => mt.isJava
+      case _ => false
+    }
+
+    /** Is this an MethodType which has implicit parameters */
```
allanrenucci: s/an/a

nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
It looks like this change makes the compilation of the dotty compiler 8% faster.

smarter:
Probably not :). Points are missing for today's PR, same as https://github.com/lampepfl/dotty/pull/3216#issuecomment-333314902

smarter:
@liufengyun Many recent PRs are missing from https://liufengyun.github.io/bench/, any idea why?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2799,19 +2800,39 @@ object Types {
...
+      def is(that: MethodKind): Boolean = (mk & that) == that
+    }
   }
 
-  object ImplicitMethodType extends MethodTypeCompanion {
-    def apply(paramNames: List[TermName])(paramInfosExp: MethodType => List[Type], resultTypeExp: MethodType => Type)(implicit ctx: Context): MethodType =
-      checkValid(unique(new ImplicitMethodType(paramNames)(paramInfosExp, resultTypeExp)))
+  object MethodType extends MethodTypeCompanion { self =>
+    import MethodKinds._
+
```
odersky: Since MethodTypes are created very often (~500K for Dotty bootstrap) and there are only a small number of different ones, I would not do the `withKind` scaffolding. I would create a finite number of companions instead and use the kind as a parameter.

Actually,  I think a simpler scheme is to forgo kinds altogether and only use companions.
Yes, it will get a little bit more complicated once we add `unused`, but it's still much easier than defining separate kinds.

nicolasstucki: That design is a lot simpler. The only drawback is that we will have an additional indirection to know is a type is a Java method type or implicit method type as that knowlage is in the companion.

nicolasstucki:
test performance please

nicolasstucki:
All requested changes have been made.

nicolasstucki:
Squashed and rebased code

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2845,18 +2843,19 @@ object Types {
   }
 
   object MethodType extends MethodTypeCompanion {
-    def apply(paramNames: List[TermName])(paramInfosExp: MethodType => List[Type], resultTypeExp: MethodType => Type)(implicit ctx: Context): MethodType =
-      checkValid(unique(new CachedMethodType(paramNames)(paramInfosExp, resultTypeExp)))
+    def withKind(isJava: Boolean = false, isImplicit: Boolean = false): MethodTypeCompanion = {
```
odersky: I don't think you need that method anymore

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2802,7 +2794,10 @@ object Types {
     def syntheticParamName(n: Int) = tpnme.syntheticTypeParamName(n)
   }
 
-  abstract class MethodTypeCompanion extends TermLambdaCompanion[MethodType] {
+  abstract class MethodTypeCompanion extends TermLambdaCompanion[MethodType] { self =>
+
+    def isJava: Boolean = false
+    def isImplicit: Boolean = false
```
odersky: I don't think these are needed either?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2845,18 +2843,19 @@ object Types {
...
+      if (isJava) JavaMethodType
+      else if (isImplicit) ImplicitMethodType
+      else MethodType
+    }
   }
 
   object JavaMethodType extends MethodTypeCompanion {
-    def apply(paramNames: List[TermName])(paramInfosExp: MethodType => List[Type], resultTypeExp: MethodType => Type)(implicit ctx: Context): MethodType =
-      unique(new JavaMethodType(paramNames)(paramInfosExp, resultTypeExp))
+    override def isJava: Boolean = true
```
odersky: Can be dropped

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2845,18 +2843,19 @@ object Types {
...
   }
 
   object ImplicitMethodType extends MethodTypeCompanion {
-    def apply(paramNames: List[TermName])(paramInfosExp: MethodType => List[Type], resultTypeExp: MethodType => Type)(implicit ctx: Context): MethodType =
-      checkValid(unique(new ImplicitMethodType(paramNames)(paramInfosExp, resultTypeExp)))
+    override def isImplicit: Boolean = true
```
odersky: Can be dropped

compiler/src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala
```diff
@@ -774,8 +774,7 @@ class Scala2Unpickler(bytes: Array[Byte], classRoot: ClassDenotation, moduleClas
         def isImplicit =
           tag == IMPLICITMETHODtpe ||
           params.nonEmpty && (params.head is Implicit)
-        val maker = if (isImplicit) ImplicitMethodType else MethodType
```
odersky: I'd leave as is.

compiler/src/dotty/tools/dotc/core/Types.scala
nicolasstucki: They will become useful if we add more method types. I will remove it and add it back if it is needed in the future.

odersky:
Now we got it to be smaller than before Great! LGTM.

nicolasstucki:
Rebased

PR #3214, opened 2017-09-29
Various performance optimizations

smarter:


smarter:
(I found this performance issue by using https://github.com/jvm-profiling-tools/async-profiler)

smarter:
test performance please

smarter:
I went ahead and did some more optimizations while I was at it :).

smarter:
test performance please: 4bb0a1fe875ad64665ab0fd1a53104e52ebf56a7 5f0ed19d45818db33b31442bfc686b534dacb259 c47f0c244db1ed444adfd1f6d050b14a5645ccf9

compiler/src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -210,9 +210,16 @@ object TreeTransforms {
...
+    private def hasRedefinedMethod(cls: Class[_], name: String): Boolean = {
+      val clsMethods = cls.getDeclaredMethods
+      var i = clsMethods.length - 1
+      while (i >= 0) {
+        if (clsMethods(i).getName == name)
+          return cls != classOf[TreeTransform]
+        i -= 1
+      }
+      hasRedefinedMethod(cls.getSuperclass, name)
+    }
```
odersky: I wonder why this method shows up as hot. It should only be called during initialization when we compute the tree transformers. How often is the method called? Back of the envelope calculation: 100 phases * 50 methods, i.e. 5000? If its significantly more than that there's something we have overlooked.

odersky (approved):
It would be good to follow up on `hasRedefinedMethod`. Otherwise LGTM.

compiler/src/dotty/tools/dotc/util/Stats.scala
```diff
@@ -21,7 +21,7 @@ import collection.mutable
   }
 
   @inline
-  def record(fn: String, n: Int = 1) =
+  def record(fn: => String, n: => Int = 1) =
```
odersky: One can do even better by using two overloaded versions of `record` and no default argument.

smarter: I don't see how, could you give more details?

PR #3215, opened 2017-09-30
Simplify substituters

odersky:
RefinedTypes and TypeAliases are not longer very common, so no need
to specialize for them.


odersky:
test performance please

odersky:
The rest of the optimizations in #3033 are a bit ugly. I'll see whether we can do a more systematic treatment based on visitors.

PR #3216, opened 2017-09-30
Some more small optimizations

smarter:


smarter:
test performance please

smarter:
Note that points for today's PRs are missing in the benchmarks linked above, so the comparison is misleading. You can look at it side-by-side with http://dotty-bench.epfl.ch/3214 instead.

smarter:
Interestingly this is the first time in a while we have a code change that affects the speed of the empty class/object/file benchmarks, and in a good way too :).

smarter:
Morale: Java reflection is not only evil, it's also very slow, be careful!

compiler/src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -209,15 +211,29 @@ object TreeTransforms {
    *  @see NXTransformations.index for format of plan
    */
   class NXTransformations {
+    private val clsMethodsCache = new java.util.IdentityHashMap[Class[_], Array[java.lang.reflect.Method]]
 
+    // TODO: We spend too much time here. See if we can call it less or make it faster,
+    // e.g. by checking `cls.getMethod(name, ...).getDeclaringClass != classOf[TreeTransform]` instead.
     private def hasRedefinedMethod(cls: Class[_], name: String): Boolean = {
-      val clsMethods = cls.getDeclaredMethods
+      if (cls.eq(classOf[TreeTransform]) || cls.eq(classOf[NoTreeTransform]) ||
```
odersky: Not clear we need this anymore, because of ed5eb24.

We have just reduced the # of calls to this method by a factor 7.

smarter: Yeah, it's more of a micro-optimization now. I can remove the added cache map if you think it's not worth the complexity now.

smarter:
test performance please: be0145700e521643de62e0882a4141eada4a4c05 6fd9569c3d3e6bd2aced0e69aed07a8bec2420e5 128afccbea4597c05f7186624e44649b623f8018

liufengyun:
FYI @smarter Now the points for one PR accumulate, so it's possible to test multiple variations of the same PR continuously and see all the points in one graph.

smarter:
Cool! That's useful, but at the same time could be misleading if master changed since the last benchmarks for the PR, maybe the info bubble for the points should mention which commit from master was merged with which commit from the PR?

liufengyun:
The simplest solution is to just update the __bot message__ to make it clear that which commit is merged with which master(hash). Does that address the concern?

Note it's also possible to provide a command `retest performance please` to erase previous points and accumulate again. This can be useful in rebasing.

smarter:
This all sounds like good ideas :). Instead of erasing previous points, I suggest using a different URL when retesting instead.

smarter:
@odersky Looks like the hasRedefinedMethod optimizations are still significant (in http://dotty-bench.epfl.ch/3216/ compare 6fd9569c3d3e6bd2aced0e69aed07a8bec2420e5 to the previous point which is master, especially for empty file/object/class)

liufengyun:
retest performance please: be0145700e521643de62e0882a4141eada4a4c05 6fd9569c3d3e6bd2aced0e69aed07a8bec2420e5 128afccbea4597c05f7186624e44649b623f8018

liufengyun:
Github doesn't always pickup the `index.html` page automatically:

http://dotty-bench.epfl.ch/3216-2/index.html

liufengyun:
test performance please

Test for the PR https://github.com/liufengyun/bench/pull/63

PR #3217, opened 2017-09-30
Some more optimizations

odersky:


compiler/src/dotty/tools/dotc/transform/LinkAll.scala
```diff
@@ -36,7 +36,7 @@ class LinkAll extends MiniPhaseTransform {
     }
 
     if (ctx.settings.XlinkOptimise.value) super.runOn(allUnits(Set.empty, units.toSet, Set.empty))
```
nicolasstucki: This came from the design that @DarkDimius showed me to add new compilation units. I mostly looked at it from a functional point of view, not a performance one. I will have a look and see if it can be done in a better way.

odersky: I see. In essence, MiniPhase.runOn calls the indexing framework that figures out the method call plan. I thought that would be done before, but it seems not (maybe because we don't have the right context before?). The problem with that approach is that it is done afresh for each compilation unit, even if it 
is always the same computation.

Nobody except LinkAll seems to call `runOn`, however.

odersky:
test performance please

compiler/src/dotty/tools/dotc/transform/LinkAll.scala
nicolasstucki: This is probably cleaner #3219, there was not a good reason for it to be a miniphase. That PR should further improve performance on `LinkAll`.

nicolasstucki: If I remember correctly `runOn` was designed specifically for the purpose of adding new compilation units. Until `LinkAll` this use case was only used in the linker fork.

PR #3223, opened 2017-10-01
Some further optimizations

odersky:


compiler/src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -606,6 +606,7 @@ object Contexts {
     /** A table for hash consing unique types */
     private[core] val uniques = new util.HashSet[Type](Config.initialUniquesCapacity) {
       override def hash(x: Type): Int = x.hash
+      override def isEqual(x: Type, y: Type) = x.eql(y)
```
smarter: Shouldn't the override also be used in the other subclasses of `HashSet` ? That is, `AppliedUniques` and `NamedTypeUniques`

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -381,25 +377,31 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
 
   def isPureExpr(tree: Tree)(implicit ctx: Context) = exprPurity(tree) == Pure
   def isIdempotentExpr(tree: Tree)(implicit ctx: Context) = exprPurity(tree) >= Idempotent
+  def isReadOnlyExpr(tree: Tree)(implicit ctx: Context) = exprPurity(tree) >= ReadOnly
```
smarter: This commit would benefit from some testcases. I suggest adding some tests like https://github.com/lampepfl/dotty/blob/1e0863708e48339759f73f38464a410df850df15/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala#L168-L189 which check that two methods end up with the same bytecode generated.

compiler/src/dotty/tools/dotc/core/Contexts.scala
odersky: No, `equals` is never called on those sets. Everything is based on `eq` for them.

smarter:
The second commit results in additional incorrect warnings:
```scala
[warn] -- Warning: /tmp/3/compiler/src/dotty/tools/dotc/core/tasty/NameBuffer.scala:47:4 
[warn] 47 |    op
[warn]    |    ^^
[warn] | a read-only expression does nothing in statement position
```

I suggest breaking off the second commit in a separate PR since it's not an optimization, and we can get the optimization in now.

odersky:
@smarter Yes, good idea.

odersky:
test performance please

smarter:
As before, these data points have to be looked at carefully because they're not including today's merged PRs. @liufengyun are the PR data points generated after every PR is merged or daily?

liufengyun:
@smarter The daily points are generated daily at mid-night.

smarter:
Would it be possible to generate them everytime a PR is merged instead?

liufengyun:
Yes, we can change it to run immediately after merge. I'll make it the next TODO in bench.

PR #3227, opened 2017-10-01
Allow running long tests with testOnly

nicolasstucki:
Improves #3194

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -465,13 +465,13 @@ object Build {
 
       test in Test := {
         // Exclude legacy tests by default
-        (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception").value
+        (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception,dotty.tools.dotc.SlowTests").value
       },
 
       vulpix := Def.inputTaskDyn {
         val args: Seq[String] = spaceDelimited("<arg>").parsed
-        val cmd = " dotty.tools.dotc.CompilationTests" + {
-          if (args.nonEmpty) " -- -Ddotty.tests.filter=" + args.mkString(" ")
+        val cmd = " dotty.tools.dotc.CompilationTests -- --exclude-categories=dotty.tools.dotc.SlowTests" + {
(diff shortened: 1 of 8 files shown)


allanrenucci:
What about a solution based on [junit categories](https://github.com/junit-team/junit4/wiki/Categories)?

project/Build.scala
```diff
@@ -465,13 +465,13 @@ object Build {
 
       test in Test := {
         // Exclude legacy tests by default
-        (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception").value
+        (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception,dotty.tools.dotc.SlowTests").value
```
allanrenucci: Does this mean that Drone will not run them?

nicolasstucki: Oh, it won't. I will have to find a workaround

project/Build.scala
```diff
@@ -465,13 +468,18 @@ object Build {
 
       test in Test := {
         // Exclude legacy tests by default
+        (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception,dotty.tools.dotc.SlowTests").value
+      },
+
+      testAll in Test := {
+        // Exclude legacy tests by default
         (testOnly in Test).toTask(" -- --exclude-categories=java.lang.Exception").value
```
allanrenucci: While you're at it, could you create a meaningful category for legacy tests?

nicolasstucki: I will

nicolasstucki:
@allanrenucci could you review the las commit

allanrenucci:
Can we put all the categories in one file. Maybe under `compiler/test/dotty/TestCategories`.  Also I think you should be able to do `@Category(classOf[SlowTests])` instead of `@Category(Array(classOf[SlowTests]))` when a test belongs to only one category. Otherwise LGTM

nicolasstucki:
It does not work with `@Category(classOf[SlowTests])`, probably implementation details of the JUnit plugin.

Issue #3008, opened 2017-08-23
Account for `@showAsInfix` in `isInfixType`


OlivierBlanvillain:
[This TODO](https://github.com/dotty-staging/dotty/blob/7bfe4d5241c4e4c1f15e18ea950b23f738f4c683/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L132)

https://github.com/scala/scala/blob/v2.12.3/src/library/scala/annotation/showAsInfix.scala

Issue #2462, opened 2017-05-17
REPL: importing an annotation does not seem to bring it in scope


smarter:
This might be the same bug as https://github.com/lampepfl/dotty/issues/2109, I'm not sure:
```scala
scala> import scala.annotation.unspecialized 
import scala.annotation.unspecialized
scala> class Foo[@unspecialized T] 
-- [E006] Unbound Identifier Error: <console>:4:11 -----------------------------
4 |class Foo[@unspecialized T]
  |           ^^^^^^^^^^^^^
  |           not found: type unspecialized

longer explanation available when compiling with `-explain`
scala> class Foo[@scala.annotation.unspecialized T] 
defined class Foo
```

PR #3228, opened 2017-10-01
Enchancement #3008

CucumisSativus:
Refined printer checks if showAsInflisAnotType annotation is used (issue #3008). 

I would be grateful if you could lend me a hand in testing my solution, couldn't find any tests for RefinedPrinter.

If something is missing please let me know, i'm happy to improve this PR :)

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
@@ -126,8 +126,8 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
     def isInfixType(tp: Type): Boolean = tp match {
       case AppliedType(tycon, args) =>
         args.length == 2 &&
-          !Character.isUnicodeIdentifierStart(tycon.typeSymbol.name.toString.head)
-          // TODO: Once we use the 2.12 stdlib, also check the @showAsInfix annotation
+        tycon.typeSymbol.getAnnotation(defn.ShowAsInfixAnnot).map(_.argumentConstant(0).forall(_.booleanValue))
+          .getOrElse(!Character.isUnicodeIdentifierStart(tycon.typeSymbol.name.toString.head))
       case _ => false
     }
 
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -661,6 +661,8 @@ class Definitions {
   def GetterMetaAnnot(implicit ctx: Context) = GetterMetaAnnotType.symbol.asClass
   lazy val SetterMetaAnnotType = ctx.requiredClassRef("scala.annotation.meta.setter")
   def SetterMetaAnnot(implicit ctx: Context) = SetterMetaAnnotType.symbol.asClass
+  lazy val ShowAsInfixAnotType = ctx.requiredClassRef("scala.annotation.showAsInfix")
+  def ShowAsInfixAnnot(implicit ctx: Context) = ShowAsInfixAnotType.symbol.asClass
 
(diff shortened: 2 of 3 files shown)


smarter:
Hi! This already looks pretty good, a few improvements that could be made:
- The commit message title should be: ``Fix #3008: Account for `@showAsInfix` in `isInfixType` ``, using "Fix #nnnn" means that the issue will automatically be closed when the PR is merged.
- It would be nice to have tests to verify that this work as expected. We already have an existing REPL test for infix type: https://github.com/lampepfl/dotty/blob/master/compiler/test-resources/type-printer/infix s you can just add the cases using `@showAsInfix` from the Scala 2 tests: https://github.com/scala/scala/blob/e2be7c498dd849a007ab3e2a923fe48c8ae74b94/test/files/run/t4700.check

smarter:
By the way, to check that the modified test works correctly, you can run just the REPL tests by typing in sbt:
```scala
dotty-compiler/test-only -- --tests=typePrinterTests
```

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -126,8 +126,8 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
     def isInfixType(tp: Type): Boolean = tp match {
       case AppliedType(tycon, args) =>
         args.length == 2 &&
-          !Character.isUnicodeIdentifierStart(tycon.typeSymbol.name.toString.head)
-          // TODO: Once we use the 2.12 stdlib, also check the @showAsInfix annotation
+          !Character.isUnicodeIdentifierStart(tycon.typeSymbol.name.toString.head) &&
```
smarter: This should be an `||` not an `&&` as we want infix types to be used either for type constructors which start with a symbol _or_ that have the `@showAsInfix` annotation.

CucumisSativus: There's one more problem. I didnt notice that this annotation takes a boolean argument, which should be checked as well.

CucumisSativus:
@smarter Thanks for pointing me to tests, looks like there's a problem  - `not found: type showAsInfix` (even after import).  I'll take a closer looks at this

smarter:
Ah you're running into a known bug (https://github.com/lampepfl/dotty/issues/2462), you can work around it by using the fully qualified name of the annotation e.g:
```scala
scala> @scala.annotation.showAsInfix type Mappy[T, U] = Map[T, U]
```

CucumisSativus:
@smarter I added tests, the only problem I have right now is using type alias `type Mappy[T,U] = Map[T,U]`, when accessing aliased type whenever i call         
`tycon.classSymbol` or `tycon.typeSymbol` i receive `Map` not `Mappy`, do you know how could I access `Mappy` there?

smarter:
Hmm this is probably the Dotty typechecker being too eager to dealias things, I'll look into it but it might be tricky to fix. We can merge this PR without the type aliases working since it already improves things anyway.

CucumisSativus:
@smarter so my proposition is to merge this pull request and open a new issue where someone (maybe even me, with some guidance :) ) can check how to approach type aliases

smarter:
LGTM, thanks for your contribution!

smarter:
> maybe even me, with some guidance :)

I'm afraid this might be a hard issue to fix :) I suggest you take a look at https://github.com/lampepfl/dotty/issues?q=is%3Aissue+is%3Aopen+label%3Aexp%3Anovice or ask on http://gitter.im/lampepfl/dotty if you're looking for other things to contribute.

CucumisSativus:
> LGTM, thanks for your contribution!
Thanks for your guidance and help :)

PR #3229, opened 2017-10-02
Micro-optimization: Make denotAt inlineable

odersky:
Reduces the size of denotAt from 37 to 32 bytes, which should
theoretically make it eligible for inlining.

denotAt showed up at 2% of total execution time in my profile.


odersky:
test performance please

liufengyun:
It's impressive, in 10 days, we reduced compilation time for Dotty from ~45s to ~35s.

PR #3231, opened 2017-10-02
Do not merge: disable RefChecks to measure its overhead

smarter:


smarter:
test performance please

smarter:
Looks like it depends a lot on the code, for Dotty it's a ~2% difference but for scala-library it's mor elike ~15%

PR #3233, opened 2017-10-03
Fix GADT-related memory leak

odersky:
There was a confusion which led to the wrong gadt map being used
for pattern-bound variables if there was no other GADT variable
in the enclosing method.

This led to the outermost gadt map in the initial context being
populated with type bounds which never went away.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Contexts.scala
+++ b/compiler/src/dotty/tools/dotc/core/Contexts.scala
@@ -700,7 +700,8 @@ object Contexts {
       myBounds = myBounds.updated(sym, b)
     def bounds = myBounds
   }
-  object EmptyGADTMap extends GADTMap(SimpleMap.Empty) {
-    override def setBounds(sym: Symbol, b: TypeBounds) = unsupported("EmptyGADTMap.setBound")
+
+  @sharable object EmptyGADTMap extends GADTMap(SimpleMap.Empty) {
+    override def setBounds(sym: Symbol, b: TypeBounds) = unsupported("EmptyGADTMap.setBounds")
   }
 }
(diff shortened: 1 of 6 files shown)


odersky:
test performance please

compiler/src/dotty/tools/dotc/core/Contexts.scala
```diff
@@ -532,7 +532,7 @@ object Contexts {
     moreProperties = Map.empty
     typeComparer = new TypeComparer(this)
     searchHistory = new SearchHistory(0, Map())
-    gadt = new GADTMap(SimpleMap.Empty)
+    gadt = new GADTMap(SimpleMap.Empty) // EmptyGADTMap
```
smarter: Was this supposed to be `= EmptyGADTMap` ?

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -289,6 +281,8 @@ class PostTyper extends MacroTransform with SymTransformer  { thisTransformer =>
             case _                                  =>
           }
           super.transform(tree)
+        case Typed(Ident(nme.WILDCARD), _) =>
+          tree // skip checking pattern type
```
smarter: Why?

odersky: Because type bounds in a pattern need not conform to selector bounds. I.e. you have 

    type Tree[T >: Null <: Type]

You are still allowed to write

    case x: Tree[_]

(which translates to)

    case x: (_: Tree[_])

odersky: I am going to refine & explain the logic better

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -256,7 +256,7 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisTrans
         case tree @ Annotated(annotated, annot) =>
           cpy.Annotated(tree)(transform(annotated), transformAnnot(annot))
         case tree: AppliedTypeTree =>
-          Checking.checkAppliedType(tree)
+          Checking.checkAppliedType(tree, boundsCheck = !ctx.mode.is(Mode.Pattern))
```
smarter: Now I'm confusing about why we would write this instead of `if (!ctx.mode.is(Mode.Pattern)) Checking.checkAppliedType(tree)` :)

odersky: Because checkAppliedType also does other things besides bounds checking.

PR #3234, opened 2017-10-03
Generate generic java signatures

Duhemm:
The code has been ported from scalac.

Two flags are added:
 - `-Yno-generic-signatures` disables the generation of generic signatures
 - `-Xverify-signatures` runs further verifications of the signatures, and emits a warning if the signature is not valid. 

As discussed, the signatures are not generated if they contain symbols that we cannot encode for the JVM.


smarter:
test performance please

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -516,6 +516,18 @@ object Symbols {
     /** The current name of this symbol */
     final def name(implicit ctx: Context): ThisName = denot.name.asInstanceOf[ThisName]
 
+    def isHigherOrderTypeParameter(implicit ctx: Context): Boolean = this.maybeOwner.isTypeParameterOrSkolem
+    def isTypeParameterOrSkolem(implicit ctx: Context): Boolean = this.isTypeParam
```
odersky: Just use `isTypeParam`. `isTypeParameterOrSkolem` would be confusing in a dotty context.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -516,6 +516,18 @@ object Symbols {
     /** The current name of this symbol */
     final def name(implicit ctx: Context): ThisName = denot.name.asInstanceOf[ThisName]
 
+    def isHigherOrderTypeParameter(implicit ctx: Context): Boolean = this.maybeOwner.isTypeParameterOrSkolem
```
odersky: It's just used once. I'd just use the RHS directly in Erasure. That would avoid having to look up what the term means.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,4 +679,485 @@ object Erasure {
...
+  }
+
+  // only refer to type params that will actually make it into the sig, this excludes:
+  // * higher-order type parameters
+  // * type parameters appearing in method parameters
+  // * type members not visible in an enclosing template
+  private def isTypeParameterInSig(sym: Symbol, initialSymbol: Symbol)(implicit ctx: Context) = {
+    !sym.isHigherOrderTypeParameter &&
+      sym.isTypeParameterOrSkolem && (
+      (initialSymbol.enclClassChain.exists(sym isContainedIn _)) ||
```
odersky: I believe equivalent, and faster would be:

    sym.isContainedIn(initialSymbol.topLevelClass)

odersky: That also means we don't need `enclClassChain`.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -516,6 +516,18 @@ object Symbols {
...
+    def isTypeParameterOrSkolem(implicit ctx: Context): Boolean = this.isTypeParam
+    def enclClassChain(implicit ctx: Context): List[Symbol] = this.maybeOwner.enclClassChain
+    final def isDerivedValueClass(implicit ctx: Context): Boolean =
+      isClass && !denot.is(Flags.Package) && !denot.is(Flags.Trait) &&
```
odersky: No need to write `denot`. It's used as an implicit everywhere else.

odersky: Symbols.scala, as most source files, imports all flags with a wildcard. So for consistency I would not use the `Flags` prefix.

odersky: You can combine the two tests like this:

    !this.is(Package | Trait)

odersky: But in the end, you should just use `ValueClasses.isDerivedValueClass`. If the two definitions differ, we should find out which is the right one and use only that one.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -516,6 +516,18 @@ object Symbols {
...
+        !ctx.phase.erasedTypes && denot.info.firstParent.typeSymbol == defn.AnyValClass && !denot.isPrimitiveValueClass
+
+    /** If this is a derived value class, return its unbox method
+     *  or NoSymbol if it does not exist.
+     */
+    def derivedValueClassUnbox(implicit ctx: Context): Symbol = NoSymbol
```
odersky: Use `ValueClasses.valueClassUnbox` instead.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -622,17 +634,30 @@ object Symbols {
...
+    override def enclClassChain(implicit ctx: Context): List[Symbol] = Nil
   }
 
   @sharable object NoSymbol extends Symbol(NoCoord, 0) {
     denot = NoDenotation
 
     override def associatedFile(implicit ctx: Context): AbstractFile = NoSource.file
+
+    override def enclClassChain(implicit ctx: Context): List[Symbol] = Nil
   }
```
odersky: So overall, none of the additions to Symbols are needed (which is good, Symbols should be kept small). If you need functionality it's better to add it to `SymDenotations`, or if it's transform-specific even better to `SymUtils`.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -739,6 +739,24 @@ object Types {
...
+        case t if t eq defn.LongClass    => nme.specializedTypeNames.Long
+        case t if t eq defn.ShortClass   => nme.specializedTypeNames.Short
+        case t if t eq defn.FloatClass   => nme.specializedTypeNames.Float
+        case t if t eq defn.UnitClass    => nme.specializedTypeNames.Void
+        case t if t eq defn.DoubleClass  => nme.specializedTypeNames.Double
+        case t if t eq defn.CharClass    => nme.specializedTypeNames.Char
+        case _                           => nme.specializedTypeNames.Object
+      }
+    }
+
```
odersky: This should probably go in `Definitions`, where we define functionality related to numeric classes. `Types` is not concerned with these. I suggest to  make it another parameter of `Definitions.valueTypeRef`.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -29,6 +29,10 @@ import TypeUtils._
 import ExplicitOuter._
 import core.Mode
 import core.PhantomErasure
+import dotty.tools.dotc.core.Annotations.Annotation
+import dotty.tools.dotc.core.classfile.ClassfileConstants
+import dotty.tools.dotc.core.TypeApplications.TypeParamInfo
+import java.lang.StringBuilder
 
 class Erasure extends Phase with DenotTransformer { thisTransformer =>
```
odersky: Erasure is already rather large as is, and therefore people are scared to touch it. We should try to signature generation in a separate sourcefile. It could be integrated as an object in erasure similarly to how Bridges.scala moves out bridge generation from erasure.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,4 +679,485 @@ object Erasure {
...
+        }
+        leaves += candidate
+      }
+      rest = rest.tail
+    }
+    leaves.toList
+  }
+
+  private def hiBounds(bounds: TypeBounds)(implicit ctx: Context): List[Type] = bounds.hi.widenDealias match {
+    case AndType(tp1, tp2) => tp1 :: tp2 :: Nil
```
odersky: What happens with nested AndTypes? Don't you need a recursive call here?

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,4 +679,485 @@ object Erasure {
...
+  }
+
+  /** Arrays despite their finality may turn up as refined type parents,
+    *  e.g. with "tagged types" like Array[Int] with T.
+    */
+  private def unboundedGenericArrayLevel(tp: Type)(implicit ctx: Context): Int = tp match {
+    case GenericArray(level, core) if !(core <:< defn.AnyRefType) =>
+      level
+    case AndType(tp1, tp2) =>
+      Math.max(unboundedGenericArrayLevel(tp1), unboundedGenericArrayLevel(tp2))
```
odersky: Can use infix `max` instead.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,4 +679,485 @@ object Erasure {
...
+  private def isTypeParameterInSig(sym: Symbol, initialSymbol: Symbol)(implicit ctx: Context) = {
+    !sym.isHigherOrderTypeParameter &&
+      sym.isTypeParameterOrSkolem && (
+      (initialSymbol.enclClassChain.exists(sym isContainedIn _)) ||
+        (initialSymbol.is(Flags.Method) && initialSymbol.typeParams.contains(sym))
+      )
+  }
+
+  final def javaSig(sym0: Symbol, info: Type, markClassUsed: Symbol => Unit)(implicit ctx: Context): Option[String] =
+    ctx.atPhase(ctx.erasurePhase) { implicit ctx => javaSig0(sym0, info, markClassUsed) }
```
odersky: Simpler: 

    javaSig0(sym0, info, markClassUsed)(ctx.withPhase(ctx.erasurePhase))

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,4 +679,485 @@ object Erasure {
...
+  // See also #2585 marker in javaSig: there, type arguments must be
+  // included (use pre.baseType(cls.owner)).
+  //
+  // This requires that cls.isClass.
+  private def rebindInnerClass(pre: Type, cls: Symbol)(implicit ctx: Context): Type = {
+    val owner = cls.owner
+    if (owner.is(Flags.PackageClass) || owner.isTerm) pre else cls.owner.info /* .tpe_* */
+  }
+
+  object GenericArray {
```
odersky: It would be good to base this on the `Definitions.MultiArrayOf` extractor. This would avoid duplication and simplify the code that needs to be added.

compiler/src/dotty/tools/dotc/core/Symbols.scala
Duhemm: I kept the original name to ease review. I don't really know what `Skolem` refers to, and was afraid to lose the change in the port from scalac.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
Duhemm: Interesting. I thought at first that it was not necessary, because all the parents would be picked up in `jsig` and the signature would be correct, but further investigation showed that I was wrong, and it's actually a bug in scalac:

Considering the following definitions:
```scala
class C1 { def c1 = 0 }
trait T1 { def t1: Int }
trait T2 { def t2: Int }
class Foo[T <: (C1 with T1) with T2] { // Note the parentheses
  def foo(t: T): Int = t.c1 + t.t1 + t.t2
}
```

The signature of `Foo` that scalac generates is: `<T:LC1;:LT2;>Ljava/lang/Object` There's no mention of `T1`! This means that javac won't complain if someone passes an instance of `C1 with T2` to `def foo`, resulting in a runtime crash.

With the recursive call that you suggest, dotty generates the expected signature: `<T:LC1;:LT1;:LT2;>Ljava/lang/Object;`

smarter: /cc @retronym

retronym: Thanks, noted as https://github.com/scala/bug/issues/10543

adriaanm: Hrm, looks like another instance of too aggressively replacing normalize (which flattens refined types) with dealias in
```
  private def hiBounds(bounds: TypeBounds): List[Type] = bounds.hi.dealiasWiden match {
    case RefinedType(parents, _) => parents map (_.dealiasWiden)
    case tp                      => tp :: Nil
  }
```

compiler/src/dotty/tools/dotc/transform/Erasure.scala
Duhemm: I simplified `GenericArray` as much as I could, but I cannot use `MultiArrayOf` directly, since it'll resolve the upper bound when encountering type bounds. This means that types such as
```scala
class Foo[T <: Array[Int]]
```
would get a wrong signature (`T <: java.lang.Object` instead of `T <: int[]`)

compiler/src/dotty/tools/dotc/core/Types.scala
Duhemm: Addressed in https://github.com/lampepfl/dotty/pull/3234/commits/fb6e1f1f75b756f68a109369c622fb5a6a7956a5. Is that what you had in mind?

Duhemm:
This is ready for another round of review 😄

compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
```diff
@@ -0,0 +1,510 @@
+package dotty.tools
+package dotc
```
allanrenucci: Should probably be in `backend/jvm`

compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
```diff
@@ -0,0 +1,510 @@
...
+import core.Types._
+import core.classfile.ClassfileConstants
+import ast.Trees._
+import TypeUtils._
+import java.lang.StringBuilder
+
+/** Helper object to generate generic java signatures, as defined in
+ *  the Java Virtual Machine Specification, §4.3.4
+ */
+object GenericSignatures {
```
allanrenucci: `GenericJavaSignatures`?

compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
odersky: Debatable where it should go. But since most of it is part of erasure I'd leave it in transform.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -745,6 +745,7 @@ object Types {
       else ctx.asSeenFrom(this, pre, cls)
     }
 
+
```
odersky: Spurious empty line?

compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
allanrenucci: It is only called from `backend/jvm/DottyBackendInterface.scala`. So my reasoning was that this code is platform specific.
But I don't have a strong opinion about where it should be.

Duhemm:
Rebased and added `-Xverify-signatures` to the default test options.

xuwei-k:
https://github.com/scala/bug/issues/12246

PR #3235, opened 2017-10-03
Turn off position checking for XML-generated blocks

smarter:
@allanrenucci Can you check if this is good enough for scalatest?


smarter:
I gave up fixing the positions and just turned off the position checker for XML blocks.

allanrenucci:
This seems to do the trick

PR #3241, opened 2017-10-03
Replace `private var` by `private[this] var` where possible

smarter:


odersky:
Why do it? It does not make the code more beautiful, for sure.

smarter:
Makes the intended usage more explicit, and it generates less bytecode (direct field access instead of accessor method calls). I agree that the syntax is not the nicest, I think @Ichoran suggested `my` as an alias for `private[this]` to encourage its use, that would help ;).

smarter:
test performance please

odersky:
How much less bytecode is generated? Performance data don't indicate a change in speed.

PR #3242, opened 2017-10-03
TastyFormat: Don't use final vals for MajorVersion/MinorVersion

smarter:
Because sbt incremental compilation for scalac is broken for final vals,
so the dotty-library-bootstrapped compiled by the non-bootstrapped
dotty-compiler may have an old MajorVersion/MinorVersion unless you do
`clean` first.


odersky:
@smarter Somehow I cannot launchIDE anymore after merging this change. Do you experience the same?

smarter:
`sbt launchIDE` seems to work OK here, what error are you getting?

PR #3254, opened 2017-10-04
Fix #3252: Change logic in adaptNoArgs

### LINKED ISSUE #3252: Implicit numeric conversion
```scala
scala> val i: Long = List(1, 2, 3).sum 
1 |val i: Long = List(1, 2, 3).sum
  |                               found:    scala.math.Numeric[Int] => Int
  |                               required: Long
```

odersky:
adaptNoArgs would do an implicit argument search only if the result type
conformed to the expected type. #3252 shows that this needs to be revised.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1994,68 +1994,76 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       }
     }
 
-    def adaptNoArgs(wtp: Type): Tree = wtp match {
-      case wtp: ExprType =>
-        adaptInterpolated(tree.withType(wtp.resultType), pt)
-      case wtp: ImplicitMethodType if constrainResult(wtp, followAlias(pt)) =>
-        val tvarsToInstantiate = tvarsInParams(tree)
-        wtp.paramInfos.foreach(instantiateSelected(_, tvarsToInstantiate))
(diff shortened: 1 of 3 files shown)


odersky:
test performance please

odersky:
To review this it's best to look at the first commit in isolation, which is reasonably small. The rest is just moving code around.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1994,169 +1994,186 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+    }
+
+    def adaptNoArgs(wtp: Type): Tree = {
+      val ptNorm = underlyingApplied(pt)
+      val functionExpected = defn.isFunctionType(ptNorm)
+      wtp match {
+        case wtp: ExprType =>
+          adaptInterpolated(tree.withType(wtp.resultType), pt)
+        case wtp: ImplicitMethodType
+        if constrainResult(wtp, followAlias(pt)) || !functionExpected =>
```
allanrenucci: Does the order of the checks matter here? I see that `constrainResult` generates new constraints if it returns `true`.
If the order does not matter, then it should be less expensive to first check `!functionExpected`.

odersky: Good observation. The order does matter, actually. `constrainResult` prunes the constraint if it succeeds so we want to always run it.

PR #3255, opened 2017-10-04
sbt-dotty: Add a setting to exclude a configuration from the Dotty IDE, and exclude the optimized projects from the Dotty build

smarter:


sbt-dotty/src/dotty/tools/sbtplugin/DottyIDEPlugin.scala
```diff
@@ -15,6 +15,16 @@ import scala.collection.mutable.ListBuffer
 import DottyPlugin.autoImport._
 
 object DottyIDEPlugin extends AutoPlugin {
+  object autoImport {
+    val excludeFromIDE = settingKey[Boolean]("If true, do not import this project or configuration in the Dotty IDE")
```
olafurpg: Not sure if it's a good idea, is it worth considering naming this key `ide-skip-project`? To match https://github.com/JetBrains/sbt-ide-settings/blob/750b993453fb3d1f31f371968d06e3fc792870a1/src/main/scala/sbtide/Keys.scala#L11-L12

Also, by convention, keys from the same plugin should ideally use the same prefix:
- ideLaunch
- ideExclude
- ideCommand
- ideRun

This convention makes it easy to explore available plugin keys from the sbt shell.

smarter: >  is it worth considering naming this key ide-skip-project?

Using the same key name as another plugin seems like a bad idea in fact, because then they will conflict with each other, I have no idea what that does in the sbt repl but in build files you'll get errors due to ambiguity.

> Also, by convention, keys from the same plugin should ideally use the same prefix

Maybe yes, feel free to open an issue for that. I have no idea what the good practices for sbt plugins are :).

Duhemm: Instead of defining a new key, can't we just use `skip in launchIDE := true`?

smarter: That wouldn't be accurate since `launchIDE` just calls `configureIDE` and I don't think we can do `skip in configureIDE` since `configureIDE` is a command and not a task. It also wouldn't be complete since we also want to disable the config for `compileForIDE`. All in all, it seems better to have a specific key for this purpose :).

project/Build.scala
```diff
@@ -228,7 +228,15 @@ object Build {
...
   // Bootstrap with -optimise
-  lazy val commonOptimisedSettings = commonBootstrappedSettings ++ Seq(scalacOptions ++= Seq("-optimise"))
+  lazy val commonOptimisedSettings = commonBootstrappedSettings ++ Seq(
+    scalacOptions ++= Seq("-optimise"),
+
+    // The *-bootstrapped and *-optimised projects contain the same sources, so
+    // we only need to import one set in the IDE. We prefer to import the
+    // non-optimized projects because optimize is slower to compile and we do
+    // not trust its output yet.
+    excludeFromIDE := true
```
Duhemm: I read the above comment as "`*-bootstrapped` projects must be skipped, so why isn't this in `commonBootstrappedSettings` instead? My guess is that it's because `dotty-language-server` exists only in bootstrapped-mode and shouldn't be skipped?

Interestingly, `commonBootstrappedSettings` contains: 
```scala
EclipseKeys.skipProject := true
```

It would probably be less surprising if kept close to `excludeFromIDE := true`.

smarter: There are three kinds of projects with the same set of sources:
- Non-bootstrapped (compiled with Scala 2.12)
- Bootstrapped (compiled with Dotty non-bootstrapped, default options)
- Optimized (compiled with Dotty non-bootstrapped, with -optimise added)

For the Dotty IDE, the non-bootstrapped don't matter: they're not Dotty-compiled projects so they will be skipped by the plugin, so we only need to decide if we want the bootstrapped or the optimized one, we prefer the bootstrapped non-optimized for the reason given by the comment.

For Eclipse, the situation is different: the only projects which we want to import are the non-bootstrapped one, because the Scala IDE can only deal with Scala 2-compiled projects, so we skip everything else (commonBootstrappedSettings is also part of the settings used for optimized projects).

Hope this clears things up :).

PR #3257, opened 2017-10-04
Some micro-optimizations around phases

odersky:


odersky:
test performance please

PR #3258, opened 2017-10-04
Actually fill the TyperState#testReporter cache

smarter:
This cache was never set before, only read from.


smarter:
I'm also thinking it might be better to always call `testReporter.reset() in the `finally` block to allow nested contexts/trees/types references from Messages stored in the testReporter to be garbage-collected faster.

PR #3263, opened 2017-10-05
Micro-optimization: Symbol#denot

odersky:


odersky:
test performance please

odersky:
test performance please

liufengyun:
The benchmark test is currently running for this PR, should finish in about 40min.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -398,19 +398,34 @@ object Symbols {
...
+      else computeDenot(lastd)
+    }
+
+    private def computeDenot(lastd: SymDenotation)(implicit ctx: Context): SymDenotation = {
+      val now = ctx.period
+      checkedPeriod = now
+      if (lastd.validFor contains now) lastd else recomputeDenot(lastd)
+    }
+
+    /** Overridden in NoSymbol */
```
smarter: Not sure if it'll help performance, but have you tried doing `if (this eq NoSymbol) NoDenotation else ...` to avoid the override?

odersky: I have not tired it. I don't think we'd see a difference either way. Would be good to know more about tradeoffs like this, though.

PR #3265, opened 2017-10-05
Optimizations: Make code inlining friendlier

odersky:
I noted that despite inlining and the printer scheme key operations had a lot of junk in them. This PR tries to address that, so that we remove unnecessary debug code from hot operations.

First, making completion printing a config option keeps the size of
key method `SymDenotations#completeFrom` small. (completeFrom showed up as taking alomst 1% in the no HT profile)

Second, we also make all tracing operations conditional on a Config option. 

Third, we massage tracing operations themselves and tweak inlining so that no trace of the tracing operation remains in the generated code.


odersky:
test performance please

odersky:
test performance please

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -217,6 +217,27 @@ class FirstTransform extends MiniPhaseTransform with InfoTransformer with Annota
...
+   *      true  || y   ==>  true
+   *      false || y   ==>  y
+   */
+  private def foldCondition(tree: Apply)(implicit ctx: Context) = tree.fun match {
+    case Select(x @ Literal(Constant(c: Boolean)), op) =>
+      tree.args match {
+        case y :: Nil if y.tpe.widen.isRef(defn.BooleanClass) =>
+          op match {
+            case nme.ZAND => if (c) y else x
+            case nme.ZOR  => if (c) x else y
```
retronym: Probably doesn't matter in this case, but I prefer to check the symbol's name (or identity), rather than the selected name, so as to avoid mis-classification in the presence of renaming imports.

PR #3267, opened 2017-10-05
Avoid creating closure for asserts

nicolasstucki:


nicolasstucki:
test performance please

odersky:
Did you check the bytecode that this does in fact inline? Just being paranoiac here.

Otherwise LGTM

nicolasstucki:
Yes it does inline, but the generated code is a bit larger.
I would prefer patching the assert I found problematic by hand.

nicolasstucki:
While compiling `dotty.tools.dotc.core` e2f4be6 reduces the time of `FlagSet.|` from 0.911 sec to 0.010 sec.

I would drop the first commit. The standard library already has this `@inline` in the `Predef.assert` but it is ignored by dotty as it is compiled in `scala2` mode. My solution may improve it a bit.

odersky:
I think we can keep the first commit. There are other asserts, where it would be nice to give them a second parameter without fear that a closure will be created.

nicolasstucki:
Removed the second commit

PR #3270, opened 2017-10-05
Do not merge: Benchmark with -optimise enabled

smarter:


smarter:
test performance please

nicolasstucki:
We have two factors to explain the slowdown, the optimised compiler is slow and the optimizations take time.
We should make sure it is only the second one.
We will need normal benchmarks executed on the optimised compiler.

PR #3272, opened 2017-10-05
Avoid most deep subtypes in test

smarter:
The only remaining testcase requiring deep subtype is t8146a.scala for
good reasons.


smarter:
test performance please

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1089,8 +1089,20 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+        val nestedCtx = ctx.fresh.setExploreTyperState()
+
+        {
+          implicit val ctx = nestedCtx
+
+          // Fully define the type so that the types of the tparams created
+          // below never contain TypeRefs whose underling types contain
+          // uninstantiated TypeVars, this could lead to cycles in `isSubType`
+          // as a TypeVar might get constrained by a TypeRef it's part of.
+          val tp1a = fullyDefinedType(tp1, "alternative", alt1.symbol.pos).asInstanceOf[PolyType]
```
odersky: I wonder whether we could optimize this so that we do the fullyDefinedType only if the type parameter bounds are non-trivial. fullyDefinedType is somewhat expensive.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1089,8 +1089,20 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
         isApplicable(alt2, formals1, WildcardType) ||
         tp1.paramInfos.isEmpty && tp2.isInstanceOf[LambdaType]
       case tp1: PolyType => // (2)
-        val tparams = ctx.newTypeParams(alt1.symbol, tp1.paramNames, EmptyFlags, tp1.instantiateBounds)
-        isAsSpecific(alt1, tp1.instantiate(tparams.map(_.typeRef)), alt2, tp2)
+        val nestedCtx = ctx.fresh.setExploreTyperState()
+
```
odersky: I believe we can use TyperState.test here/ It's cheaper and easier to use.

PR #3275, opened 2017-10-06
Fix #3252: Generalize constraint handling when harmonizing

### LINKED ISSUE #3252: Implicit numeric conversion

### LINKED ISSUE #3253: Overloaded or recursive method test needs return type
This doesn't compile:
```scala
import Test.test

object Test {
  def test = "  " * 10
}

```

```scala
[info] Running dotty.tools.dotc.Main -d out tests/allan/Test.scala -classpath dotty-library_2.12-0.4.0-bin-SNAPSHOT-nonbootstrapped.jar
-- [E043] Syntax Error: tests/allan/Test.scala:4:13 ----------------------------
4 |  def test = "  " * 10
  |             ^^^^
  |             overloaded or recursive method test needs return type
```
Dotty is confused by the import

odersky:
When harmonizing repeated arguments we needed to backtrack the
constraint. It turns out we need to do the same thing in all other
situations where harmonization takes place.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -438,16 +438,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
                 addTyped(arg, formal)
               case _ =>
                 val elemFormal = formal.widenExpr.argTypesLo.head
-                val origConstraint = ctx.typerState.constraint
-                var typedArgs = args.map(typedArg(_, elemFormal))
-                val harmonizedArgs = harmonizeArgs(typedArgs)
-                if (harmonizedArgs ne typedArgs) {
-                  ctx.typerState.constraint = origConstraint
-                    // reset constraint, we will re-establish constraint anyway when we
-                    // compare against the seqliteral. The reset is needed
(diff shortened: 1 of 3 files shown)


odersky:
Fixes #3252

allanrenucci:
Actually scalac can harmonize even further:
```scala
def totalDuration(results: List[Long], cond: Boolean): Long =
  results.map { r =>
    if (cond) r
    else if (cond) 0
    else (1: Short)
  }.sum
```

allanrenucci:
In scalac:
```scala
scala> def foo(cond: Boolean) = if (cond) (0: Short) else if (cond) 1 else 2L
foo: (cond: Boolean)Long
```
In Dotty with this patch:
```scala
scala> def foo(cond: Boolean) = if (cond) (0: Short) else if (cond) 1 else 2L 
def foo(cond: Boolean): AnyVal
```

odersky:
Yes, the differences with `scalac` are as expected.

PR #3276, opened 2017-10-06
isAsSpecific: restrict usage of fullyDefinedType

smarter:
We only need to fully define the bounds of the PolyType parameters, not the
whole PolyType.


smarter:
test performance please

PR #3277, opened 2017-10-06
Fix #3206: make sure partial function has `@unchecked`

### LINKED ISSUE #3206: PartialFunction definitions should not generate exhaustiveness warning
This started happening recently, you can see lots of spurious warnings by doing `;dotty-bootstrapped/clean;dotty-bootstrapped/compiler`, here's a minimal example:
```scala
object Test {
  val foo: PartialFunction[Option[Int], Int] = {
    case Some(x) => x
  }
}
```
```scala
-- [E028] Pattern Match Exhaustivity Warning: try/pf2.scala:3:4 ----------------
3 |    case Some(x) => x
  |    ^
  |    match may not be exhaustive.
  |    
  |    It would fail on: None
```

Would be nice to get a fix for this before releasing 0.4.0-RC1 (/cc @allanrenucci)

PS: This is why we should try to get to zero warnings and enable `-Xfatal-warnings` as soon as possible :).

liufengyun:
Fix #3206: make sure partial function has `@unchecked`.

Review @smarter .

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2251,7 +2251,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
      *  tree that went unreported. A scenario where this happens is i1802.scala.
      */
     def ensureReported(tp: Type) = tp match {
-      case err: ErrorType if !ctx.reporter.hasErrors => ctx.error(err.msg, tree.pos)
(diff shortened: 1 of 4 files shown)


smarter:
Oh, I guess this broke with the native applied type PR, makes sense.

liufengyun:
@smarter could you have a look at the last commit, it's needed to pass the tests.

liufengyun:
Thanks a lot @smarter !

PR #3278, opened 2017-10-06
fix #3243 only publish nightlies when there is a commit less than 24 hours old

### LINKED ISSUE #3243: Don't publish the same nightly twice
We publish nightly on a daily basis. The nightly version is based on the latest commit on master:
```scala
val version = baseVersion + "-bin-" + VersionUtil.commitDate + "-" + VersionUtil.gitHash + "-NIGHTLY"
```
If there is no new commit since the previous nightly, we fill fail to publish a new artirfact because one with the same version already exists.

We can either change the version to reflect the publishing date, or don't do anything if there is no new commit since the previous nightly. @smarter WDYT?

Swilvan:
This seemed like a simple fix for the issue described.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -5,6 +5,8 @@ import java.io.{File, RandomAccessFile}
 import java.nio.channels.FileLock
 import java.nio.file.{ Files, FileSystemException }
 import java.util.Calendar
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 
 import scala.reflect.io.Path
 import sbtassembly.AssemblyKeys.assembly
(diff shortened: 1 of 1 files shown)


smarter:
I think we're more likely to want this fixed in the CI publish script (https://github.com/lampepfl/dotty/blob/master/project/scripts/sbtPublish) than in the build, but I'll let @allanrenucci decide.

allanrenucci:
I am OK with it being part of the build. But I'd rather have a solution that checks if the current version has already been published. You can use`dottyLatestNightlyBuild` for that.

PR #3280, opened 2017-10-06
sbt.ExtractAPI: Support TypeArgRef

smarter:
TypeArgRef can leak to method signatures, therefore it needs to be
handled in ExtractAPI for incremental compilation to work correctly.

@odersky :  See sbt-dotty/sbt-test/source-dependencies/typeargref/A.scala for an example where a TypeArgRef leaks to the signature. This example seems dubious to me since there's no reason we couldn't use the underlying type here, so let me know if you can come up with a more reasonable example.

### DIFF
--- a/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
+++ b/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
@@ -147,6 +147,7 @@ private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder
     new api.Annotated(tp, Array(marker))
   private def marker(name: String) =
     new api.Annotation(new api.Constant(Constants.emptyType, name), Array())
+  val typeArgRefMarker = marker("TypeArgRef")
   val orMarker = marker("Or")
   val byNameMarker = marker("ByName")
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
```diff
@@ -343,6 +344,12 @@ private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder
     }
   }
 
+  // Hack to represent dotty types which don't have an equivalent in xsbti
+  def combineApiTypes(apiTps: api.Type*): api.Type = {
```
odersky: I have no idea what this is supposed to achieve.

smarter: Basically fitting a square peg in a round hole: sbt has its own representation of types which doesn't include anything like TypeArgRef, but we can still get incremental compilation to work correctly if we can find a bijection between dotty's types and sbt's types, since the only thing that matter is that the hash of the sbt type changes when the dotty type changes. Don't worry too much about it :).

PR #3283, opened 2017-10-07
move 'extend final class' to error class

m-sp:


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1770,4 +1770,12 @@ object messages {
           |"""
   }
 
+  case class ExtendFinalClass(clazz:Symbol, finalClazz: Symbol)(implicit ctx: Context) 
+    extends Message(ExtendFinalClassID) {
+    val kind = "Syntax"
+    val msg = hl"$clazz cannot extend ${"final"} $finalClazz"
+    val explanation =
+      hl"""A class marked with the ${"final"} keyword cannot be extended by an object or a class"""
```
allanrenucci: I would simply say "... cannot be extended". For instance, a `trait` cannot extends a `final class` either.

m-sp:
Thanks for the review, I've addressed your points.
I hope squashing is fine!

PR #3285, opened 2017-10-08
Fix #3273: Cook raw types

### LINKED ISSUE #3273: Dotty gets confused by raw types in bytecode signatures
Consider the following file `Test.scala`:
```scala
import org.testng.IAnnotationTransformer
import org.testng.annotations.ITestAnnotation
import java.lang.reflect.Method
import java.lang.reflect.Constructor

class SingleTestAnnotationTransformer(testName: String) extends IAnnotationTransformer {
  override def transform( annotation: ITestAnnotation, testClass: java.lang.Class[_], testConstructor: Constructor[_], testMethod: Method): Unit = {}
}
```
Running
```
./bin/dotc -classpath "$(coursier fetch -p org.testng:testng:6.8.7)" Test.scala 
```
produces the following error:
```scala
-- [E036] Reference Error: Test.scala:9:15 -------------------------
9 |  override def transform( annotation: ITestAnnotation, testClass: java.lang.Class[_], testConstructor: Constructor[_], testMethod: Method): Unit = {
  |               ^
  |               method transform overrides nothing
```

odersky:
Replace Java raw types by existentials (wildcard applications).

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
+++ b/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
@@ -23,6 +23,19 @@ object ClassfileParser {
   /** Indicate that there is nothing to unpickle and the corresponding symbols can
     * be invalidated. */
   object NoEmbedded extends Embedded
+
+  /** Replace raw types with wildcad applications */
+  def cook(implicit ctx: Context) = new TypeMap {
+    def apply(tp: Type): Type = tp match {
+      case tp: TypeRef if tp.symbol.typeParams.nonEmpty =>
+        AppliedType(tp, tp.symbol.typeParams.map(Function.const(TypeBounds.empty)))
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -23,6 +23,19 @@ object ClassfileParser {
   /** Indicate that there is nothing to unpickle and the corresponding symbols can
     * be invalidated. */
   object NoEmbedded extends Embedded
+
+  /** Replace raw types with wildcard applications */
+  def cook(implicit ctx: Context) = new TypeMap {
+    def apply(tp: Type): Type = tp match {
+      case tp: TypeRef if tp.symbol.typeParams.nonEmpty =>
```
smarter: Micro-optimization: prefer !isEmpty, `nonEmpty` is a trait method

PR #3239, opened 2017-10-03
Fix #3213: Make PolyTypes generative

odersky:
Because PolyTypes are used as carriers of constrained parameter references,
we must make sure that we can duplicate them so that the duplicate is
different from the original. Hence, it's useless to cache PolyTypes.

PR #3287, opened 2017-10-08
Fix #3250: Make HKLambda a cached type

### LINKED ISSUE #3250: Documentation generation broken by merging #3239 with typer error
Somehow, merging #3239  broke the documentation generation with the following error:
```scala
-- Error: /drone/src/github.com/lampepfl/dotty/doc-tool/src/dotty/tools/dottydoc/model/comment/HtmlParsers.scala:96:103 
96 |          s"<dl>${items map { case (t, d) => s"<dt>${inlineToHtml(t)}</dt><dd>${blockToHtml(d)}</dd>" } }</dl>"
   |                                                                                                       ^
   |ambiguous implicits: both method OnceCanBuildFrom in object TraversableOnce and method canBuildFrom in object GenIterable match type scala.collection.generic.CanBuildFrom[
   |  scala.collection.SortedMap[dotty.tools.dottydoc.model.comment.Inline, 
   |    dotty.tools.dottydoc.model.comment.Block
   |  ]
|, String, That] of parameter bf of method map in trait TraversableLike
```
(see http://dotty-ci.epfl.ch/lampepfl/dotty/627/8), you can reproduce the failure by running:
```scala
sbt genDocs
```

odersky:
This is a partial revert from #3239. #3239 made both HKLambdas and
PolyTypes uncached. It turns out that this breaks things if we do it
for HKLambdas. I have not yet figured out why, however. The test case
in #3250 is a bit too large to be able to tell.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -19,7 +19,7 @@ import Periods._
 import Designators._
 import util.Positions.{Position, NoPosition}
 import util.Stats._
-import util.{DotClass, SimpleMap}
+import util.DotClass
 import reporting.diagnostic.Message
 import reporting.diagnostic.messages.CyclicReferenceInvolving
 import ast.tpd._
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2599,7 +2599,32 @@ object Types {
     final override def toString = s"$prefixString($paramNames, $paramInfos, $resType)"
   }
 
-  trait HKLambda extends LambdaType
+  abstract class HKLambda extends CachedProxyType with LambdaType {
+    final override def underlying(implicit ctx: Context) = resType
+
+    final override def computeHash = doHash(paramNames, resType, paramInfos)
+
+    final override def equals(that: Any) = that match {
+      case that: HKLambda =>
+        paramNames == that.paramNames &&
+        paramInfos == that.paramInfos &&
+        resType == that.resType &&
+        companion.eq(that.companion)
+      case _ =>
+        false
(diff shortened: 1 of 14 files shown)


odersky:
test performance please

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2599,7 +2599,32 @@ object Types {
...
+        companion.eq(that.companion)
+      case _ =>
+        false
+    }
+
+    final override def eql(that: Type) = that match {
+      case that: HKLambda =>
+        paramNames.equals(that.paramNames) &&
+        paramInfos.equals(that.paramInfos) &&
+        resType.equals(that.resType) &&
```
smarter: That was already the case before #3239, but why is this equals and not eq?

compiler/src/dotty/tools/dotc/util/SimpleMap.scala
```diff
@@ -41,13 +41,13 @@ object SimpleMap {
   class Map1[K <: AnyRef, +V >: Null <: AnyRef] (k1: K, v1: V) extends SimpleMap[K, V] {
```
smarter: Above, there is a definition `def contains(k: K): Boolean = apply(k) != null` which could use `ne` instead of `!=`.

odersky: For null comparisons `!=` and `ne` produce exactly the same bytecode.

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: I am not even sure, `equals` is needed. But the idea is that, since many types get a compiler generated `equals` since they are case classes, the rest of the types would conform to that.

PR #3289, opened 2017-10-08
Avoid creating NumericRange instances when not needed

nicolasstucki:
Initializing NumericRange instances is expensive due to the implicit class tags used.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -111,21 +111,21 @@ class Definitions {
         val decls = newScope
         val arity = name.functionArity
         val paramNamePrefix = tpnme.scala_ ++ str.NAME_JOIN ++ name ++ str.EXPAND_SEPARATOR
-        val argParams =
-          for (i <- List.range(1, arity + 1)) yield
-            enterTypeParam(cls, paramNamePrefix ++ "T" ++ i.toString, Contravariant, decls)
+        val argParams = List.tabulate(arity) { i =>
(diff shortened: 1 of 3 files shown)


nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -111,21 +111,21 @@ class Definitions {
         val decls = newScope
         val arity = name.functionArity
         val paramNamePrefix = tpnme.scala_ ++ str.NAME_JOIN ++ name ++ str.EXPAND_SEPARATOR
-        val argParams =
-          for (i <- List.range(1, arity + 1)) yield
-            enterTypeParam(cls, paramNamePrefix ++ "T" ++ i.toString, Contravariant, decls)
+        val argParams = List.tabulate(arity) { i =>
```
odersky: Rename to argParamRefs?

compiler/src/dotty/tools/dotc/util/SourcePosition.scala
```diff
@@ -26,11 +26,12 @@ extends interfaces.SourcePosition {
...
-  def lines: List[Int] =
-    List.range(source.offsetToLine(start), source.offsetToLine(end + 1)) match {
-      case Nil => line :: Nil
-      case xs => xs
-    }
+  def lines: List[Int] = {
+    val startOffset = source.offsetToLine(start)
+    val endOffset = source.offsetToLine(end + 1)
+    if (startOffset >= endOffset) line :: Nil
+    else List.tabulate(endOffset - startOffset)(i => i + startOffset)
```
smarter: Removing the offset to add it back is confusing, I would just keep the original code here since it shouldn't be performance-sensitive, or do `(startOffset to endOffset).toList` maybe.

PR #3290, opened 2017-10-08
Avoid closure creation in hot code

nicolasstucki:
Unfortunately the `inlining` solution of  #3267 fails here as the closure is impure an hence evaluated before the inlined block of code.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Flags.scala
+++ b/compiler/src/dotty/tools/dotc/core/Flags.scala
@@ -22,7 +22,8 @@ object Flags {
       else if (that.bits == 0) this
       else {
         val tbits = bits & that.bits & KINDFLAGS
-        assert(tbits != 0, s"illegal flagset combination: $this and $that")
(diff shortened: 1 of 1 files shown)


odersky:
In what sense is the closure impure?

nicolasstucki:
After some investigation I discovered the issue, it has nothing to do with pureness. 

The problem is that `DottyPredef.assert` is only inlined when the code is compiled by `dotty-bootstrapped` (and not by `dotty-nonbootstrapped`). As `dotty-bootstrapped` is compiled with `dotty-nonbootstrapped` it does not get the `asserts` inlined. We would need a `dotty-bootstrapped-bootstrapped` to have this working.

I would suggest adding special casing this assert and adding the optimization by hand as it is in hot code. We should discuss what to do with the other ones later.

smarter:
Or you could compile the bootstrapped compiler with the non-bootstrapped compiler as usual but with the bootstrapped library on the Scala classpath

PR #3295, opened 2017-10-09
Local optimisation improvements  

gan74:
* Valify rewritten to be a lot more aggressive. 
 * RemoveUnnecessaryNullChecks rewritten to also remove redundant null checks.
 * A few new cases to ConstantFold.


smarter:
Hi! It's great that you added a bunch of tests, but It'd be even more helpful if your commit message contained details on what was done. It'd be even better if you could split your contribution into several commits, each doing one well defined thing.

PR #3296, opened 2017-10-09
Fix Missing error message on FailureToEliminateExistential

CucumisSativus:
One step forward with #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1821,9 +1821,21 @@ object messages {
 
   case class TailrecNotApplicable(method: Symbol)(implicit ctx: Context)
     extends Message(TailrecNotApplicableID) {
-      val kind = "Syntax"
-      val msg = hl"TailRec optimisation not applicable, $method is neither ${"private"} nor ${"final"}."
-      val explanation =
-       hl"A method annotated ${"@tailrec"} must be declared ${"private"} or ${"final"} so it can't be overridden."
+    val kind = "Syntax"
+    val msg = hl"TailRec optimisation not applicable, $method is neither ${"private"} nor ${"final"}."
+    val explanation =
+      hl"A method annotated ${"@tailrec"} must be declared ${"private"} or ${"final"} so it can't be overridden."
+  }
+
+  case class FailureToEliminateExistential(tp: Type, tp1: Type, tp2: Type, boundSyms: List[Symbol])(implicit ctx: Context)
+    extends Message(FailureToEliminateExistentialID) {
+    val kind = "Compatibility"
+    val msg = hl"Failure to eliminate existential type. Proceed at own risk."
+    private val originalType = ctx.dclsText(boundSyms, "; ").show
+    val explanation =
+      hl"""original type    : $tp forSome {${originalType}
+          |reduces to       : $tp1
+          |type used instead: $tp2
+          |"""
   }
 }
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1778,4 +1778,15 @@ object messages {
       hl"""A class marked with the ${"final"} keyword cannot be extended"""
   }
 
+  case class FailureToEliminateExistential(tp: Types.Type, tp1: Types.Type, tp2: Types.Type, boundSyms: List[Symbols.Symbol])(implicit ctx: Context)
+    extends Message(FailureToEliminateExistentialID) {
+    val kind = "Syntax"
```
allanrenucci: This isn't a syntax related message. Maybe: `Compatibility`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1778,4 +1778,15 @@ object messages {
       hl"""A class marked with the ${"final"} keyword cannot be extended"""
   }
 
+  case class FailureToEliminateExistential(tp: Types.Type, tp1: Types.Type, tp2: Types.Type, boundSyms: List[Symbols.Symbol])(implicit ctx: Context)
```
allanrenucci: Simply use `Type` instead of `Types.Type` and `Symbol` instead of `Symbols.Symbol`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1778,4 +1778,15 @@ object messages {
...
+    extends Message(FailureToEliminateExistentialID) {
+    val kind = "Syntax"
+    val msg = hl"failure to eliminate existential"
```
allanrenucci: `hl"Failure to eliminate existential type. Proceed at own risk."`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1778,4 +1778,15 @@ object messages {
...
+    val explanation =
+      hl"""original type    : $tp forSome {${ctx.dclsText(boundSyms, "; ").show}
+          |reduces to       : $tp1
+          |type used instead: $tp2
+          |proceed at own risk.
```
allanrenucci: I think this is better as part of the `msg`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1778,4 +1778,15 @@ object messages {
       hl"""A class marked with the ${"final"} keyword cannot be extended"""
   }
 
+  case class FailureToEliminateExistential(tp: Types.Type, tp1: Types.Type, tp2: Types.Type, boundSyms: List[Symbols.Symbol])(implicit ctx: Context)
+    extends Message(FailureToEliminateExistentialID) {
+    val kind = "Syntax"
+    val msg = hl"failure to eliminate existential"
+    val explanation =
+      hl"""original type    : $tp forSome {${ctx.dclsText(boundSyms, "; ").show}
```
allanrenucci: I would extract original type in order to fully syntax highlight it:
```scala
val explanation = {
  val originalTpe = s"$tp forSome {${ctx.dclsText(boundSyms, "; ").show}"
  hl"""original type    : $originalTpe
  ...
}
```

CucumisSativus:
Thanks for code review @allanrenucci, i'll rebase and fix the issuses later this week

CucumisSativus:
Hey @allanrenucci can you lend me a hand with writing the test as such? I was moving through the documentation but I could find the documented case when such warning should appear. Any help with this would be very welcome

allanrenucci:
@smarter I looked at our test suite and I can't find any test that exercises this warning. Could you give us an example?

smarter:
You'll need to compile some code with scalac (or find some existing code) that uses an existential type that cannot be expressed with wildcards, e.g. `Array[Array[T]] forSome { type T }`

CucumisSativus:
thanks @allanrenucci !

PR #3300, opened 2017-10-10
Fix #3246: Refine handling of postfix _ for non-functions

### LINKED ISSUE #3246: Converting from `T` to `() => T`
With scalac you can do
```scala
val foo = 1
val bar: () => Int = foo _
```
dotc reject this code snippet. Should we support it under `-language:Scala2` and emit a migration warning?

odersky:


odersky:
@allanrenucci All patterns you showed in #3246 should now be handled without error under -language:Scala2

allanrenucci:
```scala
class Test {
  def foo(fun: () => Unit) = ???
  def bar(fun: => Any) = foo(fun _)
}
```
This test case still does not compile with Dotty

odersky:
Yes, but why should it compile? I don't see how `fun _` could expand to anything but a `() => Any`. I think this is a bug in scalac.

allanrenucci:
I agree it shouldn't compile. But then, should the following code compile?
```scala
class Test {
  def foo(fun: () => Unit) = ???
  def bar(fun: => Any) = foo(() => fun)
}
```

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1547,17 +1547,19 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   def typedAsFunction(tree: untpd.PostfixOp, pt: Type)(implicit ctx: Context): Tree = {
     val untpd.PostfixOp(qual, Ident(nme.WILDCARD)) = tree
     val pt1 = if (defn.isFunctionType(pt)) pt else AnyFunctionProto
-    var res = typed(qual, pt1)
-    if (pt1.eq(AnyFunctionProto) && !defn.isFunctionClass(res.tpe.classSymbol)) {
-      ctx.errorOrMigrationWarning(i"not a function: ${res.tpe}; cannot be followed by `_'", tree.pos)
+    val nestedCtx = ctx.fresh.setNewTyperState()
+    var res = typed(qual, pt1)(nestedCtx)
```
allanrenucci: Can be a `val` now

allanrenucci:
As a side effect of this PR, we now accept this code under `-language:Scala2` (scalac rejects it)
```scala
class Test {
  def foo(x: Int) = 1
  val bar: () => Int = foo(1) _
}
```
It is definitely an improvement as we used to emit a type error:
```scala
-- [E007] Type Mismatch Error: tests/allan/Test.scala:3:26 ---------------------
3 |  val bar: () => Int = foo(1) _
  |                       ^^^^^^
  |                       found:    Int
  |                       required: () => Int
  |                       

one error found
```
Now:
```scala
-- Migration Warning: tests/allan/Test.scala:3:30 ------------------------------
3 |  val bar: () => Int = foo(1) _
  |                       ^^^^^^^^
  |                       not a function: foo(1); cannot be followed by `_'
```

allanrenucci:
```scala
class Test {
  def foo = 1
  val bar: () => Int = foo _
}
```
```scala
-- Migration Warning: tests/allan/Test.scala:3:27 ------------------------------
3 |  val bar: () => Int = foo _
  |                       ^^^^^
  |                       not a function: foo; cannot be followed by `_'
```
I think the warning is misleading in this case. If foo is defined as `def foo()`, then no warning will be emitted at all, even though the `_` is not needed. I think we should depreciate `_` in all cases.

allanrenucci:
```scala
class Test {
  def foo(x: Int) = 1
  val bar: () => Int = foo _
}
```
This now crashes under `-language:Scala2`
```
> dotc -d out tests/allan/Test.scala -language:Scala2 -migration
[info] Running dotty.tools.dotc.Main -d out tests/allan/Test.scala -language:Scala2 -migration -classpath /Users/renucci/Documents/Projects/dotty/library/target/scala-2.12/dotty-library_2.12-0.4.0-bin-SNAPSHOT-nonbootstrapped.jar
-- Migration Warning: tests/allan/Test.scala:3:27 ------------------------------
3 |  val bar: () => Int = foo _
  |                       ^^^^^
  |                       not a function: foo; cannot be followed by `_'

exception occurred while typechecking tests/allan/Test.scala

	at scala.Predef$.assert(Predef.scala:204)
	at dotty.tools.dotc.ast.Trees$Tree.withType(Trees.scala:121)
	at dotty.tools.dotc.typer.Typer.$anonfun$typedTypeTree$1(Typer.scala:1071)
	at dotty.tools.dotc.typer.Typer.typedTypeTree(Typer.scala:1056)
	at dotty.tools.dotc.typer.Typer.typedUnnamed$1(Typer.scala:1653)
	at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:1680)
	at dotty.tools.dotc.typer.Typer.$anonfun$typed$2(Typer.scala:1697)
	at dotty.tools.dotc.reporting.Reporting.traceIndented(Reporter.scala:140)
	at dotty.tools.dotc.reporting.Reporting.traceIndented$(Reporter.scala:139)
	at dotty.tools.dotc.typer.Typer.typed(Typer.scala:1693)
	at dotty.tools.dotc.typer.Namer.$anonfun$typedAheadType$1(Namer.scala:940)
	at dotty.tools.dotc.typer.Namer.typedAheadImpl(Namer.scala:933)
	at dotty.tools.dotc.typer.Namer.typedAheadType(Namer.scala:940)
	at dotty.tools.dotc.typer.Namer.valOrDefDefSig(Namer.scala:1116)
	at dotty.tools.dotc.typer.Namer$Completer.typeSig(Namer.scala:732)

```

tests/neg/i3246.scala
```diff
@@ -0,0 +1,4 @@
+class Test {
```
allanrenucci: I think we should also compile it under scala2 mode, since it only crashed in this mode

odersky: OK. Can you make the change and merge the PR when done? Thanks!

allanrenucci: Sure. Will do

allanrenucci:
Superseded by #3402

PR #3306, opened 2017-10-11
Specialize Functions

Duhemm:
Supersedes #1807 by @felixmulder 

What I changed compared to the previous version:

 - The previous version would change the parent of a class extending `FunctionN` to the specialised version, if it existed. I encountered several issues doing that, and my investigation led me to believe that doing so is not necessary in this particular case. From my reading of @dragos's [thesis, §4.4.2 "Specialized inheritance"](https://infoscience.epfl.ch/record/150270/files/EPFL_TH4820.pdf), it appears that changing the parent to the specialised version is important to have the child inherit the specialised functions rather than the generic ones. My understanding is that this is not necessary in this particular case, please let me know if I am misunderstanding the issue.

 - The optimisation wouldn't apply on by-name parameters, because it would run before `ElimByName`. I didn't manage to shuffle the phases around without adding a new traversal.


Duhemm:
test performance please

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+      var newApplys = Map.empty[Name, Symbol]
+
+      tp.parents.foreach { parent =>
+        List(0, 1, 2).foreach { arity =>
+          val func = defn.FunctionClass(arity)
+          if (parent.derivesFrom(func)) {
+            val typeParams = tp.cls.typeRef.baseType(func).argInfos
+            val interface  = specInterface(typeParams)
+
+            if (interface.exists && tp.decls.lookup(nme.apply).exists) {
```
smarter: Instead of creating a class symbol with `specInterface` and checking if it actually exists, you can use `isSpecializableFunction` like in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/FunctionalInterfaces.scala#L39

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+  override def transformTemplate(tree: Template)(implicit ctx: Context, info: TransformerInfo) = {
+    val applyBuf = new mutable.ListBuffer[Tree]
+    val newBody = tree.body.mapConserve {
+      case dt: DefDef if dt.name == nme.apply && dt.vparamss.length == 1 =>
+        val specName = nme.apply.specializedFunction(
+          dt.tpe.widen.finalResultType,
+          dt.vparamss.head.map(_.symbol.info)
+        )
+
+        val specializedApply = tree.symbol.enclosingClass.info.decls.lookup(specName)
```
smarter: Same thing here and in transformApply, you should only have to do the lookup if `isSpecializableFunction` returns true

compiler/src/dotty/tools/dotc/transform/SpecializeFunctions.scala
```diff
@@ -0,0 +1,139 @@
...
+          }
+
+          newSel.appliedToArgs(args)
+        }
+        else tree
+
+      case _ => tree
+    }
+
+  @inline private def specializedName(name: Name, args: List[Type])(implicit ctx: Context) =
```
smarter: I don't think there's any reason for this method and the next one to be marked @inline

compiler/src/dotty/tools/dotc/transform/SpecializedApplyMethods.scala
```diff
@@ -0,0 +1,96 @@
...
+      r  <- List(UnitType, BooleanType, IntType, FloatType, LongType, DoubleType)
+      t1 <- List(IntType, LongType, FloatType, DoubleType)
+    } yield specApply(func1, List(t1), r)
+
+    func2 = FunctionClass(2)
+    func2Applys = for {
+      r  <- List(UnitType, BooleanType, IntType, FloatType, LongType, DoubleType)
+      t1 <- List(IntType, LongType, DoubleType)
+      t2 <- List(IntType, LongType, DoubleType)
+    } yield specApply(func2, List(t1, t2), r)
```
smarter: This duplicates some logic from `isSpecializableFunction`

smarter:
Most of the performance cost is probably from splitting a phase group to add SpecializeFunctions. Would be great if we could figure out a way to avoid that.

Duhemm:
I think I managed to re-unify the group I split 🎉

Duhemm:
test performance please

liufengyun:
test performance please

liufengyun:
test performance please

(meta test for bench infrastructure change)

liufengyun:
test performance please

(meta test for bench infrastructure change)

Duhemm:
test performance please

smarter:
Looks like this PR still affects performance, even the empty file benchmark seems to get slower. It'd be good to try to figure out where the extra time is being spent and if we can optimize this.

Duhemm:
test performance please

PR #3307, opened 2017-10-11
Fix #3304: Launch script now requires Java 8

### LINKED ISSUE #3304: Dotty 0.4.0-RC1 shell scripts should detect Java 9 and stop with an error
Dotty won't be compatible with Java 9 until https://github.com/lampepfl/dotty/pull/3138 is done. To avoid confusing people with weird errors, we should check the Java version in the shell script and ask people to use Java 8. To get the Java version use https://github.com/sbt/sbt-launcher-package/blob/master/src/universal/bin/sbt-launch-lib.bash

allanrenucci:


dist/bin/common
```diff
@@ -61,6 +61,14 @@ if [ ! -x "$JAVACMD" ] ; then
   exit 1
 fi
 
+# parses 1.7, 1.8, 9, etc out of java version "1.8.0_91"
+JAVA_VERSION=$("$JAVACMD" -Xmx512M -version 2>&1 | grep ' version "' | sed 's/.*version "\([0-9]*\)\(\.[0-9]*\)\{0,1\}\(.*\)*"/\1\2/; 1q')
+
+if [ "$JAVA_VERSION" != "1.8" ] ; then
+  echo "Error: Java 8 is required, actual $JAVA_VERSION"
```
smarter: I think it'd be nice if the message explained that a different version of Java could be set by using the JAVA_HOME environment variable.

dist/bin/common
```diff
@@ -61,6 +61,14 @@ if [ ! -x "$JAVACMD" ] ; then
   exit 1
 fi
 
+# parses 1.7, 1.8, 9, etc out of java version "1.8.0_91"
+JAVA_VERSION=$("$JAVACMD" -Xmx512M -version 2>&1 | grep ' version "' | sed 's/.*version "\([0-9]*\)\(\.[0-9]*\)\{0,1\}\(.*\)*"/\1\2/; 1q')
```
liufengyun: A minor note: why `-Xmx512M`, it seems it's not necessary. But to be safer, let's just keep it as it is.

allanrenucci: Copy pasted form https://github.com/sbt/sbt-launcher-package/blob/master/src/universal/bin/sbt-launch-lib.bash#L183

allanrenucci: But I suppose it is not needed here

PR #3312, opened 2017-10-12
Add new style error message for @tailrec 

esarbe:
see #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1777,6 +1777,7 @@ object messages {
     val explanation =
       hl"""A class marked with the ${"final"} keyword cannot be extended"""
   }
+
   case class EnumCaseDefinitionInNonEnumOwner(owner: Symbol)(implicit ctx: Context)
     extends Message(EnumCaseDefinitionInNonEnumOwnerID) {
       val kind = "Syntax"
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1786,4 +1787,15 @@ object messages {
             |If you want to create an ${"enum"} case, make sure the corresponding ${"enum class"} exists
             |and has the ${"enum"} keyword."""
   }
+
+
+  case class TailrecNotApplicableNeitherPrivateNorFinal(owner: Symbol)(implicit ctx: Context)
```
allanrenucci: I would simply call it `TailrecNotApplicable` and `TailrecNotApplicableID`

allanrenucci: `method` is a better name than `owner`

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -1023,5 +1023,21 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+
+  @Test def tailrecNotApplicableNeitherPrivateNorFinal =
+    checkMessagesAfter("tailrec") {
+    """
+      |class Foo {
+      |
+      |  @scala.annotation.tailrec
+      |  def foo: Unit = foo
+      |
+      |}
```
allanrenucci: This file is already big enough. Let's remove unnecessary blank lines:
```scala
    """class Foo {
      |  @scala.annotation.tailrec
      |  def foo: Unit = foo
      |}
    """.stripMargin
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1786,4 +1787,15 @@ object messages {
...
+    extends Message(TailrecNotApplicableNeitherPrivateNorFinalID) {
+      val kind = "Syntax"
+      val msg = hl"TailRec optimisation not applicable, ${owner} is neither `private` nor `final` so can be overridden"
```
```scala
val msg = hl"TailRec optimisation not applicable, $owner is neither ${"private"} nor ${"final"}."
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1786,4 +1787,15 @@ object messages {
...
+      val explanation =
+        hl"""
+            |The `@tailrec` annotation can only be used on methods that are either `private` or `final`.
+          """
```
```scala
val explanation =
  hl"A method annotated ${"@tailrec"} must be declared ${"private"} or ${"final"} so it can't be overridden."
```

PR #3313, opened 2017-10-12
Fix indentation

nicolasstucki:
Just because it is messing with the IDE indentation lines.


allanrenucci:
@lampepfl/dotty-core Do we want to automate formatting with Scalafmt?

nicolasstucki:
Can we use only Scalafmt on the modified code? To avoid messing with the git history/blame.

PR #3315, opened 2017-10-12
Investigate build failure

allanrenucci:
Reopening #3301 from staging. cc/ @liufengyun


liufengyun:
@allanrenucci I just pulled the docker image and run locally, the exception is due to the missing command `tput`, which is used here: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/repl/terminal/Utils.scala#L77

However, this error message is different from the one produced in the CI.

liufengyun:
The image I'm using is `lampepfl/dotty:2017-10-07`.

Issue #3028, opened 2017-08-28
Unable to parse Java raw types


julienrf:
The following Java code is parsed and compiled by javac but dotc rejects it:

~~~ java
abstract class INodeBase<K, V> extends BasicNode {
    @SuppressWarnings("rawtypes")
    public static final AtomicReferenceFieldUpdater<INodeBase, MainNode> updater =
            AtomicReferenceFieldUpdater.newUpdater(INodeBase.class, MainNode.class, "mainnode");
}
~~~

The errors produced by dotc are the following:

~~~
[error] -- Error: /home/julien/workspace/dev/scala/collection-strawman/src/main/scala/strawman/collection/concurrent/INodeBase.java:18:52 
[error] 18 |    public static final AtomicReferenceFieldUpdater<INodeBase, MainNode> updater =
[error]    |                                                    ^^^^^^^^^
[error]    |Type argument strawman.collection.concurrent.INodeBase has not the same kind as its bound 
[error] -- Error: /home/julien/workspace/dev/scala/collection-strawman/src/main/scala/strawman/collection/concurrent/INodeBase.java:18:63 
[error] 18 |    public static final AtomicReferenceFieldUpdater<INodeBase, MainNode> updater =
[error]    |                                                               ^^^^^^^^
[error]    |Type argument strawman.collection.concurrent.MainNode has not the same kind as its bound 
~~~

The full code is available [here](https://github.com/scala/collection-strawman/blob/trie-map/src/main/scala/strawman/collection/concurrent/INodeBase.java).

julienrf:
I’m closing the issue because the code provided in the description now compiles. I can create a PR that adds a test, if you want.

PR #3317, opened 2017-10-12
Fix #3273: Extend cooking to parent types

### LINKED ISSUE #3273: Dotty gets confused by raw types in bytecode signatures

odersky:


tests/pos/i3273/client_2.scala
```diff
@@ -0,0 +1,5 @@
+class Client {
+  val a = (_: Bar_1.A[AnyRef]).foo()
+  val b = (_: Bar_1.B[AnyRef]).bar()
+  def test(x: Bar_1.A[AnyRef]): Bar_1.B[_] = x
+}
```
retronym: It would be good to test these cases under joint compilation, too (usually changes to classfile parser need to be mirrored in `JavaParsers`)

retronym:
I'm guessing some more work will be needed to deal with raw types in type parameter bound position.

smarter:
From the last commit message:
> if we compile Java files from sources we should demand that raw types are re-written as wildcard types.

Unfortunately this isn't always easy, because of `.class` and `.getClass`: `List.class` will return a `Class<? extends List>` and you cannot directly cast this to a `Class<List<?>>`, an intermediate cast is necessary:
```java
(Class<List<?>>) (Class<?>) List.class
```

Something like this was necessary to compile the strawman: https://github.com/lampepfl/dotty/issues/3028

smarter:
Of course, it's possible that the 13 years old bug report on this issue will be fixed one day: https://bugs.openjdk.java.net/browse/JDK-6184881 :)

odersky:
I have decided that as far as I am concerned I won't spend anymore cycles on this. If somebody really wants to use raw types in jointly compiled Java code, asking them to do the rewrite is justified even if it might be difficult.

PR #3319, opened 2017-10-13
Disable sbt scripted that doesn't work on the CI

allanrenucci:
Also make sbt scripted tests less verbose


allanrenucci:
>Do we have an open issue for this?

No, will open one

Issue #3208, opened 2017-09-28
typed patterns get more precise types in scalac than dotty


allanrenucci:
The following code snippet fails to compile
```scala
class Test {
  trait E
  trait Marker

  def test(es: List[E]): List[E] = es.collect { case e: Marker => e }
}
```
```scala
-- Error: tests/allan/Test.scala:5:69 ------------------------------------------
5 |  def test(es: List[E]): List[E] = es.collect { case e: Marker => e }
  |                                                                     ^
  |Cannot construct a collection of type That
  |
  |where:    That is a type variable with constraint <: scala.collection.immutable.List[Test.this.E]
  | with elements of type Test.this.Marker based on a collection of type scala.collection.immutable.List[Test.this.E].
```

smarter:
It works if you replace `case e: Marker` by `case e: E with Marker`. In scalac, `e` gets the type `E with Marker` in the original code, but in dotty it only gets the type `Marker`, I don't know what the intended/specified behavior is, but it seems that the scalac behavior is not completely consistent: https://users.scala-lang.org/t/type-inference-in-pattern-matching/1708

smarter:
Had a quick look at the spec and I couldn't find anything substantiating the current scalac behavior, in particular http://www.scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#typed-patterns says:
> A typed pattern x:T consists of a pattern variable x and a type pattern T. The type of x is the type pattern T, [...]

@odersky Are you OK with replicating the behavior of scalac here? Here's a simple example illustrating the difference:
```scala
trait X
trait Y

object Test {
  def foo(x: X) = x match {
    case y: Y =>
      val b: X = y // compiles with scalac, type mismatch with dotty
  }
}
```

odersky:
Pattern matching desperately needs a spec. It would be good if someone could start working on the issue. I think the `scalac` behavior on the test cases makes sense, though.

allanrenucci:
Here is another example that does not compile with Dotty, I found in the community build:
```scala
trait MyError { this: Throwable =>
  ...
}


try ...
catch {
  case e: MyError =>
    throw e
}
```

PR #3320, opened 2017-10-13
Remove warnings

nicolasstucki:


compiler/src/dotty/tools/repl/terminal/Utils.scala
```diff
@@ -178,8 +178,8 @@ object FrontEndUtils {
   def transpose[A](xs: List[List[A]]): List[List[A]] = {
     @tailrec def transpose(xs: List[List[A]], result: List[List[A]]): List[List[A]] = {
       xs.filter(_.nonEmpty) match {
-        case Nil =>  result
-        case ys: List[List[A]] => transpose(ys.map(_.tail), ys.map(_.head) :: result)
```
allanrenucci: This look like a bug. Scalac doesn't report a warning

nicolasstucki: You are right, we already know that the list is a list of list of A. There is no runtime check needed.

allanrenucci: It might due to #3208

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1523,9 +1523,9 @@ object Types {
           if (mySig == null) mySig = Signature.NotAMethod
         case designator: Symbol =>
           uncheckedSetSym(designator)
-        case LocalName(underlying, space) =>
```
allanrenucci: Is the warning valid here?

compiler/src/dotty/tools/repl/terminal/Utils.scala
nicolasstucki: Probably not, this is just a warning about erasure.

compiler/src/dotty/tools/dotc/core/Types.scala
nicolasstucki: Yes

compiler/src/dotty/tools/repl/terminal/Utils.scala
nicolasstucki: I created #3323

nicolasstucki: Though I would merge this change as the `: List[List[A]]` is absolutely useless.

compiler/src/dotty/tools/dotc/core/Types.scala
allanrenucci: Can you explain why? =)

nicolasstucki:
```scala
[warn] -- Warning: /Users/nicolasstucki/GitHub/dotty/compiler/src/dotty/tools/dotc/core/Types.scala:1543:22 
[warn] 1543 |        case LocalName(underlying, space) =>
[warn]      |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]      |There is no best instantiation of pattern type dotty.tools.dotc.core.Designators.LocalName[dotty.tools.dotc.core.Names.Name]
[warn]      |that makes it a subtype of selector type dotty.tools.dotc.core.Designators.Designator.
[warn]      |Non-variant type variable N cannot be uniquely instantiated.
[warn]      |(This would be an error under strict mode)
```

We are pattern matching on a `Designator { type ThisType  <: Name }` but we try to match it against a `LocalName[N <: Name] { type ThisType = N }`. Note that the designator does not define the `N` type parameter. With these constraints we could instantiate `N = Typename` or `N = TermName`.

doc-tool/src/dotty/tools/dottydoc/model/references.scala
```diff
@@ -67,6 +61,7 @@ object references {
         s"$title: $byName${ref.showReference}$repeated"
 
       case ConstantReference(title) => title
+      case EmptyReference => ""
```
allanrenucci: Does it ever happen?

nicolasstucki: I did no happen so far. Will put an assert

PR #3331, opened 2017-10-16
fix version check in Cygwin

liufengyun:
Fix version check in Cygwin

Note: I've tried following code as well, which doesn't work either on Cygwin:

``` shell
JAVA_VERSION=$(echo $JAVA_VERSION | xargs)
# or
JAVA_VERSION=$(echo $JAVA_VERSION)
```

I've tried compare `$JAVA_VERSION == "1.8 "`, which is false as well. I guess Cygwin has inserted some invisible characters somewhere.

### DIFF
--- a/dist/bin/common
+++ b/dist/bin/common
@@ -62,7 +62,7 @@ if [ ! -x "$JAVACMD" ] ; then
 fi
 
 # parses 1.7, 1.8, 9, etc out of java version "1.8.0_91"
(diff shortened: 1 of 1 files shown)


allanrenucci:
Actually I don't think there is a trailing whitespace. All the line in the CI log have a space at the end. Have you just tried `[[ "$JAVA_VERSION" != "1.8" ]]`?

liufengyun:
Yes, I also tried just use double brackets, it didn't work either.

liufengyun:
The CI shows it works for MingW and Cygwin.

https://github.com/liufengyun/packtest/pull/5

PR #3334, opened 2017-10-17
Eliminate AnnotationTransformer

odersky:
AnnotationTransformers were introducing quite a bit of irregularities

 - they designated only a few phases, but not others to run over annotations
 - they were forcing these phases to run at phase itself instead of phase.next,
    as is the standard.
 - the phase skew was enforced even for the normal run of phases that transformed
    annotations, not just when they transformed some annotation tree.

In the end, the only reason to transform annotation trees at all is to eliminate a couple of
tree forms when annotations got emitted: The eliminated forms were NamedArgs and 
Typed trees representing _* arguments. 

I believe it's conceptually simpler never to transform annotation trees and to deal with
the two added tree forms in emitArguments.

That way all transformations now run at phase + 1.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala
@@ -25,7 +25,7 @@ import TypeUtils._
 /** A transformer that removes repeated parameters (T*) from all types, replacing
  *  them with Seq types.
  */
-class ElimRepeated extends MiniPhaseTransform with InfoTransformer with AnnotationTransformer { thisTransformer =>
+class ElimRepeated extends MiniPhaseTransform with InfoTransformer { thisTransformer =>
   import ast.tpd._
 
   override def phaseName = "elimRepeated"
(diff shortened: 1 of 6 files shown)


DarkDimius:
Looks good to me.

compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala
```diff
@@ -74,9 +74,12 @@ class ElimRepeated extends MiniPhaseTransform with InfoTransformer with Annotati
     transformTypeOfTree(tree)
 
   override def transformApply(tree: Apply)(implicit ctx: Context, info: TransformerInfo): Tree = {
-    val formals = (tree.fun.tpe.widen: @unchecked) match {
-      case mt: MethodType => mt.paramInfos
-    }
+    val formals =
+      ctx.atPhase(thisTransformer) { implicit ctx =>
+        (tree.fun.tpe.widen: @unchecked) match {
```
allanrenucci: nitpick: `val mt = tree.fun.tpe.widen.asInstanceOf[MethodType]` instead?

odersky: good point.

PR #3335, opened 2017-10-17
Remove Scope.filter(...).toList

nicolasstucki:


nicolasstucki:
Note that `Scope.filter` already returns a `List[Symbol]`.

Issue #3340, opened 2017-10-17
null asInstance of Nothing not always throwing


nicolasstucki:
In the following code the cast to Nothing does not throw the expected `NullPointerException`
```scala
def f: Unit = {
  val a: Nothing = null.asInstanceOf[Nothing] // No exception
}
```
but it does in 
```scala
def f: Unit = {
  val a: Nothing = null.asInstanceOf[Nothing]
  println(a) // throws here
}
```

```scala
def f: Unit = {
  val a: Int = null.asInstanceOf[Nothing] // throws here
}
```

```scala
def f: Unit = {
  null.asInstanceOf[Nothing] // throws here
}
```
This issue was metioned prevoiusly in #1002.

nicolasstucki:
```scala
def f: Unit = {
  val a: Nothing = null.asInstanceOf[Nothing] // No exception
}
```
with `-optimise` throws, that was how I discovered the issue.

PR #3342, opened 2017-10-17
Implement unused

nicolasstucki:
Decide or propose keyword that should be used for `unused`: https://goo.gl/forms/UD44AfLUD3dkfnxs2

TODO list:
- [x] Add `unused` keyword (it can be placed where `implicit` can be placed)
- [x] Check that `unused` terms are not used where they should not
- [x] Remove references to `unused` parameters, `def`s and `val`s
- [x] Remove `unused` parameters and arguments
- [x] Add `UnusedFunctionN` and `ImplcitUnusedFunctionN` for `N>0`
- [x] Fix `tests/run/unused-2.scala` with `-optimise`
- [x] Fix erasure of path dependent types on unused terms
- [x] Put `UnusedParams` in a miniphase group
- [x] User documentation
- [x] Update Scala doc
- [x] Complete frameless example
- [x] Remove phantoms and migrate tests (#3410)
- [x] Update GenericSignatures (needs rebase)
- [ ] Make sure we decide a good keyword


liufengyun:
test performance please

nicolasstucki:
test performance please

liufengyun:
I just checked the logs, this PR failed `scalapb`.

nicolasstucki:
Rebased and changed all `MiniPhaseTransforms` with `MegaPhase.MiniPhase`.

odersky:
I would add to the TODO list: 

 - remove phantom types, and migrate tests over where they make sense

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
```diff
@@ -0,0 +1,102 @@
...
+class UnusedChecks extends MiniPhase {
+  import tpd._
+
+  override def phaseName: String = "unusedChecks"
+
+
+  /* Tree transform */
+
+  override def transformTypeDef(tree: TypeDef)(implicit ctx: Context): tree.type = {
+    if (tree.symbol.is(UnusedCommon))
```
odersky: `is(Unused)`. `UnusedCommon` should never be used in an `is` check.

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
```diff
@@ -0,0 +1,102 @@
...
+      ctx.error(tree.symbol.showKind + " cannot be unused", tree.pos)
```
odersky: I think this belongs not in this phase, but where we check the other flags whether they make sense.

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
```diff
@@ -0,0 +1,102 @@
...
+        ctx.error(msg, t.pos)
+      case _ =>
+    }
+    if (!isUnusedContext)
+      rec(tree1)
+    tree0
+  }
+
+  private def isUnusedContext(implicit ctx: Context): Boolean =
+    ctx.owner.ownersIterator.exists(_.is(Unused)) // TODO make context mode?
```
odersky: Or put it in the store, yes.

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
```diff
@@ -0,0 +1,102 @@
+package dotty.tools.dotc.transform
+
```
odersky: I think it's better to implement these checks as a top-down traversal in PostTyper. It's inherently a top-down algorithm not a bottom-up one.

compiler/src/dotty/tools/dotc/transform/UnusedArgLift.scala
```diff
@@ -2,17 +2,17 @@ package dotty.tools.dotc.transform
 
 import dotty.tools.dotc.ast.tpd
 import dotty.tools.dotc.core.Contexts._
-import dotty.tools.dotc.core.NameKinds._
+import dotty.tools.dotc.core.Symbols._
 import dotty.tools.dotc.core.Types._
 import dotty.tools.dotc.transform.MegaPhase.MiniPhase
 import dotty.tools.dotc.typer.EtaExpansion
 
 import scala.collection.mutable.ListBuffer
```
odersky: We need to have a discussion whether we want to do this. The whole point of `unused` is to avoid generating code. So if we lift most arguments we miss that point. If we assume a precise effect analysis we could lift arguments if the argument is pure only. But we do not have such an effect analysis.

An alternative is to always drop unused arguments. In a sense unused arguments are then treated similarly to pattern guards - the compiler is allowed to assume they are pure. In the future, if we have a good effect analysis we could also _require_ that unused arguments are pure (and total?).

compiler/src/dotty/tools/dotc/transform/UnusedRefs.scala
```diff
@@ -0,0 +1,58 @@
...
+import dotty.tools.dotc.core.Flags._
+import dotty.tools.dotc.core.Phases
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Types._
+import dotty.tools.dotc.transform.MegaPhase.MiniPhase
+
+/** This phase removes all references and calls to unused methods or vals
+ *
+ *   if `unused def f(x1,...,xn): T = ...`
+ *   then `f(y1,...,yn)` --> `y1; ...; yn; (default value for T)`
```
odersky: Why not use `???` instead of default value? If something goes wrong later you will fail loudly instead of silently, which is better.

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
nicolasstucki: Done

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
nicolasstucki: Done

compiler/src/dotty/tools/dotc/transform/UnusedChecks.scala
nicolasstucki: I will try this in PostTyper. My first approach was in typer and I quickly realized it was a mistake because these failures influenced typing desitions. Hence decided to do it in an isolated miniphase.

compiler/src/dotty/tools/dotc/transform/UnusedRefs.scala
nicolasstucki: These default values are intended to stay there, it is the simplest tree the can be both used when `f(...)` is a statement or an expression. These are trivially eliminated by the optimizer.
This way `UnusedRefs` does not need to know and special case unused function applications in different positions. The same applies to member selection and identifiers.

compiler/src/dotty/tools/dotc/transform/UnusedArgLift.scala
nicolasstucki: Now I'm convinced that we should not evaluate any of the arguments in an unused method. This implies that unused implicits only check for the existence of such implicit and then disregard its computation. For unused arguments that are explicitly stated we should drop them as well but if they are not pure we will emit a warning.
We should not require them to be pure or total, but rather use those as hints in the warnings to point that proof might be unsafe.

```scala
def f(unused x: A): B = ???
def g()(unused implicit x: A): B = ???

f(pureA) // pureA is not evaluated
g() // the implicit A for x is not evaluated
g()(pureA) // pureA is not evaluated

f(impureA) // warn that impureA will not be evaluated
g()(impureA) // the implicit A for x is not evaluated
```

nicolasstucki:
@odersky all changes have been made. The PR is ready for review.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2031,7 +2033,17 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
                 arg :: implicitArgs(formals1)
             }
         }
-        val args = implicitArgs(wtp.paramInfos)
+        def eraseUnusedArgs(args: List[Tree]): List[Tree] = {
+          if (!wtp.isUnusedMethod) args
+          else args.map { arg =>
+            arg.tpe match {
+              case _: AmbiguousImplicits | _: SearchFailureType => arg
```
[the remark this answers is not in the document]
odersky: I think if the argument is "stable" you want to keep it anyway. If you don't do this, dependent function types might stop working. I.e. need to add clause

    case tpe if tpe.isStable => arg

docs/docs/reference/unused-terms.md
```diff
@@ -0,0 +1,146 @@
...
+}
+
+val m = new Machine[Off]
+m.turnedOn
+m.turnedOn.turnedOn // ERROR
+//                 ^
+//                  State is must be Off
+```
+
+These constraint that only depend on the types at the call site are completly resolved at compile time and never used at runtime.
```
odersky: This phrase could be replaced with:

Note that in the code above the actual implicit arguments for `IsOff` are never used at runtime; they serve only to establish the right constraints at compile time.

docs/docs/reference/unused-terms.md
```diff
@@ -0,0 +1,146 @@
...
+
+Why unused parameters?
+----------------------
+The following examples shows an implementation of a simple state machine which can be in a state `On` or `Off`.
+The machine can change state from `Off` to `On` with `turnedOn` only if it is currently `Off`. This last constraint is
+captured with the `IsOff[S]` implicit evidence which only exists for `IsOff[Off]`.
+For example, not allowing calling `turnedOn` on in an `On` state as we would require an evidence of type `IsOff[On]` that will not be found.
+
+```scala
+sealed trait State
```
odersky: Let's try to formulate this more directly. The fist example in a section should always highlight the feature we are introducing. Then, explain what the effect of the feature is. Writing code without it to motivate is too long-winded for the format we follow in these pages.

docs/docs/reference/unused-terms.md
```diff
@@ -0,0 +1,146 @@
...
+    m.turnedOn.turnedOff
+
+    // m.turnedOff
+    //            ^
+    //            State is must be On
+
+    // m.turnedOn.turnedOn
+    //                    ^
+    //                    State is must be Off
+  }
```
odersky: Need to also give a summary of the rules that govern unused parameters at the end of the section. If something is too complicated (I hope not!) need to refer to a separate issue that explains the rules in detail.

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -114,18 +114,12 @@ class Definitions {
...
-          }
-          else (MethodType, Nil)
-        val applyMeth =
-          decls.enter(
-            newMethod(cls, nme.apply,
-              methodType(argParamRefs, resParam.typeRef), Deferred))
+        val resParamRef = enterTypeParam(cls, paramNamePrefix ++ "R", Covariant, decls).typeRef
+        val methodType = MethodType.maker(isJava = false, name.isImplicitFunction, name.isUnusedFunction)
+        val parentTraits =
+          if (!name.isImplicitFunction) Nil
```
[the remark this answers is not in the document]
nicolasstucki: I will add documentation.

compiler/src/dotty/tools/dotc/core/NameOps.scala
```diff
@@ -173,31 +173,51 @@ object NameOps {
...
     /** Is a synthetic function name
      *    - FunctionN for N > 22
      *    - ImplicitFunctionN for N >= 1
+     *    - UnusedFunctionN for N >= 1
+     *    - UnusedImplicitFunctionN for N >= 1
      *    - false otherwise
      */
     def isSyntheticFunction: Boolean = {
       functionArityFor(str.Function) > MaxImplementedFunctionArity ||
-        functionArityFor(str.ImplicitFunction) >= 1
```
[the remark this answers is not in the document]
nicolasstucki: Better 
```scala
functionArityFor(str.Function) > MaxImplementedFunctionArity ||
        functionArityFor(str.ImplicitFunction) >= 1 ||
        isUnusedFunction
```
to avoid checking `functionArityFor(str.UnusedImplicitFunction) >= 1` twice

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -486,15 +487,20 @@ object Erasure {
...
               if (args0.length > MaxImplementedFunctionArity && mt.paramInfos.length == 1) {
                 val bunchedArgs = untpd.JavaSeqLiteral(args0, TypeTree(defn.ObjectType))
                   .withType(defn.ArrayOf(defn.ObjectType))
                 args0 = bunchedArgs :: Nil
               }
               // Arguments are phantom if an only if the parameters are phantom, guaranteed by the separation of type lattices
-              val args1 = args0.filterConserve(arg => !wasPhantom(arg.typeOpt)).zipWithConserve(mt.paramInfos)(typedExpr)
-              untpd.cpy.Apply(tree)(fun1, args1) withType mt.resultType
+              val args1 = args0.filterConserve(arg => !wasPhantom(arg.typeOpt))
+              assert(args1 hasSameLengthAs mt.paramInfos)
```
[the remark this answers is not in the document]
nicolasstucki: `zipWithConserve` starts with `if (xs.isEmpty || ys.isEmpty) Nil` and therefore it will trim the longest list. Maybe we should change it to `xs.isEmpty && ys.isEmpty` and require the two list to be of the same size.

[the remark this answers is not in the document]
allanrenucci: Note that both `x.zip(y)` and `(x, y).zipped` trim the longest list

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -897,7 +915,10 @@ class Definitions {
    *  trait gets screwed up. Therefore, it is mandatory that FunctionXXL
    *  is treated as a NoInit trait.
    */
-  lazy val NoInitClasses = NotRuntimeClasses + FunctionXXLClass
+  private lazy val NoInitClasses = NotRuntimeClasses + FunctionXXLClass
+
+  def isNoInitClass(cls: Symbol): Boolean =
```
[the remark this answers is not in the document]
nicolasstucki: The only case that needs it is when we create a function class that extends a function with unused parameters, such as:
```scala
class Foo extends UnusedFunction1[Int, Int] {
  def apply(unused x: Int): Int = ???
}
```

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -730,20 +730,24 @@ object Parsers {
       }
     }
 
-    /** Type        ::=  [`implicit'] FunArgTypes `=>' Type
+    /** Type        ::=  [FunArgMods] FunArgTypes `=>' Type
      *                |  HkTypeParamClause `->' Type
      *                |  InfixType
+     *  FunArgMods  ::=  `implicit' FunArgMods
```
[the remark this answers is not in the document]
nicolasstucki: ```FunArgMods  ::=  {`implicit' | `unused'}``` does not seem to capture the case where both `implicit` and `unused` are present.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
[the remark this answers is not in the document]
nicolasstucki: We could change the semantics of ours to ensure no unexpected trimming is done.

compiler/src/dotty/tools/dotc/transform/FullParameterization.scala
```diff
@@ -229,14 +229,17 @@ trait FullParameterization {
         .appliedToTypes(allInstanceTypeParams(originalDef, abstractOverClass).map(_.typeRef))
         .appliedTo(This(originalDef.symbol.enclosingClass.asClass))
 
+    def refOrDefault(tree: Tree): Tree = // use deafult values for
```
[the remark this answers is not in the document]
nicolasstucki: It can be left out. It was an early optimization. It will be removed.

compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
```diff
@@ -239,8 +239,13 @@ object GenericSignatures {
           methodResultSig(restpe)
 
         case mtpe: MethodType =>
-          // phantom method parameters do not make it to the bytecode.
-          val params = mtpe.paramInfoss.flatten.filterNot(_.isPhantom)
+          // unused method parameters do not make it to the bytecode.
+          def effectiveParamInfoss(t: Type)(implicit ctx: Context): List[List[Type]] = t match {
```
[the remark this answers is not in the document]
nicolasstucki: Unfortunately it is not possible to create a utility method that returns the type of the method without the unused parameters in presence of path-dependent types.

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -243,6 +245,14 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
               ctx.compilationUnit.source.exists &&
               sym != defn.SourceFileAnnot)
               sym.addAnnotation(Annotation.makeSourceFile(ctx.compilationUnit.source.file.path))
+            if (sym.is(Case)) {
+              tree.rhs match {
+                case rhs: Template =>
+                  for (param <- rhs.constr.vparamss.head if param.symbol.is(Unused))
+                    ctx.error("First parameter list of case classes may not contain `unused` parameters", param.pos)
```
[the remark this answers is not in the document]
nicolasstucki: Yes, it can be placed inside `checkNotUnused`

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -297,5 +307,10 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
           println(i"error while transforming $tree")
           throw ex
       }
+
+    private def checkNotUnused(tree: RefTree)(implicit ctx: Context): Unit = {
+      if (tree.symbol.is(Unused) && !ctx.mode.is(Mode.Type))
+        ctx.error(i"`unused` value $tree can only be used as unused arguments", tree.pos)
```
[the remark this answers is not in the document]
nicolasstucki: Much better

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -58,10 +58,10 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
   /** The synthetic methods of the case or value class `clazz`. */
   def syntheticMethods(clazz: ClassSymbol)(implicit ctx: Context): List[Tree] = {
     val clazzType = clazz.appliedRef
-    lazy val accessors =
-      if (isDerivedValueClass(clazz)) clazz.paramAccessors
-      else clazz.caseAccessors
-
+    lazy val accessors = {
+      val allAccessors = if (isDerivedValueClass(clazz)) clazz.paramAccessors else clazz.caseAccessors
+      allAccessors.filterConserve(!_.is(Unused))
```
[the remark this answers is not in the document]
nicolasstucki: Yes and yes, that code is redundant.

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -377,6 +377,12 @@ object Checking {
         fail(CannotHaveSameNameAs(sym, cls, CannotHaveSameNameAs.CannotBeOverridden))
         sym.setFlag(Private) // break the overriding relationship by making sym Private
       }
+    checkApplicable(UnusedType, !sym.is(UnusedType))
+    if (sym.is(Unused)) {
+      checkApplicable(Unused, !sym.is(MutableOrLazy))
+      if (sym.info.widen.finalResultType.isBottomType)
+        fail("unused " + sym.showKind + " cannot have type Nothing")
```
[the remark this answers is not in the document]
nicolasstucki: It was because of #3340 when I still was lifting the args. The program behaved differently depending on which arguments were lifted. 

I should be able to remove this now.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -866,8 +867,9 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
                   ctx.error(ex"result type of closure is an underspecified SAM type $pt", tree.pos)
                 TypeTree(pt)
               case _ =>
-                if (!mt.isDependent) EmptyTree
-                else throw new java.lang.Error(i"internal error: cannot turn dependent method type $mt into closure, position = ${tree.pos}, raw type = ${mt.toString}") // !!! DEBUG. Eventually, convert to an error?
+                if (mt.isDependent) throw new java.lang.Error(i"internal error: cannot turn dependent method type $mt into closure, position = ${tree.pos}, raw type = ${mt.toString}") // !!! DEBUG. Eventually, convert to an error?
+                else if (mt.isUnusedMethod) TypeTree(mt.toFunctionType(env1.length))
```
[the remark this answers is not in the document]
nicolasstucki: It looks like this is not needed anymore. I will remove it.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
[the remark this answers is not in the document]
nicolasstucki: Actually this was needed, but only on the value class branch where we need `clazz.paramAccessors.filterConserve(!_.is(Unused))`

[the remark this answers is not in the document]
nicolasstucki: Or `clazz.paramAccessors.take(1)`

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -377,6 +377,12 @@ object Checking {
         fail(CannotHaveSameNameAs(sym, cls, CannotHaveSameNameAs.CannotBeOverridden))
         sym.setFlag(Private) // break the overriding relationship by making sym Private
       }
+    checkApplicable(UnusedType, !sym.is(UnusedType))
```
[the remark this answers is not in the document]
nicolasstucki: This was a result of defining the `Unused` flag as a `commonFlag`, but it can be a `termFlag` which makes this test redundant. I copied that from the flag for `implicit`, I wonder if that one can ever be a type flag. I will check that.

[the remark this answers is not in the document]
nicolasstucki: Oh, for implicits it is needed for `implicit class ...`

nicolasstucki:
@odersky Ready for review again. I rebased on master and squashed commits older than your last review (history was impossible to rebase cleanly).

compiler/src/dotty/tools/dotc/transform/Erasure.scala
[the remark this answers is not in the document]
odersky: No, it's fine. I forgot about the trimming behavior. So the current implementation is OK.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -59,7 +59,7 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
   def syntheticMethods(clazz: ClassSymbol)(implicit ctx: Context): List[Tree] = {
     val clazzType = clazz.appliedRef
     lazy val accessors =
-      if (isDerivedValueClass(clazz)) clazz.paramAccessors
+      if (isDerivedValueClass(clazz)) clazz.paramAccessors.take(1) // Tail parameters can only be `unused`
```
odersky: ??? Did not understand. Can you give an example?

compiler/src/dotty/tools/dotc/transform/UnusedDecls.scala
```diff
@@ -0,0 +1,49 @@
+package dotty.tools.dotc.transform
```
odersky: It seems to be quote a bit of code. Would the same effect not be obtained more cheaply by rolling this into erasure?

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2117,7 +2120,18 @@ class Typer extends Namer
                 arg :: implicitArgs(formals1)
             }
         }
-        val args = implicitArgs(wtp.paramInfos)
+        def eraseUnusedArgs(args: List[Tree]): List[Tree] = {
```
odersky: Move into UnusedUtil?

compiler/src/dotty/tools/dotc/typer/UnusedUtil.scala
```diff
@@ -0,0 +1,32 @@
+package dotty.tools.dotc.typer
+
+import dotty.tools.dotc.ast.tpd._
+import dotty.tools.dotc.core.Contexts._
+import dotty.tools.dotc.core.Decorators._
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Flags._
+
+/** Util methods for transformation of `unused` expressions
+ *
```
odersky: Looks a bit small for a separate object. Can we find other methods that naturally go here? If not, maybe dissolve into another module?

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
nicolasstucki: Value classes must have exactly one parameter, but we can add additional ones if they are `unused`. We already check that the first parameter list is not `unused` (same for case classes).
```scala
class Foo(v: Int)(unused ev: Ev) extends AnyVal
```
When we have that, we need to ignore the `unused` parameters when generating synthetic methods. In the case of a value class we know that the only the first parameter is needed.

compiler/src/dotty/tools/dotc/transform/UnusedDecls.scala
nicolasstucki: The idea was to prune the tree early to reduce the cost of traversing those declarations. I would have to check if it can be placed in erasure.

compiler/src/dotty/tools/dotc/typer/UnusedUtil.scala
nicolasstucki: These two methods can be moved to `Applications`

compiler/src/dotty/tools/dotc/typer/Typer.scala
nicolasstucki: Will remove `UnusedUtil`

nicolasstucki:
Rebased. Last two commits are new.

nicolasstucki:
Last two commits should be reviewed.

nicolasstucki:
Rebase

Blaisorblade:
Just voted in the poll (is it advertised more widely?).

FWIW, to me `unused` suggests not that the parameter is erased, but that it is simply not used (but the compiler shouldn't warn me  please).

Also, we might want to erase them even though they *seem* to be used. The current semantics prevent from making `=:=` erased (as soon as you call its `apply` method); yesterday I talked with Martin about how that could be fixed by marking as `inline` methods like `apply`.

docs/docs/reference/unused-terms.md
```diff
@@ -34,9 +34,7 @@ m.turnedOn.turnedOn // ERROR
 
 Note that in the code above the actual implicit arguments for `IsOff` are never used at runtime; they serve only to establish the right constraints at compile time.
 As these parameters are never used at runtime there is not real need to have them around, but they still need to be
-present at runtime to be able to do separate compilation and retain binary compatiblity. Unused parameters are contractually
-obligated to not be used at runtime, enforcing the essence of evidences on types and allows them to always be optimized away.
-
+present at runtime to be able to do separate compilation and retain binary compatiblity.
```
odersky: present at runtime --> present in some form in the generated code

nicolasstucki:
test performance please

joan38:
Really weired name to call this `unused`, it usually tells me that I should remove it if it's not used.
I would have prefered something like `erased` or `phantom`.

nicolasstucki:
I also want to change it. `erased` is not the best name as there is the concept of erasure on types that is quite different. The best contenders are `phantom` or `ghost`.

Blaisorblade:
Any chance to pick another name, maybe even before the release, since apparently nobody likes `unused`?

nicolasstucki:
I could make the change tomorrow morning.

nicolasstucki:
Renamed to `ghost` in #4061

tests/run/unused-24.scala
```diff
@@ -0,0 +1,23 @@
+object Test {
+
+  def main(args: Array[String]): Unit = {
+    println(fun(new Bar))
+  }
+
+  def fun(unused foo: Foo): foo.X = {
```
Blaisorblade: This tests is unsound because it treats ghost values as stable (#4060) /cc @nicolasstucki

PR #3345, opened 2017-10-17
Fix #3343: Make baseType work properly with type lambdas

### LINKED ISSUE #3343: Implicit view not found involving an uninstantiated type variable
This fails to compile with Dotty
```scala
class Test {
  implicit def implicitContainer[C[_], T](
    implicit t: C[T] => Traversable[T]): C[T] = ???

  implicitly[List[Char]]
}
```
```scala
-- Error: tests/allan/Test.scala:5:24 ------------------------------------------
5 |  implicitly[List[Char]]
  |                        ^
  |no implicit argument of type scala.collection.immutable.List[Char] found for parameter e of method implicitly in object Predef
```
Note that this compiles:
```scala
class Test {
  implicitly[List[Char] => Traversable[Char]]
}
```

smarter:
Previously,
```scala
    `C[Char]`.baseTypeOf(`Traversable[Char]`)
    where C := `[X] => List[X]`
```
returned `Traversable[ParamRef(A)]` instead of `Traversable[Char]`,
because the substitution was done on the type parameters of the class
symbol instead of the type parameters of the type, these are different
when type lambdas are involved.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -1646,13 +1646,11 @@ object SymDenotations {
           case tp @ AppliedType(tycon, args) =>
             val subsym = tycon.typeSymbol
             if (subsym eq symbol) tp
-            else subsym.denot match {
-              case clsd: ClassDenotation =>
-                val tparams = clsd.typeParams
-                if (tparams.hasSameLengthAs(args)) baseTypeOf(tycon).subst(tparams, args)
-                else NoType
-              case _ =>
+            else tycon.typeParams match {
+              case LambdaParam(_, _) :: _ =>
                 baseTypeOf(tp.superType)
+              case tparams: List[Symbol @unchecked] =>
+                baseTypeOf(tycon).subst(tparams, args)
             }
           case tp: TypeProxy =>
             baseTypeOf(tp.superType)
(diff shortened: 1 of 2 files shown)


odersky:
I think the real fault is in the condition. We check that the type-symbol refers a ClassDenotation. The idea was that this should ensure that the type constructor is a class, but the example shows this is wrong. So I'd try to get the condition right instead. One candidate is what use use in `substParams`: check that `typeParams` returns a list of `Symbols`. If that's not the case, form the baseType of the superType instead.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1648,9 +1648,7 @@ object SymDenotations {
             if (subsym eq symbol) tp
             else subsym.denot match {
               case clsd: ClassDenotation =>
-                val tparams = clsd.typeParams
-                if (tparams.hasSameLengthAs(args)) baseTypeOf(tycon).subst(tparams, args)
-                else NoType
+                baseTypeOf(tycon).substTypeParams(tycon, args)
```
odersky: I think the real fault is the condition. It was meant to check that we have an application of a class instead of a type lambda but failed at that. So I'd fix that first. One candidate is the condition of `substTypeParams`, that the `typeParams` of `subsym` are symbols. If that's the case, proceed as before, if not do a `baseTypeOf(superType)`, again as before.

smarter: Makes sense. PR updated.

allanrenucci:
I can confirm it fixes the issue I reported from scalacheck

PR #3346, opened 2017-10-18
Fix #2961: Scale logo size on home page with screen size

### LINKED ISSUE #2961: The slogan can not be seen in the default zoom ratio on the home page
Default:
image

80%:
image

50:
image

DSouzaM:
Changes the logo and slogan styling to scale based on screen height. This produces more suitable results on different screen sizes.

Also adds a height threshold to show mobile UI (previously, a wide but short window would not trigger the mobile UI, so the site wouldn't look very good viewing landscape on a phone).

1920x1200px (computer, high-res)
dotty-1920x1200

1280x800px (computer, low-res)
dotty-1280x800

360x640px (phone, portrait)
dotty-360x640

640x360px (phone, landscape)
dotty-640x360

I've tested this change with many other different screen sizes within Chrome as well, with good results.

### DIFF
--- a/docs/css/default.css
+++ b/docs/css/default.css
@@ -143,11 +143,10 @@ div#header li.nav-item > a.nav-link {
 }
 
 div.scala-logo-container {
-    width: 232px;
-    height: calc(400px + 13em);
+    height: calc(100vh - 80px);
     margin: auto;
     position: absolute;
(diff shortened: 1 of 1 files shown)


felixmulder:
Awesome! LGTM :)

PR #3347, opened 2017-10-18
Fix #2732: Allow wildcards in SAM types

### LINKED ISSUE #2732: SAM types don't support type bounds
The following compiles in Scala but fails in Dotty (using Scastie)
```scala
object Foo {
  val map = new java.util.HashMap[String, String]
  map.computeIfAbsent("hello", foo => "world")
}
```
```
-- Error: /tmp/scastie5756006742801572629/src/main/scala/main.scala:5:31 -------
5 |  map.computeIfAbsent("hello", foo => "world")
  |                               ^^^
  |                    missing parameter type for parameter foo, expected = ?
-- [E007] Type Mismatch Error: /tmp/scastie5756006742801572629/src/main/scala/main.scala:5:38 
5 |  map.computeIfAbsent("hello", foo => "world")
  |                                      ^^^^^^^
  |         found:    String("world")
  |         required: java.util.function.Function[_ >: String, _ <: String]#R         
```
trying
```scala
object Foo {
  val map = new java.util.HashMap[String, String]
  map.computeIfAbsent("hello", (foo: String) => "world")
}
```
still results in the second error
```
-- [E007] Type Mismatch Error: /tmp/scastie5756006742801572629/src/main/scala/main.scala:5:48 
5 |  map.computeIfAbsent("hello", (foo: String) => "world")
  |                                                ^^^^^^^
  |         found:    String("world")
  |         required: java.util.function.Function[_ >: String, _ <: String]#R
  |         
```

smarter:
The `isInstantiable` test as written is incorrect:
Given tp = `java.util.function.Function[String, _ <: String]`, we will have:
    tref = skolem TermRef with underlying info `tp`
    selfType = `java.util.function.Function[String, tref#R]` (because of the `asSeenFrom`)
`tref <:< selfType` is false, because `tref.underlying` is
just `tp` and its return type parameter is a bounded wildcard, not a reference with
`tref` as a prefix.
In any case, I cannot think of a case where this function should return
false, and no test failed when I removed it. So I'm removing it until
someone can come up with a counter-example.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3717,16 +3717,8 @@ object Types {
       case _ =>
         NoType
     }
-    def isInstantiatable(tp: Type)(implicit ctx: Context): Boolean = zeroParamClass(tp) match {
-      case cinfo: ClassInfo =>
-        val tref = tp.narrow
-        val selfType = cinfo.selfType.asSeenFrom(tref, cinfo.cls)
-        tref <:< selfType
-      case _ =>
-        false
-    }
     def unapply(tp: Type)(implicit ctx: Context): Option[SingleDenotation] =
-      if (isInstantiatable(tp)) {
+      if (zeroParamClass(tp).exists) {
         val absMems = tp.abstractTermMembers
         // println(s"absMems: ${absMems map (_.show) mkString ", "}")
         if (absMems.size == 1)
--- a/tests/pos/i2732.scala
+++ b/tests/pos/i2732.scala
@@ -1,0 +1,8 @@
+object Foo {
+  // Example 1
+  val fun: java.util.function.Function[String, _ <: String] = x => x
+
+  // Example 2
+  val map = new java.util.HashMap[String, String]
+  map.computeIfAbsent("hello", foo => "world")
+}


odersky:
Here's a test:

    class B {
      def f = 1
    }

    trait A { self: B =>

      def g1(x: Int): Int
      def g2 = g1(f)

    }

    object Test extends App {

      val x: A = (y: Int) => y + y

      println(x.g2)

    }

With master this gives:

    -- [E007] Type Mismatch Error: ../neg/i3347.scala:14:30 ------------------------
    14 |  val x: A = (y: Int) => y + y
       |                              ^
       |                              found:    Int => Int
       |                              required: A
       |        

With this PR it compiles but we get a ClassCast exception at runtime.

~/workspace/dotty/tests/pos> java Test
Exception in thread "main" java.lang.ExceptionInInitializerError
	at Test.main(i3347.scala)
Caused by: java.lang.ClassCastException: Test$$$Lambda$10/425918570 cannot be cast to B
	at A.g2(i3347.scala:8)
	at Test$.<init>(i3347.scala:16)
	at Test$.<clinit>(i3347.scala)
	... 1 more

smarter:
Indeed. Can you think of a way to fix `isInstantiable` ? Maybe we could use a TypeMap to rewrite `Function[String, _ <: String]` to `Function[String, String]` before calling `.narrow`, but that's not very nice.

liufengyun:
test performance please

(meta test of the bench infrastructure)

odersky:
> Indeed. Can you think of a way to fix isInstantiable?

For the moment it's too much of a context switch for me to think of a fix.

smarter:
Superceded by https://github.com/lampepfl/dotty/pull/4152

PR #3350, opened 2017-10-19
Run all miniphase transforms at group end

odersky:
This avoids having to change the phase of the current context on each node visit. The performance tests indicate that this gives a performance improvement of 6-7% on dotty itself.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
+++ b/compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
@@ -19,9 +19,6 @@ class RenameLifted extends MiniPhaseTransform with SymTransformer { thisTransfor
   // Not clear why this should run after restoreScopes
   // override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[RestoreScopes])
 
-  override def treeTransformPhase(implicit ctx: Context, info: TransformerInfo) =
-    groupEndPhase
-
   def transformSym(ref: SymDenotation)(implicit ctx: Context): SymDenotation =
     if (needsRefresh(ref.symbol)) ref.copySymDenotation(name = refreshedName(ref.symbol))
     else ref
--- a/compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
+++ b/compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
@@ -16,7 +16,8 @@ class RenameLifted extends MiniPhaseTransform with SymTransformer { thisTransfor
 
   override def phaseName = "renameLifted"
 
-  override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[RestoreScopes])
+  // Not clear why this should run after restoreScopes
+  // override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[RestoreScopes])
 
   override def treeTransformPhase(implicit ctx: Context, info: TransformerInfo) =
     groupEndPhase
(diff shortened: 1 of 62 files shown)


odersky:
> Won't this prevent this group from passing Ycheck ?

Seems not. Ycheck still passes the group where restoreScopes used to be run.

DarkDimius:
The last phase block is not ychecked. This will mean that flatten and RestoreScopes are no longer checked. FTR ycheck  used to find many bugs in those phases.

odersky:
If we suspect Flatten is wrong, we could always split it into its own group and Ycheck after it. But it won't run by default, needs manual intervention.

odersky:
test performance please

smarter:
> The last phase block is not ychecked.

@DarkDimius I think it is, LabelDefs is in the last group: https://github.com/lampepfl/dotty/blob/7c8f55741f7f85c26116357705b7db3e71739dfd/compiler/test/dotty/tools/vulpix/TestConfiguration.scala#L45

smarter:
Would be interesting to get some statistics on how much this reduce the number of Context objects we create.

smarter:
test performance please

compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
```diff
@@ -9,14 +9,15 @@ import dotty.tools.dotc.core.Names._
...
+import dotty.tools.dotc.transform.TreeTransforms.{MiniPhaseTransform, TransformerInfo}
 
 /** Renames lifted classes to local numbering scheme */
 class RenameLifted extends MiniPhaseTransform with SymTransformer { thisTransformer =>
 
   override def phaseName = "renameLifted"
 
-  override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[RestoreScopes])
+  // Not clear why this should run after restoreScopes
+  // override def runsAfterGroupsOf: Set[Class[_ <: Phases.Phase]] = Set(classOf[RestoreScopes])
```
smarter: @nicolasstucki Can you comment on that ?

compiler/src/dotty/tools/dotc/transform/TreeTransform.scala
```diff
@@ -471,6 +464,15 @@ object TreeTransforms {
 
     def miniPhases: Array[MiniPhase]
 
+    private var myGroupEndId: PhaseId = NoPhaseId
```
smarter: Could be `private[this]` :)

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -404,7 +406,7 @@ trait TypeAssigner {
         }
         else {
           val argTypes = preCheckKinds(args, pt.paramInfos).tpes
-          if (sameLength(argTypes, paramNames) || ctx.phase.prev.relaxedTyping) pt.instantiate(argTypes)
```
smarter: I still think it'd be nice to keep the relaxedTyping check when we can (so, at least in the Typer).

DarkDimius:
I like the idea of stores instead of mutable state inside phases.

compiler/src/dotty/tools/dotc/transform/RenameLifted.scala
nicolasstucki: Originally it was on `Flatten` because this transformation acts on the flattened names. But @DarkDimius noted that the flattening transformation finishes in `RestoreScopes`.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
odersky: OK, I'll roll that into the next PR (#3366).

PR #3353, opened 2017-10-19
Make vulpix diff readable

nicolasstucki:
This will print the diff of the vulpix outputs in two colored columns starting at the beginning and ending at the first `EOF` found.


nicolasstucki:
The new output looks like this

screen shot 2017-10-20 at 11 34 43

nicolasstucki:
Now it prints
 
screen shot 2017-10-20 at 11 43 36

nicolasstucki:
To compare, in the current master it prints

screen shot 2017-10-20 at 12 47 58

PR #3356, opened 2017-10-20
Make MissingCoreLibTests a simple unit test

allanrenucci:
I don't see pros for it being a vulpix test

### DIFF
--- a/compiler/test/dotty/tools/dotc/MissingCoreLibTests.scala
+++ b/compiler/test/dotty/tools/dotc/MissingCoreLibTests.scala
@@ -2,32 +2,21 @@ package dotty
 package tools
 package dotc
 
-import org.junit.{AfterClass, Test}
-import vulpix._
+import org.junit.Test
+import org.junit.Assert._
 
-import scala.concurrent.duration._
+import vulpix.TestConfiguration.mkClassPath
 
-class MissingCoreLibTests extends ParallelTesting {
-  import MissingCoreLibTests._
-  import TestConfiguration._
-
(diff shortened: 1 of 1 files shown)


compiler/test/dotty/tools/dotc/MissingCoreLibTests.scala
```diff
@@ -2,32 +2,21 @@ package dotty
...
-  def testFilter = Properties.testsFilter
-
-  @Test def missingDottyLib: Unit = {
+  @Test def missingDottyLib2: Unit = {
     val classPath = mkClassPath(Jars.dottyCompiler :: Jars.dottyInterfaces :: Jars.dottyExtras) // missing Jars.dottyLib
-    val options = noCheckOptions ++ checkOptions ++ yCheckOptions
-    compileFile("../tests/neg/nolib/Foo.scala", TestFlags(classPath, options)).checkExpectedErrors()
+    val source = "../tests/neg/nolib/Foo.scala"
+    val options = Array("-classpath", classPath, source)
+    val reporter = Main.process(options)
```
nicolasstucki: You are not handling the case where the compiler crashes, vulpix did that. A `try/catch` and `assertTrue(false, ex.stackTrace)` a good enough replacement.

allanrenucci: My understanding is that if something in the test crashes, the tes fail

nicolasstucki: Right

PR #3357, opened 2017-10-20
Fix #3352: defaults parameters require named type in compatibility check

### LINKED ISSUE #3352: Bad interaction between overloading, default arguments and implicit search
This code snippet does not compile with Dotty:
```scala
class Test {
  class Foo {
    def bar(x: String): Int = 1
  }

  implicit class FooOps(foo: Foo) {
    def bar(x: Int, y: Int = 2): Int = 2 // compiles with no default argument
  }

  def test(foo: Foo): Unit = {
    foo.bar(1)
  }
}
```
```scala
[info] Running dotty.tools.dotc.Main -d out Test.scala -classpath dotty-library_2.12-0.5.0-bin-SNAPSHOT-nonbootstrapped.jar
-- [E007] Type Mismatch Error: tests/allan/Test.scala:23:12 --------------------
23 |    foo.bar(1)
   |            ^
   |            found:    Int(1)
   |            required: String
   |
```
Maybe related to #3168

liufengyun:
Fix #3352: defaults parameters require named type in compatibility check

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -41,7 +41,15 @@ object ProtoTypes {
     /** Test compatibility after normalization in a fresh typerstate. */
     def normalizedCompatible(tp: Type, pt: Type)(implicit ctx: Context) = ctx.typerState.test {
       val normTp = normalize(tp, pt)
-      isCompatible(normTp, pt) || pt.isRef(defn.UnitClass) && normTp.isParameterless
+      isCompatible(normTp, pt) ||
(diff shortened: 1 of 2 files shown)


liufengyun:
test performance please

liufengyun:
@odersky : I just updated the PR with more notes. Could you check if there's a better solution to the problem? Thanks.

odersky:
Thanks for the notes. Now I understand what goes on. I am still not convinced because of the duplication in compatibility checks. This could cost us in complexity and performance. I think it might be better to make normalize not dereference a term type. Let me try this out.

liufengyun:
Superseded by #3386

PR #3358, opened 2017-10-20
Make tests for locality of unique indices not depend on order

nicolasstucki:


nicolasstucki:
The check files of these tests have changed a few times due to changes in the miniphases order and tree structure. This PR makes these test independent of the order, which was never intended to be tested in the first place.

Currently #3350 and one of my local branches are affected by this.

PR #3362, opened 2017-10-20
Reimplement LabelDefs using single transform

nicolasstucki:
Previously we had 3 transformations steps;
the MiniPhaseTransformation and 2 TreeTransforms which retransformed
the same label defs several times.

With this change, all label defs transformations are performed the
MiniPhaseTransformation. This is achieved by delaying the transformation
of label defs until their first application is found.


nicolasstucki:
Test performance please

nicolasstucki:
test performance please

PR #3365, opened 2017-10-22
Cleanup LabelDefs

nicolasstucki:


compiler/src/dotty/tools/backend/jvm/LabelDefs.scala
```diff
@@ -117,12 +114,12 @@ class LabelDefs extends MiniPhaseTransform {
     }
   }
 
-  object collectLabelDefs extends TreeMap() {
+  private object collectLabelDefs extends TreeMap() {
```
allanrenucci: Is it for performance reason that this is an object and we call `clear()` before using it, instead of making it a class, and creating a new instance when we use it?

nicolasstucki: Not yet, I am exploring what can be simplified step by step. My intention is to fix a stack overflow. My previous attempt (#3362) modified the whole code in one go an something made the bootstrapped compiler buggy. I this PR I intend to change small details at a time and run the full test suite on each change. This is just step 0 of my changes.

compiler/src/dotty/tools/backend/jvm/LabelDefs.scala
```diff
@@ -82,68 +82,48 @@ import StdNames.nme
...
-      collectLabelDefs.clear
-      val newRhs = collectLabelDefs.transform(tree.rhs)
-      var labelDefs = collectLabelDefs.labelDefs
+      val labelDefs = collectLabelDefs(tree.rhs)
 
       def putLabelDefsNearCallees = new TreeMap() {
-
         override def transform(tree: tpd.Tree)(implicit ctx: Context): tpd.Tree = {
           tree match {
+            case t: Template => t
```
smarter: I don't think this case will ever be hit since we're after LambdaLift, same for the `case _: Template` below.

smarter:
test performance please

allanrenucci:
I think you can also remove all the `tpd` prefixes

nicolasstucki:
Rebased

compiler/src/dotty/tools/backend/jvm/LabelDefs.scala
```diff
@@ -82,68 +56,47 @@ import StdNames.nme
...
-    if (tree.symbol is Flags.Label) tree
+  override def transformDefDef(tree: DefDef)(implicit ctx: Context): Tree = {
+    if (tree.symbol is Label) tree
     else {
-      collectLabelDefs.clear
-      val newRhs = collectLabelDefs.transform(tree.rhs)
-      var labelDefs = collectLabelDefs.labelDefs
+      val labelDefs = collectLabelDefs(tree.rhs)
 
       def putLabelDefsNearCallees = new TreeMap() {
```
allanrenucci: I guess this can be a `val`

nicolasstucki: Yes, but I added a small optimization to not create the transformer if `labelDefs` is empty. Therefore it will be kept as a `def`.

PR #3366, opened 2017-10-22
Rework tree transforms

odersky:
Based on #3350.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -54,8 +54,9 @@ object Trees {
     final val PackageDef = 29
     final val Thicket = 30
 
+    /** The highest tree tag that can appear in a picked tree + 1 */
     final val NumPickledTreeTags = 31
-    
+
     final val Annotated = 31
     final val SingletonTypeTree = 32
     final val AndTypeTree = 33
@@ -66,6 +67,7 @@ object Trees {
     final val ByNameTypeTree = 38
     final val TypeBoundsTree = 39
 
+    /** The highest tree tag that can appear in a typed tree + 1 */
     final val NumTypedTreeTags = 40
 
     final val TypedSplice = 40
@@ -89,6 +91,7 @@ object Trees {
     final val ContextBounds = 48
     final val PatDef = 49
 
+    /** The highest tree tag + 1 */
     final val NumTags = 50
   }
 
@@ -432,7 +435,7 @@ object Trees {
   case class Ident[-T >: Untyped] private[ast] (name: Name)
     extends RefTree[T] {
     type ThisTree[-T >: Untyped] = Ident[T]
-    def tag = Tag.Ident
+    final def tag = Tag.Ident
     def qualifier: Tree[T] = genericEmptyTree
 
     /** Is this a `BackquotedIdent` ? */
@@ -450,7 +453,7 @@ object Trees {
   case class Select[-T >: Untyped] private[ast] (qualifier: Tree[T], name: Name)
     extends RefTree[T] {
     type ThisTree[-T >: Untyped] = Select[T]
-    def tag = Tag.Select
+    final def tag = Tag.Select
   }
 
   class SelectWithSig[-T >: Untyped] private[ast] (qualifier: Tree[T], name: Name, val sig: Signature)
@@ -462,7 +465,7 @@ object Trees {
   case class This[-T >: Untyped] private[ast] (qual: untpd.Ident)
     extends DenotingTree[T] with TermTree[T] {
     type ThisTree[-T >: Untyped] = This[T]
-    def tag = Tag.This
+    final def tag = Tag.This
     // Denotation of a This tree is always the underlying class; needs correction for modules.
     override def denot(implicit ctx: Context): Denotation = {
       typeOpt match {
@@ -478,7 +481,7 @@ object Trees {
   case class Super[-T >: Untyped] private[ast] (qual: Tree[T], mix: untpd.Ident)
     extends ProxyTree[T] with TermTree[T] {
     type ThisTree[-T >: Untyped] = Super[T]
-    def tag = Tag.Super
+    final def tag = Tag.Super
     def forwardTo = qual
   }
 
@@ -493,35 +496,35 @@ object Trees {
   case class Apply[-T >: Untyped] private[ast] (fun: Tree[T], args: List[Tree[T]])
     extends GenericApply[T] {
     type ThisTree[-T >: Untyped] = Apply[T]
-    def tag = Tag.Apply
+    final def tag = Tag.Apply
   }
 
   /** fun[args] */
   case class TypeApply[-T >: Untyped] private[ast] (fun: Tree[T], args: List[Tree[T]])
     extends GenericApply[T] {
     type ThisTree[-T >: Untyped] = TypeApply[T]
-    def tag = Tag.TypeApply
+    final def tag = Tag.TypeApply
   }
 
   /** const */
   case class Literal[-T >: Untyped] private[ast] (const: Constant)
     extends TermTree[T] {
     type ThisTree[-T >: Untyped] = Literal[T]
-    def tag = Tag.Literal
+    final def tag = Tag.Literal
   }
 
   /** new tpt, but no constructor call */
   case class New[-T >: Untyped] private[ast] (tpt: Tree[T])
     extends TermTree[T] {
     type ThisTree[-T >: Untyped] = New[T]
-    def tag = Tag.New
+    final def tag = Tag.New
   }
 
   /** expr : tpt */
   case class Typed[-T >: Untyped] private[ast] (expr: Tree[T], tpt: Tree[T])
     extends ProxyTree[T] with TermTree[T] {
     type ThisTree[-T >: Untyped] = Typed[T]
-    def tag = Tag.Typed
+    final def tag = Tag.Typed
     def forwardTo = expr
   }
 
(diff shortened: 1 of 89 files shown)


odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

smarter:
Could you give a list of differences between TreeTransformer and SuperPhase?

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,628 @@
...
+import scala.util.control.NonFatal
+import reporting.trace
+
+/** A SuperPhase combines a number of mini-phases which are all executed in
+ *  a single tree traversal.
+ *
+ *  This is an evolution of the previous "TreeTransformers.scala", which was written by @DarkDimius and
+ *  is described in his thesis.
+ */
+object SuperPhase {
```
smarter: Bikeshed: I'm not a big fan of the name `SuperPhase` since `super` already has a very specific meaning in Scala. Some possible alternatives: `MegaPhase`,`FusedPhase`, `JoinedPhase`.

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,628 @@
...
+    def trans(tree: Tree)(implicit ctx: Context) = transformer(tree, ctx)
+
+    def localContext(implicit ctx: Context) = {
+      val sym = tree.symbol
+      val owner = if (sym is PackageVal) sym.moduleClass else sym
+      ctx.fresh.setOwner(owner)
+    }
+
+    { implicit val ctx = nestedCtx
+      tag match {
```
smarter: This could be `tag: @switch`

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,628 @@
...
+
+  private val cpy: TypedTreeCopier = cpyBetweenPhases
+
+  /** Transform node using all phases in this group that have idxInGroup >= start */
+  def transformNode(tree: Tree, start: Int)(implicit ctx: Context) =
+    nodeTransformer(tree.tag)(start)(tree, ctx)
+
+  /** Transform full tree using all phases in this group that have idxInGroup >= start */
+  def transformTree(tree: Tree, start: Int)(implicit ctx: Context): Tree = {
+    val tag = tree.tag
```
smarter: We could avoid the megamorphic dispatch here if we had something like:
```scala
abstract class Tree[-T >: Untyped](tag: TreeTag) ...
// ...
class Ident(...) extends Tree(Tag.Ident)
```

DarkDimius:
I should be able to review this on Wednesday.
Would you kindly give me a short summary of what motivated the changes in last 3 commits? It will give a perspective for review.

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
odersky: OK, let's go with MegaPhase. I like the sound of that :-)

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
odersky: Yes, but at the cost of one field per tree. It might still be worth it, though. One needs to experiment.

odersky:
The motivation from the change came from me trying to understand (again) TreeTransformers.scala when I did the work on running all transforms at group end. I noted that things were very convoluted, and repetitive and that, as a consequence it was hard to find the right way to do changes. For instance there was the issue that TreeTransforms, but not TeeTransformers used cpyBetweenPhases. I believe that was not intentional (or was it?), but rather caused by the fact that due to its length, TreeTransformers is hard to maintain. The new set of changes cuts the size of TreeTransformers to roughly half of what it was before and clearly separates actual implementation logic (~200loc) from initialization code.

There was also the motivation to play with the implementation a bit to see whether we could wring more performance out of it. To be able to judge performance, I believe we have to get #3350 in first and then compare. I am not yet sure performance is better or worse. But to play with it I felt it was necessary to simplify it first.

In detail, the changes can be summarized as follows:

MegaPhase is a simplified version of TreeTransformer

 - about half of the size
 - no need for separate TransformerInfo arguments
 - simpler wiring logic

The simplifications are largely due to three changes

 - use of function values, where wiring is done by a sort of staging
 - use of tree tags, which give a common index into tables of transformer lambdas
 - prepare... operations return a Context instead of a TreeTransform, so there are
    fewer "moving parts".

odersky:
For both review and performance comparisons it would be best got #3350 in first.

compiler/src/dotty/tools/dotc/util/Store.scala
```diff
@@ -0,0 +1,33 @@
...
+  val empty: Store = new Store(Array())
+}
+
+class Store(private val elems: Array[AnyRef]) extends AnyVal {
+  import Store._
+
+  def newLocation[T](): (Location[T], Store) = {
+    val elems1 = new Array[AnyRef](elems.length + 1)
+    System.arraycopy(elems, 0, elems1, 0, elems.length)
+    (new Location(elems.length), new Store(elems1))
```
liufengyun: For another PR: minor optimisations here to double the size of array and keep track of size?

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -19,6 +19,76 @@ import language.implicitConversions
...
+    final val Import = 36
+    final val PackageDef = 37
+    final val Annotated = 38
+    final val Thicket = 39
+
+    final val NumTypedTreeTags = 40
+
+    final val TypedSplice = 40
+    final val ModuleDef = 41
+    final val ParsedTry = 42
```
nicolasstucki: I would add a comment describing which tag numbers can be reused and why.

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,625 @@
...
+  /** Does `phase` contain a redefinition of method `name`?
+   *  (which is a method of MiniPhase)
+   */
+  private def defines(phase: MiniPhase, name: String) = {
+    def hasRedefinedMethod(cls: Class[_]): Boolean =
+      if (cls.eq(classOf[MiniPhase])) false
+      else {
+        var clsMethods = clsMethodsCache.get(cls)
+        if (clsMethods eq null) {
+          clsMethods = cls.getDeclaredMethods
```
DarkDimius: proposal: add `map(_.getName)` here and use `contains` instead of `exists` below.

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,625 @@
...
+   *  - the last element is always `last`, which is the identity transformer
+   *    that returns the context or tree unchanged.
+   *  - The element corresponding to phase P is `elemFn(P)` if `P` defines a
+   *    method with name `methName`, or it the same as the following element
+   *    if `P` does not define `methName`.
+   */
+  private def newTransformerArray[T, R](
+      methName: String, elemFn: IndexedTransformer[T, R], last: Transformer[T, R]) = {
+    val trans = new Array[Transformer[T, R]](miniPhases.length + 1)
+    trans(miniPhases.length) = last
```
DarkDimius: :-) Now it's actually closer to version described in the thesis.

compiler/src/dotty/tools/dotc/transform/SuperPhase.scala
```diff
@@ -0,0 +1,625 @@
...
+  private val unitNodeTransformer     = newNodeTransformerArray("Unit", transUnit)
+  private val otherContextTransformer = newContextTransformerArray("Other", prepForOther)
+  private val otherNodeTransformer    = newNodeTransformerArray("Other", transOther)
+
+  private val contextTransformer: Array[Array[Transformer[Tree, Context]]] =
+    Array.fill(Tag.NumTags)(otherContextTransformer)
+  private val nodeTransformer: Array[Array[Transformer[Tree, Tree]]] =
+    Array.fill(Tag.NumTags)(otherNodeTransformer)
+
+  private def init(name: String, tag: TreeTag, ctf: IndexedTransformer[_, Context], ntf: IndexedTransformer[_, Tree]): Unit = {
```
DarkDimius: I think `init` is too generic name and also counterintuitive given that it is called more than once.

compiler/src/dotty/tools/dotc/ast/Trees.scala
DarkDimius: 👍 It's not clear for me why are they reused.

DarkDimius:
I like the idea of stores inside Context instead of mutable state inside phases.

DarkDimius:
I've left comments on individual commits. 

I'm very curious to see what impact on performance will manual handling of tags have compared to `instanceOf` tests.

odersky:
test performance please

compiler/src/dotty/tools/dotc/ast/Trees.scala
odersky: There's no re-use: NumTypeTreeTags is a counter, not a tag. I'll add a comment to make that clearer.

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -357,6 +425,7 @@ object Trees {
   case class Ident[-T >: Untyped] private[ast] (name: Name)
     extends RefTree[T] {
     type ThisTree[-T >: Untyped] = Ident[T]
+    def tag = Tag.Ident
```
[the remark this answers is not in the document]
odersky: Final.

compiler/src/dotty/tools/dotc/util/Store.scala
odersky: That would be complicated because arrays are purely functional. I.e you have to copy anyway on an update.

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -19,6 +19,72 @@ import language.implicitConversions
 
 object Trees {
 
+  type TreeTag = Int
+
+  object Tag {
+    final val Ident = 0
```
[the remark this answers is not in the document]
odersky: We could try, but then the switches will not be dense anymore, so will most likely not be implemented by a table switch.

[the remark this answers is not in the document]
odersky: Also, in that case unpickled trees would not have tags, so we could not use tags for (e.g. typer).

PR #3368, opened 2017-10-23
Fix #3364: Name clash in class and its companion object

### LINKED ISSUE #3364: Name clash in class and its companion object
The following code snippet compiles with Dotty but not with Scalac
```scala
object Test {
  object Foo
}

class Test {
  class Foo
}
```
Same for:
```scala
object Test {
  class Foo
}

class Test {
  object Foo
}
```

allanrenucci:


compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -112,10 +112,15 @@ object RefChecks {
    *  a class or module with same name
    */
   private def checkCompanionNameClashes(cls: Symbol)(implicit ctx: Context): Unit =
-    if (!(cls.owner is ModuleClass)) {
-      val other = cls.owner.linkedClass.info.decl(cls.name).symbol
-      if (other.isClass)
+    if (!cls.owner.is(ModuleClass)) {
+      def clashes(sym: Symbol) =
+        sym.name.stripModuleClassSuffix == cls.name.stripModuleClassSuffix &&
+        (sym.isClass || sym.is(ModuleClass))
```
smarter: A ModuleClass is a class

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -112,10 +112,15 @@ object RefChecks {
    *  a class or module with same name
    */
   private def checkCompanionNameClashes(cls: Symbol)(implicit ctx: Context): Unit =
-    if (!(cls.owner is ModuleClass)) {
-      val other = cls.owner.linkedClass.info.decl(cls.name).symbol
-      if (other.isClass)
+    if (!cls.owner.is(ModuleClass)) {
+      def clashes(sym: Symbol) =
+        sym.name.stripModuleClassSuffix == cls.name.stripModuleClassSuffix &&
+        sym.isClass
```
smarter: I would reverse the order of the conditions since `sym.isClass` is cheaper to compute

PR #3369, opened 2017-10-23
Fix #3337 - Repl CompanionObject

### LINKED ISSUE #3337: REPL tab completes $ suffix for objects
I tried the following in the new REPL from tag v0.4.0-RC1

```scala
scala> scala.Pre<TAB>
Predef$  Predef
scala> scala.Predef$<ENTER>
1 |scala.Predef$
  |^^^^^^^^^^^^^
  |value `Predef$` is not a member of scala - did you mean `scala.Predef`?
```
I expected it to tab complete to `Predef` without the $ suffix.

rsoeldner:
by adding check if `ModuleClass`

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -115,7 +115,7 @@ object Interactive {
   /** Filter for names that should appear when looking for completions. */
   private[this] object completionsFilter extends NameFilter {
     def apply(pre: Type, name: Name)(implicit ctx: Context): Boolean =
-      !name.isConstructorName && name.is(SimpleNameKind)
+      !name.isConstructorName && name.toTermName.info.kind == SimpleNameKind
   }
 
   /** Find named trees with a non-empty position whose symbol match `sym` in `trees`.
--- a/compiler/test/dotty/tools/repl/TabcompleteTests.scala
+++ b/compiler/test/dotty/tools/repl/TabcompleteTests.scala
@@ -29,6 +29,12 @@ class TabcompleteTests extends ReplTest {
       comp.suggestions.sorted == List("slice", "sliding"))
   }
 
+  @Test def tabCompleteModule: Unit =
+    fromInitialState{ implicit s =>
+      val comp = tabComplete("scala.Pred")
+      assertEquals(comp.suggestions,List("Predef"))
+    }
+
   @Test def autoCompleteValAssign: Unit =
     fromInitialState { implicit s => tabComplete("val x = 5") }
 


compiler/test/dotty/tools/repl/TabcompleteTests.scala
```diff
@@ -29,6 +29,9 @@ class TabcompleteTests extends ReplTest {
       comp.suggestions.sorted == List("slice", "sliding"))
   }
 
+  @Test def classModuleHiding: Unit =
+    fromInitialState{ implicit s => assert(tabComplete("scala.Predef").suggestions.isEmpty) }
```
allanrenucci: I would name the test `tabCompleteModule` and add this to the test:
```scala
fromInitialState { implicit s =>
  val comp = tabComplete("scala.Pre")
  assertEquals(comp.suggestions, List("Predef"))
}
```

compiler/test/dotty/tools/repl/TabcompleteTests.scala
```diff
@@ -29,6 +29,13 @@ class TabcompleteTests extends ReplTest {
       comp.suggestions.sorted == List("slice", "sliding"))
   }
 
+  @Test def tabCompleteModule: Unit =
+    fromInitialState{ implicit s =>
+      val comp = tabComplete("scala.Pred").suggestions
+    assertTrue(s""""Expected single element "Predef" got: ${comp}""",
```
allanrenucci: Weird indentation

allanrenucci: You don't need braces after `$` when you only have an identifier: `$comp`

allanrenucci: Also, I actually think it is better to use: `assertEquals(comp.suggestions, List("Predef"))` than `assertTrue`. The message will be automatically generated.

compiler/src/dotty/tools/dotc/util/NameTransformer.scala
```diff
@@ -54,6 +54,13 @@ object NameTransformer {
     else name
   }
 
+  def encodeString(name: String): String = name.map{ch =>
+    lazy val repl = op2code(ch)
+    if (ch <= nops && repl != null)
+      repl
+    else ch
+  }.toString()
```
allanrenucci: Did you mean to push this?

rsoeldner: No, i rebased, sorry

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -105,6 +105,7 @@ object Interactive {
           buf ++= prefix.member(name).altsWith{ d =>
             !d.isAbsent &&
             !d.is(Synthetic) && !d.is(Artifact) &&
+            !d.is(ModuleClass) &&
```
smarter: I think there's a better fix possible. Notice that we use `completionsFilter` to restrict the symbols we complete on, the definition below is:
```scala
     def apply(pre: Type, name: Name)(implicit ctx: Context): Boolean =
       !name.isConstructorName && name.is(SimpleNameKind)
```
This is supposed to filter out all the non-simple names but the check is wrong `name.is(SimpleNameKind)` will also return true if the underlying name kind is `SimpleNameKind`, the check should be replaced by ` name.toTermName.info.kind == SimpleNameKind`. If you do this then you won't have to explicitly filter out `ModuleClass`.

rsoeldner: Thank you for explanation!

smarter:
Also I'll note in passing that there's another use of `is(SimpleNameKind)` in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/repl/ReplDriver.scala#L261 that should probably be changed too, although I haven't checked it exactly what it does.

rsoeldner:
@smarter done :+1:

PR #3375, opened 2017-10-24
Make rarely updated context fields elements of the store

odersky:
This avoids copying them with the rest of the context, so it should save space.

Based on #3366


odersky:
test performance please

odersky:
rebased

odersky:
test performance please

PR #3376, opened 2017-10-24
Publish Dotty baseVersion when generating Documentation

allanrenucci:


smarter:
I suggest also writing the latest nightly version to a file when publishing a nightly build, that way `dottyLatestNightlyBuild` can just fetch that file.

smarter:
I would call the files `versions/latest-rc` and `versions/latest-nightly`

allanrenucci:
I am only writing the base version on purpose. The reason is that I haven't found a good solution to keep the nightly version that would be on the website in sync with what is on maven.

If I write down the nightly version at the time we publish a nightly, then the website will list a version that is not yet on maven (it takes quite some time ~30min). That means that `dottyLatestNightlyBuild` will be broken for ~30min every days which us pretty bad.

allanrenucci:
The actual solution is a bit hacky, but at least `dottyLatestNightlyBuild` works all the time. Another solution would be to fetch the latest nightly version from maven and write it down when we publish the doc. The downside, is that `dottyLatestNightlyBuild` will not always return the **latest** nightly

smarter:
OK I see.

PR #3377, opened 2017-10-24
Link directly to a JAR

nicolasstucki:


smarter:
What's the motivation here?

nicolasstucki:
The motivation is to create a tool that given the classpath and an entry point, will link all classes that have a tasty. Later on optimizing them with the linker. The output is a jar to avoid overriding class files that may be in the output directory, making sure that the use of the tool is idempotent.

smarter:
Why not use `-d foo.jar` to write classfileloader directly to a jar instead of an output directory? I think scalac has this feature

nicolasstucki:
That was the intended effects thought I took a complicated solution. I will adapt it to the `-d foo.jar` scheme.

nicolasstucki:
I factored a part that handles JARs in #3382.

PR #3378, opened 2017-10-24
Include dottyLib when custom -classpath is added

nicolasstucki:


project/Build.scala
```diff
@@ -518,8 +518,8 @@ object Build {
         val args: Seq[String] = spaceDelimited("<arg>").parsed
 
         val fullArgs = args.span(_ != "-classpath") match {
-          case (beforeCp, Nil) => beforeCp ++ ("-classpath" :: dottyLib :: Nil)
-          case (beforeCp, rest) => beforeCp ++ rest
+          case (beforeCp, "-classpath" :: cp :: rest) => beforeCp ++ List("-classpath", cp + ":" + dottyLib) ++ rest
+          case (beforeCp, _) => beforeCp ++ ("-classpath" :: dottyLib :: Nil)
```
allanrenucci: Let's be consistent and use the `List(..)` syntax here too

project/Build.scala
```diff
@@ -533,8 +533,8 @@ object Build {
...
```
allanrenucci: same

project/Build.scala

project/Build.scala

Issue #3113, opened 2017-09-13
Meta-issue: Port Backend improvements


nicolasstucki:
- [x] Port https://github.com/scala/scala/pull/4963
- [x] scala/scala#6070
- [x] [Desugar string concat to java.lang.StringBuilder](https://github.com/scala/scala/pull/4737) (https://github.com/lampepfl/scala/pull/27)
- [x] [Avoid boxing primitives in string concatenation](https://github.com/scala/scala/pull/4944) (https://github.com/lampepfl/scala/pull/27)
- [x] [upgrade asm to 5.1](https://github.com/scala/scala/pull/5293) -- (https://github.com/lampepfl/scala/pull/24)
- [x] [Restrict library and reflect to compact1 profile](https://github.com/scala/scala/pull/6164)
- [x] [Add bridges for overridden methods in lambda indy call](https://github.com/scala/scala/pull/6087)
   - Maybe shouldn't be fixed in backend (["Perhaps this could be revisited for 2.13, which would simplify things for SJS (it wouldn't be a backend change anymore)"](https://github.com/scala/scala/pull/6087#issuecomment-330585723))
   - Ended up being done in erasure, see Erasure#typedClosure
- [ ] [Compiler crash in GenBCode referring to a class that extends an annotation](https://github.com/lampepfl/dotty/issues/3403)
  - Maybe shouldn't be fixed in backend (["I'm fine with the approach of repairing this in the backend for now, but agree that a special case in the typer would be worth trying."](https://github.com/scala/scala/pull/4638#issuecomment-122153586))
- [x] [Fix code gen with returns in nested try-finally blocks](https://github.com/scala/scala/pull/5509)
- [ ] [ Don't emit forwarder in mirror class for bridge methods](https://github.com/scala/scala/pull/6531)
  - We should also consider whether we can emit static forwarders before the backend (["This looks rather difficult to fix with the way forwarders are currently done. I think the better solution would be to generate static forwarders earlier in the pipeline and have them transformed normally."](https://github.com/scala/bug/issues/5289#issuecomment-292400867))

nicolasstucki:
@smarter, could you add the one you know should be ported.

allanrenucci:
Maybe scala/scala#6012 ?

smarter:
> https://github.com/scala/scala/pull/6012

[...]

PR #3382, opened 2017-10-25
Support -d with .jar paths

nicolasstucki:


compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -46,14 +47,37 @@ class GenBCode extends Phase {
...
+  override def runOn(units: List[CompilationUnit])(implicit ctx: Context) = {
+    val output = outputDir
+    if (output.isDirectory) {
+      classOutput = output
+      val res = super.runOn(units)
+      classOutput = null
+      res
+    } else {
+      assert(output.hasExtension("jar"))
+      classOutput = new PlainDirectory(Path(Path(output.file).parent + "/tmp-jar-" + System.currentTimeMillis().toHexString).createDirectory())
```
smarter: scalac has code to write directly to a JAR instead of using a temporary directory which should be faster, maybe we can reuse that: https://github.com/scala/scala/blob/d41c97f13ea2348ec6f59c24ae098d83a96f30dc/src/compiler/scala/tools/nsc/backend/jvm/ClassfileWriter.scala#L23 Although apparently in Java >= 7 there's already an API dedicated to this: https://github.com/sbt/zinc/issues/305#issuecomment-308093743

nicolasstucki: The `ClassfileWriter` came in the large refactoring made in scala/scala#6012. You said previously in #3113 that you would rather not port this one.

nicolasstucki: The Java 7 API does not allow direct access to the files in the jar.

`Once you have an instance of a zip file system, you can invoke the methods of the java.nio.file.FileSystem and java.nio.file.Path classes to perform operations such as copying, moving, and renaming files, as well as modifying file attributes.` [quote link](https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html)

allanrenucci: Can't you do something like:
```scala
def outputDir(implicit ctx: Context): AbstractFile = {
  if (isJar) {
    val jar = Paths.get(pathTojar)
    val fs = FileSystems.newFileSystem(jar)
    new PlainNioFile(fs.getPath("/"))
  }
  ...
}
```

nicolasstucki: I will try

nicolasstucki: It worked with the `java.nio.file. FileSystem` jar.

nicolasstucki:
Needs rebase on #3395 one merged

nicolasstucki:
Rebased

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -164,12 +154,12 @@ class Path private[io] (val jpath: JPath) {
...
   def stripExtension: String = name stripSuffix ("." + extension)
   // returns the Path with the extension.
-  def addExtension(ext: String): Path = Path(path + "." + ext)
+  def addExtension(ext: String): Path = new Path(jpath.resolveSibling(name + ext))
   // changes the existing extension out for a new one, or adds it
   // if the current path has none.
   def changeExtension(ext: String): Path =
     if (extension == "") addExtension(ext)
-    else Path(path.stripSuffix(extension) + ext)
+    else new Path(jpath.resolveSibling(stripExtension + "." + extension))
```
nicolasstucki: `new Path(jpath.resolveSibling(stripExtension + "." + ext))`

compiler/src/dotty/tools/io/JarFS.scala
```diff
@@ -22,30 +22,10 @@ class JarFS private (private[this] var jarFS: FileSystem) {
...
-
-  private class JarFile(path: JPath) extends PlainFile(new File(path)) {
-    override def name = "<jar-file>"
-    override def file = throw new UnsupportedOperationException
-    override def output = {
-      if (path.getParent ne null)
-        Files.createDirectories(path.getParent)
-      Files.newOutputStream(path)
-    }
-  }
```
nicolasstucki: :)

compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -52,15 +52,14 @@ class GenBCode extends Phase {
...
   def outputDir(implicit ctx: Context): AbstractFile = {
-    val path = ctx.settings.outputDir.value
-    if (path.isDirectory) new PlainDirectory(path.toDirectory)
-    else {
-      if (jarFS == null) {
-        path.delete()
+    val path = Directory(ctx.settings.outputDir.value)
+    if (path.extension == "jar") {
+      if (jarFS == null)
         jarFS = JarFS.create(path)
```
nicolasstucki: We should make sure that the jar is empty or all files are rewritten.

compiler/src/dotty/tools/io/JarFS.scala
```diff
@@ -0,0 +1,28 @@
...
+    new PlainDirectory(Directory(root))
+  }
+
+  def close() = jarFS.close()
+}
+
+object JarFS {
+  def create(path: Path): JarFS = {
+    assert(path.extension == "jar")
+    val env = Map("create" -> "true").asJava
```
nicolasstucki: @allanrenucci here is the documentation:
https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html
https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemproviderprops.html

allanrenucci:
test performance please

compiler/src/dotty/tools/io/JarFS.scala
```diff
@@ -0,0 +1,28 @@
...
+package io
+
+import java.nio.file.{FileSystem, FileSystems}
+import java.nio.file.Files
+
+import java.util.jar._
+
+import scala.collection.JavaConverters._
+
+class JarFS private (private[this] var jarFS: FileSystem) {
```
smarter: Please add some documentation :).

compiler/src/dotty/tools/io/JarFS.scala
smarter: I would add these links in comments if they're useful to understand the code

nicolasstucki:
test performance please

allanrenucci:
We definitely got a speed up with this PR. It was not intentional though :joy:

smarter:
Where does the speed-up comes from?

nicolasstucki:
We removed implicit conversion.

nicolasstucki:
@allanrenucci  parts LGTM

PR #3384, opened 2017-10-25
Remove redundant conversion of settings to array

nerush:
Taking into account that `settings` is of type `Array[String]`, there is no need to convert `settings` to array.

### DIFF
--- a/compiler/src/dotty/tools/repl/ReplDriver.scala
+++ b/compiler/src/dotty/tools/repl/ReplDriver.scala
@@ -89,7 +89,7 @@ class ReplDriver(settings: Array[String],
   /** Create a fresh and initialized context with IDE mode enabled */
   private[this] def initialCtx = {
     val rootCtx = initCtx.fresh.addMode(Mode.ReadPositions).addMode(Mode.Interactive)
-    val ictx = setup(settings.toArray, rootCtx)._2.fresh
+    val ictx = setup(settings, rootCtx)._2.fresh
     ictx.base.initialize()(ictx)
     ictx
   }


compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -89,7 +89,7 @@ class ReplDriver(settings: Array[String],
   /** Create a fresh and initialized context with IDE mode enabled */
   private[this] def initialCtx = {
     val rootCtx = initCtx.fresh.addMode(Mode.ReadPositions).addMode(Mode.Interactive)
-    val ictx = setup(settings.toArray, rootCtx)._2.fresh
+    val ictx = setup(settings, rootCtx)._2.fresh
```
nicolasstucki: This was probably intended, as it creates a new `Array` making sure that it cannot be modified from outside the `ReplDriver`.

allanrenucci: `toArray` does not create a new Array:
```scala
scala> val foo = Array(1 , 2)
foo: Array[Int] = Array(1, 2)

scala> foo eq foo.toArray
res0: Boolean = true
```

allanrenucci: @felixmulder Did you intend to make a copy?

felixmulder: Nope, I think this is a remnant from when the `ReplDriver` didn't take an array.

nerush: Great. 
Btw, I have just signed the CLA.

nerush: I have checked the classes involved  in processing `settings` and have not found any mutation of the array, only conversion to list in `CompilerCommand`. Shall I defend the array from potential yet no existing side-effects by making a copy ? Seems like premature optimisation to me, but you decide :)

PR #3386, opened 2017-10-25
Fix #3352: Avoid widening method TermRefs in compatibility checks

### LINKED ISSUE #3352: Bad interaction between overloading, default arguments and implicit search

odersky:
When normalizing for a compatibility check we should keep a method reference
instead of its widened method type as long as possible. This is because only
method references know about default arguments, once we widen to a method type
default arguments are no longer taken into account.

Supersedes #3357, from which it takes the essential solution idea.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -101,7 +101,9 @@ object ProtoTypes {
         val mbr = if (privateOK) tp1.member(name) else tp1.nonPrivateMember(name)
         def qualifies(m: SingleDenotation) =
           memberProto.isRef(defn.UnitClass) ||
-          compat.normalizedCompatible(m.info, memberProto)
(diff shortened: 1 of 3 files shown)


odersky:
> LGTM, except that the added test is incorrect.

Ah yes, missed this one by a digit.

PR #3391, opened 2017-10-26
Backend upgrade with StringBuilder and string concatenation optimizations

nicolasstucki:


nicolasstucki:
test performance please

PR #3392, opened 2017-10-26
Fix #3339: Widen access of private classes accessed from inner packages

### LINKED ISSUE #3339: Missing access checks when accessing package private classes
This code shouldn't compile:
```scala
package outer {
  private class A

  package inner {
    object Test {
        println(new A)
      }
    }
  }
}
```
Because if you try to run it you get:
```scala
% dotr outer.inner.Test
Exception in thread "main" java.lang.IllegalAccessError: tried to access class outer.A from class outer.inner.Test$
        at outer.inner.Test$.main(i3302.scala:7)
        at outer.inner.Test.main(i3302.scala
```

This code compiles and run with scalac because the class `A` is emitted as `public` whereas dotty emits it as `private`.

odersky:
Fixes #3339.

### DIFF
--- a/tests/pos/i3339.scala
+++ b/tests/pos/i3339.scala
@@ -1,0 +1,15 @@
+package outer {
+  private class A { override def toString = "A" }
+
+  package inner {
+    object Test {
+      def main(args: Array[String]): Unit = {
+        println(new A)
+      }
+    }
+  }
+}
+
+object Test extends App {
+  outer.inner.Test.main(Array())
+}
--- a/compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
@@ -71,29 +71,33 @@ class ExpandPrivate extends MiniPhase with IdentityDenotTransformer { thisPhase 
    *  If we change the scheme at one point to make static module class computations
    *  static members of the companion class, we should tighten the condition below.
    */
-  private def ensurePrivateAccessible(d: SymDenotation)(implicit ctx: Context) =
+  private def ensurePrivateAccessible(d: SymDenotation)(implicit ctx: Context): Unit =
     if (isVCPrivateParamAccessor(d))
       d.ensureNotPrivate.installAfter(thisPhase)
-    else if (d.is(PrivateTerm) && !d.owner.is(Package) && d.owner != ctx.owner.enclosingClass) {
-      // Paths `p1` and `p2` are similar if they have a common suffix that follows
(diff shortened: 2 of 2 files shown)


tests/pos/i3339.scala
```diff
@@ -0,0 +1,15 @@
...
+      def main(args: Array[String]): Unit = {
+        println(new A)
+      }
+    }
+  }
+}
+
+object Test extends App {
+  outer.inner.Test.main(Array())
+}
```
allanrenucci: Should be in `tests/run`?

retronym:
This fix appears to assume joint compilation. Scalac emits the class as public regardless of a reference in the currently compilation batch.

```
scala> :paste -raw
// Entering paste mode (ctrl-D to finish)

package p1 { private class C }

// Exiting paste mode, now interpreting.

scala> :javap -public p1.C
Compiled from "<pastie>"
public class p1.C {
  public p1.C();
}
```

smarter:
Yes, things get weird when we split the original test case in two files:
`A.scala`:
```scala
package outer {
  private class A { override def toString = "A" }
}
```
`B.scala`:
```scala
package outer {
  package inner {
    object Test {
        println(new A)
      }
    }
  }
}
```

```shell
% dotc A.scala B.scala
% dotc A.scala
% dotc B.scala
-- [E006] Unbound Identifier Error: try/i3339-B.scala:5:20 ---------------------
5 |        println(new A)
  |                    ^
  |                    not found: type A
```

I think `A` should not be seen from `B.scala` even under join compilation. That way we can still emit the class as private in the common case where we don' t have nested packages accessing private classes.

By the way, hopefully all of the mess relating to making things accessible should get easier once/if the JVM gets nestmates: http://openjdk.java.net/jeps/181

retronym:
Nestmates will likely help us for companions, no widening will be needed for:

```
object O { private def foo(o: O) = o.bar; class B { private def bar = 42; def baz = O.foo(this) }
```

And for nested classes (as is the intended use case in Java). Basically, it helps in intra-compilation unit cases as alternative to publicising or using accessibilty bridges.

But it won't help anything that admits separate compilation.

I think I'd prefer that the language just followed the JVM convention for packages. I'd be forced to declare a few more things public, but would have the confidence that the things that were non-public stayed that way. Classes that are really intended to be implementation details but need to be shared between a group of packages could be added to an `foo.internal` subpackage which would not be exported from the JPMS module.

odersky:
@retronym I think agree with you. It's probably better to change the rules so that private classes cannot be accessed from inner packages.

allanrenucci:
Closing this until we figure out what we want to do with #3339

PR #3393, opened 2017-10-26
Optimize MegaPhase

odersky:
#3366 did lose some performance relative to the previous tree transform scheme.

http://dotty-bench.epfl.ch/3366/

shows a loss of about 4% for scalapb and dotty. This PR is trying to get it back by reverting to a scheme close to the previous TreeTransform. We keep the new names, and shorten everything by about 20%, but otherwise rely on a similar dispatch algorithm as before.


odersky:
test performance please

odersky:
test performance please

odersky:
How can I find out what failed?

liufengyun:
@odersky This is a regression of the bench infrastructure. I've reverted the changes on the server to the old code.

liufengyun:
test performance please

liufengyun:
The website is already pushed to github as usual, but seems gh-pages have some problem in publishing it:

https://github.com/liufengyun/bench/tree/gh-pages/3393 

I plan to host the site on our own server instead.

odersky:
test performance please

odersky:
test performance please

odersky:
Tests are still a bit too noisy to be able to isolate effects of changes in details but we are at all-time lows for compile times for Vector, scalpb and dotty, so I'd say it's a success overall.

smarter:
Note that the removal of TreeTransform is likely to have a signficant impact on the benchmark here since you're compiling a thousand lines of code less. Is it possible that the original MegaPhase stuff is slower for dotty just because you were compiling more code?

odersky:
@smarter Yes, removal of TreeTransform is surely a large part of the improvement of dotty. But we see smaller improvements for scalapb and Vector as well.

PR #3394, opened 2017-10-27
Add an error message class for the method's missing arguments

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for the following case `Typer.scala:1898 (def missingArgs)`

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1908,7 +1908,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     def methodStr = err.refStr(methPart(tree).tpe)
 
     def missingArgs(mt: MethodType) = {
-      ctx.error(em"missing arguments for $methodStr", tree.pos)
+      ctx.error(MissingEmptyArgumentListForMethod(methPart(tree).symbol, methodStr), tree.pos)
       tree.withType(mt.resultType)
     }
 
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1908,7 +1908,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     def methodStr = err.refStr(methPart(tree).tpe)
 
     def missingArgs(mt: MethodType) = {
-      ctx.error(em"missing arguments for $methodStr", tree.pos)
+      ctx.error(MissingArgumentsForMethod(methodStr), tree.pos)
```
allanrenucci: This one is tricky. The idea of error messages is to contains useful information not only for the compiler but also for tools such as IDE. A string is not very helpful for an IDE. Ideally we would want to include the symbol of the method as well as the missing arguments. It should be something like:
```scala
case class MissingArgumentsForMethod(method: Symbol, missings: List[Symbol])(implicit ctx: Context)
```
You should look at what `def methodStr` is doing in order to get this information.

maseev: I totally agree that providing useful information for IDEs is a must, but here's the thing, this error message is about using a so-called nullary method which is described in [this](https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/auto-apply.md) document. 

So, there are actually no missing arguments which a user forgot to provide because a method doesn't accept any arguments.

Maybe the initial error message is a little bit ambiguous. Because when I see this:

```
scala> def greet(): String = "hello" 
def greet(): String
scala> greet 
1 |greet
  |^^^^^
  |missing arguments for method greet
```

It makes me think that I forgot to provide the actual arguments for the method when the actual problem is that Dotty is stricter than Scala in terms of calling nullary methods.

allanrenucci: I agree the error message as well as the name of the message is ambiguous. It this message only applies for missing empty parameter list, then I would update the message accordingly

allanrenucci:
Superseded by #3407

PR #3395, opened 2017-10-27
Use nio.Path instead of io.File to properly support PlainNioFile

nicolasstucki:


compiler/src/dotty/tools/io/AbstractFile.scala
```diff
@@ -94,7 +95,7 @@ abstract class AbstractFile extends Iterable[AbstractFile] {
   def path: String
 
   /** Returns the path of this abstract file in a canonical form. */
-  def canonicalPath: String = if (file == null) path else file.getCanonicalPath
+  def canonicalPath: String = if (jpath == null) path else jpath.toFile.getCanonicalPath
```
allanrenucci: I think it is equivalent to `jpath.normalize.toString`

compiler/src/dotty/tools/io/AbstractFile.scala
```diff
@@ -107,18 +108,26 @@ abstract class AbstractFile extends Iterable[AbstractFile] {
   def container : AbstractFile
 
   /** Returns the underlying File if any and null otherwise. */
-  def file: JFile
+  def file: JFile = try {
```
allanrenucci: Ideally we should remove this

compiler/src/dotty/tools/io/AbstractFile.scala
```diff
@@ -232,9 +241,14 @@ abstract class AbstractFile extends Iterable[AbstractFile] {
     val lookup = lookupName(name, isDir)
     if (lookup != null) lookup
     else {
-      val jfile = new JFile(file, name)
-      if (isDir) jfile.mkdirs() else jfile.createNewFile()
-      new PlainFile(jfile)
+      val path = jpath.resolve(name)
+      try {
+        if (isDir) Files.createDirectories(path)
+        else Files.createFile(path)
```
allanrenucci: This does not create intermediate directories if necessary. We need to make sure behaves as before

compiler/src/dotty/tools/io/AbstractFile.scala
```diff
@@ -232,9 +241,14 @@ abstract class AbstractFile extends Iterable[AbstractFile] {
...
+      } catch {
+        case _: FileAlreadyExistsException =>
```
allanrenucci: I guess it should never happen, otherwise `lookupName` would not have returned `null`

compiler/src/dotty/tools/io/File.scala
```diff
@@ -41,12 +41,12 @@ object File {
...
  *  ''Note:  This is library is considered experimental and should not be used unless you know what you are doing.''
  */
-class File(jfile: JFile)(implicit constructorCodec: Codec) extends Path(jfile) with Streamable.Chars {
+class File(jpath: JPath)(implicit constructorCodec: Codec) extends Path(jpath) with Streamable.Chars {
   override val creationCodec = constructorCodec
 
   override def addExtension(ext: String): File = super.addExtension(ext).toFile
   override def toAbsolute: File = if (isAbsolute) this else super.toAbsolute.toFile
-  override def toDirectory: Directory = new Directory(jfile)
+  override def toDirectory: Directory = new Directory(jfile.toPath)
```
allanrenucci: Why not simply `new Directory(jpath)`?

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -54,10 +55,10 @@ object Path {
 
   def apply(path: String): Path = apply(new JFile(path))
   def apply(jfile: JFile): Path = try {
```
allanrenucci: I think we should replace this method by `def apply(jpath: JPath): Path`

nicolasstucki:
The CI should not have passed on 72fa00d.

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -141,12 +144,12 @@ class Path private[io] (val jfile: JFile) {
...
       if (segments.nonEmpty && segments.last == "..")
         (path / "..").toDirectory
-      else jfile.getParent match {
+      else jpath.getParent match {
         case null =>
           if (isAbsolute) toDirectory // it should be a root. BTW, don't need to worry about relative pathed root
           else Directory(".")         // a dir under pwd
         case x    =>
-          Directory(x)
+          Directory(x.toFile)
```
allanrenucci: There should probably be a constructor that takes a `JPath`. And I think we should remove the `JFile` constructor

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -173,64 +176,68 @@ class Path private[io] (val jfile: JFile) {
...
    */
-  def deleteRecursively(): Boolean = deleteRecursively(jfile)
-  private def deleteRecursively(f: JFile): Boolean = {
-    if (f.isDirectory) f.listFiles match {
-      case null =>
-      case xs   => xs foreach deleteRecursively
-    }
-    f.delete()
+  def deleteRecursively(): Boolean = deleteRecursively(jpath)
+  private def deleteRecursively(p: JPath): Boolean = {
```
allanrenucci: I think the proper way to delete recursively with nio is:
```scala
def deleteRecursively(dir: Path): Unit = if (Files.exists(dir)) {
  Files.walkFileTree(dir, new SimpleFileVisitor[Path]() {
    override def visitFile(file: Path, attrs: BasicFileAttributes) = {
      Files.delete(file)
      FileVisitResult.CONTINUE
    }
    override def postVisitDirectory(dir: Path, exc: IOException) = {
      Files.delete(dir)
      FileVisitResult.CONTINUE
    }
  })
```
Not sure what should be the behavior if the path does not exist

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -173,64 +176,68 @@ class Path private[io] (val jfile: JFile) {
...
-    val res = jfile.createNewFile()
+    val res = tryCreate(Files.createFile(jpath))
     if (!res && failIfExists && exists) fail("File '%s' already exists." format name)
     else if (isFile) toFile
-    else new File(jfile)
+    else new File(jpath)
   }
 
+  private def tryCreate(create: => JPath): Boolean =
+    try { create; true } catch { case _: FileAlreadyExistsException => false }
```
allanrenucci: Semantic is different for something like `foo/bar/` if `foo` does not exist

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -173,64 +176,68 @@ class Path private[io] (val jfile: JFile) {
...
+  def deleteRecursively(): Boolean = deleteRecursively(jpath)
+  private def deleteRecursively(p: JPath): Boolean = {
+    import scala.collection.JavaConverters._
+    if (Files.isDirectory(p))
+      Files.list(p).iterator().asScala.foreach(deleteRecursively)
+    delete(p)
   }
 
+  private def delete(path: JPath): Boolean =
+    try { Files.delete(path); true } catch { case _: DirectoryNotEmptyException | _: NoSuchFileException => false }
```
allanrenucci: You can use `deleteIfExists` and not catch `NoSuchFileException`

compiler/src/dotty/tools/io/ZipArchive.scala
```diff
@@ -112,7 +112,7 @@ abstract class ZipArchive(override val file: JFile) extends AbstractFile with Eq
   }
 }
 /** ''Note:  This library is considered experimental and should not be used unless you know what you are doing.'' */
-final class FileZipArchive(file: JFile) extends ZipArchive(file) {
+final class FileZipArchive(file: JFile) extends ZipArchive(file.toPath) {
```
allanrenucci: Can't you use a `JPath` here instead?

.drone.yml
```diff
@@ -13,6 +13,7 @@ pipeline:
     image: lampepfl/dotty:2017-10-20
     commands:
       - cp -R . /tmp/1/ && cd /tmp/1/
+      - ./project/scripts/sbt compile # to ensure that side projects (such as dotty doc) compile
       - ./project/scripts/sbt testAll
       - ./project/scripts/sbt ";dotty-bench/jmh:run 1 1 tests/run/arrays.scala"
```
allanrenucci: This was meant to make sure that `dotty-bench` compiles. Should it be part of dottyRoot so that we can get rid of this line?

allanrenucci:
In `PlainFile`, I believe `def input` and `def output` can be rewritten using `nio` API

allanrenucci:
`PlainNioFile` should not exist anymore since `PlainFile` is now backed by `nio`. We should make sure, the methods in `PlainFile` are implemented as the ones in `PlainNioFile`.

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
```diff
@@ -409,7 +409,7 @@ case class Site(
   }
 
   private def toSourceFile(f: JFile): SourceFile =
```
allanrenucci: Should probably be changed to take a `JPath` instead

smarter:
I don't understand what's going on in this PR. What's the motivation?

compiler/src/dotty/tools/io/AbstractFile.scala
nicolasstucki: That will be for next step. That might need changes outside of the Path/File/Directory abstractions. I would prefer to keep the changes only inside these abstractions for this first step.

.drone.yml
nicolasstucki: That would be a solution

compiler/src/dotty/tools/io/Path.scala
nicolasstucki: As we saw together the semantics are the same. In the example, an `IOException` would be thrown in both implementations.

doc-tool/src/dotty/tools/dottydoc/staticsite/Site.scala
nicolasstucki: This will be for another PR.

nicolasstucki:
test performance please

smarter:
I still would like to know the rationale behind this PR :). Also it's likely incomplete, ExtractDependencies#recordDependency needs an extra case to handle PlainNioFiles: https://github.com/dotty-staging/dotty/blob/813d376d78ccf752b480882fbcb58cd6a35431a4/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala#L98

allanrenucci:
@smarter I believe writing to a jar required using the `java.nio` API but the `dotty.io` abstraction is mainly based on `java.io`. So it required some glue code to make it work.
The idea of this PR is to constantly use the `java.nio` everywhere and avoid conversions between `JFile` and `JPath`. Also, `java.nio` is supposed to be faster

nicolasstucki:
ExtractDependencies#recordDependency won't have any problems, I removed `PlainNioFile`.

nicolasstucki:
test performance please

smarter:
@retronym is there any reason scalac still has PlainFile and PlainNioFile or would the kind of refactoring done here make sense in scalac too?

retronym:
I added PlainNioFile as a minimal bridge from the AbtractFile to nio, as a
means to support JDK’s jrt:// filesystem. I think AbstractFile should
probably be removed from compiler APIs entirely in favour of direct use
NIO. A bit of experimentation is needed to make sure we can still support
in memory use cases (via a custom in memory filesystemprovider)

compiler/src/dotty/tools/io/Path.scala
```diff
@@ -173,64 +176,69 @@ class Path private[io] (val jfile: JFile) {
...
+    val res = tryCreate(if (force) Files.createDirectories(jpath) else Files.createDirectory(jpath))
     if (!res && failIfExists && exists) fail("Directory '%s' already exists." format name)
     else if (isDirectory) toDirectory
-    else new Directory(jfile)
+    else new Directory(jpath)
   }
   def createFile(failIfExists: Boolean = false): File = {
-    val res = jfile.createNewFile()
+    val res = tryCreate(Files.createFile(jpath))
+    jfile.createNewFile()
```
allanrenucci: Should this be here?

compiler/src/dotty/tools/io/Path.scala
allanrenucci: Still valid

compiler/src/dotty/tools/io/ZipArchive.scala
```diff
@@ -32,7 +33,7 @@ object ZipArchive {
    */
   def fromFile(file: File): FileZipArchive = fromFile(file.jfile)
```
allanrenucci: I would add `fromPath` and make the other delegate to it

nicolasstucki:
test performance please

compiler/src/dotty/tools/io/File.scala
```diff
@@ -41,23 +43,23 @@ object File {
...
   override def walkFilter(cond: Path => Boolean): Iterator[Path] =
     if (cond(this)) Iterator.single(this) else Iterator.empty
 
   /** Obtains an InputStream. */
-  def inputStream() = new FileInputStream(jfile)
+  def inputStream() = Files.newInputStream(jpath)
 
   /** Obtains a OutputStream. */
-  def outputStream(append: Boolean = false) = new FileOutputStream(jfile, append)
+  def outputStream(append: Boolean = false) = Files.newOutputStream(jpath, APPEND)
```
allanrenucci: `APPEND`?

compiler/src/dotty/tools/io/ZipArchive.scala
```diff
@@ -30,9 +31,9 @@ object ZipArchive {
    * @param   file  a File
    * @return  A ZipArchive if `file` is a readable zip file, otherwise null.
    */
-  def fromFile(file: File): FileZipArchive = fromFile(file.jfile)
-  def fromFile(file: JFile): FileZipArchive =
-    try   { new FileZipArchive(file) }
+  def fromFile(file: File): FileZipArchive = fromPath(file.jpath)
+  def fromPath(jpath: JPath): FileZipArchive =
+    try   { new FileZipArchive(jpath) }
     catch { case _: IOException => null }
```
allanrenucci: Not sure what can throw an `IOException` here

allanrenucci:
Also, I think you can implement `resolve`, `relativize`, and `parent` in `Path` in term of `nio`

allanrenucci:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
test performance please

PR #3399, opened 2017-10-29
Add an error message class for phantom types

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for the following case `Typer.scala:1427` (`ctx.error("only static objects can extend scala.Phantom", cdef.pos)`)


maseev:
That's interesting. For some reason a CI build failed on the clone phase :worried:

odersky:
Phantom types are most likely going to be removed from Dotty, being replaced by the "unused"  PR #3342. So there's no point at the moment to improve error messages for them.

maseev:
Wow, I didn't know that. If I'm not mistaken there are still a bunch of places in [this](https://github.com/lampepfl/dotty/issues/1589) issue related to the Phantom types. Maybe it would be a good idea to remove them from the list?

Should I close this PR for now?

PR #3400, opened 2017-10-29
Add an error message class for a function underscore case

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for the following case `Typer.scala:1555` (`ctx.errorOrMigrationWarning(i"not a function: ${res.tpe}; cannot be followed by _", tree.pos)`)

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1838,4 +1838,11 @@ object messages {
           |type used instead: $tp2"""
     }
   }
+
+  case class OnlyFunctionsCanBeFollowedByUnderscore(pt: Type)(implicit ctx: Context)
+    extends Message(OnlyFunctionsCanBeFollowedByUnderscoreID) {
+    val kind = "Syntax"
+    val msg = hl"not a function: $pt; cannot be followed by `_'"
+    val explanation = ""
+  }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1842,7 +1842,9 @@ object messages {
   case class OnlyFunctionsCanBeFollowedByUnderscore(pt: Type)(implicit ctx: Context)
     extends Message(OnlyFunctionsCanBeFollowedByUnderscoreID) {
     val kind = "Syntax"
-    val msg = hl"not a function: $pt; cannot be followed by `_'"
-    val explanation = ""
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1826,4 +1826,11 @@ object messages {
       val explanation =
        hl"A method annotated ${"@tailrec"} must be declared ${"private"} or ${"final"} so it can't be overridden."
   }
+
+  case class OnlyFunctionsCanBeFollowedByUnderscore(pt: Type)(implicit ctx: Context)
+    extends Message(OnlyFunctionsCanBeFollowedByUnderscoreID) {
+    val kind = "Syntax"
+    val msg = hl"not a function: $pt; cannot be followed by `_'"
+    val explanation = ""
```
```scala
val explanation =
  hl"""The syntax ${"x _"} is no longer supported if ${"x"} is not a function.
      |To convert to a function value, you need to explicitly write ${"() => x"}"""
```

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -1051,4 +1051,24 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+        |   val n = "T"
+        |   val func = n _
+        |  }
+        |}
+      """.stripMargin
+    }.expect { (ictx, messages) =>
+      implicit val ctx: Context = ictx
+
+      assertMessageCount(1, messages)
+
```
allanrenucci: remove this blank line

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -1051,4 +1051,24 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+      val OnlyFunctionsCanBeFollowedByUnderscore(pt) :: Nil = messages
+
```
allanrenucci: remove this blank line

maseev:
@allanrenucci I've noticed that quite a lot of PRs are merged with the merge commit, even if the PR contains only one commit. Is this intentional? Because Github [supports](https://help.github.com/articles/about-pull-request-merges/) different strategies of merging pull requests. 

You can merge the PR with a merge commit (with a git --no-ff option), or you can squash all commits in the PR in just one and merge the pull request with a fast-forward technique when the merge commit won't be created (this is basically is squash and a fast-forward merge).

allanrenucci:
We don't have a strict policy. I'd say we want to have a merge commit + meaningful commits.
If you squash with GitHub, you don't get a merge commit.
We tend not to ask external contributors to squash their commits, but ideally they should

PR #3402, opened 2017-10-29
Fix #3246: Refine handling of postfix _ for non-functions

### LINKED ISSUE #3246: Converting from `T` to `() => T`

allanrenucci:
Reopening #3300 from `dotty-staging`

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
+++ b/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
@@ -527,6 +527,7 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
     def unapply(tree: Tree): Option[(List[Tree], Tree, Tree)] = tree match {
       case Block(_, expr) => unapply(expr)
       case Closure(env, meth, tpt) => Some(env, meth, tpt)
+      case Typed(expr, _)  => unapply(expr)
       case _ => None
     }
   }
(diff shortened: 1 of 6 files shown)


allanrenucci:
This used to compile but doesn't compile anymore:
```scala
class Test {
  class A
  def foo(x: A)(y: x.type) = ()
  val bar = foo(new A()) _
 }
```
```scala
-- Error: tests/allan/Test.scala:4:25 ------------------------------------------
4 |  val bar = foo(new A()) _
  |            ^^^^^^^^^^^^^^
  |            not a function: foo(new A()); cannot be followed by `_'

one error found
```

allanrenucci:
Same for:
```scala
class Test {
  class A {
    type T
    def f(x: T): T = ???
  }
  def a: A = new A
  val f = a.f _
 }
```
```scala
-- Error: tests/allan/Test.scala:7:14 ------------------------------------------
7 |  val f = a.f _
  |          ^^^^^
  |          not a function: a.f; cannot be followed by `_'

one error found
```

liufengyun:
@allanrenucci The two examples don't compile by `2.12.3`, do we need to support in Dotty?

allanrenucci:
They compile on Dotty master and are extracted from our test suite.

I believe the second example should probably compile or at least this should compile (it compiles with scalac):
```scala
class Test {
  class A {
    type T = Int
    def f(x: T): T = ???
  }
  def a: A = new A
  val f = a.f _
}
```
Regarding the first example, I am not sure

liufengyun:
It works now, review @allanrenucci , thanks.

allanrenucci:
LGTM. Thanks!

PR #3414, opened 2017-10-31
Fix generic signatures of generic methods

### LINKED ISSUE #3411: Missing parameter list in JavaGeneric signature
```scala
object Foo {
  def foo[U](a: Int, b: Double)(c: Boolean, d: String)(e: Object): Int = 1
}
```
will generate the signature `public <U> int Foo$.foo(int,double)` which misses all parameters not in the first parameter list.

Duhemm:
We ignored the case where a `MethodType` had another `MethodType` as
result type. This commit fixes the issue by collecting the parameters of
the result type if it is a `MethodType`, recursively.

Fixes #3411

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
+++ b/compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
@@ -240,8 +240,8 @@ object GenericSignatures {
 
         case mtpe: MethodType =>
           // phantom method parameters do not make it to the bytecode.
-          val params = mtpe.paramInfos.filterNot(_.isPhantom)
-          val restpe = mtpe.resultType
+          val params = mtpe.paramInfoss.flatten.filterNot(_.isPhantom)
+          val restpe = mtpe.finalResultType
           builder.append('(')
           // TODO: Update once we support varargs
           params.foreach { tp =>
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
```diff
@@ -239,9 +239,14 @@ object GenericSignatures {
...
         case mtpe: MethodType =>
+          def allParamTypes(mtpe: MethodType): List[Type] =
+            mtpe.resultType match {
+              case restpe: MethodType => mtpe.paramInfos ::: allParamTypes(restpe)
+              case _ => mtpe.paramInfos
+            }
           // phantom method parameters do not make it to the bytecode.
-          val params = mtpe.paramInfos.filterNot(_.isPhantom)
-          val restpe = mtpe.resultType
+          val params = allParamTypes(mtpe).filterNot(_.isPhantom)
```
smarter: `.paramInfoss`

Duhemm: That's much better, thanks.

nicolasstucki:
Btw, the convention for the commit message fixing a bug would be`Fix #3411: Fix generic signatures of generic methods`.

PR #3420, opened 2017-11-01
Use new NotAMember new error format in PostTyper

DSouzaM:
Small fix. I started implementing a new class before I realized one existed already :smiley:

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -14,7 +14,7 @@ import util.Positions._
 import Decorators._
 import config.Printers.typr
 import Symbols._, TypeUtils._, SymUtils._
-import reporting.diagnostic.messages.SuperCallsNotAllowedInline
+import reporting.diagnostic.messages.{NotAMember, SuperCallsNotAllowedInline}
 
 /** A macro transform that runs immediately after typer and that performs the following functions:
  *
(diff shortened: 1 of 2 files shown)


DSouzaM:
A test is failing because actual output:
```
scala> import util.foo
1 | import util.foo
  |             ^^^
  |           value `foo` is not a member of util - did you mean `util.Left`?
```
doesn't match expected output:
```
scala> import util.foo
1 | import util.foo
  |             ^^^
  |             foo is not a member of util
```
as a result of the change (these errors are caught by [`PostTyper`](https://github.com/DSouzaM/dotty/blob/67a8152277f6c12fca8f8a4e6c3b3b0b8e3e58a1/compiler/src/dotty/tools/dotc/transform/PostTyper.scala#L275)).

Meanwhile, both actual and expected match for the following case, which is caught by [`TypeAssigner`](https://github.com/DSouzaM/dotty/blob/67a8152277f6c12fca8f8a4e6c3b3b0b8e3e58a1/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala#L235):
```
scala> import util.foo.bar
1 | import util.foo.bar
  |        ^^^^^^^^
  |        value `foo` is not a member of util - did you mean `util.Left`?
```
Is there a distinction between these two errors that causes them to be handled differently? Should they continue to have different error messages?

allanrenucci:
You can update the expected output of the test case

PR #3421, opened 2017-11-01
Rework implicit search

### LINKED ISSUE #3396: infinite loop in implicit search when using collection-strawman
The offending code (adapted from the tests in https://github.com/scala/collection-strawman/blob/master/collections-contrib/src/test/scala/strawman/collection/decorators/MapDecoratorTest.scala, which apparently have never been run with Dotty):

```scala
import strawman.collection.decorators._

object Test {
  def foo(map: strawman.collection.Map[Int, Int]): Unit = {
    map.zipByKeyWith(map)((a, b) => a)
  }
}
```

```scala
% dotc -classpath "$(coursier fetch -p ch.epfl.scala:collections-contrib_2.12:0.5.0)" map.scala
```

odersky:
Fixes #3396. I am not sure what we should do to add this as a test case. I thought collection-strawman was already part of the community build? If yes, the contrib directory should be included, that would have shown the error,


nicolasstucki:
The strawman is currently only a submodule used in the test suite and it has a fixed commit. We should probably update the strawman submodule as well.

It would be good to have it also on the community build to find issues faster.

allanrenucci:
In the community build, we build a fixed commit of the projects. I can make it work to compile the latest commit on master. However, if the community build fails, we then don't know if it is due a regression or newly added code.

nicolasstucki:
It is at least a way to detect that there is some bug somewere.

nicolasstucki:
I will update the strawman submodule.

odersky:
Let's update the strawman semi-regularly as it develops, Say every 2 weeks. /cc @julienrf

nicolasstucki:
Also, we do not compile the test sources from the strawan in the submodule

odersky:
test performance please

odersky:
The latest commits also fix #3430.

odersky:
test performance please

odersky:
test performance please

odersky:
> What's the purpose of doing this? In scalac they're going in the opposite direction: completely get rid of implicit sorting because it made implicit search depend on compilation order: scala/scala#6083

In fact I think we go to some degree in the same way. I would like to complement the current behavior which depends on compilation history with criteria that depend only on the definitions themselves. I believe order in a file is significant. If we communicate that it means programmers can arrange their implicits in the order they should be tried, similarly to clauses in a pattern match. We probably should also look at nesting level (deeper takes precedence) and class hierarchy (definitions in subclasses take precedence). If we do that we might be able to drop the useCount-based criterion.

Another aspect worth noting is that with d7c8604 a divergent implicit will be issued (or not) independent of compilation order.

odersky:
test performance please

odersky:
test performance please

odersky:
This is ready for review now. To summarize, there are the following main changes:

 - don't proceed with implicit search if result type cannot match, as outlined in the initial PR comment
 - correctly propagate ambiguities in nested implicit searches. This is more tricky than it first appears: We are not allowed to hide an ambiguity as a normal failure, nor can we abort the search prematurely if an ambiguity is detected. Instead the only correct course is to continue with all
remaining alternatives that would strictly dominate both ambiguous alternatives. If that gives no result, the ambiguity should be propagated.
 - Improve implicit error diagnostics. We now report the largest search tree if it is non-trivial and also explain why the search tree cannot be completed. The search tree is written in a way so that it can be pasted as an explicit argument into a program to enable further exploration.
 - Pre-sort implicit candidates so that candidates more likely to dominate others and candidates with smaller subsearch fanout are searched first. This helps keep the searches smaller, in particular for underconstrained queries that lead to ambiguity errors (i3430.scala is an example).

smarter:
Does the implicit candidate sorting affect semantics or is it purely an optimization?

odersky:
> Does the implicit candidate sorting affect semantics or is it purely an optimization?

It should be only an optimization.

smarter:
It would be nice to explicitly document it as such then. Also ideally we would have some benchmarks to confirm the effectiveness of this optimization.

OlivierBlanvillain:
I think the new handling of ambiguity affects semantics, here is an example that compiles with master and scalac but is handled differently in this branch (this is the ambiguity pattern @smarter mentioned deurring the last meeting):

```scala
object Test {

  trait Tagged[A]

  // Negation of being Tagged: implicit[NotTagged[A]] available only if there
  // are no implicitly[Tagged[A]] in scope.
  trait NotTagged[A]
  trait NotTaggedLowPrio {
    implicit def notTaggedInstance[A]: NotTagged[A] = null
  }
  object NotTagged extends NotTaggedLowPrio {
    implicit def notTaggedAmbiguity1[A](implicit ev: Tagged[A]): NotTagged[A] = null
    implicit def notTaggedAmbiguity2[A](implicit ev: Tagged[A]): NotTagged[A] = null
  }


  case class Foo[A](value: Boolean)
  trait FooLowPrio {
    implicit def fooDefault[A]: Foo[A] = Foo(true)
  }
  object Foo extends FooLowPrio {
    implicit def fooNotTagged[A](implicit ev: NotTagged[A]): Foo[A] = Foo(false)
  }


    implicit val taggedInt: Tagged[Int] = null

    assert(implicitly[Foo[Int]].value) // fooDefault

    assert(!implicitly[Foo[String]].value) // fooNotTagged

    println(1)
  }
}
```

odersky:
> I think the new handling of ambiguity affects semantics, here is an example that compiles with master and scalac but is handled differently in this branch (this is the ambiguity pattern @smarter mentioned during the last meeting):

Yes, it does affect semantics. But I would argue that the old semantics is fundamentally broken. 

Let's say do an implicit search of type `T` with candidates

    implicit val a1: T
    implicit val a2: T
    implicit val a3: T

That's ambiguous. But if you have

    implicit def bridge(implicit x: S): T
    implicit val b1: S
    implicit val b2: S
    implicit val a3: T

You still have three solutions for `T`: `bridge(b1)`, `bridge(b2)` and `a3`. So you should get an ambiguity error. But with the old semantics you get success with `a3` instead.

Not sure what we can do about this. How widely used is this (anti-)pattern?

smarter:
We should check if the community build still passes with this PR.

Duhemm:
The community build fails on scalacheck with this PR.

To reproduce:
1. Clone the community build
2. `DOTTY_REFERENCE="+refs/pull/3421/merge" run.sh`

<details><summary>log</summary>

```
Building scalacheck with Dotty 0.5.0-bin-SNAPSHOT...
git clone -b dotty https://github.com/dotty-staging/scalacheck .
Cloning into '.'...
remote: Counting objects: 8360, done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 8360 (delta 0), reused 2 (delta 0), pack-reused 8355
Receiving objects: 100% (8360/8360), 15.07 MiB | 2.74 MiB/s, done.
Resolving deltas: 100% (3452/3452), done.
sbt ++0.5.0-bin-SNAPSHOT jvm/compile
[warn] Executing in batch mode.
[warn]   For better performance, hit [ENTER] to switch to interactive mode, or
[warn]   consider launching sbt without any commands, or explicitly passing 'shell'
[info] Loading global plugins from /Users/martin/.sbt/0.13/plugins
[info] Loading project definition from /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/project
[info] Updating {file:/Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/project/}scalacheck-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/project/target/scala-2.10/sbt-0.13/classes...
Fetching latest Dotty nightly version (requires an internet connection)...
Latest Dotty nightly build version: Some(0.5.0-bin-20171108-28f6c3f-NIGHTLY)
[info] Set current project to scalacheck (in build file:/Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/)
[info] Setting version to 0.5.0-bin-SNAPSHOT
[info] Reapplying settings...
[info] Set current project to scalacheck (in build file:/Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/)
[info] Updating {file:/Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/}jvm...
[warn] Binary version (0.5.0-bin-SNAPSHOT) for dependency ch.epfl.lamp#scala-library;0.5.0-bin-SNAPSHOT
[warn] 	in org.scalacheck#scalacheck_0.5;1.14.0-SNAPSHOT differs from Scala binary version in project (0.5).
[info] Resolving ch.epfl.lamp#dotty-library_0.5;0.5.0-bin-SNAPSHOT ...
[warn] circular dependency found: ch.epfl.lamp#scala-library;0.5.0-bin-SNAPSHOT->ch.epfl.lamp#dotty-library_0.5;0.5.0-bin-SNAPSHOT->...
[warn] circular dependency found: ch.epfl.lamp#dotty-library_0.5;0.5.0-bin-SNAPSHOT->ch.epfl.lamp#scala-library;0.5.0-bin-SNAPSHOT->...
[info] Resolving org.scala-lang.modules#scala-xml_2.12;1.0.6 ...
[info] Done updating.
[info] Compiling 19 Scala sources to /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/jvm/target/scala-0.5/classes...
[error] -- Error: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/Arbitrary.scala:379:51 
[error] 379 |    Arbitrary(implicitly[Arbitrary[A => Option[B]]].arbitrary.map(Function.unlift))
[error]     |                                                   ^
[error]     |no implicit argument of type org.scalacheck.Arbitrary[A => Option[B]] was found for parameter e of method implicitly in object Predef.
[error]     |I found:
[error]     |
[error]     |    this.arbContainer2[C, T, U](
[error]     |      this.arbContainer2[C, T, U](
[error]     |        this.arbContainer2[C, T, U](
[error]     |          this.arbContainer2[C, T, U](
[error]     |            this.arbContainer2[C, T, U](
[error]     |              this.arbContainer2[C, T, U](
[error]     |                /* missing */implicitly[org.scalacheck.Arbitrary[(T, U)]]
[error]     |              , ???, ???)
[error]     |            , ???, ???)
[error]     |          , ???, ???)
[error]     |        , ???, ???)
[error]     |      , ???, ???)
[error]     |    , ???, ???)
[error]     |
[error]     |But method arbEnum in trait ArbitraryLowPriority produces a diverging implicit search when trying to match type org.scalacheck.Arbitrary[(T, U)].
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/Prop.scala:110:27 
[warn] 110 |        Test.cmdLineParser.printHelp
[warn]     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]     |        method printHelp must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/Properties.scala:78:27 
[warn] 78 |        Test.cmdLineParser.printHelp
[warn]    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]    |        method printHelp must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:244:42 
[warn] 244 |            if (doRun) runActions(sut,as, removeSut)
[warn]     |                                          ^^^^^^^^^
[warn]     |                      method removeSut must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:246:14 
[warn] 246 |              removeSut
[warn]     |              ^^^^^^^^^
[warn]     |              method removeSut must be called with () argument
[error] -- Error: /Users/martin/Documents/Projects.nosync/Duhemm/dotty-community-build/out/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:263:45 
[error] 263 |  def shrinkState: Shrink[State] = implicitly
[error]     |                                             ^
[error]     |ambiguous implicit arguments of type T found for parameter e of method implicitly in object Predef.
[error]     |I found:
[error]     |
[error]     |    org.scalacheck.Shrink.shrinkFractional[T](
[error]     |      /* ambiguous */implicitly[scala.math.Fractional[T]]
[error]     |    )
[error]     |
[error]     |But both object FloatIsFractional in object Numeric and object DoubleIsFractional in object Numeric match type scala.math.Fractional[T].
[info] Legend: Ln = line n, Cn = column n, En = error n
[error] (jvm/compile:compileIncremental) Compilation failed
[error] Total time: 12 s, completed Nov 11, 2017 5:05:42 PM
[info] Test run started
[info] Test dotty.communitybuild.CommunityBuildTest.squants started
[info] Test dotty.communitybuild.CommunityBuildTest.scalatest started
[info] Test dotty.communitybuild.CommunityBuildTest.algebra started
[info] Test dotty.communitybuild.CommunityBuildTest.ScalaPB started
[info] Test dotty.communitybuild.CommunityBuildTest.collectionstrawman started
[info] Test dotty.communitybuild.CommunityBuildTest.scopt started
[info] Test dotty.communitybuild.CommunityBuildTest.betterfiles started
[info] Test dotty.communitybuild.CommunityBuildTest.scalacheck started
[error] Test dotty.communitybuild.CommunityBuildTest.scalacheck failed: expected:<1> but was:<0>, took 31.794 sec
[info] Test run finished: 1 failed, 0 ignored, 8 total, 467.618s
[error] Failed: Total 8, Failed 1, Errors 0, Passed 7
[error] Failed tests:
[error] 	dotty.communitybuild.CommunityBuildTest
[error] (test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 473 s, completed Nov 11, 2017 5:05:42 PM
```

</details>

odersky:
@Duhemm Can you check whether scalacheck now works (with -language:Scala2)?

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -896,10 +898,23 @@ trait Implicits { self: Typer =>
         case cand :: pending1 =>
           tryImplicit(cand) match {
             case fail: SearchFailure =>
-              if (fail.isAmbiguous) healAmbiguous(pending1, fail)
-              else rank(pending1, found, fail :: rfailures)
+              if (isNot)
```
smarter: Wouldn't it be more efficient to check if we're looking for a `Not[T]` before doing the search, this way we know we only need to do a search for `T` once, whereas with the current implementation I guess the search will be done both for `Not.amb1` and `Not.amb2` ?

library/src/scala/implicits/Not.scala
```diff
@@ -0,0 +1,9 @@
+package scala.implicits
+
+final class Not[+T] private ()
```
smarter: Some documentation would be nice :).

compiler/src/dotty/tools/dotc/typer/Implicits.scala
odersky: In fact we do the search for `T` only once. if we fail at that, we immediately succeed for `Not[T]`.

library/src/scala/implicits/Not.scala
odersky: Sure :-) As well as spec the whole thing.

Duhemm:
@odersky Unfortunately, ScalaCheck still fails with the same error. I checked it runs with `-language:Scala2`:

<details>
<summary>log</summary>
> show jvm/scalacOptions
[info] * -language:Scala2
[success] Total time: 1 s, completed Nov 13, 2017 10:39:53 PM
> ++0.5.0-bin-SNAPSHOT jvm/compile
[info] Setting version to 0.5.0-bin-SNAPSHOT
[info] Reapplying settings...
[info] Set current project to scalacheck (in build file:/Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/)
[info] Compiling 19 Scala sources to /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/jvm/target/scala-0.5/classes...
[error] -- Error: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/Arbitrary.scala:379:51 
[error] 379 |    Arbitrary(implicitly[Arbitrary[A => Option[B]]].arbitrary.map(Function.unlift))
[error]     |                                                   ^
[error]     |no implicit argument of type org.scalacheck.Arbitrary[A => Option[B]] was found for parameter e of method implicitly in object Predef.
[error]     |I found:
[error]     |
[error]     |    this.arbContainer2[C, T, U](
[error]     |      this.arbContainer2[C, T, U](
[error]     |        this.arbContainer2[C, T, U](
[error]     |          this.arbContainer2[C, T, U](
[error]     |            this.arbContainer2[C, T, U](
[error]     |              this.arbContainer2[C, T, U](
[error]     |                /* missing */implicitly[org.scalacheck.Arbitrary[(T, U)]]
[error]     |              , ???, ???)
[error]     |            , ???, ???)
[error]     |          , ???, ???)
[error]     |        , ???, ???)
[error]     |      , ???, ???)
[error]     |    , ???, ???)
[error]     |
[error]     |But method arbEnum in trait ArbitraryLowPriority produces a diverging implicit search when trying to match type org.scalacheck.Arbitrary[(T, U)].
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/Prop.scala:110:27 
[warn] 110 |        Test.cmdLineParser.printHelp
[warn]     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]     |        method printHelp must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/Properties.scala:78:27 
[warn] 78 |        Test.cmdLineParser.printHelp
[warn]    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]    |        method printHelp must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:244:42 
[warn] 244 |            if (doRun) runActions(sut,as, removeSut)
[warn]     |                                          ^^^^^^^^^
[warn]     |                      method removeSut must be called with () argument
[warn] -- [E099] Syntax Migration Warning: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:246:14 
[warn] 246 |              removeSut
[warn]     |              ^^^^^^^^^
[warn]     |              method removeSut must be called with () argument
[warn] -- Migration Warning: /Users/martin/Documents/Projects.nosync/dotty-staging/scalacheck/src/main/scala/org/scalacheck/commands/Commands.scala:263:45 
[warn] 263 |  def shrinkState: Shrink[State] = implicitly
[warn]     |                                             ^
[warn]     |Ambiguous implicits object BigDecimalIsFractional in object Numeric and object FloatIsFractional in object Numeric
[warn]     |seem to be used to implement a local failure in order to negate an implicit search.
[warn]     |According to the new implicit resolution rules this is no longer possible;
[warn]     |the search will fail with a global ambiguity error instead.
[warn]     |
[warn]     |Consider using the scala.implicits.Not class to implement similar functionality.
[info] Legend: Ln = line n, Cn = column n, En = error n
[error] (jvm/compile:compileIncremental) Compilation failed
[error] Total time: 3 s, completed Nov 13, 2017 10:39:59 PM
</detail>

odersky:
Can someone help to fix ScalaCheck? It would be important to figure out whether this is a problem in ScalaCheck or in the new resolution algorithm. First questions to ask: What type argument does scalac infer at this point? Would the same argument work in dotty when given explicitly?

smarter:
The problematic definition is (from https://github.com/dotty-staging/scalacheck/blob/dotty/src/main/scala/org/scalacheck/Arbitrary.scala):
```scala
implicit def arbPartialFunction[A: Cogen, B: Arbitrary]: Arbitrary[PartialFunction[A, B]] =
  Arbitrary(implicitly[Arbitrary[A => Option[B]]]
    .arbitrary.map(Function.unlift))
```
scalac resolves this to:
```scala
  implicit def arbPartialFunction[A, B](implicit ev1: Cogen[A], ev2: Arbitrary[B]): Arbitrary[PartialFunction[A, B]] =
    Arbitrary(
      Arbitrary.arbFunction1(
        arbOption(ev2), ev1))
      .arbitrary.map(Function.unlift))
```
(`arbFunction1` is defined in https://github.com/dotty-staging/scalacheck/blob/dotty/src/main/scala/org/scalacheck/ArbitraryArities.scala)

smarter:
This PR successfully compiles scalacheck if we just specify the first level of implicit (`Arbitrary.arbFunction1`) in this method.

smarter:
Here's a minimal example:
```scala
class Foo[T]
object Foo {
  implicit def intFoo: Foo[Int] = ???
}

object Test {
  implicit def genFoo[T](implicit f: Foo[T => T]): Foo[T] = ???

  def test: Unit = {
    implicitly[Foo[Int]]
  }
}
```
Both scalac and dotty before this PR successfully compile this using `intFoo`, but with this PR we get:
```scala
-- Error: try/arb3bbb.scala:10:24 ----------------------------------------------
10 |    implicitly[Foo[Int]]
   |                        ^
   |no implicit argument of type Foo[Int] was found for parameter e of method implicitly in object Predef.
   |I found:
   |
   |    Test.genFoo[T](
   |      Test.genFoo[T](
   |        Test.genFoo[T](
   |          Test.genFoo[T](
   |            Test.genFoo[T](Test.genFoo[T](/* missing */implicitly[Foo[T => T]]))
   |          )
   |        )
   |      )
   |    )
   |
   |But method genFoo in object Test produces a diverging implicit search when trying to match type Foo[T => T].
```
```

**Comment by smarter** [comment]
scalacheck compiles with the latest commit.

File: compiler/src/dotty/tools/dotc/typer/Implicits.scala [hunk 3]
```diff
@@ -253,105 +256,101 @@ object Implicits {
...
+      else
+        if (argument.isEmpty) em"match expected type"
+        else em"convert from ${argument.tpe} to expected type"
 
     /** An explanation of the cause of the failure as a string */
     def explanation(implicit ctx: Context): String
+
+    def msg(implicit ctx: Context): Message = explanation
+
+    /** If search was a for an implicit conversion, a note describing the failure
```
smarter: typo: was a -> was

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -800,137 +818,187 @@ trait Implicits { self: Typer =>
...
+          compareCandidate(ambi.alt2, cand.ref, cand.level) < 0)
+        rank(newPending, fail, Nil).recoverWith(_ => fail)
+      }
+
+      def rank(pending: List[Candidate], found: SearchResult, rfailures: List[SearchFailure]): SearchResult = pending match {
+        case cand :: pending1 =>
+          tryImplicit(cand) match {
+            case fail: SearchFailure =>
+              if (isNot)
+                SearchSuccess(ref(defn.Not_value), defn.Not_value.termRef, 0)(ctx.typerState)
```
smarter: I don't have a minimized example to reproduce this but there is something wrong here: The `ctx` used here comes from the `ImplicitSearch` class implicit parameter. There is no guarantee that the TyperState of this Context is committable (this can be checked by adding `assert(ctx.typerState.isCommittable)`). This is bad because in `inferImplicit` we do `result.tstate.commit()` which crashes when the TyperState is not committable. This is only a problem here because the other call to `SearchSuccess` is done in `typedImplicit` which takes its own `implicit ctx` (the only call to `typedImplicit` is `typedImplicit(cand)(nestedContext().setNewTyperState().setSearchHistory(history))`, the `setNewTyperState()` ensures the state is always committable).

I found this out while trying to get the usage of ambiguous implicits in `shims` to work, see https://github.com/smarter/shims/commits/dotty-not

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -800,137 +818,187 @@ trait Implicits { self: Typer =>
...
-                case eliminated: SearchSuccess => condense(hits.filter(_ ne eliminated))
-                case _ => new AmbiguousImplicits(best.ref, alt.ref, pt, argument)
+      /** If `alt1` is also a search success, try to disambiguate as follows:
+       *    - If alt2 is preferred over alt1, pick alt2, otherwise return an
+       *      ambiguous implicits error.
+       */
+      def disambiguate(alt1: SearchResult, alt2: SearchSuccess) = alt1 match {
+        case alt1: SearchSuccess =>
+          val diff = compareCandidate(alt1, alt2.ref, alt2.level)
+          assert(diff <= 0)
```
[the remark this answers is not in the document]
odersky: A added the comment: diff > 0 candidates should already have been eliminated in `rank`

compiler/src/dotty/tools/dotc/typer/Implicits.scala
odersky: Well spotted! Yes, we need to give it a nested committable typerstate.

odersky:
I found another problem with the treatment of `Not`: We need to continue in each case as if it was a normal success/failure, instead of returning directly from `rank`.

OlivierBlanvillain:
I tried compiling my library with this PR:

- Everything works as it used to under -language:Scala2
- I could replace all my `NotX` by `implicits.Not[X]` and compile without -language:Scala2!

smarter:
@milessabin: This PR changes the behavior of implicit search, I believe this sort of things is relevant to your interests :). In particular, an ambiguous implicit now invalidates the whole search instead of just the current branch, see https://github.com/lampepfl/dotty/pull/3421#issuecomment-343657733. To compensate for this, negation of implicits is now explictly supported through https://github.com/dotty-staging/dotty/blob/fix-strawan/library/src/scala/implicits/Not.scala. Let us know if you can think of other patterns this may break.

PR #3425, opened 2017-11-01
Compile root in CI

nicolasstucki:


project/Build.scala
```diff
@@ -1134,6 +1140,7 @@ object Build {
       bootstrappedAggregate(`scala-library`, `scala-compiler`, `scala-reflect`, scalap, `dotty-language-server`).
       dependsOn(dottyCompiler).
       dependsOn(dottyLibrary).
+      dependsOn(dottyBench). // just to make sure it compiles
```
allanrenucci: I think you write `dependsOn(dottyCompiler, dottyLibrary, dottyBench)`

.drone.yml
```diff
@@ -13,16 +13,15 @@ pipeline:
     image: lampepfl/dotty:2017-10-20
     commands:
       - cp -R . /tmp/1/ && cd /tmp/1/
+      - ./project/scripts/sbt compile
```
allanrenucci: Let's do `./project/scripts/sbt ";compile ;testAll"` to avoid restarting sbt

project/Build.scala
```diff
@@ -1132,8 +1138,7 @@ object Build {
     def asDottyRoot(implicit mode: Mode): Project = project.withCommonSettings.
       aggregate(`dotty-interfaces`, dottyLibrary, dottyCompiler, dottyDoc, dottySbtBridgeReference).
       bootstrappedAggregate(`scala-library`, `scala-compiler`, `scala-reflect`, scalap, `dotty-language-server`).
-      dependsOn(dottyCompiler).
-      dependsOn(dottyLibrary).
+      dependsOn(dottyCompiler, dottyLibrary, dottyBench).
```
smarter: Please don't do this. sbt currently downloads the dotty artifact and it's dependencies, do everyone will download dotty-bench if you add it as a dependency

allanrenucci: Can we add something like a `noPublish` setting to `dottyBench` so that is is not included in the dotty artefact?

smarter: If you don't publish it, it will still be listed as a dependency, and downloading dotty will fail when the dependency is not found.

allanrenucci: Same if we put it in `aggregate` and not `dependsOn`?

nicolasstucki: I will remove this part

.drone.yml
```diff
@@ -13,16 +13,14 @@ pipeline:
     image: lampepfl/dotty:2017-10-20
     commands:
       - cp -R . /tmp/1/ && cd /tmp/1/
-      - ./project/scripts/sbt testAll
-      - ./project/scripts/sbt ";dotty-bench/jmh:run 1 1 tests/run/arrays.scala"
+      - ./project/scripts/sbt compile testAll "dotty-bench/jmh:run 1 1 tests/pos/alias.scala"
```
allanrenucci: This does not work unfortunately. The `project/script/sbt` script expects one argument. Either use the old syntax or change the script

allanrenucci: Also, do we have to run the `dotty-bench` project? Can we just compile it?

.drone.yml
```diff
@@ -59,7 +57,7 @@ pipeline:
     environment:
       - NIGHTLYBUILD=yes
     commands:
-      - ./project/scripts/sbtPublish ";dotty-bootstrapped/publishSigned ;sonatypeRelease"
+      - ./project/scripts/sbtPublish dotty-bootstrapped/publishSigned sonatypeRelease
```
allanrenucci: `project/script/sbtPublish` also expects one argument. I am fine with changing it

.drone.yml
nicolasstucki: No, I remembered that I added it because at some point we had a runtime bug on that project that stayed undetected for a while.

.drone.yml
```diff
@@ -13,16 +13,14 @@ pipeline:
...
+      - ./project/scripts/sbt ";compile ;testAll ;dotty-bench/jmh:run 1 1 tests/pos/alias.scala"
 
   test_bootstrapped:
     group: test
     image: lampepfl/dotty:2017-10-20
     commands:
       - cp -R . /tmp/2/ && cd /tmp/2/
-      - ./project/scripts/sbt dotty-bootstrapped/testAll
-      - ./project/scripts/sbt ";dotty-bench-bootstrapped/jmh:run 1 1 tests/run/arrays.scala"
+      - ./project/scripts/sbt ";dotty-bootstrapped/testAll ;dotty-bench-bootstrapped/jmh:run 1 1 tests/pos/alias.scala"
```
allanrenucci: Maybe add `;dotty-bootstrapped/compile` here too

Duhemm: Won't `dotty-bootstrapped/testAll` do the compilation step already?

allanrenucci: @nicolasstucki You said at some point that `testAll` didn't build DottyDoc, right?

nicolasstucki: It was not. Or at least it was missing some part of it.

Issue #2466, opened 2017-05-18
Broken link on getting-started page


Varunram:
While visiting the getting started page <a href="http://dotty.epfl.ch/docs/contributing/getting-started.html"> here,</a> we have a broken link to CONTRIBUTING.html(should be CONTRIBUTING.md) at the end of the page . My guess is this is something to do with parsing md to html, so would renaming CONTRIBUTING.md to simply CONTRIBUTING work?

Varunram:
CC: @felixmulder

PR #3426, opened 2017-11-01
Fix rendering of the CONTRIBUTING page on Github

maseev:
This PR is related to the following issue - #2466.

This PR adds the necessary file extension as well as fixes the link in the appropriate file so that users will be able to see the correctly rendered document written in the Markdown language. 

It also slightly modifies the `MarkdownLinkVisitor.scala` so that it filters out the 'external' links to `.md` files (so they won't be renamed in the generated `.html` documents).


maseev:
Hm, drone again failed the build on the clone phase :disappointed:

Issue #2989, opened 2017-08-16
HK type parameter`Y` not considered a subtype of `[T] => Y[T]` in some situations


smarter:
```scala
class Foo[+X[_]] {
  // OK
  def foo1[Y[_]](right: Foo[Y]): Foo[Y] = right
  // OK
  def foo2[Y[_]](right: Foo[[T] => Y[T]]): Foo[Y] = right
  // OK
  def foo3[Y[_]](right: Foo[[T] => Y[T]]): Foo[[T] => Y[T]] = right
  // Error:
  //   found:    Foo[Y](right)
  //   required: Foo[Y]
  def foo4[Y[_]](right: Foo[Y]): Foo[[T] => Y[T]] = right
}
```

PR #3428, opened 2017-11-01
Fix #3422: Add missing case to TypeComparer

### LINKED ISSUE #3422: Wrong type mismatch of HKT under separate compilation
*a.scala*:

```scala
trait Fun[L[_]]

object O1 {
  trait N[X]
}

object O2 {
  def bar: Fun[O1.N] = ???
}
```

*b.scala*:
```scala
object Test {
  def c: Fun[O1.N] = O2.bar
}
```

Breaks with separate compilation:

```scala
$ rm *.class; dotc a.scala; dotc b.scala
-- [E007] Type Mismatch Error: b.scala:2:24 ------------------------------------
2 |  val c: Fun[O1.N] = O2.bar
  |                     ^^^^^^
  |                     found:    Fun[O1.N]
  |                     required: Fun[O1.N]
  |
```

odersky:
We missed the case where we compare two hk tpes A <: B, where both A and B refer to
classes with type parameters. It seems this case usually does not arise when we compile
from source because A or B or both are eta expanded. But it can arise after unpickling.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -369,14 +369,18 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
     case _ =>
       val cls2 = tp2.symbol
       if (cls2.isClass) {
-        val base = tp1.baseType(cls2)
-        if (base.exists) {
-          if (cls2.is(JavaDefined))
-            // If `cls2` is parameterized, we are seeing a raw type, so we need to compare only the symbol
-            return base.typeSymbol == cls2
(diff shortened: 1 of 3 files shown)


odersky:
@smarter Ready to merge?

smarter:
I'll review it today

smarter:
LGTM, although I wonder if we should take a step back and try to fix our higher-kinded subtyping checks in a more systematic way. https://github.com/lampepfl/dotty/issues/2989 is a very similar issue which is still broken after this PR. I have a tentative fix at https://github.com/smarter/dotty/commit/245bed9b915c2b0fd0030e9591932d33a97a3461 but I'm not convinced it's the right thing to do.

PR #3429, opened 2017-11-01
Fix #3332: Disallow bound variables in pattern alternatives

### LINKED ISSUE #3332: Confusion between syntax for pattern alternative and typed pattern on a union type
As in the title, running this code in REPL throws `java.lang.VerifyError`:

```
scala> def f(): String | Int = 0
def f(): String | Int
scala> f() match { case s: String | Int => s }
```

```
java.lang.VerifyError: Bad local variable type
Exception Details:
  Location:
    rs$line$2$.<init>()V @33: aload_3
  Reason:
    Type top (current frame, locals[3]) is not assignable to reference type
  Current Frame:
    bci: @33
    flags: { }
    locals: { 'rs$line$2$', 'java/lang/Object' }
    stack: { 'rs$line$2$' }
  Bytecode:
    0x0000000: 2ab7 0014 2ab3 0016 2ab2 001b b600 1f4c
    0x0000010: 2bc1 0021 9900 172b c000 214d 2c4e a700
    0x0000020: 072d a700 06a7 fffc a700 1cb2 0026 2bb6
    0x0000030: 002a 9900 09a7 ffec 0000 bfbb 002c 592b
    0x0000040: b700 2fbf b500 31b1
  Stackmap Table:
    full_frame(@33,{Object[#2],Object[#4]},{Object[#2]})
    full_frame(@37,{Object[#2],Object[#4],Object[#33],Object[#33]},{Object[#2]})
    full_frame(@40,{Object[#2],Object[#4]},{Object[#2],Top})
    same_locals_1_stack_item_frame(@43,Object[#2])
    full_frame(@56,{},{Object[#53]})
    full_frame(@59,{Object[#2],Object[#4]},{Object[#2]})
    full_frame(@68,{Object[#2],Object[#4]},{Object[#2],Top})

	at rs$line$2.res0Show(rs$line$2)
	at dotty.tools.repl.Rendering.valueOf(Rendering.scala:58)
	at dotty.tools.repl.Rendering.renderVal(Rendering.scala:79)
	at dotty.tools.repl.ReplDriver.$anonfun$displayDefinitions$8(ReplDriver.scala:269)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:52)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike.map(TraversableLike.scala:234)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at dotty.tools.repl.ReplDriver.displayMembers$1(ReplDriver.scala:269)
	at dotty.tools.repl.ReplDriver.$anonfun$displayDefinitions$15(ReplDriver.scala:296)
	at scala.Option.map(Option.scala:146)
	at dotty.tools.repl.ReplDriver.$anonfun$displayDefinitions$13(ReplDriver.scala:294)
	at dotty.tools.dotc.core.Periods.atPhase(Periods.scala:26)
	at dotty.tools.dotc.core.Phases.atPhase(Phases.scala:36)
	at dotty.tools.dotc.core.Phases.atPhase$(Phases.scala:35)
	at dotty.tools.dotc.core.Contexts$Context.atPhase(Contexts.scala:57)
	at dotty.tools.repl.ReplDriver.displayDefinitions(ReplDriver.scala:289)
	at dotty.tools.repl.ReplDriver.$anonfun$compile$2(ReplDriver.scala:226)
	at scala.util.Either.fold(Either.scala:188)
	at dotty.tools.repl.ReplDriver.compile(ReplDriver.scala:220)
	at dotty.tools.repl.ReplDriver.interpret(ReplDriver.scala:190)
	at dotty.tools.repl.ReplDriver.runUntilQuit(ReplDriver.scala:143)
	at dotty.tools.repl.Main$.main(Main.scala:6)
	at dotty.tools.repl.Main.main(Main.scala)
```

Parenthesising the type, however, runs normally:
```
scala> f() match { case s: (String | Int) => s }
val res2: Any = 0
```

odersky:
Disallow bound variables where the binder appears as a pattern alternative.

Fixes #3332.

There's nothing we can do about the parsing. If we parsed

    case x: A | B 

as

    case x: (A | B)

then we would run into an error for

    case _: A | _: B

But the latter is a common idiom.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Mode.scala
+++ b/compiler/src/dotty/tools/dotc/core/Mode.scala
@@ -8,7 +8,7 @@ case class Mode(val bits: Int) extends AnyVal {
   def &~ (that: Mode) = Mode(bits & ~that.bits)
   def is (that: Mode) = (bits & that.bits) == that.bits
 
-  def isExpr = (this & PatternOrType) == None
+  def isExpr = (this & PatternOrTypeBits) == None
 
   override def toString =
     (0 until 31).filter(i => (bits & (1 << i)) != 0).map(modeName).mkString("Mode(", ",", ")")
@@ -41,6 +41,9 @@ object Mode {
 
   /** We are looking at the arguments of a supercall */
   val InSuperCall = newMode(6, "InSuperCall")
+
+  /** We are in a pattern alternative */
+  val InPatternAlternative = newMode(7, "InPatternAlternative")
 
   /** Allow GADTFlexType labelled types to have their bounds adjusted */
   val GADTflexible = newMode(8, "GADTflexible")
@@ -87,7 +90,7 @@ object Mode {
   /** Don't suppress exceptions thrown during show */
   val PrintShowExceptions = newMode(18, "PrintShowExceptions")
 
-  val PatternOrType = Pattern | Type
+  val PatternOrTypeBits = Pattern | Type | InPatternAlternative
 
   /** We are elaborating the fully qualified name of a package clause.
    *  In this case, identifiers should never be imported.
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -936,10 +936,10 @@ class Namer { typer: Typer =>
   }
 
   def typedAheadType(tree: Tree, pt: Type = WildcardType)(implicit ctx: Context): tpd.Tree =
-    typedAheadImpl(tree, typer.typed(_, pt)(ctx retractMode Mode.PatternOrType addMode Mode.Type))
+    typedAheadImpl(tree, typer.typed(_, pt)(ctx retractMode Mode.PatternOrTypeBits addMode Mode.Type))
 
   def typedAheadExpr(tree: Tree, pt: Type = WildcardType)(implicit ctx: Context): tpd.Tree =
-    typedAheadImpl(tree, typer.typed(_, pt)(ctx retractMode Mode.PatternOrType))
+    typedAheadImpl(tree, typer.typed(_, pt)(ctx retractMode Mode.PatternOrTypeBits))
 
   def typedAheadAnnotation(tree: Tree)(implicit ctx: Context): Symbol = tree match {
     case Apply(fn, _) => typedAheadAnnotation(fn)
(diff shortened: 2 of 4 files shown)


smarter:
> There's nothing we can do about the parsing.

Maybe when we report this error we could add "did you mean `case x : (A | B)` ?" if `A | B` is a valid type

smarter:
Or maybe we need more backtracking in the parser :)

smarter:
> we would run into an error for
>     case _: A | _: B
> But the latter is a common idiom.

One compromise could be: do backtracking under Scala2 mode but warn people when we backtrack that they should write `(_: A) | (_: B)` (for Scala 2 compatiblity) or `_: A | B` (which is more natural but incompatible).

odersky:
@smarter It's actually worse than that. Here's some code that is legal now, but would mean something different if we changed the parsing scheme:

    case class A
    class B

    case _: B | A

smarter:
Actually, `case class A` isn't valid, you have to write `case object A` instead, and in that case `case _: (B | A)` wouldn't compile, it would have to be `case _: (B | A.type)`

odersky:
Here's a more complete transscript of what goes wrong:

scala> case class A()
defined class A

scala> class B
defined class B

scala> (A(): Any) match { case _: B | A => "ok" case _ => "no" }
res1: String = no

compiler/src/dotty/tools/dotc/core/Mode.scala
```diff
@@ -87,7 +90,7 @@ object Mode {
   /** Don't suppress exceptions thrown during show */
   val PrintShowExceptions = newMode(18, "PrintShowExceptions")
 
-  val PatternOrType = Pattern | Type
+  val PatternOrTypeBits = Pattern | Type | InPatternAlternative
```
allanrenucci: Why did you rename this value? I think the old name is more consistent with the other mode's name

odersky: It's because otherwise we'd risk that someone thinks it's `Pattern | Type` only. And `PatternOrTypeorInPatternAlternative` was too long.

odersky:
@allanrenucci The failure should be fixed now.

smarter:
> scala> case class A()
> defined class A
> scala> class B
> defined class B
> scala> (A(): Any) match { case _: B | A => "ok" case _ => "no" }
> res1: String = no

This is a bit off-topic at this point, but I would argue that we're not doing anyone any favour by accepting this without at least a warning: it's extremely rare to want to pattern match on the companion object of a case class and is most likely to happen when trying to pattern match on the case class itself. I would rewrite `A` as `_: A.type` to make the meaning obvious.

PR #3436, opened 2017-11-03
Fix dotr -tasty

nicolasstucki:


project/Build.scala
```diff
@@ -504,15 +504,17 @@ object Build {
...
-        val _  = packageAll.value
+        val _ = packageAll.value
         val scalaLib = attList
           .map(_.data.getAbsolutePath)
           .find(_.contains("scala-library"))
           .toList.mkString(":")
 
-        if (java == "")
+        if (args.isEmpty) {
+          println("Couldn't run `dotr` without args. Use `repl` to run the repl or add args to run the dotty application")
```
smarter: Why not call the repl task at this point instead?

nicolasstucki: I tried with `repl.evaluated` but then `dotr` only ran `repl` ignoring the nonempty args. Then I just decided to show this message as we don't really need more in the sbt project.

smarter: sbt task evaluation is tricky, you probably need to use a dynamic task to get that to work.

nicolasstucki: That is what I thought, hence settled for the message. My main concerns were the factorisation of code and checking the it was executing correctly.

project/Build.scala
```diff
@@ -640,6 +617,19 @@ object Build {
       }
   )
 
+  def dotDynTask(main: String) = Def.inputTaskDyn {
```
allanrenucci: I think `runCompilerMain` is a better name

project/Build.scala
```diff
@@ -504,49 +504,26 @@ object Build {
...
+        if (args.isEmpty) {
+          println("Couldn't run `dotr` without args. Use `repl` to run the repl or add args to run the dotty application")
+        } else if (java == "") {
           println("Couldn't find java executable on path, please install java to a default location")
-        else if (scalaLib == "") {
+        } else if (scalaLib == "") {
           println("Couldn't find scala-library on classpath, please run using script in bin dir instead")
         } else {
           val dottyLib = packageAll.value("dotty-library")
           s"""$java -classpath .:$dottyLib:$scalaLib ${args.mkString(" ")}""".!
```
allanrenucci: Since you're refactoring. Triple quotes is not needed here

.drone.yml
```diff
@@ -25,7 +25,7 @@ pipeline:
     image: lampepfl/dotty:2017-10-20
     commands:
       - cp -R . /tmp/1/ && cd /tmp/1/
-      - ./project/scripts/sbt ";compile ;testAll ;dotty-bench/jmh:run 1 1 tests/pos/alias.scala"
+      - ./project/scripts/sbt ";compile ;testAll ;dotty-bench/jmh:run 1 1 tests/pos/alias.scala; ;dotc tests/run/arraycopy.scala ;dotr Test"
```
allanrenucci: `tests/pos/alias.scala; ;dotc` double semicolon?

allanrenucci: Also not sure it is the best place to test an sbt task. @smarter Do you think it should be a scripted test?

project/Build.scala
allanrenucci: I don't think we should special case for empty `args`. We need a better solution:
If I run `dotr -Xprint:typer`, `args` is not empty, but I still want to run the REPL

.drone.yml
smarter: Scripted tests are always good in my opinion :)

project/Build.scala
nicolasstucki: you can at least run `repl -Xprint:typer`.

nicolasstucki:
Next step will be to remove the `dotty.tools.dot.FromTasty` driver and integrate it to `dotty.tools.dot.Main`. This will make `-tasty` an actual `ScalaSetings` flag.

dist/bin/dotc
```diff
@@ -82,7 +81,7 @@ case "$1" in
     # Optimize for short-running applications, see https://github.com/lampepfl/dotty/issues/222
     -Oshort) addJava "-XX:+TieredCompilation -XX:TieredStopAtLevel=1" && shift ;;
         -repl) PROG_NAME="$ReplMain" && shift ;;
-       -tasty) addScala "-Yretain-trees" && PROG_NAME="$FromTasty" && shift ;;
+       -tasty) addScala "-Yretain-trees" && PROG_NAME="$CompilerMain" && shift ;;
```
smarter: I would just make `-Yretain-trees` default to true when `-tasty` is used.

smarter: I mean, in the compiler itself, instead of doing it in various scripts.

nicolasstucki:
Last commit needs review

project/scripts/sbtBootstrappedTests
```diff
@@ -2,3 +2,34 @@
...
 ./project/scripts/sbt "dotty-bench-bootstrapped/jmh:run 1 1 tests/pos/alias.scala"
+
+
+# setup for `dotc`/`dotr` script tests
+./project/scripts/sbt dist-bootstrapped/pack
+
+# check that `dotc` compiles and `dotr` runs it
+echo "testing sbt dotc and dotr"
+mkdir out/scriptedtest0
+./dist-bootstrapped/target/pack/bin/dotc tests/pos/sbtDotrTest.scala -d out/scriptedtest0
```
OlivierBlanvillain: Any reason to use `dist-bootstrapped/**/dotc` instead of `bin/dotc`?

nicolasstucki: I wanted to make sure I was testing that version of the script. We could also add some tests for `bin/dotc` in the future.

OlivierBlanvillain: `bin/dotc` just forwards to `dist-bootstrapped/target/pack/bin/dotc` and call `sbt dist-bootstrapped/pack` if needed :)

nicolasstucki: Ok, I'll change it.

PR #3437, opened 2017-11-03
new format for "package name already defined" error (+tests)

AurelienRichez:
PR from Scala.io 2017 FLOSS spree in Lyon.

For #1589 :

This PR creates a new format of error messages for `Typer.scala:1519`. The new `Message` class is `PackageNameAlreadyDefined`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1922,4 +1922,12 @@ object messages {
     val msg = hl"Traits cannot redefine final $method from ${"class AnyRef"}."
     val explanation = ""
   }
+
+  case class PackageNameAlreadyDefined(pkg: Symbol)(implicit ctx: Context) extends Message(PackageNameAlreadyDefinedID) {
+
+    override def msg: String = hl"${pkg} is already defined, cannot be a package"
+    override def kind: String = "Syntax"
+    override def explanation: String =
+      "An object cannot have the same name as an existing package. Rename ${pkg} or the package with the same name."
+  }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1924,9 +1924,8 @@ object messages {
   }
 
   case class PackageNameAlreadyDefined(pkg: Symbol)(implicit ctx: Context) extends Message(PackageNameAlreadyDefinedID) {
-
     val msg = hl"${pkg} is already defined, cannot be a ${"package"}"
-    val kind = "Syntax"
+    val kind = "Naming"
     val explanation =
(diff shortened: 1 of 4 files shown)


allanrenucci:
@smarter scalac does not emit an error here. Is the difference intended?

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -1151,4 +1151,18 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+      """
+        |package bar { }
+        |object bar { }
+        |
+      """.stripMargin
+    }.expect { (ictx, messages) =>
+      implicit val ctx: Context = ictx
+
+      assertMessageCount(1, messages)
+      assert(messages.head.isInstanceOf[PackageNameAlreadyDefined])
```
allanrenucci: Prefer:
```scala
val PackageNameAlreadyDefined(pkg) = messages.head
assertEquals(pkg.show, "object bar")
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,12 @@ object messages {
     }
     val explanation = "A statement is either an import, a definition or an expression."
   }
+
+  case class PackageNameAlreadyDefined(pkg: Symbol)(implicit ctx: Context) extends Message(PackageNameAlreadyDefinedID) {
+
+    override def msg: String = hl"${pkg} is already defined, cannot be a package"
+    override def kind: String = "Syntax"
+    override def explanation: String =
```
allanrenucci: The `override` keyword as well as the type ascription are a bit verbose. I would remove them

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,12 @@ object messages {
...
+      "An object cannot have the same name as an existing package. Rename ${pkg} or the package with the same name."
```
allanrenucci: You can use `hl` interpolator for syntax highlighting:
```scala
hl"An ${"object"} cannot have the same name as an existing ${"package"}. Rename either one of them."
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,12 @@ object messages {
     }
     val explanation = "A statement is either an import, a definition or an expression."
   }
+
+  case class PackageNameAlreadyDefined(pkg: Symbol)(implicit ctx: Context) extends Message(PackageNameAlreadyDefinedID) {
+
+    override def msg: String = hl"${pkg} is already defined, cannot be a package"
```
allanrenucci: You can syntax highlight `package` in this sentence:
```scala
hl"${pkg} is already defined, cannot be a ${"package"}"
```

AurelienRichez:
Thanks for your review ! I rebased my branch and applied the modifications.

PR #3438, opened 2017-11-04
Fix #1519: Port compiler Plugin from scalac

### LINKED ISSUE #1519: Decide upon Compiler Plugin architecture

liufengyun:
Fix #1519:  This PR ports compiler plugin from Scalac and adds the test infrastructure.

Scalac also supports Typer plugins, which is contentious, so it's not addressed in this PR. Current  design allows easy extension to support Typer plugins if we want to do that later.

Changes from Scalac plugin:

- Removed the concept `PIuginComponent`
- Research plugin: research plugins decides where to inject their own phases

TODOs:

- [x] basic plugin works
- [x] test infrastructure
- [x] ordering constraints of plugins
- [x] unit test ordering algorithm
- [x] SBT plugin test


tests/plugins/neg/divideZero/plugin_1.scala
```diff
@@ -0,0 +1,37 @@
...
+  override def init(phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
+    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
+    before ++ (List(this) :: after)
+  }
+
+  private def isNumericDivide(sym: Symbol)(implicit ctx: Context): Boolean = {
+    def test(tpe: String): Boolean =
+      (sym.owner eq ctx.requiredClass(tpe.toTermName)) && sym.name.show == "/"
+
+    test("scala.Int") || test("scala.Long") || test("scala.Short") || test("scala.FLoat") || test("scala.Double")
```
smarter: typo: FLoat -> Float

tests/plugins/neg/divideZero/plugin_1.scala
```diff
@@ -0,0 +1,37 @@
...
+  val phaseName = name
+
+  override def init(phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
+    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
+    before ++ (List(this) :: after)
+  }
+
+  private def isNumericDivide(sym: Symbol)(implicit ctx: Context): Boolean = {
+    def test(tpe: String): Boolean =
+      (sym.owner eq ctx.requiredClass(tpe.toTermName)) && sym.name.show == "/"
```
smarter: You should be able to do `sym.name == nme.DIV`

tests/plugins/neg/divideZero/scalac-plugin.xml
```diff
@@ -0,0 +1,4 @@
+<plugin>
```
smarter: We should check that sbt's support for compiler plugin still works and adapt it in sbt-dotty if it doesn't: http://www.scala-sbt.org/1.x/docs/Compiler-Plugins.html We can do that by adding scripted tests in https://github.com/lampepfl/dotty/tree/master/sbt-dotty/sbt-test/sbt-dotty

tests/plugins/neg/divideZero/plugin_1.scala
```diff
@@ -0,0 +1,37 @@
...
+import Constants.Constant
+
+class DivideZero extends MiniPhase with Plugin {
+  val name: String = "divideZero"
+  override val description: String = "divide zero check"
+
+  val phaseName = name
+
+  override def init(phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
+    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
```
smarter: It would be better to not rely on strings to identify phase names and use class types instead, similar to what we do with `Phase.replace` in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/repl/ReplCompiler.scala#L40

sjrd:
The interface with `init` is both too powerful and not enough:
```scala
  override def init(phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
    before ++ (List(this) :: after)
  }
```
First, it is too powerful, because it allows to reorder core phases, or phases of other plugins.

Also, it allows to insert micro-phases in existing macro-phases, which is extremely sensitive. I do not believe anyone outside of the core compiler team can write a correct micro-phase that would interact badly with the rest of its macro phase. IMO compiler plugins should only be able to insert macro phases (themselves possibly consisting of multiple micro-phases, all provided by the same plugin).

Then, it is not powerful enough, because it does not allow to correctly express dependencies between compiler plugins. With the scalac compiler plugin interface, I can write a compiler plugin B that knows about another compiler plugin A, and correctly declares its dependencies wrt. that one. With this interface, this is not possible. Let me illustrate this with a very concrete (and existing!) example:

The [Scala.js JUnit compiler plugin](https://github.com/scala-js/scala-js/tree/v1.0.0-M1/junit-plugin/src/main) has a phase `junit-inject` which needs to run between `mixin` and `jscode` (the latter being the code generation phase of the Scala.js compiler plugin. In scalac, this is easy to specify declaratively in `ScalaJSJUnitPlugin`. The core `ScalaJSPlugin` doesn't need to know about the JUnit compiler plugin. With the proposed interface for dotty, this dependency cannot be enforced without the core Scala.js plugin know about the JUnit compiler plugin. Indeed, even if the JUnit compiler plugin carefully manipulates the `phases` to insert itself between `mixin` (or whatever other phase is accurate for dotty in this case) and `jscode`, this only works if the `init` method of the JUnit plugin is called *after* that of the core plugin. Indeed, in the other direction, the JUnit plugin can only insert itself between `mixin` and `delambdafy` (continuing my earlier scalac-based example). Then when the core plugin's `init` method kicks in, since it does not know about `junit-inject`, it will insert itself either before `junit-inject` or after it. In the former case, things get broken.

This means that obtaining correct inter-plugin dependencies depends on the *order in which the `init` methods are called*. This is **really bad**, as this order is likely to indirectly derive from ivy resolution order.

For these reasons, I disapprove of this new interface. I strongly suggest an interface that is closer to scalac, specifying a list of `afterPhases` and a list of `beforePhases` (probably dropping `rightAfterPhase` because that one does not compose). And I would only allow plugins to insert macro phases. The design could provide for a possible future extension point that would allow expert compiler plugin authors to inject micro phases.

tests/plugins/neg/divideZero/plugin_1.scala
sjrd: Strings compose with phases coming from other compiler plugins. Class types do not.

liufengyun:
@sjrd Thanks a lot for sharing your experience, it's a valuable feedback 👍 

We are in agreement that the new design is more powerful. As compiler plugins are for very advanced users & researchers, we may assume they use the power wisely. I hope you can also agree that the protocol/API is simpler, and the implementation is easy. In general, the usability for plugin writers is better.

For the particular problem you raised, is it possible for the `junit-inject` plugin to act as a facade and also add `ScalaJS` phases to the phase plan in the `init` of `junit-inject`? 

The problem is in general about composability of plugins, which we know is very tricky: _compiler plugins just don't compose easily_. The Scalac approach solves some of the problems by allowing plugins to specify _ordering constraints_.

But the solution in Scalac is still unsatisfactory. _How can a plugin knows in advance its relative order to all other plugins_, given that 

- it's impractical for a plugin developer to test the ordering wrt. all existing plugins
- new plugins continue to appear
- logically, if `PlugA < PlugB`, should the ordering constraint be in PlugA or PlugB, or both?

The new approach instead shifts all the responsibility to programmers. This seems to be a more coherent approach to me. In practice, a few plugins will be regularly used together to solve a particular problem. Due to the frequency of the combined usage, frictions among the plugins will be detected and reduced, compatibility will improve and the relative ordering will be known. Then, it's  simple to just create an orchestrate/facade plugin to compose the phase plan _if the ordering is sensitive_.

In the new approach, creating a facade compiler plugin is easy, just a few lines of code will do. As compiler plugins usually work with sbt,  the compiler plugin facade can be authored in SBT(`Build.scala`) to orchestrate the phase plan, or come from some pre-defined SBT plugin. We do need to investigate to see if this approach works well with SBT and user experience is not impacted.

From the viewpoint of design, if practical composability problems can solved with the compiler taking ZERO responsibility, then it doesn't make sense for the compiler to take just PARTIAL responsibility.  I assume It's always better for the boundary of responsibilities to be clear.

tests/plugins/neg/divideZero/plugin_1.scala
odersky: @sjrd Good point. But can't we use `Class.forName` in that case?

sjrd: If all compiler plugins are loaded in the same `ClassLoader`, then yes. But is that the case and is it desirable?

compiler/test/dotty/tools/dotc/plugins/PluginsTest.scala
```diff
@@ -87,7 +89,7 @@ class PluginsTest {
   def orderingTwoPlugins1 = {
     object M1 extends TestPhase {
       override val runsAfter = Set(classOf[P3d])
-      override val runsBefore = Set(M2.getClass, classOf[P7], classOf[P8])
```
smarter: Opened https://github.com/lampepfl/dotty/issues/3495 to keep track of this.

smarter:
Something to keep in mind: the dynamic classloading of compiler plugins seems to have a significant performance cost, which could perhaps be avoided by better caching of classloaders, see https://github.com/scala/scala-dev/issues/458

OlivierBlanvillain:
@liufengyun what's is the status of this PR? If it needs a review please assign it to a reviewer :)

liufengyun:
@OlivierBlanvillain This is up for review. I think @sjrd has the best experience on plugins, but it's good to get more reviews.

smarter:
> @nicolasstucki nicolasstucki assigned smarter and unassigned nicolasstucki 7 days ago

Unfortunately I won't have time to review this myself in the near future.

compiler/src/dotty/tools/dotc/plugins/Plugins.scala
```diff
@@ -0,0 +1,268 @@
...
+      orderRequirements.update(phase.getClass, (runsAfter, runsBefore))
+    }
+
+    // propagate ordering constraint : reflexivity
+    pluginPhases.foreach { phase =>
+
+      var (runsAfter, runsBefore) = orderRequirements(phase.getClass)
+
+      // propagate transitive constraints to related phases
+      runsAfter.filter(!isPrimitive(_)).foreach { phaseClass =>
```
sjrd: why do you exclude primitive phases here?

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+import dotty.tools.io._
+import transform.MegaPhase.MiniPhase
+
+import java.io.InputStream
+
+import scala.collection.mutable
+import scala.util.{ Try, Success, Failure }
+
+trait PluginPhase extends MiniPhase {
+  def runsBefore: Set[Class[_ <: Phase]] = Set.empty
```
sjrd: For inter-plugin dependencies, it will be impossible for a plugin A to get hold of the `Class[_]` value for a plugin B coming from a different jar (hence in a different `ClassLoader`, so not even reflectively available).

We need to use *names* here (also for `runsAfter`). Names are usually not as good as token identities, but here they offer an essential part of the expressiveness.

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/plugin/DivideZero.scala
```diff
@@ -0,0 +1,40 @@
...
+  override val runsAfter = Set(classOf[Pickler])
+  override val runsBefore = Set(classOf[LinkAll])
+
+  override def init(options: List[String]): List[PluginPhase] = this :: Nil
+
+  private def isNumericDivide(sym: Symbol)(implicit ctx: Context): Boolean = {
+    def test(tpe: String): Boolean =
+      (sym.owner eq ctx.requiredClass(tpe.toTermName)) && sym.name.show == "/"
+
+    test("scala.Int") || test("scala.Long") || test("scala.Short") || test("scala.FLoat") || test("scala.Double")
```
sjrd: Typo: `FLoat` -> `Float`

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=0.13.15
```
sjrd: 0.13.17?

tests/plugins/neg/divideZero-research/plugin_1.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def init(options: List[String], phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
+    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
+    before ++ (List(this) :: after)
+  }
+
+  private def isNumericDivide(sym: Symbol)(implicit ctx: Context): Boolean = {
+    def test(tpe: String): Boolean =
+      (sym.owner eq ctx.requiredClass(tpe.toTermName)) && sym.name.show == "/"
+
+    test("scala.Int") || test("scala.Long") || test("scala.Short") || test("scala.FLoat") || test("scala.Double")
```
sjrd: Same typo on `FLoat`

tests/plugins/pos/divideZero/plugin_1.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def init(options: List[String], phases: List[List[Phase]])(implicit ctx: Context): List[List[Phase]] = {
+    val (before, after) = phases.span(ps => !ps.exists(_.phaseName == "pickler"))
+    before ++ (List(this) :: after)
+  }
+
+  private def isNumericDivide(sym: Symbol)(implicit ctx: Context): Boolean = {
+    def test(tpe: String): Boolean =
+      (sym.owner eq ctx.requiredClass(tpe.toTermName)) && sym.name.show == "/"
+
+    test("scala.Int") || test("scala.Long") || test("scala.Short") || test("scala.FLoat") || test("scala.Double")
```
sjrd: And again the same typo.

liufengyun:
Thanks a lot for the review @sjrd , I'll address them this weekend.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+
+  /** Use a class loader to load the plugin class.
+   */
+  def load(classname: String, loader: ClassLoader): Try[AnyClass] = {
+    import scala.util.control.NonFatal
+    try {
+      Success[AnyClass](loader loadClass classname)
+    } catch {
+      case NonFatal(e) =>
+        Failure(new PluginLoadException(classname, s"Error: unable to load class: $classname"))
```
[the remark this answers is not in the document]
som-snytt: Oddly, I recall inheriting that when I touched the code. Generally `e.getMessage == classname` for `ClassNotFoundException`, but this code could be more nuanced in avoiding repeating the name, especially for other throwables or if there's a cause. I must have added the next line in this vein. You're more likely to hit `NoClassDef` with multiple class loaders?

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+    def read(is: InputStream) =
+      if (is == null) throw new PluginLoadException(jarp.path, s"Missing $PluginXML in $jarp")
+      else PluginDescription.fromXML(is)
+
+    val xmlEntry = new java.util.jar.JarEntry(PluginXML)
+    Try(read(new Jar(jarp.jpath.toFile).getEntryStream(xmlEntry)))
+  }
+
+  private def loadDescriptionFromFile(f: Path): Try[PluginDescription] =
+    Try(PluginDescription.fromXML(new java.io.FileInputStream(f.jpath.toFile)))
```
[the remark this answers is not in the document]
som-snytt: It's not API, but it would be more consistent. Both methods could be local.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+          if (p.isDirectory) loadDescriptionFromFile(p.toDirectory / PluginXML) orElse loop(rest)
+          else if (p.isFile) loadDescriptionFromJar(p.toFile) orElse loop(rest)
+          else loop(rest)
+      }
+      loop(ps)
+    }
+    val fromPaths: PDResults = paths map (p => (p, findDescriptor(p))) map {
+      case (p, Success(pd)) => Success((pd, loaderFor(p)))
+      case (_, Failure(e))  => Failure(e)
+    }
```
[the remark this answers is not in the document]
som-snytt: Kind of puts the fusing in confusing. Maybe use placeholder notation for the `pd`.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/** ...
+ *
+ *  @author Lex Spoon
+ *  @version 1.0, 2007-5-21
+ */
+object Plugin {
+
+  private val PluginXML = "scalac-plugin.xml"
```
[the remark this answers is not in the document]
som-snytt: IIRC, the "name" in that xml is unused. Maybe it's possible to ditch the xml. See Adriaan's work to avoid dependency on scala-xml at some point. It could be a properties file, or just a file containing a class name, radical simplicity. More like java's service loader. Seems a shame to drag xml config into scala 3.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
```diff
@@ -0,0 +1,176 @@
...
+        // a nod to scala/bug#7494, take the plugin classes distinctly
+        Failure(new PluginLoadException(pd.name, s"Ignoring duplicate plugin ${pd.name} (${pd.classname})"))
+      case Success((pd, loader)) if ignoring contains pd.name =>
+        Failure(new PluginLoadException(pd.name, s"Disabling plugin ${pd.name}"))
+      case Success((pd, loader)) =>
+        seen += pd.classname
+        Plugin.load(pd.classname, loader)
+      case Failure(e)            =>
+        Failure(e)
+    }
```
som-snytt: `Try.flatMap`.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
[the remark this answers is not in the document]
liufengyun: Indeed, only the class name is actually used. Maybe we can assume a text file named `plugins` with each line corresponding to a class name.

compiler/src/dotty/tools/dotc/plugins/Plugins.scala
liufengyun: Upon reflection, I think we miss a test case, primitive phase should not be a special case.

compiler/src/dotty/tools/dotc/plugins/Plugin.scala
liufengyun: The original use of `map` is a good fit here. We are changing `Success` to `Failure` in the first two cases, `Try.flatMap` will not help here.

Blaisorblade: [Try.flatMap](https://www.scala-lang.org/api/current/scala/util/Try.html#flatMap[U](f:T=>scala.util.Try[U]):scala.util.Try[U]) allows changing `Success` to `Failure`:
```scala
(fromPaths ::: fromDirs) flatMap {
  case (pd, loader) if seen(pd.classname) => Failure(new PluginLoadException(pd.name, s"Ignoring duplicate plugin ${pd.name} (${pd.classname})"))
  case (pd, loader) if ignoring contains pd.name =>
    Failure(new PluginLoadException(pd.name, s"Disabling plugin ${pd.name}"))
  case (pd, loader) =>
    seen += pd.classname
    Plugin.load(pd.classname, loader)
}
```

liufengyun: Thanks @Blaisorblade , sure, that's the whole purpose of flatMap/bind, a temporary short-circuit in my brain ;-)

som-snytt: Now to backport to scala 2.

smarter:
Since this PR ports existing code, it would be good to add a note to https://github.com/lampepfl/dotty/blob/master/AUTHORS.md for proper attribution.

compiler/src/dotty/tools/dotc/plugins/Plugins.scala
liufengyun: @sjrd : thanks again for the review, I've addressed all raised issues. For this particular one, it turns out the code can be generalized to use the same logic for all already inserted phases. I've updated the code and added a detailed comment:

https://github.com/lampepfl/dotty/pull/3438/files#diff-3226d7606c232d5a280358cb50f2d961R149

som-snytt:
I would have smiley-faced sjrd's approval comment if github provided that function.

PR #3439, opened 2017-11-06
Replace isAsGood with ternary operator

odersky:
The aim is to have ultimately fewer comparisons. The downside is
that some of the comparisons are now more expensive, because types
have to be compared where they previously weren't.

This should be integrated in the implicits PR. It is broken out here so that performance can
be tested separately.


odersky:
test performance please

odersky:
This was integrated in #3421

PR #3445, opened 2017-11-08
Remove vulpix reflective nightmare

nicolasstucki:


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -33,10 +33,12 @@ class CompilationTests extends ParallelTesting {
 
   // @Test  // enable to test compileStdLib separately with detailed stats
   def compileStdLib: Unit = {
+    implicit val testGroup: TestGroup = TestGroup("compileStdLib")
```
smarter: You shouldn't need to give an explicit type signature here because the implicit val is in a def.

allanrenucci:
Can you explain what a test group is and what it is used for

compiler/test/dotty/tools/dotc/CompilationTests.scala
nicolasstucki: Do you want me to remove them? I added them to avoid warnings from the IDE.

smarter: Your call.

nicolasstucki:
It is used as the name of the first level of folder nesting in the output of the tests. For example all output of `@Test def runAll = ...` will be `out/runAll`. Inner folders are determined by the kind of tests you are executing. Previously the group name was generated by reflection inspecting the name of the method it was in, this was unnecessary, brittle and hard to understand in the cases where compilation tests depended on previous ones such as `@Test def tastyBootstrap`.  In that case the inner `def`s where only present to give the group name to substeps of the tests and refactoring easily broke the expected output paths.

allanrenucci:
I think it’s weird that outputDirecrory is not the actual output directory. It is also weird to add a new abstraction just to specify the output directory. Also we now have test group and test category that are two completely different things
Can we just use the existing outputdirectory parameter?

nicolasstucki:
The test group was also used to determine the title of the tests. I wonder if we could instead remove the outputdirectory parameter.

nicolasstucki:
Yes, the output directory was only used twice and was API overkill.

compiler/test/dotty/tools/vulpix/TestGroup.scala
```diff
@@ -0,0 +1,3 @@
+package dotty.tools.vulpix
+
+case class TestGroup(name: String) extends AnyVal
```
smarter: Please document what a TestGroup is.

nicolasstucki: Done

PR #3446, opened 2017-11-08
Fix #3443: adapt child compatibility check

### LINKED ISSUE #3443: Incorrect MatchCaseUnreachable warning on Coproduct
Here is an example where Dotty incorrectly reports an unreachable case in pattern marching:

```scala
object Test {
  // shapeless.Coproduct
  sealed trait Coproduct extends Product with Serializable
  sealed trait :+:[+H, +T <: Coproduct] extends Coproduct
  final case class Inl[+H, +T <: Coproduct](head : H) extends :+:[H, T]
  final case class Inr[+H, +T <: Coproduct](tail : T) extends :+:[H, T]
  sealed trait CNil extends Coproduct

  // Note that this only appears when T is a type parameter. Replaying T with
  // a concrete type (CNil or another :+:) leads to accurate warnnings
  def f[T <: Coproduct](fa: Int :+: T) =
    fa match {
      case Inl(x) => 1
      case Inr(x) => 2 // Dotty thinks this unreachable, but it is!
    }

    assert(f(Inr(Inl(-1))) == 2) // Example of reachability
}
```

liufengyun:
Fix #3443

Previously, the child compatibility check will result in a check like `CNil <: TVar(Nothing..Coproduct)`, which failed. Now we expose type param according to its variance. We only resort to `TVar` if the variance is 0.

This PR also removes the `@unchecked` check, as agreed with @allanrenucci the check should be somewhere else and handle more complex cases.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -606,20 +606,19 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       }
     }
 
-    val tvars = tp1.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
-    val protoTp1 = thisTypeMap(tp1.appliedTo(tvars))
-
-    // replace type parameter references with fresh type vars or bounds
+    // replace type parameter references with bounds
     val typeParamMap = new TypeMap {
       def apply(t: Type): Type = t match {
 
         case tp: TypeRef if tp.symbol.is(TypeParam) && tp.underlying.isInstanceOf[TypeBounds] =>
           // See tests/patmat/gadt.scala  tests/patmat/exhausting.scala
-          val bound =
-            if (variance == 0) tp.underlying.bounds      // non-variant case is not well-founded
-            else if (variance == 1) TypeBounds.upper(tp)
-            else TypeBounds.lower(tp)
-          newTypeVar(bound)
+          val exposed =
+            if (variance == 0) newTypeVar(tp.underlying.bounds)
+            else if (variance == 1) expose(tp, true)
+            else expose(tp, false)
+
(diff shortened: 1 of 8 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -778,41 +780,41 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
    *
    *  A <: X :> Y  B <: U :> V   M { type T <: A :> B }  ~~>  M { type T <: X :> V }
    */
-  def expose(tp: Type, refineCtx: Boolean = false, up: Boolean = true): Type = tp match {
+  def expose(tp: Type, up: Boolean = true): Type = tp match {
```
OlivierBlanvillain: I would remove the default value here

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -606,20 +606,19 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-            if (variance == 0) tp.underlying.bounds      // non-variant case is not well-founded
-            else if (variance == 1) TypeBounds.upper(tp)
-            else TypeBounds.lower(tp)
-          newTypeVar(bound)
+          val exposed =
+            if (variance == 0) newTypeVar(tp.underlying.bounds)
+            else if (variance == 1) expose(tp, true)
+            else expose(tp, false)
+
+          debug.println(s"$tp exposed to =====> " + exposed)
```
OlivierBlanvillain: You could put `$exposed` inside the string

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: I just removed the whole method :)

PR #3447, opened 2017-11-09
New format for trait redefined final method from AnyRef error message…

thebignet:
Further enhancement : explanation could be enhanced by telling which method from AnyRef was redefined


thebignet:
I understand what you mean.  I had the same thought as to why should there be such a specific check.  I think it is explained by this comment.

> Have to use matchingSymbol, not a method involving overridden symbols, because the scala type system understands that an abstract method here does not override a concrete method in Object. The jvm, however, does not.

According to notes on RefsChecks.scala:127, 1.2 says that final members of classes should not be overriden. Maybe it is because AnyRef is a trait (and not a class), so implementing a final def should normally work, but since methods like `wait` are implemented in the JVM for all `Object`s, there is a special case.

This is my first PR on dotty and I have no claim to be an expert.  Is this explanation correct ?

PR #3453, opened 2017-11-09
Fix #3452: Fix cycle with implicit by-name parameter

### LINKED ISSUE #3452: IndexOutOfBoundsException in implicit search
```scala
object Test {
  case class Tuple2K[H[_], T[_], X](h: H[X], t: T[X])

  trait TC[A]

  implicit def case1[F[_]](implicit t: => TC[F[Any]]): TC[Tuple2K[[_] => Any, F, Any]] = ???
  implicit def case2[A, F[_]](implicit r: TC[F[Any]]): TC[A] = ???

  implicitly[TC[Int]] // java.lang.IndexOutOfBoundsException: 0
}
```

smarter:
When creating a lazy val to represent an implicit by-name parameter, we
need to fully define its type, otherwise cycles can arise in our
constraints.

This is similar to the issue fixed in eb2fdf273f7305e79196477347593f7417f19eb4

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Implicits.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Implicits.scala
@@ -617,6 +617,7 @@ trait Implicits { self: Typer =>
      */
     val (formalValue, lazyImplicit, argCtx) = formal match {
       case ExprType(fv) =>
+        fullyDefinedType(fv, "implicit by-name parameter", pos)
         val lazyImplicit = ctx.newLazyImplicit(fv)
         (fv, lazyImplicit, lazyImplicitCtx(lazyImplicit))
       case _ => (formal, NoSymbol, ctx)
--- a/tests/neg/i3542.scala
+++ b/tests/neg/i3542.scala
@@ -1,0 +1,8 @@
+object Test {
+  trait TC[A]
+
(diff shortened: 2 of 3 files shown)


smarter:
Note that this will necessarily restrict the power of implicit by-name params (it means they cannot be used to represent [functional dependencies](https://milessabin.com/blog/2011/07/16/fundeps-in-scala/)), but I don't see a way to avoid that.

smarter:
@OlivierBlanvillain Could you check if this fixes your original problem and if this doesn't prevent you from making your library work?

OlivierBlanvillain:
@smarter It breaks other things, I'm going to minimize an example that works on master but breaks with this PR

OlivierBlanvillain:
```scala
trait ::[H, T]

trait Foo[A, R]

trait FooLowPrio {
  implicit def caseOther[A]: Foo[A, A :: Any] = null
}
object Foo extends FooLowPrio {
  implicit def caseCons[H, HR, T, TR]
    (implicit // It's a bit artificial: the by name is not required in this example...
      t: => Foo[T, TR],
      h: => Foo[H, HR]
    ): Foo[H :: T, TR] = null

  implicit def caseAny: Foo[Any, Any] = null
}

object Test {
  val implicitFoo = implicitly[Foo[Long :: Any, Any]]
}
```

PR #3457, opened 2017-11-11
Create dotty.source.Position

nicolasstucki:
All calls to dotty.source.Position.{thisLine|thisSource} are filled with the position at their call site.


nicolasstucki:
@olafurpg @jvican

compiler/src/dotty/tools/dotc/transform/SourcePositions.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  private def transformPosition(tree: RefTree)(implicit ctx: Context): Tree = {
+    if (tree.symbol.owner ne defn.DottySourcePositionModuleRef.termSymbol.moduleClass) tree
+    else if (tree.symbol.name == nme.thisLine) Literal(Constant(tree.pos.line + 1)).asInstance(tree.tpe)
+    else if (tree.symbol.name == nme.thisSource) Literal(Constant(sourceFile)).asInstance(tree.tpe)
+    else tree
+  }
+
+  private def sourceFile(implicit ctx: Context): String = {
+    // TODO This could be performed one per unit to improve performance. Worth it?
```
jvican: My results, which I will present on Monday, tell me that it's probably worth it.

compiler/src/dotty/tools/dotc/transform/SourcePositions.scala
```diff
@@ -0,0 +1,40 @@
...
+    // TODO Add a compiler flag to output absolute paths?
+    val currentPath = Paths.get(".").toAbsolutePath.getParent
```
jvican: Is doing `Paths.get(".")` safe for all the possible filesystems that NIO supports? (I'm asking, I'm not sure myself). I think a compiler flag here would be a more elegant solution, and could be configured by default from the build tool.

tests/run/sourcePosition.scala
```diff
@@ -0,0 +1,22 @@
...
+    printSourcePath()
+    printPosition()
+    println()
+    printPosition(
+
+    )
+  }
+
+  def printPosition()(implicit pos: LineNumber): Unit = println(pos)
+  def printSourcePath()(implicit src: SourcePath): Unit = println(src)
```
jvican: What would happen if you call for `SourcePath` or `LineNumber` in a named argument? I guess it's worth adding a test for that. And also for the constructors.

nicolasstucki: They work as expected.

compiler/src/dotty/tools/dotc/transform/SourcePositions.scala
ShaneDelmore: I’m not sure if this is safe I like to test with JimFS virtual filesystem as it only implements the NIO spec and no additional features.  The JVM implementation oddly implements a few things that are not part of NIO on all platforms.

nicolasstucki: I will remove the `Paths.get(".")` and use a custom base path as suggested by @jvican

library/src/dotty/source/Position.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  /** Path of a source file */
+  final class SourcePath(val value: String) extends AnyVal {
+    override def toString: String = value
+  }
+
+  /** LineNumber containing the line number of its call site.
+   *  The line number is 1-indexed.
+   */
+  implicit def thisLine: LineNumber = new LineNumber(0)
```
OlivierBlanvillain: I wonder if this should be inside `object LineNumber` to allow non-wildcard imports

nicolasstucki: Good point, I will change this.

odersky:
I think we should hold back on that until we have decided what to do with macro support.

jvican:
I'm trying Dotty out. I've tried to port https://alexn.org/blog/2017/11/10/minitest-no-crap-scala-library.html to Dotty because I want to test the code I'm writing and I don't have time to create a test library. In the process of doing so, I've realized that its only use of macros is to synthesize source locations, see `SourceLocation` https://github.com/monix/minitest/blob/2bc6f5f9efac14d9b74990f1554cbb4e48ba9a04/shared/src/main/scala/minitest/api/SourceLocation.scala.
  
I think that if this PR was revisited, I could try to fully port Minitest. It looks like macro support will take some time to hit master, and in the meanwhile I think it would be cool if early Dotty adopters have a test framework they can use. It could also be added to the community build. All these would IMO encourage early adoption.

allanrenucci:
`JUnit` should work just fine for a sbt Dotty project with https://github.com/sbt/junit-interface

OlivierBlanvillain:
@jvican This shouldn't be a blocker, you could workaround that macro by having the following Dotty only source:

```scala
final case class SourceLocation(
  fileName: Option[String],
  filePath: Option[String],
  line: Int
)

object SourceLocation {
  implicit val fromContext: SourceLocation =
    SourceLocation(Some("unsupported"), Some("unsupported"), -1)
}
```

This should be enough to cross compile minitest and your project's test. If error messages with source location is a crucial feature for your usage your I would suggest to test your project using scalac during development.

BTW is there any reason this is not implement that with exceptions? `java.lang.StackTraceElement` has filename and line number, so it might be possible to throw something when `fromContext` is called look one level up the stack...

jvican:
@OlivierBlanvillain I didnt' say this is a blocker, of course that was a possibility :wink:. This would be nice to have tho'. I want a functional test framework in Dotty, not a broken one. @allanrenucci It's true, I could have used JUnit. But my point was that it would be better to have this feature in the compiler (even if it's temporarily) because that would allow you to port some of the Scala ecosystem before having fully functional macros.
  
I went ahead and ported minitest to Dotty anyway (dotty branch: https://github.com/dotty-jvican/dotty/tree/dotty-sbt-1.0+implicit-positions, minitest branch: https://github.com/dotty-jvican/dotty/tree/dotty-sbt-1.0+implicit-positions). I had to use a Dotty version that has sbt 1.0 and this PR, and publish that locally.
  
Everything compiles, but the tests fail to run with:
  
```scala
sbt @ minitest > minitestJVM/test
[error] java.lang.ClassFormatError: Class file version does not support constant tag 15 in class file minitest/runner/Runner
[error] 	at java.lang.ClassLoader.defineClass1(Native Method)
[error] 	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
[error] 	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[error] 	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
[error] 	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
[error] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
[error] 	at java.security.AccessController.doPrivileged(Native Method)
[error] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[error] 	at java.lang.Class.getDeclaredMethods0(Native Method)
[error] 	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
[error] 	at java.lang.Class.privateGetMethodRecursive(Class.java:3048)
[error] 	at java.lang.Class.getMethod0(Class.java:3018)
[error] 	at java.lang.Class.getMethod(Class.java:1784)
[error] 	at sbt.TestFramework$.getFingerprints(TestFramework.scala:131)
[error] 	at sbt.Tests$.$anonfun$discover$1(Tests.scala:352)
[error] 	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
[error] 	at scala.collection.immutable.List.foreach(List.scala:389)
[error] 	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
[error] 	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
[error] 	at scala.collection.immutable.List.flatMap(List.scala:352)
[error] 	at sbt.Tests$.discover(Tests.scala:352)
[error] 	at sbt.Defaults$.$anonfun$detectTests$1(Defaults.scala:943)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:44)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:42)
[error] 	at sbt.std.Transform$$anon$4.work(System.scala:64)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:257)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:16)
[error] 	at sbt.Execute.work(Execute.scala:266)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:257)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:167)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:32)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] (minitestJVM/test:definedTests) java.lang.ClassFormatError: Class file version does not support constant tag 15 in class file minitest/runner/Runner
[error] Total time: 0 s, completed Dec 2, 2017 6:38:14 PM
```
  
It fails in `definedTests`, in sbt. Probably some invariant that is violated, or some error in codegen.

PR #3461, opened 2017-11-12
Avoid loading compilation units twice

nicolasstucki:


compiler/src/dotty/tools/dotc/FromTasty.scala
```diff
@@ -71,12 +71,14 @@ object FromTasty extends Driver {
...
-            List(CompilationUnit.mkCompilationUnit(clsd, unpickled, forceTrees = true))
-          case _ => Nil
+        def compilationUnit(className: TypeName): Option[CompilationUnit] = {
+          tree(className).flatMap { case (clsd, unpickled) =>
+            if (unpickled.isEmpty) None
+            else Some(CompilationUnit.mkCompilationUnit(clsd, unpickled, forceTrees = true))
+          }
         }
-        compilationUnits
+        // Try to load it from the class, or else if the class is not present try to load it from the object with that name
```
smarter: I think this comment could be improved. I would say something like:
> The TASTY section in a/b/C.class may either contain a class a.b.C, an object a.b.C, or both.
> We first try to load the class and fallback to loading the object if the class doesn't exist.
> Note that if both the class and the object are present, then loading the class will also load the object,
> this is why we use `orElse` here, otherwise we could load the object twice!

compiler/src/dotty/tools/dotc/FromTasty.scala
```diff
@@ -71,12 +71,14 @@ object FromTasty extends Driver {
...
+        compilationUnit(className).orElse(compilationUnit(className.moduleClassName)).toList
```
smarter: Return type should be an Option and not a List now I guess.

compiler/src/dotty/tools/dotc/FromTasty.scala
nicolasstucki: Done

compiler/src/dotty/tools/dotc/FromTasty.scala
nicolasstucki: Done. It was only IntelliJ that complained about the `Option` in the flat map.

PR #3463, opened 2017-11-12
Fix #3408 Implement SIP-27: Trailing Commas

### LINKED ISSUE #3408: Implement SIP-27: Trailing Commas
Accepted SIP: http://docs.scala-lang.org/sips/completed/trailing-commas.html
Scala 2 implementation: https://github.com/scala/scala/pull/5245

CucumisSativus:
Scanner implemented to accept trailing commas before new line, implementation form scala repo https://github.com/scala/scala/pull/5245

There is one negative case which should fail, but does not
`trait ImportSelectors { import foo.{ Ev0, Ev1, } }`
Would appreciate any help there

The rest seams to be working, but I would be glad if you could double check my tests


tests/neg/trailingCommas/trailingCommas.scala
```diff
@@ -0,0 +1,51 @@
+package foo
+
+// Multi-line only cases: make sure trailing commas are only supported when multi-line
+
+trait ArgumentExprs1 { f(23, "bar", )(Ev0, Ev1) } // error // error
+trait ArgumentExprs2 { f(23, "bar")(Ev0, Ev1, ) } // error // error
+trait ArgumentExprs3 { new C(23, "bar", )(Ev0, Ev1) } // error // error
+trait ArgumentExprs4 { new C(23, "bar")(Ev0, Ev1, ) } // error // error
```
liufengyun: It would be nice to avoided undefined identifiers so that we only get errors about trailing commas. Same for the code below.

CucumisSativus:
hey @liufengyun i hope i resolved the issues you addressed + your fix to `importSelectors()` worked like a charm :)

tests/neg/trailingCommas/trailingCommas.scala
```diff
@@ -0,0 +1,58 @@
+package foo
+
+// Multi-line only cases: make sure trailing commas are only supported when multi-line
+
+trait ArgumentExprs1 { validMethod(23, "bar", )(Ev0, Ev1) } // error // error
```
allanrenucci: What are the two errors emitted here? One is the invalid trailing comma. What is the other one?

CucumisSativus: The first one is `illegal start of simple expression`. The second is as follow
```
5 |trait ArgumentExprs1 { validMethod(23, "bar", )(Ev0, Ev1) } // error // error
  |                                            ^
  |too many arguments for method validMethod: (foo: Int, bar: String)(implicit ev0: foo.package.Ev0, ev1: foo.package.Ev1): Int
```

liufengyun:
LGTM, thanks a lot @CucumisSativus !

CucumisSativus:
Thank @liufengyun and @allanrenucci for your time :)

PR #3464, opened 2017-11-12
Add dependent function types

odersky:
We now support dependent function types as long as there are no inter-parameter dependencies. 
Assuming

    class C { type T }

a dependent function type is written like this:

     (x: C) => x.T

It gets represented as a normal function type that approximates the result type together with a refinement that adds an `apply` method with the precise type. E.g. the type above would be represented as

    Function1[C, C#T] { def apply(x: C): x.T }

~~For the moment, dependent function types cannot be implicit. Maybe we can lift that restriction in the future.~~

Dependent function types are important, because they are already in DOT, and they are probably a key feature to enable polymorphic effect systems.

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -184,20 +184,17 @@ class PlainPrinter(_ctx: Context) extends Printer {
       case NoPrefix =>
         "<noprefix>"
       case tp: MethodType =>
-        def paramText(name: TermName, tp: Type) = toText(name) ~ ": " ~ toText(tp)
         changePrec(GlobalPrec) {
-          (if (tp.isImplicitMethod) "(implicit " else "(") ~
-            Text((tp.paramNames, tp.paramInfos).zipped map paramText, ", ") ~
+          (if (tp.isImplicitMethod) "(implicit " else "(") ~ paramsText(tp) ~
           (if (tp.resultType.isInstanceOf[MethodType]) ")" else "): ") ~
           toText(tp.resultType)
         }
       case tp: ExprType =>
         changePrec(GlobalPrec) { "=> " ~ toText(tp.resultType) }
       case tp: TypeLambda =>
-        def paramText(name: Name, bounds: TypeBounds): Text = name.unexpandedName.toString ~ toText(bounds)
         changePrec(GlobalPrec) {
-          "[" ~ Text((tp.paramNames, tp.paramInfos).zipped.map(paramText), ", ") ~
-          "]" ~ lambdaHash(tp) ~ (" => " provided !tp.resultType.isInstanceOf[MethodType]) ~
+          "[" ~ paramsText(tp) ~ "]" ~ lambdaHash(tp) ~
+          (" => " provided !tp.resultType.isInstanceOf[MethodType]) ~
           toTextGlobal(tp.resultType)
         }
       case AnnotatedType(tpe, annot) =>
@@ -220,6 +217,11 @@ class PlainPrinter(_ctx: Context) extends Printer {
         tp.fallbackToText(this)
     }
   }.close
+
+  protected def paramsText(tp: LambdaType): Text = {
+    def paramText(name: Name, tp: Type) = toText(name) ~ toTextRHS(tp)
+    Text((tp.paramNames, tp.paramInfos).zipped.map(paramText), ", ")
+  }
 
   protected def ParamRefNameString(name: Name): String = name.toString
 
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -146,10 +146,14 @@ class PlainPrinter(_ctx: Context) extends Printer {
         toTextRef(tp) ~ ".type"
       case tp: TermRef if tp.denot.isOverloaded =>
         "<overloaded " ~ toTextRef(tp) ~ ">"
-      case tp: SingletonType =>
-        toTextLocal(tp.underlying) ~ "(" ~ toTextRef(tp) ~ ")"
       case tp: TypeRef =>
         toTextPrefix(tp.prefix) ~ selectionString(tp)
+      case tp: TermParamRef =>
+        ParamRefNameString(tp) ~ ".type"
+      case tp: TypeParamRef =>
+        ParamRefNameString(tp) ~ lambdaHash(tp.binder)
+      case tp: SingletonType =>
+        toTextLocal(tp.underlying) ~ "(" ~ toTextRef(tp) ~ ")"
       case AppliedType(tycon, args) =>
         (toTextLocal(tycon) ~ "[" ~ Text(args map argText, ", ") ~ "]").close
       case tp: RefinedType =>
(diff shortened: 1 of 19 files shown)


nicolasstucki:
We should have some neg tests for implicit dependent function types.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -684,173 +684,204 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+      val appMeth = ctx.newSymbol(ctx.owner, nme.apply, Synthetic | Deferred, mt)
+      val appDef = assignType(
+        untpd.DefDef(appMeth.name, Nil, List(params1), resultTpt, EmptyTree),
+        appMeth)
+      RefinedTypeTree(core, List(appDef), ctx.owner.asClass)
+    }
+
+    args match {
+      case ValDef(_, _, _) :: _ =>
+        typedDependent(args.asInstanceOf[List[ValDef]])(
```
OlivierBlanvillain: I'm a bit confused by this cast, args matches `ValDef()` and is casted to a `List[ValDef]`??

LPTK:
This is great. Now we can almost have a nice first-class polymorphic lambda syntax.

```scala
> def baz(f: (t: C) => List[t.T] => List[t.T]): Int = f(new C{type T = Int})(List(1,2,3)).head
> baz(t => (ls:List[t.T]) => ls.reverse)
res0: Int = 3
```

Unfortunately, we currently need to explicitly state the type of `ls`, and also the whole thing stops working when `t` is made implicit.

Ideally, is it reasonable to expect the following program to work?

```scala
def baz(f: implicit (t: C) => List[t.T] => List[t.T]): Int = f(new C{type T = Int})(List(1,2,3)).head
baz(ls => ls.reverse)
```

compiler/src/dotty/tools/dotc/typer/Typer.scala
OlivierBlanvillain: Oh I didn't see the `::`

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -684,173 +684,204 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
+          ctx.fresh.setOwner(ctx.newRefinedClassSymbol).setNewScope)
+      case _ =>
+        typed(cpy.AppliedTypeTree(tree)(untpd.TypeTree(funCls.typeRef), args :+ body), pt)
+    }
+  }
 
-      val (protoFormals, protoResult) = decomposeProtoFunction(pt, params.length)
+  def typedFunctionValue(tree: untpd.Function, pt: Type)(implicit ctx: Context) = {
+    val untpd.Function(args, body) = tree
+    val params = args.asInstanceOf[List[untpd.ValDef]]
```
OlivierBlanvillain: Maybe type annotate `args` in the pattern above?

tests/pos/depfuntype.scala
```diff
@@ -0,0 +1,35 @@
+object Test {
+
+  trait C { type M; val m: M }
+
+  type DF = (x: C) => x.M
+  val depfun1: DF = (x: C) => x.m
```
OlivierBlanvillain: I like than both the type and the value have the same syntax :)

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -49,6 +49,27 @@ object Inferencing {
   def instantiateSelected(tp: Type, tvars: List[Type])(implicit ctx: Context): Unit =
     new IsFullyDefinedAccumulator(new ForceDegree.Value(tvars.contains, minimizeAll = true)).process(tp)
 
+  /** Instantiate any type variables in `tp` whose bounds contain a reference to
+   *  one of the parameters in `tparams` or `vparamss`.
+   */
+  def instantiateDependent(tp: Type, tparams: List[Symbol], vparamss: List[List[Symbol]])(implicit ctx: Context): Unit = {
+    val dependentVars = new TypeAccumulator[Set[TypeVar]] {
+      lazy val params = (vparamss :\ tparams)( _ ::: _)
```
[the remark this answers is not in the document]
nicolasstucki: Both will not conserve the last param list in vparamss, this might be useful in all non type polymorphic methods with a single argument list ;)

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -697,160 +697,190 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
...
-        case pt: TypeVar if untpd.isFunctionWithUnknownParamType(tree) =>
-          // try to instantiate `pt` if this is possible. If it does not
-          // work the error will be reported later in `inferredParam`,
-          // when we try to infer the parameter type.
-          isFullyDefined(pt, ForceDegree.noBottom)
-        case _ =>
-      }
+    val funCls = defn.FunctionClass(args.length, isImplicit)
+
+    def typedDependent(params: List[ValDef])(implicit ctx: Context) = {
```
smarter: I would add a comment explaining to quickly describe what this function returns.

tests/pos/depfuntype.scala
```diff
@@ -15,4 +15,21 @@ object Test {
   val d: C = c
   val z = depfun3(d)
   val z1: d.M = z
+
+  // Reproduced here because the one from DottyPredef is lacking a Tasty tree and
+  // therefore can't be inlined when testing non-bootstrapped.
+  // But inlining `implicitly` is vital to make the definition of `ifun` below work.
+  inline final def implicitly[T](implicit ev: T): T = ev
```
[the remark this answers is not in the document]
smarter: The fact that adding `inline` changes how a method is typechecked doesn't seem like a great thing. I think @OlivierBlanvillain has some examples where it breaks working code.

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -146,10 +146,14 @@ class PlainPrinter(_ctx: Context) extends Printer {
...
         "<overloaded " ~ toTextRef(tp) ~ ">"
-      case tp: SingletonType =>
-        toTextLocal(tp.underlying) ~ "(" ~ toTextRef(tp) ~ ")"
       case tp: TypeRef =>
         toTextPrefix(tp.prefix) ~ selectionString(tp)
+      case tp: TermParamRef =>
+        ParamRefNameString(tp) ~ ".type"
+      case tp: TypeParamRef =>
+        ParamRefNameString(tp) ~ lambdaHash(tp.binder)
+      case tp: SingletonType =>
```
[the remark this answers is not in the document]
odersky: Should be fixed now.

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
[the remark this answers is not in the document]
odersky: `(tparams :: vparamss).flatten` is indeed much clearer.

compiler/src/dotty/tools/dotc/typer/Typer.scala
odersky: Good idea.

tests/pos/depfuntype.scala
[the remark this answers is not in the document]
odersky: I am a bit nervous about changing it to `ev.type`. The inferencer has to do major acrobatics to deal with implicit method types. It might interfere with something else. The curent type of `implicitly` is simpler and clearer.

tests/pos/depfuntype.scala
esarbe: Yes. This is something that I really like about Scala in general.

PR #3465, opened 2017-11-13
Add an error message class for a case where a trait is expected

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for the following case - `Checking.scala:551`

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1888,4 +1888,32 @@ object messages {
     }
     val explanation = "A statement is either an import, a definition or an expression."
   }
+
+  case class TraitIsExpected(tref: TypeRef)(implicit ctx: Context) extends Message(TraitIsExpectedID) {
+    val kind = "Syntax"
+    val msg = hl"$tref is not a trait"
+    val explanation = {
+      val errorCodeExample =
+        """class A
+          |class B
+          |
+          |val a = new A with B // will fail with a compile error - B is not a trait"""
+      val codeExample =
+        """class A
+          |trait B
+          |
+          |val a = new A with B // compiles normally"""
+
+      hl"""Only traits can be mixed into classes using a ${"with"} keyword.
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,32 @@ object messages {
...
+
+  case class TraitIsExpected(tref: TypeRef)(implicit ctx: Context) extends Message(TraitIsExpectedID) {
+    val kind = "Syntax"
+    val msg = hl"$tref is not a trait"
+    val explanation = {
+      val errorCodeExample =
+        """class A
+          |class B
+          |
+          |val a = new A with B // will fail with a compile error - B is not a trait"""
```
allanrenucci: You should `stripMargin` here

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,32 @@ object messages {
...
+      val codeExample =
+        """class A
+          |trait B
+          |
+          |val a = new A with B // compiles normally"""
```
allanrenucci: You should `stripMargin` here

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1888,4 +1888,32 @@ object messages {
     }
     val explanation = "A statement is either an import, a definition or an expression."
   }
+
+  case class TraitIsExpected(tref: TypeRef)(implicit ctx: Context) extends Message(TraitIsExpectedID) {
```
allanrenucci: I would rather pass the symbol instead of the `TypeRef`

maseev:
@allanrenucci all done. PTAL.

maseev:
It seems that the CLA check is still pending. Is there anything I can do to make it pass? :worried:

nicolasstucki:
Sometimes the CLA stalls

PR #3466, opened 2017-11-13
Constant folding improvements 

gan74:
Expressions with constant type are now folded correctly.
ConstantFold doesn't call typer's `ConstFold.apply` on every expression making it faster.
~~`if(!e) { a } else { b }` is now transformed into `if(e) { b } else { a }`~~

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/localopt/ConstantFold.scala
+++ b/compiler/src/dotty/tools/dotc/transform/localopt/ConstantFold.scala
@@ -21,10 +21,11 @@ import Simplify.desugarIdent
  *    out (nested) if with equivalent branches wrt to isSimilar. For example:
  *      - if (b) exp else exp → b; exp
  *      - if (b1) e1 else if (b2) e1 else e2 → if (b1 || b2) e1 else e2
+ *      - if(!b) e1 else e2 → if(b) e2 else e1
  *
  *  - Constant propagation over pattern matching.
  *
- *  @author DarkDimius, OlivierBlanvillain
+ *  @author DarkDimius, OlivierBlanvillain, gan74
  */
  class ConstantFold(val simplifyPhase: Simplify) extends Optimisation {
   import ast.tpd._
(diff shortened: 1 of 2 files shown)


DarkDimius:
>if(!e) { a } else { b } is now transformed into if(e) { b } else { a }

This case was intentionally not included.
On benchmarking that I did for original optimization this stopped branches 
from being merged by further optimizations

gan74:
>  On benchmarking that I did for original optimization this stopped branches
>  On benchmarking that I did for original optimization this stopped branches from being merged by further optimizations

Interesting. I didn't notice any adverse effects. Do you know which optimizations are affected?

compiler/src/dotty/tools/dotc/transform/localopt/ConstantFold.scala
```diff
@@ -21,18 +21,19 @@ import Simplify.desugarIdent
...
- *  @author DarkDimius, OlivierBlanvillain
+ *  @author DarkDimius, OlivierBlanvillain, gan74
  */
  class ConstantFold(val simplifyPhase: Simplify) extends Optimisation {
   import ast.tpd._
 
   def visitor(implicit ctx: Context) = NoVisitor
   def clear(): Unit = ()
 
-  def transformer(implicit ctx: Context): Tree => Tree = { x => preEval(x) match {
```
OlivierBlanvillain: I guess without `preEval` you could remove the `x => x match` altogether to turn this into a partial function:

```scala
def transformer(implicit ctx: Context): Tree => Tree = {
  case If(,,) => ...
}
```

compiler/src/dotty/tools/dotc/transform/localopt/ConstantFold.scala
```diff
@@ -76,9 +77,6 @@ import Simplify.desugarIdent
     //       cond.select(defn.Boolean_&&).appliedTo(elsep)
     //   the other case ins't handled intentionally. See previous case for explanation
 
-    case If(t @ Select(recv, _), thenp, elsep) if t.symbol eq defn.Boolean_! =>
```
OlivierBlanvillain: I'm confused, in the PR description you say that `if(!e) { a } else { b }` now `if(e) { b } else { a }`, but it seams that you just removed the case doing that... Could you had a test for this example?

gan74: Hi, I looked into it: you are right. It's an error on my part, I didn't add the feature, I only added documentation for it. 

Sorry or the confusion...

PR #3473, opened 2017-11-14
Add an error message class for sealed classes

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for the following case - `Namer.scala:880`. 

Unfortunately, this PR doesn't have a unit test because I wasn't able to find a way to trigger this particular behavior. It seems to me that it's impossible at the moment to check some code which is separated into two different compilation units inside a unit test. 

Maybe I'm simply missing something.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1966,4 +1966,10 @@ object messages {
           |impossible to instantiate an object of this class"""
     }
   }
+
+  case class UnableToExtendSealedClass(pclazz: Symbol)(implicit ctx: Context) extends Message(UnableToExtendSealedClassID) {
+    val kind = "Syntax"
+    val msg = hl"Cannot extend ${"sealed"} $pclazz in a different source file"
+    val explanation = "A sealed class or trait can only be extended in the same file as its declaration"
+  }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -117,6 +117,7 @@ public enum ErrorMessageID {
     UnapplyInvalidNumberOfArgumentsID,
     StaticFieldsOnlyAllowedInObjectsID,
     CyclicInheritanceID,
+    UnableToExtendSealedClassID,
     ;
 
     public int errorNumber() {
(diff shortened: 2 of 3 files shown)


maseev:
I'm just wondering, isn't this supposed to be 

> "Add" instead of "Added"

and not

> "Added" instead of "Add"

in the 6th rule of commit messages?

According to this [rule](https://chris.beams.io/posts/git-commit/#imperative), you need to use the present simple tense and not the past tense for your commit messages. Otherwise, it would be an indicative mood which is all about reporting facts.

allanrenucci:
>I'm just wondering, isn't this supposed to be "Add" instead of "Added"

Yes. Feel free to submit a PR. The message is in `PullRequestService.scala:270`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1922,4 +1922,36 @@ object messages {
     val msg = hl"Traits cannot redefine final $method from ${"class AnyRef"}."
     val explanation = ""
   }
+
+  case class UnableToExtendSealedClass(pclazz: Symbol)(implicit ctx: Context) extends Message(UnableToExtendSealedClassID) {
+    val kind = "Syntax"
+    val msg = hl"cannot extend sealed $pclazz in different compilation unit"
```
OlivierBlanvillain: I would refactor that and the extended message to only talk about files instead. It seams unnecessary to talk about compilation units here.

Here is what we have in the spec if that helps:

> The sealed modifier applies to class definitions. A sealed class may not be directly inherited, except if the inheriting template is defined in the same source file as the inherited class. However, subclasses of a sealed class can be inherited anywhere.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1922,4 +1922,36 @@ object messages {
...
+          |sealed class A
+          |
+          |//B.scala
+          |class B extends A""".stripMargin
+      val codeExample =
+        """//A.scala
+          |sealed class A
+          |class B extends A""".stripMargin
+
+      hl"""A ${"sealed"} keyword makes sure that all classes that extend a sealed class are
```
OlivierBlanvillain: TBH I'm not convinced a super long example with correct/incorrect examples is helpful here. [The top SO answer](https://stackoverflow.com/questions/11203268/what-is-a-sealed-trait) just gets to the point:

A `sealed` trait can be extended only in the same file as its declaration.

maseev:
@OlivierBlanvillain PTAL

OlivierBlanvillain:
:+1:, thanks!

maseev:
@OlivierBlanvillain, @allanrenucci sorry, I forgot to pull your changes first before doing a rebase. I applied the same changes you did to the message and explanation fields.

OlivierBlanvillain:
No worries, thank you for the PR!

Issue #3469, opened 2017-11-14
Unnecessary exhaustivity warning when combining refinement types w/ params


AleksanderBG:
The following code:

```scala
object O {
  sealed trait Trait[+A] { type T }
  case class CaseClass[+A](a: A) extends Trait[A] { type T = Nothing }

  def id[TT, A](v: Trait[A] { type T = TT }): Trait[A] { type T = TT } =
    v match {
      case CaseClass(a) => CaseClass(a)
    }
}
```

causes the following warning:
```
-- [E028] Pattern Match Exhaustivity Warning: sample/1:6:4 ---------------------
6 |    v match {
  |    ^
  |    match may not be exhaustive.
  |
  |    It would fail on: CaseClass(_)
```

@liufengyun - this happens because the refinement type forces usage of `typeParamMap` in `instantiate`. This in turn initializes the param in the pattern space to `Any`, so instead of `CaseClass[A]` we get `CaseClass[Any]`. This was found when rebasing #3454 - one test had to be ignored. I should also mention that with the POC typechecker integration I've described in #3454, this code is handled correctly.

liufengyun:
Thanks a lot for reporting @AleksanderBG . This problem doesn't  appear in 9166db3 -- I need to send a PR recently.

As you observed in #3454, child instantiation is very tricky, with refinement types, it gets even trickier. AFAIK, child instantiation is the first place where we use inference infrastructure to infer `prefix` and `refinement`, it's largely unexplored land.

Note child instantiation is needed no matter what algorithm we use. For refinement types, I don't yet see a principled way to do it.

AleksanderBG:
For clarity's sake: refinement types are not needed to trigger this behaviour. Here's a minimized test from #3454 which doesn't use refinement types, but triggers similar warning (missing pattern for `NCons`):

```scala
object O {
  sealed trait Nat
  type Zero = Nat
  sealed trait Succ[N <: Nat] extends Nat

  sealed trait NVec[N <: Nat, +A]
  case class NCons[N <: Nat, +A](head: A, tail: NVec[N, A]) extends NVec[Succ[N], A]

  def id[N <: Nat, A](v: NVec[N, A]): NVec[N, A] = v match {
    case NCons(h, t) => NCons(h, t)
  }
}
```

[...]

Issue #3455, opened 2017-11-10
Patmat exhaustivity checks sometimes don't consider path-dependant children


AleksanderBG:
The following code compiles without any warnings:

```scala
trait AxisCompanion {
   sealed trait Format
   object Format {
      case object Decimal extends Format
      case object Integer extends Format
   }
}
object Axis extends AxisCompanion
class Axis {
   import Axis._
   def test( f: Format ) = f match {
      case Format.Integer => "Int"
   }
}
```

Note that `Axis#test` is inexhaustive. The underlying problem is that when children of `Format` are to be located, symbols for `AxisCompanion.Format.Decimal` and `.Integer` are obtained and are subsequently ousted in `refine`, as they are not subtypes of `Axis.Format`. The paths of `.Decimal` and `.Format` should be adapted to the path of their parent type.

@liufengyun - this was discovered when implementing intersection-based useless pattern checks in #3454. I found the cause for it, but as I'm running out of time and I'd prefer to focus on typechecker integration, I'm simply raising the issue here.

liufengyun:
@AleksanderBG Thanks for reporting, it has to do with _prefix inference_ in _child instantiation_. I'll have a look.

liufengyun:
Fixed in https://github.com/lampepfl/dotty/pull/3475

PR #3475, opened 2017-11-14
Fix #3333: adapt child instantiation

### LINKED ISSUE #3333: Patmat exhaustivity checks do not consider all possible type children

liufengyun:
Fix #3333 #3455 #3469: adapt child instantiation

Structural types in #3333 is left open.

- If a child type parameter is not constrained, instantiate it to Wildcard
- In prefix inference, forget `ThisType` for modules.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -584,10 +584,9 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     // precondition: `tp1` should have the shape `path.Child`, thus `ThisType` is always covariant
     val thisTypeMap = new TypeMap {
       def apply(t: Type): Type = t match {
-        case tp @ ThisType(tref) if !tref.symbol.isStaticOwner && !tref.symbol.is(Module)  =>
-          // TODO: stackoverflow here
-          // newTypeVar(TypeBounds.upper(mapOver(tp.underlying)))
-          newTypeVar(TypeBounds.upper(mapOver(tref & tref.classSymbol.asClass.givenSelfType)))
+        case tp @ ThisType(tref) if !tref.symbol.isStaticOwner  =>
+          if (tref.symbol.is(Module)) mapOver(tref)
+          else newTypeVar(TypeBounds.upper(tp.underlying))
         case _ =>
           mapOver(t)
       }
@@ -633,7 +632,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     else {
       val protoTp2 = typeParamMap(tp2)
       if (protoTp1 <:< protoTp2) {
-        isFullyDefined(protoTp1 & protoTp2, force)
+        isFullyDefined(AndType(protoTp1, protoTp2), force)
         instUndetMap(protoTp1)
       }
       else {
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -626,14 +626,14 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     val protoTp1 = thisTypeMap(tp1.appliedTo(tvars))
 
     if (protoTp1 <:< tp2) {
-      isFullyDefined(protoTp1, force)
-      instUndetMap(protoTp1)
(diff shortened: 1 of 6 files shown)


liufengyun:
@AleksanderBG Do you want to review? This may be helpful to your WIP.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -611,13 +609,32 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+      minimizeAll = false
+    )
+
     val tvars = tp1.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
     val protoTp1 = thisTypeMap(tp1.appliedTo(tvars))
 
-    if (protoTp1 <:< tp2 && isFullyDefined(protoTp1, ForceDegree.noBottom)) protoTp1
+    if (protoTp1 <:< tp2) {
+      isFullyDefined(protoTp1, force)
+      instUndetMap(protoTp1)
```
AleksanderBG: If I understand correctly how `isFullyDefined` behaves, then if it returns `true`, all TVars in `protoTp1` should be already instantiated and `instUndetMap` will be a no-op. Would it make sense to only use `instUndetMap` if `isFullyDefined` returned `false`?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -584,10 +584,9 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     // precondition: `tp1` should have the shape `path.Child`, thus `ThisType` is always covariant
     val thisTypeMap = new TypeMap {
       def apply(t: Type): Type = t match {
-        case tp @ ThisType(tref) if !tref.symbol.isStaticOwner && !tref.symbol.is(Module)  =>
-          // TODO: stackoverflow here
-          // newTypeVar(TypeBounds.upper(mapOver(tp.underlying)))
```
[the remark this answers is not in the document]
liufengyun: I forgot exactly what's the problem, but it's related to `ClassInfo#selfType`. `tp.underlying` is supposed to be decreasing (I assume), so I think the case you mention will not occur. For the task you want to deal with, the memoization in your code is correct.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: Yes, good catch, I've updated the code, thanks a lot.

PR #3441, opened 2017-11-07
Upgrade to sbt 1

Duhemm:
Projects from the community build to port to sbt1:

- [x] [algebra](https://github.com/dotty-staging/algebra/tree/sbt1)
- [x] [better-files](https://github.com/dotty-staging/better-files/tree/sbt1)
- [x] [scalacheck](https://github.com/dotty-staging/scalacheck/tree/sbt1)
- [x] [ScalaPB](https://github.com/dotty-staging/ScalaPB/tree/sbt1)
- [x] [scalatest](https://github.com/dotty-staging/scalatest/tree/sbt1)
- [x] [scopt](https://github.com/dotty-staging/scopt/tree/sbt1)
- [x] [squants](https://github.com/dotty-staging/squants/tree/sbt1)
- [x] [collections-strawman](https://github.com/dotty-staging/collection-strawman/tree/sbt1)

Issue #1574, opened 2016-10-10
Scala.JS sandbox is broken


felixmulder:
On master doing: `sbt sjsSandbox/compile` results in:

```
[info] Loading project definition from /home/fixel/Projects/dotty-master/project
[info] Set current project to dotty (in build file:/home/fixel/Projects/dotty-master/)
> sjsSandbox/compile
[info] Resolving org.scala-js#scalajs-ir_2.11;0.6.8 ...
[info] Compiling 1 Scala sources to /home/fixel/Projects/dotty-master/sandbox/scalajs/target/scala-2.11/classes...
[error] exception occurred while compiling /home/fixel/Projects/dotty-master/sandbox/scalajs/hello.scala
[error] Exception in thread "main" java.lang.AssertionError: assertion failed: phase getClass has unmet requirement: class dotty.tools.dotc.transform.FunctionalInterfaces should precede this phase
[error]     at scala.Predef$.assert(Predef.scala:165)
[error]     at dotty.tools.dotc.core.Phases$PhasesBase$class.checkRequirements$1(Phases.scala:164)
[error]     at dotty.tools.dotc.core.Phases$PhasesBase$$anonfun$usePhases$2.apply(Phases.scala:177)
[error]     at dotty.tools.dotc.core.Phases$PhasesBase$$anonfun$usePhases$2.apply(Phases.scala:176)
[error]     at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
[error]     at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
[error]     at dotty.tools.dotc.core.Phases$PhasesBase$class.usePhases(Phases.scala:176)
[error]     at dotty.tools.dotc.core.Contexts$ContextBase.usePhases(Contexts.scala:527)
[error]     at dotty.tools.dotc.Run$$anonfun$compileUnits$1.apply$mcV$sp(Run.scala:66)
[error]     at dotty.tools.dotc.Run$$anonfun$compileUnits$1.apply(Run.scala:62)
[error]     at dotty.tools.dotc.Run$$anonfun$compileUnits$1.apply(Run.scala:62)
[error]     at dotty.tools.dotc.util.Stats$.monitorHeartBeat(Stats.scala:76)
[error]     at dotty.tools.dotc.Run.compileUnits(Run.scala:62)
[error]     at dotty.tools.dotc.Run.compileSources(Run.scala:59)
[...]

PR #3481, opened 2017-11-15
[WIP] Resurrect the Scala.js backend

smarter:
Happy birthday @sjrd  :stuck_out_tongue: 

This partially reverts 06a3d47 to bring
back Dotty.js from the dead. Many things are broken still, but this is
enough to do some experiments:
- The backend was initially removed because of #1574, I don't see why
  GetClass has a requirement on FunctionalInterfaces, so I just removed
  it ¯\\\_(ツ)\_/¯
- Scala.js was upgraded to 0.6.19. More recent versions of Scala.js cannot be
  used because they require sbt 0.13.16 and we're blocked by
  sbt/sbt#3460, this won't be an issue anymore
  once we switch to sbt 1 (see #3441)
- GenSJSIR was hacked to compile with 0.6.19 but is missing most of the
  changes to the Scala.js backend phase for Scala 2 that happened
  since it was initially ported to Dotty from Scala.js 0.6.8
- scalajs-ir does not compile with Dotty anymore, see comments in Build.scala
- The Scala.js backend explodes when extending a Scala 2 trait because
  of the way LinkScala2Impls transforms supercalls to these traits.
  Luckily, we don't need to extend js.JSApp anymore to make a Hello World :P.

I verified that the following works:
```scala
sbt sjsSandbox/run
```
But I don't know if anything else does!


Blaisorblade:
Closing this for now as outdated — hopefully somebody (@sjrd?) will resurrect it.

PR #3483, opened 2017-11-15
Add an error message for a cyclic inheritance case

maseev:
This PR is related to #1589. Specifically, it adds a new error message class for a cyclic inheritance case - `Namer.scala:872`. I decided not to include an explanation in this particular message because it's pretty self-explanatory.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1949,4 +1949,21 @@ object messages {
     val explanation =
       hl"${"@static"} members are only allowed inside objects."
   }
+
+  case class CyclicInheritance(symbol: Symbol, addendum: String)(implicit ctx: Context) extends Message(CyclicInheritanceID) {
+    val kind = "Syntax"
+    val msg = hl"Cyclic inheritance: $symbol extends itself$addendum"
+    val explanation = {
+      val codeExample = "class A extends A"
+
+      hl"""Cyclic inheritance is prohibited in Dotty.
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1922,4 +1922,10 @@ object messages {
     val msg = hl"Traits cannot redefine final $method from ${"class AnyRef"}."
     val explanation = ""
   }
+
+  case class CyclicInheritance(symbol: Symbol, addendum: String)(implicit ctx: Context) extends Message(CyclicInheritanceID) {
+    val kind = "Syntax"
+    val msg = hl"cyclic inheritance: $symbol extends itself$addendum"
+    val explanation = ""
```
nicolasstucki: This would need a better explanation.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1922,4 +1922,10 @@ object messages {
     val msg = hl"Traits cannot redefine final $method from ${"class AnyRef"}."
     val explanation = ""
   }
+
+  case class CyclicInheritance(symbol: Symbol, addendum: String)(implicit ctx: Context) extends Message(CyclicInheritanceID) {
+    val kind = "Syntax"
+    val msg = hl"cyclic inheritance: $symbol extends itself$addendum"
```
nicolasstucki: Start with uppercase `Cyclic`

maseev:
@nicolasstucki PTAL

PR #3485, opened 2017-11-16
 Create dotc -decompiler

nicolasstucki:
Create infrastructure to support `dotp` tasty decompiler. The implementation of the decompiler is a tasty compiler that compiles to a text representation of the loaded code (currently with Tree.show).

### DIFF
--- a/compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
@@ -24,8 +24,7 @@ class DecompilationPrinter extends Phase {
     println(unit.source)
     println(line)
 
-    val code = unit.tpdTree.show
-    println(if (ctx.useColors) printing.SyntaxHighlighting(code) else code)
+    println(unit.tpdTree.show)
     println(line)
   }
 }
--- a/compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
@@ -1,0 +1,31 @@
+package dotty.tools.dotc
+package decompiler
+
+import dotty.tools.dotc.core.Contexts._
+import dotty.tools.dotc.core.Phases.Phase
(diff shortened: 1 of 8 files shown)


nicolasstucki:
Needs rebase on #3436

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -96,6 +96,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val YnoPatmatOpt = BooleanSetting("-Yno-patmat-opt", "disable all pattern matching optimizations.")
   val YplainPrinter = BooleanSetting("-Yplain-printer", "Pretty-print using a plain printer.")
   val YprintSyms = BooleanSetting("-Yprint-syms", "when printing trees print info in symbols instead of corresponding info in trees.")
+  val YprintNoPrivate = BooleanSetting("-Y-print-no-private", "when printing trees print do not print private/protected members.")
```
smarter: should -Yprint, not -Y-print

nicolasstucki:
Rebased and squashed

compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
```diff
@@ -0,0 +1,31 @@
...
+
+    val doubleLine = "=" * pageWidth
+    val line = "-" * pageWidth
+
+    println(doubleLine)
+    println(unit.source)
+    println(line)
+
+    val code = unit.tpdTree.show
+    println(if (ctx.useColors) printing.SyntaxHighlighting(code) else code)
```
allanrenucci: It should really be `printing.SyntaxHighlighting(code).mkString`. We shouldn't assume that the underlying `Iterable[Char].toString` works as we expect.

nicolasstucki: I will remove the syntax highlighting as it is not compatible with all the output formats of `show`. I will work on coloring in a separate PR.

PR #3487, opened 2017-11-16
Fix MatchError: InfixOp on TreeAccumulator

OlivierBlanvillain:
In case of errors it may be that typed trees point to untyped ones. I think it makes sense to avoid the MatchError in that case, not only in Interative mode.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -1325,7 +1325,8 @@ object Trees {
             this(this(x, arg), annot)
           case Thicket(ts) =>
             this(x, ts)
-          case _ =>
+          case _ if ctx.reporter.errorsReported || ctx.mode.is(Mode.Interactive) =>
+            // In interactive mode, errors might come from previous runs.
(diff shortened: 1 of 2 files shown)


allanrenucci:
This code snippet also causes a match error because of an untyped tree.
```scala
class Test[T](f: List[String] => T)

object o {
  implicit object a extends Test(_ map identity)
  implicit object b extends Test(_ map identity)
}
```
Not sure it is related

OlivierBlanvillain:
The test case included in the PR throws a match error on master (note that `F` is undefined)

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -1325,7 +1325,7 @@ object Trees {
             this(this(x, arg), annot)
           case Thicket(ts) =>
             this(x, ts)
-          case _ if ctx.mode.is(Mode.Interactive) =>
+          case _ =>
```
odersky: Proposal: replace with 

    case _ if ctx.errorsReported || ctx.mode.is(Mode.Interactive) =>

and adapt the comment to say that in interactive mode, errors might come from previous runs.

PR #3491, opened 2017-11-16
Fix #3480: Fix expanded name

### LINKED ISSUE #3480: Dotty sometimes generates method names with spaces in them for partial functions
```scala
class Test {
  def foo(x: PartialFunction[Int, Int]) = x(0)

  foo({ case i => i})
}
```
```java
% javap Test
Compiled from "pfi.scala"
public class Test {
  public Test();
  public int foo(scala.PartialFunction<java.lang.Object, java.lang.Object>);
  public static int Test$$local Test$$$$init$$$anonfun$1$1(int);
  public static boolean Test$$local Test$$$isDefinedAt$1(int);
}
```
@odersky Do you understand what's going on here?

odersky:
Expanded name behaved different from what would be expected
by the fullNameSeparated conventions. This led to local
dummies creeping in the expanded names for local definitions
in templates.

### DIFF
--- a/tests/run/i3006b.check
+++ b/tests/run/i3006b.check
@@ -1,3 +1,3 @@
-Foo$$init$$$bar$1
-Foo$$init$$$bar$2
-Bar$$init$$$bar$1
+Foo$$_$bar$1
+Foo$$_$bar$2
+Bar$$_$bar$1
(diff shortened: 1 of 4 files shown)


tests/run/i3006b.check
```diff
@@ -1,3 +1,3 @@
-Foo$$init$$$bar$1
-Foo$$init$$$bar$2
-Bar$$init$$$bar$1
+Foo$$_$bar$1
```
smarter: Why was this changed? Isn't this going to make debugging more difficult?

odersky: It makes things more regular. Fully qualified names do skip terms like this. The problem was that we treated the last separator of an expanded name as different from the others. This looks wrong. It's a different question what fully qualified names should do. AFAIK scalac skips term components completely.

PR #3492, opened 2017-11-16
Keep loaded pickled tasty when compiling from tasty

nicolasstucki:


compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -604,22 +604,38 @@ object Symbols {
...
         val body = unpickler.body(ctx.addMode(Mode.ReadPositions))
         myTree = body.headOption.getOrElse(tpd.EmptyTree)
+        if (ctx.settings.tasty.value)
+          myTasty = unpickler.unpickler.bytes
         unpickler = null
       }
-      myTree
     }
     private[this] var myTree: tpd.Tree /* tpd.PackageDef | tpd.TypeDef | tpd.EmptyTree */ = tpd.EmptyTree
+    private[this] var myTasty: Array[Byte] = _
```
smarter: If possible, I would rather not add a field to every ClassSymbol for this.

nicolasstucki: The alternative is to keep the `unpickler` longer. I will go back to that solution.

OlivierBlanvillain:
I guess the title should now be "Keep loaded pickled tasty when compiling from tasty", but the diff is LGTM! Did you try testing than re-pickling these trees leads to the same tasty?

PR #3493, opened 2017-11-16
Fix #3477: Add custom classpath to dotr script

### LINKED ISSUE #3477: `dist/bin/dotr` can't run with a custom `-classpath`
```
sbt dist-bootstrapped/pack
echo "object Main { def main(a: Array[String]): Unit = println(\"OK\") }" > Main.scala
./dist-bootstrapped/target/pack/bin/dotc -d out Main.scala
./dist-bootstrapped/target/pack/bin/dotr -classpath out Main
```

The last line will default to the repl instead running the application.

Note that `dist-bootstrapped/target/pack/bin/dotr` is generated from `dist/bin/dotr` when packing.

rsoeldner:
@nicolasstucki This addresses issue #3477

### DIFF
--- a/dist/bin/dotr
+++ b/dist/bin/dotr
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
-IFS=""
 # Try to autodetect real location of the script
 
 if [ -z "$PROG_HOME" ] ; then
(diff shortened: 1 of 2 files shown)


dist/bin/dotr
```diff
@@ -31,18 +31,39 @@ source "$PROG_HOME/bin/common"
...
+run_app=flase
+while [[ $# -gt 0 ]]; do
+key=$1
+case $key in
+        -repl)
+         run_repl=true
+         shift
+         ;;
+        -classpath)
+         CLASS_PATH+=":$2"
```
nicolasstucki: We should probably add the user-defined classpaths to the left of the classpath to give them the priority.

dist/bin/dotr
```diff
@@ -31,18 +31,39 @@ source "$PROG_HOME/bin/common"
...
+          shift
+          shift
+          ;;
+          *)
+           addResidual "$1"
+           shift
+           ;;
+        -d)
+        DEBUG="$DEBUG_STR"
```
nicolasstucki: Is this ever reached with the `*)` before it?

dist/bin/dotr
```diff
@@ -31,18 +31,39 @@ source "$PROG_HOME/bin/common"
...
+         shift
+         ;;
+  esac
+done
+
+if [ -z $residual_args ]; then
     echo "Starting dotty REPL..."
     eval "$PROG_HOME/bin/dotc -repl"
-elif [[ ${first_arg:0:1} == "-" ]]; then
-    eval "$PROG_HOME/bin/dotc -repl $@"
```
nicolasstucki: We should still be able to call the repl with a custom classpath and custom args. You can assume that the `-repl` is set in that case and hence `run_repl` should be true.

nicolasstucki:
You could also enable the two tests in https://github.com/lampepfl/dotty/blob/master/project/scripts/sbtBootstrappedTests#L29. That way we can avoid some regressions

nicolasstucki:
Otherwise the script code LGTM

dist/bin/dotr
```diff
@@ -31,18 +31,49 @@ source "$PROG_HOME/bin/common"
...
+           ;;
+
+  esac
+done
+
+CLASS_PATH="${classpath_args[@]}$CLASS_PATH"
+
+if [ $run_repl == true ]; then
+    echo "Starting dotty REPL"
+    echo "$PROG_HOME/bin/dotc -classpath $CLASS_PATH -repl ${residual_args[@]}"
```
nicolasstucki: Here the classpath should only contain `classpath_args` if it is not empty.

What does it mean exactly `classpath_args[@]`? I'm not familiar with the `[@]` part.

rsoeldner: @nicolasstucki Afaik, the `[*]` or `[@]` subscripts differ only within double quotes, with `[*]` each element inside the array is seperated by `IFS`

rsoeldner:
@nicolasstucki If everything is ok, I would rebase to a single commit.

allanrenucci:
I’d like to review this before we merge

allanrenucci:
@nicolasstucki can you review the last commit

Issue #3488, opened 2017-11-16
Compiler fails to inline


allanrenucci:
The following code snippet does not compile with Dotty:
```scala
class Set[A] {

  def incl(elem: A): Set[A] = ???

  inline final def + (elem: A): Set[A] = incl(elem)
}

object Set {
  def apply[A](elems: A*): Set[A] = ???
}

class Test {
  Set(1) + 1
}
```
```scala
-- Error: tests/allan/Test.scala:13:9 ------------------------------------------
13 |  Set(1) + 1
   |  ^^^^^^^^^^
   |undefined: Set_this.incl # 191: TermRef(TermRef(NoPrefix,val Set_this),incl[with sig Signature(List(java.lang.Object),Set)])
```
It compiles without the `inline` keyword.

odersky:
Fixed by #3494

PR #3494, opened 2017-11-17
Rework NamedTypes

odersky:
When analyzing #3488, the root cause seems to be that constructing TermRefs with signatures is fragile. Signatures might not be fully known at the point a TermRef is constructed, which leads to complicated compensation actions. 

In this PR I try to make NamedTypes by default symbolic (i.e. all designators should ideally be symbols, not names). We need to compensate for cross compilation unit references in Tasty trees.
These need names and signatures as before. But hopefully we can be symbol-only in the internal types. 

Postscript: Some NamedTypes are still created with a name only, in order not to force too much. But the scheme should now be much more robust than before.


odersky:
Something is not right with the CI. It seems to pick up an outdated version of SymDenotations.scala. The line that leads to the compile error (420) has been deleted and now contains something unrelated to the error message.

odersky:
It turned out to be a rebase problem.

odersky:
test performance please

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1707,84 +1754,27 @@ object Types {
...
-      val d = asMemberOf(prefix, allowPrivate = false)
-      if (d.exists || ctx.phaseId == FirstPhaseId || !lastDenotation.isInstanceOf[SymDenotation])
-        if (mySig != null && mySig.ne(Signature.OverloadedSignature)) d.atSignature(mySig).checkUnique
-        else d
-      else { // name has changed; try load in earlier phase and make current
-        val d = loadDenot(ctx.withPhase(ctx.phaseId - 1)).current
-        if (d.exists) d
-        else throw new Error(s"failure to reload $this of class $getClass")
-      }
+    /** Is this a reference to a class or object member? */
```
smarter: This comment does not seem to match the implementation: `infoDependsOnPrefix` calls `membersNeedAsSeenFrom(prefix)` which will return false if `prefix` is a statically accessible object.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1924,15 +1895,54 @@ object Types {
...
-    def withNameSpace(nameSpace: NameSpace)(implicit ctx: Context): NamedType =
-      if (nameSpace == this.nameSpace) this
-      else NamedType(prefix, designator.withNameSpace(nameSpace))
+    /** A reference like this one, but with the given denotation, if it exists */
+    final def withDenot(denot: Denotation)(implicit ctx: Context): ThisType =
+      if (denot.exists) {
+        val adapted = withSym(denot.symbol)
+        if (adapted ne this) adapted.withDenot(denot).asInstanceOf[ThisType]
+        else {
+          setDenot(denot)
```
smarter: Given the name of this method and its documentation, it's surprising that it can mutate `this`.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1945,18 +1955,18 @@ object Types {
...
+  /** A reference to an implicit definition. This can be either a TermRef or a
+   *  Implicits.RenamedImplicitRef.
+   */
+  trait ImplicitRef {
+    def implicitName(implicit ctx: Context): TermName
+    def underlyingRef: TermRef
   }
 
-  abstract case class TermRef(override val prefix: Type, designator: TermDesignator) extends NamedType with SingletonType {
+  abstract case class TermRef(override val prefix: Type, var designator: Designator)
```
smarter: I see `var designator` here, doesn't that make the setter of `TermRef#designator` public?

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -444,34 +447,31 @@ object Symbols {
...
 
+    /** Is symbol valid in current run? */
     final def isValidInCurrentRun(implicit ctx: Context): Boolean =
-      lastDenot.validFor.runId == ctx.runId || ctx.stillValid(lastDenot)
-
-    /** Designator overrides */
-    final override def isTerm(implicit ctx: Context): Boolean =
+      (lastDenot.validFor.runId == ctx.runId || ctx.stillValid(lastDenot)) &&
+      (lastDenot.symbol eq this)
+        // the last condition is needed because under option -Yupdate-stale overwritten
```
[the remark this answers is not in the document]
odersky: No that would not work. The condition is _redundant_ unless stale symbols are accepted.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1593,109 +1629,120 @@ object Types {
...
-          lastDenotation = d
-          lastSymbol = d.symbol
-          checkedPeriod = ctx.period
-          checkDenot()
-        }
+          setDenot(d)
         d
       }
+
+      def fromDesignator = designator match {
```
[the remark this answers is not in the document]
odersky: True. I left it like this for better debuggability. If checkDenot (called from setDenot) fails with an AssertionError, we know exactly how we got there.

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: Good point. I'll update the docs.

PR #3497, opened 2017-11-17
Avoid boxing FlagSets in union and allOf

nicolasstucki:
When bootstrapping dotty, this reduces the allocated `FlagSets` from 2.7M to 400K.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Flags.scala
+++ b/compiler/src/dotty/tools/dotc/core/Flags.scala
@@ -172,12 +172,24 @@ object Flags {
   }
 
   /** The union of all flags in given flag set */
-  def union(flagss: FlagSet*) = (EmptyFlags /: flagss)(_ | _)
+  def union(flagss: FlagSet*): FlagSet = {
+    var flag = EmptyFlags
+    for (f <- flagss)
+      flag |= f
+    flag
+  }
 
   /** The conjunction of all flags in given flag set */
-  def allOf(flagss: FlagSet*) = {
-    assert(flagss forall (_.numFlags == 1), "Flags.allOf doesn't support flag " + flagss.find(_.numFlags != 1))
-    FlagConjunction(union(flagss: _*).bits)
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -172,12 +172,24 @@ object Flags {
   }
 
   /** The union of all flags in given flag set */
-  def union(flagss: FlagSet*) = (EmptyFlags /: flagss)(_ | _)
+  def union(flagss: FlagSet*): FlagSet = {
+    var flag = EmptyFlags
+    for (f <- flagss)
```
smarter: If you want to go deep in micro-optimizations, you could use a while loop

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -172,12 +172,24 @@ object Flags {
   }
 
   /** The union of all flags in given flag set */
-  def union(flagss: FlagSet*) = (EmptyFlags /: flagss)(_ | _)
+  def union(flagss: FlagSet*): FlagSet = {
```
smarter: This is still going to box, isn't it? `FlagSet*` becomes `Seq[FlagSet]`, value classes in generic positions are always boxed.

compiler/src/dotty/tools/dotc/core/Flags.scala
nicolasstucki: Should avoid `FlagSet*` altogether.

compiler/src/dotty/tools/dotc/core/Flags.scala
nicolasstucki: We avoid reboxing in the fold, half of the boxes are gone.

nicolasstucki:
test performance please

nicolasstucki:
I do not believe the performance should change much

PR #3501, opened 2017-11-17
Add an error message for "invalid unapply return type" error.

markarasev:
Part of #1589.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -116,6 +116,7 @@ public enum ErrorMessageID {
     TraitRedefinedFinalMethodFromAnyRefID,
     PackageNameAlreadyDefinedID,
     UnapplyInvalidNumberOfArgumentsID,
+    UnapplyInvalidReturnTypeID,
     StaticFieldsOnlyAllowedInObjectsID,
     CyclicInheritanceID,
     BadSymbolicReferenceID,
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -28,7 +28,7 @@ import TypeApplications._
 import reporting.diagnostic.Message
 import reporting.trace
 import Constants.{Constant, IntTag, LongTag}
-import dotty.tools.dotc.reporting.diagnostic.messages.{NotAnExtractor, UnapplyInvalidNumberOfArguments}
+import dotty.tools.dotc.reporting.diagnostic.messages.{UnapplyInvalidReturnType, NotAnExtractor, UnapplyInvalidNumberOfArguments}
 import Denotations.SingleDenotation
 import annotation.constructorOnly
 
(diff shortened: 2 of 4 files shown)


maseev:
> Use the imperative mood in the subject line ("Added" instead of "Add")

I thought I already [fixed](https://github.com/lampepfl/dotty/commit/1278611b7894e0185bf0dfc114f2ce6dab4c9f5e) this. It looks like the dotty-bot hasn't been rebuilt since then :worried:

markarasev:
I fixed the commit message =)

senia-psm:
@markarasev This new feature is called [Name Based Pattern](http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html#name-based-pattern). Please note there there is also a [Product Pattern](http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html#product-pattern).

markarasev:
@senia-psm I updated the explanation to cover these cases.

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
```diff
@@ -113,6 +113,7 @@
     IllegalStartOfStatementID,
     TraitIsExpectedID,
     TraitRedefinedFinalMethodFromAnyRefID,
+    InvalidUnapplyReturnTypeID,
```
gosubpl: Or `UnapplyInvalidReturnTypeID` ? I am now struggling with this for the other one, and decided that `UnapplyInvalidNumberOfArgumentsID` sounds better than `InvalidUnapplyNumberOfArgumentsID`. WDYT?

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -92,7 +93,7 @@ object Applications {
     def getTp = extractorMemberType(unapplyResult, nme.get, pos)
 
     def fail = {
-      ctx.error(i"$unapplyResult is not a valid result type of an $unapplyName method of an extractor", pos)
+      ctx.error(InvalidUnapplyReturnType(unapplyResult, unapplyName))
```
gosubpl: How is the `pos` being passed here? (sorry for the probably naive question, I am just starting on those error messages)

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
markarasev: OK for me, I updated the names.

compiler/src/dotty/tools/dotc/typer/Applications.scala
markarasev: I actually misunderstood this value's usage, fixed this.

odersky:
In Dotty, Product subclasses are also legal Unapply result types. So this specific PR is invalid. It would be good to 

 - update the "Changes" section in the references to reflect that, if that is not already the case
 - check whether the error message for types that are still illegal is correct, and update it if necessary.

markarasev:
@odersky do you mean that this part of the explanation is not explicit enough about Product subclasses?
> To be used as an extractor, an unapply method has to return a type that either:
> - [...]
> - is a Product (like a "Tuple2[T1, T2]")

By "changes section in the references" do you mean these docs http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html ? They seem to be up to date.

For your last point, do you mean I should add more unit tests?

odersky:
@OlivierBlanvillain This has been inactive for a year. Let's merge or close this now.

PR #3504, opened 2017-11-19
Rework denotations

odersky:
Make PreDenotation a superclass of Denotation. The immediate need
for this is that we would like to add an `asSeenFrom` to Denotations,
but if we do that it clashes with the `asSeenFrom` in Predenotations.

This is a part of the NamedTypes rework. Submitted independently so that we can test
performance in isolation.


odersky:
test performance please

odersky:
Seems it's a win, performance-wise.

PR #3508, opened 2017-11-20
Add error message for wrong number of argument patterns

gosubpl:
Refs #1589

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1929,4 +1929,17 @@ object messages {
     val explanation =
       hl"An ${"object"} cannot have the same name as an existing ${"package"}. Rename either one of them."
   }
+
+  case class UnapplyInvalidNumberOfArguments(qual: untpd.Tree, argTypes: List[Type])(implicit ctx: Context)
+    extends Message(UnapplyInvalidNumberOfArgumentsID) {
+    val kind = "Syntax"
+    val msg = hl"Wrong number of argument patterns for $qual; expected: ($argTypes%, %)"
+    val explanation =
+      hl"""The Unapply method of $qual was used with incorrect number of arguments.
+          |Expected usage would be something like:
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1929,4 +1929,16 @@ object messages {
...
+  case class UnapplyInvalidNumberOfArguments(qual: untpd.Tree, argTypes: List[Type])(implicit ctx: Context)
+    extends Message(UnapplyInvalidNumberOfArgumentsID) {
+    val kind = "Syntax"
+    val msg = hl"Wrong number of argument patterns for $qual; expected: ($argTypes%, %)"
+    val explanation =
+      hl"""The Unapply method of $qual was used with incorrect number of arguments.
+          |Expected usage would be something like:
+          |case $qual(${argTypes.map(_ => '_')}%, %) => ...
+          |
+        |where subsequent arguments would have following types: ($argTypes%, %).""".stripMargin
```
nicolasstucki: `""".stripMargin` should be on the next line.

gosubpl:
@nicolasstucki thanks for such a quick review! I have reflected your comments.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
allanrenucci: The issue with `stripMargin` being on the next line is that it adds an extra line to the explanation which is not desired.

gosubpl: @nicolasstucki @allanrenucci can you please decide which option is recommended? I will gladly oblige.
Also - whoever does the merge can fix it as I have checked 'Allow edits from maintainers'.

nicolasstucki: Use Allan's.

gosubpl:
Many thanks!

PR #3510, opened 2017-11-20
A simple way to check exhaustivity of GADTs

liufengyun:
This is an alternative to #3454.

The idea (implemented in OCaml, briefly mentioned in the `space` paper) is simple: just exclude counterexamples that are not satisfiable.

The PR passed all GADT tests, and it should have better performance by only enabling the checking for `invariant` GADTs. It doesn't do any changes to the core algorithm.

Many thanks to @AleksanderBG for his work on checking GADTs and pushing the limits of exhaustivity check for Scala.

cc/ @AleksanderBG: we're still interested in your algorithm , it would be nice to have detailed evaluation of the two approaches when #3454 is done.


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -708,6 +708,41 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
 
+  /** Whether the counterexample is satisfiable. The space is flattened and non-empty. */
+  def satisfiable(sp: Space): Boolean = {
+    def genConstraint(space: Space): List[(Type, Type)] = space match {
+      case Prod(tp, unappTp, unappSym, ss, _) =>
+        val tps = signature(unappTp, unappSym, ss.length)
+        ss.zip(tps).flatMap {
+          case (sp : Prod, tp) => sp.tp -> tp :: genConstraint(sp)
+          case (Typ(tp1, _), tp2) => tp1 -> tp2 :: Nil
+          case _ => ???  // impossible
```
OlivierBlanvillain: I feel that we should use something other than `???` for unreachable cases, to me `???` really has the semantic of a TODO: "implementation is missing here, to be added later"

Maybe something like this? Not sure where should that definition could go...

```scala
def impossible: Nothing = throw new AssertionError("This expression shouldn't be reachable.")
```

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -775,6 +810,15 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     res
   }
 
+  def checkGADT(tp: Type): Boolean = {
```
OlivierBlanvillain: I'm not sure what you mean by `checkGADT` the implementation looks like it's computing `isDeeplyInvariant`

AleksanderBG:
@liufengyun Is it possible to use `ctx.gadt` to further restrict initial tvar bounds? There are cases like the following one:
```scala
object O {
  sealed trait NumExpr[T]
  case class IntLit(v: Int) extends NumExpr[Int]
  case class DoubleLit(v: Double) extends NumExpr[Double]
  case class Mul[T](n1: NumExpr[T], n2: NumExpr[T]) extends NumExpr[T]

  def simplify[T](e: NumExpr[T]): NumExpr[T] = e match {
    case Mul(IntLit(i), ie) => ie match {
      case IntLit(0) => IntLit(0)
      case IntLit(1) => IntLit(i)
      case IntLit(_) | Mul(_, _) => e
    }
    case _ => e
  }
}
```
where the typechecker already knows that `T ~ Int` from the outer match. If I understand correctly that these bounds are recorded in `ctx.gadt`, using it would improve the warnings while barely requiring any adjustments.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
allanrenucci: `assert(false, "Unreachable code")`?

OlivierBlanvillain: @allanrenucci assert has type Unit

allanrenucci: My bad...

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: Yes, something like `shouldCheckGADT` with a comment is more friendly.

liufengyun:
@AleksanderBG Good observation. The `ctx.gadt` information is only available during type checking (phase `Frontend`), in pattern matcher it's too late to get that information.

PR #3511, opened 2017-11-20
RemoveUnnecessaryNullChecks improvements

gan74:
Rewrite of RemoveUnnecessaryNullChecks:

* Remove redundant checks
* More consistently remove checks on values that can not be null (such as `this` and `new`)
* Remove checks after statements that would have thrown a NPE

Added relevant tests.


compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,118 @@ import core.Contexts.Context
...
+  // return nullchecked symbol in expr (if any)
+  private def nullChecksInExpr(expr: Tree)(implicit ctx: Context): Option[Symbol] =
+    expr match {
+      case t @ Apply(Select(id: Ident, op), List(rhs)) if (t.symbol == defn.Object_eq || op == nme.EQ) && !isVar(id.symbol) && isNull(rhs) => Some(id.symbol)
+      case t @ Apply(Select(lhs, op), List(id: Ident)) if (t.symbol == defn.Object_eq || op == nme.EQ) && !isVar(id.symbol) && isNull(lhs) => Some(id.symbol)
+      case _ => None
+    }
 
+  // return all sub-trees of tree containing a nullcheck for symbol
+  private def findNullChecksFor(symbol: Symbol, tree: Tree)(implicit ctx: Context) =
```
OlivierBlanvillain: Could you also add return types on private methods?

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,118 @@ import core.Contexts.Context
...
+      case t @ Apply(Select(lhs, op), List(id: Ident)) if (t.symbol == defn.Object_eq || op == nme.EQ) && !isVar(id.symbol) && isNull(lhs) => Some(id.symbol)
+      case _ => None
+    }
 
+  // return all sub-trees of tree containing a nullcheck for symbol
+  private def findNullChecksFor(symbol: Symbol, tree: Tree)(implicit ctx: Context) = 
+    tree.filterSubTrees(nullChecksInExpr(_) match { case Some(s) if (s == symbol) => true; case _ => false })
+
+  // fill replace using findNullChecksFor
+  private def flagAsNull(symbol: Symbol, tree: Tree)(implicit ctx: Context) = replace ++= findNullChecksFor(symbol, tree).map((_, Literal(Constant(true))))
```
OlivierBlanvillain: You could maybe assert here than the result of this `findNullChecksFor` is non empty

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,118 @@ import core.Contexts.Context
...
+  private def flagAsNotNull(symbol: Symbol, tree: Tree)(implicit ctx: Context) = replace ++= findNullChecksFor(symbol, tree).map((_, Literal(Constant(false))))
+
+  private def isNotNull(tree: Tree)(implicit ctx: Context): Boolean =
```
OlivierBlanvillain: It's a detail but I find the names are slightly misleading. I would call these `isNeverNull` and `isAlwaysNull`, otherwise it sounds like either `isNotNull(t)` or `isNull(t)` would be true for any `t`.

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,118 @@ import core.Contexts.Context
...
-          else if (sym == defn.Object_ne) block(true)
-          else check
-        } else check
-
-      case t => t
-    }
-    transformation
-  }
+
+  private def isVar(s: Symbol)(implicit ctx: Context) = s.is(Mutable, Lazy) && !s.is(Method) && !s.owner.isClass
```
OlivierBlanvillain: Why do you need anything more than `.is(Mutable)` here? Also I think this version of `is` was intended to always be used with named second argument: `s.is(Mutable, butNot = Lazy)`.

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,118 @@ import core.Contexts.Context
...
+          cond match {
+            case cond @ Select(expr, _) if cond.symbol == defn.Boolean_! =>
+              visitBranches(expr, el, th) 
+            case cond @ Apply(Select(lhs, _), List(rhs)) if cond.symbol == defn.Boolean_&& =>
+              visitBranches(lhs, th, el)
+              visitBranches(rhs, th, el)
+            case expr => 
+              nullChecksInExpr(expr) match {
+                case Some(symbol) => 
+                  flagAsNull(symbol, th)
```
OlivierBlanvillain: I'm missing something here. I'm I correct saying than `flagAsNull` will flag `Trees` as null/notNull? If so, where do you keep track that this nullness is only valid within this specific branch of the if?

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,120 @@ import core.Contexts.Context
...
-      case _ => false
+  private def isNeverNull(tree: Tree)(implicit ctx: Context): Boolean =
+    tree match {
+      case Block(_, expr) => isNeverNull(expr)
+      case If(_, th, el) => isNeverNull(th) && isNeverNull(el)
+      case t: Typed => isNeverNull(t.expr)
+      case t: Literal => t.const.tag != NullTag
+      case _: This => true
+      case _: New => true
+      case t: Apply if t.symbol.isPrimaryConstructor => true
```
OlivierBlanvillain: why `Primary`?

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
```diff
@@ -6,104 +6,120 @@ import core.Contexts.Context
...
 
-          case t: Apply if t.symbol.isPrimaryConstructor => true
-
-          case t: Literal => t.const.value != null
+  def transformer(implicit ctx: Context): Tree => Tree = {
+    // transform tree recursively replacing nullchecks for symbols in nullness
+    def transform(tree: Tree, nullness: mutable.Map[Symbol, Boolean]) =
+      if (nullness.isEmpty) tree
+      else new TreeMap() {
+          override def transform(tree: Tree)(implicit ctx: Context): Tree = {
```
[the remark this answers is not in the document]
gan74: You are right. However nullness isn't static, even for `val`, it can change within a block. There is no way to evaluate if a value is null without being aware of what code has been executed before. 

I wasn't able to get the information I needed using the `transformer` so I resorted to explore the whole tree manually, from top to bottom. 

This can be optimized to basically  `O(n) ` by running the transformer only on the top level block (and with a few other modifications). I am not aware of any way to do this however.

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
gan74: It's taken from [here](https://github.com/lampepfl/dotty/blob/9de3905ca9e287d42ef293025d0ff3369b7d8f12/compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala#L60).

compiler/src/dotty/tools/dotc/transform/localopt/RemoveUnnecessaryNullChecks.scala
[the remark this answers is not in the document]
OlivierBlanvillain: This point needs to be addressed one way or another before merging.

Maybe a solution to prevent the quadratic behavior would be to cut the inner traversal to not got into nested If & Block (or accept a bounded level of nesting).

nicolasstucki:
test performance please

nicolasstucki:
The local optimizations where removed. This PR is no longer needed.

PR #3514, opened 2017-11-20
Fix build failure after last merge

smarter:
tasty got renamed to fromTasty in
331245e444752eed1595c09e9fa3bda000c84def, but tasty was used in
51bc54a681ba243b545e72b114899562f60847c4 which was merged at the same
time.


smarter:
@nicolasstucki That's the one which is fixed in this PR.

nicolasstucki:
My mistake, LGTM

nicolasstucki:
I was testing the same fix on my machine before this PR.

PR #3516, opened 2017-11-20
#1589: add dedicated error message for @static annotation not inside object

esarbe:


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1929,4 +1929,11 @@ object messages {
     val explanation =
       hl"An ${"object"} cannot have the same name as an existing ${"package"}. Rename either one of them."
   }
+
+  case class StaticFieldsOnlyAllowedInObjects(field: Symbol)(implicit ctx: Context) extends Message(StaticFieldsOnlyAllowedInObjectsID) {
+    val msg = "@static fields are only allowed inside objects"
```
allanrenucci: I would say:
```scala
hl"${"@static"} $member in ${member.owner} must be defined inside an ${"object"}."
```
And use
```scala
hl"${"@static"} members are only allowed inside objects."
```
for the explanation

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1929,4 +1929,11 @@ object messages {
     val explanation =
       hl"An ${"object"} cannot have the same name as an existing ${"package"}. Rename either one of them."
   }
+
+  case class StaticFieldsOnlyAllowedInObjects(field: Symbol)(implicit ctx: Context) extends Message(StaticFieldsOnlyAllowedInObjectsID) {
```
allanrenucci: Substitute field by member. It is not necessarily a field. Can be a class, a val...

esarbe:
should be better now

allanrenucci:
Thanks @esarbe! 🎉

PR #3517, opened 2017-11-20
Fix #3515: Do not save tasty in .class with -YemitTasty

### LINKED ISSUE #3515: Don't emit a TASTY section in classfiles when we're also emitting .tasty files
`-YemitTasty` can be used to generate .tasty files, but when we do this we still store TASTY in classfiles, this is unnecessary duplication.

nicolasstucki:
Only emit .tasty file

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
+++ b/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
@@ -13,7 +13,7 @@ import scala.collection.{ mutable, immutable }
 import scala.collection.mutable.{ ListBuffer, ArrayBuffer }
 import scala.annotation.switch
 import typer.Checking.checkNonCyclic
-import io.AbstractFile
+import io.{AbstractFile, PlainFile}
 import scala.util.control.NonFatal
 
 object ClassfileParser {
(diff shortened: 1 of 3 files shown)


nicolasstucki:
We are currently not loading tasty from the `.tasty` files. Will need to add this functionality in this PR.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -781,7 +781,15 @@ class ClassfileParser(
 
       if (scan(tpnme.TASTYATTR)) {
         val attrLen = in.nextInt
-        return unpickleTASTY(in.nextBytes(attrLen))
+        if (attrLen == 0) {
+          // A tasty attribute implies the existence of the .tasty file
+          val tastyFilePath =
+            classfile.jpath.resolveSibling(classfile.jpath.getFileName.toString.dropRight("class".length) + "tasty")
```
allanrenucci:
```scala
io.File(classfile.jpath).changeExtension("class")
```

nicolasstucki: Thanks

nicolasstucki:
We have them in the idempotency tests, that is why ed61847 failed the tests.

nicolasstucki:
I will create one test that will execute in local tests

PR #3519, opened 2017-11-21
Add -print-tasty setting

nicolasstucki:


compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -43,6 +43,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val rewrite = OptionSetting[Rewrites]("-rewrite", "When used in conjunction with -language:Scala2 rewrites sources to migrate to new syntax")
   val silentWarnings = BooleanSetting("-nowarn", "Silence all warnings.")
   val fromTasty = BooleanSetting("-from-tasty", "Compile classes from tasty in classpath. The arguments are used as class names.")
+  val printTasty = BooleanSetting("-print-tasty", "Prints the raw tasty when decompiling.")
```
liufengyun: Nit-picky: Should we add that `-print-tasty` can only be used together with `-fromTasty`? Or, as a blank before L45, to put tasty related options together.

nicolasstucki: It must not necessarily be the case. In the future we might use it to print the pickled tasty while compiling without `-from-tasty`.

nicolasstucki: Currently it can only be used with `-decompile` which is stated. `-from-tasty` has nothing to do with it.

PR #3522, opened 2017-11-21
Fix #3518: Erase explicit Unit.box and Unit.unbox in erasure

### LINKED ISSUE #3518: `Unit.box(())` crashes backend
`Unit.box(())` crashes backend:

```Scala
class Test {
  Unit.box(())
}
```

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -479,6 +479,7 @@ object Erasure {
       val Apply(fun, args) = tree
       if (fun.symbol == defn.cbnArg)
         typedUnadapted(args.head, pt)
+      else if ((tree.symbol.name == nme.box || tree.symbol.name == nme.unbox) && tree.symbol.owner.asClass.companionClass == defn.UnitClass) ref(defn.BoxedUnit_UNIT)
```
sjrd: That drops the side effects of `args.head`, unless I'm mistaken.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -477,7 +477,7 @@ object Erasure {
 	 */
     override def typedApply(tree: untpd.Apply, pt: Type)(implicit ctx: Context): Tree = {
       val Apply(fun, args) = tree
-      if (fun.symbol == defn.cbnArg)
+      if (fun.symbol == defn.cbnArg || (tree.symbol.name == nme.box || tree.symbol.name == nme.unbox) && tree.symbol.owner.asClass.companionClass == defn.UnitClass)
```
allanrenucci: I would factor out the newly added tests into an inner function, to make it clear what is tested: `def isUnitBoxOrUnbox(fun: Symbol)`

nicolasstucki:
Fixes are in backend https://github.com/lampepfl/scala/pull/30

nicolasstucki:
scalac also ignores `Unit` when defining `isBox` and `isUnbox` 

https://github.com/scala/scala/blob/07d61ecf134dbba143531f5a1bd3c1289c437296/src/reflect/scala/reflect/internal/Definitions.scala#L84 

https://github.com/scala/scala/blob/07d61ecf134dbba143531f5a1bd3c1289c437296/src/reflect/scala/reflect/internal/Definitions.scala#L1470-L1473

nicolasstucki:
@smarter now it does the same as in `scalac`

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -169,8 +169,10 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     s eq defn.newArrayMethod
   }
 
-  def isBox(sym: Symbol): Boolean = Erasure.Boxing.isBox(sym)
-  def isUnbox(sym: Symbol): Boolean = Erasure.Boxing.isUnbox(sym)
+  def isBox(sym: Symbol): Boolean = Erasure.Boxing.isBox(sym) && !isInUnit(sym)
```
smarter: Why not change the definition of Erasure.Boxing.isBox directly?

nicolasstucki: Erasure does transformations on Unit box inbox based on Boxing.isBox. If I add that condition directly there erasure starts failing.

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -169,8 +169,10 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
...
+  def isUnbox(sym: Symbol): Boolean = Erasure.Boxing.isUnbox(sym) && !isInUnit(sym)
+
+  private def isInUnit(sym: Symbol) = sym.denot.owner.denot.linkedClass == defn.UnitClass
```
smarter: I would add a defn.UnitModule instead.

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
smarter: ok

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
nicolasstucki: Done

PR #3525, opened 2017-11-21
Give position to symbols when unpickling

smarter:
I started working on this to fix a crash when compiling the collection-strawman contrib tests. I minimized it to `anonBridge.scala` (in the collection-strawman, unpickling happened because of `inline def`, here the test is in `from-tasty` so it will be compiled from an unpickled tree). Eventually, I would also like to use the unpickled symbol positions in the IDE, but this will require more work. (Right now, going from an arbitrary Position to a SourcePosition is annoying, it requires creating a SourceFile which requires converting the source file content to an array of bytes).

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
+++ b/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
@@ -96,6 +96,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val YdetailedStats = BooleanSetting("-Ydetailed-stats", "show detailed internal compiler stats (needs Stats.enabled to be set to true).")
   val Yheartbeat = BooleanSetting("-Ydetailed-stats", "show heartbeat stack trace of compiler operations (needs Stats.enabled to be set to true).")
   val YprintPos = BooleanSetting("-Yprint-pos", "show tree positions.")
+  val YprintPosSyms = BooleanSetting("-Yprint-pos-syms", "show symbol definitions positions.")
   val YnoDeepSubtypes = BooleanSetting("-Yno-deep-subtypes", "throw an exception on deep subtyping call stacks.")
   val YnoPatmatOpt = BooleanSetting("-Yno-patmat-opt", "disable all pattern matching optimizations.")
   val YplainPrinter = BooleanSetting("-Yplain-printer", "Pretty-print using a plain printer.")
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -288,7 +288,7 @@ trait Symbols { this: Context =>
     val tparamBuf = new mutable.ListBuffer[TypeSymbol]
     val trefBuf = new mutable.ListBuffer[TypeRef]
     for (name <- names) {
-      val tparam = newNakedSymbol[TypeName](NoCoord)
+      val tparam = newNakedSymbol[TypeName](owner.coord)
       tparamBuf += tparam
       trefBuf += TypeRef(owner.thisType, tparam)
     }
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -443,10 +443,11 @@ object Symbols {
       if (lastDenot == null) NoRunId else lastDenot.validFor.runId
 
     /** Does this symbol come from a currently compiled source file? */
-    final def isDefinedInCurrentRun(implicit ctx: Context): Boolean = {
-      // FIXME: broken now now that symbols unpickled from TASTY have a position
-      pos.exists && defRunId == ctx.runId
-    }
+    final def isDefinedInCurrentRun(implicit ctx: Context): Boolean =
+      pos.exists && defRunId == ctx.runId && {
+        val file = associatedFile
+        file != null && ctx.runInfo.files.contains(file)
+      }
 
     /** Is symbol valid in current run? */
     final def isValidInCurrentRun(implicit ctx: Context): Boolean =
@@ -540,7 +541,7 @@ object Symbols {
      *  Overridden in ClassSymbol
      */
     def associatedFile(implicit ctx: Context): AbstractFile =
-      denot.topLevelClass.symbol.associatedFile
+      if (lastDenot == null) null else lastDenot.topLevelClass.symbol.associatedFile
 
     /** The class file from which this class was generated, null if not applicable. */
     final def binaryFile(implicit ctx: Context): AbstractFile = {
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -130,8 +130,8 @@ trait Symbols { this: Context =>
     newClassSymbol(owner, name, flags, completer, privateWithin, coord, assocFile)
   }
 
-  def newRefinedClassSymbol = newCompleteClassSymbol(
-    ctx.owner, tpnme.REFINE_CLASS, NonMember, parents = Nil)
+  def newRefinedClassSymbol(coord: Coord = NoCoord) =
+    newCompleteClassSymbol(ctx.owner, tpnme.REFINE_CLASS, NonMember, parents = Nil, coord = coord)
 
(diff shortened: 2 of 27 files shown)


smarter:
test performance please

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -93,6 +93,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val YdetailedStats = BooleanSetting("-Ydetailed-stats", "show detailed internal compiler stats (needs Stats.enabled to be set to true).")
   val Yheartbeat = BooleanSetting("-Ydetailed-stats", "show heartbeat stack trace of compiler operations (needs Stats.enabled to be set to true).")
   val Yprintpos = BooleanSetting("-Yprintpos", "show tree positions.")
+  val YprintposSyms = BooleanSetting("-Yprintpos-syms", "show symbol definitions positions.")
```
odersky: Since `-Yprintpos` is about to change to `Yprint-pos`, I guess we also need an additional hypen here: `-Yprint-pos-syms`.

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -437,7 +437,7 @@ object Symbols {
 
     /** Does this symbol come from a currently compiled source file? */
     final def isDefinedInCurrentRun(implicit ctx: Context): Boolean = {
-      pos.exists && defRunId == ctx.runId
+      pos.exists && defRunId == ctx.runId && sourceFile == ctx.source.file
```
odersky: That's not the right condition anymore. See for instance how it is used in `checkNoConflict`. We want to flag top-level definitions in other source files if they clash with a definition in the current source file.

odersky: We could probably use as third criterion:

     { val file = assocatedFile
       file != null && !file.path.endsWith(".class")
     }

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1155,7 +1155,8 @@ object Parsers {
         t
     }
 
-    def ascription(t: Tree, location: Location.Value) = atPos(startOffset(t), in.skipToken()) {
+    def ascription(t: Tree, location: Location.Value) = atPos(startOffset(t)) {
+      in.skipToken()
```
odersky: Why the change? The previous implementation would place the `^` under the `:` of an ascription, which seems to be what we want, no?

compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
```diff
@@ -90,8 +90,8 @@ class ExpandPrivate extends MiniPhase with IdentityDenotTransformer { thisPhase
       }
 
       assert(d.symbol.sourceFile != null &&
-             isSimilar(d.symbol.sourceFile.path, ctx.source.file.path),
-          s"private ${d.symbol.showLocated} in ${d.symbol.sourceFile} accessed from ${ctx.owner.showLocated} in ${ctx.source.file}")
+             isSimilar(d.symbol.sourceFile.path, ctx.owner.sourceFile.path),
+          s"private ${d.symbol.showLocated} in ${d.symbol.sourceFile} accessed from ${ctx.owner.showLocated} in ${ctx.owner.sourceFile}")
```
odersky: When would this make a difference?

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
smarter: Sure, in which PR is Yprintpos changed to Yprint-pos ?

compiler/src/dotty/tools/dotc/core/Symbols.scala
smarter: Hmm that's tricky because associatedFile forces things, I'll give it a try.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
smarter: In all other cases, the point is under the first letter of the name of the definition (e.g. for `def foo: Int = 1` the point is under `f`). This change means that I can always use the point of the definition tree to set the symbol position.

compiler/src/dotty/tools/dotc/transform/ExpandPrivate.scala
smarter: When unpickling from tasty, `ctx.source.file` will be "Foo.class" whereas `ctx.owner.sourceFile` will be the source file name found in the `SourceFile` annotation of the tasty tree.

smarter:
@odersky There's another difficulty with `isDefinedInCurrentRun` If we're compiling from TASTY then `associatedFile` may very well end with `.class` but `isDefinedInCurrentRun` should return true. The only solution I see here is doing a check like `ctx.runInfo.units.contains(associatedFile)`, but that requires adding `units` to `RunInfo` (currently it's only in `Run` which isn't accessible from `Context`), what do you think?

smarter:
Alternative proposal to make `isDefinedInCurrentRun` easy to define: track whether a symbol comes from a compilation unit when the symbol is created. This could be done by giving a special meaning to one bit of the `Coord` class

smarter:
test performance please

smarter:
@odersky Could you review the last commit please?

smarter:
Ping for review of the last commit

odersky:
I am worried that this seems to cause a slowdown of 6% for Dotty itself. What could be the cause of this?

odersky:
Last commit LGTM, but it would be good to drill down on the performance issue.

smarter:
test performance please

odersky:
I am still not convinced this is the right thing to do. We add a lot more positions to Tasty files which makes them larger. Symbols get larger too. It seems this has a significant performance implications. So what exactly are we trying to solve here? Can't this be obtained more cheaply? 

If it's just the inlining, it would be easy to fix with an additional pass when we create the Inline Body annotation.

smarter:
> We add a lot more positions to Tasty files which makes them larger.

Not at all! The positions of symbols are derived from the position of trees, so we're not storing anything more.

>  Symbols get larger too.

I don't see how, symbols already had a `coord` field. It's just that for many of them it was set to `NoCoord` previously.

>  It seems this has a significant performance implications.

I think the benchmark result was a fluke, we'll see. If there are performance issues I'll fix them.

> So what exactly are we trying to solve here?

Unpickled trees should be as close to their pre-pickling state as possible. This reduces the number of surprises when compiling from tasty since we're using the same code paths than when compiling from sources. Positions for symbols may also allow for better error messages, and could be used in the IDE to reduce the number of tree traversals done for operations such as jump-to-definition (the current PR doesn't do this because it's not trivial).

smarter:
Hmm it looks like we have no new benchmarks recorded for master since 20/11 (see http://dotty-bench.epfl.ch), so it's difficult to measure the effect of this PR. Ping @liufengyun

smarter:
test performance please: f92f278 6d5c226

smarter:
Oops, used the wrong syntax to test multiple commits, trying again...

smarter:
test performance please 7848ef6 6b19932

smarter:
@liufengyun Looks like testing multiple commits doesn't work anymore, or did I get the syntax wrong?

liufengyun:
You need a comma after the please.

Le 6 déc. 2017 19:43, "Guillaume Martres" <notifications@github.com> a
écrit :

odersky:
OK, I see the rationale now, and it seems performance implications are smaller than first feared. So I am OK to get this in. But maybe finish the performance evaluation first.

smarter:
test performance please; 18767e0d 6b19932

liufengyun:
@smarter Sorry, it's colon instead semicolon, I've a little problem with the punctuation names.

smarter:
I tried using colon in https://github.com/lampepfl/dotty/pull/3525#issuecomment-349654481 but it didn't seem to work either ?

liufengyun:
Let's try a new curve and see:

retest performance please: 18767e0d 6b19932

liufengyun:
retest performance please: 18767e0d 6b19932

PR #3527, opened 2017-11-22
Fix position of desugared tree wrapped in parentheses

smarter:
Before this commit, the added testcase produced the following result:
```scala
scala> while ((( foo ))) {}
1 |while ((( foo ))) {}
  |      ^^^^^^^^^^^
  |      not found: foo
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1017,7 +1017,17 @@ object desugar {
     }
 
     // begin desugar
+
+    // Special case for `Parens` desugaring: unlike all the desugarings below,
+    // its output is not a new tree but an existing one whose position should
+    // be preserved, so we shouldn't call `withPos` on it.
     tree match {
+      case Parens(t) =>
+        return t
+      case _ =>
+    }
+
+    val desugared = tree match {
       case SymbolLit(str) =>
         Apply(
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1017,7 +1017,17 @@ object desugar {
...
+    // Special case for `Parens` desugaring: unlike all the desugarings below,
+    // its output is not a new tree but an existing one whose position should
+    // be preserved, so we shouldn't call `withPos` on it.
     tree match {
+      case Parens(t) =>
+        return t
+      case _ =>
+    }
+
+    val desugared = tree match {
```
allanrenucci: You can remove the `Parens` case in this pattern match

compiler/test-resources/repl/errmsgs
```diff
@@ -64,3 +64,7 @@ scala> { def f: Int = g; val x: Int = 1; def g: Int = 5; }
 1 | { def f: Int = g; val x: Int = 1; def g: Int = 5; }
   |                ^
   |           `g` is a forward reference extending over the definition of `x`
+scala> while (((  foo ))) {}
+1 |while (((  foo ))) {}
```
allanrenucci: You need to shift the error message by one space after the `|`. This is a weird issue. I haven't figure it out yet

PR #3528, opened 2017-11-22
Port part of scala/scala#5293

allanrenucci:
Original commit from Irytz: https://github.com/scala/scala/commit/e619b033350a3378d650db4c3e5b1bfc83b73d81

Upgrade asm to 5.1

The constructor of scala.tools.asm.Handle now takes an additional
boolean parameter to denote whether the owner is an interface.


allanrenucci:
The rest of the PR should be ported to our backend fork

smarter:
The rest should already be handled by https://github.com/lampepfl/scala/pull/24

PR #3529, opened 2017-11-22
Add color to tasty printer

nicolasstucki:
Makes clear the distinction between name indices, tree indices and tree lengths.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
@@ -10,6 +10,7 @@ import TastyUnpickler._
 import TastyBuffer.{Addr, NameRef}
 import util.Positions.{Position, offsetToInt}
 import collection.mutable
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala
```diff
@@ -22,8 +23,8 @@ class TastyPrinter(bytes: Array[Byte])(implicit ctx: Context) {
 
   def printNames() =
     for ((name, idx) <- nameAtRef.contents.zipWithIndex) {
-      val index = "%4d: ".format(idx)
-      println(index + nameToString(name))
+      val index = Magenta("%4d".format(idx))
```
smarter: I would add a layer of abstraction, e.g. "NameColor" instead of "Magenta"

nicolasstucki: Will do.

PR #3534, opened 2017-11-22
-Xverify-signatures: emit errors, not warnings

smarter:
Otherwise the tests will still succeed.


smarter:
`-Xverify-signatures` is not enabled by default, so in what situation would people be blocked by this?

Duhemm:
🤦‍♂️ You're absolutely right

PR #3541, opened 2017-11-23
Fix all warnings and add -Xfatal-warnings

nicolasstucki:


nicolasstucki:
Needs https://github.com/lampepfl/scala/pull/28, which removes warnings in the backend

project/Build.scala
```diff
@@ -574,6 +574,11 @@ object Build {
       // Add git-hash used to package the distribution to the manifest to know it in runtime and report it in REPL
       packageOptions += ManifestAttributes(("Git-Hash", VersionUtil.gitHash)),
 
+      scalacOptions ++= Seq(
```
allanrenucci: Should this go into `commonSettings`?

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -722,23 +722,23 @@ object Symbols {
     def clear(): Unit = value.clear()
 
     def filter(p: ((Symbol, T)) => Boolean): Map[Symbol, T] = {
-      import scala.collection.JavaConversions._
-      value.toMap.filter(p)
+      import scala.collection.JavaConverters._
```
smarter: I would just move the import at the top of the file.

project/Build.scala
```diff
@@ -574,6 +574,11 @@ object Build {
...
+        "-Xfatal-warnings",
```
smarter: I would enable this on all projects but only on the CI, when doing local development warnings as errors is just annoying.

nicolasstucki: Good idea

project/Build.scala
nicolasstucki: Cannot do it due to #3543

nicolasstucki: We should push that for another PR

nicolasstucki:
Rebased on top of #3545 (last warning fix).

nicolasstucki:
@allanrenucci, I rebased it and fixed a new warning that came from some recently merged PR

allanrenucci:
@nicolasstucki Can you open an issue for the unchecked reference warning?

PR #3545, opened 2017-11-23
Fix #3543: compare extractor with TermRef instead of MethodType

### LINKED ISSUE #3543: Unreachable code warning for reachable match case
```scala
import scala.util.matching.Regex

object Test {
    foo("c" :: Nil, false)
  }

  def foo(remaining: List[String], inCodeBlock: Boolean): Unit = {
    remaining match {
      case CodeBlockEndRegex(before) :: ls =>
      case SymbolTagRegex(name) :: ls if !inCodeBlock => println("OK")
      case _ =>
    }
  }

  val CodeBlockEndRegex = new Regex("(b)")
  val SymbolTagRegex = new Regex("(c)")
}
```
Emits the folowing warining but should not.

```scala
11 |      case SymbolTagRegex(name) :: ls if !inCodeBlock => println("OK")
   |                                                      ^^^^^^^^^^^^^^^^
   |                                                      unreachable code
```

The code is correctly compiled and executed.

liufengyun:
Fix #3543: compare extractor with TermRef instead of MethodType

### DIFF
--- a/tests/patmat/3543.scala
+++ b/tests/patmat/3543.scala
@@ -9,7 +9,7 @@ class Test {
 
     xs match {
       case Yes(x) :: ls => println("Yes")
-      case No(y) :: ls => println("No") // unreachable code
+      case No(y) :: ls => println("No")
       case _ =>
     }
   }
@@ -26,7 +26,7 @@ class Test2 {
 
     xs match {
       case No() :: ls => println("No")
-      case Yes() :: ls => println("Yes") // unreachable code
+      case Yes() :: ls => println("Yes")
       case _ =>
     }
   }
--- a/tests/patmat/3543.scala
+++ b/tests/patmat/3543.scala
@@ -1,0 +1,52 @@
+class Test {
+  class Foo {
+    def unapply(x: String): Option[String] = ???
+  }
+
+  def test(xs: List[String]): Unit = {
+    val Yes = new Foo
+    val No = new Foo
+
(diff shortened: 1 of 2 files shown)


allanrenucci:
If there is no check file, the tests expect that the compiler emit not warning?

tests/patmat/3543.scala
```diff
@@ -0,0 +1,52 @@
...
+    def unapply(x: String): Option[String] = ???
+  }
+
+  def test(xs: List[String]): Unit = {
+    val Yes = new Foo
+    val No = new Foo
+
+    xs match {
+      case Yes(x) :: ls => println("Yes")
+      case No(y) :: ls => println("No") // unreachable code
```
allanrenucci: I would remove the comment

tests/patmat/3543.scala
```diff
@@ -0,0 +1,52 @@
...
+    def unapply(y: String): Boolean = x
+  }
+
+  def test(xs: List[String]): Unit = {
+    val Yes = new Foo(true)
+    val No = new Foo(false)
+
+    xs match {
+      case No() :: ls => println("No")
+      case Yes() :: ls => println("Yes") // unreachable code
```
allanrenucci: I would remove the comment

liufengyun:
@allanrenucci Yes, if there's no check file, there should be no warning.

PR #3550, opened 2017-11-24
Fix #3549: Load .tasty from jar if necessary

### LINKED ISSUE #3549: Cannot load .tasty files from jar
When trying to load a `.tasty` file fro a jar produced with `-YemitTasty`,  the `FileZipArchive#LeakyEntry` for the `.class` returns a `null` java path (`AbstractFile.jpath`) which ends with a NPE.

```
echo "class Foo" > Foo.scala
sbt "dotc -d foo.jar -YemitTasty Foo.scala"
sbt "dotc -decompile -classpath foo.jar Foo"
```
Putting an `assert(classfile.jpath ne null, classfile)` just before 
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala#L786 makes it explicit.

nicolasstucki:


compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -781,11 +781,32 @@ class ClassfileParser(
...
+        if (attrLen == 0) { // A tasty attribute implies the existence of the .tasty file
+          def readBytes(file: io.Path): Array[Byte] = {
+            val plainFile = new PlainFile(file)
+            if (plainFile.exists) new AbstractFileReader(plainFile).nextBytes(plainFile.sizeOption.get)
+            else {
+              ctx.error("Could not find " + file)
+              Array.empty
+            }
+          }
+          val tastyBytes = classfile match {
```
allanrenucci: I propose something along the line of:
```scala
val tastyBytes = classfile.underlyingSource match {
  case None =>
    readBytes(File(classfile.jpath).changeExtension("tasty"))
  case Some(jar) => // We are in a jar
    val jarArchive = JarArchive.open(File(jar.jpath))
    val tastyFile = File(jarFile.jpath).resolve(classfile.name).changeExtension("tasty")
    val bytes = readBytes(tastyFile)
    jarArchive.close()
    bytes
}
```
It's weird to pattern match on `FileZipArchive#LeakyEntry`. This is internal implementation detail, should probably be private.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -781,11 +781,32 @@ class ClassfileParser(
...
         val attrLen = in.nextInt
-        if (attrLen == 0) {
-          // A tasty attribute implies the existence of the .tasty file
-          val file = new PlainFile(io.File(classfile.jpath).changeExtension("tasty"))
-          if (file.exists) return unpickleTASTY(new AbstractFileReader(file).nextBytes(file.sizeOption.get))
-          else ctx.error("Could not find " + file)
+        if (attrLen == 0) { // A tasty attribute implies the existence of the .tasty file
+          def readBytes(file: io.Path): Array[Byte] = {
+            val plainFile = new PlainFile(file)
+            if (plainFile.exists) new AbstractFileReader(plainFile).nextBytes(plainFile.sizeOption.get)
```
allanrenucci: There is `toByteArray` defined on `AbstractFile`

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
allanrenucci: Also can you add a comment: `// TODO: simplify when #3552 is fixed`

allanrenucci: I have updated the code snippet: `underlyingSource` seems to return `None` only for virtual files. In the case there is no `underlyingSource`, we cannot do anything...

PR #3551, opened 2017-11-24
Fix #3548: set outer select type to be the type enclosing the expected type

### LINKED ISSUE #3548: ClassCastException during pattern match on inner classes defined in base traits
The following code crashes with a ClassCastException:

```scala
    trait Common {
      case class Data(a: String)
    }
    object O1 extends Common
    object O2 extends Common

    val data = O2.Data("test")

    // Runtime error: java.lang.ClassCastException: O2$ cannot be cast to O1$
    data match {
        case O1.Data(s) => println("O1 data")
        case O2.Data(s) => println("O2 data")
        case _ => println("Unknown")
    }
```

(Observed on 0.4.0-RC1)

liufengyun:
Fix #3548: set outer select type to be the type enclosing the expected type.

Otherwise, `ElimOuterSelect` will insert `ensureConforms(tp)` in outer test, which causes run-time exception.

### DIFF
--- a/tests/run/i3548.scala
+++ b/tests/run/i3548.scala
@@ -9,11 +9,12 @@ object Test {
 
     val data = O2.Data("test")
 
-    // Runtime error: java.lang.ClassCastException: O2$ cannot be cast to O1$
-    data match {
-        case O1.Data(s) => println("O1 data")
-        case O2.Data(s) => println("O2 data")
-        case _ => println("Unknown")
+    val result = data match {
+      case O1.Data(s) => 1
+      case O2.Data(s) => 2
+      case _ => 3
     }
+
+    assert(result == 2)
   }
 }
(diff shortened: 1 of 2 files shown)


tests/run/i3548.scala
```diff
@@ -0,0 +1,19 @@
...
+
+    val data = O2.Data("test")
+
+    // Runtime error: java.lang.ClassCastException: O2$ cannot be cast to O1$
+    data match {
+        case O1.Data(s) => println("O1 data")
+        case O2.Data(s) => println("O2 data")
+        case _ => println("Unknown")
+    }
+  }
```
allanrenucci: I propose a slight modification to the test:
```scala
val result = data match {
  case O1.Data(s) => 1
  case O2.Data(s) => 2
  case _ => 3
}
assert(result == 2)
```

odersky:
I had doubts at first, but after looking into it more deeply this appears to be the right fix. LGTM

PR #3556, opened 2017-11-26
#1589: Add error message for not emitting @switch.

vitorsvieira:
Regarding #1589, this PR addresses "could not emit switch" (PatternMatcher.911).

Noticed that `ctx.warning` does not propagate the error messages. 

@smarter Suggested to refactor `checkMessagesAfter` in a way that errors and warning messages are reported.

Would be something like:

```
  def checkMessagesAfter(checkAfterPhase: String)(source: String): Report = {
    ctx = newContext
    val runCtx = checkCompile(checkAfterPhase, source) { (_, _) => () }

    if (!runCtx.reporter.hasErrors && !runCtx.reporter.hasWarnings) new EmptyReport  // adding && !runCtx.reporter.hasWarnings
    else {
      val rep = runCtx.reporter.asInstanceOf[StoreReporter]
      val msgs = rep.removeBufferedMessages(runCtx).map(_.contained()).reverse
      new Report(msgs, runCtx)
    }
  }
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1972,4 +1972,36 @@ object messages {
     val msg = hl"Cannot extend ${"sealed"} $pclazz in a different source file"
     val explanation = "A sealed class or trait can only be extended in the same file as its declaration"
   }
+
+  case class UnableToEmitSwitch()(implicit ctx: Context)
+  extends Message(UnableToEmitSwitchID) {
+    val kind = "Syntax"
+    val msg = hl"""Could not emit switch for ${"@switch"} annotated match"""
+    val explanation = {
+      val errorCodeExample =
+        """val middle = 'm'
+          |val last   = 'z'
+          |
+          |val number   = (middle: @switch) match {
+          |  case 'a'  => 1
+          |  case 'm'  => 13
+          |  case last => 26  //a non-literal may prevent switch generation: this would not compile.
+          |  case _    => 0
+          |}""".stripMargin
+
+      hl"""The compiler verifies that the match has been compiled to a tableswitch or lookupswitch.
+          |An error will be issued in case it compiles into a series of conditional expressions.
+          |
+          |Consider the following example:
+          |
+          |$errorCodeExample
+          |
+          |The example above would fail as it matches a non-literal value.
+          |The compiler will apply the tableswitch optimization when:
+          |-The matched value is a known integer or types implicitly convertible to integer.
+          |-The matched expression only matches literal values, without type checks, if statements, or extractors.
+          |-The expression have its value available at compile time.
+          |-There are more than two case statements."""
+    }
+  }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -118,6 +118,7 @@ public enum ErrorMessageID {
     StaticFieldsOnlyAllowedInObjectsID,
     CyclicInheritanceID,
     UnableToExtendSealedClassID,
+    UnableToEmitSwitchID,
     ;
 
     public int errorNumber() {
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -14,6 +14,7 @@ import Decorators._
 import patmat.Space
 import NameKinds.{UniqueNameKind, PatMatStdBinderName, PatMatCaseName}
 import config.Printers.patmatch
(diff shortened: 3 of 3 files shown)


senia-psm:
@vitorsvieira The example in you error message is incorrect. `case _` is unreachable. You might want to either replace
```
case last => ...
```
with
```
case `last` => ...
```
or rename `last` to `Last`

You could also add a test to your PR. See [this example](https://github.com/lampepfl/dotty/pull/3516/files#diff-1c6a63362aa319e461b47ac777ddde58).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1972,4 +1972,36 @@ object messages {
...
+    val msg = hl"""Could not emit switch for ${"@switch"} annotated match"""
+    val explanation = {
+      val errorCodeExample =
+        """val middle = 'm'
+          |val last   = 'z'
+          |
+          |val number   = (middle: @switch) match {
+          |  case 'a'  => 1
+          |  case 'm'  => 13
+          |  case last => 26  //a non-literal may prevent switch generation: this would not compile.
```
senia-psm: `case _` is unreachable. You might want to either replace
```
case last => ...
```
with
```
case `last` => ...
```
or rename `last` to `Last`

vitorsvieira: Fixed.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1972,4 +1972,36 @@ object messages {
     val msg = hl"Cannot extend ${"sealed"} $pclazz in a different source file"
     val explanation = "A sealed class or trait can only be extended in the same file as its declaration"
   }
+
+  case class UnableToEmitSwitch()(implicit ctx: Context)
+  extends Message(UnableToEmitSwitchID) {
+    val kind = "Syntax"
+    val msg = hl"""Could not emit switch for ${"@switch"} annotated match"""
```
allanrenucci: No need for triple quotes

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1972,4 +1972,36 @@ object messages {
...
+      val errorCodeExample =
+        """val Middle = 'm'
+          |val Last   = 'z'
+          |
+          |val number   = (Middle: @switch) match {
+          |  case 'a'  => 1
+          |  case 'm'  => 13
+          |  case Last => 26  //a non-literal may prevent switch generation: this would not compile.
+          |  case _    => 0
+          |}""".stripMargin
```
allanrenucci: A more complete example:
```scala
val ConstantB = 'B'
final val ConstantC = 'C'
def tokenMe(ch: Char) = (ch: @switch) match {
  case '\t' | '\n' => 1
  case 'A'         => 2
  case ConstantB   => 3  // a non-literal may prevent switch generation: this would not compile
  case ConstantC   => 4  // a constant value is allowed
  case _           => 5
}
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1972,4 +1972,36 @@ object messages {
...
+          |Consider the following example:
+          |
+          |$errorCodeExample
+          |
+          |The example above would fail as it matches a non-literal value.
+          |The compiler will apply the tableswitch optimization when:
+          |-The matched value is a known integer or types implicitly convertible to integer.
+          |-The matched expression only matches literal values, without type checks, if statements, or extractors.
+          |-The expression have its value available at compile time.
+          |-There are more than two case statements."""
```
```
If annotated with ${"@switch"}, the compiler will verify that the match has been compiled to a
tableswitch or lookupswitch and issue an error if it instead compiles into a series of conditional
expressions. Example usage:

$codeExample

The compiler will not apply the optimisation if:
- the matched value is not of type ${"Int"}, ${"Byte"}, ${"Short"} or ${"Char"}
- the matched value is not a constant literal
- there are less than three cases
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Updated.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Awesome! Updated.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Updated.

nicolasstucki:
Please change the commit message to `Fix #1589: Add error message for not emitting switch`

vitorsvieira:
Cool! 👍

PR #3558, opened 2017-11-26
Ref #1589: Add error message for forward reference not allowed from self constructors

vitorsvieira:
Regarding #1589, this PR addresses:
- "Forward reference not allowed from self constructor invocation" (RefCheck:874).

Wasn't able to add a test for this error message as dotty seems to not have it ported from `scalac` yet. Evidence: https://scastie.scala-lang.org/lnrxIGq3SUCpoCXpUw3yGA

With scalac:
```
class A(a: Any) {
  def this() = {
    this(b)             //forward reference not allowed from self constructor invocation
    def b = new {}
  }
}
```

With dotty:
```
Starting dotty REPL...
scala> class A(a: Any) {
           def this() = {
             this(b)
             def b = new {}
           }
         }
3 |      this(b)
  |           ^
  |           not found: b
```

@smarter recommended to open an issue for this case. #3559

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorMessageID.java
@@ -118,6 +118,7 @@ public enum ErrorMessageID {
     StaticFieldsOnlyAllowedInObjectsID,
     CyclicInheritanceID,
     UnableToExtendSealedClassID,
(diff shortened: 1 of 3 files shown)


nicolasstucki:
Please change the commit message to `Fix #1589: Add error message for forward reference not allowed from self constructors`.

allanrenucci:
@vitorsvieira Since the part emitting this error seems to be dead code, we prefer not to add a new error message that will never be used. Thanks for the contribution anyway! 😄

PR #3560, opened 2017-11-26
Simplify REPL

allanrenucci:
- Remove monadic API when not needed: methods returning `Result[T]` now return `T` if failure is not possible
- Other minor polishing

### DIFF
--- a/compiler/test-resources/repl/patdef
+++ b/compiler/test-resources/repl/patdef
@@ -22,5 +22,5 @@ val x: Int = 1
 scala> val List(_ @ List(x)) = List(List(2))
 val x: Int = 2
 scala> val B @ List(), C: List[Int] = List()
-val B: List[Int] = Nil
-val C: List[Int] = Nil
+val B: List[Int] = List()
+val C: List[Int] = List()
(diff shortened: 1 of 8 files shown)


compiler/test-resources/repl/patdef
```diff
@@ -22,5 +22,5 @@ val x: Int = 1
 scala> val List(_ @ List(x)) = List(List(2))
 val x: Int = 2
 scala> val B @ List(), C: List[Int] = List()
-val B: List[Int] = Nil
-val C: List[Int] = Nil
+val B: List[Int] = List()
+val C: List[Int] = List()
```
felixmulder: ?

library/src/dotty/Show.scala
```diff
@@ -72,12 +73,12 @@ object Show {
 
   implicit def showList[T](implicit st: Show[T]): Show[List[T]] = new Show[List[T]] {
     def show(xs: List[T]) =
-      if (xs.isEmpty) "Nil"
+      if (xs.isEmpty) "List()"
```
felixmulder: ?

library/src/dotty/Show.scala
```diff
@@ -72,12 +73,12 @@ object Show {
...
       else "List(" + xs.map(_.show).mkString(", ") + ")"
   }
 
   implicit val showNil: Show[Nil.type] = new Show[Nil.type] {
-    def show(xs: Nil.type) = "Nil"
+    def show(xs: Nil.type) = "List()"
```
felixmulder: ?

allanrenucci: It is what scalac does:
```scala
scala> List()
res0: List[Nothing] = List()

scala> List(1, 2)
res1: List[Int] = List(1, 2)
```
I think it is more consistent

felixmulder: ¯\_(ツ)_/¯

Sounds like a debate for the ages. I'm definitely fine with the change, just wanted to know if there was some other motive beside the superficial one :)

PR #3562, opened 2017-11-26
Ref #1589: Add warning messages for migration annotation

vitorsvieira:
Regarding #1589, this PR addresses:

- "{symbol} has an unparsable version number..." (RefCheck:694)
- "{symbol} has changed semantics in version..." (RefCheck:698)


nicolasstucki:
Please change the commit message to `Fix #1593: Add warning messages for migration annotation`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
   }
 
+  case class SymbolHasUnparsableVersionNumber(symbol: Symbol, migrationMessage: String)(implicit ctx: Context)
+  extends Message(SymbolHasUnparsableVersionNumberID) {
+    val kind = "Syntax"
+    val msg = hl"${symbol.showLocated} has an unparsable version number: $migrationMessage"
+    val explanation =
+      hl"""$migrationMessage
+          |
+          |The symbol is marked with ${"@migration"} indicating it has changed semantics
```
nicolasstucki: You should not talk about symbols as this is a compiler internal concept, the user should not know about them. You should probably replace it with `${symbol.showLocated}` or `${symbol.show}`.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
+          |between versions and the ${"-Xmigration"} settings is used to warn about constructs
+          |whose behavior may have changed since version.""".stripMargin
```
nicolasstucki: "may have changed since version" -> "may have changed since version change"

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
+    migrationVersion: scala.util.Try[ScalaVersion]
+  )(implicit ctx: Context) extends Message(SymbolChangedSemanticsInVersionID) {
+    val kind = "Syntax"
+    val msg =
+      hl"""${symbol.showLocated} has changed semantics in version $symbolVersion:
+          |$migrationVersion""".stripMargin
+    val explanation = {
+      hl"""The symbol is marked with ${"@migration"} indicating it has changed semantics
+          |between versions and the ${"-Xmigration"} settings is used to warn about constructs
+          |whose behavior may have changed since version.""".stripMargin
```
nicolasstucki: same

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
+    symbol: Symbol,
+    symbolVersion: scala.util.Try[ScalaVersion],
+    migrationVersion: scala.util.Try[ScalaVersion]
+  )(implicit ctx: Context) extends Message(SymbolChangedSemanticsInVersionID) {
+    val kind = "Syntax"
+    val msg =
+      hl"""${symbol.showLocated} has changed semantics in version $symbolVersion:
+          |$migrationVersion""".stripMargin
+    val explanation = {
+      hl"""The symbol is marked with ${"@migration"} indicating it has changed semantics
```
nicolasstucki: same (symbol)

vitorsvieira:
I was always curious about this annotation. Is there any reason why it is not part of the standard library? ATM, it is available only internally ` private[scala] final class migration(message: String, changedIn: String) extends scala.annotation.StaticAnnotation`.

Just out of curiosity, Swift provides the `@available` attribute, that IMO is an expanded 
 representation of scala's `@migration`. Maybe a combination of `@deprecated` and  `@migration` 

In Swift, `@available(Platform+VersionNumber*, Args*)` attribute can be used to indicate the whole `declaration`'s  lifecycle, relative to certain Swift language versions or certain platforms and operating system versions. Some of its arguments:

- unavailable(`version number`)
- introduced(`version number`)
- deprecated(`version number`)
- obsoleted(`version number`)
- message(`message`)
- renamed(`new name`)

In scala's ecosystem, this would mean the possibility to achieve something like:
`@available(JVM1_8, ScalaJS1_0, ScalaNative1_0, introduced:2.4, deprecated:2.6, obsoleted:2.10, renamed: "newMethodName")`

Reference: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html#//apple_ref/doc/uid/TP40014097-CH35-ID347

PS: Not a suggestion, just toughts ;)

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -691,11 +691,11 @@ object RefChecks {
         case scala.util.Success(v) =>
           ctx.settings.Xmigration.value < v
         case Failure(ex) =>
-          ctx.warning(s"${sym.showLocated} has an unparsable version number: ${ex.getMessage()}", pos)
+          ctx.warning(SymbolHasUnparsableVersionNumber(sym, ex.getMessage()), pos)
           false
       }
       if (changed)
-        ctx.warning(s"${sym.showLocated} has changed semantics in version $symVersion:\n${sym.migrationMessage.get}")
+        ctx.warning(SymbolChangedSemanticsInVersion(sym, symVersion, sym.migrationMessage.get))
```
allanrenucci: While you're at it. Can you move this into the match. There is no need for this `changed` variable.
Also, `symVersion` is `sym.migrationVersion.get` so you don't need both of them

vitorsvieira: Sure. Updated.

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -685,17 +685,14 @@ object RefChecks {
...
     // Similar to deprecation: check if the symbol is marked with @migration
     // indicating it has changed semantics between versions.
-    if (sym.hasAnnotation(defn.MigrationAnnot) && ctx.settings.Xmigration.value != NoScalaVersion) {
-      val symVersion: scala.util.Try[ScalaVersion] = sym.migrationVersion.get
-      val changed = symVersion match {
-        case scala.util.Success(v) =>
-          ctx.settings.Xmigration.value < v
+    val xMigrationValue = ctx.settings.Xmigration.value
+    if (sym.hasAnnotation(defn.MigrationAnnot) && xMigrationValue != NoScalaVersion) {
+      sym.migrationVersion.get match {
```
allanrenucci: Your match is not exhaustive

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -685,17 +685,14 @@ object RefChecks {
...
+        case scala.util.Success(symVersion) if xMigrationValue < symVersion =>
+          ctx.warning(SymbolChangedSemanticsInVersion(sym, symVersion, sym.migrationMessage.get))
```
allanrenucci: You can remove `sym.migrationMessage.get`. It is the same as `symVersion`. I don't know why it was duplicated in the first place

allanrenucci: You're missing the `pos`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
+  case class SymbolHasUnparsableVersionNumber(symbol: Symbol, migrationMessage: String)(implicit ctx: Context)
+  extends Message(SymbolHasUnparsableVersionNumberID) {
+    val kind = "Syntax"
+    val msg = hl"${symbol.showLocated} has an unparsable version number: $migrationMessage"
+    val explanation =
+      hl"""$migrationMessage
+          |
+          |The ${symbol.showLocated} is marked with ${"@migration"} indicating it has changed semantics
+          |between versions and the ${"-Xmigration"} settings is used to warn about constructs
+          |whose behavior may have changed since version change.""".stripMargin
```
allanrenucci: You don't need to `stripMargin`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1973,6 +1974,34 @@ object messages {
...
+    migrationVersion: scala.util.Try[ScalaVersion]
+  )(implicit ctx: Context) extends Message(SymbolChangedSemanticsInVersionID) {
+    val kind = "Syntax"
+    val msg =
+      hl"""${symbol.showLocated} has changed semantics in version $symbolVersion:
+          |$migrationVersion""".stripMargin
+    val explanation = {
+      hl"""The ${symbol.showLocated} is marked with ${"@migration"} indicating it has changed semantics
+          |between versions and the ${"-Xmigration"} settings is used to warn about constructs
+          |whose behavior may have changed since version change.""".stripMargin
```
allanrenucci: You don't need to `stripMargin`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Fixed.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Fixed.

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
vitorsvieira: Fixed.

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
vitorsvieira: Fixed.

vitorsvieira: I would have to change the final message if that is ok. Currently:
```
${symbol.showLocated} has changed semantics in version $symbolVersion:
$migrationVersion
```
The migrationMessage provides either a version string or an error message from `ScalaVersion.parse`:

```
 def failure = Failure(new NumberFormatException(
      s"There was a problem parsing ${versionString}. " +
       "Versions should be in the form major[.minor[.revision]] " +
       "where each part is a positive number, as in 2.10.1. " +
       "The minor and revision parts are optional."
    ))
```

allanrenucci: It cannot be a failure since you're in the `Success` case of the match

vitorsvieira: OK

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -685,17 +685,15 @@ object RefChecks {
...
+    val xMigrationValue = ctx.settings.Xmigration.value
+    if (sym.hasAnnotation(defn.MigrationAnnot) && xMigrationValue != NoScalaVersion) {
+      sym.migrationVersion.get match {
+        case scala.util.Success(symVersion) if xMigrationValue < symVersion=>
+          ctx.warning(SymbolChangedSemanticsInVersion(sym, symVersion), pos)
         case Failure(ex) =>
-          ctx.warning(s"${sym.showLocated} has an unparsable version number: ${ex.getMessage()}", pos)
-          false
+          ctx.warning(SymbolHasUnparsableVersionNumber(sym, ex.getMessage()), pos)
+        case _ => ()
```
allanrenucci: Remove `()`

vitorsvieira: Removed.

PR #3565, opened 2017-11-27
Progress towards full SIP-23 support: correct inference for scala.Singleton, support for symbol literals

smarter:


smarter:
@milessabin The narrowing trick from scalac does not work in Dotty:
```scala
scala> def narrow[T <: Singleton](x: T): T {} = x 
def narrow[T <: Singleton](x: T): T
scala> narrow(1) 
val res0: Int = 1
```
However, the following works in both scalac and dotty:
```scala
scala> type Identity[T] = T 
// defined alias type Identity = [T] => T
scala> def narrow[T <: Singleton](x: T): Identity[T] = x 
def narrow[T <: Singleton](x: T): ([T] => T)[T]
scala> narrow(1) 
val res1: ([T] => T)[1] = 1
scala> val b = res51
val b: ([T] => T)[1] = 1
scala> val c: 1 = b 
val c: 1 = 1
```
(The fact that the dotty repl prints `([T] => T)[T]` instead of `Identity[T]` seems to be a pretty-printer issue, it shouldn't matter).

It might be useful to add a `scala.Identity[T]` as well as a definition of `narrow` to the Predef at some point.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -315,6 +316,7 @@ object TastyFormat {
   final val STRINGconst = 77
   final val IMPORTED = 78
   final val RENAMED = 79
+  final val SCALASYMBOLconst = 80
```
odersky: Why add Scala? I'd prefer we just called it `SYMBOLconst`.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -331,6 +333,8 @@ class TreeUnpickler(reader: TastyReader, nameAtRef: NameRef => TermName, posUnpi
           ConstantType(Constant(readType()))
         case ENUMconst =>
           ConstantType(Constant(readTermRef().termSymbol))
+        case SCALASYMBOLconst =>
```
odersky: I don't think you need the duplication. SYMBOLconsts are simple types, so they should just occur in readSimpleType. I believe we duplication of CLASSconst and ENUMconst is also wrong and should be removed.

PR #3566, opened 2017-11-27
REPL: do not dealias before printing types

smarter:
Also simplify the mechanism used to avoid printing prefixes. This is now
handled in the PlainPrinter and limited to List, Map, Set, Seq.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -357,8 +357,15 @@ class Definitions {
     def newGenericArrayMethod(implicit ctx: Context) = DottyArraysModule.requiredMethod("newGenericArray")
     def newArrayMethod(implicit ctx: Context) = DottyArraysModule.requiredMethod("newArray")
 
+  lazy val ListType = ctx.requiredClassRef("scala.collection.immutable.List")
+  def ListClass(implicit ctx: Context) = ListType.symbol.asClass
   lazy val NilModuleRef = ctx.requiredModuleRef("scala.collection.immutable.Nil")
   def NilModule(implicit ctx: Context) = NilModuleRef.symbol
+
+  lazy val immutableMapType = ctx.requiredClassRef("scala.collection.immutable.Map")
+  def immutableMapClass(implicit ctx: Context) = immutableMapType.symbol.asClass
+  lazy val immutableSetType = ctx.requiredClassRef("scala.collection.immutable.Set")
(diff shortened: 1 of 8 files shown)


compiler/test-resources/type-printer/prefixless
```diff
@@ -0,0 +1,8 @@
+scala> List(1,2,3)
+val res0: List[Int] = List(1, 2, 3)
+scala> Map("foo" -> 1)
+val res1: Map[String, Int] = Map("foo" -> 1)
+scala> Seq('a','b')
+val res2: Seq[Char] = List(a, b)
+scala> Set(4, 5)
+val res6: Set[Int] = Set(4, 5)
```
allanrenucci: Should be `res3`

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -357,9 +357,16 @@ class Definitions {
     def newGenericArrayMethod(implicit ctx: Context) = DottyArraysModule.requiredMethod("newGenericArray")
     def newArrayMethod(implicit ctx: Context) = DottyArraysModule.requiredMethod("newArray")
 
+  lazy val ListType = ctx.requiredClassRef("scala.collection.immutable.List")
+  def ListClass = ListType.symbol.asClass
```
odersky: This looks wrong. Don't we need to pass a context to `ListClass`, like we do everywhere else? (same for the other new definitions)

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -147,7 +152,12 @@ class PlainPrinter(_ctx: Context) extends Printer {
       case tp: TermRef if tp.denot.isOverloaded =>
         "<overloaded " ~ toTextRef(tp) ~ ">"
       case tp: TypeRef =>
-        toTextPrefix(tp.prefix) ~ selectionString(tp)
+        printWithoutPrefix.find(sym => tp.isDirectRef(sym)) match {
```
odersky: I would prefer not to hard-code these. Can't we instead check when printing whether a class has an alias in `Predef` or `scala` with the same name?

smarter: Possibly, but that would be more expensive and may require forcing things.

compiler/src/dotty/tools/dotc/core/Definitions.scala
smarter: Indeed, fixed.

smarter:
test performance please

PR #3567, opened 2017-11-27
Fix #3467: Don't pre-check kinds in TypeAssigner

### LINKED ISSUE #3467: Kind error in simple inheritance
We're running into an error compiling collection-strawman on Dotty 0.4.0-RC1. Here's the branch with the reproduction: https://github.com/scala/collection-strawman/compare/wip/dotty-kind-error

Steps to reproduce:

```
# Cleaning in sbt is bugged in our build; do it manually to make sure you rebuild everything
$ rm -r collections/.jvm/target/

$ sbt ++0.4.0-RC1 collectionsJVM/test:compile
[...]
[info] Compiling 86 Scala sources and 5 Java sources to /Users/szeiger/code/scala/collection-strawman/collections/.jvm/target/scala-0.4/classes...
[...]
[warn] 12 warnings found
[info] Compiling 1 Scala source to /Users/szeiger/code/scala/collection-strawman/collections/.jvm/target/scala-0.4/test-classes...
[error]
[error] Type argument strawman.collection.immutable.IndexedSeq has not the same kind as its bound  <: [_$63] => scala.Any
[error] one error found
[error] (collectionsJVM/test:compileIncremental) Compilation failed
```

The error message does not give any position. It is caused by refining the type of `ArrayOps.withFilter`:

```scala
  override def withFilter(p: A => Boolean): ArrayOps.WithFilter[A] = new ArrayOps.WithFilter[A](p, this)
```

Without the type annotation here (or by explicitly annotating it with the original type `collection.WithFilter[A, immutable.IndexedSeq]`) the test class compiles. I have been unable to come up with any workaround that would allow us to refine the type (as required to get some extra methods).

The `IndexedSeq` type argument mentioned in the error message comes from the definition of `ArrayOps.WithFilter`:

```
class WithFilter[A](p: A => Boolean, ao: ArrayOps[A]) extends collection.WithFilter[A, immutable.IndexedSeq]
```

This error is a blocker for https://github.com/scala/collection-strawman/pull/278

odersky:
In the existing tests we only need to pre-check kinds in Typer in order
to avoid self-applications that can lead to crashes. By avoiding them in
TypeAssigner we also avoid the error exhibited by i3467 in the strawman
collections.

This change required some changes in the tests because errors now get reported
at different times.

### DIFF
--- a/tests/neg/kinds1.scala
+++ b/tests/neg/kinds1.scala
@@ -5,14 +5,12 @@ object Test {
 
   class B
 
-  val x: C[C] = ??? // error: missing type parameter(s)
+  val x: C[C] = ??? // error: Type argument has not the same kind as its bound
   val y: C2[C] = ???
 
   def f[T] = ???
 
   def f2[T[X]] = ???
 
-  f[C] // error: missing type parameter(s)
   f2[C]
-
 }
--- a/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
@@ -13,7 +13,7 @@ import NameOps._
 import collection.mutable
 import reporting.diagnostic.Message
 import reporting.diagnostic.messages._
-import Checking.{preCheckKind, preCheckKinds, checkNoPrivateLeaks}
+import Checking.checkNoPrivateLeaks
 
 trait TypeAssigner {
   import tpd._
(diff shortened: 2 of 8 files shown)


tests/neg/kinds1.scala
```diff
@@ -0,0 +1,16 @@
+object Test {
```
smarter: This file and kinds2.scala are almost identitical, the other one could be removed.

allanrenucci: `kinds2.scala` tests another error. So I think we should keep it

smarter: Any reason to not test `f[C]` in this file?

allanrenucci: I believe because the error is not reported with this file:
> This change required some changes in the tests because errors now get reported
at different times.

PR #3568, opened 2017-11-27
Fix ambiguous implicit for Show[Nothing]

allanrenucci:
PR #3421 changed implicit search:
In the old scheme, a method with an implicit argument and a default
value would give you the default value if called with an ambiguous
implicit argument. The new scheme emits an error.

We now provide an implicit instance of `Show[Nothing]` in order to avoid
ambiguous implicit error in the REPL for expressions like:
```scala
scala> List()
val res0: List[Nothing] = List()
scala> Option.empty
val res1: Option[Nothing] = None
scala> Map()
val res2: Map[Nothing, Nothing] = Map()
```

### DIFF
--- a/library/src/dotty/Show.scala
+++ b/library/src/dotty/Show.scala
@@ -1,21 +1,16 @@
 package dotty
 
-import scala.annotation.implicitNotFound
-
-@implicitNotFound("No member of type class Show could be found for ${T}")
-trait Show[-T] {
+trait Show[T] {
   def show(t: T): String
 }
 
-/** Ideally show would only contain `defaultShow` and the pimped generic class,
-  * but since we can't change the current stdlib, we're stuck with providing
-  * default instances in this object
-  */
-object Show {
-  private[this] val defaultShow: Show[Any] = new Show[Any] {
-    def show(x: Any) = x.toString
+trait LowPrioShow {
+  implicit def defaultShow[T]: Show[T] = new Show[T] {
+    def show(x: T) = x.toString
   }
+}
 
+object Show extends LowPrioShow {
   /** This class implements pimping of all types to provide a show method.
     * Currently it is quite permissive, if there's no instance of `Show[T]` for
     * any `T`, we default to `T#toString`.
@@ -46,16 +41,8 @@ object Show {
     }
   }
 
-  implicit val intShow: Show[Int] = new Show[Int] {
-    def show(i: Int) = i.toString
-  }
-
   implicit val floatShow: Show[Float] = new Show[Float] {
     def show(f: Float) = f + "f"
-  }
-
-  implicit val doubleShow: Show[Double] = new Show[Double] {
-    def show(d: Double) = d.toString
   }
 
   implicit val charShow: Show[Char] = new Show[Char] {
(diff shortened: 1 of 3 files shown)


library/test/dotty/ShowTests.scala
```diff
@@ -64,6 +64,5 @@ class ShowTests {
     case class Car(model: String, manufacturer: String, year: Int)
 
     assertEquals("Car(Mustang,Ford,1967)", Car("Mustang", "Ford", 1967).show)
-    assertEquals("Map()", Map().show)
```
allanrenucci: I removed this. For some reason, it now causes a diverging implicit search in scalac

library/src/dotty/Show.scala
```diff
@@ -71,6 +71,10 @@ object Show {
     }) + "'"
   }
 
+  implicit val nothingShow: Show[Nothing] = new Show[Nothing] {
+    def show(x: Nothing) = ???
+  }
```
felixmulder: Should we really need this? Does it show up for `showList`, have you tried to demand `implicit ev: T =:!= Nothing`?

library/test/dotty/ShowTests.scala
felixmulder: T_T

library/src/dotty/Show.scala
allanrenucci: Currently:
```scala
scala> List() 
1 |List()
  |^
  |ambiguous implicit arguments of type dotty.Show[scala.collection.immutable.List[Nothing]] found for parameter ev of method show in class ShowValue.
  |I found:
  |
  |    dotty.Show.showList[T](/* ambiguous */implicitly[dotty.Show[T]])
  |
  |But both getter intShow in object Show and getter stringShow in object Show match type dotty.Show[T].
```
> have you tried to demand implicit ev: T =:!= Nothing? (in List / Map)

`=:!=` is not available by default. I quickly look it up online and it scared me. @OlivierBlanvillain This is your area of expertise. What do you think? Do you have an implementation of `=:!=` that works for Dotty? Is is preferable to an implicit `Show[Nothing]`?

OlivierBlanvillain:
```scala
type =:!=[A, B] = implicits.Not[=:=[A, B]]
```

OlivierBlanvillain: Not sure which one is worse, that or a  `Show[Nothing]` :smile:

felixmulder: Maybe our `Show` instances should not be so all-encompassing? :)

I mean, we need something equivalent to this mechanism for the REPL. But that doesn't mean we have to expose our users to it..

OlivierBlanvillain:
I think I prefer the first commit. We need to resist Felix trolling us from abroad :P

OlivierBlanvillain:
You could try to make `Show` invariant and add a default for any `T` (instead of `Any`):

```scala
private[this] def defaultShow[T]: Show[T] = new Show[T] {
  def show(x: T) = x.toString
}
```

The the following instances become redundant: `intShow`, `doubleShow`, `showNone`, `showMapOfNothing`.

allanrenucci:
Credits go to @OlivierBlanvillain. @felixmulder What do you think about the current fix?

OlivierBlanvillain:
LGTM

felixmulder:
It's incredible

PR #3570, opened 2017-11-27
Do not fail on warnings in sbt-dotty scripted tests

nicolasstucki:
To avoid errors seen in http://dotty-ci.epfl.ch/lampepfl/dotty/1700/7

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -851,6 +851,7 @@ object Build {
   // https://github.com/lampepfl/dotty-example-project for usage.
   lazy val `sbt-dotty` = project.in(file("sbt-dotty")).
     settings(commonSettings).
+    settings(scalacOptions -= "-Xfatal-warnings").
     settings(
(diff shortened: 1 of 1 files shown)


nicolasstucki:
This could potentially be the source of the unexpected cancellation of http://dotty-ci.epfl.ch/lampepfl/dotty/1700/8

nicolasstucki:
The strange thing is that with fatal warnings we have errors reporter in the console but they are not failures. I would prefer to keep them as warnings to make sure those tests are run to the end an not interrupted. 
There is also the question why those errors do not make the tests fail. Could actual failures also be swallowed by these tests?

allanrenucci:
>The strange thing is that with fatal warnings we have errors reporter in the console but they are not failures.

What errors are you talking about?

allanrenucci:
[Here](http://dotty-ci.epfl.ch/lampepfl/dotty/1707) the build for the latest push on master passed.

PR #3574, opened 2017-11-28
Prune impossible types from exhaustivity checker

OlivierBlanvillain:
The exhaustivity checker sometimes looks for impossible types.

The test case added by this PR is an example taken from the shapeless test suite where a sealed trait is extended in two different scopes. The added case classes`One` and `Two` can only appear withing object `Atoms01` and  `Atoms02` respectively. On master the exhaustivity checker reports missing cases of the following form: `PathVariantDefns.Atom2 & PathVariantDefns.Atoms01.type(PathVariantDefns.Atoms01).Two(_)`. Given that `PathVariantDefns.Atoms01.type` is the type of an object, the intersection between this type and any unrelated type is always empty. This PR adds a check to `instantiate` that detect and remove these cases.

@liufengyun WDYT?

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -267,8 +267,8 @@ trait SpaceLogic {
         else
           // `(_, _, _) - (Some, None, _)` becomes `(None, _, _) | (_, Some, _) | (_, _, Empty)`
           Or(ss1.zip(ss2).map((minus _).tupled).zip(0 to ss2.length - 1).map {
-              case (ri, i) => Prod(tp1, fun1, sym1, ss1.updated(i, ri), full)
-            })
+            case (ri, i) => Prod(tp1, fun1, sym1, ss1.updated(i, ri), full)
+          })
 
     }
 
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -640,6 +640,24 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
+        false
+    })
+
+    if (hasEmptyIntersections(false, result)) {
+      debug.println(s"hasEmptyIntersections($protoTp1) = true")
+      NoType
+    } else {
+      result
+    }
   }
```
liufengyun: In the check, there's a facility named `implementability` from a contributor which does more or less the same thing. What do you think we can generalise/adapt `implementability` to handle the case?

In particular, I think it's good to rename it to `def inhabited(tp: Type): Boolean`, and it should be able to handle the current case as well. WDYT?

OlivierBlanvillain: I think `implementability` computes if a type can be implemented by other types (singleton types can't!), while `inhabited` would need to check if there could actually exists an instance of that type, so I'm not sure two logics can be merged. @AleksanderBG what's your opinion on that?

liufengyun: I think `implementability` should be refactored to have the semantics of `inhabited`. According to the core algorithm of exhaustivity check, it should only depend on the semantics of `def inhabited(tp: Type): Boolean`.

AleksanderBG: @OlivierBlanvillain I believe @liufengyun is correct - the function can be safely changed to check if a type is inhabited instead. This won't change the result for and-types, which is the only result `intersectUnrelatedAtomicTypes` should care about. I considered a similar refactoring myself when working on #3454. With the benefit of hindsight, type inhabitation is a far more generic notion and I should've tried to implement it from the beginning.

PR #3599, opened 2017-11-29
Fix dotr for Cygwin and MingW

liufengyun:
This PR fix the dotr script for Cygwin and MingW.

@allanrenucci : Is it possible to run a service on test machine to listen for commands like "test scripts please" or "test community build" please?  If it's possible, I'm motivated to write such scripts.

### DIFF
--- a/dist/bin/dotr
+++ b/dist/bin/dotr
@@ -61,7 +61,7 @@ if [ $run_repl == true ] || [ ${#residual_args[@]} -eq 0 ]; then
     cp_arg="-classpath \"$CLASS_PATH\""
   fi
   echo "Starting dotty REPL..."
-  eval "$PROG_HOME/bin/dotc $cp_arg -repl ${residual_args[@]}"
+  eval "\"$PROG_HOME/bin/dotc\" $cp_arg -repl ${residual_args[@]}"
 else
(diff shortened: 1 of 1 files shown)


liufengyun:
The test for the fix is green: https://github.com/lampepfl/packtest/pull/8

dist/bin/dotr
```diff
@@ -58,16 +58,16 @@ done
 
 if [ $run_repl == true ] || [ ${#residual_args[@]} -eq 0 ]; then
   if [ "$CLASS_PATH" ]; then
-    cp_arg="-classpath $CLASS_PATH"
+    cp_arg="$CLASS_PATH"
   fi
   echo "Starting dotty REPL..."
-  eval "$PROG_HOME/bin/dotc $cp_arg -repl ${residual_args[@]}"
+  eval "$PROG_HOME/bin/dotc -classpath \"$cp_arg\" -repl ${residual_args[@]}"
```
allanrenucci: The semantic is different here. Now if you pass no classpath, you will run `dotc -classpath "" ...`,
before `dotc  ...`. `dotc` is not happy if you pass an empty classpath

liufengyun: I've updated the PR, and added a test to packtest to guard against regression for empty classpath: https://github.com/lampepfl/packtest/pull/9/files

PR #3600, opened 2017-11-29
Do not clean the sbt bridge cache on every bootstrapped compile

smarter:
Previously, any compilation of a bootstrapped project removed the sbt
bridge cache which then had to be recompiled. This is too extreme: we
should only do it when the bridge needs to be recompiled.


smarter:
Yes, exactly :).

Issue #3598, opened 2017-11-29
Error with wildcard type from TASTY


nicolasstucki:
Fail to compile from TASTY
```scala
class Foo[A] {
  def baz(foo: Foo[_]): Unit = bar(foo)
  def bar[A](foo: Foo[A]): A = ???
}
```

Can be reproduced with 
```
dotc <source>
dotc -decompile -Ycheck:all Test
```

```
exception while typing this.bar[Foo[_]#<parameter A of class Foo>](foo) of class class dotty.tools.dotc.ast.Trees$Apply # 68
exception while typing {
  this.bar[Foo[_]#<parameter A of class Foo>](foo)
  ()
} of class class dotty.tools.dotc.ast.Trees$Block # 70
exception while typing  def baz( foo: Foo[_ >: Nothing <: Any]): Unit = 
  {
    this.bar[Foo[_]#<parameter A of class Foo>](foo)
    ()
  } of class class dotty.tools.dotc.ast.Trees$DefDef # 50
exception while typing @scala.annotation.internal.SourceFile("tests/pos/i2250.scala")  class T ()
   extends
 Object() { 
   def baz( foo: Foo[_ >: Nothing <: Any]): Unit = 
    {
      this.bar[Foo[_]#<parameter A of class Foo>](foo)
      ()
    }
   def bar[ A]( foo: Foo[A]): A = ???
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 15
exception while typing package <empty> {
  @scala.annotation.internal.SourceFile("tests/pos/i2250.scala")  class T ()
     extends
   Object() { 
     def baz( foo: Foo[_ >: Nothing <: Any]): Unit = 
      {
        this.bar[Foo[_]#<parameter A of class Foo>](foo)
        ()
      }
     def bar[ A]( foo: Foo[A]): A = ???
  }
} of class class dotty.tools.dotc.ast.Trees$PackageDef # 22
*** error while checking ./T.class after phase frontend ***
Exception in thread "main" java.lang.AssertionError: assertion failed: found:    Foo[_](foo)
required: Foo[Foo[_]#<parameter A of class Foo>]


tree = foo
	at scala.Predef$.assert(Predef.scala:219)
	at dotty.tools.dotc.transform.TreeChecker$Checker.adapt(TreeChecker.scala:447)
	at dotty.tools.dotc.typer.ProtoTypes$FunProto.typedArg(ProtoTypes.scala:242)
[...]

Issue #3462, opened 2017-11-12
Typing error while compiling value classes from TASTY


nicolasstucki:
When compiling from tasty (tests in `tests/posFromTasty/`)
```scala
package foo
class Foo(a: Int) extends AnyVal
```
When recompiling the code loaded from tasty there is a type mismatch in 

```
exception while typing $this.a.==(x$0.a) of class class dotty.tools.dotc.ast.Trees$Apply # 7640
exception while typing x$0 match 
  {
    case x$0 @ _: foo.Foo => $this.a.==(x$0.a)
    case _: Any => false
  } of class class dotty.tools.dotc.ast.Trees$Match # 7648

...
checking simpleValueClass1.scala after phase MegaPhase{checkStatic, elimRepeated, normalizeFlags, extmethods, expandSAMs, tailrec, byNameClosures, liftTry, hoistSuperArgs, classOf, refchecks}
java.lang.AssertionError: assertion failed: found:    x$0.a.type
required: Any
```

The issue is that `x$0.a.type` as no type.

nicolasstucki:
This can be reproduced with 
```
echo "package foo\nclass Foo(a: Int) extends AnyVal\n" > Foo.scala
sbt "dotc Foo.scala"
sbt "dotc -tasty -classpath . foo.Foo"
```

nicolasstucki:
Now it can also be reproduced with
```
dotc <source>
dotc -decompile -Ycheck:all Test
```

PR #3602, opened 2017-11-29
Replace TypeArgRefs by normal TypeRefs

odersky:
Replace TypeArgRefs by normal TypeRefs. This streamlines the treatment of wildcard arguments and brings it in line with path dependent types. We can also eliminate that way both a fundamental type constructor and a Tasty node class.

Based on #3494.

The change unveiled two errors, which are tracked in #3603 and #3604.


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -183,6 +182,18 @@ object Types {
       loop(this)
     }
 
+    /** True iff `symd` is a denotation of a class type parameter and the reference
+     *  `<this> . <symd>` is an actual argument reference, i.e. `this` is different
+     *  from the ThisType of `symd`'s owner.
+     */
+    def isArgPrefix(symd: SymDenotation)(implicit ctx: Context) =
```
smarter: Suggestion: call the method `isArgPrefixOf`.

odersky: Good point. In fact, I still want to try something else: Make internal class parameter references take a `NoPrefix`. Then any class parameter reference with some other prefix would be considered an external (type-arg) reference.

nicolasstucki:
This PR is also fixing #3462, #3598 and #3607 (as shown in #3610) :smiley:

compiler/test-resources/repl/i2554
```diff
@@ -1,7 +0,0 @@
-scala> object foo { trait ShapeLevel; trait FlatShapeLevel extends ShapeLevel; trait ColumnsShapeLevel extends FlatShapeLevel; abstract class Shape[Level <: ShapeLevel, -Mixed, Unpacked, Packed]; object Shape extends TupleShapeImplicits { }; trait TupleShapeImplicits { implicit final def tuple2Shape[Level <: ShapeLevel, M1,M2, U1,U2, P1,P2](implicit u1: Shape[_ <: Level, M1, U1, P1], u2: Shape[_ <: Level, M2, U2, P2]): Shape[Level, (M1,M2), (U1,U2), (P1,P2)] = ???; }; }
```
smarter: This commit can be removed now that https://github.com/lampepfl/dotty/issues/3603 is fixed.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -231,38 +224,48 @@ Standard Section: "Positions" Assoc*
...
+
+    final val QUALIFIED = 2          // A fully qualified name `<prefix>.<suffix>`.
+
+    final val EXPANDED = 3           // An expanded name `<prefix>$$<suffix>`,
+                                     // used by Scala-2 for private names.
+
+    final val EXPANDPREFIX = 4       // An expansion prefix `<prefix>$<suffix>`,
+                                     // used by Scala-2 for private names.
+
+    final val UNIQUE = 10            // A unique name `<name>$<num>` where `<num>`
```
smarter: Since we're breaking compat anyway, we could renumber the tags so that all the tasty tags are continuous (and non-tasty nametags could have numbers starting with 1000 to distinguish them)

odersky: I tried this but got weird bootstrap errors and don't have the time to chase them down. Maybe somebody else wants to give it a try? Since so much else is sitting on this commit, I am getting it merged now.

PR #3605, opened 2017-11-29
Ref #1589: Add error message for bad symbolic reference.

vitorsvieira:
Regarding #1589, this PR addresses the following error:
- "Bad symbolic reference. A signature..." - SymDenotations.scala:1964

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -19,7 +19,7 @@ import java.util.WeakHashMap
 import config.Config
 import config.Printers.{incremental, noPrinter}
 import reporting.diagnostic.Message
-import reporting.diagnostic.messages._
+import reporting.diagnostic.messages.BadSymbolicReference
 import reporting.trace
 
 trait SymDenotations { this: Context =>
@@ -1957,12 +1957,7 @@ object SymDenotations {
 
     def complete(denot: SymDenotation)(implicit ctx: Context): Unit = {
       val sym = denot.symbol
-      val file = sym.associatedFile
-      val (location, src) =
-        if (file != null) (s" in $file", file.toString)
-        else ("", "the signature")
-      val name = ctx.fresh.setSetting(ctx.settings.YdebugNames, true).nameString(denot.name)
-      val errMsg = BadSymbolicReference(location, name, denot.owner, src)
+      def errMsg = BadSymbolicReference(denot)
       ctx.error(errMsg, sym.pos)
       if (ctx.debug) throw new scala.Error()
       initializeToDefaults(denot, errMsg)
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1968,6 +1968,24 @@ object messages {
     }
   }
 
+  case class BadSymbolicReference(
+    location: String,
+    name: String,
+    denotationOwner: Symbol,
+    src: String
+  )(implicit ctx: Context) extends Message(BadSymbolicReferenceID) {
```
allanrenucci: I would change `BadSymbolicReference ` so that it only has `denot: SymDenotation`. You can derive `location`, `name`, `owner` and `src` from `denot`.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -18,6 +18,8 @@ import util.Stats
 import java.util.WeakHashMap
 import config.Config
 import config.Printers.{incremental, noPrinter}
+import reporting.diagnostic.Message
+import reporting.diagnostic.messages._
```
allanrenucci: Only import what you need

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1968,16 +1968,18 @@ object messages {
...
-    src: String
-  )(implicit ctx: Context) extends Message(BadSymbolicReferenceID) {
+  case class BadSymbolicReference(denot: SymDenotation)(implicit ctx: Context) extends Message(BadSymbolicReferenceID) {
     val kind = "Reference"
+
+    private val file = denot.symbol.associatedFile
+    private val (location, src) =
+      if (file != null) (s" in $file", file.toString)
+      else ("", "the signature")
+    private val name = ctx.fresh.setSetting(ctx.settings.YdebugNames, true).nameString(denot.name)
```
allanrenucci: We usually don't create fields to only store temporary values. If you need to precompute something for `msg` , you can use a block:
```scala
val msg = {
  val tmp1 = ...
  val tmp2 = ...
  doSomethingWith(tmp1, tmp2)
}
```
If a temporary computation is shared like `name`, we usually make it a `def`. We might want to create a field if we don't want to perform a computation multiple times. In this case I guess it make sense to keep `name` a field since `ctx.fresh...` may be expensive.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1968,16 +1968,18 @@ object messages {
...
+
     val msg =
       hl"""Bad symbolic reference. A signature$location
-          |refers to $name in ${denotationOwner.showKind} ${denotationOwner.showFullName} which is not available.
+          |refers to $name in ${denot.owner.showKind} ${denot.owner.showFullName} which is not available.
```
allanrenucci: Here you can also store the result of `owner` in a temporary `val` within the block

PR #3610, opened 2017-11-30
Add regression test #3598

### LINKED ISSUE #3462: Typing error while compiling value classes from TASTY

nicolasstucki:
Based on #3602


odersky:
Was integrated into #3602

PR #3612, opened 2017-11-30
 Fix #3581: Reorganise neg tests

### LINKED ISSUE #3581: Subdirectories in neg tests are not actually tested
```shell
mkdir tests/neg/foo
echo "dsqfsdfq" > tests/neg/foo/foo.scala
sbt
> test-only -- --tests=compileNeg
[success]
```

This happens because `compileShallowFilesInDir` is used in `compileNeg`. I have no idea why this is setup that way.

allanrenucci:


tests/neg-custom-args/structural.scala
```diff
@@ -36,20 +38,21 @@ object RClose {
...
+    idMap(Responder.constant(5)) // error: type mismatch: found Responder[Int], required Object{map: [U](f: Any => U): Any}
   }
 }
 }
 package p4 {
 
 trait A { self: Any { def p: Any } =>
   def f(b: => Unit): Unit = {}
-  f { p } // error: cannot access member 'p' from structural type
+  f { p } // OK
```
allanrenucci: Is it intended?

allanrenucci:
I had to move `tests/neg/inlineAccess` to `tests/pending/inlineAccess`. It crashes the compiler

compiler/test/dotty/tools/dotc/CompilerTest.scala
```diff
@@ -149,7 +149,8 @@ abstract class CompilerTest {
 
   private def expectedErrors(filePath: String): List[ErrorsInFile] = expectedErrors(List(filePath))
 
-  private def isNegTest(testPath: String) = testPath.contains("/neg/")
+  private def isNegTest(testPath: String) =
+    testPath.contains("/neg/") || testPath.contains("/neg-custom-args/") || testPath.contains("/neg-tailcall/")
```
smarter: `testPath.startsWith("/neg")` ?

smarter:
> I had to move tests/neg/inlineAccess to tests/pending/inlineAccess. It crashes the compiler

Could you open an issue for that?

compiler/test/dotty/tools/dotc/CompilerTest.scala
allanrenucci: I believe the full path is given here. Something like `../tests/neg/Foo.scala`. I wouldn't bother to do something smarter since this is for the legacy tests

smarter: ok.

PR #3616, opened 2017-12-01
Fix #3614: Do not allow the SyntaxFormatter to recolor code

### LINKED ISSUE #3614: Color leaks in error messages
```scala
trait Test[A]

object Test {
  val a = new Test[Int, Int, Int, Int, Int] {}
}
```

```scala
-- [E022] Syntax Error: /home/olivier/colol.scala:38:14 
38 |  val a = new Test[Int, Int, Int, Int, Int] {}
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |            Too many type arguments for Test[A]
   |            expected: [A]
   |            actual:   [35mInt0m, 35mInt0m, 35mInt0m, 35mInt0m, 35mInt0m]
```

nicolasstucki:


compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -86,7 +86,8 @@ object Formatting {
         case hb: HighlightBuffer =>
           hb.toString
         case str: String if ctx.settings.color.value != "never" =>
-          new String(SyntaxHighlighting(str).toArray)
+          if (str.matches(".*\u001b\\[.*?m.*")) str // was already colored by some other printer
```
allanrenucci: This seems very expensive. Error messages are colored because of the `hl` string interpolator. The interpolator call show on its arguments. Can we instead disable syntax highlighting when calling show for the argument of the interpolator?

allanrenucci: Redefining `hl` as:
```scala
def hl(args: Any*)(implicit ctx: Context): String = {
  val nestedCtx = ctx.fresh.setSetting(ctx.settings. color, true)
  new SyntaxFormatter(sc).assemble(args)(nestedCtx).stripMargin
)
```
should work. I am wondering how expensive is `ctx.fresh` though.

nicolasstucki: This might work if the argument in `args` is still of type `Text`.

nicolasstucki: At this point it is too late. I created instead an `AlreadyColored` that fixes the issue for each specific case. It should be added to any printed tree that is already colored.

compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -77,17 +77,16 @@ object Formatting {
...
-          new String(highlighted.toArray)
         case hl: Highlight =>
           hl.show
         case hb: HighlightBuffer =>
           hb.toString
         case str: String if ctx.settings.color.value != "never" =>
           new String(SyntaxHighlighting(str).toArray)
-        case _ => super.showArg(arg)
+        case x :: xs =>
+          showArg(x) + (if (xs.isEmpty) "" else ", " + showArg(xs))
```
allanrenucci: I would remove this special case. This seems completely arbitrary to special case `List`.
If a user want to pretty print a `List` like this, then he should do it himself.

PR #3629, opened 2017-12-03
Fix #3623: Special case MethodType in PlainPrinter

### LINKED ISSUE #3623: Error message for unimplemented members does not pretty print methods correctly
```
[error] -- Error: /data/rw/code/scala/loop/rework/src/main/scala/bloop/State.sca
la:110:8
[error] 110 |  class MyRunner(msg: String) extends sbt.testing.Runner {
[error]     |        ^
[error]     |class MyRunner needs to be abstract, since:[error]     |it has 4 unimplemented members.
[error]     |/** As seen from class MyRunner, the missing signatures are as foll
ows.
[error]     | *  For convenience, these are usable as stub implementations.
[error]     | */
[error]     |  def args: (): Array[String] = ???
[error]     |  def done: (): String = ???[error]     |  def remoteArgs: (): Array[String] = ???
[error]     |  def tasks: (x$0: Array[sbt.testing.TaskDef]): Array[sbt.testing.T
ask] = ???
```
  
As you see, `def args: (): Array[String] = ???` has an extra `: ` before the parameter list.

allanrenucci:
There is still an issue for method with no parameter list:
```scala
def bar: Int // printed as: def bar: => Int
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -342,6 +342,8 @@ class PlainPrinter(_ctx: Context) extends Printer {
           else dclsText(trueDecls)
         tparamsText ~ " extends " ~ toTextParents(tp.parents) ~ "{" ~ selfText ~ declsText ~
           "} at " ~ preText
+      case mt: MethodType =>
+        toTextGlobal(mt)
       case tp =>
         ": " ~ toTextGlobal(tp)
     }


odersky:
What was fixed? We should add a test case.

nicolasstucki:
We currently don't have a good way to test the output of the printers. Or do we?

allanrenucci:
No. For this PR, I could add either A REPL test or an error message test.
I guess it would be nice to have tests for the printers though.

nicolasstucki:
I think it is fine, but we should design some easy way to test printers. I also need for tests in the decompiler. I already have some ideas and will start working on them.

PR #3631, opened 2017-12-04
Add prefix types

odersky:
Allow the syntax `prefixOp stableId` as a type, where `prefixOp` is one of `+, -, !, ~` and `stableId` is a stable term identifier. This is not the same as the second part of SIP 33, which proposed `~` as an operator over types.

Essentially we have a choice how to interprete an expression

    ~ t

It could be either

    t.unary_~

or

    t#unary_~

This PR picks the first alternative, since it seems to be a smaller delta wrt the status quo and also because it actually has a use case, namely expressing splice operations in quoted types.


sjrd:
This does not make much sense to me. It's definitely inconsistent with infix types. By the logic where `~T` stands for `T.unary_~` to mimic `~t` = `t.unary_~`, `T Either U` should stand for `T.Either[U]` to mimic `t op u` = `t.op(u)`.

olafurpg:
I am a bit concerned about overloading prefix + and - since they already represent contravariance/covariance in definition type position. This could be confusing for many

```scala
trait F[-T]
val x: -T
```

odersky:
> I am a bit concerned about overloading prefix + and - since they already represent contravariance/covariance in definition type position. This could be confusing for many

I don't see it. `-T` where `T` is a type is a variance annotation. `-t` where `t` is a term is a type, provided `t` contains a definition

    type unary_- = ...

odersky:
@sjrd I agree it looks inconsistent with infix definitions. Without a convincing use case the proposal does not make much sense (neither does the original SIP 33 prefix types proposal, really). 

So I am going to withdraw this as a separate PR and roll it into the staging PR, which does provide a use case.

PR #3634, opened 2017-12-05
A symmetric meta programming framework

odersky:
A first implementation of the compiler parts of symmetric meta programming. 

Still to be done in separate PRs:

 - Add unpickling
 - Implement `Liftable` natively
 - Decide on packaging -- where does `quoted` reside?
 - Work out how to represent TypeTags
 - Add more tests and examples


nicolasstucki:
I will integrate the TASTY reification from the prototype in https://github.com/dotty-staging/dotty/tree/tasty-quote

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -49,7 +49,8 @@ class Compiler {
       List(new LinkAll),            // Reload compilation units from TASTY for library code (if needed)
       List(new FirstTransform,      // Some transformations to put trees into a canonical form
            new CheckReentrant,      // Internal use only: Check that compiled program has no data races involving global vars
-           new ElimJavaPackages),   // Eliminate syntactic references to Java packages
+           new ElimPackagePrefixes, // Eliminate syntactic references to Java packages
+           new ReifyQuotes),        // Eliminate syntactic references to Java packages
```
nicolasstucki: We might need to do this before `FirstTranform`, when we will pickle the contents of the quotes we should probably give that pickler a tree that is in the same shape as the tree was in `Pickler`.

odersky: FirstTransforms does some simplifications. It eliminates NamedArgs and Imports and replaces  all trees representing types by TypeTree nodes. This is arguable good, since it simplifies the trees to reify quite a bit. The reason we keep the more complicated trees in the TASTY is for the IDE, so that all elements of source files are represented. But for macros and staging that should not be necessary.

odersky: On second thought, I now see you were right. It does become easier if we reify exactly in the pickled format.

smarter:
> The PR is based on the various TASTY fix PRs which are in turn based on #3602 - would be good to get that bunch in soon so we can consolidate.

I left some comments on https://github.com/lampepfl/dotty/pull/3494, once that's in the rest should be easy to get in.

tests/pos/quoteTest.scala
```diff
@@ -0,0 +1,15 @@
+import scala.meta._
```
olafurpg: Is the plan to use this for the public API or only tests? `scala.meta.{Tree,Type,Term,...}` are already defined in https://github.com/scalameta/scalameta/blob/v2.1.3/scalameta/trees/shared/src/main/scala/scala/meta/Trees.scala and get ~50,000 monthly downloads.

nicolasstucki: We could use `scala.meta.quote` instead.

odersky: > We could use scala.meta.quote instead.

Good idea.

smarter: Still seems a bit too easy to confuse with scala.meta. How about one of: scala.quote, scala.reify, scala.staging

nicolasstucki: +1 for `scala.staging`

odersky: How about `scala.quoted`? `staging` and `reify` is both expert-speak.

odersky: Also, it's a stretch to call macro part of this "staging".

odersky:
test performance please

nicolasstucki:
I found a small blocker when inlining quotes in separate compilations

```scala
object Macros {
  inline def f(): Unit = '(1)
}
```

```scala
class Test {
  Macros.f()
}
```

```
-- Error: tests/pos/quoted2.scala:6:14 -----------------------------------------
6 |
  |              ^
  |undefined: <special-ops>.' # 51: TermRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),module <special-ops>),') at frontend
one error found
```
We are probably missing a case in the `TreeUnpickler` that trasforms calls to `<special-ops>.'` to `Quoted` trees.

nicolasstucki:
@odersky, @smarter, @OlivierBlanvillain, @allanrenucci @liufengyun  we need to rebase all open all PRs on top of #3665. Without this, the CI will try to execute a test command that doesn't exist anymore.

tests/pos/liftable.scala
```diff
@@ -17,7 +17,7 @@ object Test {
       if (b) '(true) else '(false)
   }
 
-  implicit def ListIsQuotable[T: Type: Quotable]: Quotable[List[T]] = new {
+  implicit def ListIsQuotable[T: Quotable]: Quotable[List[T]] = new {
```
nicolasstucki: `Quotable` -> `Liftable`
In several places of this file

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+If `program` is treated as a quoted expression, the call to
+`Macro.assertImpl` becomes phase correct even if macro library and
+program are conceptualized as local definitions.
+
+But what about the call from `assert` to `assertImpl? Here, we need a
+tweak of the typing rules. An inline function such as `assert` that
+contains a splice operation outside an enclosing quote is called a
+_macro_. Macros are supposed to be expanded in a subsequent phase,
+i.e. in a quoted context. Therefore, they are also type checked as if
+they were in a quoted context, For instance, the definition of
```
allanrenucci: s/context,/context.

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+
+    class Array[T] {
+      inline def map[U](f: T => U): Array[U] = ~ Macros.mapImpl[T, U](’[U], ’(this), ’(f))
+    }
+
+Here’s the definition of the `mapImpl` macro, which takes quoted types and expressions to a quoted expression:
+
+    object Macros {
+
+      def mapImpl[T, U](u: Type[U], arr: Expr[Array[T]], op: Expr[T => U])(implicit ctx: Context): Expr[Array[U]] = ’{
```
nicolasstucki: We do not need the implicit context for this one.

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+        env(x)
+      case Let(x, e, body) =>
+        ’{ val y = ~compile(e, env); ~compile(body, env + (x -> ’(y))) }
+    }
+
+Running `compile(letExp, Map())` would yield the following Scala code:
+
+    ’{ val y = 3; (2 + y) + 4 }
+
+The body the first clause (`case Num(n) => n`) looks suspicious. `n`
```
allanrenucci: the body **of**

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+several types including all underlying types of literals. For instance
+`Int` values can be converted to `Expr[Int]` values by wrapping the
+value in a `Literal` tree node. This makes use of the underlying tree
+representation in the compiler for efficiency. But the `Liftable`
+instances are nevertheless not "magic" in the sense that they could
+all be defined in a user program without knowing anything about the
+representation of `Expr` trees. For instance, here is a possible
+instance of `Liftable[Boolean]`:
+
+    implicit def BooleanIsLiftable: Liftable[Boolean] = new {
```
allanrenucci: new **Liftable** {

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+      implicit def toExpr(b: Boolean) = if (b) ’(true) else ’(false)
+    }
+
+Once we can lift bits, we can work our way up. For instance, here is a
+possible implementation of `Liftable[Int]` that does not use the underlying
+tree machinery:
+
+    implicit def IntIsLiftable: Liftable[Int] = new {
```
allanrenucci: new **Liftable** {

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,743 @@
...
+        case 0               => ’(0)
+        case _ if n % 2 == 0 => ’(~toExpr(n / 2) * 2)
+        case _               => ’(~toExpr(n / 2) * 2 + 1)
+      }
+    }
+
+Since `Liftable` is a type class, instances can be conditional. For instance
+a `List` is liftable if its element type is:
+
+    implicit def ListIsLiftable[T: Liftable]: Liftable[List[T]] = new {
```
allanrenucci: new **Liftable** {

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,747 @@
+# Symmetric Meta Programming
+
+Symmetric meta programming is a new framework for staging and for some
+forms of macros. It is is expressed as strongly and statically typed
```
nicolasstucki: is is

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,747 @@
...
+add constructors that create typed trees directly without going
+through quotes. Most likely, those constructors would work over `Expr`
+types which lack a known type argument. For instance, an `Apply`
+constructor could be typed as follows:
+
+    def Apply(fn: Expr[_], args: List[Expr[_]]): Expr[_]
+
+This would allow constructing applications from lists of arguments
+without having to match the arguments one-by-one with the
+corresponding formal parameter types of the function. We need then "at
```
allanrenucci: We then need?

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,747 @@
...
+structure of `Expr` is a subtype of the type structure representing
+`T`.
+
+Before going down that route, we should evaluate in detail the tradeoffs it
+presents.  Constructing trees that are only verified _a posteriori_
+to be type correct loses a lot of guidance for constructing the right
+trees.  So we should wait with this addition until we have more
+use-cases that help us decide whether the loss in type-safety is worth
+the gain in flexibility. In this context, it seems that deconstructing types is
+less error-prone than deconstructing tersm, so one might also
```
allanrenucci: terms

docs/docs/reference/symmetric-meta-programming.md
```diff
@@ -0,0 +1,747 @@
...
+
+### Operational semantics:
+
+We define a small step reduction relation `-->` with the following rules:
+
+                ((x: T) => t) v  -->  [x := v]t
+
+                          ~(’t)  -->  t
+
+                              t  -->  t’
```
nicolasstucki: We should avoid using both `'` for quote and `'` prime. Maybe we should use unstead if we use `t --> t2`.

nicolasstucki:
With separate compilation, this code produces an unexpected error message.
```scala
import scala.quoted._
object Macros {
  inline def assert(expr: => Boolean): Unit =  ~ assertImpl('(expr))
  def assertImpl(expr: Expr[Boolean]) = '{ () }
}
```

```scala
class Test {
  import Macros._
  val x = 1
  assert(x != 0)
}
```

```scala
8 |
  |                    ^^^^
  |         value of type scala.quoted.Type[Boolean] does not take parameters
```

docs/docs/reference/symmetric-meta-programming.md
odersky: no, that's actually legal syntax now.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -225,7 +227,7 @@ object TastyFormat {
 
   final val header = Array(0x5C, 0xA1, 0xAB, 0x1F)
   val MajorVersion = 2
-  val MinorVersion = 0
+  val MinorVersion = 1
```
nicolasstucki: We do not need this since we have not published `2.0` yet

compiler/src/dotty/tools/dotc/transform/MacroTransformWithImplicits.scala
```diff
@@ -0,0 +1,90 @@
...
+import Contexts._
+import Symbols._
+import Decorators._
+import collection.mutable
+import annotation.tailrec
+
+/** A Macrotransform that maintains the necessary infrastructore to support
+ *  contxtual implicit searches (type-scope implicits are supported anyway).
+ */
+abstract class MacroTransformWithImplicits extends MacroTransform {
```
nicolasstucki: Should this really go in if we do not use it? What are the other use cases where it might be needed?

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
smarter: No sure what you mean by "published". We do have nightly builds out that use 2.0 and we should bump the version when we change something because it's good practice and because it's valuable information to anyone wanting to make a separate implementation of tasty.

nicolasstucki: OK

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1086,7 +1074,6 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   }
 
   def typedQuote(tree: untpd.Quote, pt: Type)(implicit ctx: Context): Tree = track("typedQuote") {
-    ctx.compilationUnit.containsQuotes = true
```
nicolasstucki: We do need `ctx.compilationUnit.containsQuotesOrSplices = true` in case there are only quotes in the tree. I can add the fix with the test case in #3662.

tests/pending/pos/quotedSepComp/Macro_1.scala
```diff
@@ -0,0 +1,5 @@
+import scala.quoted._
+object Macros {
+  inline def assert(expr: => Boolean): Unit =  ~ assertImpl('(expr))
+  def assertImpl(expr: Expr[Boolean]) = '{ () }
```
nicolasstucki: Indeed, this one works fine as a run test in #3662

compiler/src/dotty/tools/dotc/transform/MacroTransformWithImplicits.scala
ritschwumm: someone might want to fix the typos in the doc-comment, too...

PR #3639, opened 2017-12-07
Remove obsolete package members in new runs

smarter:
Before this commit, typing and then deleting "class Foo" at the
top-level in the IDE resulted in the creation of symbols "F", "Fo" and
"Foo" that were never deleted. We fix this by always deleting package members
coming from a previous run of the current file when indexing a package.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Scopes.scala
+++ b/compiler/src/dotty/tools/dotc/core/Scopes.scala
@@ -333,8 +333,12 @@ object Scopes {
     }
 
     /** remove symbol from this scope if it is present */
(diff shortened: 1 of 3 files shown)


smarter:
test performance please

liufengyun:
We've a problem with benchmarks which stopped run since Nov 20th (one merge in master doesn't compile). Now It's running regressionally to test merges since that time, which take several days.

I've temporarily disabled the regressional test in order for this test to run. It should run this afternoon.

I'll refine the test a little to ensure master test runs in pieces and PR test always takes precedence in the job queue.

Issue #2464, opened 2017-05-18
Implicit class without arguments


nicolasstucki:
```scala
object Foo {
  object Bar
  implicit class Bar
}
```

fails with

```
-- Error: tests/run/phantom-implicit-class.scala:3:2 ---------------------------
3 |  object Bar
  |  ^
  |  Foo.Bar is not stable
-- Error: tests/run/phantom-implicit-class.scala:3:9 ---------------------------
3 |  object Bar
  |         ^
  |         object Bar is already defined as method Bar: ()Foo.Bar

two errors found
```

the signature of the lazy val of the module `object Bar` and the signature of desugared implicit method of `implicit class Bar` are the same and they have the same name.

The issues are:
* The error message is not helpful unless you know about the desugaring and the signature internals.
* Using `implicit class` on phantoms generates these kinds of ambiguities because the parameter will be erased.

allanrenucci:
Fixed in #3641

PR #3641, opened 2017-12-07
Fix #3540: Implicit class with two arguments does not fail compilation

### LINKED ISSUE #3540: Implicit class with two arguments does not fail compilation
```scala
object Test {
  implicit class Foo(i: Int, s: String)
}
```
Fails in `scalac` with
```
Foo.scala:2: error: implicit classes must accept exactly one primary constructor parameter
  implicit class Foo(i: Int, s: String)
                 ^
one error found
```

But not on `dotc`. This error should probably be emitted when desugaring the `implicit class`.

vitorsvieira:
This PR aims #3540 and #2464.

Where:

```
object Test {
  implicit class Foo(i: Int, s: String)
}
```

And

```
object Foo {
  object Bar
  implicit class Bar
}
```

Compilation must fail providing `Implicit classes must accept exactly one primary constructor parameter` error message and further explanation.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -455,7 +455,7 @@ object messages {
   case class ImplicitClassPrimaryConstructorArity()(implicit ctx: Context)
   extends Message(ImplicitClassPrimaryConstructorArityID){
     val kind = "Syntax"
-    val msg = hl"Implicit classes must accept exactly one primary constructor parameter"
+    val msg = "Implicit classes must accept exactly one primary constructor parameter"
     val explanation = {
       val example = "implicit class RichDate(date: java.util.Date)"
       hl"""Implicit classes may only take one non-implicit argument in their constructor. For example:
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -20,7 +20,6 @@ import printing.Formatting
 import ErrorMessageID._
 import Denotations.SingleDenotation
 import dotty.tools.dotc.ast.Trees
-import dotty.tools.dotc.ast.untpd.Modifiers
 import dotty.tools.dotc.config.ScalaVersion
 import dotty.tools.dotc.core.Flags.{FlagSet, Mutable}
 import dotty.tools.dotc.core.SymDenotations.SymDenotation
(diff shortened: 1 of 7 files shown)


tests/neg/i3540.scala
```diff
@@ -0,0 +1,3 @@
+object Test {
+  implicit class Foo(i: Int, s: String) // error: Implicit classes must accept exactly one primary constructor parameter
+}
```
allanrenucci: Can you add more tests here:
```scala
implicit class Foo0
implicit class Foo1()
implicit class Foo2()(x: Int)
```

nicolasstucki: You should also add 
```scala
implicit case class Bar0
implicit case class Bar1()
implicit case class Bar2()(x: Int)
```

vitorsvieira: OK. Adding these tests,

nicolasstucki: Probably

allanrenucci: You simply need to add `// error` next to the lines that do not compile

vitorsvieira: 👍 Added.

allanrenucci: Can you also add the error messages emitted by the compiler:
```scala
implicit class Foo0 // error: <the error message here>
```
You can get the error messages by running in sbt:
```
dotc tests/neg/i3540.scala
```

vitorsvieira:
```

Comment (vitorsvieira) [reply tests/neg/i3540.scala]: Added error messages to the comments.

File: compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala [hunk 2]
```diff
@@ -453,6 +452,33 @@ object messages {
...
+          |
+          |$allowedImplicit1
+          |
+          |Or as:
+          |
+          |$allowedImplicit2
+          |
+          |It’s possible to create an implicit class with more than one
+          |non-implicit argument, but such classes are not used during implicit lookup.
+          |""" + implicitClassRestrictionsText
```
allanrenucci: I would say instead:
```scala
val explanation = {
  val example = "implicit class RichDate(date: java.util.Date)"
  hl"""Implicit classes may only take one non-implicit argument in their constructor. For example:
      |
      | $example
      |
      |While it’s possible to create an implicit class with more than one non-implicit argument,
      |such classes aren’t used during implicit lookup.
      |""" + implicitClassRestrictionsText
}
```

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -801,6 +801,23 @@ class ErrorMessagesTests extends ErrorMessagesTest {
...
+        |  implicit class Foo(i: Int, s: String)
+        |}
+      """.stripMargin
+    }
+    .expect { (itcx, messages) =>
+      implicit val ctx: Context = itcx
+
+      assertMessageCount(1, messages)
+      val err :: Nil = messages
+
```
allanrenucci: Remove this white line

tests/neg/i3540.scala
```diff
@@ -0,0 +1,9 @@
+object Test {
+  implicit class Foo(i: Int, s: String) // error: Implicit classes must accept exactly one primary constructor parameter
+  implicit class Foo0                   // error: Implicit classes must accept exactly one primary constructor parameter
+  implicit class Foo1()                 // error: Implicit classes must accept exactly one primary constructor parameter
+  implicit class Foo2()(x: Int)         // error: Implicit classes must accept exactly one primary constructor parameter
+  implicit case class Bar0              // error: A case class must have at least one parameter list
+  implicit case class Bar1()            // error: A case class may not be defined as implicit
+  implicit case class Bar2()(x: Int)    // error: A case class may not be defined as implicit
+}
```
allanrenucci: Add new line at the end of a file

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
vitorsvieira: Perfect! Updated.

tests/neg/i3540.scala
vitorsvieira: Added.

PR #3643, opened 2017-12-09
Micro-optimize Positions

smarter:
- Use final vals to inline constants
- Remove always-true assertions


smarter:
test performance please

PR #3644, opened 2017-12-10
More aggressive valification

gan74:
Valify optimization pass rewritten to be a lot more aggressive.
This implementation replaces every `var` assignation by a `val` def then replaces every (unconditional) access to the mutable variable by an access to it's valified counterpart. If not all accesses can be valified the assignation is kept, otherwise it is discarded.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/localopt/Valify.scala
+++ b/compiler/src/dotty/tools/dotc/transform/localopt/Valify.scala
@@ -7,87 +7,193 @@ import core.Types._
 import core.Flags._
 import ast.Trees._
 import scala.collection.mutable
+import java.util.IdentityHashMap
+import core.NameKinds.LocalOptValifyName
 
-/** Rewrite vars with exactly one assignment as vals.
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/localopt/Valify.scala
```diff
@@ -7,87 +7,193 @@ import core.Types._
...
+        else {
+          // abort the valification for vars that are modified in a nested (ie: potentially conditional) subtree
+          tree.foreachSubTree { 
+            case VarAssign(sym, _) if valified.contains(sym) => 
+              valified -= sym
+            case _ => 
+          }
+
+          if (valified.isEmpty) tree
+          else new TreeMap() {
```
OlivierBlanvillain: Same comment that on #3511, this quadratic traversal needs to be addressed before merge

nicolasstucki:
test performance please

nicolasstucki:
The local optimizations where removed. This PR is no longer needed.

PR #3646, opened 2017-12-10
fix #3645: handle type alias in child instantiation

### LINKED ISSUE #3645: GADT pattern matching unsoundness.
```scala
object App {
    trait AgeT {
      type T
      def subst[F[_]](fa: F[Int]): F[T]
    }
    type Age = Age.T
    final val Age: AgeT = new AgeT {
      type T = Int
      def subst[F[_]](fa: F[Int]): F[T] = fa
    }

    sealed abstract class K[A]
    final case object KAge extends K[Age]
    final case object KInt extends K[Int]

    val kint: K[Age] = Age.subst[K](KInt)
    def get(k: K[Age]): String = k match {
      case KAge => "Age"
    }

    get(kint)
  }
}
```

Produces no warnings but results in a runtime `MatchError` failure.

[Somewhat relevant paper](https://repository.upenn.edu/cgi/viewcontent.cgi?article=1598&context=cis_papers) (it describes a similar problem in Haskell that was solved by the introduction of `role`s, but I am not sure how applicable it is in the context of Scala).

liufengyun:
Fix #3645: handle type alias in child instantiation


smarter:
I think this code would benefit from having more comments explaining what it does :).

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -582,20 +582,31 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
     // precondition: `tp1` should have the shape `path.Child`, thus `ThisType` is always covariant
-    val thisTypeMap = new TypeMap {
-      def apply(t: Type): Type = t match {
+    def childTypeMap(implicit ctx: Context) = new TypeMap {
+      def apply(t: Type): Type = t.dealias match {
         case tp @ ThisType(tref) if !tref.symbol.isStaticOwner  =>
           if (tref.symbol.is(Module)) mapOver(tref)
           else newTypeVar(TypeBounds.upper(tp.underlying))
+        case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
+          // Note that the logic for contra- and co-variance is reverse of `typeParamMap`
```
smarter: typeParamMap is now renamed parentTypeMap :)

PR #3656, opened 2017-12-11
Fix #3636: Better disambiguation in withPrefix

### LINKED ISSUE #3636: Compiler fails to inline
Code snippet minimised from collection strawman:
```scala
trait Iterable[A] {
  def concat[B >: A](that: Iterable[B]): Iterable[B] = ???
  @`inline` final def ++ [B >: A](that: Iterable[B]): Iterable[B] = concat(that)
}

class BitSet extends Iterable[Int] {
  def concat(that: Iterable[Int]): BitSet = ???
  @`inline` final def ++ (that: Iterable[Int]): BitSet = concat(that)
}

class Test {
  def test(x: BitSet, y: Iterable[Int]): Unit = {
    val foo = x ++ y
  }
}
```
```scala
-- Error: tests/allan/Test.scala:68:16 -----------------------------------------
68 |    val foo = x ++ y
   |              ^^^^^^
   |    overloaded alternatives of method concat in class BitSet with types
   |     (that: Iterable[Int]): BitSet
   |     [B >: Int](that: Iterable[B]): Iterable[B] does not take parameters
   | This location is in code that was inlined at tests/allan/Test.scala:68
```
scalac can compile. Dotty can compile without the `inline` annotations.

odersky:
Use the kind of disambiguation in withPrefix that we apply already in memberDenot.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1698,9 +1698,10 @@ object Types {
       finally ctx.typerState.ephemeral |= savedEphemeral
     }
 
-    private def disambiguate(d: Denotation)(implicit ctx: Context): Denotation = {
-      val sig = currentSignature
-      //if (ctx.isAfterTyper) println(i"overloaded $this / $d / sig = $sig") // DEBUG
+    private def disambiguate(d: Denotation)(implicit ctx: Context): Denotation =
+      disambiguate(d, currentSignature)
+
+    private def disambiguate(d: Denotation, sig: Signature)(implicit ctx: Context): Denotation =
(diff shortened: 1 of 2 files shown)


odersky:
> There's one thing I don't quite understand: is it safe to do overloading resolution for inlined names? Is it possible that an inlined name during overloading resolution gets bound to another symbol?

For inlining, they are always the same symbol, since the new prefix has the same widened type as the original one. But the mechanism we are using for inlining (TreeTypeMaps, which use in turn `withPrefix`) can rebind symbols. We could have fixed the problem by fixing the symbol explicitly when inlining. But I thought it was better to make the underlying mechanism better at disambiguating.

PR #3662, opened 2017-12-12
Add TASTY pickling of quotes and implement `~` on quotes

nicolasstucki:
Based on #3634


compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
```diff
@@ -0,0 +1,20 @@
+package dotty.tools.dotc.core.tasty
+
+/** Utils for String representation of TASTY */
+object TastyString {
+
+  /** Decode the TASTY String into TASTY bytes */
+  def stringToTasty(str: String): Array[Byte] = { // TODO factor out this and tastyToString
```
allanrenucci: `str.getBytes`?

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
```diff
@@ -0,0 +1,20 @@
...
+    val bytes = new Array[Byte](str.length)
+    for (i <- str.indices) bytes(i) = str.charAt(i).toByte
+    bytes
+  }
+
+  /** Encode TASTY bytes into a TASTY String */
+  def tastyToString(bytes: Array[Byte]): String = {
```
allanrenucci: `new String(bytes)`?

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
nicolasstucki: No, the string encoding messes up the bytes

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
nicolasstucki: Same problem

nicolasstucki:
@odersky this PR is ready for review. I fixed the issue while unpicking quotes with holes that I mentioned before.

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
odersky: This is a tricky problem. Looking at Stackoverflow, people say you should use a Codec for this, typically Base64. The scheme of mapping all bytes to ranges 0..255 looks like it would work, but it's not optimal. Strings are represented in Classfiles as UTF8 characters, with one byte for ranges 0.127 and two bytes for ranges 128-255. This means that, assuming a uniform bit distribution you get an overhead of 50%. Doing a 8->7 bit codec would give an overhead of less than 15%. 

There's another problem of string size. Strings are limited to 65365 characters. This might not be enough for a larger quoted program. 

scalac solves both of these problems when serializing its pickles as annotations. I think we should copy that scheme. I tried to find it but could not. @retronym @lrytz @adriaanm does one of you have an idea where the code that serializes a Pickle as an annotation is?

odersky: I think we can leave it like this for this PR, but then we should open an issue for future improvements.

nicolasstucki: I will start looking at the alternatives. I also think we should start with this for now to unblock the next PRs and allow people to use it.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -0,0 +1,129 @@
...
+  /** Returns the interpreted result of interpreting the code represented by the tree.
+   *  Return Some of the result or None if some error happen during the interpretation.
+   */
+  def interpretTree[T](tree: Tree)(implicit ct: ClassTag[T]): Option[T] = {
+    try {
+      interpretTreeImpl(tree) match {
+        case obj: T => Some(obj)
+        case obj =>
+          ctx.error(s"Interpreted tree returned a result of an unexpected type. Expected ${ct.runtimeClass} but was ${obj.getClass}", tree.pos)
+          throw new StopInterpretation
```
odersky: just return `None` here.

odersky: We probably should upgrade that to a full type tag, or whatever our equivalent quoted structure is. But that's future work.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -0,0 +1,129 @@
...
+          val msg =
+            if (tree.tpe.derivesFrom(defn.QuotedExprClass)) "Quote needs to be explicit or a call to a static method"
+            else "Value needs to be a explicit or a call to a static method"
+          ctx.error(msg, tree.pos)
+          throw new StopInterpretation
+      }
+    } catch {
+      case ex: NoSuchMethodException =>
+        ctx.error("Could not find interpreted method in classpath: " + ex.getMessage, tree.pos)
+        throw new StopInterpretation
```
odersky: Seems like there's repetition worth factoring out.

odersky: But having two try/catches in `interpretTree` and `interpretTreeImpl` looks like overkill. Maybe propagate everything to `interpretTree`?

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -0,0 +1,129 @@
...
+        case tree: RefTree if tree.symbol.isStatic =>
+          val clazz = classLoader.loadClass(tree.symbol.owner.companionModule.fullName.toString)
+          val method = clazz.getMethod(tree.name.toString)
+          method.invoke(null)
+
+        case tree: RefTree if tree.symbol.is(Module) =>
+          ??? // TODO
+
+        case Inlined(_, bindings, expansion) =>
+          if (bindings.nonEmpty) ??? // TODO evaluate bindings and add environment
```
odersky: Once we address this TODO we could also interpret blocks; it's the same problem.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -0,0 +1,129 @@
...
+          interpretTreeImpl(expansion)
+        case _ =>
+          val msg =
+            if (tree.tpe.derivesFrom(defn.QuotedExprClass)) "Quote needs to be explicit or a call to a static method"
+            else "Value needs to be a explicit or a call to a static method"
```
odersky: In the `else` case it's a QuotedType right? In that case the error message seems strange, however.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -0,0 +1,129 @@
...
+        throw new StopInterpretation
+      case ex: RuntimeException =>
+        ex.printStackTrace()
+        ctx.error("A runtime exception occurred while interpreting: " + ex.getMessage, tree.pos)
+        throw new StopInterpretation
+    }
+  }
+
+  /** List of classes of the parameters of the signature of `sym` */
+  private def paramsSig(sym: Symbol): List[Class[_]] = {
```
odersky: This looks like an extremely roundabout and inefficient way to get what you want!

odersky: The names of a signature are type names, you can compare them with an `eq`.

compiler/src/dotty/tools/dotc/interpreter/RawExpr.scala
```diff
@@ -0,0 +1,7 @@
+package dotty.tools.dotc.interpreter
+
+import dotty.tools.dotc.ast.tpd
+
+class RawExpr(val tree: tpd.Tree) extends quoted.Expr[Any] with RawQuoted {
```
odersky: All top-level classes merit a doc comment. Also, it would be good to have somewhere a diagram that shows the complete hierarchy rooted in `Quoted`.

compiler/src/dotty/tools/dotc/quoted/PickledQuotes.scala
```diff
@@ -0,0 +1,91 @@
...
+  def pickleQuote(tree: Tree)(implicit ctx: Context): String = {
+    if (ctx.reporter.hasErrors) "<error>"
+    else {
+      val encapsulated = encapsulateQuote(tree)
+      val pickled = pickle(encapsulated)
+      TastyString.tastyToString(pickled)
+    }
+  }
+
+  /** Transform the expression into it's fully spliced Tree */
```
odersky: it's -> its

compiler/src/dotty/tools/dotc/quoted/PickledQuotes.scala
```diff
@@ -0,0 +1,91 @@
...
+    unpickled match { // Expects `package _root_ { val ': Any = <tree> }`
+      case PackageDef(_, (vdef: ValDef) :: Nil) => vdef.rhs
+    }
+  }
+
+  /** Encapsulate the tree in a top level `val`
+   *    `<tree>` ==> `package _root_ { val ': Any = <tree> }`
+   *
+   *  Note: Trees for types are also encapsulated this way to preserve the holes in the tree.
+   *        Encapsulating the type of the tree in a `type ' = <tree.tpe>` can potentially
```
odersky: Did not understand why.

compiler/src/dotty/tools/dotc/quoted/QuoteUnpickler.scala
```diff
@@ -0,0 +1,23 @@
+package dotty.tools.dotc.quoted
+
+import dotty.tools.dotc.core.tasty._
+
+object QuoteUnpickler {
```
odersky: Should this be in `tasty` instead of `quoted`?

compiler/src/dotty/tools/dotc/transform/Splicer.scala
```diff
@@ -0,0 +1,35 @@
...
+/** Utility class to slice quoted expressions */
+object Splicer {
+  import tpd._
+
+  /** Splice the Tree for a Quoted expression. `~'(xyz)` becomes `xyz`
+   *  and for `~xyz` the tree of `xyz` is interpreted for which the
+   *  resulting expression is return as a `Tree`
+   */
+  def splice(tree: Tree)(implicit ctx: Context): Tree = tree match {
+    case Apply(quote, quoted :: Nil) if quote.symbol == defn.quoteMethod => quoted
```
odersky: These seem to be used in a few separate places now. Make an extractor?

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1074,6 +1074,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
   }
 
   def typedQuote(tree: untpd.Quote, pt: Type)(implicit ctx: Context): Tree = track("typedQuote") {
+    ctx.compilationUnit.containsQuotesOrSplices = true
```
odersky: Are you sure this is needed? The field should be set in PostTyper anyway.

library/src/scala/quoted/Liftable.scala
```diff
@@ -13,6 +13,23 @@ abstract class Liftable[T] {
...
  */
 object Liftable {
-  implicit def IntIsLiftable: Liftable[Int] = ???
-  implicit def BooleanIsLiftable: Liftable[Boolean] = ???
+
+  sealed abstract class PrimitiveExpr[T] extends Expr[T] {
+    def value: T
+  }
+
+  private class ValueExpr[T <: AnyVal](val value: T) extends PrimitiveExpr[T]
```
odersky: Doesn't look like we need `ValueExpr` as a separate class.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
nicolasstucki: Furthermore, we could also call non-static methods on any object available in the environment.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
nicolasstucki: No. This is what would happen when interpreting an expression which is not a quote, not a constant or not a static method. For example, if we interpret `foo(x: Int)` and we have no way to interpret `x`. I will update the error message to make it clear.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
nicolasstucki: The issue is that the error message should know which method or class was not found which is not accessible in `interpretTree`. Instead I would create local methods for `loadClass` and `invoke` that would catch the Java exceptions and throw the `StopInterpretation` with the error message and position.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
nicolasstucki: Is there an efficient way to get from Scala primitives to java primitive classes?

compiler/src/dotty/tools/dotc/quoted/QuoteUnpickler.scala
nicolasstucki: We already have `TastyUnpickler` that is used inside on the `QuoteUnpickler`, it might be confusing.

compiler/src/dotty/tools/dotc/typer/Typer.scala
nicolasstucki: `tests/run/quote-and-splice` fails otherwise. I will move it to PostTyper.

nicolasstucki: Actually it is not needed anymore. I forgot you added the logic for that in `handleMeta` after my comment on #3634.

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
odersky: Agreed.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
odersky: Ah, I see.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
odersky: OK, makes sense.

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
odersky: `valueTypeNameToJavaType` in Definitions.scala. It's currently commented out, but I see no reason why we could not uncomment it.

compiler/src/dotty/tools/dotc/quoted/QuoteUnpickler.scala
odersky: I don't see it as confusing.

nicolasstucki: Then I'll change it

compiler/src/dotty/tools/dotc/transform/Splicer.scala
nicolasstucki: Yes

library/src/scala/quoted/Liftable.scala
nicolasstucki: I will remove it

compiler/src/dotty/tools/dotc/quoted/PickledQuotes.scala
nicolasstucki: Actually, I managed to encode it into a top level `type`.

nicolasstucki:
All requested changes have been made.

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
lrytz: @odersky https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/backend/jvm/BCodeHelpers.scala#L1036-L1120

nicolasstucki:
Rebased to make sure we do not have regressions.

compiler/src/dotty/tools/dotc/core/tasty/TastyString.scala
nicolasstucki: @lrytz thanks for the link. Could you also point me to the place where the `String`/`Array[Strings]` are converted back into an `Array[Byte]`. Thanks.

lrytz: It took me a while to find it.. Need to clean this up / document. Method [`parseScalaSigBytes`](https://github.com/scala/scala/blob/68d66968eede247adf6d784e0d8f0f343d9124d4/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala#L1125) calls `ConstantPool.getBytes` which goes through `ByteCodecs.decode`.

The encoding is explained here http://www.scala-lang.org/old/sites/default/files/sids/dubochet/Mon,%202010-05-31,%2015:25/Storage%20of%20pickled%20Scala%20signatures%20in%20class%20files.pdf
  - first map all 8-bit bytes to 7 bits (shifting the rest)
  - then increment all by 1 (in 7 bits), so 0x7f becomes 0x00
  - then encode 0x00 as 0xc0 0x80, which is an overlong utf 8 encoding for zero. it's what the jvm classfile spec uses to avoid having 0x00 in strings. it's called "modified utf 8".

the reason for the incrementing by 1 that 0x7f is expected to be less common than 0x00, so the two byte encoding hits less often.

The confusing part is that the class `ScalaSigBytes` used in the backend to encode the signature uses `ByteCodecs.encode8to7`, but does the `+1` itself. It doesn't need to map 0x00 to the two byte version because ASM will do it when writing the annotation to the classfile. However, in the unpickler, we don't use ASM to read the annotation, but just get the bytes from the classfile directly. So there we'll see the two byte encoding. `ByteCodecs.decode` does that work.

PR #3670, opened 2017-12-15
Print diff on space characters with background color

nicolasstucki:


nicolasstucki:
This change will make the diff on whitespaces visible as shown in the example bellow where the indentation was wrong. Previously it was all printed without colors.

screen shot 2017-12-15 at 15 40 47

nicolasstucki:
That was in a previous version. We stopped doing that because on the limited ANSI color palette it was hard to read.

PR #3671, opened 2017-12-15
Split Compiler.phases into smaller logical groups

nicolasstucki:
This change aims to simplify overriding of `phases` in classes that extend `Compiler`. It also splits the macro phases into groups that share a same logical purpose.

### DIFF
--- a/compiler/src/dotty/tools/dotc/Compiler.scala
+++ b/compiler/src/dotty/tools/dotc/Compiler.scala
@@ -40,78 +40,93 @@ class Compiler {
    *     plain SymDenotation, as opposed to a UniqueRefDenotation.
    */
   def phases: List[List[Phase]] =
-    List(
-      List(new FrontEnd),           // Compiler frontend: scanner, parser, namer, typer
-      List(new sbt.ExtractDependencies), // Sends information on classes' dependencies to sbt via callbacks
-      List(new PostTyper),          // Additional checks and cleanups after type checking
-      List(new sbt.ExtractAPI),     // Sends a representation of the API of classes to sbt via callbacks
-      List(new Pickler),            // Generate TASTY info
-      List(new LinkAll),            // Reload compilation units from TASTY for library code (if needed)
-      List(new ReifyQuotes),        // Turn quoted trees into explicit run-time data structures
-      List(new FirstTransform,      // Some transformations to put trees into a canonical form
-           new CheckReentrant,      // Internal use only: Check that compiled program has no data races involving global vars
-           new ElimPackagePrefixes), // Eliminate references to package prefixes in Select nodes
-      List(new CheckStatic,         // Check restrictions that apply to @static members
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -40,77 +40,92 @@ class Compiler {
...
-      List(new Constructors,        // Collect initialization code in primary constructors
-                                    // Note: constructors changes decls in transformTemplate, no InfoTransformers should be added after it
-           new FunctionalInterfaces, // Rewrites closures to implement @specialized types of Functions.
-           new GetClass,            // Rewrites getClass calls on primitive types.
-           new Simplify),           // Perform local optimizations, simplified versions of what linker does.
-      List(new LinkScala2Impls,     // Redirect calls to trait methods defined by Scala 2.x, so that they now go to their implementations
-           new LambdaLift,          // Lifts out nested functions to class scope, storing free variables in environments
+    frontendPhases ::: picklerPhases ::: transformPhases ::: backendPhases
+
+  /** Phases dealing with the the frontend up to trees ready for TASTY pickling */
```
smarter: typo: the the

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -40,77 +40,92 @@ class Compiler {
...
+    List(new sbt.ExtractAPI) ::    // Sends a representation of the API of classes to sbt via callbacks
+    Nil
+
+  /** Phases dealing with TASTY tree pickling and unpickling */
+  protected def picklerPhases: List[List[Phase]] =
+    List(new Pickler) ::           // Generate TASTY info
+    List(new LinkAll) ::           // Reload compilation units from TASTY for library code (if needed)
+    Nil
+
+  /** Phases dealing transforming from pickled trees to backend trees */
```
smarter: "dealing transforming" is not correct grammar

compiler/src/dotty/tools/dotc/Compiler.scala
```diff
@@ -40,77 +40,92 @@ class Compiler {
...
+  protected def transformPhases: List[List[Phase]] =
+    List(new FirstTransform,        // Some transformations to put trees into a canonical form
+         new CheckReentrant,        // Internal use only: Check that compiled program has no data races involving global vars
+         new ElimJavaPackages) ::   // Eliminate syntactic references to Java packages
```
smarter: Why use `::` instead of `List(List(...), ...)` ?

allanrenucci: I personally prefer `List(List(...), ...)` 😄

nicolasstucki: I thought it would make a better distinction between micro and macro phases and it is a little bit more efficient.

nicolasstucki:
Rebased after addition of `ReifyQuotes` phase

PR #3674, opened 2017-12-16
Fix #3669: Treat java repeated arguments like scalac's

### LINKED ISSUE #3669: Ycheck failure after frontend and MegaPhase{firstTransform, checkReentrant, elimJavaPackages}
```scala
import java.nio.file._

class Test {
  def test = {
    val p = Paths.get("")
  }
}
```
```
dotc Test.scala -Ycheck:frontend
dotc Test.scala -Ycheck:elimJavaPackages
```
Both fail

allanrenucci:
Java repeated arguments are now treated as scala repeated arguments. They
are now replaced by a java array during ElimRepeated instead of Typer.

This change let us Ycheck one more MegaPhase. Previously java repeated
arguments would be replaced by a java array during typechecking which
prevented us from retypechecking the tree: Array[T] found, expected T*

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -549,10 +549,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       val args = typedArgBuf.takeRight(n).toList
       typedArgBuf.trimEnd(n)
       val elemtpt = TypeTree(elemFormal)
-      val seqLit =
-        if (methodType.isJavaMethod) JavaSeqLiteral(args, elemtpt)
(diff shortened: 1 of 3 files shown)


smarter:
I think you may want to make a similar change in the unpickler: https://github.com/lampepfl/dotty/blob/88d016c6f0d991698a42bfca2d593f8ec1ae608d/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala#L949-L950

allanrenucci:
@smarter Is this code snippet meant to handle java repeated?
This seems to be dead code. If the arg is a java repeated, then then `readTerm()` will return:
- `Typed(SeqLiteral, <repeated>)` after this fix
- `Typed(JavaSeqLiteral, Array)` before this fix

PR #3665, opened 2017-12-13
Fix #3664: Run `test` in CI instead of `testAll`

allanrenucci:
We remove `testAll` which is now equivalent to `test`. If one only want to run the compilation tests, he can use the `vulpix` command.

This commit also, clean the sbt build

PR #3677, opened 2017-12-18
Update documentation after #3665

allanrenucci:


smarter:
LGTM. Though I find it weird to call this testCompilation when it also runs tests :).

allanrenucci:
Do you a better alternative? I am happy to change the name. Martin wanted something that starts with "test" so that the command can be discovered using autocompletion in sbt

smarter:
testCompiler ?

allanrenucci:
Merging this since `testCompilation` has been around for some time now.

PR #3678, opened 2017-12-19
Remove dead code for unpickling java repeated

allanrenucci:
Repeated arguments of a method call are encoded as
`Typed(SeqLiteral,<repeated>)`. `case SeqLiteral` is dead code.


compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -944,13 +944,7 @@ class TreeUnpickler(reader: TastyReader, nameAtRef: NameRef => TermName, posUnpi
...
-              def readArg() = readTerm() match {
-                case SeqLiteral(elems, elemtpt) if isJava =>
-                  JavaSeqLiteral(elems, elemtpt)
-                case arg => arg
-              }
-              tpd.Apply(fn, until(end)(readArg()))
+              tpd.Apply(fn, until(end)(readTerm()))
             case TYPEAPPLY =>
               tpd.TypeApply(readTerm(), until(end)(readTpt()))
             case TYPED =>
```
smarter: Should the use of JavaSeqLiteral in this case be removed too?

PR #3681, opened 2017-12-19
Implement SIP-23 semantics for isInstanceOf[Singleton] 

smarter:


compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
```diff
@@ -153,6 +153,33 @@ object TypeTestsCasts {
...
+            // Does the expression type contain a union type?
+            var containsUnion: Boolean = false
+
+            def isSingleton(tp: Type): Boolean = tp.dealias match {
+              case tp: SingletonType =>
+                true
+              case AndType(tp1, tp2) =>
+                isSingleton(tp1) || isSingleton(tp2)
+              case tp =>
+                if (tp.isInstanceOf[OrType])
```
odersky: Why not simply:

    caes OrType(tp1, tp2) => isSingleton(tp1) && isSingleton(tp2)

(with a comment that this case currently cannot arise, but is left there anyway for future-proofing)?

smarter: The union of two singleton types might not be a singleton type, e.g. "1 | 2" contains two values, not one.

smarter:
After discussion with @milessabin, I think it's better to just statically disallow `isInstanceOf[Singleton]`. Checking whether or not a type has exactly one value is not trivial, and not very useful anyway.

sjrd:
> Checking whether or not a type has exactly one value is not trivial, and not very useful anyway.

It's not only useless, it's incorrect. `x.isInstanceOf[T]` tests whether the *dynamic* type of `x` is a subtype of `T`, not whether its static type is a subtype of `T`. And I would argue that dynamically, at run-time, any (non-null) `x` has a dynamic type which is `x.type`, which is therefore a subtype of `Singleton`. According to that reasoning, `x.isInstanceOf[Singleton]` should always be equivalent to `x ne null`.

smarter:
> According to that reasoning, x.isInstanceOf[Singleton] should always be equivalent to x ne null.

Unfortunately, in both scalac and dotty, singleton types of reference types are actually doubletons because they're inhabited by null:
```scala
scala> val x = "" 
val x: String = ""
scala> val a: x.type = a 
val a: x.type = null
```

sjrd:
Yes but `x.isInstanceOf[T]`'s specification says that it answers `false` when `x eq null`. For example `null.isInstanceOf[String]` is `false`.

milessabin:
I'm persuaded by @sjrd's argument here. I'm going to update Scala SIP-23 implementation to make `x.isInstanceOf[Singleton]` equivalent to `x ne null`.

milessabin:
Done here: https://github.com/scala/scala/pull/5310/commits/e2f3b0d8f8ea47168008c36a49a82c34838754ba.

PR #3685, opened 2017-12-21
Implement Expr.{run|show}

nicolasstucki:
* Create implicit `Runner` for `run` and `show`
* Implement quote compiler for `run`
* Implement quote decompiler for `show`
* Add `-with-compiler` flag for `dotr`/`repl` to setup the compiler dependencies in the classpath

Based on #3662


nicolasstucki:
Rebased

compiler/test/dotty/tools/dotc/FromTastyTests.scala
```diff
@@ -89,6 +89,7 @@ class FromTastyTests extends ParallelTesting {
          "t493.scala",
          "t8395.scala",
          "t3613.scala",
+         "quote-no-splices.scala",
```
nicolasstucki: I will fix this issue in another PR, it is unrelated to the`Expr.{run|show}` implementation.

nicolasstucki:
Some improvements on the format of the output of `Expr.show` will come with PR #3691.

nicolasstucki:
Test performance please

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/quoted/ExprCompiler.scala
```diff
@@ -0,0 +1,36 @@
...
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.{Mode, Phases}
+import dotty.tools.dotc.core.Phases.Phase
+import dotty.tools.dotc.transform.Pickler
+import dotty.tools.io.VirtualDirectory
+
+/** Compiler that takes the contents of a quoted expression `expr` and produces
+ *  a class file with `class ' { def apply: Object = expr }`.
+ */
+class ExprCompiler(directory: VirtualDirectory) extends Compiler {
```
odersky: Since these are all small files, should ExprRun and ExprFrontEnd be made part of ExprCompiler?

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -203,7 +203,9 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
               // might be a type constructor.
               Checking.checkInstantiable(tree.tpe, nu.pos)
               withNoCheckNews(nu :: Nil)(super.transform(tree))
-            case _ =>
+            case meth =>
+              if (meth.symbol.name.eq(nme.QUOTE) && meth.symbol.owner.eq(defn.OpsPackageClass))
```
odersky: Just `meth.symbol.isQuote` should do. But I don't see why this is needed. Does `handleMeta` in the `Select` case earlier not catch this? Or maybe we need a `handleMeta` for `Ident`s as well?

nicolasstucki: This is half of the fix in #3777. I will explore the different solutions in that PR.

compiler/src/dotty/tools/dotc/quoted/ExprCompiler.scala
nicolasstucki: Good idea. Done.

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
nicolasstucki: `handleMeta` seems to be the simplest solution. I will merge #3777 into this PR.

nicolasstucki:
test performance please

nicolasstucki:
Needs review for the last commits.

nicolasstucki:
test performance please

nicolasstucki:
test performance please

liufengyun:
test performance please

(The machine powered off for unknown reasons, maybe due to electricity problem)

compiler/src/dotty/tools/dotc/CompilationUnit.scala
```diff
@@ -39,13 +40,21 @@ object CompilationUnit {
...
+      val force = new Force
       force.traverse(unit1.tpdTree)
+      unit1.containsQuotesOrSplices = force.containsQuotes
+    }
     unit1
   }
 
   /** Force the tree to be loaded */
-  private object force extends TreeTraverser {
-    def traverse(tree: Tree)(implicit ctx: Context): Unit = traverseChildren(tree)
```
odersky: I don't understand why the change here is necessary. `forceTrees` is true only in the IDE,, right? But in the IDE we never get to ReifyQuotes.

nicolasstucki: No, force tree is also used when compiling from TASTY. 
This covers the case when we recompile from TASTY a file with a quote and no splices.

PR #3686, opened 2017-12-21
Allow completion from source

odersky:
If -sourcepath is set, we now allow completion from sources that are in the canonical place (i.e. fully qualified source name ~ path name of source file relative to source path). If both class and source file exist, the newer one is chosen.

To do: it would be nice to optionally scan all sources so that classes can be found even if their
fully qualified name does not correspond to the file name. Logically, this should be done in `ClassPath.scala`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/Run.scala
+++ b/compiler/src/dotty/tools/dotc/Run.scala
@@ -18,6 +18,7 @@ import transform.TreeChecker
 import rewrite.Rewrites
 import java.io.{BufferedWriter, OutputStreamWriter}
 import printing.XprintMode
+import parsing.Parsers.Parser
 import typer.ImplicitRunInfo
 
 import scala.annotation.tailrec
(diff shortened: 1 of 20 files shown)


compiler/src/dotty/tools/dotc/Run.scala
```diff
@@ -169,6 +170,20 @@ class Run(comp: Compiler, ictx: Context) extends ImplicitRunInfo with Constraint
     runPhases(runCtx)
     if (!ctx.reporter.hasErrors) Rewrites.writeBack()
   }
+  
+  /** Enter top-level definitions of classes and objects contain in Scala source file `file`.
+   *  The newly added symbols replace any previously entered symbols.
+   */
+  def enterRoots(file: AbstractFile)(implicit ctx: Context): Unit =
+    if (!files.contains(file)) {
+      val unit = new CompilationUnit(getSource(file.path))
```
smarter: Shouldn't we add the unit to `units` at this point ?

odersky:
@smarter Do you suggest further changes?

compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
```diff
@@ -98,16 +104,65 @@ class SymbolLoaders {
...
+          case Select(qual: RefTree, name: TermName) => name :: addPrefix(qual, path)
+          case _ => path
+        }
+
+        def enterScanned(unit: CompilationUnit)(implicit ctx: Context) = {
+
+          def checkPathMatches(path: List[TermName], what: String, tree: MemberDef): Boolean = {
+            val ok = filePath == path
+            if (!ok)
+              ctx.warning(i"""$what ${tree.name} is in wrong directory.
```
smarter: "in wrong" -> "in the wrong" ?

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -750,7 +758,7 @@ class Namer { typer: Typer =>
   }
 
   /** The completer of a symbol defined by a member def or import (except ClassSymbols) */
-  class Completer(val original: Tree)(implicit ctx: Context) extends LazyType {
+  class Completer(val original: Tree)(implicit ctx: Context) extends LazyType with SymbolLoaders.SecondCompleter {
```
smarter: Why is `SecondCompleter` needed here now?

compiler/src/dotty/tools/dotc/Run.scala
```diff
@@ -68,6 +69,12 @@ class Run(comp: Compiler, ictx: Context) extends ImplicitRunInfo with Constraint
   private[this] var myUnitsCached: List[CompilationUnit] = _
   private[this] var myFiles: Set[AbstractFile] = _
 
+  /** Units that are added because of source completers but that are not
+   *  compiled in current run.
+   */
+  val lateUnits = mutable.ListBuffer[CompilationUnit]()
```
smarter: Would be nicer if the mutability was hidden from outside this class, like for units and files.

PR #3688, opened 2017-12-21
Add distinction between quoted and unquoted code

nicolasstucki:
screen shot 2017-12-21 at 16 42 14


OlivierBlanvillain:
I'm not so convinced now that I see it... What about

nicolasstucki:
That one was not working on my terminal. It is not widely supported as far as I saw.

nicolasstucki:
After having a look with @OlivierBlanvillain we found out that there is no good ANSI format that works on all terminals an looks good.

Issue #3594, opened 2017-11-29
Compiling pos and run tests from TASTY


nicolasstucki:
These are the tests that still fail from tasty. Several of these tests fail due to same issues (will classify them later). This list was added as tests in #3593 which allows easy reproduction of the issues.

- [ ] tests/pos/Meter.scala
- [ ] tests/pos/NoCyclicReference.scala (issue #3816)
- [ ] tests/pos/depfuntype.scala
- [ ] tests/pos/hklub0.scala
- [ ] tests/pos/i0306.scala (issue #3500)
- [ ] tests/pos/i1365.scala (issue #3500)
- [ ] tests/pos/i1795.scala (issue #3816)
- [ ] tests/pos/i2345.scala
- [ ] tests/pos/i2888.scala (issue #3817)
- [ ] tests/pos/i2944.scala (issue #3500)
- [ ] tests/pos/i2980.scala
- [ ] tests/pos/i3000.scala
- [ ] tests/pos/i536.scala
- [ ] tests/pos/i974.scala (issue #3817)
- [ ] tests/pos/liftable.scala
- [ ] tests/pos/quoteTest.scala
- [ ] tests/pos/quoted.scala
- [ ] tests/pos/stagedInterpreter.scala
- [ ] tests/pos/superacc.scala
- [ ] tests/pos/t0231.scala
- [ ] tests/pos/t1203a.scala
- [ ] tests/pos/t2260.scala
- [ ] tests/pos/t3612.scala
- [ ] tests/pos/t3800.scala
- [ ] tests/pos/t4579.scala
- [ ] tests/pos/t8023.scala (issue #3500?)
- [ ] tests/pos/tcpoly_ticket2096.scala
- [ ] tests/pos/t247.scala
- [ ] tests/run/Course-2002-13.scala
- [ ] tests/run/NestedClasses.scala
- [ ] tests/run/bridges.scala
- [ ] tests/run/eff-dependent.scala
- [ ] tests/run/enum-approx.scala
- [ ] tests/run/i2337.scala
- [ ] tests/run/i2337b.scala
- [ ] tests/run/inlineForeach.scala
- [ ] tests/run/patmat-bind-typed.scala (issue #3500?)
- [ ] tests/run/phantom-decls-1.scala
- [ ] tests/run/phantom-decls-3.scala
- [ ] tests/run/phantom-decls-5.scala
- [ ] tests/run/phantom-hk-1.scala
- [ ] tests/run/phantom-hk-2.scala
[...]

PR #3693, opened 2017-12-22
Add regression tests for #3594

nicolasstucki:


nicolasstucki:
Yes, I will change it to a black list.

PR #3694, opened 2017-12-22
Implement liftables

nicolasstucki:
Based on #3685


nicolasstucki:
Added on top of #3662

library/src/scala/runtime/quoted/Runner.scala
```diff
@@ -0,0 +1,9 @@
+package scala.runtime.quoted
+
+import scala.annotation.implicitNotFound
+import scala.quoted.Expr
+
+@implicitNotFound("Could not find implicit Runner. Default runner can must be imported with `import dotty.tools.dotc.quoted.Runners._`")
```
senia-psm: Typo `can must`

library/src/scala/runtime/quoted/Show.scala
```diff
@@ -0,0 +1,9 @@
+package scala.runtime.quoted
+
+import scala.annotation.implicitNotFound
+import scala.quoted.Expr
+
+@implicitNotFound("Could not find implicit Show. Default runner can must be imported with `import dotty.tools.dotc.quoted.Runners._`")
```
senia-psm: Typo `can must`

library/src/scala/runtime/quoted/Runner.scala
nicolasstucki: Thanks, I will fix it in the other PR that contains this commit.

PR #3698, opened 2017-12-23
Fixes and extensions to symmetric meta programming docs

odersky:


docs/docs/reference/simple-smp.md
```diff
@@ -0,0 +1,229 @@
+# Simple Symmetric Metaprogramming
```
nicolasstucki: This file should have a header like I added in #3696

library/src/scala/quoted/Liftable.scala
```diff
@@ -15,4 +15,5 @@ abstract class Liftable[T] {
 object Liftable {
   implicit def IntIsLiftable: Liftable[Int] = ???
   implicit def BooleanIsLiftable: Liftable[Boolean] = ???
+  implicit def StringIsLiftable: Liftable[String] = ???
```
nicolasstucki: This was already added along with all other constants

PR #3700, opened 2017-12-24
Remove compilation units for classes that have already been loaded

nicolasstucki:
If not, we would create two compilation units that try to load the same class.

### DIFF
--- a/compiler/src/dotty/tools/dotc/fromtasty/ReadTastyTreesFromClasses.scala
+++ b/compiler/src/dotty/tools/dotc/fromtasty/ReadTastyTreesFromClasses.scala
@@ -28,9 +28,12 @@ class ReadTastyTreesFromClasses extends FrontEnd {
             else {
               val unit = CompilationUnit.mkCompilationUnit(clsd, unpickled, forceTrees = true)
               val cls = clsd.symbol.asClass
-              unit.pickled += (cls -> cls.unpickler.unpickler.bytes)
-              cls.unpickler = null
-              Some(unit)
+              if (cls.unpickler == null) None // Has already been loaded by some other compilation unit
+              else {
+                unit.pickled += (cls -> cls.unpickler.unpickler.bytes)
+                cls.unpickler = null
+                Some(unit)
+              }
             }
         }
       }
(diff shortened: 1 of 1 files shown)


nicolasstucki:
Now in both cases the file/class names are deduplicated.

compiler/src/dotty/tools/dotc/Driver.scala
```diff
@@ -26,7 +26,7 @@ class Driver extends DotClass {
     if (fileNames.nonEmpty)
       try {
         val run = compiler.newRun
-        run.compile(fileNames)
+        run.compile(fileNames.distinct)
```
smarter: Not sure this is a great thing to do:
- It doesn't deduplicate between `Foo.scala` and `/a/b/c/Foo.scala`
- List#distinct is O(n^2), that's not great when compiling 10000 files.

nicolasstucki: Then maybe I should keep only the fix on the TASTY ones. With those the compiler crashes. With `dotc Foo.scala Foo.scala` the compiler will emit the error saying that the class is duplicated, which is already fine.

The implementation of `List@distinct` I see is in `SeqLike` and it does look `O(n)`.

smarter: Ah indeed, List#distinct uses a hashset so it's probably more like n*log(n) which is acceptable. But why does the compiler crash when compiling twice the same TASTY file? Could we get it to output an error instead?

compiler/src/dotty/tools/dotc/fromtasty/ReadTastyTreesFromClasses.scala
```diff
@@ -28,9 +28,12 @@ class ReadTastyTreesFromClasses extends FrontEnd {
             else {
               val unit = CompilationUnit.mkCompilationUnit(clsd, unpickled, forceTrees = true)
               val cls = clsd.symbol.asClass
-              unit.pickled += (cls -> cls.unpickler.unpickler.bytes)
-              cls.unpickler = null
-              Some(unit)
+              if (cls.unpickler == null) None // Has already been loaded by some other compilation unit
```
smarter: Maybe make it an error?

nicolasstucki: Yes and done

PR #3710, opened 2017-12-29
Fix #1905: Detect case where bridge would clash with the member

### LINKED ISSUE #1905: Duplicate symbol on bridge generation

odersky:
With value classes it is possible that a bridge is needed because the semi erased
types of a member and its overridden method differ, yet the fully erased types
of the member and the bridge are the same. With this commit we flag this situation
as an error. Before it generated a ClassCastException at runtime instead.

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
@@ -170,6 +170,12 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
       case tp: TypeRef =>
         if (tp.symbol.isAnonymousClass && !ctx.settings.uniqid.value)
           return toText(tp.info)
+        if (tp.symbol.is(Param))
+          tp.prefix match {
+            case pre: ThisType if pre.cls == tp.symbol.owner =>
+              return nameString(tp.symbol)
+            case _ =>
+          }
       case ExprType(result) =>
         return "=> " ~ toText(result)
       case ErasedValueType(tycon, underlying) =>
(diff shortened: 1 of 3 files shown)


odersky:
test performance please

Issue #3658, opened 2017-12-11
HK type aliasing bug


alexknvl:
```scala
object App {
  def main(args: Array[String]): Unit = {
    trait ModuleSig {
      type F[_]
      type Type = F

      def subst[F[_[_]]](fa: F[List]): F[Type]
    }
    val Module: ModuleSig = new ModuleSig {
      type F[A] = List[A]

      def subst[F[_[_]]](fa: F[List]): F[Type] = fa
    }
  }
}
```
doesn't compile, but if you change
```scala
type F[A] = List[A]
```
to
```scala
type F = List
```
it does...

odersky:
I'd argue the program should be rejected. `F[Type]` is a self-application is you follow aliases and as such kind-incorrect.

liufengyun:
@odersky I think there's some missing dealias somewhere, there's no self-application as the two `F` in `F[F]` are different: one from the method type parameter, one from a type member.

smarter:
It works if you do:
```scala
object App {
  def main(args: Array[String]): Unit = {
    trait ModuleSig {
      type F[_]
      type Type = F

      def subst[F[_[_]]](fa: F[[X] => List[X]]): F[Type]
    }
    val Module: ModuleSig = new ModuleSig {
      type F[A] = List[A]

      def subst[F[_[_]]](fa: F[[X] => List[X]]): F[Type] = fa
    }
  }
}
```
This is the same kind of bug as https://github.com/lampepfl/dotty/issues/2989, (I have a potential fix for https://github.com/lampepfl/dotty/issues/2989 at https://github.com/lampepfl/dotty/issues/2989 but it doesn't seem to handle this case. I really wish we could find a more principled way to write hk subtyping checks).

odersky:
> doesn't compile, but if you change

    type F[A] = List[A]

> to

    type F = List

> it does...

Turns out, if you change it to

    type F[+A] = List[A]

is compiles, too. So the question is, is `[A] => List[A]` a subtype of `[+A] => List[A]`? By our normal reasoning about variances, it should not be. The reason it works with `type F = List` is that this definition expands to

    type F[+A] = List[A]

not

    type F[A] = List[A]

PR #3713, opened 2017-12-30
Add missing case to TypeComparer

odersky:
The missing case was for

    hk-type ref <:< hk type lambda

E.g.

    type Y[X]
    Foo[Y] <:< Foo[[T] => Y[T]]

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1065,7 +1065,7 @@ object desugar {
             AppliedTypeTree(ref(seqType), t),
             New(ref(defn.RepeatedAnnotType), Nil :: Nil))
         } else {
-          assert(ctx.mode.isExpr || ctx.reporter.hasErrors, ctx.mode)
+          assert(ctx.mode.isExpr || ctx.reporter.hasErrors || ctx.mode.is(Mode.Interactive), ctx.mode)
           Select(t, op.name)
         }
       case PrefixOp(op, t) =>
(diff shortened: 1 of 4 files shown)


smarter:
Missing testcases? I assume this should fix https://github.com/lampepfl/dotty/issues/3658 and/or https://github.com/lampepfl/dotty/issues/2989 ?

odersky:
Just added the test case for #2989. #3658 still fails.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -475,12 +475,17 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
-          if (!tp1.isHK) {
-            tp2 match {
-              case EtaExpansion(tycon2) if tycon2.symbol.isClass =>
-                return isSubType(tp1, tycon2)
-              case _ =>
-            }
+          if (tp1.isHK) {
+            val tparams1 = tp1.typeParams
+            return isSubType(
+              HKTypeLambda.fromParams(tparams1, tp1.appliedTo(tparams1.map(_.paramRef))),
```
smarter: Can be written `EtaExpansion(tp1)` I think.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1065,7 +1065,7 @@ object desugar {
             AppliedTypeTree(ref(seqType), t),
             New(ref(defn.RepeatedAnnotType), Nil :: Nil))
         } else {
-          assert(ctx.mode.isExpr || ctx.reporter.hasErrors, ctx.mode)
+          assert(ctx.mode.isExpr || ctx.reporter.hasErrors || ctx.mode.is(Mode.Interactive), ctx.mode)
```
smarter: Any idea why we manage to hit this assertion in the IDE?

odersky: No idea, I am afraid.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
odersky: EtaExpansion only works for class type constructors. I tried to change that, but it would be quote involved, so I thought it was simpler to write the code directly here.

PR #3717, opened 2017-12-31
Fix #3647: Define topLevelClass for NoDenotation

### LINKED ISSUE #3647: AssertionError: NoDenotation.owner
```scala
object App {
    trait FooT {
      type T
      type Bar[A]

      def get(k: Bar[T]): String
    }
    val test: FooT = new FooT {
      type T = String
      type Bar[A] = J[A]
      sealed abstract class J[A]
      final case object JName extends J[T]
      final case object JInt extends J[Int]

      def get(k: J[T]): String = k match {
        case JName => "Age"
      }
    }
  }
}
```

kills the compiler with:
```scala
error when pickling type {...}.this.J
error when pickling type {...}.this.J[A]
error when pickling type {...}.this.J
error when pickling type  = {...}.this.J
error when pickling type FooT{T = String; Bar = {...}.this.J}
error when pickling type {z1 => FooT{T = String; Bar = z1.J}}
error when pickling tree {z1 => FooT{T = String; Bar = z1.J}}
error when pickling tree new Object with FooT{...}(): {z1 => FooT{T = String; Bar = z1.J}}
...
Exception in thread "main" java.lang.AssertionError: NoDenotation.owner
	at dotty.tools.dotc.core.SymDenotations$NoDenotation$.owner(SymDenotations.scala:1874)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.topLevel$1(SymDenotations.scala:928)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.topLevelClass(SymDenotations.scala:931)
	at dotty.tools.dotc.core.tasty.TreePickler.isLocallyDefined(TreePickler.scala:82)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleNewType(TreePickler.scala:191)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleType(TreePickler.scala:132)
	at dotty.tools.dotc.core.tasty.TreePickler.$anonfun$pickleNewType$1(TreePickler.scala:148)
	at dotty.tools.dotc.core.tasty.TreePickler.withLength(TreePickler.scala:32)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleNewType(TreePickler.scala:148)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleType(TreePickler.scala:132)
	at dotty.tools.dotc.core.tasty.TreePickler.$anonfun$pickleMethodic$1(TreePickler.scala:264)
	at dotty.tools.dotc.core.tasty.TreePickler.withLength(TreePickler.scala:32)
	at dotty.tools.dotc.core.tasty.TreePickler.pickleMethodic(TreePickler.scala:263)
...
```

odersky:


compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -925,7 +925,7 @@ object SymDenotations {
     /** The top-level class containing this denotation,
      *  except for a toplevel module, where its module class is returned.
      */
-    final def topLevelClass(implicit ctx: Context): Symbol = {
+    def topLevelClass(implicit ctx: Context): Symbol = {
```
smarter: Why not do `if (!exists) symbol` instead of adding an override? That's more consistent with other definitions in this class like `lexicallyEnclosingClass`

odersky: Agreed, I'll change it.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -925,11 +925,14 @@ object SymDenotations {
     /** The top-level class containing this denotation,
      *  except for a toplevel module, where its module class is returned.
      */
-    final def topLevelClass(implicit ctx: Context): Symbol = {
-      def topLevel(d: SymDenotation): Symbol = {
-        if (d.isEffectiveRoot || (d is PackageClass) || (d.owner is PackageClass)) d.symbol
-        else topLevel(d.owner)
-      }
+    def topLevelClass(implicit ctx: Context): Symbol = {
```
allanrenucci: `final`?

PR #3718, opened 2017-12-31
Fix #3630: Handle classtag searches with wildcard arguments

### LINKED ISSUE #3630: crash on compilable code: no TypeBounds allowed

$ dotc -version
```
Dotty compiler version 0.5.0-bin-SNAPSHOT-git-4fefb64 -- Copyright 2002-2017, LAMP/EPFL
```

$ dotc abc.scala
```

Exception in thread "main" java.lang.AssertionError: no TypeBounds allowed
	at dotty.tools.dotc.core.TypeApplications$.noBounds(TypeApplications.scala:29)
	at dotty.tools.dotc.core.TypeApplications$.argTypes$extension$$anonfun$1(TypeApplications.scala:447)
	at dotty.tools.dotc.core.TypeApplications$.argTypes$extension(TypeApplications.scala:447)
	at dotty.tools.dotc.typer.Implicits.synthesizedClassTag$1(Implicits.scala:551)
	at dotty.tools.dotc.typer.Implicits.inferImplicitArg(Implicits.scala:629)
	at dotty.tools.dotc.typer.Implicits.synthesizedClassTag$1(Implicits.scala:555)
	at dotty.tools.dotc.typer.Implicits.inferImplicitArg(Implicits.scala:629)
	at dotty.tools.dotc.typer.Typer.implicitArgs$2(Typer.scala:2021)
	at dotty.tools.dotc.typer.Typer.addImplicitArgs$1(Typer.scala:2034)
	at dotty.tools.dotc.typer.Typer.adaptNoArgsImplicitMethod$2(Typer.scala:2086)
	at dotty.tools.dotc.typer.Typer.adaptNoArgs$1(Typer.scala:2202)
	at dotty.tools.dotc.typer.Typer.adaptInterpolated(Typer.scala:2347)
<snipped>
```
$ cat abc.scala
```scala
class U { Array(Array(2), Array("a")) }
```

odersky:


compiler/src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -320,6 +320,7 @@ object TypeErasure {
         case _ => false
       }
     case tp: TypeParamRef => false
+    case tp: TypeBounds => false
```
smarter: Would `hasStableErasure(tp.lo) && hasStableErasure(tp.hi)` be incorrect?

tests/neg/i3630.scala
```diff
@@ -0,0 +1 @@
+class U { Array(Array(2), Array("a")) } // error: no classtag
```
smarter: Do we really need to disallow this? It seems to work OK in scalac.

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
odersky: No, because we fall into this case also via TypeProxy.

tests/neg/i3630.scala
odersky: The full error message says:

    No ClassTag available for Array[_ >: Int & String <: Int | String]

The inferred type of the outer array is

    Array[Array[_ >: Int & String <: Int | String]]

That is indeed the best type. But there's no ClassTag. We *could* infer some wider type such as 

    Array[AnyRef](Array(2), Array("a"))

for the outer array. But it would be a special case. And I am not sure whether in the end it would not be better to reject this weird construction as an error and demand that the user write

    Array[AnyRef](Array(2), Array("a"))

explicitly.

PR #3719, opened 2017-12-31
Fix #3624: Lazily evaluate static module accessors

### LINKED ISSUE #3624: Objects in traits are instantiated on every access
```scala
trait T {
  case object Foo
}

object Bar extends T

object Test {
    assert(Bar.Foo == Bar.Foo) // false
  }
}
```

Looking at the bytecode we see than `Bar.Foo` calls `T.Foo` which creates a new instance on every call:

```java
public interface T {
    }

    default public Foo. Foo() {
        return new Foo.();
    }
}

public final class Bar$
implements T {
    public static final Bar$ MODULE$;

    public static {
        new Bar$();
    }

    public Bar$() {
        MODULE$ = this;
    }

    @Override
    public final T.Foo. Foo() {
        return T.super.Foo();
    }
}
```

For comparison, this the code generated by scalac:

```java
public interface T {
    public Foo. Foo();

    default public static void $init$(T $this) {
    }
}

public final class Bar$
implements T {
    public static Bar$ MODULE$;
    private volatile T.Foo. Foo$module;

    public static {
        new Bar$();
    }

    @Override
    public T.Foo. Foo() {
        if (this.Foo$module == null) {
            this.Foo$lzycompute$1();
        }
        return this.Foo$module;
    }

    private final void Foo$lzycompute$1() {
        Bar$ bar$ = this;
        synchronized (bar$) {
            if (this.Foo$module == null) {
                this.Foo$module = new T.Foo.(null);
            }
        }
    }

    private Bar$() {
        MODULE$ = this;
        T.$init$(this);
    }
}
```

odersky:
We did not use lazy evaluation for static objects, because these are usually evauated
lazily by the backend. But that does not hold if the static module is an (accessor) def
instead of a val.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/LazyVals.scala
+++ b/compiler/src/dotty/tools/dotc/transform/LazyVals.scala
@@ -62,8 +62,8 @@ class LazyVals extends MiniPhase with IdentityDenotTransformer {
   def transformLazyVal(tree: ValOrDefDef)(implicit ctx: Context): Tree = {
     val sym = tree.symbol
     if (!(sym is Flags.Lazy) ||
-        sym.owner.is(Flags.Trait) ||
-        (sym.isStatic && sym.is(Flags.Module, butNot = Flags.Method)))
+        sym.owner.is(Flags.Trait) || // val is accessor, lazy field will be implemented in subclass
+        (sym.isStatic && sym.is(Flags.Module, butNot = Flags.Method))) // static module vals are implemented in the JVM by lazy loading
       tree
     else {
       val isField = sym.owner.isClass
--- a/compiler/src/dotty/tools/dotc/transform/LazyVals.scala
+++ b/compiler/src/dotty/tools/dotc/transform/LazyVals.scala
@@ -36,7 +36,7 @@ class LazyVals extends MiniPhase with IdentityDenotTransformer {
   class OffsetInfo(var defs: List[Tree], var ord:Int)
   val appendOffsetDefs = mutable.Map.empty[Symbol, OffsetInfo]
 
-  override def phaseName: String = "LazyVals"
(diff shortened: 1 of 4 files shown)


smarter:
The issue number referenced in the commit message is wrong, it should be https://github.com/lampepfl/dotty/issues/3624, not https://github.com/lampepfl/dotty/pull/3634

compiler/src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -61,7 +61,10 @@ class LazyVals extends MiniPhase with IdentityDenotTransformer {
 
   def transformLazyVal(tree: ValOrDefDef)(implicit ctx: Context): Tree = {
     val sym = tree.symbol
-    if (!(sym is Flags.Lazy) || sym.owner.is(Flags.Trait) || (sym.isStatic && sym.is(Flags.Module))) tree
+    if (!(sym is Flags.Lazy) ||
+        sym.owner.is(Flags.Trait) ||
+        (sym.isStatic && sym.is(Flags.Module, butNot = Flags.Method)))
+      tree
```
smarter: Maybe add a comment here explaining why this is OK in these cases?

tests/run/i3634.scala
```diff
@@ -0,0 +1,11 @@
+trait T {
+  case object Foo
+}
+
+object Bar extends T
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    assert(Bar.Foo == Bar.Foo) // false
```
smarter: I would use `eq` instead of `==` to be extra sure that nothing funny is going on.

smarter: Also add a `object Baz extends T` and `assert(Bar.Foo ne Baz.Foo)`

Issue #3427, opened 2017-11-01
@unchecked annotation propagated to return type of the pattern match expression


allanrenucci:
```scala
scala> val foo = List(1) match { case x: List[Int] @unchecked => x } 
val foo: List[Int] @unchecked = List(1)
```
`@unchecked` shouldn't show up here, the unchecked annotation on a pattern match expression is meant to be seen only by the pattern match checker and shouldn't be propagated to the return type of the pattern match

odersky:
Will be fixed when #3723 is in.

PR #3723, opened 2018-01-01
Fix #3564: Allow annotated singleton types

### LINKED ISSUE #3564: Annotations shadow singleton types
There is a weird interaction between annotations and singleton types. In the following, the singleton type is lost when it shouldn't:

```scala
scala> def g(arg: "abc" @uncheckedVariance) = 1
def f(t: String @uncheckedVariance): Int
```
If `"abc"` is put in a type alias it's survives:

```scala
import scala.annotation.unchecked.uncheckedVariance
scala> type ST = "abc"
// defined alias type ST = abc
scala> def f(arg: ST @uncheckedVariance) = 1
def f(t: abc @uncheckedVariance): Int
```

This problems manifests itself when trying to create a case class with a singleton typed field, the copy method doesn't typecheck because of the above:

```scala
scala> case class Foo(s: "abc")
1 |case class Foo(s: "abc")
  |                       ^
  |                       found:    String @uncheckedVariance
  |                       required: String("abc")
  |
```

odersky:
Previously, singleton types were widened before an annotation was added. Now such
a widening is no longer performed. The change necessitated a systematic generalization
of what it means to be a singleton, where we now use in many places
an extractor instead of a type test. This has the added benefit that we now also
classify as singletons intersections involving a singleton type and aliases to a singleton
type.


odersky:
test performance please

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -802,7 +802,7 @@ object PatternMatcher {
           }
 
           expectedTp.dealias match {
-            case expectedTp: SingletonType =>
+            case SingletonType(expectedTp) =>
```
smarter: This is wrong because the extractor for SingletonType will take only one branch of an AndType:
```scala
trait Foo {
  def hi: Int
}

object Test {
  def foo(x: Any) = x match {
    case x: ("1" & Foo) =>
      x.hi
    case _ =>
  }

    foo("1")
  }
}
```
This code used to run correctly, but now crashes with `java.lang.ClassCastException: java.lang.String cannot be cast to Foo`. In general I would rather not consider `AndType` to be singleton types, I think it's too easy to get incorrect semantics that way like what's happening here.

odersky: I see. It looks easiest to revert and not let AndTypes be SingletonTypes, even though semantically it would be nice if they could be.

odersky:
test performance please

odersky:
@smarter I changed the scheme so that AndTypes no longer qualify as singletons

odersky:
> Somehow we now loose annotations in some situations

Yes, it's because widen drops annotations and local type inference does a widen. It is a change of behavior but it's not clear that the old behavior was the correct one. I tried to leave the annotation on when widening but that's unintuitive (for comparison, we drop the annotation on dealiasing), and it also caused a lot of failures e.g. in compileStdLib. So I think the new scheme is the least bad one.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1593,7 +1593,7 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
     if (ctx.mode is Mode.Type)
       assignType(cpy.Annotated(tree)(arg1, annot1), arg1, annot1)
     else {
-      val tpt = TypeTree(AnnotatedType(arg1.tpe.widen, Annotation(annot1)))
+      val tpt = TypeTree(AnnotatedType(arg1.tpe, Annotation(annot1)))
```
smarter: Actually thinking about this more, I think we should keep the `.widen` here.
I think we can only reach this case when the user write:
```scala
val x = 1: @unchecked
```
In this case I think it's fine to interpret this code as if the user had written:
```scala
val x = 1: Int @unchecked
```
If the user really wants to put the annotation on the singleton type, which is rather unusual, he can still explicitly write:
```scala
val x = 1: 1 @unchecked
```
This one will work fine because of the `.widen` which was dropped in `TypeAssigner`

odersky: But that means #3564 fails, because we have precisely the case where we annotate a singleton with @uncheckedVariance. 

Also, I would argue that someone writing   

    expr: @unchecked

is really annotating the expression `expr`, not its type. The type is an artifact of the way we
express annotated expressions. So it is better that the annotation does _not_ leak into inferred types of the lhs, say. See also #3427, which is another bug that would be fixed by this PR, but would not be fixed by the proposed change.

smarter: > But that means #3564 fails, because we have precisely the case where we annotate a singleton with @uncheckedvariance.

I don't think so, in #3564 a singleton type is annotated, so `ctx.mode is Type` is true, this case is handled two lines above the current one, and only the `.widen` that was in `TypeAssigner` is relevant there.

> So it is better that the annotation does not leak into inferred types of the lhs, say.

OK. I'm fine with this but note that this is a change of behavior with respect to Scala 2.

odersky: No, I changed the logic in Typer only, and #3564 started failing again.

PR #3729, opened 2018-01-02
Fix #3532: Drop -Yinfer-argument-types

### LINKED ISSUE #3532: -Yinfer-argument-types is not rejected as bad option but code using it causes the typechecker to crash
Trying to compile in dotty some code that compiles fine in scalac I found this issue. Dottyc does warn about other invalid settings but not this one, and code like

 ```scala
new EventHandler[Evt] {
  override def handle(evt) = ???
}
```

causes the typechecker to crash.

odersky:


propensive:
Wow, this is a *relic* from another era! Don't worry: I'm not using it!

Issue #2906, opened 2017-07-22
Syntax error with Enums


allanrenucci:
This fails to compile
```scala
enum Foo {
  case Bar
  private case Bat
}
```
```scala
-- [E039] Syntax Error: <console>:3:2 ------------------------------------------
3 |  private case Bat
  |  ^^^^^^^
  |  ';' expected, but 'private' found
```
It can be reproduced with any modifier.

biboudis:
Thanks @allanrenucci. Modifiers are currently not supported with Enums.

Enums are discussed [here](http://dotty.epfl.ch/docs/reference/enums/desugarEnums.html).

I am marking it as a bug as the error needs to be adjusted.

allanrenucci:
Note that dotc is fine with both
```scala
enum Foo {
  case Bar;
  private case Bat
}
```
and 
```scala
enum Foo {
  private case Bar
} 
```
So the reported error is not the only issue here.

nicolasstucki:
Is it actually making `Bar`/`Bat` private? or is the modifier ignored.
What is the use case for a private enum?

allanrenucci:
This is what I get in the repl
```scala
scala> enum Foo { case Bar; private case Bat } 
defined class Foo
defined module Foo
scala> Foo.Bar 
val res1: Foo = Bar
scala> Foo.Bat 
-- Error: <console>:6:4 --------------------------------------------------------
6 |Foo.Bat
  |^^^^^^^
  |value Bat in object Foo cannot be accessed as a member of Foo.type from module class $iw$.
```
I used `private` to showcase the issue. It can be reproduced with any modifier. I can see use cases for annotations (e.g. `@deprecated`)

biboudis:
I wonder what happens in the case of pattern matching over mixed enum cases.

ping @odersky

PR #3730, opened 2018-01-02
Fix #3460: Fix semicolon insertion confusion after enum case

### LINKED ISSUE #3460: Syntax error when putting `def` in enum object after simple `case`
This doesn't compile:
```scala
enum class Ducks

object Ducks {
  case Dewey

  def wooohoo: Int = 1
}
```
```scala
-- [E039] Syntax Error: enumo.scala:6:6 ------------------------------------
6 |  def woohoo: Int = 1
  |      ^^^^^^
  |      ';' expected, but identifier found
```
This can be worked around by writing `case Dewey;` or `case Dewey extends Ducks`.

odersky:
After seeing an enum case, the Scanner thought it is in a pattern, and
therefore does not do semicolon insertion on newlines. We now get it
out of that mode by pretending to be passed the `=>`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -2055,6 +2055,7 @@ object Parsers {
         typeDefOrDcl(start, posMods(start, mods))
       case CASE =>
         enumCase(start, mods)
+          .reporting(t => i"case $t, current = ${in.show} / ${in.sepRegions}%, %")
       case _ =>
         tmplDef(start, mods)
     }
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2060,6 +2060,7 @@ object Parsers {
         typeDefOrDcl(start, posMods(start, mods))
       case CASE =>
         enumCase(start, mods)
+          .reporting(t => i"case $t, current = ${in.show} / ${in.sepRegions}%, %")
```
smarter: Stray debug statement.

allanrenucci:
Also fixes #2906

PR #3733, opened 2018-01-02
Fix #3433: Fix determining typeParams for structural type members

### LINKED ISSUE #3433: Incorrect kind of type projection
```scala
scala> type Const[t] = { type L[x] = t }
defined type alias Const

scala> type Bar[a] = Const[Int]#L[a] 
-- [E052] Reference Error: <console>:5:15 --------------------------------------
5 | type Bar[a] = Const[Int]#L[a]
  |               ^^^^^^^^^^^^^^^
  |               Const[Int]#L does not take type parameters
```

Interestingly if we "partially apply" this `Const` the result is neither in `*` nor in `* → *`:

```scala
scala> type Foo = Const[Int]#L
defined type alias Foo

scala> val a: Foo[Int] = 1
-- [E052] Reference Error: <console>:6:7 ---------------------------------------
6 |val a: Foo[Int] = 1
  |       ^^^^^^^^
  |       Foo does not take type parameters

scala> val a: Foo = 1
-- [E055] Syntax Error: <console>:6:7 ------------------------------------------
6 |val a: Foo = 1
  |       ^^^
  |       missing type parameter for Foo
```

(Note that in Dotty we would just define `Const` as `type Const[C] = [X] => C` which works as expected, the above is a way to encode a constant type function in scalac)

odersky:


compiler/src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -172,6 +172,7 @@ class TypeApplications(val self: Type) extends AnyVal {
       case self: TypeRef =>
         val tsym = self.symbol
         if (tsym.isClass) tsym.typeParams
+        else if (!tsym.exists) self.info.typeParams
```
smarter: .How come we don't have the symbol for the refinement member here? That seems likely to be problematic in a lot of other cases, for example in the IDE how do we do "go to definition" if the tree has no symbol?

odersky: Refinements of types that do not exist in the baseclass don't have a symbol. Which symbol could they have?

PR #3734, opened 2018-01-02
Fix #3383: Don't fail on unreported errors

### LINKED ISSUE #3383: `posTwice` `testNonCyclic` fails if `dotc.CompilationUnit` does not import `Types.Type`
If we remove the unused import `import dotty.tools.dotc.core.Types.Type` in `dotty.tools.dotc.CompilationUnit` the `testNonCyclic` fails in `posTwice`.

```scala
-- [E006] Unbound Identifier Error: ../compiler/src/dotty/tools/dotc/ast/Trees.scala:817:42 
817 |  abstract class Instance[T >: Untyped <: Type] extends DotClass { inst =>
    |                                          ^^^^
    |                                          not found: type Type
```

Note that the error manifests itself in another class.

odersky:
#3383 shows that there are valid cases where an error type is assigned to a tree
before the error is reported. The error will then be reported later in `adapt`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -112,9 +112,9 @@ object Trees {
      *  type. (Overridden by empty trees)
      */
     def withType(tpe: Type)(implicit ctx: Context): ThisTree[Type] = {
-      if (Config.checkUnreportedErrors)
-        if (tpe.isInstanceOf[ErrorType]) assert(ctx.reporter.errorsReported)
-      if (Config.checkTreesConsistent)
+      if (tpe.isInstanceOf[ErrorType])
+        assert(!Config.checkUnreportedErrors || ctx.reporter.errorsReported)
+      else if (Config.checkTreesConsistent)
         checkChildrenTyped(productIterator)
       withTypeUnchecked(tpe)
     }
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -112,9 +112,9 @@ object Trees {
      *  type. (Overridden by empty trees)
      */
     def withType(tpe: Type)(implicit ctx: Context): ThisTree[Type] = {
-      if (tpe.isInstanceOf[ErrorType])
-        assert(ctx.mode.is(Mode.Interactive) || ctx.reporter.errorsReported)
-      else if (Config.checkTreesConsistent)
+      if (Config.checkUnreportedErrors)
+        if (tpe.isInstanceOf[ErrorType]) assert(ctx.reporter.errorsReported)
```
nicolasstucki: Why drop the `ctx.mode.is(Mode.Interactive)`?

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -112,9 +112,9 @@ object Trees {
...
+      if (Config.checkTreesConsistent)
```
nicolasstucki: Now, if `tpe.isInstanceOf[ErrorType]` is `true` we would go into this branch. Is this intentional?

odersky: Good point. No, we need to fix that.

compiler/src/dotty/tools/dotc/ast/Trees.scala
odersky: No need I think. Since Config.checkUnreportedErrors is by default `false`, we don't need the additional test. If someone turns Config.checkUnreportedErrors on for the IDE I assume they know what they are doing.

PR #3735, opened 2018-01-02
Fix #3339: Disallow accesses to private package members from nested pkgs

### LINKED ISSUE #3339: Missing access checks when accessing package private classes

odersky:
Change of the accessibility rules: Disallow accesses to private package members from nested packages.

Such accesses were causing an access error at runtime before.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -674,10 +674,7 @@ object SymDenotations {
     final def isAccessibleFrom(pre: Type, superAccess: Boolean = false, whyNot: StringBuffer = null)(implicit ctx: Context): Boolean = {
 
       /** Are we inside definition of `boundary`? */
-      def accessWithin(boundary: Symbol) =
-        ctx.owner.isContainedIn(boundary) &&
-          (!(this is JavaDefined) || // disregard package nesting for Java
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -733,7 +730,9 @@ object SymDenotations {
              (  !(this is Local)
              || (owner is ImplClass) // allow private local accesses to impl class members
              || isCorrectThisType(pre)
-             )
+             ) &&
```
smarter: This long chain of conditions is getting really hard to read, factor it out into a few vals?

odersky: I tried but it does not seem to help. The key is treating !A || B as an implication then things get clearer. I was wondering whether we should add an inline helper method `implies` on `Boolean` to do this. But it's quite a lot of effort.

PR #3739, opened 2018-01-03
Fix #3736: Fix owner checking logic

### LINKED ISSUE #3736: `scala.language.dynamics` import is "shadowed" by any other import
The `scala.language.dynamics` implicit for enabling use of the `Dynamic` trait appears to be strangely "shadowed" by *any* subsequent import.

This does not compile,
```scala
import scala.language.dynamics
import scala.List
class Foo() extends Dynamic
```
with
```
dotc test.scala
```
whereas omitting
```
import scala.List
```
will allow it to compile, as will reordering the imports like so,
```scala
import scala.List
import scala.language.dynamics
class Foo() extends Dynamic
```

Compiler version:
```
Dotty compiler version 0.5.0-RC1 -- Copyright 2002-2017, LAMP/EPFL
```

Compiler output:
```
-- Error: test.scala:3:6 -------------------------------------------------------
3 |class Foo() extends Dynamic
  |^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |extension of type scala.Dynamic needs to be enabled
  |by making the implicit value scala.language.dynamics visible.
  |This can be achieved by adding the import clause 'import scala.language.dynamics'
  |or by setting the compiler option -language:dynamics.
  |See the Scala docs for value scala.language.dynamics for a discussion
  |why the feature needs to be explicitly enabled.
one error found
```

nicolasstucki:


smarter:
LGTM but I'd like @odersky to double-check this one.

odersky:
It's not possible to remove the test as in the original PR because otherwise e.g.

    import foo.dynamic

would also count. The logic of `hasImport` needs to be changed instead.

compiler/src/dotty/tools/dotc/core/TypeOps.scala
```diff
@@ -292,16 +292,16 @@ trait TypeOps { this: Context => // TODO: Make standalone object.
       if (!sym.exists || (sym eq defn.LanguageModuleClass)) ""
       else toPrefix(sym.owner) + sym.name + "."
     def featureName = toPrefix(owner) + feature
-    def hasImport(implicit ctx: Context): Boolean = {
-      if (ctx.importInfo == null || (ctx.importInfo.site.widen.typeSymbol ne owner)) false
-      else if (ctx.importInfo.excluded.contains(feature)) false
```
smarter: Previously, we stopped and returned false for an unimported feature, but we now continue looking up the chain for an import, this breaks the last three errors in the neg test https://github.com/lampepfl/dotty/blob/master/tests/neg/dynamicNoImport.scala

odersky: Yes, indeed, the logic is quite subtle here (and it was not documented either).

nicolasstucki:
@smarter @odersky I reimplemented the fix, added an `import foo.dynamic` test case and all tests pass.

PR #3741, opened 2018-01-03
Add `’` as an alias of `'` for quotes

nicolasstucki:


smarter:
Do we really want to do this? I don't think I can type `’` with my keyboard layout and I'm far from the only one.

smarter:
Oh wait, actually I found where it is on my layout :). But it's not part of the standard US QWERTY layout many people use.

nicolasstucki:
The idea is to support it in case someone copy pastes it from the documentation.

smarter:
If we use it in the documentation people will be confused when they can't figure out how to write it themselves.

PR #3744, opened 2018-01-03
Add error message for polymorphic methods in structured refinements

Wojtechnology:
Addresses `Typer.scala:1109` in https://github.com/lampepfl/dotty/issues/1589. Polymorphic methods are not allowed in structural refinements as explained here: https://github.com/lampepfl/dotty/blob/88d016c6f0d991698a42bfca2d593f8ec1ae608d/docs/docs/reference/changed/structural-types.md#notes

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2064,4 +2064,14 @@ object messages {
     val explanation =
       hl"An object that contains ${"@static"} members must have a companion class."
   }
+
+  case class PolymorphicMethodMissingTypeInParent(rsym: Symbol, parentTpt: tpd.Tree)(implicit ctx: Context)
+  extends Message(PolymorphicMethodMissingTypeInParentID) {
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2064,4 +2064,14 @@ object messages {
...
       hl"An object that contains ${"@static"} members must have a companion class."
   }
+
+  case class PolymorphicMethodMissingTypeInParent(rsym: Symbol, parentTpt: tpd.Tree)(implicit ctx: Context)
+  extends Message(PolymorphicMethodMissingTypeInParentID) {
+    val kind = "Syntax"
+    val msg = hl"polymorphic refinement $rsym without matching type in parent $parentTpt is no longer allowed"
+    val explanation =
+      hl"""Polymorphic $rsym is not allowed in the structural refinement of $parentTpt because
+          |$rsym does not override any symbol in $parentTpt. Structural refinement does not allow for
```
allanrenucci: any **method** in ...

PR #3746, opened 2018-01-04
Added error message for Symbol is not a value #1589

benkobalog:
1. Added SymbolIsNotAValue error Message class 
2. Substituted the original error message at Checking.scala:526 with the new class and added a test
3. Typo fix at Checking.scala:211

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2068,6 +2068,21 @@ object messages {
   case class JavaSymbolIsNotAValue(symbol: Symbol)(implicit ctx: Context) extends Message(JavaSymbolIsNotAValueID) {
     val msg = hl"$symbol is not a value"
     val kind = "Type Mismatch"
-    val explanation = hl"Java packages cannot be used as a value"
+    val explanation = {
+      val javaCodeExample = """class A {public static int a() {return 1;}}"""
+
+      val scalaCodeExample =
+        """val objectA = A     // This does not compile
+          |val aResult = A.a() // This does compile""".stripMargin
+
+      hl"""Java statics and packages cannot be used as a value.
+          |For Java statics consider the following Java example:
+          |
+          |$javaCodeExample
+          |
+          |When used from Scala:
+          |
+          |$scalaCodeExample"""
+    }
   }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -208,7 +208,7 @@ object messages {
 
   case class WildcardOnTypeArgumentNotAllowedOnNew()(implicit ctx: Context)
   extends Message(WildcardOnTypeArgumentNotAllowedOnNewID) {
-    val kind = "syntax"
+    val kind = "Syntax"
     val msg = "type argument must be fully defined"
 
     val code1 =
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -208,7 +208,7 @@ object messages {
 
   case class WildcardOnTypeArgumentNotAllowedOnNew()(implicit ctx: Context)
   extends Message(WildcardOnTypeArgumentNotAllowedOnNewID) {
-    val kind = "syntax"
+    val kind = "Syntax"
```
benkobalog: This was the only place where this string started with a lowercase. I assumed it was a typo, if somebody could confirm it, it would be good.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2064,4 +2064,11 @@ object messages {
     val explanation =
       hl"An object that contains ${"@static"} members must have a companion class."
   }
+
+  case class SymbolIsNotAValue(symbol: Symbol)(implicit ctx: Context) extends Message(SymbolIsNotAValueID) {
+    val msg = hl"${symbol.show} is not a value"
+    val kind = "Type Mismatch"
+    val explanation =
+      hl"Scala or Java packages cannot be assigned to a value"
```
nicolasstucki: Scala packages can be used as values if they have a package object.

`Java statics and packages cannot be used value.`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2064,4 +2064,11 @@ object messages {
     val explanation =
       hl"An object that contains ${"@static"} members must have a companion class."
   }
+
+  case class SymbolIsNotAValue(symbol: Symbol)(implicit ctx: Context) extends Message(SymbolIsNotAValueID) {
```
nicolasstucki: `SymbolIsNotAValue ` -> `JavaSymbolIsNotAValue`
Same for ID.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
nicolasstucki: Good

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2064,4 +2064,11 @@ object messages {
...
+    val msg = hl"${symbol.show} is not a value"
```
allanrenucci: You don't need show here `$symbol`.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
benkobalog: What should be the explanation then?
What are Java statics in this context? I guess not static fields because """object O {val a = Math.PI}""" compiles.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
benkobalog: I removed the show method.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
benkobalog: I found the relevant  [conversation](https://github.com/lampepfl/dotty/pull/1787). I will write a better explanation based on this.

benkobalog: I updated the explanation.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
benkobalog: I changed the class name, ID and test name too.

benkobalog:
I will close this pull request and open a new one, because I couldn't properly resolve the conflicts (I'm new to this part of git)

PR #3747, opened 2018-01-04
Fix #3248: Handle repeated arguments in results of unapply selectors

### LINKED ISSUE #3248: Some[(A, Seq[B])] is not a valid result type of an unapplySeq
This code snippet fails to compile with Dotty
```scala
class Test {
  class Foo(val name: String, val children: Int *)
  object Foo {
    def unapplySeq(f: Foo) = Some((f.name, f.children))
  }

  def foo(f: Foo) = f match {
    case Foo(name, ns @ _*) => 1
  }
}
```

```scala
[info] Running dotty.tools.dotc.Main -d out -language:Scala2 tests/allan/Test.scala -classpath dotty-library_2.12-0.4.0-bin-SNAPSHOT-nonbootstrapped.jar
-- Error: tests/allan/Test.scala:8:12 ------------------------------------------
8 |    case Foo(name, ns: _*) => 1
  |         ^^^^^^^^^^^^^^^^^
  |Some[(String, scala.collection.Seq[Int] @Repeated)] is not a valid result type of an unapplySeq method of an extractor
```

odersky:
Normal unapply selectors can return results that have a repeated
argument as last element. This needs to be handled. This also subsumes
the previous way to express such arguments with `unapplySeq`.

In the long run, we should be able to do without `unapplySeq` altogether.
This is an intermediate step to get there.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -77,7 +77,10 @@ class Definitions {
         val cls = denot.asClass.classSymbol
         val paramDecls = newScope
         val typeParam = enterSyntheticTypeParam(cls, paramFlags, paramDecls)
-        val parents = parentConstrs.toList
+        def instantiate(tpe: Type) =
+          if (tpe.typeParams.nonEmpty) tpe.appliedTo(typeParam.typeRef)
+          else tpe
+        val parents = parentConstrs.toList map instantiate
         denot.info = ClassInfo(ScalaPackageClass.thisType, cls, parents, paramDecls)
       }
     }
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1363,6 +1363,15 @@ object Types {
      */
     def signature(implicit ctx: Context): Signature = Signature.NotAMethod
 
+    def annotatedToRepeated(implicit ctx: Context): Type = this match {
+      case tp @ ExprType(tp1) => tp.derivedExprType(tp1.annotatedToRepeated)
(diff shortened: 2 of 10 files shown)


odersky:
This currently fails the exhaustivity checks. @liufengyun Can you see whether this is fixable?

liufengyun:
@odersky : The bug is fixed. There are 3 problems: 

1. The parent `Seq` is incorrectly instantiated for `RepeatedParamClass`, causing exception in `typedSeqLiteral`, due to `baseType` failure in `elemType`.
2. `SeqLiteral` is not handled in pattern matcher.
3. `SeqLiteral` is not handled in exhaustivity check.

odersky:
@liufengyun Looks good. Thanks for the fix!

nicolasstucki:
This PR improved the performance of the compiler in http://dotty-bench.epfl.ch/.
Dotty has never been so fast before :)

odersky:
Wow! I looked at the changes and have no idea what could have caused this. Maybe Fengyun's changes to the pattern matcher?

liufengyun:
It's related to the machine restart, where Turbo Disable setting failed.

The machine powered off around Jan 10th for unknown causes. Today I restarted machine, ran the config to fixed the CPU frequency and disable Turbo. For some reason, the Turbo mode setting doesn't take effect.

I'll fix the setting and rerun the benchmarks for the last 2-3 merges.

OlivierBlanvillain:
I'm tagging this with a "needs spec" as the affects what's written [here](https://www.scala-lang.org/files/archive/spec/2.11/08-pattern-matching.html#pattern-sequences).

allanrenucci:
@OlivierBlanvillain I would open a new issue instead. It will probably get lost othwerwise

Issue #2977, opened 2017-08-15
REPL behaves strangely on first try, OK afterwards.


odersky:
Here's a transcript:
```scala
scala> trait A 
defined trait A
scala> trait B 
defined trait B
scala> class C extends A with B 
-- Error: <console>:6:23 -------------------------------------------------------
6 |class C extends A with B
  |                       ^
  |unexpected parent type: HKApply(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),A),List(TypeRef(ThisType(TypeRef(NoPrefix,<empty>)),B)))
scala> class C extends A with B 
defined class C
```

felixmulder:
Seems to be classpath issues from discussion in meeting.

DarkDimius:
I'm pretty sure this is due to classpath pollution.

PR #3751, opened 2018-01-04
Fix #2997: Add regression test

### LINKED ISSUE #2997: Special handling in the typer of scala.Singleton is missing
```scala
  case class Foo[T <: Int with Singleton](t : T)
  val one = 1
  final val final_one = 1
  val a : 1 = Foo(1).t //Type Mismatch Error
  val b : Int = Foo(one).t //OK, but unexpected
  val c : 1 = Foo(final_one).t //Type Mismatch Error
```
Scastie link: https://scastie.scala-lang.org/DcDeG4eYQTG17exjb2FigQ

allanrenucci:


compiler/test-resources/repl/i2997
```diff
@@ -0,0 +1,8 @@
+scala> trait A
```
smarter: Wrong testcase?

allanrenucci: Wrong issue. I meant #2977 😃. Will update the PR

PR #3753, opened 2018-01-04
Fix #2789

### LINKED ISSUE #2789: Anonymous functions in REPL (and expressions) must be wrapped in parentheses
Dotty's REPL does not wrap the expression to evaluate in parens (like Scala's REPL). But because of #1424, wrapping expressions in parentheses can be needed to preserve behavior.

That bug is still open but confusingly titled, and the actual issue is hidden. Hence the separate one.

TL;DR. Dotty's REPL rejects`(x: Any) => println(1)` but accepts `((x: Any) => println(1))`. Martin suggested a fix in https://github.com/lampepfl/dotty/issues/1424#issuecomment-236449912

> So it seems the REPL has to special case closures in any case.
> So closures have to be put in parentheses by the REPL anyway.

I'm not sure if closures need special casing, or you can just wrap every expression in parentheses. Of course, definitions can't be wrapped :-|.

```scala
scala>  x => println(1)
-- [E080] Syntax Error: <console>:5:1 ------------------------------------------
5 | x => println(1)
  | ^
  | missing parameter type for parameter x, expected = ?

scala> (x: Any) => println(1)
-- Error: <console>:1:0 --------------------------------------------------------
1 |(x: Any) => println(1)
  |^^^^^^^^
  |not a legal self type clause
```

allanrenucci:
Closing #1424 too?

### DIFF
--- a/compiler/test/dotty/tools/repl/ReplCompilerTests.scala
+++ b/compiler/test/dotty/tools/repl/ReplCompilerTests.scala
@@ -115,4 +115,9 @@ class ReplCompilerTests extends ReplTest {
     //   storedOutput().startsWith("scala.MatchError: null")
     // }
   }
+
+  @Test def i2789: Unit = fromInitialState { implicit state =>
+    compile("(x: Int) => println(x)")
+    assertTrue(storedOutput().startsWith("val res0: Int => Unit ="))
+  }
 }
--- a/compiler/test/dotty/tools/repl/ReplCompilerTests.scala
+++ b/compiler/test/dotty/tools/repl/ReplCompilerTests.scala
@@ -2,7 +2,7 @@ package dotty.tools
 package repl
 
 import org.junit.Assert._
-import org.junit.Test
+import org.junit.{Test, Ignore}
 
 import dotc.core.Contexts.Context
 import dotc.ast.Trees._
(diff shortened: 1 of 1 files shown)


compiler/test/dotty/tools/repl/ReplCompilerTests.scala
```diff
@@ -115,4 +115,9 @@ class ReplCompilerTests extends ReplTest {
     //   storedOutput().startsWith("scala.MatchError: null")
     // }
   }
+
+  @Test def i2789: Unit = fromInitialState { implicit state =>
+    compile("(x: Int) => println(x)")
```
smarter: Why not put this in a repl script test?

compiler/test/dotty/tools/repl/ReplCompilerTests.scala
```diff
@@ -93,27 +93,27 @@ class ReplCompilerTests extends ReplTest {
       )
     }
 
-  @Test def i3305: Unit = {
+  // FIXME: Tests are not run in isolation, the classloader is corrupted after the first exception
```
smarter: Open an issue for that?

smarter:
Do you know what PR fixed the underlying problem with anonymous function parsing?

compiler/test/dotty/tools/repl/ReplCompilerTests.scala
allanrenucci: Because it prints something like this:
```scala
val res0: Int => Unit = Lambda$1423/1112073835@3e6534e7
```
which is not deterministic

compiler/test/dotty/tools/repl/ReplCompilerTests.scala
allanrenucci: #3754

allanrenucci:
>Do you know what PR fixed the underlying problem with anonymous function parsing?

https://github.com/lampepfl/dotty/commit/a886cd481eecc26bb32c4af119d7cbc781c8dffc#diff-69c8fd9135f88af62a8b7073df7c69c7R102

PR #3755, opened 2018-01-04
Fix #3533: Fix parsing of raw types appearing in generic position

### LINKED ISSUE #3533: Interaction with Java's raw generic types is not possible
See for instance Javafx's method https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/TreeTableView.html#setColumnResizePolicy-javafx.util.Callback- it takes a `Callback<TreeTableView.ResizeFeatures,Boolean>` but do notice that `ResizeFeatures` defines one generic parameter, the method defines a callback that takes it raw though. In scalac, you would pass a `javafx.util.Callback[javafx.scene.control.TreeTableView.ResizeFeatures[_], Boolean]` and it would typecheck, but with dottyc it doesn't work

smarter:


compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -261,11 +261,11 @@ class ClassfileParser(
           addConstructorTypeParams(denot)
         }
 
-        denot.info = cook.apply(pool.getType(in.nextChar))
+        denot.info = pool.getType(in.nextChar)
         if (isEnum) denot.info = ConstantType(Constant(sym))
         if (isConstructor) normalizeConstructorParams()
         setPrivateWithin(denot, jflags)
-        denot.info = translateTempPoly(parseAttributes(sym, denot.info))
+        denot.info = translateTempPoly(cook.apply(parseAttributes(sym, denot.info)))
```
odersky: With the change, the two applications of `cook.apply` are in 1-to-1 correspondence with the two calls to `parseAttributes`. Add the `cook.apply` to the result type of `parseAttributes` instead? 

Otherwise LGTM

smarter: Makes sense, PR updated.

PR #3763, opened 2018-01-05
Make Ycheck stricter

allanrenucci:
I found out that in some cases Ycheck would print an error but not fail. Is it intended?

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -56,16 +56,12 @@ class TreeChecker extends Phase with SymTransformer {
 
   def isValidJVMMethodName(name: Name) = name.toString.forall(isValidJVMMethodChar)
 
-  def printError(str: String)(implicit ctx: Context) = {
-    ctx.echo(Console.RED + "[error] " + Console.WHITE  + str)
-  }
-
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
```diff
@@ -56,17 +56,18 @@ class TreeChecker extends Phase with SymTransformer {
...
 
   def testDuplicate(sym: Symbol, registry: mutable.Map[String, Symbol], typ: String)(implicit ctx: Context) = {
     val name = sym.fullName.mangledString
-    if (this.flatClasses && registry.contains(name))
-        printError(s"$typ defined twice $sym ${sym.id} ${registry(name).id}")
-    registry(name) = sym
+    if (this.flatClasses) {
+      registry.get(name) match {
+        case Some(prev) =>
+          assert(sym eq prev, i"$typ defined twice $sym ${sym.id} ${prev.id}")
```
allanrenucci: The same symbol goes through `def transformSym(symd: SymDenotation)(implicit ctx: Context): SymDenotation` multiple times. Is this expected?

smarter: Fix in https://github.com/lampepfl/dotty/pull/4224

PR #3768, opened 2018-01-05
Fix #1619: Reduce collision rate in NameTable

### LINKED ISSUE #1619: Port NameTable collision rate reducing change from Scalac
https://github.com/scala/scala/pull/5474

I originally though we're not affected, but I've misunderstood the issue. We are affected.

allanrenucci:
Rebased #1727

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Names.scala
+++ b/compiler/src/dotty/tools/dotc/core/Names.scala
@@ -546,13 +546,15 @@ object Names {
   private[this] var size = 1
 
   /** The hash of a name made of from characters cs[offset..offset+len-1].  */
-  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int =
-    if (len > 0)
-      (len * (41 * 41 * 41) +
-        cs(offset) * (41 * 41) +
-        cs(offset + len - 1) * 41 +
-        cs(offset + (len >> 1)))
-    else 0
+  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int = {
+    var i = offset
+    var hash = 0
+    while (i < len + offset) {
+      hash = 31 * hash + cs(i)
+      i += 1
+    }
+    hash
+  }
 
   /** Is (the ASCII representation of) name at given index equal to
    *  cs[offset..offset+len-1]?


allanrenucci:
test performance please

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -546,13 +546,15 @@ object Names {
...
-    if (len > 0)
-      (len * (41 * 41 * 41) +
-        cs(offset) * (41 * 41) +
-        cs(offset + len - 1) * 41 +
-        cs(offset + (len >> 1)))
-    else 0
+  private def hashValue(cs: Array[Char], offset: Int, len: Int): Int = {
+    var i = offset
+    var hash = 0
+    while (i < len + offset) {
```
nicolasstucki: Compute `len+offset` only once.

smarter: Pretty sure the jvm should figure it out :).

compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -549,7 +549,8 @@ object Names {
   private def hashValue(cs: Array[Char], offset: Int, len: Int): Int = {
     var i = offset
     var hash = 0
-    while (i < len + offset) {
+    val until = offset + len
```
odersky: I doubt this has any effect. Hoisting constant expressions out of loops is among the easiest optimizations of a JIT compiler. So I think it's better to revert this commit. 

Otherwise LGTM.

smarter:
Not sure if we should merge this. It seems to have negligible effects on our benchmarks, but the corresponding scalac PR (https://github.com/scala/scala/pull/5474) hasn't been merged because of a %2 slowdown on scalap. Unless someone can find a benchmark where we actually do better with this PR in, what's the point anyway?

odersky:
The situation between dotc and scalac is a bit different because with semantic names we have on average much shorter names in the NameTable. So the fix is likely less costly (and potentially also less useful) than for scalac. I think it would be good to benchmark `scalap` and if that does not show a slow-down we can go ahead and merge. 

I believe the benefit of this fix is not so much a speedup on average but avoiding outliers.

smarter:
@retronym How come https://github.com/scala/scala/pull/5474 changes the magic prime used in the hashcode from 41 to 31?

retronym:
I wanted to use the same values as `java.lang.String`. I think it makes sense to go ahead with this change, even if you're seeing flat benchmark results.

retronym:
For reference, we ended up merging https://github.com/scala/scala/pull/5506 instead, which avoids needlessly creating new names for fresh type symbols created during overload resolution.

allanrenucci:
test performance please

allanrenucci:
test performance please

allanrenucci:
test performance please

PR #3770, opened 2018-01-06
Fix #3067: Flag missing parent type of implicit object as error

### LINKED ISSUE #3067: Cyclic reference error involving CanBuildFrom in implicit objects
The following code compiles on Scala but produces a `cyclic reference involving trait CanBuildFrom` compiler error on Dotty:

```scala
class Test[T](f: List[String] => T)

object o {
  implicit object a extends Test(_ map identity)
  implicit object b extends Test(_ map identity)
}
```

odersky:
Implicit objects should not allow parent types to be computed from argument
expressions of an application. This is analogous to demanding an explicit type
for implicit vals and defs.

This fix was straightfoward, however it exhibited another problem: Some errors issued
during completion were never reported because the completion happened inside a 
`TyperState#test` that reset the reporter subsequently. The fix for this one was a bit
more involved.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Contexts.scala
+++ b/compiler/src/dotty/tools/dotc/core/Contexts.scala
@@ -282,6 +282,11 @@ object Contexts {
     /** The current reporter */
     def reporter: Reporter = typerState.reporter
 
+    /** Run `op` as if it was run in a fresh explore typer state, but possibly
+     *  optimized to re-use the current typer state.
+     */
+    final def test[T](op: Context => T): T = typerState.test(op)(this)
+
     /** Is this a context for the members of a class definition? */
     def isClassDefContext: Boolean =
       owner.isClass && (owner ne outer.owner)
--- a/compiler/src/dotty/tools/dotc/core/TyperState.scala
+++ b/compiler/src/dotty/tools/dotc/core/TyperState.scala
@@ -64,6 +64,14 @@ class TyperState(previous: TyperState /* | Null */) extends DotClass with Showab
   def isGlobalCommittable: Boolean =
     isCommittable && (previous == null || previous.isGlobalCommittable)
 
+  private[this] var isShared = false
+
+  /** Mark typer state as shared (typically because it is the typer state of
+   *  the creation context of a source definition that potentially still needs
+   *  to be completed). Members of shared typer states are never overwritten in `test`.
+   */
+  def markShared(): Unit = isShared = true
+
   private[this] var isCommitted = false
 
   /** A fresh typer state with the same constraint as this one. */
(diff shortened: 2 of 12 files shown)


odersky:
@nicolasstucki Can you check why i3067 fails the fromTasty test?

smarter:
Could we instead restrict the implicit scope used when typing the parents of an `implicit object` to not contain the object itself?

nicolasstucki:
I minimized the failure of i3067 from TASTY.

```scala
class Test(f: List[String] => List[String])
object o extends Test(_.map(x => x))
```

The bug is that when we compile from TASTY not all `def`s of closures are lifted in `lambdaLift`. In this case `x => x` is not lifted. 

```scala
 @scala.annotation.internal.SourceFile("tests/pos/i3067.scala") final module 
    class
   o$ extends Test { 
    def <init>(): Unit = 
      {
        super(o$#o$$superArg$1())
        ()
      }
    private <static> def o$$superArg$1(): Function1 = 
      {
        closure(this.o$$superArg$1$$anonfun$1)
      }
    private <static> def o$$superArg$1$$anonfun$1$$anonfun$1(x: String): String
       = 
    x
    private <static> def o$$superArg$1$$anonfun$1(
      _$1: scala.collection.immutable.List
    ): scala.collection.immutable.List = 
      _$1.map(
        {
          closure(this.o$$superArg$1$$anonfun$1$$anonfun$1)
        }
      , scala.collection.immutable.List.canBuildFrom()).asInstanceOf[
        scala.collection.immutable.List
      ]
  }
```


```scala
  @scala.annotation.internal.SourceFile("tests/pos/i3067.scala") final module 
    class
   o$ extends Test { 
    def <init>(): Unit = 
      {
        super(o$#o$$superArg$1())
        ()
      }
    private <static> def o$$superArg$1(): Function1 = 
      {
        closure(this.o$$superArg$1$$anonfun$1)
      }
    private <static> def o$$superArg$1$$anonfun$1(
      _$1: scala.collection.immutable.List
    ): scala.collection.immutable.List = 
      _$1.map(
        {
          def $anonfun(x: String): String = x
          closure(o.$anonfun)
        }
      , scala.collection.immutable.List.canBuildFrom()).asInstanceOf[
        scala.collection.immutable.List
      ]
  }
```

odersky:
> Could we instead restrict the implicit scope used when typing the parents of an implicit object to not contain the object itself?

And what to do about indirect recursions? Since we decided on demanding full types elsewhere I think it makes sense to not make an exception for objects.

odersky:
@nicolastucki What is a simple command to compile from Tasty that does not involve composing a classpath manually?

nicolasstucki:
Not really. I will create a way to test from TASTY compilation directly from the sources.

For now, you can replace `compileTastyInDir(....)` in `FromTastyTests` with `compileTastyInDir("../tests/pos/i3067.scala", defaultOptions)` to only test that file only.

nicolasstucki:
#3786 will add a way to debug from Tasty directly from the sources.

smarter:
The markShared scheme seems fragile. It's hard to know for sure all the situations where it could be necessary. I can think of at least FunProto (whose peculiar use of Context is already causing problems: https://github.com/lampepfl/dotty/issues/2412)

odersky:
`FunProto` is still a sticky point. For now, I know neither of a better solution in general nor have I a concrete amelioration of `FunProto`. I tried some variations but they all fail tests. We should leave at least one issue open that mentions that or create a new one.

PR #3772, opened 2018-01-07
Fix #3012: Avoid swallowing multiple ()'s

### LINKED ISSUE #3012: Non-nullary apply is permitted on dotty nullary method overriding java method
Copy-pasteable repro.
```scala
cat << EOF > Fuzbar.java
package fuz;
public interface Fuzbar {
  public String str();
}
EOF
cat << EOF > a.scala
object a extends fuz.Fuzbar {
  override def str = ""
  str()()()()()()
}
EOF
dotc a.scala Fuzbar.java
// compiles OK, expected error
$ dotc -version
0.4.0-bin-SNAPSHOT-nonbootstrapped-git-d70b8f4
```

I expected dotty to report an error like below

```
-- Error: a.scala:5:6 ----------------------------------------------------------
5 |  str()()()()
  |  ^^^^^
  |  missing argument for parameter index of method apply: (index: Int): Char

```

Related https://github.com/lampepfl/dotty/issues/2570

odersky:
Allow dropping at most one () if a method is called that exists
in both ()T and => T forms.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -64,6 +64,7 @@ object Typer {
 
   private val ExprOwner = new Property.Key[Symbol]
   private val InsertedApply = new Property.Key[Unit]
+  private val DroppedEmptyArgs = new Property.Key[Unit]
 }
 
 class Typer extends Namer with TypeAssigner with Applications with Implicits with Dynamic with Checking with Docstrings {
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -64,6 +64,7 @@ object Typer {
 
   private val ExprOwner = new Property.Key[Symbol]
   private val InsertedApply = new Property.Key[Unit]
+  private val DroppedEmptyArgs = new Property.Key[Unit]
```
smarter: I think every key should have a documentation comment :).

tests/neg/i3012/a.scala
```diff
@@ -0,0 +1,5 @@
+
+object a extends fuz.Fuzbar {
+  override def str = ""
+  str()()()()()()  // error: missing argument
```
smarter: Add a line with just `str()()` to check that it's also an error? A pos test with `str` and `str()` would be useful too.

PR #3773, opened 2018-01-07
Treat singleton ExprTypes as stable values

### LINKED ISSUE #3005: Add support for lazy (explicit|implicit) parameters
#1998 makes implicit parameters consistent with explicit parameters by allowing implicit parameters to also be by-name. This change was motivated in Dotty (and work in progress on Scala 2) to support the same sort of type level programming use cases which are currently enabled by shapeless's `Lazy` type.

However, the bynamity of a parameter prevents it from being stable which means that the following,
```scala
trait Foo {
  type Out
  def out: Out
}

object Test {
  implicit def bar(implicit foo: => Foo): foo.Out = foo.out
}
```
doesn't compile (either in Dotty or Scala 2 with by-name implicit support) because the by-name argument `foo` is not stable,
```
lazy-implicits-7.scala:12: error: stable identifier required, but foo found.
  implicit def bar(implicit foo: => Foo): foo.Out = foo.out
                                          ^
one error found
```
This pattern is very important in implicit-based type level programming, so this would be a disappointing limitation. By contrast, shapeless's `Lazy` captures the value in a stable context and hence supports the following with vanilla Scala 2,
```scala
implicit def bar(implicit foo: Lazy[Foo]): foo.value.Out = foo.value.out
```
Currently the following is a workaround with by-name implicits which compiles both on my Scala 2 branch and with Dotty,

```scala
trait Foo {
  type Out
  def out: Out
}

object Foo {
  type Aux[Out0] = Foo { type Out = Out0 }
}

object Test {
  implicit def bar[T](implicit foo: => Foo.Aux[T]): T = foo.value
}
```
ie. we use the Aux pattern to avoid the need for a stable value. This is clumsy however, and the additional type parameter (which typically must be inferred) can make it hard to express methods which also require explicit type arguments.

So, whilst by name implicits get us close be being able to replace shapeless's `Lazy` macro, it's not quite there.

It was proposed in the discussion on #1998 to use the `lazy` modifier on method parameters to express by-need (ie. at most once) evaluation semantics. If such parameters could participate in stable paths (at least modulo the restrictions described [here](http://scala-lang.org/blog/2016/02/17/scaling-dot-soundness.html#plugging-the-loopholes) and the issues enumerated in the [Scala 2 ticket](https://github.com/scala/bug/issues/240)) then a `lazy implicit` parameter would more or less exactly replicate the semantics of shapeless's `Lazy`.

odersky:
Two main changes:

 - values with type => T are stable if T is stable
 - types deriving from scala.Singleton are considered stable

The two changes together allow to declare by-name-parameters to be stable
by declaring them to have a Singleton type.

Currently such programs fail -Ycheck between elimByName and erasure because elimByName replaces e.g. `foo.out` for cbn foo with `foo.apply().out` but no corresponding operation can be performed on the types. So if we had `foo.out: foo.Out` before elimByName,  this is no longer true afterwards.

The most direct way to avoid the problem is probably to merge elimByName with erasure.

This fixes the test case of #3005, but is probably not enough to address the root of the problem. So I am not sure we should merge this.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -595,8 +595,11 @@ object SymDenotations {
       )
 
     /** Is this a denotation of a stable term (or an arbitrary type)? */
-    final def isStable(implicit ctx: Context) =
-      isType || is(Stable) || !(is(UnstableValue) || info.isInstanceOf[ExprType])
+    final def isStable(implicit ctx: Context) = {
+      def isUnstable =
+      	is(UnstableValue) || info.isInstanceOf[ExprType] && !info.isStable
+      isType || is(Stable) || !isUnstable
+    }
 
     /** Is this a "real" method? A real method is a method which is:
      *  - not an accessor
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -115,7 +115,7 @@ object Types {
       case tp: RefinedOrRecType => tp.parent.isStable
       case tp: ExprType => tp.resultType.isStable
       case tp: AnnotatedType => tp.tpe.isStable
-      case _ => false
+      case tp => tp.derivesFrom(defn.SingletonClass)
     }
 
     /** Is this type a (possibly refined or applied or aliased) type reference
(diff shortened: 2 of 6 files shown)


odersky:
This is actually not a fix for #3005 because it requires arguments to Singleton by name parameters to be stable. As is, it is also unsound, because it admits code like this:

    def x: T & Singleton = x

`x` is declared to be a Singleton yet is not realizable, so not a legal path. So the `Singleton` trick really works only for by name parameters, not for general parameterless `def`s. But then it pretty useless - if the only legal arguments for a singleton parameter are paths and constants, why make
the parameter by name in the first place?

PR #3774, opened 2018-01-07
Fix #2998: Constrain type from above before interpolating variables

### LINKED ISSUE #2997: Special handling in the typer of scala.Singleton is missing

### LINKED ISSUE #2998: When ExprType(T) has expected type Foo, we should add a constraint T <: Foo
Type inference should work the same for `zero` and `one`, but it doesn't, `foo(1).zero` has expected type `1` but this isn't added as a constraint, so we instantiate `T` to `Int` (because the only constraint is `>: 1` and we widen when instantiating a singleton lower bound)

```scala
class Foo[T] {
  def zero: T = ???
  def one(): T = ???
}

object Test {
  def foo[T](x: T): Foo[T] = new Foo

  val a: 1 = foo(1).zero // Fails: Found: Int, required: Int(1)
  val b: 1 = foo(1).one() // OK
}
```

odersky:
When interpolating type variables we should use all available information
at this point, including the expected result type. One situation where this makes
a difference is if the expected type is a singleton type, because type variables
are instantiated to singleton types only if their upper bounds are singleton types.

Also fixes #2997.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -345,7 +345,13 @@ trait Inferencing { this: Typer =>
       def ensureConstrained() =
         if (!isConstrained) {
           isConstrained = true
-          constrainResult(tree.tpe, pt)
+          tree match {
+            case tree: Apply => // already constrained
+            case _ => tree.tpe match {
+              case _: MethodOrPoly => // already constrained
+              case tp => constrainResult(tp, pt)
+            }
+          }
         }
 
       // Avoid interpolating variables if typerstate has unreported errors.
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -222,72 +222,6 @@ object Inferencing {
       case _ => NoType
     }
 
-  /** Interpolate those undetermined type variables in the widened type of this tree
-   *  which are introduced by type application contained in the tree.
-   *  If such a variable appears covariantly in type `tp` or does not appear at all,
-   *  approximate it by its lower bound. Otherwise, if it appears contravariantly
-   *  in type `tp` approximate it by its upper bound.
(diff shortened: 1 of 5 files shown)


odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
The last commit seemed to have helped performance, so we are good on that count, I think.

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -375,6 +309,93 @@ object Inferencing {
...
+      constr.println(s"interpolate undet vars in ${tp.show}, pos = ${tree.pos}, mode = ${ctx.mode}, undets = ${constraint.uninstVars map (tvar => s"${tvar.show}@${tvar.bindingTree.pos}")}")
+      constr.println(s"qualifying undet vars: ${constraint.uninstVars filter qualifies map (tvar => s"$tvar / ${tvar.show}")}, constraint: ${constraint.show}")
+
+      val vs = variances(tp, qualifies)
+      val hasUnreportedErrors = ctx.typerState.reporter match {
+        case r: StoreReporter if r.hasErrors => true
+        case _ => false
+      }
+
+      var isConstrained = false
```
smarter: I think it would be cleaner to do:
```scala
var isConstrained = tree.isInstanceOf[Apply] || tree.tpe.isInstanceOf[MethodOrPoly]
```

odersky: I saw that comment to late, after merging. I'll attach the change to a subsequent PR.

PR #3410, opened 2017-10-31
Remove scala.Phantom (replaced by `unused`)

nicolasstucki:

PR #3779, opened 2018-01-09
Added error message for non object phantom extensions

Wojtechnology:
Fixes `Typer.scala:1519` in https://github.com/lampepfl/dotty/issues/1589.


nicolasstucki:
Unfortunately some major changes to the phantoms are coming in #3410. Making error messages for phantoms is useless until we merge that PR as they might completely change.

PR #3780, opened 2018-01-09
Attempt at fixing `Applications.scala:95` #1589

aesteve:
First try at implementing an error message, please tell me if I'm doing a mistake in some way.

Especially : 

* is `val kind = "Reference"` the right kind of error
* I don't really know how to test the `explain` part 
* is the explain message relevant, correct, and not too big
* should I also be testing `unapplySeq` ?

Thanks a lot for the review.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2074,4 +2074,21 @@ object messages {
           |$rsym does not override any method in $parentSym. Structural refinement does not allow for
           |polymorphic methods."""
   }
+
+  case class NotAValidResultTypeOfUnapply(sym: Symbol, resultType: Type, args: List[untpd.Tree])(implicit ctx: Context) extends Message(NotAValidResultTypeOfUnapplyID) {
+    val kind = "Reference"
+    val resultClassName = resultType.classSymbol.name
+    val msg = hl"Cannot extract (${sym.name}) from ${sym.owner}. $resultClassName is not a valid extractor return type"
+    val explanation =
+      hl"""${sym.name} method of ${sym.owner} has been tried as a candidate to extract ${args.map(_.show).mkString("[", ", ", "]")}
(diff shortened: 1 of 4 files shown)


aesteve:
I signed the CLA. Should I open a new PR ?

liufengyun:
@aesteve You can update the PR (e.g. by force push), and it will re-trigger the CI.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2074,4 +2074,21 @@ object messages {
...
+    val resultClassName = resultType.classSymbol.name
+    val msg = hl"Cannot extract (${sym.name}) from ${sym.owner}. $resultClassName is not a valid extractor return type"
+    val explanation =
+      hl"""${sym.name} method of ${sym.owner} has been tried as a candidate to extract ${args.map(_.show).mkString("[", ", ", "]")}
+          |Its return type: ${s"$resultClassName"} is not a valid return type for an extractor.
+          |
+          |The return type of an extractor should be chosen as follows:
+          | - If it is just a test, return a ${"Boolean"}. For instance ${"case even()"}
+          | - If it returns a single sub-value of type ${s"$resultClassName"}, return an ${s"Option[$resultClassName]"}
+          | - If you want to return several sub-values ${"T1,...,Tn"}, group them in an optional tuple ${"Option[(T1,...,Tn)]"}.
```
OlivierBlanvillain: This is not telling the full story, we also need to mention [option-less pattern matching
](http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html) and talk about unapplySeq.

allanrenucci: I believe `unapplySeq` has been deprecated in #3747

OlivierBlanvillain:
By the way, this is a duplicate of #3501.

odersky:
I believe a lot more work needs to be done on good unapply error messages. That work has to be in the Typer checker, which has to become better at finding root causes of unapply errors. Adding a fixed error message for the failure would make things more complicated at the present time

PR #3785, opened 2018-01-10
Fix #3784 : add an enum for Error kind

### LINKED ISSUE #3784: Use enumerations for Error kinds
For a newcomer implementing Error Messages, choosing the right kind of error is a bit difficult. What is "kind" ? What does it mean ? Is it documented somewhere ? Which one should I pick ?

A very simple first step could be to gather all the existing kinds into an enumeration, this is simple, and allows to gather every kind of error in a single place.

This would allow newcomers to, at least, quickly get a glimpse of the different kind of errors already used (`Syntax`, `Reference`, etc.). 

After that, maybe documenting the enum would be a very good idea.

aesteve:
First step to give newcomers a better view of the different kind of errors the compiler may throw.

Having a quick glimpse at existing error kinds should help picking the right one (or adding a new one).


compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorKind.java
```diff
@@ -0,0 +1,28 @@
+package dotty.tools.dotc.reporting.diagnostic;
+
+public enum ErrorKind {
```
nicolasstucki: It would be better to have this in Scala.

```scala
class ErrorKind(name: String) {
  def toString(): String = name
}
object ErrorKinds {
  val COMPATIBILITY = new ErrorKind("Compatibility")
  val DEFINITION_NOT_FOUND = new ErrorKind("Definition Not Found")
  ...
}
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/Message.scala
```diff
@@ -117,7 +117,7 @@ class ExtendMessage(_msg: () => Message)(f: String => String) { self =>
 /** The fallback `Message` containing no explanation and having no `kind` */
 class NoExplanation(val msg: String) extends Message(ErrorMessageID.NoExplanationID) {
   val explanation = ""
-  val kind = ""
+  val kind = ErrorKind.NO_KIND
```
nicolasstucki: With the change I proposed, this should become `NO_KIND`

nicolasstucki:
Otherwise it looks good

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorKind.java
allanrenucci: We are already using java enums for ErrorMessageID. I think it is fine to stick to java enum until we migrate everything to Dotty enum.

I would however use Scala naming convention: `NoKind` instead of `NO_KIND`

nicolasstucki: Ok.

But move `NoKind` to the first position.

aesteve:
Ok thanks, I wasn't sure about the Java enum vs. Scala object vs. the new enum system coming. I read the full discussion on #1970 and I guess it ended up confusing me ^^

I changed it.

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorCategory.scala
```diff
@@ -0,0 +1,20 @@
+package dotty.tools.dotc.reporting.diagnostic
+
+sealed case class ErrorCategory(name: String) {
```
allanrenucci: If you intend to make it behave like an enum, I suggest you use this pattern:
```scala
sealed abstract class ErrorCategory(category: String) {
   ...
}
object ErrorCategory {
  val NoKind = new ErrorCategory("") {}
  ...
}
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/ErrorCategory.scala
```diff
@@ -0,0 +1,20 @@
...
+  override def toString: String = name
+}
+object ErrorCategories {
+  val COMPATIBILITY = ErrorCategory("Compatibility")
```
allanrenucci: Please use Scala naming convention: `NoKind` instead of `NO_KIND`. Also I would put `NoKind` at the top.

aesteve: I can definitely rename everything, but it unfortunately clashes with the `case class PatternMatchExhaustivity(uncovered: String)` (both the message and kind clash together). What would you recommend ? Keep the same name but prefix like ` val kind = ErrorKind.PatternMatchExhaustivity` or rename the enum members ? (if so, how ?). Thanks for your insights

allanrenucci: Yes you can disambiguate with `val kind = ErrorCategory.PatternMatchExhaustivity`

aesteve:
That should be OK now.

aesteve:
@allanrenucci is it OK for you now ?

The longer this PR is opened, the more conflicts with people implementing new error messages we may get :)

aesteve:
Sorry I must have messed up with the merge... 

I have a lot of compiler errors now.

allanrenucci:
You can reopen a new PR if you feel like it is easier

odersky:
Let's revisit using Dotty enums once we have the bootstrap (hopefully real soon now!)

aesteve:
Good, and thanks.
Please ping me once the bootstrap thing ( ? :\ ) is merged.
I'll then take my PR back (sorry for abandoning it), review/rewrite it then submit a new one. 
Thanks for the work on dotty.

OlivierBlanvillain:
> Please ping me once the bootstrap thing ( ? :\ ) is merged.

ping @aesteve

aesteve:
I'm giving it a try with a proper Scala enum [here](https://github.com/lampepfl/dotty/pull/6759) but tests are failing with `failed: java.lang.NoSuchFieldError: NoKind`, still need to figure out why :)

b-studios:
@aesteve Is this PR still active, or do you want to close it?

aesteve:
Hello and sorry.
Let's just close it. I still have this contribution in mind, but it would be better to restart from scratch.
Fortunately closing it doesn't mean it'll disappear I'll still have the diff as inspiration if I have time to rework it.

Sorry once again, never found the actual time to do this properly.

b-studios:
Don't worry -- thanks for your quick feedback.

PR #3786, opened 2018-01-10
Add -Yfrom-tasty to debug from TASTY compilation direct from sources

nicolasstucki:


compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,60 @@
...
+    // see <https://groups.google.com/forum/#!topic/scala-user/kte6nak-zPM>.
+    val _ = NonFatal
+
+
+    println("From tasty debug driver")
+    assert(!args.contains("-d"))
+
+    val fromSourcesOut = Files.createTempDirectory(Paths.get("out").toAbsolutePath, "from-sources-tmp")
+
+    println(s"Compiling scala to sources to $fromSourcesOut")
```
allanrenucci: No sure it useful to print the output directory since the file will be deleted when the program exit

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,60 @@
...
+    // Preload scala.util.control.NonFatal. Otherwise, when trying to catch a StackOverflowError,
+    // we may try to load it but fail with another StackOverflowError and lose the original exception,
+    // see <https://groups.google.com/forum/#!topic/scala-user/kte6nak-zPM>.
+    val _ = NonFatal
+
+
+    println("From tasty debug driver")
+    assert(!args.contains("-d"))
+
+    val fromSourcesOut = Files.createTempDirectory(Paths.get("out").toAbsolutePath, "from-sources-tmp")
```
allanrenucci: Make sure `out` exist?

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    if (compilation1.hasErrors) {
+      println("Failed compilation from sources")
+      sys.exit(1)
+    }
+
+    val fromTastyOut = Files.createTempDirectory(Paths.get("out").toAbsolutePath, "from-tasty-tmp")
+
+    val ext = "hasTasty"
+    val classes = Path(fromSourcesOut).walk.filter(x => x.isFile && x.extension == ext).map { x =>
```
allanrenucci: `Directory`

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    println(s"Compiling scala to sources to $fromSourcesOut")
+    val compilation1 = dotc.Main.process("-d" +: fromSourcesOut.toString +: args)
+
+    if (compilation1.hasErrors) {
+      println("Failed compilation from sources")
+      sys.exit(1)
+    }
+
+    val fromTastyOut = Files.createTempDirectory(Paths.get("out").toAbsolutePath, "from-tasty-tmp")
```
allanrenucci: Why a temporary directory?

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    println(s"Compiling TASTY to sources from $fromSourcesOut to $fromTastyOut")
+    val compilation2 = dotc.Main.process(fromTastyArgs.toArray)
+
+    if (compilation2.hasErrors) {
+      println("Failed compilation from sources")
+      sys.exit(1)
+    }
+
+    Path(fromSourcesOut).deleteRecursively()
```
allanrenucci: Is this needed if the directory is temporary?

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
nicolasstucki: It works only on `sbt dotc`, `out` should exist.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
nicolasstucki: We do not delete them in case of failure. I will print the directories only in case of failure.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
nicolasstucki: Those temporary directories do not delete themselves. The `deleteOnExit` is not set by default on them.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
nicolasstucki: To ensure it is a clean directory and the only contents are the ones coming from this compilation.

nicolasstucki:
Any good ideas for the name of the flag?

smarter:
-Ythrough-tasty ? :)

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,67 @@
...
+      println("Failed compilation from sources")
+      sys.exit(1)
+    }
+
+    val fromTastyOut = Files.createDirectory(tmpOut.resolve("from-tasty"))
+
+    val ext = "hasTasty"
+    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && x.extension == ext).map { x =>
+      val source = x.toString
+      source.substring(fromSourcesOut.toString.length + 1, source.length - ext.length - 1).replace('/', '.')
```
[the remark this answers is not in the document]
nicolasstucki: ok

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,67 @@
...
+    assert(!args.contains("-d"))
+
+    val outPath = Paths.get("out")
+    Directory(outPath).createDirectory()
+
+    val tmpOut = Files.createTempDirectory(outPath.toAbsolutePath, "from-tasty-tmp")
+
+    val fromSourcesOut = Files.createDirectory(tmpOut.resolve("from-source"))
+
+    println(s"Compiling .scala")
```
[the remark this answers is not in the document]
nicolasstucki: I will change it to `Compiling from .scala sources` as in the second step the sources are the `.class` files.
I will change the second one to `Compiling TASTY from .class sources` to also be explicit.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    val tmpOut = Files.createTempDirectory(Paths.get("out").toAbsolutePath, "from-tasty-tmp")
+
+    val fromSourcesOut = Files.createDirectory(tmpOut.resolve("from-source"))
+
+    println(s"Compiling .scala")
+    val compilation1 = dotc.Main.process("-d" +: fromSourcesOut.toString +: args)
+
+    if (compilation1.hasErrors) {
+      println("Failed compilation from sources")
```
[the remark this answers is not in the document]
nicolasstucki: It would not tell me which of the two compilation failed.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -0,0 +1,64 @@
...
+import java.nio.file.{Files, Paths}
+
+object Debug {
+  def main(args: Array[String]): Unit = {
+    // Preload scala.util.control.NonFatal. Otherwise, when trying to catch a StackOverflowError,
+    // we may try to load it but fail with another StackOverflowError and lose the original exception,
+    // see <https://groups.google.com/forum/#!topic/scala-user/kte6nak-zPM>.
+    val _ = NonFatal
+
+    assert(!args.contains("-d"))
```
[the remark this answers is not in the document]
nicolasstucki: If someone needs that feature in the future, he can implement it.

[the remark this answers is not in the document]
nicolasstucki: That would actually be unsafe. For the test to be reproducible the output directory needs to be empty.

PR #3792, opened 2018-01-10
Fix #1380: Add missing part of the message

### LINKED ISSUE #1380: Custom message when missing type for expanded a function.
``` scala
scala> { case x: Int => 1 }
<console>:11: error: missing parameter type for expanded function
The argument types of an anonymous function must be fully known. (SLS 8.5)
Expected type was: ?
       { case x: Int => 1 }
       ^
```

``` scala
dotty> { case x: Int => 1 } 
  ^
{ case x: Int => 1 }
<console>:5: error: missing parameter type for parameter x$1 of expanded function x$1 => 
  x$1 match {
    case x: Int => 
      1
  }, expected = ?
```

nicolasstucki:


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -193,7 +193,9 @@ object messages {
         else
           ""
 
-      i"missing parameter type for parameter ${param.name}$ofFun, expected = $pt"
+      i"""missing parameter type for parameter ${param.name}$ofFun, expected = $pt
+         |The argument types of an anonymous function must be fully known. (SLS 8.5)
+         |"""
```
allanrenucci: This create a spurious new line at the end

PR #3800, opened 2018-01-11
Move some pending tests to neg/

smarter:


nicolasstucki:
`tests/neg/t7868.scala` failed with `-optimise`, you can move it to `tests/neg-no-optimise/t7868.scala`

PR #3802, opened 2018-01-11
Fix #2631: Add custom error message

### LINKED ISSUE #2631: Improve error message for field accesses in secondary constructors
```scala
class Foo(x : Any) {
  val foo : Integer = 0

  def this() = {
    this(foo)
  }
}
```

```shell
-- [E006] Unbound Identifier Error: /Users/bibou/Projects/EPFL/dotty/compiler/../tests/run/tMain.scala:5:9 -------------
5 |    this(foo)
  |         ^^^
  |         not found: foo
    Compilation failed for: '../tests/run/tMain.scala'    
```

nicolasstucki:


nicolasstucki:
@biboudis do not forget this one

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -389,6 +389,8 @@ class Typer extends Namer with TypeAssigner with Applications with Implicits wit
       }
       else if (name.toTermName == nme.ERROR)
         UnspecifiedErrorType
+      else if (ctx.owner.isConstructor && ctx.owner.owner.unforcedDecls.toList.exists(_.name == tree.name))
```
allanrenucci: I would factor out the second test into a def. It is not clear what it does. Maybe:
```scala
def isNotAccesibleFromConstructor(name: Name) = {
  val classMembers = ctx.owner.owner.unforcedDecls
  classMembers.lookup.(name).exists
}
```

nicolasstucki: If the test is factored out it is not sufficient to ensure that it is not accessible from the constructor.

nicolasstucki: I will use the `.lookup.(name).exists` though.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -396,7 +396,11 @@ class Typer extends Namer
       }
       else if (name.toTermName == nme.ERROR)
         UnspecifiedErrorType
-      else
+      else if (ctx.owner.isConstructor && ctx.owner.owner.unforcedDecls.lookup(tree.name).exists) {
+        // If the field existed but was not found we assume that
+        // we where in the context of an argument of the super constructor
+        errorType(ex"$tree is not accessible from super constructor arguments", tree.pos)
```
nicolasstucki: @smarter do you think this condition is enough?

smarter: Can't you check `ctx.mode.is(InSuperCall)` ? (The name is misleading, it's set for both this- and super-call, see the method superOrThisCallContext)

smarter: Also, don't say "super constructor", a this call is not a super call, it's just a constructor call

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -396,7 +396,12 @@ class Typer extends Namer
       }
       else if (name.toTermName == nme.ERROR)
         UnspecifiedErrorType
-      else
+      else if (ctx.owner.isConstructor && ctx.mode.is(Mode.InSuperCall) &&
```
allanrenucci: Doesn't `ctx.mode.is(Mode.InSuperCall)` implies `ctx.owner.isConstructor`?

nicolasstucki: It might be a method super call with `super.xys`.

PR #3804, opened 2018-01-11
Fix #2495: Improve error message

### LINKED ISSUE #2495: Monomorphism check missing on anonymous fns
Consider this snippet:

```scala
object TypeParam {

  sealed trait F[A]
  case object FI extends F[Int]
  case object FS extends F[String]

  trait FFn { def apply[A](f: F[A]): A }

  val test4: FFn = {
    case FI => 1
    case FS => ""
  }
}
```

In scalac it fails saying:
```
  // [error] missing parameter type for expanded function
  // [error] The argument types of an anonymous function must be fully known. (SLS 8.5)
  // [error] Expected type was: TypeParam.FFn
  // [error]   val test4: FFn = {
  // [error]                    ^
```

In dotc it fails saying:
```
-- Error: <console>:9:6 --------------------------------------------------------
9 |      case FI => 1
  |      ^
  |     missing parameter type for parameter x$1 of expanded function x$1 => 
  |       x$1 match 
  |         {
  |           case FI => 
  |             1
  |           case FS => 
  |             ""
  |         }, expected = FFn
```

which is much more confusing a message. It seems a monomorphism check is present in Scala and provides a specific error message where as in Dotty it's missing then fails further down the line.

nicolasstucki:


nicolasstucki:
The idea of this test is to show that the pretty printed version looks as expected. This is the best that can be done for this issue. The alternative is not testing it.

allanrenucci:
Yeah, I don't think this deserves a test.

My opinion for error messages is to test that the right error message is emitted and that it contains the right semantic information (if any). But not test how it is pretty printed.

I'll leave the last word to @smarter 😄

nicolasstucki:
I will remove it, it will be partially tested by the old scripted test.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -189,12 +189,15 @@ object messages {
...
-          i" of expanded function $tree"
+          i" of expanded function:\n$tree"
         else
           ""
 
-      i"""missing parameter type for parameter ${param.name}$ofFun, expected = $pt
-         |The argument types of an anonymous function must be fully known. (SLS 8.5)"""
+      i"""missing parameter type
+         |
+         |The argument types of an anonymous function must be fully known. (SLS 8.5)
```
liufengyun: It's better to have this general reference at the end, and put more specific information before.

nicolasstucki: This is part of the `missing parameter type` explaination.

nicolasstucki: It also follows the old format in scalac that users said was clearer.

PR #3806, opened 2018-01-11
Remove almost all entries from the stdlib blacklist

smarter:


smarter:
I think it could be useful to keep it around in case something changes in the future and we need to add back an entry.

Issue #1774, opened 2016-12-05
Spec: Are blocks allowed in patterns?


liufengyun:
Scalac doesn't parse following programs syntactically, but support it if it's synthesized.

Dotty accepts the code syntactically, but the typer generates a typing error.

```Scala
object Test {
  val a = (3, 4) match {
    case ({name: Int}, {parent: Int}) => (name, parent)
  }

  val name = a._1
  val parent = a._2
}
```

Error message:

```
-- Error: examples/block.scala -------------------------------------------------
6 |  val name = a._1
  |             ^
  |             cyclic reference involving value a
-- Error: examples/block.scala -------------------------------------------------
7 |  val parent = a._2
  |               ^
  |               cyclic reference involving value a
-- Error: examples/block.scala -------------------------------------------------
3 |    case ({name: Int}, {parent: Int}) => (name, parent)
  |                                          ^^^^
  |                                        cyclic reference involving value a

three errors found
```

julienrf:
Support could be useful for this kind of stuff:

~~~ scala
(3, 4) match {
  case { val single = MyTupleExtractor(Snd); single(4) } => …
}
~~~

(It makes it possible to manipulate extractors in rhs position, for instance to configure them)

julienrf:
But I’m not sure it is worth it… :)

liufengyun:
The implementation of Quasiquote in scala.meta ([link](https://github.com/scalameta/scalameta/blob/master/scalameta/quasiquotes/src/main/scala/scala/meta/internal/quasiquotes/ReificationMacros.scala#L411)) relies on anonymous extractors as follows:

```Scala
          new {
            def unapply(input: Joy) = {
              ..$preamble
              input match {
                case $lifted => $thenp
                case _       => $elsep
              }
            }
          }.unapply(..$args)
```

which gets expanded to a block:

```Scala
      {
          class $anon {
            def unapply(input: Joy) = {
              ..$preamble
              input match {
                case $lifted => $thenp
                case _       => $elsep
              }
            }
          }

         new $anon
      }.unapply(..$args)
```

odersky:
I think it would be good to allow this! Needs a spec change, though.

liufengyun:
As a side note, we don't need this feature in order to implement Quasiquote in dotty. The quasi quote in Dotty now pass all regression set in scalameta.

liufengyun:
Close this for now, I don't think it's a good idea to have blocks in patterns, the code is not readable.

[...]

PR #3810, opened 2018-01-11
Fix #1846: add regression test

### LINKED ISSUE #1846: Expanding a pattern match leads to "head of empty list"
When trying to compile this syntactically correct, but nonsensical program the compiler crashes:

```scala
object Bug {
  val x = 42
  x match {case {42}.toString => 42}
}
```
Beginning of the trace:

```
exception while transforming Bug.x match
  {
    case 42.toString() => 42
  } of class class dotty.tools.dotc.ast.Trees$Match # 79
exception while transforming () extends Object() {
  val x: Int = 42
  Bug.x match
    {
      case 42.toString() => 42
    }
} of class class dotty.tools.dotc.ast.Trees$Template # 81
exception while transforming @scala.annotation.internal.SourceFile("examples/bug.scala") final module class
  Bug$
() extends Object() {
  val x: Int = 42
  Bug.x match
    {
      case 42.toString() => 42
    }
} of class class dotty.tools.dotc.ast.Trees$TypeDef # 82
  final lazy module val Bug: Bug$ = new Bug$()
  @scala.annotation.internal.SourceFile("examples/bug.scala") final module class

  Bug$() extends Object() {
    val x: Int = 42
    Bug.x match
      {
        case 42.toString() => 42
      }
  }
} of class class dotty.tools.dotc.ast.Trees$PackageDef # 83

exception occurred while compiling examples/bug.scala
Exception in thread "main" java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:420)
	at scala.collection.immutable.Nil$.head(List.scala:417)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$ScalacPatternExpanders$ScalacPatternExpander$class.applyMethodTypes(PatternMatcher.scala:1863)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$ScalacPatternExpanders$alignPatterns$.applyMethodTypes(PatternMatcher.scala:1904)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$ScalacPatternExpanders$alignPatterns$.apply(PatternMatcher.scala:1955)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$ScalacPatternExpanders$alignPatterns$.apply(PatternMatcher.scala:1978)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$ExtractorCall$.apply(PatternMatcher.scala:1412)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.extractor$lzycompute(PatternMatcher.scala:1127)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.extractor(PatternMatcher.scala:1127)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.paramType$1(PatternMatcher.scala:1185)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.extractorStep(PatternMatcher.scala:1201)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.nextStep(PatternMatcher.scala:1222)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$BoundTree.translate(PatternMatcher.scala:1231)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$class.translatePattern(PatternMatcher.scala:1338)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$OptimizingMatchTranslator.translatePattern(PatternMatcher.scala:72)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$class.translateCase(PatternMatcher.scala:1335)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$OptimizingMatchTranslator.translateCase(PatternMatcher.scala:72)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$$anonfun$29.apply(PatternMatcher.scala:1299)
	at dotty.tools.dotc.transform.PatternMatcher$Translator$MatchTranslator$$anonfun$29.apply(PatternMatcher.scala:1299)
	at scala.collection.immutable.List.map(List.scala:273)
```

liufengyun:
Fix #1846: add regression test

This PR does the following:

- removes the block syntax in patterns 
- fix a crash under `-Ycheck`
- check that stable identifier patterns are really stable -- scalac compatible

Note that by removing the block syntax from patterns, following code is still supported:

```Scala
implicit class Regex(sc: StringContext) {
  def r = new util.matching.Regex(sc.parts.mkString, sc.parts.tail.map(_ => "x"): _*)
}

scala> "123" match { case r"\d+" => true case _ => false }
res34: Boolean = true

scala> "123" match { case r"(\d+)$d" => d.toInt case _ => 0 }
res36: Int = 123
```


allanrenucci:
Wow! I didn't know we could do that. We should probably spec it and test it thoroughly.

For instance, there are some inconsistencies:
```scala
42 match { case { 42 } => 42 } // OK
42 match { case { 42 }.toString => 42 } // OK, should error?
42 match { case { 42.toString } => 42 } // error
```

allanrenucci:
@liufengyun I see you closed #1774 saying it is a bad idea. Then I think we should make this a neg test and make sure the compiler rejects block in patterns. WDYT?

liufengyun:
@allanrenucci It's actually consistent, I've updated the test.

The following doesn't type-check because the pattern type and scrutinee type are irrelevant:

```Scala
42 match { case { 42.toString } => 42 }
```

The following code works:

```Scala
 "h" match { case { 42.toString } => 42 }
```

liufengyun:
@allanrenucci Note that blocks are allowed in the syntax specification for Dotty:

```Scala
SimplePattern     ::=  PatVar                                                   Ident(wildcard)
                    |  Literal                                                  Bind(name, Ident(wildcard))
                    |  ‘(’ [Patterns] ‘)’                                       Parens(pats) Tuple(pats)
                    |  XmlPattern
                    |  SimplePattern1 [TypeArgs] [ArgumentPatterns]
SimplePattern1    ::=  Path
                    |  ‘{’ Block ‘}’
                    |  SimplePattern1 ‘.’ id
```

I think in #1774, I misunderstood the semantics of blocks in patterns.

allanrenucci:
I understand why this typechecks:
```scala
42 match { case { 42.toString } => 42 }
```
But then I don't understand why this typechecks:
```scala
42 match { case { 42 }.toString => 42 }
```

liufengyun:
The following code shows the pattern:

```Scala
    val x = 42
    val Y = "h"

    x match { case { "h" }.toString => println(42) }  // ok
    x match { case { "h".toString } => println(42) }  // error
    x match { case Y => println(42) }  // error
    x match { case Y.toString => println(42) }  // ok
```

But arguably, all of them should be invalid.

allanrenucci:
> But arguably, all of them should be invalid.

Can you open an issue?

tests/run/i1846.scala
```diff
@@ -0,0 +1,22 @@
+object Test {
```
allanrenucci: I think is it better to use `assert` instead of a check file for simple test cases like this one. Also one more think you could test:
```scala
val foo = 42 match { case x @ { 42 } => x }
assert(foo == 42)
```

tests/run/i3812.scala
```diff
@@ -0,0 +1,15 @@
...
+    val Y = "42"
+
+    x match { case { 42 }           => () } // ok
+    x match { case { "42".toInt }   => () } // ok
+    x match { case { "42" }.toInt   => () } // ok
+    x match { case { "42".toInt }   => () } // ok
+    x match { case { Y.toInt }      => () } // ok
+    x match { case { Y }.toInt      => () } // ok
+
+    // x match { case Y.toInt          => () } // compiles but rejected by Ycheck
```
allanrenucci: @liufengyun `Ycheck` fails on this one. Should this be allowed?

liufengyun: Good catch, I think it should be allowed.

allanrenucci: Note that `case "42".toInt =>` does not parse. It might be confusing that one is allowed but not the other

liufengyun: The syntax specification requires the prefix has to be a` path` or `block`.

```
SimplePattern     ::=  PatVar                                                   Ident(wildcard)
                    |  Literal                                                  Bind(name, Ident(wildcard))
                    |  ‘(’ [Patterns] ‘)’                                       Parens(pats) Tuple(pats)
                    |  XmlPattern
                    |  SimplePattern1 [TypeArgs] [ArgumentPatterns]
SimplePattern1    ::=  Path
                    |  ‘{’ Block ‘}’
                    |  SimplePattern1 ‘.’ id
```

tests/run/i3812.scala
```diff
@@ -0,0 +1,8 @@
+object Test {
+  def main(args: Array[String]): Unit = {
+    val x = 42
+    val Y = "42"
+
+    x match { case Y.toInt          => () } // ok
```
allanrenucci: This is an error in scalac, I don't think we should support it either

liufengyun: Good point, a check for stable prefix is missing for patterns here.

liufengyun: Now I'm a little unsure about Scalac behavior, as it also rejects the following code:

```
    var X = 42
    42 match { case X          => () }
```

It seems to be it can be accepted without any issue.

liufengyun: According to scala spec 2.13, the scalac behavior is correct.

https://www.scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#stable-identifier-patterns

liufengyun:
@odersky  This PR does the following:

- removes the block syntax in patterns 
- fix a crash under `-Ycheck`
- check that stable identifier patterns are really stable -- scalac compatible

Note that by removing the block syntax from patterns, following code is still supported:

```Scala
implicit class Regex(sc: StringContext) {
  def r = new util.matching.Regex(sc.parts.mkString, sc.parts.tail.map(_ => "x"): _*)
}

scala> "123" match { case r"\d+" => true case _ => false }
res34: Boolean = true

scala> "123" match { case r"(\d+)$d" => d.toInt case _ => 0 }
res36: Int = 123
```

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -75,6 +75,21 @@ object RefChecks {
     }
   }
 
+  /** Check that a stable identifier pattern is indeed stable (SLS 8.1.5)
+   */
+  private def checkStableIdentPattern(tree: Tree)(implicit ctx: Context) = {
```
odersky: I believe it would be simpler to put this check in Typer. 

Here, we could just add `if (ctx.mode.is(Mode.Pattern) && !tree.tpe.isStable) error` to the checks for Ident and Select.

odersky: But I am also no sure why we should demand stability? What would go wrong if we didn't?

liufengyun: The following check seems too strong to me:

```
if (ctx.mode.is(Mode.Pattern) && !tree.tpe.isStable) error
```

As a selection of `unapply` would fail this check. The check should only happen for _Stable Identifier Pattern_ (SLS 8.1.5). I don't know what's the motivation for the Scala specification. Arguably, a `var` could be allowed in a pattern, but then it complicates the specification and check.

odersky: OK, so let's keep the test for stable idents but move it to Typer.

liufengyun: OK, now it's done.

PR #3811, opened 2018-01-11
Fix #2333: Check if class is deprecated

### LINKED ISSUE #2333: @deprecated on a class is ignored
```scala
@deprecated("bla", "2.11.0") class Foo

object Test {
  new Foo
}
```
```scala
% scalac -deprecation try/depr.scala
try/depr.scala:4: warning: class Foo is deprecated (since 2.11.0): foo
  new Foo
      ^
one warning found
% dotc -deprecation try/depr.scala
%
```

nicolasstucki:


allanrenucci:
What about this instead?
```patch
diff --git a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
index aaadbb66e..36ab320d7 100644
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -879,6 +879,7 @@ class RefChecks extends MiniPhase { thisPhase =>
   }
 
   override def transformNew(tree: New)(implicit ctx: Context) = {
+    checkUndesiredProperties(tree.tpe.typeSymbol, tree.pos)
     currentLevel.enterReference(tree.tpe.typeSymbol, tree.pos)
     tree
   }
```

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -879,6 +879,7 @@ class RefChecks extends MiniPhase { thisPhase =>
   }
 
   override def transformNew(tree: New)(implicit ctx: Context) = {
+    checkUndesiredProperties(tree.tpe.typeSymbol, tree.pos)
     currentLevel.enterReference(tree.tpe.typeSymbol, tree.pos)
```
allanrenucci: Maybe store `tree.tpe.typeSymbol` in a temporary val

tests/neg-custom-args/i2333.scala
```diff
@@ -0,0 +1,5 @@
+@deprecated("bla", "2.11.0") class Foo
+
+object Test {
+  new Foo // error
+}
```
allanrenucci: I would slightly modify the test, to also test secondary constructors:
```scala
@deprecated("bla", "2.11.0") class Foo {
  def this(x: Int) = this()
}

object Test {
  new Foo // error: deprecated
  new Foo(1) // error: deprecated
}
```

allanrenucci:
Sorry for the multiple paths 😄

PR #3813, opened 2018-01-11
Simplify ClassfileParser#innerClasses#classSymbol

smarter:
- Avoid redundant calls to get
- Fix indentation of getMember and add some brackets to make it clearer

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
+++ b/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
@@ -106,7 +106,7 @@ class ClassfileParser(
 
   /** Return the class symbol of the given name. */
   def classNameToSymbol(name: Name)(implicit ctx: Context): Symbol = innerClasses.get(name) match {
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -890,54 +890,42 @@ class ClassfileParser(
...
+            staticScope.lookup(name)
+          else {
+            var module = sym.companionModule
+            if (!module.exists && sym.isAbsent)
+              module = sym.scalacLinkedClass
+            module.info.member(name).symbol
+          }
+        }
+        else {
+          if (sym == classRoot.symbol)
```
allanrenucci: You could reduce nesting:
```scala
if (isStatic(entry.jflags)) {
  ...
}
else if (sym == classRoot.symbol)
  instanceScope.lookup(name)
else
  sym.info.member(name).symbol
```

smarter:
Comments addressed, thanks @allanrenucci !

PR #3819, opened 2018-01-13
Encode large quotes

nicolasstucki:
Based on #3685


odersky:
Needs a rebase

library/src/scala/runtime/quoted/Unpickler.scala
```diff
@@ -8,7 +8,7 @@ object Unpickler {
   /** Representation of pickled trees. For now it's String, but it
```
odersky: Comment is wrong

odersky: I'd probably make it a List[String] instead. SeqLiterals are expensive but `::` is cheap. So single string quotes could be represented as `quoted :: Nil`

PR #3824, opened 2018-01-13
Fix Tasty unpickling of a TEMPLATE with a SHARED parent

smarter:


smarter:
Note: would be good if someone double-checked that there aren't other places where we use `nextByte` when `nextUnsharedTag` would be more appropriate.

smarter:
The test failure is due to a missing position for an unpickled tree, I'll investigate.

smarter:
Ready for review.

PR #3828, opened 2018-01-14
Avoid pickling the same position multiple times

smarter:
Decreases the size of the bootstrapped dotty-compiler jar from 8662 KB
to 8647 KB.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/PositionPickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/PositionPickler.scala
@@ -19,7 +19,7 @@ class PositionPickler(pickler: TastyPickler, addrOfTree: tpd.Tree => Option[Addr
   import buf._
   import ast.tpd._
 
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/tasty/PositionPickler.scala
```diff
@@ -19,7 +19,7 @@ class PositionPickler(pickler: TastyPickler, addrOfTree: tpd.Tree => Option[Addr
   import buf._
   import ast.tpd._
 
-  private val remainingAddrs = new java.util.IdentityHashMap[Tree, Iterator[Addr]]
+  private val pickledIndices = new mutable.HashSet[Int]
```
nicolasstucki: Maybe use a bitset

odersky: Yes, this looks ideal for a BitSet, since it is bounded and fairly dense.

smarter: Done.

PR #3830, opened 2018-01-14
Enable optimisations in `Expr.run`

nicolasstucki:


smarter:
Optimizations have a significant impact on compilation time, and not much effect on running time, why should they be enabled by default?

nicolasstucki:
It should probably be an option of the `run` method. I will change it to be configurable and of by default.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -13,10 +13,9 @@ import java.nio.charset.StandardCharsets
 
 class QuoteDriver extends Driver {
 
-  def run[T](expr: Expr[T]): T = {
+  def run[T](expr: Expr[T], optimise: Boolean): T = {
```
smarter: That will work for now, but in the future if you want to be able to pass more options, it might make more sense to pass a `Settings` object.

nicolasstucki: I will add it now to avoid future changes on the API.

nicolasstucki:
@smarter this PR needs another review, I added `Settings` and a second useful setting for debugging purposes.

compiler/src/dotty/tools/dotc/quoted/Runners.scala
```diff
@@ -31,6 +31,7 @@ object Runners {
   }
 
   case class RunSettings(
-    optimise: Boolean = false
+    optimise: Boolean = false,
+    outDir: Option[String] = None
```
smarter: outputDir, also document the meaning of Some/None here.

smarter: You can do `java -Djdk.internal.lambda.dumpProxyClasses=out` to be able to debug the classes generated for lambda calls at runtime. Maybe we should have a `-Ddotty.meta.dumpClasses` or something for similar purpose.

PR #3832, opened 2018-01-14
Shave of more time from quote-run-large.scala

nicolasstucki:


nicolasstucki:
Execution time went from 15s to 9s on my machine

PR #3837, opened 2018-01-15
Triage FromTasty tests

nicolasstucki:


compiler/test/dotty/tools/dotc/FromTastyTests.scala
```diff
@@ -24,28 +24,104 @@ class FromTastyTests extends ParallelTesting {
     implicit val testGroup: TestGroup = TestGroup("posTestFromTasty")
     val (step1, step2) = compileTastyInDir("../tests/pos", defaultOptions,
       blacklist = Set(
+        /* assertion failed: owner discrepancy for type CC, expected: class <refinement>, found: class <refinement>
+            at scala.Predef$.assert(Predef.scala:219)
+            at dotty.tools.dotc.core.tasty.TreeUnpickler$TreeReader.symbolAtCurrent(TreeUnpickler.scala:195)
+         */
```
liufengyun: What's the motivation for adding error message in the code?

nicolasstucki: Easier to know which is the reason for the failure without rerunning all of them each time. We also had that in the stib blacklist.

smarter: I think it would make more sense to open issues for each of them.

compiler/test/dotty/tools/dotc/FromTastyTests.scala
```diff
@@ -24,28 +24,104 @@ class FromTastyTests extends ParallelTesting {
...
+              at scala.Predef$.assert(Predef.scala:219)
+              at dotty.tools.dotc.transform.TreeChecker$Checker.adapt(TreeChecker.scala:448)
+         */
         "i2944.scala",
+        "t8023.scala",
+
+        /* scala.MatchError: 136 (of class java.lang.Integer)
+            at dotty.tools.dotc.core.tasty.TreeUnpickler$TreeReader.readLengthType$1(TreeUnpickler.scala:231)
+            at dotty.tools.dotc.core.tasty.TreeUnpickler$TreeReader.readType(TreeUnpickler.scala:361)
+         */
```
smarter: Handled by https://github.com/lampepfl/dotty/pull/3824

compiler/test/dotty/tools/dotc/FromTastyTests.scala
nicolasstucki: I will work on minimizing them and open an issue with the minimized code. Then would refer to the group of tests that is likely to be failing because of that particular issue.

PR #3840, opened 2018-01-15
Fix #3478: Optimize away unused defs and lazy vals

### LINKED ISSUE #3478: Optimise away local lazy vals that are not used
Compiling the following code with `-optimise`

```scala
object Test {
    lazy val foo = 1
  }
}
```

currently produces 
```scala
 class Test$ extends Object { 
    def <init>(): Unit = 
      {
        super()
        ()
      }
      {
        lazy var foo$lzy1: dotty.runtime.LazyInt = new dotty.runtime.LazyInt()
        ()
      }
    private def foo$lzyINIT1$1(foo$lzy1$1: dotty.runtime.LazyInt): Int = 
      scala.Int.unbox(
        foo$lzy1$1.synchronized(
          scala.Int.box(
            if foo$lzy1$1.initialized() then foo$lzy1$1.value() else 
              {
                foo$lzy1$1.initialized_=(true)
                foo$lzy1$1.value_=(1)
                foo$lzy1$1.value()
              }
          )
        )
      )
    private lazy <accessor> def foo$1(foo$lzy1$2: dotty.runtime.LazyInt): Int = 
      if foo$lzy1$2.initialized() then foo$lzy1$2.value() else 
        Test.foo$lzyINIT1$1(foo$lzy1$2)
  }
```

The optimizer should be able to eliminate the declaration of `foo$lzy1`. Hopefully before the creation of `foo$lzyINIT1$1` and `foo$1`.

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/localopt/Devalify.scala
```diff
@@ -159,6 +159,9 @@ class Devalify extends Optimisation {
         if (replacements.contains(t.symbol))
           deepReplacer.transform(replacements(t.symbol)).ensureConforms(t.tpe.widen)
         else t
+      case t: DefDef if timesUsed.getOrElse(t.symbol, 0) == 0 =>
```
allanrenucci: Does this remove only local definitions? Otherwise, it is not safe to remove public members and definitions

nicolasstucki: Yes, the Simplifier only transforms local trees.

Issue #3816, opened 2018-01-12
Owner discrepancy when compiling from TASTY


nicolasstucki:
Compiling the following minimization of [`tests/pos/NoCyclicReference.scala`](https://github.com/lampepfl/dotty/blob/master/tests/pos/NoCyclicReference.scala) with `-Ytest-pickler`

```scala
trait Iterable { self =>
  type CC <: Iterable { type CC = self.CC }
}
```
fails with 
```
> dotc -Ytest-pickler Test.scala
exception occurred while compiling tests/allan/Test.scala
Exception in thread "main" java.lang.AssertionError: assertion failed: Cyclic reference while unpickling definition at address 29 in unit null
	at dotty.DottyPredef$.assertFail(DottyPredef.scala:36)
	at dotty.tools.dotc.core.tasty.TreeUnpickler$TreeReader.readIndexedDef(TreeUnpickler.scala:637)
	at dotty.tools.dotc.core.tasty.TreeUnpickler$Completer.complete(TreeUnpickler.scala:91)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.completeFrom(SymDenotations.scala:246)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.completeInfo$1(SymDenotations.scala:209)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.info(SymDenotations.scala:211)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.computeAsSeenFrom(Denotations.scala:1070)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.computeAsSeenFrom(Denotations.scala:1063)
	at dotty.tools.dotc.core.Denotations$PreDenotation.asSeenFrom(Denotations.scala:131)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.findMember(SymDenotations.scala:1599)
	at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:514)
	at dotty.tools.dotc.core.Types$Type.goAnd$1(Types.scala:646)
	at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:532)
	at dotty.tools.dotc.core.Types$Type.goThis$1(Types.scala:615)
	at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:518)
	at dotty.tools.dotc.core.Types$Type.findMember(Types.scala:658)
[...]

PR #3842, opened 2018-01-15
Add cyclic reference detection to TASTY unpickling

smarter:


smarter:
This help find the root causes of issues like https://github.com/lampepfl/dotty/issues/3816 faster:
```scala
> run -Ythrough-tasty tests/pos/NoCyclicReference.scala
[info] Running dotty.tools.dotc.fromtasty.Debug -classpath /home/smarter/opt/dotty/library/target/scala-2.12/dotty-library_2.12-0.6.0-bin-SNAPSHOT-nonbootstrapped.jar tests/pos/NoCyclicReference.scala
Compiling from .scala sources
Compiling TASTY from .class sources
Exception in thread "main" java.lang.AssertionError: assertion failed: Cyclic reference while unpickling tree at address 61
...
```

Issue #2887, opened 2017-07-18
IndexOutOfBoundsException when attempting to reduce ill-kinded types


sstucki:
The following snippet is a variant of that causing #2771:

```Scala
trait A { type L[G[F[_],_],H[F[_],_]] }
trait B { type L[F[_],_] }
trait C { type M <: A }
trait D { type M >: B }

object Test {
  def test(x: C with D): Unit = {
    def foo(a: A, b: B)(z: a.L[b.L,b.L]) = z
    def bar(y: x.M, b: B) = foo(y, b)
    def baz(b: B) = bar(b, b)
    baz(new B { type L[F[_],X] = F[X] })(1)
  }
}
```

The underlying issue is likely the same, though the error message is different:

    exception occurred while compiling Test.scala
    Exception in thread "main" java.lang.IndexOutOfBoundsException: 1
        at scala.collection.LinearSeqOptimized$class.apply(LinearSeqOptimized.scala:65)
        at scala.collection.immutable.List.apply(List.scala:84)
        at dotty.tools.dotc.core.Substituters$class.substParams(Substituters.scala:220)
        ...

The error disappears when the last line (the call to `baz`) is commented out. See #2771 for a more in-depth discussion.

sstucki:
Here is another example, producing a stack overflow again (like for #2771). It uses the untyped SKI fixpoint combinator and absurd bounds to create a non-terminating term.

```Scala
trait A { type S[X[_] <: [_] => Any, Y[_]] <: [_] => Any; type I[_] }
trait B { type S[X[_],Y[_]]; type I[_] <: [_] => Any }
trait C { type M <: B }
trait D { type M >: A }

object Test {
  def test(x: C with D): Unit = {
    def foo(a: A, b: B)(z: a.S[b.I,a.I][b.S[a.I,a.I]]) = z
    def bar(a: A, y: x.M) = foo(a,y)
    def baz(a: A) = bar(a, a)
    baz(new A { type S[X[_] <: [_] => Any, Y[_]] = [Z] => X[Z][Y[Z]]; type I[X] = X })(1)
  }
}
```
Again, the error disappears when the call to `baz` is commented out.

Note the use of curried type operators. One can also write an uncurried version, resulting in yet another error:
```Scala
trait A { type S[X[_,_], Y[_],_]; type I[_] }
trait B { type S[X[_],Y[_]]; type I[_,_] }
trait C { type M <: B }
trait D { type M >: A }

object Test {
[...]

PR #3844, opened 2018-01-15
Sketch of possible typelevel extension of principled metaprogramming

odersky:
This is a sketch of a possible extension to support functional typelevel programming.
It's prose only, there is no implementation. The intention is that, before embarking
on an implementation we discuss the approach in general and the specific details.

### DIFF
--- a/docs/docs/typelevel.md
+++ b/docs/docs/typelevel.md
@@ -95,7 +95,7 @@ containing an expression of type `scala.quoted.Type[T]`, for some type
 A type macro may have a declared upper bound, e.g. `<: Nat` in the
 example above. If an upper bound is given, the computed type `T` must
 conform to it.  If no upper bound is given, `Any` is assumed. The
-upper bound is used to type check recursive calls.
+upper bound is used to type-check recursive calls.
 
 The `toNat` function can now be given a more precise type:
 
@@ -130,7 +130,7 @@ symbolic evaluation and it is not yet clear what the details of that
 would be.
 
 But there is a simpler way: We can _delay_ checking the precise result
-type of the body of an `macro` until the macro is inlined.  The
+type of the body of a macro until the macro is inlined.  The
 declared result type can be fully unfolded based on the static
 information available at the inlining point. At that point the inlined
 expansion of the macro is type-checked using the fully computed types.
@@ -236,7 +236,7 @@ the use of quotes and rely on lifting to go from `Int` to
 constant without relying on constant folding:
 
     inline def toInt[N <: Nat]: Int = ~{
-      case N =: Z => 0
+      case N =:= Z => 0
       case N =:= S[type N1] => 1 + toInt[N1]
     }
 
@@ -286,7 +286,7 @@ typechecks because `=:=` (and `<:<`) will inherit from
 `U`. But we still need to explain why a search for `Xs =:= HCons[type X, type Xs1]`
 will succeed in the right hand side of the case. What happens in detail is this:
 
-When type checking a query case, the type pattern is type-checked
+When type-checking a query case, the type pattern is type-checked
 according to the usual rules. Then the guard and right-hand side are type-checked
 under the assumption that an implicit value of the type pattern is available.
 For instance, when checking
@@ -379,7 +379,7 @@ We can define type-level Booleans analogously to Peano numbers:
 
 Here's a type function that returns the result of comparing two `Nat`
 types `X` and `Y`, returning `True` iff the peano number defined by
-`X` is smaller than the peano number defined by `Y` and `Falsesy` otherwise:
+`X` is smaller than the peano number defined by `Y` and `False` otherwise:
 
     type < [X <: Nat, Y <: Nat] <: Bool = {
       case X =:= Z, Y =:= S[type T1] => True
--- a/docs/docs/typelevel.md
+++ b/docs/docs/typelevel.md
@@ -61,7 +61,7 @@ would not always work, so we'd end up sometimes with code being
 evaluated at runtime to compute a result that is already known at
 compile-time. Since typelevel programming is based on information
 known at compile-time, staging is a natural complement to it since it
-distinuguishes syntactically between computations done at compile time
+distinguishes syntactically between computations done at compile time
 and computations done at runtime. The syntactic distinctions also help
 understanding the code better because they make clear what gets
 computed when.
@@ -117,18 +117,18 @@ methods.
 
 ### Second Step: Nicer Syntax for Type Definitions
 
-The syntax for `ToNat` was a bit clunky, and it turns out that most
-inline type definitions use similar staging constructions. All
+The syntax for `ToNat` was a bit clunky. This is no outlier - it turns out
+that most inline type definitions use similar staging constructs. All
 parameters tend to be inline, and the body is typically a conditional
 expression which returns a quoted type in each branch. We can cut down
-this boilerplate by introducing a shorthand form for parameterized type
-definitions, which allows us to express `ToNat` as follows:
+this boilerplate by introducing a shorthand form for parameterized
+type definitions, which allows us to express `ToNat` as follows:
 
     type ToNat(n: Int) =
       if n == 0 then Z
       else S[ToNat(n = 1)]
 
-The short form expands to exactly the previous definition of `ToNat`.
+The short form expands precisely to the previous definition of `ToNat`.
 To get from a short form type definition to a long form type macro,
 the following rewritings are performed:
 
@@ -156,11 +156,11 @@ as conditions in spliced code. Example:
       case N =:= S[type N1] => '(1 + toInt[N1])
     }
 
-A query conditional is written ` { <cases> }` where each case of the form
+A query conditional is written ` { <cases> }` where each case is of the form
 
     case <type-pattern> <guard> => <expr>
 
-<guard> is either empty or of the form `if <expr>`. Some examples of legal cases are
+`<guard>` is either empty or of the form `if <expr>`. Some examples of legal cases are
 
     case Ord[X] => e1
     case X <:< List[Y] if n == 0 => e2
@@ -208,7 +208,7 @@ constant without relying on constant folding:
 
 ### Example: HLists
 
-Consider the standard definition of `HLists`:
+Consider the standard definition of `HList`s:
 
     class HList
 
@@ -222,7 +222,7 @@ A type-level length function on HLists can be written as follows:
       case Xs =:= HCons[type X, type Xs1] => S[Length[Xs1]]
     }
 
-If we don't want to introduce the type variable `X`, which is undused`, we could
+If we don't want to introduce the type variable `X`, which is unused, we could
 also formulate the type pattern with `<:<` instead of `=:=`:
 
     type Length[Xs <: HList] = {
@@ -241,10 +241,6 @@ Here's a `Concat` type with associated `concat` method:
       case Xs =:= HNil => 'ys
       case Xs =:= HCons[type X, type Xs1] => '(xs.hd :: concat(xs.tl, ys))
     }
-
-    case Xs =:= HNil => 'ys
-
-    type Concat[X, Xs] = typeOf(concat[X, Xs](X.init, Xs.init))
 
 ### Typechecking Query Conditionals
 
@@ -363,7 +359,7 @@ We can then use such comparisons as follows:
       Select[Xs, N]
 
 The example demonstrates how one can get from the world of functional
-programming to te world of logic programming and back. The `<` type
+programming to the world of logic programming and back. The `<` type
 performs implicit queries and wraps the result in a function. The
 `SafeSelect` type uses the result of the function as an implicit
 evidence parameter. This works because `True` is implicit but `False`
@@ -440,7 +436,7 @@ following language extensions:
     of the implicit search may also be bound to a variable. Query conditionals
     can only appear in spliced code.
 
-(1-3) together constitute a logical complement to `inline `def`
+(1-3) together constitute a logical complement to `inline def`
 macros.  After all, if we can abstract over a splice that maps
 expressions of type `Expr[T]` to expressions of type `T`, we should
 also be able to abstract over a splice that maps expresssions of type
@@ -482,14 +478,15 @@ now types can be computed as macros, which means that splice expansion
 must be done during typing, so it cannot be a separate phase anymore.
 This style of "white box macros" poses implementation challenges, such
 as how to keep compilers and IDEs working well in the presence of
-resource hungry or mis-behaving macros. A mis-behaving macros could
+resource hungry or mis-behaving macros. A mis-behaving macro could
 slow down typechecking to unacceptable degrees, consume too much memory,
 or pose a security risk.
 
 Maybe the best way to mitigate the risks is to interpret all macro
 code in the compiler. That way, one can impose limits on the number of
 interpretaton steps per macro or prevent macros from calling external
-libraries or doing system calls.
+libraries or doing system calls. Precedent for this is the D Language
+implementation of meta programming, which seems to be well accepted.
 
 A typer-based implementation of macros also strongly suggests that
 splicing should be treated as syntax instead of being a user-defined
--- a/docs/docs/typelevel.md
+++ b/docs/docs/typelevel.md
@@ -126,7 +126,7 @@ Another tricky aspect is how to ensure that the right-hand side of
 easy to see that we use the same recursive decomposition of `n` in
 both cases and that the result types of each case correspond. But to
 prove this correspondence in general would require some sort of
-symbolic computation and it is not yet clear what the details of that
+symbolic evaluation and it is not yet clear what the details of that
 would be.
 
 But there is a simpler way: We can _delay_ checking the precise result
@@ -136,7 +136,7 @@ information available at the inlining point. At that point the inlined
 expansion of the macro is type-checked using the fully computed types.
 
 That still begs the question how we are going to typecheck the
-_definition_ of a macro. The idea is to to approximate any computed
+_definition_ of a macro. The idea is to approximate any computed
 type at the definition that depends on unknown parameters by its upper
 bound.
 
@@ -144,8 +144,8 @@ E.g., in the case of `toNat`, the type of the recursive call
 `toNat(n - 1)` would be the upper bound `Nat` of the declared return type
 `ToNat(n - 1)`. This gives `S[Nat]` as the type of the else
 clause. The full type of the right hand side `Z | S[Nat]` would then
-be checked against the upper approximation of the result type
-`Nat`. This succeeds, so the definition is deemed type-correct.
+be checked against the upper approximation of the declared result type
+`ToNat(n)`, which is again `Nat`. This succeeds, so the definition is deemed type-correct.
 
 
 ### Second Step: Nicer Syntax for Type Definitions
@@ -379,7 +379,7 @@ We can define type-level Booleans analogously to Peano numbers:
 
 Here's a type function that returns the result of comparing two `Nat`
 types `X` and `Y`, returning `True` iff the peano number defined by
-`X` is smaller than the peano number defined by `Y` and `False otherwise:
+`X` is smaller than the peano number defined by `Y` and `Falsesy` otherwise:
 
     type < [X <: Nat, Y <: Nat] <: Bool = {
       case X =:= Z, Y =:= S[type T1] => True
@@ -489,8 +489,8 @@ grammar](../internal/syntax.md).
                    |  ‘'’ ‘[’ Type ‘]’
 
     Def          ::=  ...
-                   |  ‘type’ ‘def’ DefSig ‘=’ ‘~’ SimpleExpr
-                   |  ‘type’ id [TypTypeParamClause] DefParamClauses ‘=’ Type
+                   |  ‘type’ ‘def’ DefSig [‘<:’ Type] ‘=’ ‘~’ SimpleExpr
+                   |  ‘type’ id [TypTypeParamClause] DefParamClauses [‘<:’ Type] ‘=’ Type
 
     SimpleExpr   ::=  ...
                    |  ‘{’ {Query ‘=>’ Block} ‘}’
--- a/docs/docs/typelevel.md
+++ b/docs/docs/typelevel.md
@@ -140,8 +140,8 @@ _definition_ of a macro. The idea is to to approximate any computed
 type at the definition that depends on unknown parameters by its upper
 bound.
 
-E.g., in the case of `toNat`, the type of the recursive call `toNat(n
-- 1)` would be the upper bound `Nat` of the declared return type
+E.g., in the case of `toNat`, the type of the recursive call
+`toNat(n - 1)` would be the upper bound `Nat` of the declared return type
 `ToNat(n - 1)`. This gives `S[Nat]` as the type of the else
 clause. The full type of the right hand side `Z | S[Nat]` would then
 be checked against the upper approximation of the result type
--- a/docs/docs/typelevel.md
+++ b/docs/docs/typelevel.md
@@ -1,0 +1,501 @@
+# Functional Typelevel Programming
+
+This note is an exploration how we might combine staging with
+typelevel metaprogramming in Dotty.
+
+### State of the Art
+
+Currently typelevel programming uses implicits as the basic mechanism,
+resulting in a programming style much like Prolog. Amazing feats have
+been achieved using this scheme, but things are certainly far from
+ideal. In particular:
+
+ - The logic programming style requires a shift of mindset compared to the
+   usual functional programming style in Scala.
+ - The ways to control implicit search are underdeveloped,
+   leading to complicated schemes, requiring rule prioritization or other hacks.
+ - Because of their conceptual complexity the resulting typelevel programs are often
+   fragile.
+ - Error diagnostics are usually very bad, making programming with implicits somewhat
+   of a black art. Dotty has greatly improved error dignostics for recursive implicits,
+   but the fundamental problem remains.
+
+In the following we develop a system and methodology that lets us do
+typelevel programming in a functional style. The system is based on
+Dotty's symmetric meta programming system, adding some fairly modest
+language extensions.
+
+### Prequel: Conditions in Staged Code
+
+As a perparatory step, we allow if-then-else in spliced code. Given an
+if-then-else expression, the macro-interpreter interprets its
+condition (which must be a compile-time constant) and chooses one of
+the two branches depending on the condition's value. Consider the
+following example mapping an integer value `n` to a peano number
+represented as an instance of type `Nat`.
+
+    trait Nat
+
+    case object Z extends Nat
+    case class S[N <: Nat] extends Nat
+
+    inline def toNat(inline n: Int): Nat = ~{
+      if n == 0 then 'Z
+      else 'S[toNat(n - 1)]
+    }
+
+In order to avoid clutter we use quote syntax without mandatory
+parentheses, e.g. `'Z` instead of `'(Z)`. (This assumes that symbol
+literals are confined to Scala-2 mode.)
+
+Note that the use of staging for `toNat` is recommended but not required. We could have
+written
+
+    inline def toNat(inline n: Int): Nat =
+      if n == 0 then Z
+      else S[toNat(n - 1)]
+
+But then we would rely on inlining's nornalizations to simplify the
+code to the branch that was taken. For more complicated examples this
+would not always work, so we'd end up sometimes with code being
+evaluated at runtime to compute a result that is already known at
+compile-time. Since typelevel programming is based on information
+known at compile-time, staging is a natural complement to it since it
+distinuguishes syntactically between computations done at compile time
+and computations done at runtime. The syntactic distinctions also help
+understanding the code better because they make clear what gets
+computed when.
+
+The ability to use conditionals in spliced code does not add anything
+fundamental, since the same effect can be achieved by defining an
+intermediate function that contains the conditional and is called in a
+splice. By contrast, the next section does introduce a fundamental new
+capability.
+
+### First Step: Type Macros
+
+In the previous example we would like to establish a stronger link
+between the parameter `n` and the result type. For instance `toNat(3)`
+should have static type `S[S[S[Z]]]` instead of `Nat`. To
+get there, we allow macros that produce types. Just like an `inline
+def` defines a macro yielding a value, an `inline type def` defines a macro
+yielding a type.
+
+Example:
+
+    inline type def ToNat(inline n: Int) = ~{
+      if n == 0 then '[Z]
+      else '[S[ToNat(n - 1)]]
+    }
+
+The right-hand side of an `inline type def` must be a splice
+containing an expression of type `scala.quoted.Type[T]`, for some type
+`T`.  A call to the macro in type position is then expanded to `T`.
+
+The `toNat` function can now be given a more precise type:
+
+    inline def toNat(inline n: Int): ToNat(n) = ~{
+      if n == 0 then 'Z
+      else 'S[toNat(n - 1)]
+    }
+
+
+### Handling Recursion
+
+A tricky aspect of recursive type functions is that their expansion
+need not terminate. One could impose syntactic criteria to ensure
+termination, but these interact badly with the implicit conditions
+defined in the next section, since the compiler does in general not
+know the relationship between the different types appearing in such a
+definition.
+
+Therefore, we propose to impose a (user-configurable) limit on the
+number of expansion-steps or the time spent to expand in order to
+prevent the compiler from going into an infinite loop or producing a
+stack overflow, exactly in the same way it is done for normal `inline`
+methods.
+
+### Second Step: Nicer Syntax for Type Definitions
+
+The syntax for `ToNat` was a bit clunky, and it turns out that most
+inline type definitions use similar staging constructions. All
+parameters tend to be inline, and the body is typically a conditional
+expression which returns a quoted type in each branch. We can cut down
+this boilerplate by introducing a shorthand form for parameterized type
+definitions, which allows us to express `ToNat` as follows:
+
+    type ToNat(n: Int) =
+      if n == 0 then Z
+      else S[ToNat(n = 1)]
+
+The short form expands to exactly the previous definition of `ToNat`.
+To get from a short form type definition to a long form type macro,
+the following rewritings are performed:
+
+ - `type` is expanded to `inline type def`.
+ - All value parameters get an `inline` modifier.
+ - If the right hand side consists of a conditional, it is put in a spliced block `~{...}`.
+ - The right hand sides of all conditionals are put in type quotes `'[...]`.
+
+We expect that most computed types coming up in practice will use the
+short form. The long form is kept around, to cater for more advanced
+use cases, and to give a precise meaning to the short form.
+
+### Third Step: Query Conditionals
+
+What about defining the inverse of `ToNat`? This should be a function
+that takes a _type_ of peano numbers and produces a compile-time
+integer _value_. We can emulate the effect of such a function with
+implicits but it's a bit roundabout since we would need several
+overloaded methods each taking a different implicit parameter. A more
+direct, closed formulation is possible if we allow implicit searches
+as conditions in spliced code. Example:
+
+    inline def toInt[N <: Nat]: Int = ~{
+      case N =:= Z => '0
+      case N =:= S[type N1] => '(1 + toInt[N1])
+    }
+
+A query conditional is written ` { <cases> }` where each case of the form
+
+    case <type-pattern> <guard> => <expr>
+
+<guard> is either empty or of the form `if <expr>`. Some examples of legal cases are
+
+    case Ord[X] => e1
+    case X <:< List[Y] if n == 0 => e2
+    case _ if n < 0 => e3
+
+Query conditionals may only appear in spliced code. That sets them
+apart from blocks with normal case clauses which define partial
+functions and therefore can only appear in normal code.  Every query
+conditional is evaluated by the macro interpreter. The macro
+interpreter evaluates each case of a query conditional by performing
+an implicit search for the given type pattern. If the search succeeds,
+it continues with evaluating the guard, if there is one, followed by
+the right-hand side, otherwise it continues with the next case.
+
+Type patterns may bind type variables. For instance,
+`N =:= S[type N1]` binds the type variable `N1`. Bound type variables are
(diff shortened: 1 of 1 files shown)


docs/docs/typelevel.md
```diff
@@ -0,0 +1,498 @@
...
+The syntax for `ToNat` was a bit clunky. This is no outlier - it turns out
+that most inline type definitions use similar staging constructs. All
+parameters tend to be inline, and the body is typically a conditional
+expression which returns a quoted type in each branch. We can cut down
+this boilerplate by introducing a shorthand form for parameterized
+type definitions, which allows us to express `ToNat` as follows:
+
+    type ToNat(n: Int) =
+      if n == 0 then Z
+      else S[ToNat(n = 1)]
```
senia-psm: Typo. Should be `n - 1`

docs/docs/typelevel.md
```diff
@@ -0,0 +1,498 @@
...
+patterns. Currently type identifiers in patterns are treated as bound
+if they start with a lower-case letter and as free otherwise. This
+mirrors the convention for names, but is not very visible. Requiring
+`type` prefixes instead would make programs clearer.
+
+In the code above, the macro interpreter would perform an implicit
+search for the type `N =:= S[type N1]` where `N` is the actual type
+passed to `toInt` and `N1` is a type variable bound by the search.
+If the search succeeds, it returns one plus the result of evaluating
+`toInt[N1]`. If the search fails it returns zero.
```
senia-psm: There is no `case _ => '0` brunch. Should it be something like "If the search fails it results in compilation failure"?

docs/docs/typelevel.md
```diff
@@ -0,0 +1,498 @@
...
+The implicit definition backing a query pattern can be bound to a name `x` using the syntax
+
+    case x: T => ...
+
+Using this device we can make the conversions in the `concat` function above explicit:
+
+    inline def concat[Xs <: HList, Ys <: HList](xs: Xs, ys: Ys): Concat[Xs, Ys] = ~ {
+      case Xs =:= HNil => 'ys
+      case toCons: Xs =:= HCons[type X, type Xs1] => '(toCons(xs).hd :: concat(toCons(xs).tl, ys))
+    }
```
liufengyun: It's not clear to me how the typer checks that the result type is `Concat[Xs, Ys]` at the definition site of `concat`?

odersky: Very good question. I added a section explaining this. Essentially, we use only approximate typing for definitions and do the full typecheck at all expansion sites.

docs/docs/typelevel.md
```diff
@@ -92,14 +92,18 @@ The right-hand side of an `inline type def` must be a splice
 containing an expression of type `scala.quoted.Type[T]`, for some type
 `T`.  A call to the macro in type position is then expanded to `T`.
 
+A type macro may have a declared upper bound, e.g. `<: Nat` in the
+example above. If an aupper bound is given, the computed type `T` must
```
senia-psm: Typo "**a**upper"

docs/docs/typelevel.md
```diff
@@ -115,6 +119,35 @@ prevent the compiler from going into an infinite loop or producing a
...
+would be.
+
+But there is a simpler way: We can _delay_ checking the precise result
+type of the body of an `macro` until the macro is inlined.  The
+declared result type can be fully unfolded based on the static
+information available at the inlining point. At that point the inlined
+expansion of the macro is type-checked using the fully computed types.
+
+That still begs the question how we are going to typecheck the
+_definition_ of a macro. The idea is to to approximate any computed
```
senia-psm: Typo `to to`

docs/docs/typelevel.md
```diff
@@ -379,7 +379,7 @@ We can define type-level Booleans analogously to Peano numbers:
 
 Here's a type function that returns the result of comparing two `Nat`
 types `X` and `Y`, returning `True` iff the peano number defined by
-`X` is smaller than the peano number defined by `Y` and `False otherwise:
+`X` is smaller than the peano number defined by `Y` and `Falsesy` otherwise:
```
smarter: Falsesy?

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+if-then-else expression, the macro-interpreter interprets its
+condition (which must be a compile-time constant) and chooses one of
+the two branches depending on the condition's value. Consider the
+following example mapping an integer value `n` to a peano number
+represented as an instance of type `Nat`.
+
+    trait Nat
+
+    case object Z extends Nat
+    case class S[N <: Nat] extends Nat
```
Blaisorblade: I’m reading and commenting as I go along. This is confusing. It is similar but not identical to “singleton types” à la Haskell (from Tim Sheard’s Ωmega), but unlike those it seems to mix type-level and value level. For numbers this seems to work, but how do you encode correctly the type `Eq[A, B]` so that it’s only inhabited if `A` and `B` are equal? It’s OK to use Peano numbers as examples, but I’d be wary of using them as evaluation.

With singletons you have both type-level numbers:
```scala
trait TNat
case object TZ extends TNat
case class TS(n: TNat) extends TNat
```

and value-level numbers, indexed by type-level numbers, so that `Nat[N]` has *one* inhabitant which is isomorphic to `N`, and so that you can operate on this number at both value and type-level since you have the number at both levels.

```
trait Nat[N <: TNat]
case object Z extends Nat[TZ]
case object S[TN](n: Nat[TN]) extends Nat[TS[TN]]
```

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+intermediate function that contains the conditional and is called in a
+splice. By contrast, the next section does introduce a fundamental new
+capability.
+
+### First Step: Type Macros
+
+In the previous example we would like to establish a stronger link
+between the parameter `n` and the result type. For instance `toNat(3)`
+should have static type `S[S[S[Z]]]` instead of `Nat`. To
+get there, we allow macros that produce types. Just like an `inline
```
Blaisorblade: Are these macros or type-level functions? And if the latter, is it F/FOmega functions or can you also pattern-match on types? Here you just show pattern-matching on compile-time values.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Checking Correspondence Between Types and Terms
+
+Another tricky aspect is how to ensure that the right-hand side of
+`toNat` corresponds to its declared type. In the example above it is
+easy to see that we use the same recursive decomposition of `n` in
+both cases and that the result types of each case correspond. But to
+prove this correspondence in general would require some sort of
+symbolic evaluation and it is not yet clear what the details of that
+would be.
```
Blaisorblade: Dependently-typed languages give one answer, Ωmega gives a slightly different one, SMT solvers yet another one. There can be reasons to ignore all that, but I’m curious which ones they are.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+
+But there is a simpler way: We can _delay_ checking the precise result
+type of the body of an `macro` until the macro is inlined.  The
+declared result type can be fully unfolded based on the static
+information available at the inlining point. At that point the inlined
+expansion of the macro is type-checked using the fully computed types.
```
Blaisorblade: In this case, we would not be *typechecking* the macro, we would be writing a contract on it and enforcing it at “macro runtime” (where “runtime” means the runtime of the staged code, so it’s still while Dotty runs, but it’s not when defining the type function but when running it; as usual with staging, we don’t have just compile-time and runtime). This is an interesting idea and it has certain advantages, but there are well-understood tradeoffs between typechecking and (macro) runtime enforcements of contracts and we should make sure those tradeoffs are appropriate.

It’d be IMHO important to not talk about types because they suggest the wrong assumptions. I can already foresee an example puzzler: “my type function ‘typechecks’ but fails at this new call site”. More annoyingly, bugs in a type-level library might be detected by some user. Calling these “contracts” might reduce the confusion a bit (though it might not be enough); but would this be worth using?
But since many languages can type-check type-level functions, I’d like a reason to do instead research on something new and unproven.
Maybe dealing with a typechecker for type functions is too hard for our target users?
But are there really people who want type-level programming but are willing to admit the downsides of typecheckers (beyond me)?

And if those people exist, are they willing to give up separate typechecking of type-level functions to avoid those downsides? I’m not. Sometimes it’s too hard faking dependent types, as Chlipala will tell to anybody who listens — but then, just don’t fake them at all.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+      else S[ToNat(n - 1)]
+
+The short form expands precisely to the previous definition of `ToNat`.
+To get from a short form type definition to a long form type macro,
+the following rewritings are performed:
+
+ - `type` is expanded to `inline type def`.
+ - All value parameters get an `inline` modifier.
+ - If the right hand side consists of a conditional, it is put in a spliced block `~{...}`.
+ - The right hand sides of all conditionals are put in type quotes `'[...]`.
```
Blaisorblade: It might be good to see more examples of the core language before choosing what’s the sugar we need.

I’ll need to think about this and I might well be wrong, but I doubt non-recursive desugarings are worth the surprises and bugs they cause. Doesn’t type inference on LMS already do better?
As a rule of thumb, metaprograms that aren’t structurally recursive are bugs. When TAing our programming language lab, it was easy to show bugs just by reading the metaprogram; here it’d take longer to figure out if there’s any and which they are.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Third Step: Query Conditionals
+
+What about defining the inverse of `ToNat`? This should be a function
+that takes a _type_ of peano numbers and produces a compile-time
+integer _value_. We can emulate the effect of such a function with
+implicits but it's a bit roundabout since we would need several
+overloaded methods each taking a different implicit parameter. A more
+direct, closed formulation is possible if we allow implicit searches
+as conditions in spliced code. Example:
```
Blaisorblade: If the goal is to avoid needing logic programming, why not allow matching on types? Allowing for implicit searches is pretty interesting idea, but detracts from that goal... and requiring implicit search to use pattern-matching/unification on types seems pretty roundabout?

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+    case _ if n < 0 => e3
+
+Query conditionals may only appear in spliced code. That sets them
+apart from blocks with normal case clauses which define partial
+functions and therefore can only appear in normal code.  Every query
+conditional is evaluated by the macro interpreter. The macro
+interpreter evaluates each case of a query conditional by performing
+an implicit search for the given type pattern. If the search succeeds,
+it continues with evaluating the guard, if there is one, followed by
+the right-hand side, otherwise it continues with the next case.
```
Blaisorblade: This sounds like a mixture of logic and functional programming. How does this interact with backtracking in implicit search? After some experiences with this mixture I’d like a good answer, because this sort of combination can be pretty tricky to program (or too hard to use) with if you don’t have control of backtracking.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+
+A type-level length function on HLists can be written as follows:
+
+    type Length[Xs <: HList] <: Nat = {
+      case Xs =:= HNil => Z
+      case Xs =:= HCons[type X, type Xs1] => S[Length[Xs1]]
+    }
+
+If we don't want to introduce the type variable `X`, which is unused, we could
+also formulate the type pattern with `<:<` instead of `=:=`:
```
Blaisorblade: Why can’t we just use `_` and `<:<` together, and write `Xs =:= HCons[_, type Xs1]`? The semantics would be a conventional match, followed by throwing away `X`, per the usual semantics of `_`. No subtyping should be necessary.

Subtyping might instead be needed if `Xs` were a *subtype* of `HCons[X, Xs1]`. It’d be good to check how `=:=` and subtyping interact and if the interaction is acceptable in a wider range of use cases.

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Typechecking Query Conditionals
+
+Why does the last case of `concat` typecheck, given that `xs` is of
+type `Xs`, which does not have `hd` or `tl` fields? Roughly, it
+typechecks because `=:=` (and `<:<`) will inherit from
+`ImplicitConverter` so in the presence of an implicit search result of
+`T =:= U` or `T <:< U` we also get an implicit conversion from `T` to
+`U`. But we still need to explain why a search for `Xs =:= HCons[type X, type Xs1]`
+will succeed in the right hand side of the case. What happens in detail is this:
```
Blaisorblade: Let me assume `concat` is a `def` not an `inline def` (see above).

We also need to make sure that the recursive call to `concat` typechecks — I can imagine ways in which it’d work, but one needs to pick a typechecking algorithm.

The sketched process does not seem to use type refinement — in particular, it seems that in this branch we don’t know that `Xs` is equal to `HCons[X, Xs1]`, only that they’re convertible both ways. But then, the type of the RHS is `HCons[X, Concat[Xs1, Ys]]`, while the desired return type is `Concat[Xs, Ys]` — without unifying `Xs` with `HCons[X, Xs1]`, why would the two coincide? I now see `concat` is here a query conditional, but see above for why that might not work well enough.

Also, will the code still typecheck if I write the following?

```scala
    inline def concat[Xs <: HList, Ys <: HList](xs: Xs, ys: Ys): Concat[Xs, Ys] = ~ {
      case Xs <:< HNil => 'ys
      case Xs <:< HCons[_, type Xs1] => 'HCons(xs.hd, concat(xs.tl, ys))
    }
```

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+Here's a `Concat` type with associated `concat` method:
+
+    type Concat[Xs <: HList, Ys <: HList] <: HList = {
+      case Xs =:= HNil => Ys
+      case Xs =:= HCons[type X, type Xs1] => HCons[X, Concat[Xs1, Ys]]
+    }
+
+    inline def concat[Xs <: HList, Ys <: HList](xs: Xs, ys: Ys): Concat[Xs, Ys] = ~ {
```
Blaisorblade: Making `concat` inline allows using the rules for typechecking query conditionals, but it does not seem expressive enough. Usually type-level programming lets me define a proper *runtime* function `concat` and typecheck it once and for all; if we can’t do that, it seems we forbid lots of applications of `HList`. And to write a runtime `concat` we have to either do contract-checking for `concat` at the actual runtime, and define erasure for types like `Concat[Xs, Ys]` (so that type errors might manifest not just while running macros but while running functions), or to introduce symbolic evaluation.

docs/docs/typelevel.md
Blaisorblade: I forgot to mention the application of SMT solvers, refinement types and Liquid Haskell/“Liquid Scala” (is that the name @gsps?), like in @gsps Scala’16 paper (https://conf.researchr.org/event/scala-2016/scala-2016-smt-based-checking-of-qualified-types-for-scala). I suspect refinement types have closer to industrial use than dependent types, since their typecheckers end up being “smarter” in many common scenarios.

Blaisorblade:
> The intention is that, before embarking on an implementation we discuss the approach in general and the specific details.

I think that’s an excellent approach, and that’s probably necessary for an extension of such meta-theoretic complexity. I couldn’t resist making some comments as I read the proposal; apologies for any misunderstandings or nonsense I wrote — and sorry if I was too harsh at times.

I’ll get back to preparing for my defense, I hope we’ll have a chance to talk more about this!

docs/docs/typelevel.md
odersky: There's one bit of information I glossed over here: In the definition of `case object Z` I assumed it defines a type as well as an object. I.e. that we expand to

     lazy val Z: Z = new Z
     class Z

Current Scala does not do this. We could of course always use the more explicit definition, in which case type `Nat` would become:

    trait Nat
    class Z extends Nat
    val Z = new Z
    class S[N <: Nat] extends Nat
    def S[N <: Nat] = new S[N]

So that's what I meant in painful detail when I wrote the definitions of `Nat`. The other types are analogous.

docs/docs/typelevel.md
odersky: They are macros that give rise to typelevel functions, as is explained in the following.

docs/docs/typelevel.md
odersky: I believe the fact that you cite in one sentence there _different_ and competing approaches is sufficient reason to be wary whether any of them is ready. If one works out satisfactorily, fine. But right now we do not know, and unless you provide me with proof that one of these approaches would work in all cases at good efficiency without restricting the language in any way, I will stay with my position that "it is not yet clear".

docs/docs/typelevel.md
odersky: Your position is that we should not do any of this until Scala = Idris, or Ur or whatever. Fine. I don't think we have a lot of common ground then.

docs/docs/typelevel.md
odersky: Matching on types is problematic because that's super fragile for a rich type system with subtyping. This observation was the blocker why we did not do any of this before. Implicit search overcomes this blocker, so is IMO really important.

docs/docs/typelevel.md
odersky: Not sure what you are getting at here. How is this not structurally recursive?

docs/docs/typelevel.md
odersky: The implicit query must be resolved at this point. If it is ambiguous it's an error. There is no backtracking over query conditionals.

docs/docs/typelevel.md
odersky: We could probably introduce

    Xs =:= HCons[type _, type Xs1]

if we want to be fancy. But

    Xs =:= HCons[_, type Xs1]

already means something else: That the type `Xs` is equal to the type `HCons[_, T]` for some `T`, where `_` is a wildcard. If we write this out in terms of DOT it would be something like

    HCons { type Hd; type Tl = T }

docs/docs/typelevel.md
odersky: Good question. I would assume yes, but without coming up with a detailed algorithm I would not know for sure.

docs/docs/typelevel.md
odersky: I think this point illustrates that we have different expectations. Defining a runtime `concat` is possible only if the _representation_ of types is homogeneous, so `HLists` or tuples are represented as normal lists at runtime. But there's no way we can or want to do this! So compile-time is the only option.

I think the root issue is that you are coming from a world that is not very applicable to our constraints. We want to achieve similar expressiveness as dependently typed languages, but in a language where efficiency matters, representations have to be specialized, and everything is subject to suptyping. This means we need to be inspired by the likes of C++ or D as well as the sources you cite.

Also, the whole proposal really makes sense only in conjunction with the new meta-programming proposal:

    http://dotty.epfl.ch/docs/reference/symmetric-meta-programming.html

That's where the synergy comes from! Seeing this as an isolated approach to typelevel programming will lead you astray, IMO.

docs/docs/typelevel.md
Blaisorblade: Uh, the last paragraph was misleading — I just meant “looks like I’m not the target user”, which might be wrong. I’m not expecting Scala = Idris/Agda/..., I could rant about those langs for hours :-). But I’d like to tease out advantages and disadvantages as early as possible.

docs/docs/typelevel.md
Blaisorblade: I was assuming that `Z.type` exists; but `S[N]` surprised me because it is partly static and partly dynamic. So for instance I can’t convert `Nat` to a normal number at runtime. Anyway, I’ll try to figure out concretely what can be encoded in this style.

docs/docs/typelevel.md
Blaisorblade: Good point, but doesn’t `_` in `Xs <:< HCons[_, type Xs1]` also refer to a wildcard? Or what do I miss?

odersky: `HCons[Int, Nil] <:< HCons[_, Nil]` but `=:=` between the types does not hold.

docs/docs/typelevel.md
Blaisorblade: Thanks for linking to http://dotty.epfl.ch/docs/reference/symmetric-meta-programming.html, I’ll catch up and sorry if I jumped ahead too much. I also had to look at https://github.com/lampepfl/dotty/pull/2199 to understand the plans for tuples—but there seems to be a common `Tuple` type?

Overall, I’d just like to (help) understand and clarify limitations and tradeoffs of proposals like this, and get to some sweet spot — that is, try to remove any limitations that don’t buy something in exchange — not impose my preference about a certain sweet spot. Of course I’ll compare with related work I’m familiar with, even though that’s not perfect.

Here, you talk about usual type-level programming and standard HLists. So just to understand: is this meant to subsume existing Scala approaches, as the introduction suggests, or is this specialized for a few common scenarios? Current Scala designs allow a runtime `concat` in Scala today.* Now that you mention C++, maybe I understand better what you have in mind; I might later propose edits to the introduction to clarify what you mean. Yet, in C++ having to specialize everything can cause excessive code bloat, so I’d be afraid if we have to go there.

And is a runtime `concat` impossible also if I use some other data structure where I make different choices?

*I checked again https://apocalisp.wordpress.com/2010/07/08/type-level-programming-in-scala-part-6b-hlist%C2%A0folds/ and there are runtime operations—where the type is computed statically. Copy-pasted sample:

```scala
// construct a heterogeneous list of length 3 and type
//  Int :: String :: List[Char] :: HNil
val a = 3 :: "ai4" :: List('r','H') :: HNil
 
// construct a heterogeneous list of length 4 and type
//  Char :: Int :: Char :: String :: HNil
val b = '3' :: 2 :: 'j' :: "sdfh" :: HNil
 
   // append the two HLists
val ab = a ::: b
```

senia-psm: @Blaisorblade please take a look at this code sample, I hope you'll find it useful: https://scastie.scala-lang.org/senia-psm/zqIQMbtgTk6YQj6nlY7Ppw/12

In object `concat_using_=:=` you can find code that works in current version of dotty, but looks similar to `concat` implementation from this PR.

In part "using type macto" I describe my understanding of possible runtime representations of `concat(a, b)`.

Also you may take a look at `manual_concatter` for runtime representation of `a ::: b`.

docs/docs/typelevel.md
odersky: > So for instance I can’t convert Nat to a normal number at runtime. 

Indeed. Type arguments are erased, so there is not enough information available at runtime.

docs/docs/typelevel.md
odersky: I think it's OK to use the contract analogy. But for me these are still types. It's quite analogous to (predicate) refinement types a la Liquid Haskell. Here again we might check them early or late, or check them at run-time.

docs/docs/typelevel.md
odersky: Here's a tongue in cheek version to do a runtime `concat`:

    trait Concatenizer[Xs <: HList, Ys <: HList] {
       def apply(xs: Xs, ys: Ys): Concat[Xs, Ys]
    }

    implicit inline def concatenizer[Xs <: HList, Ys <: HList]: Concatenizer[Xs, Ys] = new {
       def apply(xs: Xs, ys: Ys) = concat(xs, ys)
    }

   def runtimeConcat[Xs <: HList, Ys <: HList](xs: Xs, ys: Ys)(implicit conc: Concatenizer[Xs, Ys]) = 
      conc.apply(xs, ys)

That's indeed a runtime concat but it defers to an implicit that does the actual work and that is synthesized at compile time. The example you were pasting is the same in spirit. And yes, that can generate _lots_ of code.

docs/docs/typelevel.md
LPTK: This is a really important problem in practice, and not just in the context of metaprogramming – the type-checker does not leverage `<:<` evidence to refine the set of known subtyping relations.

One way I've found that works in Dotty (but not scalac) to help work around this problem is to define `<:<` as:

```scala
abstract class <:< [-A,+B] { type Ev >: A <: B }
```

Then we can use explicit type ascriptions to lead the type from one side to the other, as in:

```scala
def foo[S,T](xs: List[S])(implicit ev: S <:< T): List[T] = (xs: List[ev.Ev])
```

docs/docs/typelevel.md
Blaisorblade: Ah the existing example is already using the wildcard semantics, not the throw-away semantics! Hadn’t realized. For later: a syntax less ambiguous for readers might be nice.

docs/docs/typelevel.md
Blaisorblade: IIRC the Scala’17 paper on algorithmic DOT describes the very need for such ascriptions.

Though in Scala you can *call* `<:<` to use it as a coercion function, though IIRC Scalac (used to?) produce for this an actual method call (which JIT can probably inline).

docs/docs/typelevel.md
Blaisorblade: Well, here at least the values *can* be picked at runtime, so we keep a phase distinction and specialize per shape and not per values. I’m curious how much of the shape one needs to reify at compile-time, and if staging lets us have control of this.

But I’ll have to study more closely both HList and @senia-psm’s snippet!

docs/docs/typelevel.md
Blaisorblade: What I was concerned about was “if the right hand side consists of a conditional”. What if the conditional is only part of the RHS, or nested inside other conditionals?

docs/docs/typelevel.md
LPTK: > IIRC the Scala’17 paper on algorithmic DOT describes the very need for such ascriptions.

That's right! I was probably inspired by that talk when I found the trick.

> Though in Scala you can call <:< to use it as a coercion function

Note that this is not sufficient by any account. In my example, it means we have to rebuild a new list `xs.map(ev)` which is purely accidental runtime overhead (not present in `foo` above). In some cases, it's not even possible to rebuild the value this way. 

An even harder case is to use `S` in type argument position where a bound `<: T` is required. To make that one work, I think we'd need to use `ev.Ev` instead of `S` but also make `ev.Ev` equivalent to `S`, as in:

```scala
abstract class <:< [A,+B] { type Ev >: A <: B & A }
```

docs/docs/typelevel.md
LPTK: What's not explicit in the proposal is when a case-defined type like `Concat[A,B]` is expanded. In `trait Concatenizer`, it can clearly not be expanded because we don't know `Xs` and `Ys` yet. This seems to imply that expansion of a case-defined type `T` would happen only while checking that a term `t` has type `T`, as in the definition of `apply` in `concatenizer` where we check that `concat(xs, ys)` has type `Concat[Xs, Ys]`. Is that right?

LPTK: The status of the `Concat[Xs, Ys]` type in `trait Concatenizer` in relation with logic programming makes it appear that this approach is somewhat related to type projections.

A case-defined type such as `Concat[A,B]` can be encoded as `type Concat[A,B] { type Out }` (as in @senia-psm's code) where `Out` is the result of the type-level computation and where each case of the case-defined type is an implicit definition (potentially mutually-recursive with the other cases). Then `Concat[Xs, Ys]` in `trait Concatenizer` has to be `Concat[Xs, Ys] # Out` because there is no implicit value in scope at this point.

So this seems to regain some of the extra flexibility that we had with type projections (which was arguably one reason why they were so pervasive in type-level programming). Hopefully this new approach does not have the same soundness problems as type projections.

senia-psm: Usage of `Concat` in trait looks really confusing for me. I would expect something like this:
```
trait Concatenizer[Xs <: HList, Ys <: HList] {
  type Out <: HList
  def apply(xs: Xs, ys: Ys): Out
}
implicit inline def concatenizer[Xs <: HList,
                             Ys <: HList]: Concatenizer[Xs, Ys] { type Out = Concat[Xs, Ys] } = new {
   def apply(xs: Xs, ys: Ys): Concat[Xs, Ys] = concat(xs, ys)
}
```

docs/docs/typelevel.md
odersky: > What I was concerned about was “if the right hand side consists of a conditional”. What if the conditional is only part of the RHS, or nested inside other conditionals?

Yes, we need to be more precise here. Any nested tree of conditionals (both if-then-else and query) qualifies.

docs/docs/typelevel.md
odersky: Let's forget about the example, which was anyway tongue-in-cheek.

docs/docs/typelevel.md
Blaisorblade: (Should we move this topic to a separate issue?)

> Note that this is not sufficient by any account.

Good point, so I suspect I’m in favor of adding `Ev`. I think scalaz/cats have replacements for `<:<` (`Liskov`?), implemented with casts. Similarly, Haskell added the whole `Coercible` typeclass and friends, to play the role of `=:=` while solving this problem. `Coercible` (and *roles*) seem overly complicated but at least can address similar problems.

> An even harder case [...]

Argh. I think you’re right, but I might rewrite the call site to take `S <:< T` instead. But I’m not sure that’s satisfactory, and it currently adds overhead (not sure if Dotty phantom types can help).

I and @sstucki had reasons to consider some *built-in* `<:<` for DOT that the typechecker understands as subtyping evidence, but we never discussed exposing it to users. I’d prefer an encoding such as what you show, though the asymmetry there bugs me (and would be trouble when using `T` for a type variable with bound `>: S`.
What about
```scala
abstract class <:<[-A, +B <: A] {
  type Ev >: A <: B
  type EvB >: A <: A & B /* = B */
  type EvA >: A | B <: B /* = A, sort-of-dual to the above */
}
```
? I’ve also thought about just having `abstract class <:<[-A, +B <: A]` but it has no type members, and type members don’t have declaration-site variance annotations.

LPTK: > (Should we move this topic to a separate issue?)

I would, but @odersky told us at POPL (jokingly I believe) that if we want to open a new issue, we have to close two old ones first 😅 
Anyway, it's probably easier to ask him directly.

It's actually pretty cool that in Dotty you don't need unsafe casts to achieve the type coercion. It does flow naturally from the theory, where objects with bounded abstract type members can truly be witnesses of subtyping relations. AFAIK this constituted one of the difficult parts of proving the core calculus sound, so we better make use of it 😄 

I like the symmetry of your alternative, but then `Ev` seems really redundant, as a type floating in between without providing anything useful. So I'd remove it:

```scala
abstract class <:<[-A, +B <: A] {
  type ConstrainedB >: A <: A & B // = B with constraints
  type ConstrainedA >: A | B <: B // = A with constraints
}
```

I think the next logical step would be to for the type checker to automatically leverage subtyping relations found as members of values in the current scope (maybe only consider those from implicit values?), though as explained in the Scala'17 talk IIRC it's impossible to make this approach complete in general.

> I’ve also thought about just having abstract class `<:<[-A, +B <: A]`

I've tried that too, but it doesn't work with how subtyping knowledge is propagated:

```scala
def foo[A,B](x: A <:< B) = x
                      ^
                      Type argument B does not conform to upper bound A
```

Blaisorblade: > [On removing Ev]
Sure.

> [automatically leverage subtype relations]

I should look at the paper again, but guessing for now: we can’t look arbitrarily deep into objects to find these relations (I guess), but a limited search might be useful?

> I've tried that too, but it doesn't work with how subtyping knowledge is propagated:

One could then write `foo[A, B <: A]`, and you have to with Scalac, but I thought Dotty tried to accept code like yours? If you can call `foo`, then `A <:< B` already proves that `A <: B`, no need to propagate the constraint at all call sites.

sstucki: > > Note that this is not sufficient by any account.
>
> Good point, so I suspect I’m in favor of adding `Ev`. I think scalaz/cats have replacements for `<:<` (`Liskov`?), implemented with casts.

I just wanted to mention that [as of v2.13 of the standard library](https://github.com/scala/scala/pull/5623), `<:<` finally behaves like Liskov inequality, i.e. you can now do casts at almost no costs (they are the identity function) even if the related types are deeply nested in other type operators, using `substitute` of `liftCo`/`liftContra`. E.g. if you have evidence `p: A <:< B` you get `p.liftCo[List]: List[A] <: List[B]` for "free" (you still need to call `liftCo` manually).

Adding `Ev` to `<:<` seems appealing, but – as you discussed – has the drawback of working only in dotty. It's also potentially brittle since it uses inconsistent bounds. Since the `Ev` version is "extensional" (i.e. the type checker can reason about `Ev` directly), it might cover more cases, but I'm not sure about this. Maybe someone has an example where the `Ev` version works but the Liskov version (`liftCo`/`liftContra` + explicit cast) does not work?

LPTK: Thanks @sstucki, I had missed that. It's good to know!
 
It's true that `liftCo`/`liftContra` are sufficient in the `List` example, but they can become quite more cumbersome to use, as the user needs to pass in a type lambda when the coerced type does not have the straightforward one-argument shape of `T[_]`.

Here is an example where `liftCo`/`liftContra` actually cannot be used but where our approach works – or at least I don't see an easy way to use `liftCo`/`liftContra` in that case:

```scala
abstract class <:<[A, B] {
  type ConstrainedB >: A <: A & B // = B with constraints
  type ConstrainedA >: A | B <: B // = A with constraints
}
def foo[A,B<:A] = Set.empty[(A,B)]
def bar[A,B](implicit ev: B <:< A): Set[(A,B)] = foo[A,ev.ConstrainedB]
```

Not sure I understand why you say the bounds in `<:<` are inconsistent. Having an instance of it is a proof that they are indeed consistent. This might not happen in practice (for runtime performance reasons), but conceptually we would create a new instance of `A<:<B` for any `A` and `B` where we know `A<:B`, and encode that knowledge in the abstract type members.

EDIT: maybe you meant that they are not _syntactically_ consistent.

sstucki: You welcome @LPTK! Regarding your comments:

> Here is an example where `liftCo`/`liftContra` actually cannot be used (...)

Actually, this is a case for `liftContra`, but it's very cumbersome indeed. Here's a possible implementation:
```Scala
trait QFun[A, -X] { def apply[Y <: X]: Set[(A, Y)] }
def bar[A,B](implicit ev: B <:< A): Set[(A,B)] = {
  type Ctx[-X] = QFun[A, X]
  def qFoo[A] = new QFun[A, A] { def apply[Y <: A]: Set[(A, Y)] = foo }
  liftContra[B, A, Ctx](ev)(qFoo).apply
}
```
The `QFun` and `qFoo` machinery is necessary because polymorphism/universal quantification is restricted to trait and method definitions in Scala. If the language had System-F-style quantifiers, we could have written the example as follows:
```Scala
def bar[A,B](implicit ev: B <:< A): Set[(A,B)] = {
  type Ctx[-X] = ∀Y <: X. Set[(A, Y)]
  liftContra[B, A, Ctx](ev)(ΛY <: A.foo[A, Y])
}
```
But that's still much more clumsy than your version using bounds-based evidence directly.

> Not sure I understand why you say the bounds in `<:<` are inconsistent.

Sorry for the confusion re. consistency. I'm using the following terminology (adopted from my thesis): the bounds `A`, `B` of an abstract type `X >: A <: B` are
 * _consistent_ if we can prove that `A <: B` (in the context where `X` is defined),
 * _inconsistent_ if we cannot prove that `A <: B`,
 * _absurd_ if they are inconsistent even as closed types.

The difference between inconsistent and absurd bounds has to do with assumptions. E.g. the bounds of `X >: Any <: Nothing` are inconsistent, no matter the context, so we say they are absurd. But in your definition of `<:<`, we can neither prove that the bounds of `Ev >: A <: B` are consistent nor that they are absurd because we really don't know anything about `A` and `B` (they are abstract).

One could instead write "possibly inconsistent" but, personally, I prefer the shorter version.

As for the safety of inconsistent assumptions, note that, while inconsistency doesn't break safety of DOT, it [can make your compiler crash](https://github.com/lampepfl/dotty/issues/2887#issuecomment-316177441). In essence: reduction of open expressions under inconsistent bounds is unsafe, both for open terms _and_ open types. And contrary to open terms, open types _are_ routinely being reduced (at compile time).

nafg:
Possibly somewhat relevant: https://github.com/Microsoft/TypeScript/pull/21316

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+literals are confined to Scala-2 mode.)
+
+Note that the use of staging for `toNat` is recommended but not required. We could have
+written
+
+    inline def toNat(inline n: Int): Nat =
+      if n == 0 then Z
+      else S[toNat(n - 1)]
+
+But then we would rely on inlining's nornalizations to simplify the
```
DanielRosenwasser: nornalizations -> normalizations

docs/docs/typelevel.md
```diff
@@ -0,0 +1,532 @@
...
+        }
+      case _ => throw new BadIndex
+    }
+
+### Summary of Extensions
+
+Compared to the basic meta programming system of Scala, we propose the
+following language extensions:
+
+ 1. `implicit type def` macros which work like `inline def` macros, but map a right hand side
```
Jasper-M: I'm guessing you meant `inline type def`.

docs/docs/typelevel.md
LPTK: @sstucki
> The QFun and qFoo machinery is necessary because polymorphism/universal quantification is restricted to trait and method definitions in Scala

Interestingly, since the [recent introduction](https://github.com/lampepfl/dotty/pull/3464) of dependent function types, we now have a lightweight way to do first-class polymorphism! Indeed, we can write types such as `(x: {type T}) => List[x.T] => x.T`...

So, using it to define `liftContra` as in:

```scala
abstract class <:<[-From, +To] {
  protected type Ev >: From <: To
  def liftContra[F[-_]](f: F[To]): F[From] = f: F[Ev]
}
```

We can, today, write with Dotty:

```scala
def bar[A,B](implicit ev: B <:< A): Set[(A,B)] =
  ev.liftContra[
    [-X] => (p:{type T<:X}) => Set[(A,p.T)]
  ](p => foo[A,p.T])(new Poly{type T=B})
```

Where `trait Poly { type T }` is just used to avoid a type inference problem we run into with a bare `new{type T=B}`.

odersky:
The (current state of my thinking)[https://github.com/dotty-staging/dotty/blob/add-transparent/docs/docs/typelevel.md] is now a part of #4616.

PR #3845, opened 2018-01-16
Quote lib

nicolasstucki:
Part of this code would be added as part of some external library for operations on quotes.
The rest is to add more test cases on complex quotes.

### DIFF
--- a/tests/run-with-compiler/quote-lib.check
+++ b/tests/run-with-compiler/quote-lib.check
@@ -1,0 +1,1 @@
+quote lib ok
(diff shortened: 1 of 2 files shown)


nicolasstucki:
This contains some good regression tests

propensive:
LBTM or LBGTM?

nicolasstucki:
I have a fix for the syntax highlighting issue in for VS code in [this extension](https://marketplace.visualstudio.com/items?itemName=nicolasstucki.scala). We will try to push this fix in the syntax template to GitHub as soon as possible.

PR #3846, opened 2018-01-16
Add quote-run-large.scala to pending

nicolasstucki:
This is a temporary to avoid timeouts in the CI
while performance in improved for large quotes.


nicolasstucki:
will do

PR #3848, opened 2018-01-16
Revert "Fix #3782: Fix type parameter inference with scala.Singleton"

### LINKED ISSUE #3782: failure to infer literal types from literal terms
```scala
  class Foo[T](val value : T) {type Out = T} 
  object Foo {
    def fromXInt[T <: Int with Singleton](i : T): Foo[T] = new Foo[T](i)
  }
  val f1 = Foo.fromXInt(1)
```
```
8 |  val f1 = Foo.fromXInt(1)
  |               ^
  |        Type argument Int does not conform to upper bound Int & Singleton
```

nicolasstucki:
This reverts commit 6a8c29a78e9eaeb9176fce75549133389324c443.


nicolasstucki:
It has a possible conflict with 6ce1ac777e383f18cd6b05ac8fccd264d4b81cd7

allanrenucci:
Merging to unblock CI

PR #3849, opened 2018-01-16
Add toExpr extension method to all liftable classes

nicolasstucki:
Requires importing Liftable._


odersky:
LGTM, but needs a rebase

PR #3852, opened 2018-01-16
Fix interpretation of Typed trees

nicolasstucki:


compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -111,10 +111,13 @@ class Interpreter(implicit ctx: Context) {
         val env2 = bindings.foldLeft(env)((acc, x) => interpretStat(x, acc))
         interpretTreeImpl(expansion, env2)
 
+      case Typed(expr, _) =>
+        interpretTreeImpl(expr, env)
```
liufengyun: Do we need a check here? Maybe not.

tests/run/quote-splice-interpret-1.check
```diff
@@ -0,0 +1,2 @@
+true
+false
```
liufengyun: How can I be sure the output is from compile-time, not run-time?

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
nicolasstucki: What kind of check? This is a fully typed tree, we evaluate it's erased version.

liufengyun: I was thinking to check the type, but it's not the semantics of the language. So ignore the comment.

tests/run/quote-splice-interpret-1.check
nicolasstucki: Because the `inline def` has a top level `~` it can always be inlined. Then the inlined tree is interpreted to evaluate the `Expr[T]` which is then spliced in (see #3853 for a bit of info on that regard).
In short a macro call is guaranteed to be inlined or the compilation fails.

liufengyun:
Thanks for the explanation, it makes sense. Move the code to a uncalled
method can also remove any doubt.

Le 17 janv. 2018 6:30 PM, "Nicolas Stucki" <notifications@github.com> a
écrit :

liufengyun:
I mean in the test it's possible to move `println(isZero(0))` out of the `main` method.

PR #3865, opened 2018-01-18
Fix #2760: Support defaults in Java annotations parsed from sources

### LINKED ISSUE #2760: "missing argument" for default annotation parameter to Java interface
```scala
$ cat <<EOT >> bench.scala
package foo

@Fork(value = 16)
class Bar
EOT
$ cat <<EOT >> fork.java
package foo;

public @interface Fork {
    int value() default -1;
    int warmups() default -1;
}
EOT
$ scalac bench.scala fork.java # using 2.12.2
$ dotc bench.scala fork.java
-- Error: bench.scala:3:0 ------------------------------------------------------
3 |@Fork(value = 16)
  |^^^^^^^^^^^^^^^^
  |missing argument for parameter warmups of constructor Fork: (value: Int, warmups: Int)foo.Fork

one error found

```

smarter:


tests/pos-java-interop/i2760/Test.scala
```diff
@@ -0,0 +1 @@
+@Fork(value = 16) class Test
```
odersky: How do we know we get the right default in the backend? Can we make this a run test?

smarter: Yes, done :).

smarter:
Hmm, the run test revealed another issue:
```scala
Exception in thread "main" java.lang.NoSuchFieldError: value
        at Test$.main(Test.scala:6)
        at Test.main(Test.scala)
```

Calling `fork.value` fails because `Fork` has a method `value`, not a field `value`, but the Java parser outputs a val:
```scala
  abstract class Fork(val value: Int = _root_.scala.Predef.???, 
    val warmups: Int = _root_.scala.Predef.???
  ) extends _root_.scala.annotation.Annotation with _root_.java.lang.annotation.
    Annotation
   with _root_.scala.annotation.ClassfileAnnotation {}
```

and the `Getters` mini-phase wil not transform `JavaDefined` symbols. I think the simplest thing to do here is to use defs instead of vals and create default getters by hand like ClassfileParser does: https://github.com/lampepfl/dotty/blob/7058f69fe1852b7776e703692555a0b4e4dcf37f/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala#L640-L645

smarter:
Fixed by tweaking the parser output, see last commit.

Issue #3721, opened 2017-12-31
Slight performance regressions in recent PRs


smarter:
Looking at http://dotty-bench.epfl.ch/ it looks like we've become about ~2% slower on some of the tests recently. It's hard to see due the variance but I think the following PRs are responsible for this:
- [#3634:  A symmetric meta programming framework](https://github.com/lampepfl/dotty/pull/3634) by @odersky 
- [#3575: Reduce the depth of trees generated in PatternMatcher](https://github.com/lampepfl/dotty/pull/3575) by @nicolasstucki 

I'm opening this issue because I think it would be nice to know why these PRs had a performance impact and whether this can be mitigated in some way. 2% isn't a big deal so if we can't figure it out we can just close the issue.

odersky:
This will be fixed by #3866.

PR #3867, opened 2018-01-18
Fix #3866: Add support for typetags

### LINKED ISSUE #3866: Unpickling type in quote fails when not given explicitly
The code in the quote in `f` cannot be unpickled
```
import dotty.tools.dotc.quoted.Runners._
import scala.quoted._
object Test {
    def f[T](x: Expr[T])(implicit t: Type[T]) = '{
      val z = ~x
    }
   println(f('(2))(Type.IntTag).show)
  }
}
```

This quote will be pickled as

```
type T = [[1| ]]
{
  val z: T = [[0| ]]
  ()
}
```

After `ReifyQuotes` the tree is
```scala
package <empty> {
  import dotty.tools.dotc.quoted.Runners._
  import scala.quoted._
  @scala.annotation.internal.SourceFile("Foo.scala") final module class Test$()
     extends
   Object() { this: Test.type => 
      {
        def f[T >: Nothing <: Any](x: quoted.Expr[T])(implicit t: quoted.Type[T]
          )
        : scala.quoted.Expr[Unit] = 
          scala.runtime.quoted.Unpickler.unpickleExpr[Unit](
            scala.collection.immutable.Nil.::[String](
              
                "\\¡«\037\0203\0200\00\00\00\00\00\00\00\00\00\00\00\00\016\02M0£\01\0204ASTs\01\0206_root_\01\0201\'\01\0203Any\01\0205scala\01\0201z\01\0201T\0200¤\0200¢6\0201\0201\0236\0202u\02036\0204\0214\0226\0214\0211\02\0201\0206\02055\0230ÿ\0201\0200\0203\0211\0206£\0205ÿ\0201\02012\0235\025\025"
              
            )
          , [x,new scala.quoted.Type[T]() : Any])
        println(
          f[Int](
            scala.runtime.quoted.Unpickler.unpickleExpr[Int](
              scala.collection.immutable.Nil.::[String](
                
                  "\\¡«\037\0203\0200\00\00\00\00\00\00\00\00\00\00\00\00\n%fð\0235\01\0204ASTs\01\0206_root_\01\0201\'\01\0203Any\01\0205scala\0200\0216\0200\02146\0201\0201\0210\0202u\02036\0204<\0202\025"
                
              )
            , [ : Any])
          )(quoted.Type.IntTag).show(
            dotty.tools.dotc.quoted.Runners.runner[Unit]
          )
        )
      }
  }
}
```

Note that the arguments for the holes are `x` and `new scala.quoted.Type[T]()` hence when unpickling we get an instance of `scala.quoted.Type` with an erased `T`. Instead of `new scala.quoted.Type[T]()` we should have `t`.

If we have 
```scala
def f[T](x: Expr[T])(implicit t: Type[T]) = '{
   val z: ~t= ~x
}
```
then the hole receives the `t` and works as expected.

odersky:
Based on feedback with the early implementation, we found out that we need type lifting
in the framework. The first commit explains what needs to be done in the meta programming doc.

The other commits are a first implementation. I am still not quite happy with it because we are lacking a way to do quoting and splicing directly on types. So for now we generate a type tree, but this cannot express all possible positions of typetags. For instance, we could not splice
a type tag in the position of a type parameter of a refined method since refined methods are
represented as types, not as type trees.

So maybe we need to take a step back and allow that. But that would upset quite a few core structures of the compiler.


odersky:
test performance please

docs/sidebar.yml
```diff
@@ -47,8 +47,8 @@ sidebar:
               url: docs/reference/trait-parameters.html
             - title: Inline
               url: docs/reference/inline.html
-            - title: Symmetric Meta Programming
-              url: docs/reference/symmetric-meta-programming.html
+            - title: Meta Programming
+              url: docs/reference/principled-meta-programming.html
```
liufengyun: The file name needs to be changed to `principled-meta-programming.md` as well.

library/src/scala/quoted/Type.scala
```diff
@@ -9,6 +9,8 @@ abstract class Type[T] extends Quoted {
 /** Some basic type tags, currently incomplete */
 object Type {
 
+  implicit def apply[T]: Type[T] = ???
```
nicolasstucki: Are calls to this methods going to be replaced by the tags below when possible and replaced `'[T]` if the type is statically known?

odersky:
I think we found our performance regression culprit causing #3721: `isQuote` and `isSplice` did a member lookup on every call because they were implemented in a sub-optimal way in `Definitions`.
Fixed by e6664ed

docs/docs/reference/principled-meta-programming.md
```diff
@@ -72,13 +72,18 @@ The two types can be defined in package `scala.quoted` as follows:
 
     package scala.quoted
 
-    abstract class Expr[T] {
+    sealed abstract class Expr[T] {
       def unary_~: T   // splice operation
     }
-    class Type[T] {
+    sealed abstract class Type[T] {
```
nicolasstucki: @odersky I created #3883 to make these classes `sealed`, no need to implement it on this PR.

docs/docs/reference/principled-meta-programming.md
```diff
@@ -163,14 +168,57 @@ quote but no splice between the parameter binding of `T` and its
...
+
+In this version of `reflect`, the type of `x` is now the result of
+splicing the `Type` value `t`. This operation _is_ splice correct -- there
+is one quote and one splice between the use of `t` and its definition.
+
+To avoid clutter, the Scala implementation tries to convert any phase-incorrect
+reference to a type `T` to a type-splice, by rewriting `T` to `~implicitly[Type[T]]`.
+For instance, the user-level definition of `reflect`:
+
+    def reflect[T: Type, U](f: Expr[T] => Expr[U]) Expr[T => U] =
```
nicolasstucki: Still missing a `:`

docs/docs/reference/principled-meta-programming.md
```diff
@@ -163,14 +168,57 @@ quote but no splice between the parameter binding of `T` and its
...
+To avoid clutter, the Scala implementation tries to convert any phase-incorrect
+reference to a type `T` to a type-splice, by rewriting `T` to `~implicitly[Type[T]]`.
+For instance, the user-level definition of `reflect`
+
+    def reflect[T: Type, U](f: Expr[T] => Expr[U]) Expr[T => U] =
+      ’{ (x: T) => ~f(’(x)) }
+
+would be rewritten to
+
+    def reflect[T: Type, U](f: Expr[T] => Expr[U]) Expr[T => U] =
```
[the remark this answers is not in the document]
nicolasstucki: Same here

PR #3868, opened 2018-01-18
Fix #3721: Roll back SubstMap functionality into general TreeTypeMap

### LINKED ISSUE #3721: Slight performance regressions in recent PRs

odersky:
Adding SubstMap is a bit clunky and might have been a performance regression.
Trying to do without it by rolling back its hehavior to regular TreeTypeMaps.
Downsides:

 - loss of modularity (TreeTypeMaps have to know about holes)
 - possible loss of error checking (normal TreeTypeMaps no longer crash when seeing a hole)

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/TreeTypeMap.scala
+++ b/compiler/src/dotty/tools/dotc/ast/TreeTypeMap.scala
@@ -7,6 +7,7 @@ import Types._, Contexts._, Constants._, Names._, Flags._
 import SymDenotations._, Symbols._, Annotations._, Trees._, Symbols._
 import Denotations._, Decorators._
 import dotty.tools.dotc.transform.SymUtils._
+import core.tasty.TreePickler.Hole
 
 /** A map that applies three functions and a substitution together to a tree and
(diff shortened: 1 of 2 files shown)


odersky:
test performance please

odersky:
No change in performance, and we have identified another likely culprit. So it comes down to whether the simplification is worth it.

odersky:
I think the simplification is probably large enough to justify the change.

PR #3869, opened 2018-01-18
Fix #3675: don't unnecessarily wrap arrays when eliminating java varargs

### LINKED ISSUE #3675: Dotty unnecessarily wraps arrays used in place of java repeated
```scala
import java.nio.file._
class Test {
  def test(xs: Array[String]) = {
    val p4 = Paths.get("Hello", xs: _*)
  }
}
```
The code above is compiled by Dotty to:
```scala
public class Test {
    public void test(String[] xs) {
        Path p4 = Paths.get("Hello", (String[])Arrays$.MODULE$.seqToArray((Seq)Predef$.MODULE$.wrapRefArray((Object[])xs), String.class));
    }
}
```
And scalac:
```scala
public class Test {
    public void test(String[] xs) {
        Path p4 = Paths.get("Hello", xs);
    }
}
```

abeln:
When eliminating varargs in a java method call, we need to convert Seqs
into Arrays. However, if the object we're passing is already a wrapped
array, we just need to unwrap it.

Testing
-------
Added a unit test that checks that
```
import java.nio.file._
class Test {
  def test(xs: Array[String]) = {
     val p4 = Paths.get("Hello", xs: _*)
  }
}
```
now gives

`val p4: java.nio.file.Path = java.nio.file.Paths.get("Hello", xs)` (i.e. there's no call to wrapXArray)

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -9,6 +9,7 @@ import scala.collection.{ mutable, immutable }
 import PartialFunction._
 import collection.mutable
 import util.common.alwaysZero
+import dotty.tools.dotc.transform.TreeGen
 
 object Definitions {
 
@@ -338,6 +339,12 @@ class Definitions {
     def Predef_classOf(implicit ctx: Context) = Predef_classOfR.symbol
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
+    // The set of all wrap{X, Ref}Array methods, where X is a value type
+    lazy val Predef_wrapArrayR: Set[TermRef] = {
+      val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
+      methodNames.map(ScalaPredefModule.requiredMethodRef).toSet
+    }
+    def Predef_wrapArray(implicit ctx: Context): Set[Symbol] = Predef_wrapArrayR.map(_.symbol)
 
   lazy val ScalaRuntimeModuleRef = ctx.requiredModuleRef("scala.runtime.ScalaRunTime")
   def ScalaRuntimeModule(implicit ctx: Context) = ScalaRuntimeModuleRef.symbol
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -340,11 +340,10 @@ class Definitions {
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
     // The set of all wrap{X, Ref}Array methods, where X is a value type
-    lazy val Predef_wrapArrayR: collection.Set[TermRef] = {
+    val Predef_wrapArray = new PerRun[collection.Set[Symbol]]({ implicit ctx =>
       val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
-      methodNames.map(ScalaPredefModule.requiredMethodRef)
-    }
-    val Predef_wrapArray = new PerRun[collection.Set[Symbol]](implicit ctx => Predef_wrapArrayR.map(_.symbol))
+      methodNames.map(ScalaPredefModule.requiredMethodRef(_).symbol)
+    })
 
   lazy val ScalaRuntimeModuleRef = ctx.requiredModuleRef("scala.runtime.ScalaRunTime")
   def ScalaRuntimeModule(implicit ctx: Context) = ScalaRuntimeModuleRef.symbol
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -340,11 +340,11 @@ class Definitions {
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
     // The set of all wrap{X, Ref}Array methods, where X is a value type
-    lazy val Predef_wrapArrayR: Set[TermRef] = {
+    lazy val Predef_wrapArrayR: collection.Set[TermRef] = {
       val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
-      methodNames.map(ScalaPredefModule.requiredMethodRef).toSet
+      methodNames.map(ScalaPredefModule.requiredMethodRef)
     }
-    def Predef_wrapArray(implicit ctx: Context): Set[Symbol] = Predef_wrapArrayR.map(_.symbol)
+    val Predef_wrapArray = new PerRun[collection.Set[Symbol]](implicit ctx => Predef_wrapArrayR.map(_.symbol))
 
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/TreeGen.scala
```diff
@@ -23,4 +23,15 @@ object TreeGen {
       .select(wrapArrayMethodName(elemtp))
       .appliedToTypes(if (elemtp.isPrimitiveValueType) Nil else elemtp :: Nil)
       .appliedTo(tree)
+
+  /** Determines whether the given `tree` an application of `wrapRefArray` or `wrapXArray` */
+  def isRefOrPrimWrappedArray(tree: Tree, elemTpe: Type)(implicit ctx: Context): Boolean = {
+    tree match {
+      case Apply(fun, _) =>
+        val sname = fun.symbol.name
+        sname == nme.wrapRefArray || sname == nme.wrapXArray(elemTpe.classSymbol.name)
```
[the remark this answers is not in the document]
allanrenucci: You can try something like:
```scala
private def isWrapArrayMethodCall(tree: Apply)(implicit ctx: Context): Boolean = {
  val elemType = tree.tpe.elemType
  val wrapArrayMethodSym = defn.ScalaPredefModuleRef.requiredMethodRef(wrapArrayMethodName(elemTpe))
  tree.fun.symbol == wrapArrayMethodSym
}
```
Also I would move this method to `ElimRepeated`

compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala
```diff
@@ -106,6 +108,14 @@ class ElimRepeated extends MiniPhase with InfoTransformer { thisPhase =>
           // Because of phantomclasses, the Java array's type might not conform to the return type
   }
 
+  /** Determines whether the given `tree` represents a wrapped array */
+  private def isWrappedArray(tree: Apply)(implicit ctx: Context): Boolean = {
```
abeln: @allanrenucci That's much nicer! Thanks! PTAL

compiler/src/dotty/tools/dotc/transform/TreeGen.scala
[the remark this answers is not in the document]
smarter: I would store the result of `requiredMethodRef` in a val in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/Definitions.scala (see existing examples like https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/Definitions.scala#L339-L340)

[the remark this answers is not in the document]
smarter: Though here it's more complicated since there's one method per primitive type + Ref. You could store all of them in a Set in Definitions and then just check if the current symbol is contained in the Set.

compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala
abeln: Done.

compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala
```diff
@@ -187,4 +187,32 @@ class TestBCode extends DottyBytecodeTest {
...
+      val method = getMethod(moduleNode, "test")
+
+      val arrayWrapped = instructionsFromMethod(method).exists { instr: Instruction =>
+        instr match {
+          case inv: Invoke => inv.name.contains("wrapRefArray")
+          case _ => false
+        }
+      }
+
+      assert(!arrayWrapped, "Arrays should not be wrapped when passed to a Java varargs method\n")
```
abeln: I verified that this fails if I comment out the changes to ElimRepeated.

abeln:
Thanks for the pointer to the BC unit tests. Didn't know they existed.

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -338,6 +339,12 @@ class Definitions {
     def Predef_classOf(implicit ctx: Context) = Predef_classOfR.symbol
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
+    // The set of all wrap{X, Ref}Array methods, where X is a value type
```
abeln: @smarter pre-calculated the set of methods here, as suggested
PTAL

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -338,6 +339,12 @@ class Definitions {
...
+    lazy val Predef_wrapArrayR: Set[TermRef] = {
+      val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
+      methodNames.map(ScalaPredefModule.requiredMethodRef).toSet
+    }
+    def Predef_wrapArray(implicit ctx: Context): Set[Symbol] = Predef_wrapArrayR.map(_.symbol)
```
smarter: To avoid redoing the `.map(_.symbol)` every time this method is called it would make sense to cache its result. Symbols are valid at least for one run so you can use the `PerRun` class for this, grep for `PerRun` in this file for some examples.

abeln: Done.

abeln:
PTAL

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -338,6 +339,12 @@ class Definitions {
     def Predef_classOf(implicit ctx: Context) = Predef_classOfR.symbol
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
+    // The set of all wrap{X, Ref}Array methods, where X is a value type
+    lazy val Predef_wrapArrayR: collection.Set[TermRef] = {
+      val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
+      methodNames.map(ScalaPredefModule.requiredMethodRef)
+    }
```
allanrenucci: I would inline this in `Predef_wrapArray` since it is only used there:
```scala
val Predef_wrapArray = new PerRun[collection.Set[Symbol]] { implicit ctx =>
  val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
  methodNames.map(name => ScalaPredefModule.requiredMethodRef(name).toSymbol)
}
```

compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala
```diff
@@ -187,4 +187,32 @@ class TestBCode extends DottyBytecodeTest {
...
+        |  }
+        |}
+      """.stripMargin
+
+    checkBCode(source) { dir =>
+      val moduleIn = dir.lookupName("Test.class", directory = false)
+      val moduleNode = loadClassNode(moduleIn.input)
+      val method = getMethod(moduleNode, "test")
+
+      val arrayWrapped = instructionsFromMethod(method).exists { instr: Instruction =>
```
allanrenucci: You can do:
```scala
val arrayWrapped = instructionsFromMethod(method).exists {
  case inv: Invoke => inv.name.contains("wrapRefArray")
  case _ => false
}
```

abeln: Done.

abeln:
Made requested changes. @allanrenucci PTAL
Thanks for patiently shepherding this!

PR #3871, opened 2018-01-19
Fix classpath in QuoteDriver when executing from sbt

nicolasstucki:


nicolasstucki:
Found the bug and fixed it in https://github.com/nicolasstucki/quoted-util/commit/39087d4fd89c5aa588a0d3187dc3fc6f18860768

smarter:
I have no idea what the context is so I can't really review this. I'll just note that you can pass `-usejavacp` to the compiler so that the Scala classpath contains the Java classpath.

nicolasstucki:
Maybe the  `-usejavacp` solves this issue cleanly. I will try it out.

nicolasstucki:
The `-usejavacp` did not work, the issue is that in this case the java classpath does not contain the current classpath. The classpath only contains the SBT launcher which then uses a custom classloader to load the classpath that will be used to run the application.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -55,7 +56,15 @@ class QuoteDriver extends Driver {
 
   override def initCtx: Context = {
     val ictx = super.initCtx.fresh
-    val classpath = System.getProperty("java.class.path")
+    var classpath = System.getProperty("java.class.path")
+    this.getClass.getClassLoader match {
+      case cl: URLClassLoader =>
+        // Loads the classes load by this class loader
```
allanrenucci: the classes **loaded** by

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -55,7 +56,15 @@ class QuoteDriver extends Driver {
...
+        // When executing `run` or `test` in sbt the classpath is not in th property java.class.path
```
allanrenucci: in **the** property

nicolasstucki:
Rebased and fixed comments

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -61,7 +62,15 @@ class QuoteDriver extends Driver {
 
   override def initCtx: Context = {
     val ictx = super.initCtx.fresh
-    val classpath = System.getProperty("java.class.path")
```
smarter: What about using ctx.settings.classpath instead, isn't that enough?

nicolasstucki: No, the classpath of the library in not in there. That is why I needed to add it explicitly. We also need the current classpath in case the quote refers to something in the application (such as class or static method).

nicolasstucki: For example
```
class Foo
object Foo {
  def main(...) = {
    val q = '(new Foo)
    q.run // Foo need to be present in the classpath of the compiler that will compile q
  }
```

smarter:
I assume macros in scalac have the same problem. We should check how they handle classpaths.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
liufengyun: In scala.reflect macros, it assumes separation (ahead-of) compilation of macro definitions. The user-app must depend on the lib that contains `Foo` in order to use that macro, thus it's automatically on the path.

PR #3872, opened 2018-01-19
Upgrade to sbt 1

allanrenucci:
Rebased #3441. Reviews still need to be addressed

Unintentionally closed #3687


compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -139,23 +139,25 @@ class ExtractDependencies extends Phase {
       ctx.sbtCallback.binaryDependency(file, binaryClassName, fromClassName, sourceFile, context)
 
     def processExternalDependency(depFile: AbstractFile) = {
-      def binaryClassName(classSegments: List[String]) =
-            classSegments.mkString(".").stripSuffix(".class")
+
+      val binaryClassName = ctx.atPhase(ctx.flattenPhase) { implicit ctx =>
```
smarter: Was the previous mechanism used to find the binary class name incorrect in some situations?

compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -139,23 +139,25 @@ class ExtractDependencies extends Phase {
...
-          val packages = dep.to.ownersIterator
-            .filter(x => x.is(PackageClass) && !x.isEffectiveRoot).length
-            // We can recover the fully qualified name of a classfile from
-            // its path
-            val classSegments = pf.givenPath.segments.takeRight(packages + 1)
-            binaryDependency(pf.file, binaryClassName(classSegments))
+          val classFile =
+            if (dep.to.is(ModuleClass, butNot = Scala2x))
+              new File(pf.path.stripSuffix(".class") + "$.class")
+                // Module Class loaded from TASTY are loaded from their companion class file.
```
smarter: I don't understand what's going on here. Do you have a testcase where this is needed? Does it do the right thing for inner objects too?

compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -274,24 +270,32 @@ private class ExtractDependenciesCollector extends tpd.TreeTraverser { thisTreeT
...
-    if (fromClass ne NoSymbol) {
+    if (fromClass.exists) {
       assert(fromClass.isClass)
       addUsedName(fromClass, name, scope)
     }
   }
 
+  /** Mangle a JVM symbol name in a format better suited for internal uses by sbt. */
+  private def mangledName(sym: Symbol)(implicit ctx: Context): Name = {
+    def constructorName = sym.enclosingClass.fullName ++ ";init;"
```
allanrenucci: @smarter I didn't use the exact same scheme as in sbt/zinc#288. Let me know if you think it is an issue

smarter: I remember weird things happening in zinc when I used "." in the naming scheme, that's why I switched to ";". It would be prudent to do the same here.

smarter: (I don't remember what the weird things where sadly)

liufengyun:
Benchmarks work without problem 👍

PR #3880, opened 2018-01-21
 Fix #3876: Implement Expr.AsFunction

### LINKED ISSUE #3876: Implement Expr.AsFunction
Implementing `Expr.AsFunction` will require a new type of expression that is special-cased in the compiler to implement the optimization described in the documentation.

* [Expr.AsFunction](https://github.com/lampepfl/dotty/blob/master/library/src/scala/quoted/Expr.scala#L15-L17)
* [Documentation](https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/symmetric-meta-programming.md#from-exprs-to-functions-and-back)

nicolasstucki:


compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -111,4 +107,42 @@ object PickledQuotes {
...
+      case ClassTag.Double => defn.FloatType
+    }
+    TypeTree(tpe)
+  }
+
+  private def functionAppliedTo(f: Tree, x: Tree)(implicit ctx: Context): Tree = {
+    val x1 = SyntheticValDef(NameKinds.UniqueName.fresh("x".toTermName), x)
+    def x1Ref() = ref(x1.symbol)
+    def rec(f: Tree): Tree = f match {
+      case Block((ddef: DefDef) :: Nil, _: Closure) =>
```
odersky: Use the `closure` extractor instead (in TreeInfo).

nicolasstucki: `closure` only extracts the `Closure` and here we also need its `DefDef`. I will instead factor out a common extractor from `closureBody` and this me method.

compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
```diff
@@ -534,9 +534,18 @@ trait TypedTreeInfo extends TreeInfo[Type] { self: Trees.Instance[Type] =>
     }
   }
 
+  /** An extractor for closures with their def contained in a block. */
```
odersky: I'd rename it to `closureDef` and return just the `def`. The closure part is used nowhere.

PR #3883, opened 2018-01-22
 Seal quoted.Expr and quoted.Type

nicolasstucki:


nicolasstucki:
Will need to rebase after #3880 is merged.

PR #3887, opened 2018-01-22
Implement AppliedTermRef (singleton types for term-level applications)

gsps:
This PR implements `AppliedTermRef`, a singleton type representing a term-level application precisely. `AppliedTermRef`s are currently only introduced when they are stable, that is, when both the function and the arguments are.

The PR includes several additional changes to allow interaction and testing of `AppliedTermRef`s:

- Previously no methods were marked stable, making it impossible to take advantage of the new singletons. As a starting point, we now mark methods on primitive value classes stable. Until an effect system capable of proving purity has landed, it might make sense to allow users to assume purity for user-defined methods (e.g. using an annotation).
- Similarly to other precise types `AppliedTermRef`s are widened away unless specifically expected on declaration sites. This prompts the need for an explicit syntax of singleton types more complex than `p.a.t.h.type`. For this purpose we add an alternative syntax for singleton types: a type tree of the form `{ expr }` is translated to the singleton type `T` of `expr`, provided that `T` is stable.

A simple example:
```scala
val x = 1
val y: {x} = x
implicitly[{x + 1} =:= {y + 1}]  // works
```

To Dos:

- [x] ~~Bug: Missing denotation / NoType upon accessing underlying type of inter-parameter dependency.~~ _Deferred. I suggest tackling this as a separate issue -- I think the problem lies in how parameter types are integrated into `MethodType`s and existed before this PR, but afaict could not be exercised previously._
- [ ] Implement translation of `AppliedTermRef` for sbt API extraction. _Not sure how to proceed with this. Can we map `AppliedTermRef`s to existing `xsbti.api.Tree`s, or do we need a new subclass?_

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -699,6 +699,15 @@ class TypeComparer(initctx: Context) extends ConstraintHandling[AbsentContext] w
             false
         }
         compareTypeBounds
+      case tp2: AppliedTermRef =>
+        // TODO(gsps): Check whether rule or position thereof should change
+        def compareAppliedTerm = tp1 match {
+          case tp1: AppliedTermRef =>
+            sameLength(tp1.args, tp2.args) && isSubType(tp1.fn, tp2.fn) &&
+              tp1.args.zip(tp2.args).forall((arg1, arg2) => isSubType(arg1, arg2))
+          case _ => fourthTry
+        }
+        compareAppliedTerm
       case tp2: AnnotatedType if tp2.isRefining =>
         (tp1.derivesAnnotWith(tp2.annot.sameAnnotation) || defn.isBottomType(tp1)) &&
         recur(tp1, tp2.parent)
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1044,9 +1044,11 @@ object Types {
       case _ => this
     }
 
-    /** Strip PolyType prefixes */
+    /** Strip PolyType and AppliedTermRef prefixes */
+    // TODO(gsps): Rename this to also reflect the removal of AppliedTermRefs
     def stripPoly(implicit ctx: Context): Type = this match {
       case tp: PolyType => tp.resType.stripPoly
+      case tp: AppliedTermRef => tp.resType.stripPoly
       case _ => this
     }
 
@@ -2606,6 +2608,56 @@ object Types {
     override def toString: String = s"LazyRef(${if (computed) myRef else "..."})"
     override def equals(other: Any): Boolean = this.eq(other.asInstanceOf[AnyRef])
     override def hashCode: Int = System.identityHashCode(this)
+  }
+
+  // --- AppliedTermRef -------------------------------------------------------
+
+  /** A precise representation of a term-level application `fn(... args)`. **/
+  abstract case class AppliedTermRef(fn: /*TermRef | AppliedTermRef*/ SingletonType, args: List[Type])
+    extends CachedProxyType with SingletonType
+  {
+    private[this] var myResType: Type = _
+    def resType(implicit ctx: Context): Type = {
+      if (myResType == null)
+        fn.widen match {
+          case methTpe: MethodType => myResType = ctx.typer.applicationResultType(methTpe, args)
+        }
+      myResType
+    }
+
+    def underlying(implicit ctx: Context): Type = resType
+
+    def derivedAppliedTermRef(fn: Type, args: List[Type])(implicit ctx: Context): Type =
+      if ((this.fn eq fn) && (this.args eq args)) this
+      else AppliedTermRef(fn, args)
+
+    override def computeHash(bs: Binders) = doHash(bs, fn, args)
+    override def hashIsStable: Boolean = fn.hashIsStable && args.forall(_.hashIsStable)
+
+    override def eql(that: Type) = that match {
+      case that: AppliedTermRef => (this.fn eq that.fn) && this.args.eqElements(that.args)
+      case _ => false
+    }
+
+    // TODO(gsps): AppliedTermRef#iso?
+  }
+
+  final class CachedAppliedTermRef(fn: SingletonType, args: List[Type]) extends AppliedTermRef(fn, args)
+
+  object AppliedTermRef {
+    def apply(fn: Type, args: List[Type])(implicit ctx: Context): Type = {
+      assertUnerased()
+      fn.dealias match {
+        case fn: TermRef => unique(new CachedAppliedTermRef(fn, args))
+        case fn: AppliedTermRef => unique(new CachedAppliedTermRef(fn, args))
+        case _ =>
+          fn.widenDealias match {
+            case methTpe: MethodType => ctx.typer.applicationResultType(methTpe, args)
+            case _: WildcardType => WildcardType
+            case tp => throw new AssertionError(i"Don't know how to apply $tp.")
+          }
+      }
+    }
   }
 
   // --- Refined Type and RecType ------------------------------------------------
@@ -4800,6 +4852,8 @@ object Types {
       tp.derivedSuperType(thistp, supertp)
     protected def derivedAppliedType(tp: AppliedType, tycon: Type, args: List[Type]): Type =
       tp.derivedAppliedType(tycon, args)
+    protected def derivedAppliedTermRef(tp: AppliedTermRef, fn: Type, args: List[Type]): Type =
+      tp.derivedAppliedTermRef(fn, args)
     protected def derivedAndType(tp: AndType, tp1: Type, tp2: Type): Type =
       tp.derivedAndType(tp1, tp2)
     protected def derivedOrType(tp: OrType, tp1: Type, tp2: Type): Type =
@@ -4858,6 +4912,9 @@ object Types {
               nil
           }
           derivedAppliedType(tp, this(tp.tycon), mapArgs(tp.args, tp.tyconTypeParams))
+
+        case tp: AppliedTermRef =>
+          derivedAppliedTermRef(tp, this(tp.fn), tp.args.mapConserve(this))
 
         case tp: RefinedType =>
           derivedRefinedType(tp, this(tp.parent), this(tp.refinedInfo))
@@ -5173,6 +5230,26 @@ object Types {
           else tp.derivedAppliedType(tycon, args)
       }
 
+    // TODO(gsps): Double-check for changes in similar derivations
+    override protected def derivedAppliedTermRef(tp: AppliedTermRef, fn: Type, args: List[Type]): Type =
+      fn match {
+        case Range(fnLo, fnHi) =>
+          range(derivedAppliedTermRef(tp, fnLo, args), derivedAppliedTermRef(tp, fnHi, args))
+        case _ =>
+          if (fn.isBottomType) {
+            fn
+          } else if (args.exists(isRange)) {
+            val loBuf, hiBuf = new mutable.ListBuffer[Type]
+            args foreach {
+              case Range(lo, hi) => loBuf += lo; hiBuf += hi
+              case arg => loBuf += arg; hiBuf += arg
+            }
+            range(tp.derivedAppliedTermRef(fn, loBuf.toList), tp.derivedAppliedTermRef(fn, hiBuf.toList))
+          } else {
+            tp.derivedAppliedTermRef(fn, args)
+          }
+      }
+
     override protected def derivedAndType(tp: AndType, tp1: Type, tp2: Type): Type =
       if (isRange(tp1) || isRange(tp2)) range(lower(tp1) & lower(tp2), upper(tp1) & upper(tp2))
       else tp.derivedAndType(tp1, tp2)
@@ -5274,6 +5351,9 @@ object Types {
             foldArgs(acc, tparams.tail, args.tail)
           }
         foldArgs(this(x, tycon), tp.tyconTypeParams, args)
+
+      case tp: AppliedTermRef =>
+        foldOver(this(x, tp.fn), tp.args)
 
       case _: BoundType | _: ThisType => x
 
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2627,9 +2627,12 @@ object Types {
 
     def underlying(implicit ctx: Context): Type = resType
 
+    /** Compute the derived AppliedTermRef, widening to the result type if any of its
+     *  components is unstable.
+     */
     def derivedAppliedTermRef(fn: Type, args: List[Type])(implicit ctx: Context): Type =
       if ((this.fn eq fn) && (this.args eq args)) this
-      else AppliedTermRef(fn, args)
+      else AppliedTermRef.make(fn, args)
 
     override def computeHash(bs: Binders) = doHash(bs, fn, args)
     override def hashIsStable: Boolean = fn.hashIsStable && args.forall(_.hashIsStable)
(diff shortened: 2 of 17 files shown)


compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -861,6 +861,11 @@ object Parsers {
       }
     }
 
+    def emptyRefinementOrSingletonExpr(): Tree = {
```
gsps: Maybe call this `refinementOfEmptyOrSingletonExpr()`.

gsps:
Rebased on top of master, but the most recently merged PR (#3841) seems to interfere, as now the fromtasty version of the added test fails. Investigating.

Blaisorblade:
Wow! Isn’t this supporting what OCaml calls *applicative functors*? That would be exciting news, I can elaborate why.

Do you also support type members of such applications? That would be a logical consequence.

To elaborate a bit (not sure if this will be clear): this would enable usable type members on implicits. Without such a feature, two separate occurrences of `implicitly[Ord[(A, B)].FooTypeMember` (assuming implicits for`oa: Ord[A]`, `ob: Ord[B]`, and `implicit def ordPair[A: Ord, B: Ord]: Ord[(A, B)]`) would not be compatible types, because they’re ultimately `ordPair(oa, ob).FooTypeMember`. (You’d also need a stricter type for `implicitly`, `def implicitly[T](t: T): t.type = t`, but this should work). If `ordPair` is marked as pure/stable, separate occurrences of `implicitly[Ord[(A, B)].FooTypeMember` would become type compatible.

For proper credit, this isn’t my idea: it goes back to a talk by Derek Dreyer in 2009, and I and Tillmann Rendel discussed applications to Scala a few years ago.

gsps:
Indeed, this PR should enable applicative functors as in OCaml (though I have no experience with OCaml in that regard). I haven't written any test cases involving implicit search, specifically, but in principle that should work too. Feel free to play with the branch and let me know if it doesn't behave as you'd expect!

Blaisorblade:
I should definitely do that! It’ll take me a while to get productive and be able to do this though — I’m already commenting on more issues I can look at seriously!

gsps:
test performance please

gsps:
@liufengyun When I try to access the performance test results at http://dotty-bench.epfl.ch/3887/ I get a 404, any idea why?

Blaisorblade:
@gsps That link works here now.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -561,6 +561,17 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
             false
         }
         compareTypeBounds
+      case tp2: AppliedTermRef =>
+        def compareAppliedTerm = tp1 match {
+          case _ if !ctx.phase.isTyper =>
+            // Need to be more permissive when checking later phases, applications may have been rewritten
+            isSubType(tp1, tp2.resType)
+          case tp1: AppliedTermRef =>
+            tp1.args.size == tp2.args.size && isSubType(tp1.fn, tp2.fn) &&
```
smarter: `sameLength(tp1.args, tp2.args)`

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -561,6 +561,17 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
             false
         }
         compareTypeBounds
+      case tp2: AppliedTermRef =>
+        def compareAppliedTerm = tp1 match {
+          case _ if !ctx.phase.isTyper =>
+            // Need to be more permissive when checking later phases, applications may have been rewritten
```
smarter: Can you give an example?

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -561,6 +561,17 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
+            isSubType(tp1, tp2.resType)
+          case tp1: AppliedTermRef =>
+            tp1.args.size == tp2.args.size && isSubType(tp1.fn, tp2.fn) &&
+              (tp1.args zip tp2.args).forall(t => isSubType(t._1, t._2))
```
smarter: Slightly cleaner: `(tp1.args, tp2.args).zipped.map(isSubType)`

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -561,6 +561,17 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
             false
         }
         compareTypeBounds
+      case tp2: AppliedTermRef =>
```
smarter: Any reason to have this in thirdTry and not firstTry? Also I think it'd be good to have more testcases that stress-test the subtype checks for AppliedTermRef. For example, what happens when an AppliedTermRef is hidden in a type alias?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -148,6 +148,7 @@ object Types {
     /** Does this type denote a stable reference (i.e. singleton type)? */
     final def isStable(implicit ctx: Context): Boolean = stripTypeVar match {
       case tp: TermRef => tp.termSymbol.isStable && tp.prefix.isStable || tp.info.isStable
+      case tp: AppliedTermRef => tp.fn.isStable && tp.args.forall(_.isStable)
```
smarter: Aren't all AppliedTermRefs stable by definition?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2169,6 +2171,53 @@ object Types {
       apply(prefix, if (denot.symbol.exists) denot.symbol.asType else name).withDenot(denot)
   }
 
+  // --- AppliedTermRef ---------------------------------------------------------------------
+
+  /** A precise representation of a term-level application `fn(... args)`. **/
+  abstract case class AppliedTermRef(fn: /*TermRef | AppliedTermRef*/ SingletonType, args: List[Type])
+    extends CachedProxyType with SingletonType
+  {
+    protected[this] var myResType: Type = _
```
smarter: Could be `private[this]`

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3912,6 +3961,8 @@ object Types {
       tp.derivedSuperType(thistp, supertp)
     protected def derivedAppliedType(tp: AppliedType, tycon: Type, args: List[Type]): Type =
       tp.derivedAppliedType(tycon, args)
+    protected def derivedAppliedTerm(tp: AppliedTermRef, fn: Type, args: List[Type]): Type =
```
smarter: should be `derivedAppliedTermRef` for consistency (or rename `AppliedTermRef` to `AppliedTerm` maybe)

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2169,6 +2171,53 @@ object Types {
...
+
+  object AppliedTermRef {
+    def apply(fn: Type, args: List[Type])(implicit ctx: Context): Type = {
+      assertUnerased()
+      fn.dealias match {
+        case fn: TermRef => unique(new CachedAppliedTermRef(fn, args))
+        case fn: AppliedTermRef => unique(new CachedAppliedTermRef(fn, args))
+        case _ =>
+          fn.widenDealias match {
+            case methTpe: MethodType => ctx.typer.applicationResultType(methTpe, args)
```
smarter: It's kind of weird that `AppliedTermRef#apply` can produce things which are not `AppliedTermRef`, when does this happen?

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -416,6 +416,7 @@ object TastyFormat {
     METHODtype + implicitOffset + erasedOffset
   }
 
+  final val APPLIEDTERMREF = 180
```
smarter: Need to add this to the comment listing all tags at the top of this file and to bump the MinorVersion of TASTY. Also may need to add a case to  TastyPrinter (To check if TastyPrinter is working correctly, enable "pickling" in Printers.scala)

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -863,7 +863,7 @@ object Parsers {
           makeTupleOrParens(inParens(argTypes(namedOK = false, wildOK = true)))
         }
       else if (in.token == LBRACE)
-        atPos(in.offset) { RefinedTypeTree(EmptyTree, refinement()) }
+        atPos(in.offset) { inBraces(emptyRefinementOrSingletonExpr()) }
```
smarter: If you're changing the grammar, it'd be nice to update the grammar comment above (and reflect the update in https://github.com/lampepfl/dotty/blob/master/docs/docs/internals/syntax.md)

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -60,6 +60,8 @@ class PlainPrinter(_ctx: Context) extends Printer {
           homogenize(tp.ref)
         case AppliedType(tycon, args) =>
           tycon.dealias.appliedTo(args)
+        case tp: AppliedTermRef =>
+          homogenize(tp.underlying)
```
smarter: That seems pretty drastic to me, it means that AppliedTermRef will always be widened before being pretty-printed with -Ytest-pickler, so you may miss differences.

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -141,6 +143,8 @@ class PlainPrinter(_ctx: Context) extends Printer {
         ParamRefNameString(tp) ~ ".type"
       case tp: TypeParamRef =>
         ParamRefNameString(tp) ~ lambdaHash(tp.binder)
+      case tp: AppliedTermRef =>
+        toTextRef(tp) ~ ".type"
```
smarter: shouldn't you wrap in `{` and `}` to reflect the syntax instead?

compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
```diff
@@ -445,6 +446,9 @@ private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder
         val apiTycon = apiType(tycon)
         val apiArgs = args.map(processArg)
         api.Parameterized.of(apiTycon, apiArgs.toArray)
+      case tp: AppliedTermRef =>
+        val apiTps = (tp.fn :: tp.args).map(apiType)
+        withMarker(combineApiTypes(apiTps: _*), appliedTermRefMarker)
```
smarter: Could you add an sbt scripted test to check that this is working properly? See https://github.com/lampepfl/dotty/tree/master/sbt-dotty/sbt-test/source-dependencies/by-name for an example.

compiler/src/dotty/tools/dotc/core/Types.scala
Blaisorblade: I'd think they can't be stable if the arguments aren't stable; OTOH, I suspect `AppliedTermRef` with unstable arguments shouldn't be just unstable but illegal.

smarter: That's what I meant.

gsps: In principle `f(unstableThing)` is still meaningful, if you think of it as the image of `f` on set `unstableThing`. E.g., we could still type `sign(impure()): 1.type` where `def impure(): PositiveInt`. That being said, I think it's fine to just make `AppliedTermRef`s stable by construction for now, and I'll come back to this in my branch if I find a compelling use-case.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
gsps: We do have a test that relies on a type application:
```scala
val t: Id[{a + 1}] = a + 1
```
I'll also add one that relies on a simple alias.

compiler/src/dotty/tools/dotc/core/Types.scala
gsps: I think for the purposes of this PR it's only in `derivedAppliedTermRef` for the case when the `fn` part was widened, so that we need to widen the ATR altogether (In the ptyper branch there are additional uses).

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
gsps: Indeed, though it does require some state to track when we hit the first `AppliedTermRef`. Is it okay to keep this kind of flag in `PlainPrinter` itself?

smarter: I guess so, we already have state for rectypes.

compiler/src/dotty/tools/dotc/core/Types.scala
smarter: But is there any other type node in dotty where the #apply can return a different type?

gsps: True, I guess #apply should take the dealiased `fnTpe` check that it's an `(Applied)?TermRef` and just return the cached instance. The rest can go into a separate factory method (or be inlined into derivedAppliedTermRef, for now).

LPTK:
This is really exciting for the reason @Blaisorblade mentioned. It could help solve a long-lasting problem in Scala (and other languages with type classes but no dependent types).

I've just played a little bit with it, but my experimentation was hampered because of this:

```scala
scala> class Ord[A]
scala> implicit object io extends Ord[Int]
scala> implicit def lo[A](implicit ord: Ord[A]): Ord[List[A]] = new Ord
scala> val r: {lo(io)} = lo(io)
1 |val r: {lo(io)} = lo(io)
  |       ^^^^^^^^
  |       Ord[List[A]] is not stable
  |
  |       where:    A is a type variable which is an alias of Int
```

Is there currently a way (annotation) to tell Dotty that the `lo` function is stable, or this is yet to be added?

Also, it would be nice and generally more consistent if we could write type `f(x).T` instead of having to do `{f(x)}#T`.

gsps:
There isn't in this branch, but the ptyper one has it, i.e., `scala.annotation.assumePure`. If you're feeling brave, https://github.com/gsps/dotty/tree/implement-ptype ;-) I noticed just recently that the REPL is somewhat broken, but I'm in the middle of the somewhat major change that is moving to dotty-compiled inox, so that'll have to wait a bit.  

I agree on the syntax, and I'm not sure what would be the preferred notation here. It's a pity that in general we can't just use term-level syntax (without the curly braces) to denote singleton types, since the type and the term namespaces might clash.

LPTK:
> I agree on the syntax, and I'm not sure what would be the preferred notation here.

The thing is what comes before `.T` should be a stable term even in current Scala. So it makes sense to allow `f(x).T` as meaning `{f(x)}#T`. Syntax `{f(x)}.T` wouldn't be consistent.

Blaisorblade:
> It's a pity that in general we can't just use term-level syntax (without the curly braces) to denote singleton types, since the type and the term namespaces might clash.

We have a few exceptions, but they can be tricky to handle — `Int => 1` is valid syntax, but it's hard for Dotty to always tell it's in fact a type and to not crash on `class Foo extends Int => 1` (see https://github.com/lampepfl/dotty/pull/4487).

odersky:
I believe if we want to revive this we need a decision on syntax, and a clear goal. For me the goal would be to implement applicative functors. But that should be fleshed out in detail first before we make a move on the underlying type structure. Otherwise we just increase entropy.

anatoliykmetyuk:
There was no activity on this one for a long while, so let's close it.

PR #3888, opened 2018-01-23
Refactor `inhabited` in exhaustivness check

liufengyun:
Previously there is both `implementability` and `inhabited`. The refactoring merges the two and simplify the logic.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -510,10 +510,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     // convert top-level type shape into "conjunctive normal form"
     def cnf(tp: Type): Type = tp match {
       case AndType(OrType(l, r), tp)      =>
-        val tp1 = cnf(tp)
-        val l1  = cnf(l)
-        val r1  = cnf(r)
-        OrType(cnf(AndType(l1, tp1)), cnf(AndType(r1, tp1)))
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -574,21 +499,97 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-
-    !emptySingletonIntersection(false, tpe)
+  def inhabited(tp: Type)(implicit ctx: Context): Boolean = {
+    // convert top-level type shape into "conjunctive normal form"
+    def cnf(tp: Type): Type = tp match {
+      case AndType(OrType(l, r), tp)      =>
+        val tp1 = cnf(tp)
+        val l1  = cnf(l)
+        val r1  = cnf(r)
+        OrType(cnf(AndType(l1, tp1)), cnf(AndType(r1, tp1)))
```
OlivierBlanvillain: I think you don't need to compute cnf(l), cnf(r) and cnf(tp) here, it's already handled by the recursion, the body of this case could be changed to the following:

```scala
OrType(cnf(AndType(l, tp)), cnf(AndType(r, tp)))
```

liufengyun: You are right, good catch.

liufengyun:
@OlivierBlanvillain The motivation for CNF is to ease employing the facts of "single inheritance" and "final class". For checks that employ the fact about "singleton types", CNF is not required.

PR #3889, opened 2018-01-23
Fix #2675: Properly handle undetermined type variables in unapply

### LINKED ISSUE #2675: pattern matching generic case classes might be too strict
Consider the following example:

~~~scala
class Ref[-T]

sealed trait Op
case class Send[T](ref: Ref[T], msg: T) extends Op

object Main {
  val ref: Ref[String] = ???
  val op: Op = Send(ref, "hello")
  op match {
    case Send(ref, msg) => 
  }
}
~~~

Compiling this with dotty 0.1.2-RC1 yields:

~~~
[warn] -- Warning: /Users/rkuhn/comp/test/dotty/dotty-project-template/src/main/scala/Main.scala:10:13 
[warn] 10 |    case Send(target, msg) => 
[warn]    |         ^^^^^^^^^^^^^^^^^
[warn]    |         There is no best instantiation of pattern type Send[Any^]
[warn]    |         that makes it a subtype of selector type Op.
[warn]    |         Non-variant type variable T cannot be uniquely instantiated.
[warn]    |         (This would be an error under strict mode)
[warn] one warning found
~~~

The important knowledge encapsulated by the `Send` type is that it has been constructed with ref and msg matching each other. When pattern matching on such a type, I don’t care what the type parameter was, I only care that it was the same for the ref and the msg. Why can’t dotty just make up a type parameter that is not bound and use that in the case statement?

odersky:
In non-variant contexts, we used to always maximize undetermined type variables, but this is unsound. See neg/patternUnsoundness.scala. We now create fresh abstract types in
these situations. This forced us to also fix the pickling and unpickling of Bind nodes.


tests/pickling/i0290-type-bind.scala
```diff
@@ -0,0 +1,59 @@
...
+
+object bar3 { def f0 = {
+
+  class C[T <: Seq[_]]
+
+  val x: AnyRef = new C
+
+  x match {
+    case x: C[u] =>
+      def x: u = x
```
smarter: Is the self-recursive call intended here?

odersky: I assume so, it was an old test, just with more variations that exercise forward binds in pickling.

PR #3893, opened 2018-01-23
Fix dotty-doc exit code

allanrenucci:


allanrenucci:
@liufengyun Can you have a look again. I took a different approach to share more code with `Driver.scala`. I now override `process` instead of `main`.

doc-tool/src/dotty/tools/dottydoc/DocDriver.scala
```diff
@@ -33,16 +32,12 @@ class DocDriver extends Driver {
 
   override def newCompiler(implicit ctx: Context): Compiler = new DocCompiler
 
-  def compiledDocs(args: Array[String]): collection.Map[String, Package] = {
```
allanrenucci: I removed this method since I found no reference for it

liufengyun: Nice refactoring! LGTM

PR #3894, opened 2018-01-23
Implement f"" with quotes

nicolasstucki:
Todo list

- [x] support basic feature to print
- [ ] type checking
- [ ] error message & positions
- [x] port tests from scalac

Requirements

- error reporting facility
- get position of trees
- get type of tree
- subtype checking
- infer implicit

Problem Found

- cannot select in macro body (`sc.parts`), need to move to implementation
- cannot interpreter correctly with explicit call `new Interpolation.Helper(sc)`.
- `Seq` works, but not `List` in the API
- doesn't support `f"${x + 1}%d"`: phase error

### DIFF
--- a/compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
+++ b/compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
@@ -102,23 +102,6 @@ class Interpreter(implicit ctx: Context) {
       case tree: Ident if env.contains(tree.symbol) =>
         env(tree.symbol)
 
-      case tree: Apply =>
-        def prefixAndArgs(t: Tree, allArgs: List[Tree]): (Tree, List[Tree]) = t match {
-          case Apply(qual, args2) => prefixAndArgs(qual, args2 ::: allArgs)
-          case TypeApply(qual, _) => prefixAndArgs(qual, allArgs)
-          case Select(qual, _) => (qual, allArgs)
-        }
-        val (prefix, args) = prefixAndArgs(tree, Nil)
-
-        val evaluatedPrefix = interpretTreeImpl(prefix, env)
-
-        val clazz = evaluatedPrefix.getClass
-        val paramClasses = paramsSig(tree.symbol)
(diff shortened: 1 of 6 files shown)


smarter:
If this ends up implementing the full f"", we'll be able to close https://github.com/lampepfl/dotty/issues/687 :)

compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -138,7 +138,7 @@ class Interpreter(implicit ctx: Context) {
         env(qualifier.symbol)
 
       case SeqLiteral(elems, _) =>
-        Seq(elems.map(elem => interpretTreeImpl(elem, env)))
+        elems.map(elem => interpretTreeImpl(elem, env))
```
liufengyun: 👍

nicolasstucki: Failed tests because the check file was not updated, but this bug is fixed :)

liufengyun:
Looking closely at the commented code that doesn't compile, it has two problems:

1. When it's safe to interpret code (`Splicer.splice`)? The current implementation (`level == 0`) seems not to be the correct condition. It boils down to a theoretical question: when it's safe to run the code of `Expr[T]`?

2. There is some bug related to level checking for inline bindings. Need to investigate more.

```Scala
val x = 5
println(ff"$x%d"              // works
println(ff"${x + 1}%d")   // doesn't work
```

nicolasstucki:
Inline binding might be lifting computations from within a quoted inlined tree which can break phase consistency. This is what is happening with the varargs.

nicolasstucki:
Will revisit this PR when we have proper extension methods.

liufengyun:
@nicolasstucki Should we revive this PR?

nicolasstucki:
What a coincidence! A student has just started today a semester project on this. https://github.com/alemannosara/f-interpolators-in-Dotty-macros

PR #3897, opened 2018-01-23
Fix #2681: re-enable check for potentially overridden classfiles

### LINKED ISSUE #2681: Figure out why overwriting protection doesn't work.
I recall dotty having protection from overwriting a user-defined class with a synthetic class with the same name by name-mangling sythetic one:
https://github.com/lampepfl/dotty/pull/313#commitcomment-9050025

https://github.com/lampepfl/dotty/issues/2673#issuecomment-306032190 reports that it doesn't work in some cases.

biboudis:


smarter:
I'm not sure it's worth adding a mini-phase for this. There is already code in the backend to do this kind of check that just needs to be enabled I think: https://github.com/lampepfl/dotty/blob/3004af37018d2f64c08f33500becf939141ec2e8/compiler/src/dotty/tools/backend/jvm/GenBCode.scala#L190-L202 see https://github.com/lampepfl/dotty/issues/1077

odersky:
The problem with the miniphase approach is that it catches only files declared in the same sourcefile. A backend solution is better because it is more complete, IMO.

compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -188,18 +188,14 @@ class GenBCodePipeline(val entryPoints: List[Symbol], val int: DottyBackendInter
...
           case None =>
             caseInsensitively.put(lowercaseJavaClassName, claszSymbol)
           case Some(dupClassSym) =>
-            reporter.warning(
-              claszSymbol.pos,
-              s"Class ${claszSymbol.javaClassName} differs only in case from ${dupClassSym.javaClassName}. " +
-                "Such classes will overwrite one another on case-insensitive filesystems."
-            )
-        }*/
+            ctx.warning(s"Class ${claszSymbol.name.toString.toLowerCase} differs only in case from ${dupClassSym.name.toString}. " +
```
allanrenucci: I would keep the old error message, if you convert to lower case, the user won't see the differences.

tests/neg/i2673.scala
```diff
@@ -0,0 +1,4 @@
+package Foos //error
```
allanrenucci: This should fo into `neg-custom-args` and be compiled with `-Xfatal-warnings`. Here is an example https://github.com/lampepfl/dotty/blob/78bd3dd8ce5822171ccf252864e2e5c3d76747a4/compiler/test/dotty/tools/dotc/CompilationTests.scala#L197

smarter:
I have too many things to review.

biboudis:
If this is merged #1077 can be closed as well.

odersky:
LGTM

PR #3902, opened 2018-01-24
Add tests files from benchmarks

nicolasstucki:


allanrenucci:
@nicolasstucki Can you move this test to pending, it is causing spurious failure in the CI

PR #3903, opened 2018-01-24
Fix #2663: More refined handling of enum case apply results

### LINKED ISSUE #2663: Type parameter inference broken for parent of an enum case
This works:
```scala
class Foo[T](x: T)
class Bar[T](y: T) extends Foo(y)
```
But this doesn't:
```scala
enum Foo[T](x: T) {
  case Bar[T](y: T) extends Foo(y)
}
```
```scala
-- [E055] Syntax Error: try/parents.scala:2:28 ---------------------------------
2 |  case Bar[T](y: T) extends Foo(y)
  |                            ^^^
  |                            missing type parameter for [Foo$$T] => Foo[T]
```
Manually adding the missing type parameter works:
```scala
enum Foo[T](x: T) {
  case Bar[T](y: T) extends Foo[T](y)
}
```

odersky:
i2663 demonstrates that we do not always have a complete result type
for the `apply` method of an enum case. In that case our only recourse
is to add a generic widening method that upcasts the enum case class to the
enum base class.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -456,19 +456,14 @@ object desugar {
     // For all other classes, the parent is AnyRef.
     val companions =
       if (isCaseClass) {
-        def extractType(t: Tree): Tree = t match {
-          case Apply(t1, _) => extractType(t1)
-          case TypeApply(t1, ts) => AppliedTypeTree(extractType(t1), ts)
-          case Select(t1, nme.CONSTRUCTOR) => extractType(t1)
-          case New(t1) => t1
-          case t1 => t1
-        }
         // The return type of the `apply` method
-        val applyResultTpt =
-          if (isEnumCase)
-            if (parents.isEmpty) enumClassTypeRef
-            else parents.map(extractType).reduceLeft(AndTypeTree)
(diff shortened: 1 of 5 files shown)


liufengyun:
It doesn't work for the following case:

```Scala
enum Foo[T](x: T) {
  case Bar[S, T](y: T) extends Foo(y)
}
```

While the following compiles without no problem:

```Scala
class Foo[T](x: T)
case class Bar[S, T](y: T) extends Foo(y)
```

odersky:
@liufengyun Excellent catch! This turned out to be an even bigger can of worms than I thought.

liufengyun:
The fix looks good, but I found another related issue about the return type of `apply`:

```Scala
enum Foo[T](x: T) {
  case Bar[S, T](y: T) extends Foo[y.type](y)
}
```

odersky:
That's devious (or clever, depending how you look at it)! It seems to fix this properly we need #3920 to be solved, and it is uncertain whether we will ever get there.

The only quick fix I see is to disallow parameter references in extension clauses of enum cases.

odersky:
Or we leave it as is, and just accept the fact that the enum expansion is not type correct in some cases.

odersky:
I filed a separate issue #3935 to keep track of this.

PR #3905, opened 2018-01-24
Fix #2104: Instantiate unapply result type variables

### LINKED ISSUE #2104: Type inference on nested GADTs
Type inference fails to infer `i: Int` in the following pattern matching:

```scala
case class Pair[A, B](_1: A, _2: B)

trait Cons[+H, +T]

object Cons {
  def apply[H, T](h: H, t: T): Cons[H, T] = ???
  def unapply[H, T](t: Cons[H, T]): Option[Pair[H, T]] = ???
}

object Test {
    Cons(Option(1), None) match {
      case Cons(Some(i), None) =>
        i: Int // error: found: Any(i), requires: Int
        assert(i == 1)
    }
  }
}
```

-Xprint:frontend shows a union type in the unapply:

```scala
case Cons.unapply[Option[Int] | Some[Any], None$](
  Some.unapply[Any](i @ _): Some[Any], None
)
```

odersky:
Type variables in an unapply result should be fully instantiated before nested
pattern matches. This is analogous to instantiating the scrutinee type of a match.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -920,6 +920,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
         val ownType =
           if (selType <:< unapplyArgType) {
             unapp.println(i"case 1 $unapplyArgType ${ctx.typerState.constraint}")
+            fullyDefinedType(unapplyArgType, "pattern selector", tree.pos)
             selType
           } else if (isSubTypeOfParent(unapplyArgType, selType)(ctx.addMode(Mode.GADTflexible))) {
             maximizeType(unapplyArgType) match {
@@ -952,7 +953,6 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
               ex"Pattern type $unapplyArgType is neither a subtype nor a supertype of selector type $selType",
               tree.pos)
           }
-
         val dummyArg = dummyTreeOfType(ownType)
         val unapplyApp = typedExpr(untpd.TypedSplice(Apply(unapplyFn, dummyArg :: Nil)))
         val unapplyImplicits = unapplyApp match {
--- a/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
+++ b/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
@@ -811,8 +811,7 @@ class ErrorMessagesTests extends ErrorMessagesTest {
     .expect { (ictx, messages) =>
       implicit val ctx: Context = ictx
 
-      assertMessageCount(1, messages)
-      val AnonymousFunctionMissingParamType(param, args, _, pt) :: Nil = messages
+      val AnonymousFunctionMissingParamType(param, args, _, pt) = messages.last
       assertEquals("x$1", param.show)
       assertEquals(s"List(ValDef(${param.show},TypeTree,EmptyTree))", args.toString)
       assertEquals("?", pt.show)
--- a/tests/pos/i2104.scala
+++ b/tests/pos/i2104.scala
@@ -1,0 +1,20 @@
+case class Pair[A, B](_1: A, _2: B)
+
+trait Cons[+H, +T]
+
+object Cons {
+  def apply[H, T](h: H, t: T): Cons[H, T] = ???
+  def unapply[H, T](t: Cons[H, T]): Option[Pair[H, T]] = ???
+}
+
+
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    Cons(Option(1), None) match {
+      case Cons(Some(i), None) =>
+        i: Int // error: found: Any(i), requires: Int
+        assert(i == 1)
+    }
+  }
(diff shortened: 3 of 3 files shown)


compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
```diff
@@ -811,8 +811,7 @@ class ErrorMessagesTests extends ErrorMessagesTest {
     .expect { (ictx, messages) =>
       implicit val ctx: Context = ictx
 
-      assertMessageCount(1, messages)
-      val AnonymousFunctionMissingParamType(param, args, _, pt) :: Nil = messages
+      val AnonymousFunctionMissingParamType(param, args, _, pt) = messages.last
```
odersky: This is very myserious: If I copy out the code of the test I always see 4 messages - one "missing param type", followed by 3 "cannot resolve overloading". that happens no matter whether I am on master or this PR. 

On master the ErrorMessageTests succeeded, i.e. they report only one failure instead of the 4 we see in the terminal. On this PR this is no longer the case. Instead, it reports all 4 tests in `messages`, but in reverse order. 

I have no idea what could be the cause of these effects.

allanrenucci: The reporter for the error message tests is mixed in with `dotty.tools.dotc.reporting.UniqueMessagePositions` which suppresses multiple error messages per position. This is why only one error message is reported. This PR must have changed the position of the reported error messages, so that now the test sees 4 messages as well.

liufengyun: The problem is that in master, there is no overlap in positions of the 4 errors, and this PR doesn't change positions of the errors, and the order doesn't change either.

liufengyun: The root cause is the following line in `Inferencing.scala` (https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Inferencing.scala#L339-L342):

```Scala
      val hasUnreportedErrors = ctx.typerState.reporter match {
        case r: StoreReporter if r.hasErrors => true
        case _ => false
      }

      // ....

      if (!hasUnreportedErrors)
        vs foreachBinding { (tvar, v) =>
          if (v != 0 && ctx.typerState.constraint.contains(tvar)) {
            // previous interpolations could have already instantiated `tvar`
            // through unification, that's why we have to check again whether `tvar`
            // is contained in the current constraint.
            typr.println(s"interpolate ${if (v == 1) "co" else "contra"}variant ${tvar.show} in ${tp.show}")
            ensureConstrained()
            tvar.instantiate(fromBelow = v == 1)
          }
        }

```

In normal command line compilation, we are using `ConsoleReporter`, `hasUnreportedErrors` will be false, despite there's error in `ConsoleReporter` ( unspecified type of anonymous function), thus resulting instantiating of `List.unapply[A](null)` with `A = Nothing`. The instantiation in turn leads to 3 overloading errors.

In the testing, we are using `ConsoleReporter`, thus `hasUnreportedErrors` will be `true` due to the unspecified type of the anonymous function. Now, instantiation of `A` in ``List.unapply[A](null)` will be delayed, after type checking `(1, 2, 3)`. So there will not be overloading errors and `A = Int` eventually.

 After the change in this PR, we are triggering early instantiation even for `StoreReporter`, thus now in test we get 4 errors.

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -920,6 +920,7 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
         val ownType =
           if (selType <:< unapplyArgType) {
             unapp.println(i"case 1 $unapplyArgType ${ctx.typerState.constraint}")
+            fullyDefinedType(unapplyArgType, "pattern selector", tree.pos)
```
liufengyun: From the diagnosis, the type parameter could be further constrained by patterns, thus I'm not sure if this change is the right fix. However, given that there is not test case invalidated this change, maybe we can accept the fix for now.

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
odersky: Thanks for clearing this up, Fengyung!

compiler/src/dotty/tools/dotc/typer/Applications.scala
odersky: Yes, I think it is analogous to fully instantiating the match scrutinee. In pattern matching all information flows from the scrutinee via the expected type to the pattern, never the other way round.

compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala
liufengyun: Do we need to change or document `hasUnreportedErrors` somehow? The fact that different reporters will influence type inference doesn't  sound correct.

PR #3906, opened 2018-01-24
Split run test to avoid timeouts

nicolasstucki:
Such as in http://dotty-ci.epfl.ch/lampepfl/dotty/2813/5


nicolasstucki:
I will split some other tests

nicolasstucki:
Now the run tests that were timing out are 3~4X faster when running the compiled test.

PR #3913, opened 2018-01-25
Allow interpretation of inline methods in value classes and varargs

nicolasstucki:


nicolasstucki:
I will merge it like this. It will serve as a regression test for the definition of macro string interpolators. We will create an issue for the case where we inline an explicit value class.

PR #3915, opened 2018-01-25
Fix #3900: add check for inline class parameters

### LINKED ISSUE #3900: inline val class parameter in backend
```scala
class Foo(inline val i: Int)
```

emits the following error
```
Concrete method has no definition: Concrete method has no definition: DefDef(i,List(),List(List()),TypeTree[TypeRef(ThisType(TypeRef(NoPrefix,module class scala)),class Int)],EmptyTree)
one error found
```

biboudis:


allanrenucci:
From the discussion in #3900, this PR should also check for inline parameters for secondary constructors

PR #3918, opened 2018-01-25
Fix #1754: Don't narrow GADTs to lower bounds

### LINKED ISSUE #1754: Soundness issue with type pattern
This doesn't compile, as expected:
```scala
case class One[T](fst: T)

object Test {
  def good[T](e: One[T]) = e match {
    case One(_) =>
      val t: T = e.fst
      val nok: Nothing = t // error, as expected
  }
}
```
But this does:
```scala
case class One[T](fst: T)

object Test {
  def bad[T](e: One[T]) = e match {
    case foo: One[a] =>
      val t: T = e.fst
      val nok: Nothing = t // should not compile
  }
}
```

odersky:
neg/i1754.scala succeeded because a GADT bound for `A` in `A <: B`
was narrowed to the lower bound of `B` (which was `Nothing`) instead of
`B` itself. Fixing this uncovered several other problems that were
hidden by the overly aggressive narrowing "feature".

In the end the PR had to do a major cleanup so that now type bounds and type
variables in patterns are treated the same. I.e. it makes no difference whether
one writes a type pattern `C[_ >: L <: H]` or `C[t >: L <: H]` where `t` is otherwise
not used.

It also involved a major refactoring in TypeComparer to make it harder to accidentally narrow to bounds that are too tight.

Based on #3889.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -354,9 +354,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
               narrowGADTBounds(tp2, tp1, approx, isUpper = false)) &&
             GADTusage(tp2.symbol)
         }
-        val tryLowerFirst = frozenConstraint || !isCappable(tp1)
-        if (tryLowerFirst) isSubType(tp1, lo2, approx.addHigh) || compareGADT || fourthTry
-        else compareGADT || fourthTry || isSubType(tp1, lo2, approx.addHigh)
+        isSubType(tp1, lo2, approx.addHigh) || compareGADT || fourthTry
 
       case _ =>
         val cls2 = tp2.symbol
@@ -389,7 +387,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
           val alwaysTrue =
             // The following condition is carefully formulated to catch all cases
             // where the subtype relation is true without needing to add a constraint
-            // It's tricky because we might need to either appriximate tp2 by its
+            // It's tricky because we might need to either approximate tp2 by its
             // lower bound or else widen tp1 and check that the result is a subtype of tp2.
             // So if the constraint is not yet frozen, we do the same comparison again
             // with a frozen constraint, which means that we get a chance to do the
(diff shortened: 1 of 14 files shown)


odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -188,8 +189,27 @@ class FirstTransform extends MiniPhase with InfoTransformer { thisPhase =>
...
+      case _ =>
+        foldOver(annots, t)
+    }
+  }
+
+  /** Replace type tree `t` of type `T` with `TypeTree(T)`, but make sure all
+   *  binders in `t` are maintained by rewrapping binders around the type tree.
+   *  E.g. if `t` is  `C[t @ (>: L <: H)]`, replace with
+   *  `t @ TC[_ >: L <: H]`. The body of the binder `t` is now wrong, but this does
+   *  not matter, as we only need the info of `t`.
```
smarter: I understand what the code does but I'm confused by the documentation :). What are `C` and `TC` here? Are you referring to two differents `t` when you say "`t` is  `C[t @ (>: L <: H)]`" ? The comment also says that we rewrap binders around the type tree but the code wraps the type tree with an annotation, not a bind.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -106,736 +107,761 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
...
+      *  If both bounds are the same, continue with `tp1 <:< app2`.
+      *  otherwise continue with either
+      *
+      *    tp1 <:< tp2    using fourthTry (this might instantiate params in tp1)
+      *    tp1 <:< app2   using isSubType (this might instantiate params in tp2)
+      */
+      def compareLower(tycon2bounds: TypeBounds, tyconIsTypeRef: Boolean): Boolean =
+        if (tycon2bounds.lo eq tycon2bounds.hi)
+          if (tyconIsTypeRef) recur(tp1, tp2.superType)
+          else isSubApproxHi(tp1, tycon2bounds.lo.applyIfParameterized(args2))
```
smarter: Is subapprox needed here given that `tycon2bounds.lo eq tycon2bounds.hi` ?

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -826,6 +826,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         false
       } else isSubType(tp1, tp2, approx.addLow)
 
+    def isSubApproxHi(tp1: Type, tp2: Type): Boolean =
+      (tp2 ne NothingType) && isSubType(tp1, tp2, approx.addHigh)
```
smarter: Is the Nothing check an optimization? Any chance it could do the wrong thing if `tp1` is a refinement of Nothing like `Nothing { type T }` ?

smarter:
After rebasing, the bytecode test for efficient try/catch added in https://github.com/lampepfl/dotty/pull/3929 started failing. I pushed a commit to fix this by adding a `.stripAnnots`.

odersky:
I am going to merge now because other PR #3841 depends on the fixes here.

PR #3921, opened 2018-01-26
Fix #1857: Allow lower bounds to influence implicit search

### LINKED ISSUE #1857: implicit summon fails in dotty, works in 2.*

odersky:
The previous implementation claimed to conform to the spec:

      We exclude lower bounds to conform to SLS 7.2:
      "The parts of a type T are: [...] if T is an abstract type, the parts of its upper bound"

But in fact we do not need to exclude lower bounds from `namedPartsWith` to do that -
TypeRefs are already lifted to classes, which implements SLS 7.2. Including lower bounds
does affect wildcard types (i1857.scala) and wildcard parameters in implicits (implicit-lower-bound.scala), but the spec does not exclude either of these.

### DIFF
--- a/tests/pos/implicit-lower-bound.scala
+++ b/tests/pos/implicit-lower-bound.scala
@@ -9,6 +9,6 @@ class Test {
   def get1(implicit lf: List[_ <: Bar]) = {}
   def get2(implicit lf: List[_ >: Bar]) = {}
 
-  get1 // works
-  get2 // error
+  get1
+  get2
 }
--- a/compiler/src/dotty/tools/dotc/typer/Implicits.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Implicits.scala
@@ -459,10 +459,7 @@ trait ImplicitRunInfo { self: Run =>
             }
             tp.classSymbols(liftingCtx) foreach addClassScope
           case _ =>
-            // We exclude lower bounds to conform to SLS 7.2:
-            // "The parts of a type T are: [...] if T is an abstract type, the parts of its upper bound"
-            for (part <- tp.namedPartsWith(_.isType, excludeLowerBounds = true))
-              comps ++= iscopeRefs(part)
+            for (part <- tp.namedPartsWith(_.isType)) comps ++= iscopeRefs(part)
         }
         comps
       }
--- a/tests/pos/i1857.scala
+++ b/tests/pos/i1857.scala
@@ -1,0 +1,38 @@
+package commandeer
+
+
+trait CommandeerDSL[Host] {
+  trait Operation[T]
+  type Op[T] <: Operation[T]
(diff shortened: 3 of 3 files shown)


odersky:
Test optimized hung. It's doing that a lot now. Can we find out the root cause of this?

tests/pos/implicit-lower-bound.scala
```diff
@@ -9,6 +9,6 @@ class Test {
   def get1(implicit lf: List[_ <: Bar]) = {}
   def get2(implicit lf: List[_ >: Bar]) = {}
 
-  get1 // works
-  get2 // error
+  get1
+  get2
```
smarter: I'll note that this line doesn't compile with scalac, so scalac interpretation of SLS 7.2 seems to match what I had in mind.

PR #3892, opened 2018-01-23
Fix #3886: add tests for dotc/dotr/dotd

liufengyun:
Fix #3886: add tests for dotc/dotr/dotd

PR #3924, opened 2018-01-26
Fix #3901: Survive parameter dependencies on by-name parameters

### LINKED ISSUE #3901: Intersection of by-name and singleton type as result of application
The following code produces an error complaining about `cond.type` not being stable. In the subsequent function application we nonetheless try to compute the intersection of non-value type `=> Boolean` and `Singleton`, which is caught by an assertion in `AndType`.

```scala
object Crash {
  def f(cond: => Boolean): cond.type = ???
  f(true)
}
```

odersky:


liufengyun:

PR #3926, opened 2018-01-26
Cleanup tests

OlivierBlanvillain:
This PR cleans up the tests, to summarize:

- Activate all the passing tests from tests/pending/* and tests/untried/*

- Disable all the the pending tests that use `scala.reflect._`, existential types and are depending on things from `scala.tools.partest._`

- Clean up warnings from pending tests

- Clean up all orphan .check and .flags files

This is best reviewed commit by commit


liufengyun:

OlivierBlanvillain:
@allanrenucci This is ready for review!

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -55,14 +55,14 @@ trait ParallelTesting extends RunnerOrchestration { self =>
...
-    def title: String = self match {
+    def title: String = (self match {
       case self: JointCompilationSource =>
         if (self.files.length > 1) name
         else self.files.head.getPath
 
       case self: SeparateCompilationSource =>
         self.dir.getPath
-    }
+    }).stripPrefix("../")
```
allanrenucci: What about we change the working directory when running tests? Something like:
```scala
baseDirectory in test := baseDirectory in ThisBuild
```
We could get rid of all the `"../` in `CompilationTest`

tests/run/Meter.scala
```diff
@@ -69,7 +69,7 @@ object Test extends dotty.runtime.LegacyApp {
 
 
   println("x.hashCode: "+x.hashCode)
-  println("x == 1: "+(x == 1))
+  println("x == 1: "+false) // Values of types a.Meter and Int cannot be compared with == or !=
```
allanrenucci: I would remove this line or comment it out

tests/run/MeterCaseClass.scala
```diff
@@ -66,7 +66,7 @@ object Test extends dotty.runtime.LegacyApp {
 
 
   println("x.hashCode: "+x.hashCode)
-  println("x == 1: "+(x == 1))
+  println("x == 1: "+false) // Values of types a.Meter and Int cannot be compared with == or !=
```
allanrenucci: I would remove this line or comment it out

tests/run/literals.scala
```diff
@@ -90,10 +90,10 @@ object Test {
     check_success("1e1f == 10.0f", 1e1f, 10.0f)
     check_success(".3f == 0.3f", .3f, 0.3f)
     check_success("0f == 0.0f", 0f, 0.0f)
-    check_success("01.23f == 1.23f", 01.23f, 1.23f)
+    check_success("01.23f == 1.23f", /*0*/1.23f, 1.23f) // Non-zero numbers may not have a leading zero.
```
allanrenucci: I would remove this line or comment it out

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -189,7 +189,6 @@ class CompilationTests extends ParallelTesting {
     compileFile("../tests/neg-custom-args/xfatalWarnings.scala", defaultOptions.and("-Xfatal-warnings")) +
     compileFile("../tests/neg-custom-args/i3561.scala", defaultOptions.and("-Xfatal-warnings")) +
     compileFile("../tests/neg-custom-args/pureStatement.scala", defaultOptions.and("-Xfatal-warnings")) +
-    compileFile("../tests/neg-custom-args/i3589-a.scala", defaultOptions.and("-Xfatal-warnings")) +
```
allanrenucci: Nice! We support emitting switch on value classes. Maybe open an issue to figure out if this is expected or not.

OlivierBlanvillain:
@allanrenucci 

- No idea about the tests that fail only on legacy or with dotty bootstrapped. But that fact that you see these in the commits is just an artifact of the way I did moves things around: first move the tests that were green under `sbt testCompilation`, then fix what broke.

- I don't see how the tests in pending/run would check warnings, then only tests that could potentially check warnings would be in pending/neg with the fatal warnings flag. I manually checked all the flags touched by 6a950cd, and none of them where neg/*.flags, so I think I didn't lose any warning related flags.

- I tried the change in Build.scala that you suggested, let see if the CI is happy with it!

OlivierBlanvillain:
@allanrenucci There is an unfortunate side effect of `baseDirectory in Test := baseDirectory.value / ".."`, which is that `.class` at the root of the dotty repo will be picked up when compiling the tests. I'm not sure that's better than the original `.stripPrefix("../")` solution...

OlivierBlanvillain:
@allanrenucci Review addressed, let me know what you think!

PR #3927, opened 2018-01-26
Fix #1169: Fix strange type formatting in error message

### LINKED ISSUE #1169: Incorrect bridge gen for method that has value class with underlying Object type as result type

odersky:
We already generate an error message for i1169.scala, but it looked weird:

    |bridge generated for member method produce=> X in class XProducer
    |which overrides method produce=> T in trait Producer
    |clashes with definition of the member itself; both have erased type (): Object.

The fix changes how ExprTypes in method signatures are handled. We now get:

    |bridge generated for member method produce: X in class XProducer
    |which overrides method produce: T in trait Producer
    |clashes with definition of the member itself; both have erased type (): Object."

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Bridges.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Bridges.scala
@@ -56,7 +56,13 @@ class Bridges(root: ClassSymbol)(implicit ctx: Context) {
       bridgesScope.lookupAll(member.name).exists(bridge =>
         bridgeTarget(bridge) == member && bridge.signature == other.signature)
     def info(sym: Symbol)(implicit ctx: Context) = sym.info
-    def desc(sym: Symbol)= i"$sym${info(sym)(preErasureCtx)} in ${sym.owner}"
+    def desc(sym: Symbol)= {
+      val infoStr = info(sym)(preErasureCtx) match {
+        case ExprType(info) => i": $info"
+        case info => info.show
+      }
+      i"$sym$infoStr in ${sym.owner}"
+    }
     if (member.signature == other.signature) {
       if (!member.info.matches(other.info))
         ctx.error(em"""bridge generated for member ${desc(member)}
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/Bridges.scala
```diff
@@ -56,7 +56,13 @@ class Bridges(root: ClassSymbol)(implicit ctx: Context) {
       bridgesScope.lookupAll(member.name).exists(bridge =>
         bridgeTarget(bridge) == member && bridge.signature == other.signature)
     def info(sym: Symbol)(implicit ctx: Context) = sym.info
-    def desc(sym: Symbol)= i"$sym${info(sym)(preErasureCtx)} in ${sym.owner}"
+    def desc(sym: Symbol)= {
```
allanrenucci: Maybe:
```scala
val infoStr = sym.showDcl(preErasureCtx)
i"$infoStr in ${sym.owner}"
```
This will print `def produce: => X`. But I think we could configure the printer to not print the `=>`. Other part of the compiler (such as error messages) could benefit from it.

liufengyun:

compiler/src/dotty/tools/dotc/transform/Bridges.scala
odersky: The problem is that printing the type will always print the `=>`. Otherwise it would be confusing. I would should work on the Printer but it has to look at the larger context to do the right thing.

Issue #746, opened 2015-08-04
Overzealous pattern elimination for try expressions.


odersky:
Trying to fix NonLocalReturns I tried to compile this:

```
object Test {

  def f(x: Int) = try {
    ???
  }
  catch {
    case ex: scala.runtime.NonLocalReturnControl[T @scala.unchecked] =>
      ???
  }
}
```

and got this output:

``` scala
result of test.scala after TreeTransform:{patternMatcher, explicitOuter, splitter}:
package <empty> {
  final lazy module val Test: Test$ = new Test$()
  final module class Test$() extends Object() { this: Test.type => 
    def f(x: Int): Nothing = 
      try {
        ???
      } catch {
        case ex1 @ _ => 
          case val selector11: Throwable = ex1
          {
            def case11(): Nothing = {
              def case21(): Nothing = {
                def matchFail11(): Nothing = throw new MatchError(selector11)
                {
                  throw ex1
                }
              }
              if 
                selector11.isInstanceOf[
                  runtime.NonLocalReturnControl[T @unchecked]
                ]
               then {
                val ex: runtime.NonLocalReturnControl[T @unchecked] = 
                  selector11.asInstanceOf[
                    runtime.NonLocalReturnControl[T @unchecked](ex)
                  ]
                {
                  {
                    ???
                  }
                }
              } else case21()
            }
            case11()
          }
      }
  }
}
```

Needless to say, this makes every try very expensive. We need to back off here and keep typed patterns until the backend can convert them into proper exception handlers.

This currently blocks the implementation of non local returns because I do not know what code should be generated; i.e. what the precise code of the given example is after pattern matching.

smarter:
Looks like this is still an issue in 2018 :).

PR #3928, opened 2018-01-26
Fix #756: Generate efficient try cases for parameterized exceptions

### LINKED ISSUE #756: dotc may generate duplicate super accessors, resulting in ClassFormatError

odersky:
We were missing a case for AppliedType

### DIFF
--- a/tests/pos/i756.scala
+++ b/tests/pos/i756.scala
@@ -1,0 +1,10 @@
+class Test[T] {
+
+  def f(x: Int) = try {
+    ???
+  }
+  catch {
+    case ex: scala.runtime.NonLocalReturnControl[T @scala.unchecked] =>
+      ???
+  }
+}
--- a/compiler/src/dotty/tools/dotc/transform/TryCatchPatterns.scala
+++ b/compiler/src/dotty/tools/dotc/transform/TryCatchPatterns.scala
@@ -76,6 +76,8 @@ class TryCatchPatterns extends MiniPhase {
       (pre == NoPrefix || pre.widen.typeSymbol.isStatic) && // Does not require outer class check
       !tp.symbol.is(Flags.Trait) && // Traits not supported by JVM
       tp.derivesFrom(defn.ThrowableClass)
+    case tp: AppliedType =>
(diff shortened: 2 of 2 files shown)


smarter:
Wrong issue number, it's #746, not #756

tests/pos/i756.scala
```diff
@@ -0,0 +1,10 @@
+class Test[T] {
+
+  def f(x: Int) = try {
+    ???
+  }
+  catch {
+    case ex: scala.runtime.NonLocalReturnControl[T @scala.unchecked] =>
```
smarter: This should probably be a BytecodeTest that checks that we don't use the instanceof opcode: https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala

odersky:
I pushed a new PR with the corrected issue number

PR #3929, opened 2018-01-26
Fix #746: Generate efficient try cases for parameterized exceptions

### LINKED ISSUE #746: Overzealous pattern elimination for try expressions.

odersky:
We were missing a case for AppliedType


odersky:
I agree with @smarter's comment on #3928 that it would be good to have a bytecode test. I don't know enough about the bytecode testing framework to be able to do one quickly, though. So if someone else wants to add one, please do!

allanrenucci:
The CI failure should go away after rebasing. I can do it while adding a bytecode test

odersky:
@allanrenucci Thanks for the test!

PR #3934, opened 2018-01-28
Restore caching of CachedMethodType

smarter:
Looks like this was accidentally removed in
81dc91db0ef630dd5cc6527f1aef5a048f8115ed

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2887,7 +2887,7 @@ object Types {
     }
 
     final def apply(paramNames: List[TermName])(paramInfosExp: MethodType => List[Type], resultTypeExp: MethodType => Type)(implicit ctx: Context): MethodType =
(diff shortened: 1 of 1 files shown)


smarter:
test performance please

allanrenucci:
test performance please

smarter:
> [check /data/workspace/bench/logs/pull-3934-01-28-14.09.out for more information]

@liufengyun Where can I find the logs for this failure?

liufengyun:
There is a stackoverflow testing Quotes: https://github.com/liufengyun/bench/blob/master/tests/power-macro/PowerInlined-1k.scala

I've disabled the test and now the test is running -- now per bench test takes about 2h.

odersky:
Good catch.

odersky:
Let's move the Quote test to pending for now, so that performance tests will work more reliably.

smarter:
The version of this test we have in Dotty is currently disabled: https://github.com/lampepfl/dotty/blob/master/tests/pos/power-macro/PowerInlined-1k_2.scala

PR #3939, opened 2018-01-29
Fix #3938: adapt prefix to ease prefix inference

### LINKED ISSUE #3938: Erroneous unreachable warning with path dependant type
```scala
class Foo {
  val bar = new Bar
  class Bar {
    sealed abstract class A
    case class B() extends A
    case class C() extends A
  }
}

class Test {
  val foo = new Foo
  import foo.bar._

  def test(a: A) = {
    a match {
      case B() => 1
      case _ => 2 // unreachable code
    }
  }
}
```

liufengyun:
Fix #3938: adapt prefix to ease prefix inference

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -597,9 +597,9 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
    *  Otherwise, return NoType.
    *
    */
-  def instantiate(tp1: Type, tp2: Type)(implicit ctx: Context): Type = {
+  def instantiate(tp1: NamedType, tp2: Type)(implicit ctx: Context): Type = {
     // expose abstract type references to their bounds or tvars according to variance
-    abstract class AbstractTypeMap(maximize: Boolean)(implicit ctx: Context) extends TypeMap {
+    class AbstractTypeMap(maximize: Boolean)(implicit ctx: Context) extends TypeMap {
       def expose(tp: TypeRef): Type = {
         val lo = this(tp.info.loBound)
         val hi = this(tp.info.hiBound)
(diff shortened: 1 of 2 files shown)


liufengyun:
@odersky : we are doing _prefix inference_ in exhaustivity check, this PR does some magic in order for inference to work smoothly. Could you have a look and see if it makes sense?

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -638,15 +640,31 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       }
     }
 
+    // Fix subtype checking for child instantiation,
+    // such that `Foo(Test.this.foo) <:< Foo(Foo.this)`
```
odersky: I could not correlate the claim `Foo(Test.this.foo) <:< Foo(Foo.this)` in i3939.scala with the code. In i3938, foo is not a module, so how does the `removeThisType` affect it?

Also, note that if we do have a 

     object Foo

then we have already `Foo.type <:< Foo.this` and `Foo.this <:< Foo.type` by the rules for `ThisType` in TypeComparer's firstTry and secondTry methods. So I am sure why the code would be necessesary.

liufengyun: Thanks for the clarification. The correlation is that in the `else` branch, it throws away of the `ThisType`. For the `module` case, I'll try if I can throw it away.

liufengyun: I just checked, the patch for `module` cannot be removed, as the subtyping for `Foo.type <:< Foo.this` only works if it's static owner:

```Scala
          case tp1: NamedType if cls2.is(Module) && cls2.eq(tp1.widen.typeSymbol) =>
            cls2.isStaticOwner ||
            isSubType(tp1.prefix, cls2.owner.thisType) ||
            secondTry(tp1, tp2)
```

The test case `tests/patmat/3455.scala` shows that we need the patch:

```Scala
trait AxisCompanion {
   sealed trait Format
   object Format {
      case object Decimal extends Format
      case object Integer extends Format
   }
}
object Axis extends AxisCompanion
class Axis {
   import Axis._
   def test( f: Format ) = f match {
      case Format.Integer => "Int"
   }
}
```

PR #3940, opened 2018-01-29
Override Console.out and Console.err the REPL

allanrenucci:
This make it possible to test command such as `println` in the REPL
scripted tests.

### DIFF
--- a/compiler/src/dotty/tools/repl/ReplDriver.scala
+++ b/compiler/src/dotty/tools/repl/ReplDriver.scala
@@ -81,7 +81,7 @@ case class Completions(cursor: Int,
 
 /** Main REPL instance, orchestrating input, compilation and presentation */
 class ReplDriver(settings: Array[String],
-                 protected val out: PrintStream = System.out,
+                 protected val out: PrintStream = Console.out,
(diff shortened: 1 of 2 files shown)


allanrenucci:
I am not 100% this is the right solution. I guess hijacking `Console.in` and `Console.out` when the JVM is shared can lead to surprising behaviors (like when running inside sbt).

allanrenucci:
`Console.out` is initialised to `System.out`. Then if `System.out` is later on mutated, it will not impact `Console.out`. However `Console.out` is effectively mutated by `Console.withOut(out) { ... }`. The old value is restored while returning but the mutation can be seen by another thread.

>It looks good to me in that it solves the issue - but how resilient would it be if there is logging to stdout for instance? Eg logback and friends.

This implementation is not mutating `System.out`, so anything that you run within the REPL that uses the Java API will output to `System.out`. We could also set `System.out` and `System.err` in `withRedirectedOutput`, but unless there is a use case, I would be conservative and not play with 🔥

felixmulder:
Sounds reasonable, we did something similar in Vulpix FYI. 

But yeah, LGTM :)

PR #3942, opened 2018-01-29
Splitter: Avoid unnecessary allocations (20% fewer Apply nodes)

smarter:
When compiling dotty-compiler, this reduces the total number of
allocated Apply nodes from 649879 to 506536, and the total number of
allocated TypeApply nodes from 111097 to 95094


smarter:
test performance please

PR #3944, opened 2018-01-30
Fix #3690: Move RestoreScopes after ExpandPrivate

### LINKED ISSUE #3690: Question on RestoreScopes and ExpandPrivate breaking scopes
According to the doc of RestoreScopes:
```scala
/** The preceding lambda lift and flatten phases move symbols to different scopes
 *  and rename them. This miniphase cleans up afterwards and makes sure that all
 *  class scopes contain the symbols defined in them.
 */
```
But ExpandPrivate comes later in the same group and renames denotations. This means the scopes can be broken since members are entered in scopes by names. This could be fixed by moving ExpandPrivate before RestoreScopes, but we haven't done this so far and nothing broke, so:
- Is this actually a problem? Or is breaking scopes inside classes at this point ok for some reason?
- If it's actually a problem, is moving ExpandPrivate before RestoreScopes the correct way to fix this?

/cc @odersky

odersky:


odersky:
I know the commit has the wrong number, but I am willing to let it go instead of wasting more time fixing it.

PR #3948, opened 2018-01-30
PostTyper#normalizeTypeArgs: fast path to avoid allocations

smarter:
According to -Yprofile-enabled, this reduces the amount of memory
allocated in PostTyper when compiling re2s from 39 MB to 26 MB


smarter:
@nicolasstucki I had to disable tests/run-with-compiler/quote-run-constants-extract.scala because it times out.

odersky:
I believe some of the quote tests are too aggressive, causing stackoverflows or timeouts in normal situations. We should make them less resource hungry.

PR #3951, opened 2018-01-30
Erasure: reuse TypeTrees when possible

smarter:
If the erased type is equal to the unerased type, we can reuse
TypeTrees.
When compiling re2s, this reduces the number of allocated TypeTrees from
34870 to 34477. It should be possible to do better if unerased types
used TermRef less often, for example:
```scala
TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),module scala),class Boolean)
```
gets erased to:
```scala
TypeRef(ThisType(TypeRef(NoPrefix,module class scala)),class Boolean)
```
but the latter type could also have been used pre-erasure.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/Erasure.scala
+++ b/compiler/src/dotty/tools/dotc/transform/Erasure.scala
@@ -535,9 +535,12 @@ object Erasure {
       }
     }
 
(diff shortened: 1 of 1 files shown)


smarter:
Yeah I'm not sure, but it might enable future improvements. Feel free to close the PR.

PR #3953, opened 2018-02-01
Experiment: optimize TimeTravellingTreeCopier

smarter:
Currently, we always recompute types for Apply, TypeApply and Closure
nodes but this is a bit extreme, I think that the same conditions as in
TypedTreeCopier can be used to avoid recomputation if we also check for
widened types.


smarter:
test peformance please

smarter:
test performance please

smarter:
Closing for now, I'll come back to this.

PR #3954, opened 2018-02-01
Close #3933 Improvements to the IntelliJ IDEA getting started guide

### LINKED ISSUE #3933: Improvements to the IntelliJ IDEA getting started guide
I would like to create a pull request to update the IDEA getting started guide (https://github.com/lampepfl/dotty/blob/master/docs/docs/contributing/intellij-idea.md), so this is an associated issue.

Current weaknesses:

* The last step of setup seems out of date. Using the newest version of the community version of IDEA, it does not ask which modules to load. You have to manually unload unnecessary ones after the project loads.
* It is confused at one point. At the beginning of the running section, it says it is presenting an sbt command and instead there is a shell command which incidentally will not run as written. Then it neglects how to run dotty using a shell command (granted, it's easy enough to guess it's in the bin directory).
* The ending is vague. It does not specify that the `dotc` command has to be run from within `sbt` to work, and for me who had not used remote debugging, it took googling and experimenting to understand.
* There are grammatical errors.

expz:
This pull request makes improvements to the IntelliJ IDEA getting started guide. The types of improvements are listed in the associated issue #3933.

### DIFF
--- a/docs/docs/contributing/intellij-idea.md
+++ b/docs/docs/contributing/intellij-idea.md
@@ -20,40 +20,46 @@ its installed on your local machine). Otherwise, specify it by pressing *New*.
 
 ![](../../images/idea/idea-sdk.png "Select the JDK")
 
-On the final window we must select which modules we can import. Here, we are presented with the full list of SBT projects
-that are defined in Dotty. You can either select all (expect performance degradation by IDEA if you select all) or
+Next we must select which modules we can import. IDEA version 2017.1 and earlier presents the full list of SBT modules
+that are defined in Dotty. You can either select all (expect performance degradation if you select all) or
 select only the `dotty` module. In order to do that, unselect all modules and select on `dotty`. IDEA, then, automatically
 selects all the necessary dependencies and you press OK.
 
 ![](../../images/idea/idea-sbt.png "Select modules to import")
 
+In IDEA version 2017.2 and later, wait for the project to load. Then right click on `dotty` in the project explorer
+and select `Load/Unload Modules...`. Select only `dotty`, accept when IDEA asks to add its dependencies and press OK.
+
 Running/Debugging
 -------
 
-To run the compiler you can do it either as an sbt command or via debugging the compiler.
-For the first option you can fire up sbt from the `Terminal` window of IDEA or you can do it externally.
-For example, to run a test you can write with or without flags:
+To run the compiler you can do it either as an sbt command or a shell script. Open an external terminal.
+For the first option, a test can be run as follows.
 
 ```shell
-$ dotc tests/pos/Arrays.scala
+$ sbt
+> dotc tests/pos/Arrays.scala
+```
+For the second option, the compiler can be run using a script in the `bin` directory:
+
+```shell
+$ ./bin/dotc tests/pos/Arrays.scala
 ```
 
-If you are interested in debugging the compiler you can enable the necessary agent on the JVM.
-This is done in two steps. For the first step you need to pass the 
-necessary flag to the running VM. For convenience, this is already in comments on the `Build.scala` file under the
-`project` directory. The string you need to uncomment is the following:
-
-> `"-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"`
-
-Uncomment the line and reload the project file with the `reload` command on sbt. 
-Now, each time you run `dotc` the compiler listens at the designated address and waits the agent to connect.
-For, the second step you need to create a configuration for Debug through IDEA: 
+If you are interested in debugging the compiler, you can use a remote debugging configuration.
+This is done by launching dotc's JVM with the JDWP agent loaded. To that end, run
+```shell
+$ ./bin/dotc -debug tests/pos/Arrays.scala
+```
+Then when dotc starts, it will suspend and wait for a debugger to connect on port `5005`.
+Next a configuration for Debug must be created in IDEA: 
 
 > Run > Edit Configurations > Add New Configuration > (select) Remote
 
 ![](../../images/idea/idea-debug.png "Create a Debug Configuration")
 
-Now every time you run `dotc` you can set your breakpoints and hit the `Debug dotty-debug` button (since we used that name for 
- the sample configuration above). The default data on the configuration match the enabled agent on the VM so, probably,
+Under `Search sources using module's classpath` select `dotty`. Set breakpoints as desired.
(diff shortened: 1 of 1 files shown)


expz:
Please hold off on this for a few hours. I realized the explanation at the end has a small error, although the steps are right. sbt and dotc run in separate JVMs, but I wrote as if they are in the same one.

expz:
I updated the pull request 5 days ago, but it still has not run the checks. Do I need to do something to initiate them?

OlivierBlanvillain:
Something went wrong with the CI, no idea that that is... But since your diff is only on a markdown file I guess this could be review / merged without running the test suite.

allanrenucci:
The CLA is [signed](https://www.lightbend.com/contribute/cla/scala/check/expz)

docs/docs/contributing/intellij-idea.md
```diff
@@ -20,40 +20,51 @@ its installed on your local machine). Otherwise, specify it by pressing *New*.
 
 ![](../../images/idea/idea-sdk.png "Select the JDK")
 
-On the final window we must select which modules we can import. Here, we are presented with the full list of SBT projects
+In some versions of IDEA, we must select which modules we can import. Here, we are presented with the full list of SBT projects
```
biboudis: If you have that handy can you give a "v > XX.XX" format example of which versions?

docs/docs/contributing/intellij-idea.md
```diff
@@ -20,40 +20,51 @@ its installed on your local machine). Otherwise, specify it by pressing *New*.
...
+In other versions of IDEA, no choice will be given. After the project loads, right click on `dotty` in the project explorer
+and select `Load/Unload Modules...`. Then select only `dotty`, accept when IDEA asks to add its dependencies and press OK.
+
 Running/Debugging
 -------
 
-To run the compiler you can do it either as an sbt command or via debugging the compiler.
-For the first option you can fire up sbt from the `Terminal` window of IDEA or you can do it externally.
+To run the compiler you can do it either as an sbt command or a shell script.
+For the first option you can fire up sbt from the `Terminal` window of IDEA or you from an external terminal.
```
biboudis: Although error messages are linkable and nice from my personal experience it's better to run it on an external terminal so personally I would give only one option here.

docs/docs/contributing/intellij-idea.md
```diff
@@ -20,40 +20,51 @@ its installed on your local machine). Otherwise, specify it by pressing *New*.
...
+For the second option, the compiler can be run using a script in the `bin` directory:
+
+```shell
+$ ./bin/dotc tests/pos/Arrays.scala
 ```
 
-If you are interested in debugging the compiler you can enable the necessary agent on the JVM.
-This is done in two steps. For the first step you need to pass the 
-necessary flag to the running VM. For convenience, this is already in comments on the `Build.scala` file under the
+If you are interested in debugging the compiler, you can use a remote debugging configuration.
```
biboudis: Actually this step is automated now. We run `dist/pack` from sbt and then we pass the `-debug` flag when compiling with `dotc`. I think this is a great opportunity to write this small improvement down.

docs/docs/contributing/intellij-idea.md
Blaisorblade: Good question, but is there any reason why we should currently support multiple IDEA versions? Supporting only one (in the docs) seems a reasonable tradeoff.

Maybe people who don't want to upgrade can install a new version next to the other.

Gotta setup IDEA (already run into bad bugs from ScalaIDE), so I can take a look at this PR.

expz: After some testing, I found that the change happened in version 2017.2 released in November 2017.

@Blaisorblade You have a good point in general, but I can think of two reasons to support multiple versions. First, the change was made a few months ago, so many people will need the first directions. Second, this page is unique in that its main purpose is to make on-boarding as frictionless as possible. That seems to call for setting aside the general principle of supporting only the newest version that applies elsewhere.

docs/docs/contributing/intellij-idea.md
expz: Done.

docs/docs/contributing/intellij-idea.md
expz: Thanks, that way is more convenient.

I noticed it does not work from within sbt, although that is not important. It fails with a `bad option '-debug' was ignored` error. Also, perhaps it is by design, but neither the sbt `dotc` nor the script `dotc` list `-debug` in the options under the `-help` command.

expz: I just found that the `dist/pack` command was not necessary for debug to work. Do you have a guess as to what I am doing differently? I am using the most recent commit of the master branch.

Blaisorblade:
Hi @expz thanks! I'm new here and also setting up IDEA, so I'm also going through the guide! Do you mind if I also help?

docs/docs/contributing/intellij-idea.md
allanrenucci: > I noticed it does not work from within sbt

Indeed, `debug` does not work  within sbt. It might be tricky to make it work though

> I just found that the dist/pack command was not necessary for debug to work.

Yeh, `dist/pack` is not necessary. If you run `bin/dotc`, the script will run `dist/pack` for you if needed.

> Also, perhaps it is by design, but neither the sbt dotc nor the script dotc list -debug in the options under the -help command.

I don't think we should list the `debug` option with the `help` command. This an internal option meant to debug the compiler.

Blaisorblade:
I'm watching Dmitry's intro to Dotty internals, and he suggests modifying `/Applications/IntelliJ\ IDEA\ CE.app/Contents/bin/idea.vmoptions` to give IDEA more memory. IntelliJ documents a cleaner way to access that file: https://intellij-support.jetbrains.com/hc/en-us/articles/206544869-Configuring-JVM-options-and-platform-properties

That's not in our instructions. But memory usage hasn't decreased, especially if one adds more projects.

@expz Your PR already improves docs, so I'm happy to move other changes in separate PRs (and ping you if you want), rather than iterate here. Any preference?

Blaisorblade:
Merging this as-is. Further improvements can go in later PRs.

expz:
Sorry I did not see your questions in time!

Blaisorblade:
@expz Don't worry, better to have this in I guess — we can do further PRs, and I didn't want to hold this up too much. I'll tag you for the next PR.

PR #3949, opened 2018-01-30
Add sourcepath to IDE setup

odersky:
This allows the compiler to succeed even if the project is not fully built.
It also gives navigation capability into directly dependent files.

PR #3960, opened 2018-02-04
Context-sensitive IDE completions

odersky:
This provides support for full context-sensitive completions. Local scopes and imports are considered. Member completions take possible implicit conversions into account.

Based on #3949.

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/Printers.scala
+++ b/compiler/src/dotty/tools/dotc/config/Printers.scala
@@ -24,7 +24,7 @@ object Printers {
   val implicits: Printer = noPrinter
   val implicitsDetailed: Printer = noPrinter
   val inlining: Printer = noPrinter
-  val interactiv: Printer = new Printer
+  val interactiv: Printer = noPrinter
   val overload: Printer = noPrinter
   val patmatch: Printer = noPrinter
   val pickling: Printer = noPrinter
--- a/compiler/src/dotty/tools/dotc/config/Printers.scala
+++ b/compiler/src/dotty/tools/dotc/config/Printers.scala
@@ -19,12 +19,12 @@ object Printers {
   val cyclicErrors: Printer = noPrinter
   val dottydoc: Printer = noPrinter
   val exhaustivity: Printer = noPrinter
-  val incremental: Printer = noPrinter
   val gadts: Printer = noPrinter
   val hk: Printer = noPrinter
   val implicits: Printer = noPrinter
   val implicitsDetailed: Printer = noPrinter
   val inlining: Printer = noPrinter
+  val interactiv: Printer = new Printer
   val overload: Printer = noPrinter
   val patmatch: Printer = noPrinter
   val pickling: Printer = noPrinter
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -667,7 +667,7 @@ object Symbols {
         case None =>
           val idSet = mutable.SortedSet[String]()
           tree.foreachSubTree {
-            case tree: tpd.RefTree => idSet += tree.name.toString
+            case tree: tpd.NameTree => idSet += tree.name.toString
             case _ =>
           }
           val ids = idSet.toArray
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -641,7 +641,7 @@ object Symbols {
       case _ =>
         denot.ensureCompleted()
         myTree match {
-          case fn: TreeProvider => myTree = fn.getTree(ctx)
+          case fn: TreeProvider => myTree = fn.tree
           case _ =>
         }
         myTree.asInstanceOf[Tree]
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -667,7 +667,8 @@ object Symbols {
         case None =>
           val idSet = mutable.SortedSet[String]()
           tree.foreachSubTree {
-            case tree: tpd.NameTree if tree.name.isInstanceOf[SimpleName] => idSet += tree.name.toString
+            case tree: tpd.NameTree if tree.name.toTermName.isInstanceOf[SimpleName] =>
+              idSet += tree.name.toString
             case _ =>
           }
           val ids = idSet.toArray
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -510,12 +510,6 @@ object Symbols {
     final def isStatic(implicit ctx: Context): Boolean =
       lastDenot != null && lastDenot.initial.isStatic
 
-    /** A unique, densely packed integer tag for each class symbol, -1
-     *  for all other symbols. To save memory, this method
-     *  should be called only if class is a super class of some other class.
-     */
-    def superId(implicit ctx: Context): Int = -1
-
     /** This symbol entered into owner's scope (owner must be a class). */
     final def entered(implicit ctx: Context): this.type = {
       assert(this.owner.isClass, s"symbol ($this) entered the scope of non-class owner ${this.owner}") // !!! DEBUG
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -21,7 +21,7 @@ import StdNames._
 import NameOps._
 import NameKinds.LazyImplicitName
 import ast.tpd
-import tpd.{Tree, TreeProvider}
+import tpd.{Tree, TreeProvider, TreeOps}
 import ast.TreeTypeMap
 import Constants.Constant
 import reporting.diagnostic.Message
@@ -29,7 +29,7 @@ import Denotations.{ Denotation, SingleDenotation, MultiDenotation }
 import collection.mutable
 import io.AbstractFile
 import language.implicitConversions
-import util.{NoSource, DotClass}
+import util.{NoSource, DotClass, Property}
 import scala.collection.JavaConverters._
 
 /** Creation methods for symbols */
@@ -402,6 +402,9 @@ object Symbols {
 
   implicit def eqSymbol: Eq[Symbol, Symbol] = Eq
 
+  /** Tree attachment containing the identifiers in a tree as a sorted array */
+  val Ids = new Property.Key[Array[String]]
+
   /** A Symbol represents a Scala definition/declaration or a package.
    *  @param coord  The coordinates of the symbol (a position or an index)
    *  @param id     A unique identifier of the symbol (unique per ContextBase)
@@ -654,7 +657,7 @@ object Symbols {
             }
             else tpd.EmptyTree
           case tree: Tree @ unchecked =>
-            tree
+            if (id.isEmpty || mightContain(tree, id)) tree else tpd.EmptyTree
         }
     }
 
@@ -662,6 +665,22 @@ object Symbols {
 
     private[dotc] def treeOrProvider_=(t: TreeOrProvider)(implicit ctx: Context): Unit =
       myTree = t
+
+    private def mightContain(tree: Tree, id: String)(implicit ctx: Context): Boolean = {
+      val ids = tree.getAttachment(Ids) match {
+        case Some(ids) => ids
+        case None =>
+          val idSet = mutable.SortedSet[String]()
+          tree.foreachSubTree {
+            case tree: tpd.RefTree => idSet += tree.name.toString
+            case _ =>
+          }
+          val ids = idSet.toArray
+          tree.putAttachment(Ids, ids)
+          ids
+      }
+      ids.binarySearch(id) >= 0
+    }
 
     /** The source or class file from which this class was generated, null if not applicable. */
     override def associatedFile(implicit ctx: Context): AbstractFile =
--- a/compiler/src/dotty/tools/dotc/core/Symbols.scala
+++ b/compiler/src/dotty/tools/dotc/core/Symbols.scala
@@ -21,7 +21,7 @@ import StdNames._
 import NameOps._
 import NameKinds.LazyImplicitName
 import ast.tpd
-import tpd.Tree
+import tpd.{Tree, TreeProvider}
 import ast.TreeTypeMap
 import Constants.Constant
 import reporting.diagnostic.Message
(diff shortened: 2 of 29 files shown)


odersky:
I have one issue that's still remaining: When doing selection completions, let's say following

    sym.in

we often get a completion on an error tree `sym.<error>`. This is a pity since it means we cannot pre-filter on the prefix `in` (yes, VSCode does the filtering as well, but it usually includes way too many symbols, and also cannot distinguish between type names and term names).

Is there a way to get round this? I.e. force the client to give us the latest source?

odersky:
Separate TODO: The REPL should use the new completer in Interactive instead of the old one.

smarter:
> Is there a way to get round this? I.e. force the client to give us the latest source?

An error tree created by `errorTree` is just an untyped tree with an ErrorType set, so you should still be able to see the name.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -165,38 +310,91 @@ object Interactive {
...
+      ctx
+    case first :: _ if first eq stat =>
+      ctx.exprContext(stat, exprOwner)
+    case (imp: Import) :: rest =>
+      contextOfStat(rest, stat, exprOwner, ctx.importContext(imp, imp.symbol(ctx)))
+    case _ =>
+      ctx
+  }
+
+  def contextOfPath(path: List[Tree])(implicit ctx: Context): Context = path match {
```
smarter: This looks very fragile. Since Typer already has to do a similar job for typedIdent to work, I was wondering if instead we could make a subclass of Typer that:
1. Skips trees outside of the current position
2. As soon as it reaches a leaf, find all completions and then exit.

odersky:
> An error tree created by errorTree is just an untyped tree with an ErrorType set, so you should still be able to see the name.

That's not what I am seeing. I a see a lot of trees where the selector is literally "<error>". These are generated by the parser when the "." is not followed by an identifier. My theory is that the client is giving the server these trees to complete even though the user has since typed some more characters. But I don't know why that happens or how to work around it.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: That's pretty much how scalac's PC works. It's a possibility, but would be a much larger rewrite compared to what we have. Other downsides: (1) it could be a lot slower, depending how much needs to be tpechecked. (2) we create ne symbols, which might confuse thiings. (3) Generally, more entanglement of normal compiler and presentation code.

retronym:
Sort of related, let's think about how the parser enable completions like `global.def<TAB>`. I want `definitions` there, not the keyword `def`.

I worked around this in the completion work I did for the 2.12 REPL by inserting some dummy text at the cursor, but you then need to account for the length of that text when looking at positions.

So maybe the parser should be told the cursor location directly in interactive mode, and promote unexpected keywords to idents or selections.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
smarter: > (2) we create new symbols, which might confuse things.

But a ReTyper doesn't create symbols, does it? It also should be possible to not retypecheck anything and rely on the existing typechecked tree.

smarter:
> So maybe the parser should be told the cursor location directly in interactive mode, and promote unexpected keywords to idents or selections.

Couldn't we just do that all the time?

retronym:
Also, during my REPL work, I added a layer of abstraction over the scope- and member-completion APIs, [`completionsAt`](https://github.com/scala/scala/blob/ec5e2416143ccb7168fd7aab1b99389e9c1eb8d4/src/interactive/scala/tools/nsc/interactive/Global.scala#L1211-L1250). That handled some corner cases that are tested in [`CompletionsTest`](https://github.com/scala/scala/blob/ec5e2416143ccb7168fd7aab1b99389e9c1eb8d4/test/junit/scala/tools/nsc/interpreter/CompletionTest.scala). While that test is at the REPL completion level, it covers some interesting test cases for general completion.

retronym:
> Couldn't we just do that all the time?

I guess that's an implementation choice. So long as `"foo".def` remains a parse error, and we don't end up emitting a `no such member "def"` type error.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: That's true. Retypers don't create symbols. But I think messing with them will ultimately be more fragile. It comes down to whether we can coax a 3000 line module to do something quite different than what it was intended for, or whether we write a 40 line method that extracts the information directly. I was actually quite pleased how straightforward that was. True, there are some small differences, i.e. we ignore the "in supercall" distinctions. But I don't think these matter. And it's not clear that a retyper approach would not have similar differences.

odersky:
I reverted 349d79b in 858d52c. Without the wider matchSymbol definition, we would not count `apply` methods of case classes as references, for instance.

odersky:
I did some more investigation. If I type e.g. `sym.in` CTRL-TAB, even very quickly, the source for the completion contains only `sym.`. If I wait a couple of seconds and then press CTRL-TAB again, the source is `sym.in`. I am not sure why that happens and how we can mitigate it.

It's not so bad, since we will simply return all completions from the `.`, and let the client choose which one to present. When presenting, VSCode _does_ know about the prefix, but it throws a lot of other completions out which contain the prefix letters somewhere in the identifiers, not start with the prefix.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -65,7 +66,8 @@ class DottyLanguageServer extends LanguageServer
       myDrivers = new mutable.HashMap
       for (config <- configs) {
         val classpathFlags = List("-classpath", (config.classDirectory +: config.dependencyClasspath).mkString(File.pathSeparator))
-        val settings = defaultFlags ++ config.compilerArguments.toList ++ classpathFlags
+        val sourcepathFlags = List("-sourcepath", config.sourceDirectories.mkString(File.pathSeparator), "-scansource")
+        val settings = defaultFlags ++ config.compilerArguments.toList ++ classpathFlags ++ sourcepathFlags
```
smarter: When running launchIDE on Dotty we get:
> [error] Flag -sourcepath set repeatedly

Because `-sourcepath` is part of the settings we set for dotty-library in our build. This could be solved by not adding `-sourcepath` when it's already set, or by appending source directories to an existing `-sourcepath` found in `config.compilerArguments`

compiler/src/dotty/tools/dotc/config/Printers.scala
```diff
@@ -19,12 +19,12 @@ object Printers {
   val cyclicErrors: Printer = noPrinter
   val dottydoc: Printer = noPrinter
   val exhaustivity: Printer = noPrinter
-  val incremental: Printer = noPrinter
   val gadts: Printer = noPrinter
   val hk: Printer = noPrinter
   val implicits: Printer = noPrinter
   val implicitsDetailed: Printer = noPrinter
   val inlining: Printer = noPrinter
+  val interactiv: Printer = new Printer
```
smarter: Should be noPrinter

compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -625,30 +622,59 @@ object Symbols {
...
+    private[dotc] def treeOrProvider_=(t: TreeOrProvider)(implicit ctx: Context): Unit =
+      myTree = t
+
+    private def mightContain(tree: Tree, id: String)(implicit ctx: Context): Boolean = {
+      val ids = tree.getAttachment(Ids) match {
+        case Some(ids) => ids
+        case None =>
+          val idSet = mutable.SortedSet[String]()
+          tree.foreachSubTree {
+            case tree: tpd.NameTree => idSet += tree.name.toString
```
smarter: Should be restricted to `SimpleName`, like in `DottyUnpickler#mightContain` I think.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -62,13 +71,34 @@ object Interactive {
...
+   *  or `include` is `overriding`, and `tree` overrides `sym`.
+   */
+  def matchSymbol(tree: Tree, sym: Symbol, include: Include.Set)(implicit ctx: Context): Boolean = {
+
+    def overrides(sym1: Symbol, sym2: Symbol) =
+      sym1.owner.derivesFrom(sym2.owner) && sym1.overriddenSymbol(sym2.owner.asClass) == sym2
+
+    (  sym == tree.symbol
+    || sym.exists && sym == sourceSymbol(tree.symbol)
+    || include != 0 && sym.name == tree.symbol.name && sym.owner != tree.symbol.owner
```
smarter: `sym.owner` will crash if `sym` is `NoSymbol`

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -62,13 +71,34 @@ object Interactive {
...
+  }
+
   private def safely[T](op: => List[T]): List[T] =
     try op catch { case ex: TypeError => Nil }
 
   /** Get possible completions from tree at `pos`
    *
    *  @return offset and list of symbols for possible completions
    */
+  // deprecated
```
smarter: Add a TODO for removing this

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -96,18 +126,140 @@ object Interactive {
...
+      case (ref: RefTree) :: _ =>
+        if (ref.name == nme.ERROR)
+          (ref.pos.point, "", false, false)
+        else
+          (ref.pos.point,
+           ref.name.toString.take(pos.pos.point - ref.pos.point),
+           ref.name.isTermName,
+           ref.name.isTypeName)
+      case _ =>
+        println(i"COMPUTE from ${path.headOption}")
```
smarter: Stray println

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -96,18 +126,140 @@ object Interactive {
...
+    }
+
+    /** Include in completion sets only symbols that
+     *   - start with given name prefix
+     *   - do not contain '$' except in prefix where it is explicitly written by user
+     *   - have same term/type kind as name prefix given so far
+     */
+    def include(sym: Symbol) =
+      sym.name.startsWith(prefix) &&
+      !sym.name.toString.drop(prefix.length).contains('$') &&
```
smarter: Can't we just check if the name kind is SimpleName after a `stripModuleClassSuffix` ?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -96,18 +126,140 @@ object Interactive {
...
+
+    def getImportCompletions(ictx: Context): Unit = {
+      implicit val ctx = ictx
+      val imp = ctx.importInfo
+      if (imp != null) {
+        def addImport(name: TermName) = {
+          addMember(imp.site, name)
+          addMember(imp.site, name.toTypeName)
+        }
+        for (renamed <- imp.reverseMapping.keys) addImport(renamed)
```
smarter: I don't think this can work since the name `renamed` doesn't exist in `imp.site`

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -96,18 +126,140 @@ object Interactive {
...
+            addMember(imp.site, mbr.name)
+      }
+    }
+
+    def getScopeCompletions(ictx: Context): Unit = {
+      implicit val ctx = ictx
+
+      if (ctx.owner.isClass) {
+        for (mbr <- accessibleMembers(ctx.owner.thisType))
+          addMember(ctx.owner.thisType, mbr.name)
```
smarter: Equivalent to `addAccessibleMembers(ctx.owner.thisType)`

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -165,38 +312,91 @@ object Interactive {
...
+      // FIXME: We shouldn't need a cast. Change NavigateAST.pathTo to return a List of Tree?
+      val path = NavigateAST.pathTo(pos, tree, skipZeroExtent = true).asInstanceOf[List[untpd.Tree]]
+      path.dropWhile(!_.hasType).asInstanceOf[List[tpd.Tree]]
     }
+    else Nil
+
+  def contextOfStat(stats: List[Tree], stat: Tree, exprOwner: Symbol, ctx: Context): Context = stats match {
+    case Nil =>
+      ctx
+    case first :: _ if first eq stat =>
```
smarter: Why is the first statement handled specially?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -165,38 +312,91 @@ object Interactive {
...
+          contextOfStat(stats, nested, ctx.owner, localCtx)
+        case tree @ CaseDef(pat, guard, rhs) if nested `eq` rhs =>
+          val localCtx = outer.fresh.setNewScope
+          pat.foreachSubTree {
+            case bind: Bind => localCtx.enter(bind.symbol)
+            case _ =>
+          }
+          localCtx
+        case tree @ Template(constr, parents, self, _) =>
+          if ((constr :: self :: parents).exists(nested `eq` _)) ctx
```
smarter: Any reason to not use `.contains(nested)` here?

smarter:
In `InteractiveDriver.scala`, doing a `Find All References` query on the definition of `allTrees` does not bring up the usages in `DottyLanguageServer`.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
odersky: Not sure what is the right thing to do here. I was assuming config.courseDirectories gives us all the info. Does it make sense to include the dotty-library sourcepath?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: I want to exclude all compile-generated names, not just module names.

odersky:
> In InteractiveDriver.scala, doing a Find All References query on the definition of allTrees does not bring up the usages in DottyLanguageServer.

Yes, I think that's because these are different projects. I am not sure what needs to be done to also search external projects. Open a fresh issue?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
smarter: sourceDirectories does give us all the info, it's just that when compiling dotty-library itself we already use -sourcepath in our build to avoid accidentally linking against the dotty-library on the classpath. In general we cannot assume that -sourcepath is not part of `compilerArguments`.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
smarter: But all these names will have a kind different from SimpleName, no?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: Good point. I removed the code and added a FIXME. Will open a separate issue.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: It's not. There's a tail recursion in the third clause.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
odersky: None. I overlooked that tree comparison is by `eq` anyway.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
odersky: Ah, OK, so we can just filter out the existing sourcepath, right?

smarter:
Right, but to be conservative I would keep it and append our sourceDirectories to it if it already exists

odersky:
> Right, but to be conservative I would keep it and append our sourceDirectories to it if it already exists

The problem is that uncontrolled source directories can be surprising and expensive. So I'd rather keep a tight lid on it and not include some random stuff on the classpath.

Issue #3308, opened 2017-10-11
Optimize the raw"" and s"" string interpolators


smarter:
Similar to https://github.com/scala/scala/pull/6093 but should be done in a mini-phase if possible.

nerush:
Very interesting story for novice contributors to Dotty. I can give it a shot.

Wojtechnology:
@nerush are you still working on this? If not, I can take it.

nerush:
@Wojtechnology Thanks for asking. Taking into account that this issue has low priority, I am considering sending PR with port as soon as original scala/scala#6093 is merged to scala:2.12.x branch. If members think that we can port it before scala/scala#6093 is merged, you can take it for sure.

smarter:
I think it's fine to work on this now, I'll let you decide who works on what when :).

Wojtechnology:
@nerush I can start working on this if you haven't already. Let me know :).

nerush:
@Wojtechnology Sure. If you think I can help you with the implementation or review it once it is implemented you can always let me know on gitter.

PR #3961, opened 2018-02-04
String interpolation optimization miniphase

Wojtechnology:
Addresses https://github.com/lampepfl/dotty/issues/3308. `s` and `raw` string interpolators use string concatenation instead of `StringContext`. Seeing a 4x increase in performance on a very simple initial benchmark:

With optimization:
```
[info] Benchmark                  Mode  Cnt      Score       Error   Units
[info] Test.testRawInterpolator  thrpt    3  42419.376 ±  8525.679  ops/ms
[info] Test.testSInterpolator    thrpt    3  41732.471 ±  7622.324  ops/ms
[info] Test.testStrConcat        thrpt    3  40346.563 ± 14475.086  ops/ms
```
Without:
```
[info] Benchmark                  Mode  Cnt      Score      Error   Units
[info] Test.testRawInterpolator  thrpt    3   9304.561 ±  996.445  ops/ms
[info] Test.testSInterpolator    thrpt    3   8325.524 ±  677.976  ops/ms
[info] Test.testStrConcat        thrpt    3  40857.625 ± 3487.954  ops/ms
```
Performance for string interpolators is on par with string concatenation.


compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -0,0 +1,59 @@
...
+  private object StringContextIntrinsic {
+    def unapply(tree: Apply)(implicit ctx: Context): Option[(List[Tree], List[Tree])] = {
+      tree match {
+        case Apply(Select(Apply(Select(Ident(nme.StringContext), nme.apply),
+                   List(SeqLiteral(strs, _))), id), List(SeqLiteral(elems, _))) =>
+          if (id == nme.raw_) Some(strs, elems)
+          else if (id == nme.s) {
+            try {
+              val escapedStrs = strs.mapConserve { str =>
+                val strValue = str.asInstanceOf[Literal].const.stringValue
```
allanrenucci: I don't think you can assume that `str` is always a constant literal. One could write:
```scala
val foo = "Hello"
StringContext(foo).s()
```

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -0,0 +1,59 @@
...
+            }
+          } else None
+        case _ => None
+      }
+    }
+  }
+
+  override def transformApply(tree: Apply)(implicit ctx: Context): Tree = {
+    tree match {
+      case StringContextIntrinsic(strs: List[Tree], elems: List[Tree]) =>
```
allanrenucci: I would to something similar to https://github.com/scala/scala/blob/c2a5883891a68180b143eb462c8b0cebc8d3b021/src/library/scala/StringContext.scala#L119. It is very hard to understand what the code below does

Wojtechnology:
Do we usually write unit tests for mini phases? In particular, I'd like to check if the correct tree is generated after passing it to the phase. Is there any example of this kind of test?

smarter:
We rarely unit test mini phases (I think there's some in https://github.com/lampepfl/dotty/tree/master/compiler/test/dotty/tools/dotc/transform), we usually do end-to-end testing instead. Here I would suggest adding some tests that check that the same bytecode is generated when using the interpolator or when writing the equivalent code by hand, see https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/InlineBytecodeTests.scala for example.

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  override def phaseName: String = "stringInterpolatorOpt"
+
+  private object StringContextIntrinsic {
+    def unapply(tree: Apply)(implicit ctx: Context): Option[(List[Tree], List[Tree])] = {
+      tree match {
+        case Apply(Select(Apply(Select(ident, nme.apply), List(SeqLiteral(strs, _))), fn),
+            List(SeqLiteral(elems, _))) =>
+          val clsSym = ident.symbol.companionClass
+          if (clsSym.eq(defn.StringContextClass) && strs.forall(_.isInstanceOf[Literal])
```
smarter: You could check if `ident.symbol` is equal to `defn.StringContextModuleClass` instead after adding `StringContextModuleClass` to `Definitions.scala`

odersky:
What is the status of the PR? Given the performance improvements it would be great to get this in.

Wojtechnology:
Just waiting on code review from my end.

compiler/test/dotty/tools/backend/jvm/StringInterpolatorOptTest.scala
```diff
@@ -0,0 +1,64 @@
...
+      """
+                   |class Foo {
+                   |  val one = 1
+                   |  val two = "two"
+                   |  val three = 3.0
+                   |
+                   |  def meth1: String = raw"$one plus $two$three\n"
+                   |  def meth2: String = "" + one + " plus " + two + three + "\\n"
+                   |}
+                 """.stripMargin
```
allanrenucci: Use normal indentation:
```scala
      """
        |class Foo {
        |  val one = 1
        |  val two = "two"
        |  val three = 3.0
        |
        |  def meth1: String = raw"$one plus $two$three\n"
        |  def meth2: String = "" + one + " plus " + two + three + "\\n"
        |}
      """.stripMargin
```

compiler/test/dotty/tools/backend/jvm/StringInterpolatorOptTest.scala
```diff
@@ -0,0 +1,64 @@
...
+      """
+                   |class Foo {
+                   |  val one = 1
+                   |  val two = "two"
+                   |  val three = 3.0
+                   |
+                   |  def meth1: String = s"$one plus $two$three\n"
+                   |  def meth2: String = "" + one + " plus " + two + three + "\n"
+                   |}
+                 """.stripMargin
```
allanrenucci: Use normal indentation:

tests/run/interpolation-opt.scala
```diff
@@ -0,0 +1,23 @@
...
+  println(s"a$one$two${three}b")
+
+  // Test empty string interpolators
+  println(raw"")
+  println(s"")
+
+  // Make sure that StringContext still works with idents
+  val foo = "Hello"
+  val bar = "World"
+  println(StringContext(foo, bar).s(" "))
```
allanrenucci: Add a test for:
```scala
def myStringContext= { println("Side effect!"); StringContext }
println(myStringContext(foo, bar).s(" ")) // this shouldn't be optimised away
```
I am not sure it is handled by your implementation

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -583,6 +583,16 @@ class Definitions {
...
 
+  lazy val StringContextType: TypeRef       = ctx.requiredClassRef("scala.StringContext")
+  def StringContextClass(implicit ctx: Context) = StringContextType.symbol.asClass
+    lazy val StringContextSR = StringContextClass.requiredMethodRef(nme.s)
+    def StringContextS(implicit ctx: Context) = StringContextSR.symbol
+    lazy val StringContextRawR = StringContextClass.requiredMethodRef(nme.raw_)
+    def StringContextRaw(implicit ctx: Context) = StringContextRawR.symbol
+  def StringContextModule(implicit ctx: Context) = StringContextClass.companionModule
+    lazy val StringContextModule_applyR = StringContextModule.requiredMethodRef(nme.apply)
+    def StringContextModule_apply(implicit ctx: Context) = StringContextModule_applyR.symbol
```
allanrenucci: @smarter Can you comment on this? Should we expose only the ones used in the optimisation (i.e. `StringContextRaw`, `StringContextS`, `StringContextModule_apply`)? Should these values be inlined in the mini phase and be initialised in `prepareForUnit`?

tests/run/interpolation-opt.scala
Wojtechnology: Added the test and looks like it handles it correctly.

allanrenucci: Sorry, my example was erroneous. What I had is mind was:
```scala
println(myStringContext("Foo", "Bar").s(" ")) // this shouldn't be optimised away

```

allanrenucci: This one too, should not be optimised:
```scala
println({ println("Side effect n2!"); StringContext }.apply("Titi", "Toto").s(" ")) // this shouldn't be optimised away
```

Wojtechnology: Added tests and a fix.

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -32,15 +32,29 @@ class StringInterpolatorOpt extends MiniPhase {
     }
   }
 
+  private object StringContextIdent {
+    def unapply(tree: Ident)(implicit ctx: Context): Option[Ident] = {
+      if (tree.symbol.eq(defn.StringContextModule)) Some(tree) else None
+    }
+  }
```
allanrenucci: Inline this extractor into `StringContextApply`

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -32,15 +32,29 @@ class StringInterpolatorOpt extends MiniPhase {
...
+
+  private object StringContextApply {
```
allanrenucci: I see that this extractor is just a test and does not return any result (but the tree itself). You can return a `Boolean` instead:
```scala
private object StringContextApply {
  def unapply(tree: Select)(implicit ctx: Context): Boolean =
    tree.symbol.eq(defn.StringContextModule_apply) && {
      val qualifier = tree.qualifier
      qualifier.isInstanceOf[Ident] && qualifier.symbol.eq(efn.StringContextModule)
    }
}
```

allanrenucci:
test performance please

OlivierBlanvillain:
@Wojtechnology thanks!

PR #3967, opened 2018-02-05
Restart presentation compilers if memory is low

odersky:
I have noted during long editing sessions (lasting several days,
typically) that memory can get full because the Dotty compiler has some
space leaks. The leaks look really hard to fix, and we don't know
yet whether it's at all possible. To mitigate the leaks, this commit
makes the language server watch available memory, and, if it
is low (i.e. free memory after a GC < 10% of maximal available)
restart all interactive drivers. This will free all memory of
the compiler(s) except the shared nametable.

There's a stressTest option in `Memory.scala`, which, when turned on,
causes a restart every 10 editing actions. I verified that the compiler
stays functional and reasonably responsive in that mode.

Based on #3960

### DIFF
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -25,6 +25,7 @@ import reporting._, reporting.diagnostic.MessageContainer
 import util._
 import interactive._, interactive.InteractiveDriver._
 import Interactive.Include
+import config.Printers.interactiv
 
 import languageserver.config.ProjectConfig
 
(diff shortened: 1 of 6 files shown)


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -68,12 +69,40 @@ class DottyLanguageServer extends LanguageServer
...
+        // new Object() necessary or we get a BootstrapMethodError:
+        //
+        // Caused by: java.lang.invoke.LambdaConversionException: Type mismatch for lambda expected return: void is not convertible to class java.lang.Object
+        // at java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:286)
+        // at java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:303)
+        // at java.lang.invoke.CallSite.makeSite(CallSite.java:302)
+        // ... 11 more
+        //
+        // This looks like a problem with Dottys code generation for void-returning closures passed
+        // to Java methods. (or SAM functions in general?)
```
smarter: Hmm we should be returning a BoxedUnit here, not sure could you open an issue so we can keep track of this?

smarter:
Rebased.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
odersky: See #3984

odersky:
rebased

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
odersky: Yes, you are right.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -173,6 +204,7 @@ class DottyLanguageServer extends LanguageServer
   }
 
   override def didChange(params: DidChangeTextDocumentParams): Unit = thisServer.synchronized {
+    checkMemory()
```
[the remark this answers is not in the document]
odersky: Since they do not involve new runs, I don't think it's a problem.  The leak comes from the memory of previous runs not being completely reclaimed.

odersky:
Merging this now since I am losing too much productivity running into the problem this solves.

PR #3970, opened 2018-02-06
Fix #3965: Correct equality for higher-kinded types

### LINKED ISSUE #3965: Regression in community build: collection-strawman
https://travis-ci.org/lampepfl/dotty-community-build/jobs/336895282

odersky:
What happened was that a higher-kinded type was added repeatedly
to a constraint. Because equality of higher-kinded types was "broken",
the compiler did not realize that the type had already been added.
"Broken" means: Two isomorphic instances of the same higher kinded type
such as `[X] => C[X]` were treated as different. 

For similar reasons the equality check in monitoredSubtype did not
detect loops involving higher-kinded types.

To fix the issue, rework hashing and equality so that two isomorphic types
are identified even if they are dependent (i.e. have back edges from a
`BoundType` such as `ParamRef` or `RecThis` to its `LambdaType` or `RecType`
binder. 

While we are at it, also use this for hash-consing method and poly types.

Aside: `scalac` also uses the naive version of type equality that treats `PolyTypes`
as generative. But it does not have the same issues because it does not use `PolyTypes`
as higher-kinded types. I believe similar issues could be engineered by making up
type lambdas using refinement types and putting these in certain F-bounds. But so far nobody 
seems to have pushed things far enough to notice.


odersky:
test performance please

odersky:
test performance please

compiler/src/dotty/tools/dotc/core/StructEquality.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def equals(tp1: Type, tp2: Any): Boolean = (tp1 `eq` tp2.asInstanceOf[AnyRef]) || tp1.iso(tp2, this)
+
+  @tailrec final def equals(tps1: List[Type], tps2: List[Type]): Boolean =
+    (tps1 `eq` tps2) || {
+      if (tps1.isEmpty) tps2.isEmpty
+      else tps2.nonEmpty && equals(tps1.head, tps2.head) && equals(tps1.tail, tps2.tail)
+    }
+
+  final def equalBinders(tp1: BindingType, tp2: BindingType): Boolean =
```
smarter: This method is unused

odersky: No, there are two calls in `Types`.

odersky:
test performance please

odersky:
Subsumed by #3978

PR #3972, opened 2018-02-07
Fix #3914: Add compiler args to quote run and show Settings

### LINKED ISSUE #3914: Pass any compiler flag to the `QuoteDriver` for debugging
We need to add a way to pass any compiler flag to `QuoteDriver.run` and `QuoteDriver.show`. We might have it as a field`compilerArgs: List[String]` in `Runners.RunSettings`.

This is important when a new bug is found in a project that uses quotes.

nicolasstucki:


nicolasstucki:
This will be useful for debugging.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -11,12 +11,13 @@ import scala.quoted.Expr
...
+import Runners.{Settings, Run, Show}
+
 class QuoteDriver extends Driver {
   import tpd._
 
-  def run[T](expr: Expr[T], settings: Runners.RunSettings): T = {
-    val ctx: Context = initCtx.fresh
-    ctx.settings.optimise.update(settings.optimise)(ctx)
+  def run[T](expr: Expr[T], settings: Settings[Run]): T = {
+    val (_, ctx: Context) = setup(settings.compilerArgs.toArray :+ "dummy.scala", initCtx.fresh)
```
liufengyun: `dummy.scala`?

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -39,25 +40,25 @@ class QuoteDriver extends Driver {
...
     }
-    withTree(expr, show)
+    withTree(expr, show, settings)
   }
 
-  def withTree[T](expr: Expr[_], f: (Tree, Context) => T): T = {
-    val ctx: Context = initCtx.fresh
-    ctx.settings.color.update("never")(ctx) // TODO support colored show
+  def withTree[T](expr: Expr[_], f: (Tree, Context) => T, settings: Settings[_]): T = {
+    val (_, ctx: Context) = setup(settings.compilerArgs.toArray :+ "dummy.scala", initCtx.fresh)
```
liufengyun: what's `dummy.scala`?

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
nicolasstucki: `setup` expects some input file. But it is not required here. Hence I add a dummy file name that is then ignored. It is the `_` in `val (_, ctx: Context) = `

PR #3973, opened 2018-02-07
Fix test log folder location

allanrenucci:
After #3926, test logs were outputted to the dotty project parent folder


OlivierBlanvillain:
:+1:

PR #3974, opened 2018-02-07
Use WildcardType in place of a type variable in `pretypeArgs`

allanrenucci:
As discussed during the meeting. Let's see if it passes the tests

### DIFF
--- a/tests/pos/overloaded.scala
+++ b/tests/pos/overloaded.scala
@@ -29,6 +29,11 @@ object overloaded {
   val r2 = map(x => x.toInt)
   val t2: Seq[Int] = r2
 
+  val rp1 = map { case x => x.toUpper }
+  val tp1: String = rp1
+  val rp2 = map { case x => x.toInt }
+  val tp2: Seq[Int] = rp2
+
   def flatMap(f: Char => String): String = ???
   def flatMap[U](f: Char => Seq[U]): Seq[U] = ???
   val r3 = flatMap(x => x.toString)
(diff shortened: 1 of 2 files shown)


tests/pos/overloaded.scala
```diff
@@ -29,22 +29,40 @@ object overloaded {
...
+  bar { case x => () }
 
   def combine(f: (Char, Int) => Int): Int = ???
   def combine(f: (String, Int) => String): String = ???
   val r5 = combine((x: Char, y) => x + y)
   val t5: Int = r5
   val r6 = combine((x: String, y) => x ++ y.toString)
   val t6: String = r6
+
+  // Errors: The argument types of an anonymous function must be fully known
```
allanrenucci: @odersky Is this intended? Should I open an issue for it?

odersky: I think it's fine; we don't look inside tuples here. What does `scalac` say?

allanrenucci: Tested on Scala 2.13.0-M2:
```scala
[error] constructor cannot be instantiated to expected type;
[error]  found   : (T1, T2)
[error]  required: Char
[error]   bar { case (x, y) => () }
[error]              ^

[warn] match may not be exhaustive.
[warn] It would fail on the following input: (_, _)
[warn]   val rp5 = combine { case (x: Char, y) => x + y }
[warn]                     ^

[warn] match may not be exhaustive.
[warn] It would fail on the following input: (_, _)
[warn]   val rp6 = combine { case (x: String, y) => x ++ y.toString }
[warn]                     ^
[warn] two warnings found
```
In two out of tree of the cases scalac is able to find the correct overload. It emits a spurious exhaustivity warning though

PR #3978, opened 2018-02-09
Fix #3965: Make higher-kinded equality correct and efficient

### LINKED ISSUE #3965: Regression in community build: collection-strawman

odersky:
#3970 fixed equality for higher-kinded types but came at a steep performance penalty. This PR is trying to avoid the penalty. 

Investigation showed that the main culprit was that equality reverted to `eq` for some ProtoType case classes that inherit from type. Since the multiple copies of these case class instances all had the same hashcode, hashtables saw a significant drop in performance. This shows the danger of redefining equality without also redefining hashcode at the same time. We now dropped the shortcut of having a global `equals` in Type, preferring to overwrite `equals` in each subclass where necessary.

To make detection of such issues easier in the future e9388bc adds the numbers of hash-cons table accesses and collisions in the statistics output.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -2663,7 +2663,6 @@ object Types {
     final def isHigherKinded = isInstanceOf[TypeProxy]
 
     private[this] var myParamRefs: List[ParamRefType] = null
-    private[this] var myStableHash: Byte = 0
 
     def paramRefs: List[ParamRefType] = {
       if (myParamRefs == null) myParamRefs = paramNames.indices.toList.map(newParamRef)
@@ -2696,20 +2695,24 @@ object Types {
           x => paramInfos.mapConserve(_.subst(this, x).asInstanceOf[PInfo]),
           x => resType.subst(this, x))
 
+    protected def prefixString: String
+    final override def toString = s"$prefixString($paramNames, $paramInfos, $resType)"
+  }
+
+  abstract class HKLambda extends CachedProxyType with LambdaType {
+    final override def underlying(implicit ctx: Context) = resType
+
     override def computeHash(bs: Binders) =
       doHash(new Binders(this, bs), paramNames, resType, paramInfos)
 
-    override def stableHash = {
-      if (myStableHash == 0) myStableHash = if (resType.stableHash && paramInfos.stableHash) 1 else -1
-      myStableHash > 0
-    }
+    override def stableHash = resType.stableHash && paramInfos.stableHash
 
     final override def equals(that: Any) = equals(that, null)
 
     // No definition of `eql` --> fall back on equals, which calls iso
 
     final override def iso(that: Any, bs: BinderPairs) = that match {
-      case that: LambdaType =>
+      case that: HKLambda =>
         paramNames.eqElements(that.paramNames) &&
         companion.eq(that.companion) && {
           val bs1 = new BinderPairs(this, that, bs)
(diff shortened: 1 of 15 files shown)


odersky:
Test performance please

liufengyun:
test performance please

odersky:
Test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
It seems that caching MethodTypes and PolyTypes is not worth it, so we'll revert the last commit.

odersky:
test performance please

liufengyun:
Performance test finished successfully:

Visit http://dotty-bench.epfl.ch/3978/ to see the changes.

Benchmarks is based on merging with master (5943331)

odersky:
OK, I think performance is roughly on par with what it was before and I don't see any immediate ways to improve. So, ready for review.

smarter:
> Reduce number of calls to Type#equals 

Is this useful at all given that it just forwards to referential equality?

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -397,9 +397,20 @@ object ProtoTypes {
...
       }
 
-    val added =
-      if (state.constraint contains tl) tl.newLikeThis(tl.paramNames, tl.paramInfos, tl.resultType)
+    /** Ensure that `tl` is not already in constraint, make a copy of necessary */
+    def ensureFresh(tl: TypeLambda): TypeLambda =
+      if (state.constraint contains tl) {
+      	var paramInfos = tl.paramInfos
+      	if (tl.isInstanceOf[HKLambda]) {
+      	  // HKLambdas care hash-consed, need to create an artificial difference by adding
```
smarter: typo: care -> are

smarter: Would something break if we did `tl.clone()` instead to get a different instance?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1394,15 +1394,30 @@ object Types {
      */
     def simplified(implicit ctx: Context) = ctx.simplify(this, null)
 
+    final def equals(that: Any, bs: BinderPairs): Boolean =
+      (this `eq` that.asInstanceOf[AnyRef]) || this.iso(that, bs)
```
smarter: I'm not completely convinced that the default `equals` should call `iso`. Why not explicitly use `iso` in the few cases where we need structural equality (monitored subtyping checks, constraint handing, anything else ?)

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1507,10 +1522,27 @@ object Types {
...
+     *  Otherise the standard identity hash.
+     */
+    override def identityHash(bs: Binders) = {
+       def recur(n: Int, tp: BindingType, rest: Binders): Int =
+        if (this `eq` tp) finishHash(hashing.mix(hashSeed, n), 1)
+        else if (rest == null) System.identityHashCode(this)
+        else recur(n + 1, rest.tp, rest.next)
+      avoidSpecialHashes(
+        if (bs == null) System.identityHashCode(this)
+        else recur(1, bs.tp, bs.next))
```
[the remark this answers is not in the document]
odersky: I just don't know how a 0 behaves for hashing.mix

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -2774,7 +2835,7 @@ object Types {
   abstract case class MethodType(paramNames: List[TermName])(
       paramInfosExp: MethodType => List[Type],
       resultTypeExp: MethodType => Type)
-    extends CachedGroundType with MethodOrPoly with TermLambda with NarrowCached { thisMethodType =>
+    extends MethodOrPoly with TermLambda with NarrowCached { thisMethodType =>
```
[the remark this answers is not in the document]
odersky: Good point; we need one since MethodTypes can appear in refinements.

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
odersky: Yes, they would be assumed as `equals`, so e.g. pendingSubtypes would not distinguish between them.

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: It's just a shorthand to avoid the repeated `eq` tests everywhere it is called.

PR #3981, opened 2018-02-11
Fix #3945: Fix eta expansion for partially applied methods

### LINKED ISSUE #3945: Fail to infer parameter types of eta-expanded anonymous functions
In the following example I compose ```int``` with ```int```. 

```scala
object Test {
  def int[A](k: String => A)(s: String)(x: Int): A = ???

  // composing directly: ok in scalac, error in dotc
  val c: (String => String) => (String) => (Int) => (Int) => String = (int[Int => String] _).compose(int[String] _) 

  // unwrapping composition: ok in scalac, ok in dotc
  val q: (String => Int => String) => (String) => (Int) => (Int => String) = int[Int => String]
  val p: (String => String) => (String) => (Int) => String = int[String]
  val c2: (String => String) => (String) => (Int) => (Int) => String = q.compose(p) 
}
```

Doing it directly raises an inference error for both parameters of the anonymous function in dotty with the following error:

```
error missing parameter type

The argument types of an anonymous function must be fully known. (SLS 8.5)
Expected type: class dotty.tools.dotc.typer.ProtoTypes$AnyFunctionProto$
Missing type for parameter s
```

odersky:
This was a tricky one to fix.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3286,7 +3286,7 @@ object Types {
    *  `owningTree` and `owner` are used to determine whether a type-variable can be instantiated
    *  at some given point. See `Inferencing#interpolateUndetVars`.
    */
-  final class TypeVar(val origin: TypeParamRef, creatorState: TyperState, val bindingTree: untpd.Tree, val owner: Symbol) extends CachedProxyType with ValueType {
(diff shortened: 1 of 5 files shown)


odersky:
Test output for failed tests is missing. I just see `exit code 1`.

allanrenucci:
>Test output for failed tests is missing. I just see exit code 1.

Looking into it

PR #3982, opened 2018-02-11
Fix #3976: Fix Eq method for enums with higher-kinded parameters

### LINKED ISSUE #3976: Can't define an enum with higher kinded types
I get an error in Dotty 0.6.0-RC1.

```scala
scala> enum class Hoge[F[_]]
1 |enum class Hoge[F[_]]
  |^
  |Type argument F$1 has not the same kind as its bound
```

I guessed that it failed to derive an Eq instance.

odersky:


compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -341,8 +341,21 @@ object desugar {
       (if (args.isEmpty) tycon else AppliedTypeTree(tycon, args))
         .withPos(cdef.pos.startPos)
 
-    def appliedRef(tycon: Tree, tparams: List[TypeDef] = constrTparams) =
-      appliedTypeTree(tycon, tparams map refOfDef)
+    def appliedRef(tycon: Tree, tparams: List[TypeDef] = constrTparams, widenHK: Boolean = false) = {
+      var targs = for (tparam <- tparams) yield {
```
smarter: var -> val

tests/pos/i3976.scala
```diff
@@ -0,0 +1 @@
+enum class Hoge[F[_]]
```
smarter: I would add a test with an `M[_[_]]` param, and perhaps some code that exercises the generated Eq implicits too to make sure it works as expected.

PR #3986, opened 2018-02-12
Remove AndOrType

OlivierBlanvillain:
This PR remove AndOrType to help the JVM's inliner (+2 other small cleanups based on the same data).

The pattern implemented by `AndOrType` is very problematic for the inliner as the type profiles would be around 50/50 between AndType and OrType, meaning that there is no way to inline further. IIUC, if the ratio goes above 95/5, there the JVM will inline for the common 95% cases and deopt in the other 5% cases.

To get that info from the VM I bootstrapped Dotty with `-XX:+UnlockDiagnosticVMOptions -XX:+PrintInlining`, then grepped for `not inlineable`. [Here is the full log](https://gist.github.com/OlivierBlanvillain/a0b82a8f98ac28ff4a7699b408b403de). Note that each of these are on the hot path, so there should be some performance to be gained in each of them. All the `_1` & `_2` on types look like other low hanging fruits.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/MarkupParsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/MarkupParsers.scala
@@ -314,22 +314,6 @@ object MarkupParsers {
       done
     }
 
-    /** Some try/catch/finally logic used by xLiteral and xLiteralPattern.  */
-    private def xLiteralCommon(f: () => Tree, ifTruncated: String => Unit): Tree = {
-      try return f()
-      catch {
-        case c @ TruncatedXMLControl  =>
-          ifTruncated(c.getMessage)
-        case c @ (MissingEndTagControl | ConfusedAboutBracesControl) =>
-          parser.syntaxError(c.getMessage + debugLastElem + ">", debugLastPos)
-        case _: ArrayIndexOutOfBoundsException =>
-          parser.syntaxError("missing end tag in XML literal for <%s>" format debugLastElem, debugLastPos)
-      }
-      finally parser.in resume Tokens.XMLSTART
-
-      parser.errorTermTree
-    }
-
     /** Use a lookahead parser to run speculative body, and return the first char afterward. */
     private def charComingAfter(body: => Unit): Char = {
       try {
@@ -343,8 +327,8 @@ object MarkupParsers {
     /** xLiteral = element { element }
      *  @return Scala representation of this xml literal
      */
-    def xLiteral: Tree = xLiteralCommon(
-      () => {
+    def xLiteral: Tree = {
+      try return {
         input = parser.in
         handle.isPattern = false
 
@@ -367,15 +351,24 @@ object MarkupParsers {
           assert(ts.length == 1)
           ts(0)
         }
-      },
-      msg => parser.incompleteInputError(msg)
-    )
+      } catch {
+        case c @ TruncatedXMLControl  =>
+        parser.incompleteInputError(c.getMessage)
+        case c @ (MissingEndTagControl | ConfusedAboutBracesControl) =>
+          parser.syntaxError(c.getMessage + debugLastElem + ">", debugLastPos)
+        case _: ArrayIndexOutOfBoundsException =>
+          parser.syntaxError("missing end tag in XML literal for <%s>" format debugLastElem, debugLastPos)
+      }
+      finally parser.in resume Tokens.XMLSTART
+
+      parser.errorTermTree
+    }
 
     /** @see xmlPattern. resynchronizes after successful parse
      *  @return this xml pattern
      */
-    def xLiteralPattern: Tree = xLiteralCommon(
-      () => {
+    def xLiteralPattern: Tree = {
+      try return {
         input = parser.in
         saving[Boolean, Tree](handle.isPattern, handle.isPattern = _) {
           handle.isPattern = true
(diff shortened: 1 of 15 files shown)


OlivierBlanvillain:
test performance please

compiler/src/dotty/tools/dotc/parsing/MarkupParsers.scala
```diff
@@ -314,22 +314,6 @@ object MarkupParsers {
       done
     }
 
-    /** Some try/catch/finally logic used by xLiteral and xLiteralPattern.  */
-    private def xLiteralCommon(f: () => Tree, ifTruncated: String => Unit): Tree = {
```
smarter: Why not use `inline def`?

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1717,58 +1717,72 @@ class Typer extends Namer
...
-          case tree: untpd.TypeTree => typedTypeTree(tree, pt)
+          case tree: untpd.Apply             => if (ctx.mode is Mode.Pattern) typedUnApply(tree, pt) else typedApply(tree, pt)
+          case tree: untpd.This              => typedThis(tree)
+          case tree: untpd.Literal           => typedLiteral(tree)
+          case tree: untpd.New               => typedNew(tree, pt)
+          case tree: untpd.Typed             => typedTyped(tree, pt)
+          case tree: untpd.NamedArg          => typedNamedArg(tree, pt)
+          case tree: untpd.Assign            => typedAssign(tree, pt)
+          case tree: untpd.Block             => typedBlock(desugar.block(tree), pt)(ctx.fresh.setNewScope)
+          case _ => typedUnnamed1(tree)
```
smarter: I doubt this kind of "outlining" is useful since I expect that the JIT will just inline it right back.

OlivierBlanvillain: I pushed the last commit (doing the outlining stuff) by mistake, I'm started some benchmarks locally to see how this helps...

OlivierBlanvillain:
test performance please

compiler/src/dotty/tools/dotc/parsing/MarkupParsers.scala
OlivierBlanvillain: It breaks the pickling tests :/
I will try to minimize into an issue

allanrenucci:
```scala
class Foo {
  inline def foo: Int = try { 1 } finally println("Hello")
  foo
}
```
```
Exception in thread "main" class dotty.tools.dotc.reporting.diagnostic.messages$Error at ?: pickling difference for class Foo in tests/allan/Test.scala, for details:
                   |
                   |  diff before-pickling.txt after-pickling.txt
	at dotty.tools.dotc.reporting.Reporting.error(Reporter.scala:88)
	at dotty.tools.dotc.transform.Pickler.testSame(Pickler.scala:107)
	at dotty.tools.dotc.transform.Pickler.testUnpickler$$anonfun$4(Pickler.scala:99)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:18)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:789)
	at scala.collection.mutable.HashMap.$anonfun$foreach$1(HashMap.scala:138)
	at scala.collection.mutable.HashTable.foreachEntry(HashTable.scala:236)
	at scala.collection.mutable.HashTable.foreachEntry$(HashTable.scala:229)
	at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:40)
	at scala.collection.mutable.HashMap.foreach(HashMap.scala:138)
	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:788)
	at dotty.tools.dotc.transform.Pickler.testUnpickler(Pickler.scala:100)
	at dotty.tools.dotc.transform.Pickler.runOn(Pickler.scala:83)
	at dotty.tools.dotc.Run.runPhases$4$$anonfun$4(Run.scala:123)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:18)
	at scala.compat.java8.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:32)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:29)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:191)
	at dotty.tools.dotc.Run.runPhases$5(Run.scala:136)
	at dotty.tools.dotc.Run.compileUnits$$anonfun$1(Run.scala:141)
	at scala.compat.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at dotty.tools.dotc.util.Stats$.maybeMonitored(Stats.scala:88)
	at dotty.tools.dotc.Run.compileUnits(Run.scala:143)
	at dotty.tools.dotc.Run.compileSources(Run.scala:94)
	at dotty.tools.dotc.Run.compile(Run.scala:78)
	at dotty.tools.dotc.Driver.doCompile(Driver.scala:29)
	at dotty.tools.dotc.Driver.process(Driver.scala:127)
	at dotty.tools.dotc.Driver.process(Driver.scala:96)
	at dotty.tools.dotc.Driver.process(Driver.scala:108)
	at dotty.tools.dotc.Driver.main(Driver.scala:135)
	at dotty.tools.dotc.Main.main(Main.scala
```

OlivierBlanvillain:
@odersky on master, [this large pattern matching](https://github.com/lampepfl/dotty/pull/3986/files#diff-e6057cacdd6d97cf1079e08b9f18db66L30) is executed in the constructor of every single Constant (the JIT also complains that it's also too large to be inlined). With this refactoring, if the type of the constant is known on the call site we avoid doing that work by directly putting the correct tag:

```scala
object Constant {
    def apply(x: Unit) = new Constant(x, UnitTag)
}
```

Note that the "dynamic" version is still there as a `def apply(x: Any): Constant`, IIRC it's only used in one of two places.

compiler/src/dotty/tools/dotc/parsing/MarkupParsers.scala
allanrenucci: This has been fixed. Should revert this change?

PR #3990, opened 2018-02-12
Show GADT bounds in type mismatch message

smarter:
The inferred GADT bounds can be non-trivial so it makes sense to display
them.

This commit also showcases the now-improved GADT support in Dotty (both
tests/pos/gadt-eval.scala and tests/neg/gadt-eval.scala used to compile,
and both compile in Scala 2).

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/Formatting.scala
+++ b/compiler/src/dotty/tools/dotc/printing/Formatting.scala
@@ -168,7 +168,12 @@ object Formatting {
       case param: TypeParamRef =>
         s"is a type variable${addendum("constraint", ctx.typeComparer.bounds(param))}"
       case sym: Symbol =>
-        s"is a ${ctx.printer.kindString(sym)}${sym.showExtendedLocation}${addendum("bounds", sym.info)}"
+        val info =
+          if (ctx.gadt.bounds.contains(sym))
+            sym.info & ctx.gadt.bounds(sym)
+          else
+            sym.info
+        s"is a ${ctx.printer.kindString(sym)}${sym.showExtendedLocation}${addendum("bounds", info)}"
       case tp: SkolemType =>
         s"is an unknown value of type ${tp.widen.show}"
     }
@@ -183,7 +188,11 @@ object Formatting {
     def needsExplanation(entry: Recorded) = entry match {
       case param: TypeParamRef => ctx.typerState.constraint.contains(param)
       case skolem: SkolemType => true
-      case _ => false
+      case sym: Symbol =>
+        ctx.gadt.bounds.contains(sym) && ctx.gadt.bounds(sym) != TypeBounds.empty
+      case _ =>
+        assert(false, "unreachable")
+        false
     }
 
     val toExplain: List[(String, Recorded)] = seen.toList.flatMap {
--- a/tests/neg/gadt-eval.scala
+++ b/tests/neg/gadt-eval.scala
@@ -1,0 +1,33 @@
+sealed trait Exp[T]
+case class Lit(value: Int) extends Exp[Int]
+case class Pair[A, B](fst: Exp[A], snd: Exp[B]) extends Exp[(A, B)]
+
+object Test {
+  def eval[T](e: Exp[T]): T = e match {
+    case Lit(x) =>
+      x
+    case Pair(a, b) =>
+      (eval(a), eval(a)) // error:
+        // -- [E007] Type Mismatch Error: tests/neg/gadt-eval.scala:10:6 ------------------
+        // 10 |      (eval(a), eval(a))
+        // |      ^^^^^^^^^^^^^^^^^^
+        // |    found:    (_$1, _$1)
+        // |    required: T
+        // |
+        // |    where:    T is a type in method eval which is an alias of (_$1, _$2)
+  }
+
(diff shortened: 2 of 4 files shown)


tests/neg/gadt-eval.scala
```diff
@@ -0,0 +1,33 @@
...
+      x
+    case Pair(a, b) =>
+      (eval(a), eval(a))
+        // -- [E007] Type Mismatch Error: tests/neg/gadt-eval.scala:10:6 ------------------
+        // 10 |      (eval(a), eval(a))
+        // |      ^^^^^^^^^^^^^^^^^^
+        // |    found:    (_$1, _$1)
+        // |    required: T
+        // |
+        // |    where:    T is a type in method eval which is an alias of (_$1, _$2)
```
Blaisorblade: This overall testcase is cool!

Wondering how to explain where these type variables come from. They come from the match (hard to show), and they show up in the types of a and b, which is easier to show, most easily by showing the whole typing context. That might be too much, but I think GHC sometimes does it, and both Agda and Coq absolutely do it and have to do it. Not sure where to draw a line.

smarter: Yeah, if you can think of a good way to display this information I'm interested, I haven't checked what other compilers do.

compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -183,7 +188,10 @@ object Formatting {
     def needsExplanation(entry: Recorded) = entry match {
       case param: TypeParamRef => ctx.typerState.constraint.contains(param)
       case skolem: SkolemType => true
-      case _ => false
+      case sym: Symbol =>
+        ctx.gadt.bounds.contains(sym) && ctx.gadt.bounds(sym) != TypeBounds.empty
+      case _ =>
+        false
```
allanrenucci: `assert(false, "unreachable")`?

tests/neg/gadt-eval.scala
odersky: I don't get it. If "T is a type in method eval which is an alias of (_$1, _$2)", why is there a type error?

allanrenucci:
```
found: (_$1, _$1) 
expected: (_$1, _$2) 
```
The second type parameter of the tuple differs

compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -168,7 +168,12 @@ object Formatting {
       case param: TypeParamRef =>
         s"is a type variable${addendum("constraint", ctx.typeComparer.bounds(param))}"
       case sym: Symbol =>
-        s"is a ${ctx.printer.kindString(sym)}${sym.showExtendedLocation}${addendum("bounds", sym.info)}"
+        val info =
+          if (ctx.gadt.bounds.contains(sym))
+            sym.info & ctx.gadt.bounds(sym)
+          else
+            sym.info
```
allanrenucci: micro optimisation:
```scala
val bounds = ctx.gadt.bounds(sym)
val info = if (bounds != null) sym.info & bounds else sym.info
```

compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -183,7 +188,11 @@ object Formatting {
     def needsExplanation(entry: Recorded) = entry match {
       case param: TypeParamRef => ctx.typerState.constraint.contains(param)
       case skolem: SkolemType => true
-      case _ => false
+      case sym: Symbol =>
+        ctx.gadt.bounds.contains(sym) && ctx.gadt.bounds(sym) != TypeBounds.empty
```
allanrenucci: micro optimisation:
```scala
ctx.gadt.bounds(sym) != TypeBounds.empty // since it returns null if not in the gadt bounds map
```

compiler/src/dotty/tools/dotc/printing/Formatting.scala
```diff
@@ -168,7 +168,12 @@ object Formatting {
       case param: TypeParamRef =>
         s"is a type variable${addendum("constraint", ctx.typeComparer.bounds(param))}"
       case sym: Symbol =>
-        s"is a ${ctx.printer.kindString(sym)}${sym.showExtendedLocation}${addendum("bounds", sym.info)}"
+        val info =
+          if (ctx.gadt.bounds.contains(sym))
+            sym.info & ctx.gadt.bounds(sym)
```
allanrenucci: Do you need to do the intersection? Can you simply return `ctx.gadt.bounds(sym)`?

smarter: I think so, we used to do something similar in PostTyper.

compiler/src/dotty/tools/dotc/printing/Formatting.scala
smarter: we would also have to check != null

compiler/src/dotty/tools/dotc/printing/Formatting.scala
smarter: I don't think micro-optimizing error messages is worth it.

PR #3991, opened 2018-02-12
ClassfileParser: skip synthetic methods

smarter:
When compiling re2s, this reduces the number of symbols from 24663 to
24515 and the number of denotations from 216669 to 195374.


smarter:
test performance please

smarter:
Turns out we cannot do that, because when we lookup class members, we may get a member from a parent class, and this member may have the Deferred flag set, thus breaking stuff (here the test failures a re due to the wrong method signature being used in a lambdametafactory call because we incorrectly find two abstract methods in an interface).

Issue #3989, opened 2018-02-12
Unsoundness when inheriting conflicting type parameters


szeiger:
This compiles without warnings in both Dotty master and Scala 2.12:

```scala
trait A[+X]
case class B[X](x: X) extends A[X]
class C[X](x: Any) extends B[Any](x) with A[X]
def f(a: A[Int]): Int = a match {
  case B(i) => i
  case _ => 0
}
f(new C[Int]("foo"))
```

There used to be unsoundness warnings but they are gone since (presumably) https://github.com/lampepfl/dotty/issues/3918 even though the pattern match is unsound.

Furthermore the following compiles without warnings:

```scala
trait A[+X] { def get: X }
case class B[X](x: X) extends A[X] { def get: X = x }
class C[X](x: Any) extends B[Any](x) with A[X]
def g(a: A[Int]): Int = a.get
g(new C[Int]("foo"))
```

In this case, which doesn't rely on pattern matching, the definition of `C` itself is unsound but Dotty accepts it anyway. Scala 2.12 rightfully complains:

```
scala> class C[X](x: Any) extends B[Any](x) with A[X]
<console>:14: error: overriding method get in trait A of type => X;
 method get in class B of type => Any has incompatible type
       class C[X](x: Any) extends B[Any](x) with A[X]
             ^
```

szeiger:
Discussion leading to this ticket: https://gitter.im/lampepfl/dotty?at=5a81a707e217167e2c742a4f

Blaisorblade:
The second one seems GADT-unrelated.

I disagree on “conflicting”: `C` *refines* a covariant type parameter, and I’m aware of no way to get issues from it without pattern matching (in Scalac) — this is an old bug (from my Scala’13 short paper, and there’s some issue for Scalac). As long as defining `C` is allowed, that pattern match (more precisely, the type refinement in that pattern match) should be forbidden  — whether the compiler sees `C` or not (since it can be added later — unless there’s `sealed`).

Maybe defining `C` should indeed be forbidden but it’s not clear — that idea was ultimately rejected in the Scalac tracker (though I’m not sure the rationale for it was too compelling).
Maybe one could also separate more GADTs and normal classes, even though that’s irregular — I seem to recall Dotty’s sugar for ADTs has some special features anyway.

smarter:
[...]

PR #3999, opened 2018-02-14
add more gadt exhaustivity check tests

liufengyun:
add more gadt exhaustivity check tests

### DIFF
--- a/tests/patmat/gadt2.scala
+++ b/tests/patmat/gadt2.scala
@@ -1,0 +1,16 @@
+sealed trait Nat[+T]
+case class Zero() extends Nat[Nothing]
+case class Succ[T]() extends Nat[T]
+
+// +N is incorrect, as in `foo` we can have `N = Zero | Succ[Zero]`,
+// then it's correct for exhaustivity check to produce two warnings.
+sealed trait Vect[N <: Nat[_], +T]
+case class VN[T]() extends Vect[Zero, T]
+case class VC[T, N <: Nat[_]](x: T, xs: Vect[N, T]) extends Vect[Succ[N], T]
+
+object Test {
+  def foo[N <: Nat[_], A, B](v1: Vect[N, A], v2: Vect[N, B]) = (v1, v2) match {
+    case (VN(), VN()) => 1
+    case (VC(x, xs), VC(y, ys)) => 2
+  }
+}
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -600,38 +600,28 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   def instantiate(tp1: NamedType, tp2: Type)(implicit ctx: Context): Type = {
     // expose abstract type references to their bounds or tvars according to variance
     class AbstractTypeMap(maximize: Boolean)(implicit ctx: Context) extends TypeMap {
-      def expose(tp: TypeRef): Type = {
-        val lo = this(tp.info.loBound)
-        val hi = this(tp.info.hiBound)
-        val exposed =
-          if (variance == 0)
-            newTypeVar(TypeBounds(lo, hi))
-          else if (variance == 1)
-            if (maximize) hi else lo
-          else
-            if (maximize) lo else hi
-
-        debug.println(s"$tp exposed to =====> $exposed")
-        exposed
-      }
+      def expose(lo: Type, hi: Type): Type =
+        if (variance == 0)
+          newTypeVar(TypeBounds(lo, hi))
+        else if (variance == 1)
+          if (maximize) hi else lo
+        else
+          if (maximize) lo else hi
 
       def apply(tp: Type): Type = tp match {
         case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
+          val lo = this(tp.info.loBound)
+          val hi = this(tp.info.hiBound)
           // See tests/patmat/gadt.scala  tests/patmat/exhausting.scala  tests/patmat/t9657.scala
-          expose(tp)
+          val exposed = expose(lo, hi)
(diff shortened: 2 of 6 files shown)


tests/patmat/gadt2.scala
```diff
@@ -0,0 +1,16 @@
+sealed trait Nat[+T]
+case class Zero() extends Nat[Nothing]
+case class Succ[T]() extends Nat[T]
```
Blaisorblade: This is not intended to be a correct representation of Peano numerals, right? Not only for the `+T` but also because `T` in `Succ` is completely unconstrained...

liufengyun:
Yes, good catch, I forgot where I got the test case, I tried adding the
bounds and remove the variance on Nat, it doesn't change the result.

Le 14 févr. 2018 6:07 PM, "Paolo G. Giarrusso" <notifications@github.com> a
écrit :

Blaisorblade:
I should have spelled out the encoding I had in mind. Please take a look especially at the comments.

```scala
object Try1 {
  // type-level naturals
  sealed trait TNat
  sealed trait TZero extends TNat
  sealed trait TSucc[T] extends TNat

  //reflect TNat at the value level; given n: Nat[T], n and T represent the same number.
  //This is what Haskellers call "singleton types" and took from Omega.
  // The point is that we can use such a number both at compile-time and at runtime,
  // but of course the runtime representation is not erased.
  //And this constrains the type argument of `Nat` to be of type `TNat` — though you could add bounds for it.

  sealed trait Nat[+T]
  case class Zero() extends Nat[TZero]
  case class Succ[T](n: Nat[T]) extends Nat[TSucc[T]]

  //We can index Vect with the types of value-level Nat, but this is a bit overkill. Still, no warnings.
  sealed trait Vect[N <: Nat[_], +T]
  case class VN[T]() extends Vect[Zero, T]
  case class VC[T, N <: Nat[_]](x: T, xs: Vect[N, T]) extends Vect[Succ[N], T]

  object Test {
    def foo[N <: Nat[_], A, B](v1: Vect[N, A], v2: Vect[N, B]) =
      (v1, v2) match {
        case (VN(), VN())           => 1
        case (VC(x, xs), VC(y, ys)) => 2
      }
  }
}

//Since we didn't need value-level numbers, let's drop them:
object Try2 {
  sealed trait TNat
  sealed trait TZero extends TNat
  sealed trait TSucc[T] extends TNat

  sealed trait Vect[N <: TNat, +T]
  case class VN[T]() extends Vect[TZero, T]
  case class VC[T, N <: TNat](x: T, xs: Vect[N, T]) extends Vect[TSucc[N], T]

  object Test {
    def foo[N <: TNat, A, B](v1: Vect[N, A], v2: Vect[N, B]) =
      (v1, v2) match {
        case (VN(), VN())           => 1
        case (VC(x, xs), VC(y, ys)) => 2
      }
  }
}

//Same as Try2, but now `Vect` is covariant in `N` so we get the warning you described.
object Try3 {
  sealed trait TNat
  sealed trait TZero extends TNat
  sealed trait TSucc[T] extends TNat

  sealed trait Vect[+N <: TNat, +T]
  case class VN[T]() extends Vect[TZero, T]
  case class VC[T, N <: TNat](x: T, xs: Vect[N, T]) extends Vect[TSucc[N], T]

  object Test {
    def foo[N <: TNat, A, B](v1: Vect[N, A], v2: Vect[N, B]) =
      //Warnings here!
      (v1, v2) match {
        case (VN(), VN())           => 1
        case (VC(x, xs), VC(y, ys)) => 2
      }
  }
}
```

liufengyun:
@Blaisorblade Thanks for the encoding. If I were correct, as long as `N` is covariant in `Vect`, it's possible to have `N = TZero | TSucc[_]` in `foo`, thus it's correct to report the warnings.

Blaisorblade:
@liufengyun IIUC, triggering `N = TZero | TSucc[_]` in `Try3` requires extending `VN` or `VC` and refining the covariant type parameter ? If so, the warning can be avoided if both classes are final; IIRC Scalac manages to figure this out in some cases.

But if we forbid such covariant refinements as @smarter would like (see #3989), the warning becomes unnecessary.
And maybe case classes extending `sealed` should be final automatically, again making this warning unnecessary. In my 2013 paper I claimed I had use cases in SQuOpt for non-final children of sealed traits, I'll have to dig for them and see how crucial they are.

And to be sure: Even though all variants give the same warnings, we might want to add (with more priority) the encodings we want to keep supporting, more than the ones which might or might not work.

liufengyun:
@Blaisorblade Triggering `N = TZero | TSucc[_]` doesn't require extending `VN` and `VC`, it's just normal type application with union types. Note that this behaviour only exists in Dotty because it support union types. In another word, in any theory that supports union types, we can not be sure that the same type parameter `T` in different positions always refer to the same type.

Blaisorblade:
> Triggering N = TZero | TSucc[_] doesn't require extending VN and VC, it's just normal type application with union types.

Ah, do you mean `foo[TZero | TSucc[_], Int, String](VN(), VC("", VN()))`? Ah, the crucial point is you can upcast `Vec`, I see! Very interesting and annoying.

> In another word, in any theory that supports union types, we can not be sure that the same type parameter T in different positions always refer to the same type.

I see what you mean, but a type variable still is bound to *one* type at a time. We just have more type variables: due to covariance the argument for `N` to `foo` need not be the same as the one given to the `Vec` constructor, and union types give a surprising way to exploit that; intersection types and covariant type refinement give another.

Blaisorblade:
@liufengyun Hope adding my commits here is OK

liufengyun:
Sure, thanks a lot @Blaisorblade !

Blaisorblade:
Also assigning myself so I come back to it.

Blaisorblade:
@liufengyun looks good to go, or do you want to add something? Seems mergeable!

liufengyun:
Thanks @Blaisorblade , I've no more things on my side.

PR #4001, opened 2018-02-15
Fix trees not being loaded from tasty when class symbol has NoCompleter

nicolasstucki:


compiler/src/dotty/tools/dotc/core/Symbols.scala
```diff
@@ -638,22 +638,22 @@ object Symbols {
...
-            }
-            else tpd.EmptyTree
-          case tree: Tree @ unchecked =>
-            if (id.isEmpty || mightContain(tree, id)) tree else tpd.EmptyTree
-        }
+    def treeContaining(id: String)(implicit ctx: Context): Tree = {
+      denot.infoOrCompleter match {
+        case _: NoCompleter =>
+        case _ => denot.ensureCompleted()
+      }
```
allanrenucci: Why not simply `denot.ensureCompleted()`? I expect it to be NoOp if  `infoOrCompleter` returns `NoCompleter`.

allanrenucci: Apparently I am wrong

nicolasstucki: Will revert that change

PR #4002, opened 2018-02-15
Harden IDE: Catch NoSuchFileException

odersky:
A NoSuchFileException was observed for a non-existing directory on the classpath:

    /Users/odersky/workspace/dotty/compiler/../out/bootstrap/dotty-compiler-bootstrapped/scala-0.7/classes

It was after a dotty-bootstrapped/clean. I believe this should be ignored, or maybe handled with
an error message. But certainly not a crash.

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -344,7 +344,7 @@ object Interactive {
     if (tree.pos.contains(pos)) {
       // FIXME: We shouldn't need a cast. Change NavigateAST.pathTo to return a List of Tree?
       val path = NavigateAST.pathTo(pos, tree, skipZeroExtent = true).asInstanceOf[List[untpd.Tree]]
-      path.dropWhile(!_.hasType).asInstanceOf[List[tpd.Tree]]
+      path.dropWhile(!_.hasType) collect { case t: tpd.Tree @unchecked => t }
     }
     else Nil
 
@@ -365,7 +365,7 @@ object Interactive {
     case nested :: encl :: rest =>
       import typer.Typer._
       val outer = contextOfPath(encl :: rest)
-      encl match {
+      try encl match {
         case tree @ PackageDef(pkg, stats) =>
           assert(tree.symbol.exists)
           if (nested `eq` pkg) outer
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala
```diff
@@ -148,20 +148,24 @@ class InteractiveDriver(settings: List[String]) extends Driver {
...
-      Files.walkFileTree(root, new SimpleFileVisitor[Path] {
-        override def visitFile(path: Path, attrs: BasicFileAttributes) = {
-          if (!attrs.isDirectory) {
-            val name = path.getFileName.toString
-            for {
-              tastySuffix <- tastySuffixes
-              if name.endsWith(tastySuffix)
-            } {
-              names += root.relativize(path).toString.replace("/", ".").stripSuffix(tastySuffix)
+      try
```
allanrenucci: `if (Files.exists(root)) ...` instead of catching an exception?

sjrd:
The standard behavior is to (silently) ignore non-existent entries in a classpath. Both scalac and the JVM do that, and historically I can tell you that people rely on this behavior because we had a bug report in Scala.js reporting that we did not ignore them: https://github.com/scala-js/scala-js/issues/2198

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -344,7 +344,7 @@ object Interactive {
     if (tree.pos.contains(pos)) {
       // FIXME: We shouldn't need a cast. Change NavigateAST.pathTo to return a List of Tree?
       val path = NavigateAST.pathTo(pos, tree, skipZeroExtent = true).asInstanceOf[List[untpd.Tree]]
-      path.dropWhile(!_.hasType).asInstanceOf[List[tpd.Tree]]
+      path.dropWhile(!_.hasType) collect { case t: tpd.Tree @unchecked => t }
```
allanrenucci: If `path` can contain non-tree elements then `hasType` will also throws. Maybe change `NavigateAST.pathTo` to discard non-tree element and return a `List[Tree]` instead (as suggested in the comment).

odersky: Good point.

odersky: I made the change on the other PR (harden-ide)

PR #4007, opened 2018-02-16
Fix #3797: Add support for @implicitAmbiguous

### LINKED ISSUE #3797: Add support for @implicitAmbiguous to customize the error message when an implicit search finds multiple ambiguous value
Added in Scala 2.12: https://github.com/scala/scala/pull/4673

Jasper-M:
There are no tests yet. Can I add a test to https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/dotc/reporting/ErrorMessagesTests.scala if this error doesn't have a dedicated case class yet?

Dotc will not warn like scalac when the user defined error message references a non-existent type parameter, but that's consistent with @implicitNotFound in dotty.


compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,44 @@ trait Implicits { self: Typer =>
...
+          val params = alt.ref.underlying match {
+            case lambda: TypeLambda => lambda.typeParams.map(_.paramName.unexpandedName.toString)
+            case _                  => Nil
           }
-          yield {
-            err.implicitNotFoundString(
-              raw,
-              pt.typeSymbol.typeParams.map(_.name.unexpandedName.toString),
-              pt.argInfos)
+          alt.tstate.commit()
```
smarter: We should try avoiding committing a TyperState with errors if possible. Instead you might be able to make a new Context with the correct TyperState to report the error: `val altCtx = ctx.fresh.setTyperState(alt.tstate)`, and then move the code below to be in a method that takes an `implicit ctx: Context` and give it `altCtx` as argument.

smarter:
> f this error doesn't have a dedicated case class yet?

I suggest just making a case class for the error, it doesn't have to be very fancy, e.g. it's okay if `def explanation` is left empty.

smarter:
> Dotc will not error out like scalac when the user defined error message references a non-existent type parameter, but that's consistent with @implicitnotfound in dotty.

Could you open an issue for that?

Jasper-M:
> I suggest just making a case class for the error, it doesn't have to be very fancy, e.g. it's okay if def explanation is left empty.

That would be a general `ImplicitNotFound` message then? Untangling the different subtypes of errors in `missingArgMsg` looks difficult to me, especially as I don't really understand the distinction between `case arg: Trees.SearchFailureIdent[_]` and `case _` underneath it.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
Jasper-M: Sounds good.

Jasper-M:
> Could you open an issue for that?

It's actually a warning apparently. But I'll open an issue.

smarter:
> That would be a general ImplicitNotFound message then?

Sure, but you could have an AmbiguousImplicit subclass or something.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,46 @@ trait Implicits { self: Typer =>
         }
     }
     def location(preposition: String) = if (where.isEmpty) "" else s" $preposition $where"
+    def userDefinedMessage(annot: Annotation, params: List[String], args: List[Type]): Option[String] =
+      for(Trees.Literal(Constant(raw: String)) <- annot.argument(0)) yield {
```
smarter: space after `for`

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,46 @@ trait Implicits { self: Typer =>
...
-          yield {
-            err.implicitNotFoundString(
-              raw,
-              pt.typeSymbol.typeParams.map(_.name.unexpandedName.toString),
-              pt.argInfos)
+          val args = alt.tree match {
+            case TypeApply(_, targs) if params.nonEmpty =>
+              def resolveTypes(targs: List[tpd.Tree])(implicit ctx: Context) = {
+                targs.map(a => fullyDefinedType(a.tpe, "type parameter", a.pos))
+              }
```
smarter: No need for braces since the body is  an expression.

Jasper-M: I thought it was better for readability in this case, but your call.

Jasper-M:
I added tests, but I think it's better if I don't add a half assed custom `Message` case class in this pull request. I think it's ready for (final) review.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -159,11 +159,11 @@ object ErrorReporting {
       TypeMismatch(found2, expected2, whyNoMatchStr(found, expected), postScript)
     }
 
-    /** Format `raw` implicitNotFound argument, replacing all
-     *  occurrences of `${X}` where `X` is in `paramNames` with the
+    /** Format `raw` implicitNotFound or implicitAmbiguous argument, replacing
```
Blaisorblade: Looks like this function might be useful for more annotations (@smarter mentioned a couple more ones on Gitter), as suggested by the new name you picked, but not by the comment. What do you think?

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,45 @@ trait Implicits { self: Typer =>
         }
     }
     def location(preposition: String) = if (where.isEmpty) "" else s" $preposition $where"
+    def userDefinedMessage(annot: Annotation, params: List[String], args: List[Type]): Option[String] =
+      for (Trees.Literal(Constant(raw: String)) <- annot.argument(0)) yield {
```
Blaisorblade: Wondering about the magic constant `0`, I guess because the string is the 0th argument for all annotations currently supported? But if you try to make `userDefinedMessage` a good abstraction, *maybe* `annot.argument(0)` should move to the call site (even if, indeed, it gets duplicated).

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
Jasper-M: I sort of agree. I'm open to suggestions 😅

smarter: I think it's not worth overgeneralizing, we can change this when we need to.

Jasper-M: I may have misinterpreted your question a little bit. It is useful for all annotations whose messages can reference type parameters, which could potentially be more than the 2 listed here, but not all annotations carrying a custom message. I'm open to suggestions for a better name.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,45 @@ trait Implicits { self: Typer =>
...
-          for {
-            notFound <- pt.typeSymbol.getAnnotation(defn.ImplicitNotFoundAnnot)
-            Trees.Literal(Constant(raw: String)) <- notFound.argument(0)
+        val maybeAnnot = ambi.alt1.ref.symbol.getAnnotation(defn.ImplicitAmbiguousAnnot).map(
+          (_, ambi.alt1)
+        ).orElse(
+          ambi.alt2.ref.symbol.getAnnotation(defn.ImplicitAmbiguousAnnot).map(
+            (_, ambi.alt2)
+          )
+        )
```
Blaisorblade: Hm, can this be abstracted with `for { alt <- List(ambi.alt1, ambi.alt2) } .... alt.ref.symbol...`? Or better, in terms of a new `ambi.alts`?

(Except if people feel this is too slow and we're in a performance-critical path).

Actually surprised the error only lists two implicits, apparently since this class' birth (in ea640a32264cb78efbf267d5c2be89e3e99dcccf) only the first two candidate implicits are reported.

Potential bug (not blocking the PR): if you have *three* equally good candidates that are ambiguous, the `@implicitAmbiguous` annotation will be ignored on the third candidate, which is against the specification in http://www.scala-lang.org/api/2.12.3/scala/annotation/implicitAmbiguous.html. If that bug can indeed be triggered (haven't tried, but it's not clear why that shouldn't happen), fixing it seems to require a refactoring of the search algorithm to return all the ambiguous candidates — which is, again, *out of scope* for the PR.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
Blaisorblade: Hm, I suspect the name is fine and would write "Format `raw` user-defined error message" in the doc. WDYT? Anyway, this isn't critical, just nice to have. (Warning, I've been told I'm a perfectionist).

compiler/src/dotty/tools/dotc/typer/Implicits.scala
Jasper-M: Yes I also thought about this. I think the Scalac implementation has the same limitation. I suspect the most realistic fix would probably be in the documentation of `@implicitAmbiguous`.

Blaisorblade: Ah, fair enough — this wasn't a well-known issue around here. Not sure how to specify it, 
so for now I've opened https://github.com/scala/bug/issues/10739?

For the PR, please go ahead!

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,45 @@ trait Implicits { self: Typer =>
...
+        val userDefined = maybeAnnot.flatMap { case (annot, alt) =>
+          val params = alt.ref.underlying match {
+            case lambda: TypeLambda => lambda.typeParams.map(_.paramName.unexpandedName.toString)
+            case _                  => Nil
```
Blaisorblade: @allanrenucci Couldn't quite review code from here on, please take a look. Otherwise LGTM.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
allanrenucci: I would do it the other way around. First look for a user defined annotation message and then try try to recover type params only if needed:
```scala
def userDefinedMsg(sym: Symbol, cls: Symbol) = for {
  ann <- symbol.getAnnotation(cls)
  Trees.Literal(Constant(msg: String)) <- ann.argument(0)
} yield msg

object AmbiguousImplicitMsg {
  def unapply(search: SearchSuccess) =
    userDefinedMsg(search.ref.symbol, defn.ImplicitAmbiguousAnnot)
}

(ambi.alt1, ambi.alt2) match {
  case (AmbiguousImplicitMsg(msg), _) => ... // recover type params from alt1
  case (_, AmbiguousImplicitMsg(msg)) => ... // recover type params from alt2
  case _ =>
    msg(s"ambiguous implicit arguments: ${ambi.explanation}${location("of")}")(
        s"ambiguous implicit arguments of type ${pt.show} found${location("for")}")
}
```

Jasper-M:
How are you supposed to handle `LazyTree`s? If I assign them a type ascription in a pattern match (`rhs: Tree`) I get an error because of an unchecked pattern. If I try to match it with a pattern (`Apply(TypeApply(_, targs), _)`) and then do targs.map(_.tpe) I get a `ClassCastException: TypeVar cannot be cast to Null$`, nor can I use targs where a `List[tpd.Tree]` is expected (while it obviously is, hence the exception).

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -696,22 +697,45 @@ trait Implicits { self: Typer =>
...
+            case TypeApply(_, targs) if params.nonEmpty =>
+              def resolveTypes(targs: List[tpd.Tree])(implicit ctx: Context) =
+                targs.map(a => fullyDefinedType(a.tpe, "type parameter", a.pos))
+              resolveTypes(targs)(ctx.fresh.setTyperState(alt.tstate))
+            case _ =>
+              Nil
           }
+          userDefinedMessage(annot, params, args)
+        }
+        userDefined.map(msg(_)()).getOrElse(
```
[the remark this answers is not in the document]
Jasper-M: That's very likely. Is the same true for the `ImplicitNotFound` case?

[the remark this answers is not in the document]
allanrenucci: Actually I am not sure anymore. I revert what I said before. Let's be conservative and also wrap the user defined message in `msg(_)()`

Jasper-M:
As far as I'm concerned this is pretty much ready (except for the cast maybe). Please let me know if there's still something that is blocking.

Blaisorblade:
@Jasper-M what about the CI failure?

Jasper-M:
> what about the CI failure?

Not a clue. My guess is either it's spurious or a rebase or somesuch is required. Someone who understands the CI job and test framework better will have to point me in the right direction.

Blaisorblade:
OK, we’ll take a look on our side. Delays are possible, but we’ll do our best.

Blaisorblade:
(Build restarted, it was weird that only *test* failed.)

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -685,22 +686,54 @@ trait Implicits { self: Typer =>
...
-            notFound <- pt.typeSymbol.getAnnotation(defn.ImplicitNotFoundAnnot)
-            Trees.Literal(Constant(raw: String)) <- notFound.argument(0)
-          }
-          yield {
-            err.implicitNotFoundString(
-              raw,
-              pt.typeSymbol.typeParams.map(_.name.unexpandedName.toString),
-              pt.argInfos)
+        object AmbiguousImplicitMsg {
+          def unapply(search: SearchSuccess) =
```
smarter: Write down the result type of this method for clarity.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -685,22 +686,54 @@ trait Implicits { self: Typer =>
...
+            userDefinedMsg(search.ref.symbol, defn.ImplicitAmbiguousAnnot)
+        }
+
+        def userDefinedAmbiguousImplicitMsg(alt: SearchSuccess, raw: String) = {
```
smarter: Add a documentation comment for this method.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -685,22 +686,54 @@ trait Implicits { self: Typer =>
         }
     }
     def location(preposition: String) = if (where.isEmpty) "" else s" $preposition $where"
+
+    def userDefinedMsg(sym: Symbol, cls: Symbol) = for {
```
smarter: Add a documentation comment for this method.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -685,22 +686,54 @@ trait Implicits { self: Typer =>
...
+          //   implicit def f[A](x: Int): String = ...
+          //   implicitly[Int => String]
+          val call = closureBody(alt.tree) // the tree itself if not a closure
+          val (_, targs, _) = decomposeCall(call)
+          val args = resolveTypes(targs)(ctx.fresh.setTyperState(alt.tstate))
+          err.userDefinedErrorString(raw, params, args)
+        }
+
+        (ambi.alt1, ambi.alt2) match {
+          case (AmbiguousImplicitMsg(msg), _) =>
```
smarter: If you write `alt @ AmbiguousImplicitMsg(msg)`, you can use `alt` below instead of `ambi.alt1`, same in the next case.

allanrenucci:
Merging! Thanks @Jasper-M 🎉

PR #4012, opened 2018-02-19
Fix #3912: Account for units and blocks in inline and macros

### LINKED ISSUE #3912: Macro with Unit type splicing non Unit fails with `splice outside quotes`
```
import scala.quoted._
object Macros {
  inline def foo(): Unit = ~impl()
  def impl(): Expr[Int] = '(1)
}
```
```
import scala.quoted._
import Macros._
class Test {
  val a: Unit = foo()
}
```

fails with 
```
5 |    ~impl()
  |    ^^^^^^^
  |    splice outside quotes
```

because `inline def foo(): Unit = ~impl()` is transformed to 
`inline def foo(): Unit = { ~impl(); () }`, which is correct.

We need a better error message for this.

biboudis:


compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -338,5 +346,20 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+
+    /** InlineSplice is used to detect cases where the expansion
+     *  consists of a (possibly multiple & nested) block or a sole expression.
+     */
+    object InlineSplice {
+      def unapply(tree: Tree)(implicit ctx: Context): Option[Select] = {
+        tree match {
+          case expansion: Select if expansion.symbol.isSplice =>
+            Some(expansion)
+          case Block(List(stat), Literal(Constant(()))) => unapply(stat)
```
allanrenucci: Can you give an example of when this happen?

biboudis: Yes, this is related to the implicit conversion related to value discarding (simply put when an `e` returns a value and it's called from a function of expected type `Unit` and it is converted into `{ e; () }`). This manifested in quoted code as well.

nicolasstucki: For example `inline foo: Unit = 4` will become `inline foo: Unit = { 4; () }` and then be inlined as `{ 4; () }`. In our case we have a `~xyz` instead of `4`.

allanrenucci: Got it! I would add a comment explaining why this is needed (i.e. the compiler can insert `()`).

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -338,5 +346,20 @@ class ReifyQuotes extends MacroTransformWithImplicits {
             checkLevel(mapOverTree(enteredSyms))
         }
       }
+
+    /** InlineSplice is used to detect cases where the expansion
+     *  consists of a (possibly multiple & nested) block or a sole expression.
+     */
+    object InlineSplice {
```
allanrenucci: I don't know how performance sensitive this code is but you could return a `Tree` instead of `Option[Select]` to avoid boxing (and `EmptyTree` instead of `None`)

biboudis: We need that `.get` method on `unapply`s return type so isn't `Option` unavoidable?

allanrenucci: I believe `Tree` qualifies for both [product pattern](http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html#product-pattern) and [name based pattern](http://dotty.epfl.ch/docs/reference/changed/pattern-matching.html#name-based-pattern). And scalac supports name based pattern

odersky:
What's the relationship with #4019?

nicolasstucki:
No relationship.

nicolasstucki:
Needs rebase

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -321,16 +321,22 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
             // To maintain phase consistency, we move the binding of the this parameter into the spliced code
             val (splicedBindings, stagedBindings) = bindings.partition {
               case vdef: ValDef => vdef.symbol.is(Synthetic) // Assume that only _this bindings are tagged with Synthetic
               case _ => false
             }
+
             val tree1 =
               if (level == 0) cpy.Inlined(tree)(call, stagedBindings, Splicer.splice(seq(splicedBindings, body)))
               else seq(stagedBindings, cpy.Select(expansion)(cpy.Inlined(tree)(call, splicedBindings, body), name))
             transform(tree1)
```
nicolasstucki: This should be a`tree2`

`val tree2 = transform(tree1)`

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -321,16 +321,22 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+
+            // due to value-discarding which converts an { e } into { e; () })
+            if (tree.tpe =:= defn.UnitType) Block(tree1 :: Nil, Literal(Constant(())))
```
nicolasstucki: And use `tree2` here and in next line

biboudis:
Approve @nicolasstucki? ⭐️

PR #4018, opened 2018-02-20
Fix #4005 command-line options help is incomplete

### LINKED ISSUE #4005: Command-line options help is incomplete
The option help text doesn't document correctly special options
```
$ dotc -help
...
-color             Colored output
...
$ ./bin/dotc -color /Users/pgiarrusso/git/dotty-example-project/src/main/scala/TestGADT.scala
missing argument for option -color
  dotc -help  gives more information
```

The culprit is https://github.com/lampepfl/dotty/blob/36cc65ad12ce809f7a9cd08d6ba3ba5432b75eed/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala#L74

That code just prints name and description of the `Setting`, ignoring choices, default and other fields.

ioganegambaputifonguser:
- added default value output
- added choices values output

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
+++ b/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
@@ -2,10 +2,12 @@ package dotty.tools.dotc
 package config
 
 import java.nio.file.{Files, Paths}
+
 import Settings._
 import core.Contexts._
 import util.DotClass
 import Properties._
+
 import scala.collection.JavaConverters._
 
 object CompilerCommand extends DotClass {
@@ -71,15 +73,23 @@ object CompilerCommand extends DotClass {
       val ss                  = (ctx.settings.allSettings filter cond).toList sortBy (_.name)
       val width               = (ss map (_.name.length)).max
       def format(s: String)   = ("%-" + width + "s") format s
-      def formatSettings(name: String, value: String) = {
-        if (value.nonEmpty)
-        // the format here is helping to make empty padding and put the additional information exactly under the description.
-          s"\n${format("")} $name: $value."
-        else
-          ""
+      def helpStr(s: Setting[_]) = {
+        def defaultValue(setting: Setting[_]) = setting.default match {
+          case _: Int | _: String => s.default.toString
+          case _ =>
+            // For now, skip the default values that do not make sense for the end user.
+            // For example 'false' for the version command.
+            ""
+        }
+        def formatSetting(name: String, value: String) = {
+          if (value.nonEmpty)
+          // the format here is helping to make empty padding and put the additional information exactly under the description.
+            s"\n${format("")} $name: $value."
+          else
+            ""
+        }
+        s"${format(s.name)} ${s.description} ${formatSetting("Default", defaultValue(s))} ${formatSetting("Choices", s.legalChoices)}"
       }
-      def helpStr(s: Setting[_]) =
-        s"${format(s.name)} ${s.description} ${formatSettings("Default", s.defaultValue)} ${formatSettings("Choices", s.legalChoices)}"
       ss map helpStr mkString "\n"
     }
 
--- a/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
+++ b/compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
@@ -71,7 +71,14 @@ object CompilerCommand extends DotClass {
       val ss                  = (ctx.settings.allSettings filter cond).toList sortBy (_.name)
       val width               = (ss map (_.name.length)).max
       def format(s: String)   = ("%-" + width + "s") format s
-      def helpStr(s: Setting[_]) = s"${format(s.name)} ${s.description}"
+      def formatSettings(name: String, value: String) = if (value.nonEmpty) format("\n") ++ s"  $name: $value." else ""
+      def helpStr(s: Setting[_]) = {
+        val help = StringBuilder.newBuilder
+        help.append(s"${format(s.name)} ${s.description}")
+        help.append(formatSettings("Default", s.defaultValue))
+        help.append(formatSettings("Choices", s.legalChoices))
+        help.toString
+      }
       ss map helpStr mkString "\n"
     }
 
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -88,7 +88,13 @@ object Settings {
         state.update(idx, x.asInstanceOf[T])
     }
 
-    def isDefaultIn(state: SettingsState) = valueIn(state) == default
+    def isDefaultIn(state: SettingsState): Boolean = valueIn(state) == default
+
+    def defaultValue: String = implicitly[ClassTag[T]] match {
+      case StringTag => default.asInstanceOf[String]
+      case IntTag => default.asInstanceOf[Int].toString
+      case _ => ""
```
Blaisorblade: Both branches could be `default.toString`, but it seems you want to exclude defaults that aren't String and Int? Could you explain why?

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
```diff
@@ -71,7 +71,14 @@ object CompilerCommand extends DotClass {
...
       val width               = (ss map (_.name.length)).max
       def format(s: String)   = ("%-" + width + "s") format s
-      def helpStr(s: Setting[_]) = s"${format(s.name)} ${s.description}"
+      def formatSettings(name: String, value: String) = if (value.nonEmpty) format("\n") ++ s"  $name: $value." else ""
+      def helpStr(s: Setting[_]) = {
+        val help = StringBuilder.newBuilder
+        help.append(s"${format(s.name)} ${s.description}")
+        help.append(formatSettings("Default", s.defaultValue))
+        help.append(formatSettings("Choices", s.legalChoices))
+        help.toString
```
Blaisorblade: Since this code already using the `s` formatter, what about just one expression?
`s"${format(s.name)} ${s.description}${formatSettings("Default", s.defaultValue)}${formatSettings("Choices", s.legalChoices)}"`? You can put arbitrary expressions inside `${}` 🙂

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
```diff
@@ -71,7 +71,14 @@ object CompilerCommand extends DotClass {
       val ss                  = (ctx.settings.allSettings filter cond).toList sortBy (_.name)
       val width               = (ss map (_.name.length)).max
       def format(s: String)   = ("%-" + width + "s") format s
-      def helpStr(s: Setting[_]) = s"${format(s.name)} ${s.description}"
+      def formatSettings(name: String, value: String) = if (value.nonEmpty) format("\n") ++ s"  $name: $value." else ""
```
Blaisorblade: Hm, why do we need to use `format` on `\n`? I'm probably being slow (or rushed)... 
And expressions can go inside the s formatter...
Any chance we can use `s"\n${format(name)} ${format(value)}"`? That's closer to what we do in `helpStr`, and hopefully clearer.

compiler/src/dotty/tools/dotc/config/Settings.scala
ioganegambaputifonguser: I found that only for these tag classes there are meaningful default values. For example for the `-version` there is `default: Boolean == false`, what does not make sense.

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
ioganegambaputifonguser: I use `format("\n")` as a hack to start the second line with an empty first column for Default or Choices
screen shot 2018-02-20 at 16 12 08

compiler/src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -88,7 +88,16 @@ object Settings {
         state.update(idx, x.asInstanceOf[T])
     }
 
-    def isDefaultIn(state: SettingsState) = valueIn(state) == default
+    def isDefaultIn(state: SettingsState): Boolean = valueIn(state) == default
+
+    def defaultValue: String = implicitly[ClassTag[T]] match {
```
allanrenucci: I would make it an inner method of `def helpStr`. Also you don't need a cast here:
```scala
def defaultValue(setting: Setting[_]): String = setting.default match {
  case _: Int | _: String => setting.default.toString
  case _ => ""
}
```

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
```diff
@@ -71,7 +71,15 @@ object CompilerCommand extends DotClass {
       val ss                  = (ctx.settings.allSettings filter cond).toList sortBy (_.name)
       val width               = (ss map (_.name.length)).max
       def format(s: String)   = ("%-" + width + "s") format s
-      def helpStr(s: Setting[_]) = s"${format(s.name)} ${s.description}"
+      def formatSettings(name: String, value: String) = {
```
allanrenucci: You can make it an inner method of `def helpStr`

compiler/src/dotty/tools/dotc/config/CompilerCommand.scala
```diff
@@ -71,14 +71,15 @@ object CompilerCommand extends DotClass {
...
-        help.toString
+      def formatSettings(name: String, value: String) = {
+        if (value.nonEmpty)
+        // the format here is helping to make empty padding and put the additional information exactly under the description.
+          s"\n${format("")} $name: $value."
+        else
+          ""
       }
+      def helpStr(s: Setting[_]) =
+        s"${format(s.name)} ${s.description} ${formatSettings("Default", s.defaultValue)} ${formatSettings("Choices", s.legalChoices)}"
```
Blaisorblade: @ioganegambaputifonguser Thanks for the changes, I think they address most of my suggestions :-)

ioganegambaputifonguser: @Blaisorblade  you are welcome. also added comments to make it more clear.

compiler/src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -88,7 +88,7 @@ object Settings {
         state.update(idx, x.asInstanceOf[T])
     }
 
-    def isDefaultIn(state: SettingsState) = valueIn(state) == default
+    def isDefaultIn(state: SettingsState): Boolean = valueIn(state) == default
```
Blaisorblade: I guess we don’t need to add this type annotation, no? Otherwise, LGTM.

ioganegambaputifonguser: I followed the Boy Scout Rule and added this small thing :) Is it true that the public methods have to have the return type annotation? Please correct me if I am not right.

ioganegambaputifonguser:
@allanrenucci done. Could you please take a look?

PR #4022, opened 2018-02-20
Fix #3200: bind var for singleton pattern get scrutinee type

### LINKED ISSUE #3200: Compatibility with `-Xfuture` in type checking patmat
The following code fails with `scalac -Xfuture test.scala`, but Dotty compiles without problem:

```Scala
object Test {
  def f(x: Any): Bob.type = x match { case x @ Bob => x }
}
```

This is a separate issue extracted from #1463.

liufengyun:
Fix #3200: bind var for singleton pattern get scrutinee type

Related issues:

https://issues.scala-lang.org/browse/SI-1503
https://github.com/lampepfl/dotty/issues/1463

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1320,7 +1320,12 @@ class Typer extends Namer
       case _ =>
         if (tree.name == nme.WILDCARD) body1
         else {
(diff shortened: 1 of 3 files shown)


tests/neg/3200b.scala
```diff
@@ -0,0 +1,5 @@
+object Test {
+  def main(args: Array[String]): Unit = {
```
odersky: Rename to i3200b.scala

PR #4024, opened 2018-02-21
Fix #3861: Spec changes to extractors implemented in #3747

### LINKED ISSUE #3861: Spec followup to #3747
#3747 changes pattern matching to allow "Pattern Sequences" to be used on `unapply` (where `unapplySeq` was previously required). This needs a spec change, [this page](http://dotty.epfl.ch/docs/reference/changed/vararg-patterns.html) would probably be a good place to do that.

odersky:
Fixes #3861

### DIFF
--- a/docs/docs/reference/changed/vararg-patterns.md
+++ b/docs/docs/reference/changed/vararg-patterns.md
@@ -18,4 +18,30 @@ supported:
     /*!*/ case List(1, 2, xs @ _*)       // syntax error
     /*!*/ case List(1, 2, _*) => ...     // syntax error
 
+Another change relates to extractors producing values with `T*` types, which can then be
+matched by vararg patterns. Previously, such extractors used an `unapplySeq` whereas now they use an `unapply` (in the long term, we plan to get rid of `unapplySeq` altogether, replacing all its usages with `unapply`).
+
+Example: Previously, this was correct:
+
+    class Person(val name: String, val children: Person *)
+    object Person {
+      def unapplySeq(p: Person) = Some((p.name, p.children))
+    }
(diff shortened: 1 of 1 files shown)


docs/docs/reference/changed/vararg-patterns.md
```diff
@@ -18,4 +18,30 @@ supported:
...
     /*!*/ case List(1, 2, _*) => ...     // syntax error
 
+Another change relates to extractors producing values with `T*` types, which can then be
+matched by vararg patterns. Previously, such extractors used an `unapplySeq` whereas now they use an `unapply` (in the long term, we plan to get rid of `unapplySeq` altogether, replacing all its usages with `unapply`).
+
+Example: Previously, this was correct:
+
+    class Person(val name: String, val children: Person *)
+    object Person {
+      def unapplySeq(p: Person) = Some((p.name, p.children))
```
allanrenucci: Maybe add a paragraph explaining the rules that make an `unapply` qualify for vararg extraction. I am not even sure what are the rules myself.  The return type must be of the form `Option[Seq[T]]` or `Option[(..., Seq[T])]`?

odersky: I can't context switch into that, sorry. Constant context switching is what kills my productivity. So either somebody else digs out the details and adds them or we leave it as is. Again, my plea stands that somebody will have to take over the pattern matching problematic.

odersky:
No activity for 3 weeks + 1 day. We either merge or close. Since I can't see a reason why this PR could make things worse I am merging. But we have to get better moving things along.

Issue #1551, opened 2016-10-02
Allow Singleton types in Union types


odersky:
#1550 disallows singleton types in union types to avoid the problems uncovered by #829. But it would be good to allow them, if we can find a solid solution.

The current scheme is: (1) singleton types in unions are not allowed. (2) Therefore, when taking the lub, we widen every singleton type. Thats simple and consistent, but also restrictive.

smarter:
Note that singleton types in unions can still slip through the cracks:
```scala
scala> def or[T](x: T | Int) = x 
or: [T](x: T | Int)T | Int
scala or["foo"]("bar") 
res13: String | Int = bar
```

It's probably not worth trying to disallow this case before we try to make singleton types in unions actually work.

propensive:
I uncovered another valid -- I think -- use-case while experimentitng with `enum`s, which doesn't involve singleton *literal* types.

This is intended to be a linked-list-like datastructure, but with alternating types, and it might make a good test case, but currently doesn't compile due to the restriction.

```
enum Fence[+T, +S] {
  case End
  case Post(value: T, next: Panel[T, S] | End.type)
  case Panel(value: S, next: Post[T, S])
}

import Fence._
val fence = Post(1, Panel("two", Post(3, End)))
```

soronpo:
Would have been nice to do:
`type Binary = 0 | 1`
There are other ways to achieve this constraint, though.

panacekcz:
Just a curiosity, the `or["foo"]("bar")` example doesn't compile anymore, but it is still possible to get unsoundness with literal types in a union.

```scala
object Get2As1 {
  class OrIntroFn[T, U, TU >: T|U]{
    type V = TU
    def tuToV(): (TU => V) = p => p
  }
  class Or11X[X] extends OrIntroFn[1&1, X, (1&1|X)]{
    def get2as11X:V = tuToV()(2)
  }
  class Or11Nothing extends Or11X[Nothing]
  def get2as1:1 = new Or11Nothing().get2as11X

  def main(a:Array[String]) = {
    println(get2as1) // prints 2
    val one:1 = get2as1
    println(one) // prints 1
  }
}
```

godenji:
In TypeScript this isn't an issue:
```scala
type ABC = 'A' | 'B' | 'C'
type A2F = ABC | 'D' | 'E' | 'F'

def foo(x: A2F) = ???
foo('F') // ok
foo('G') // fails
```
[...]

PR #4028, opened 2018-02-21
Add opaque types

odersky:
Implements SIP 35


odersky:
Still TODO: Make implicit scope include companion objects of opaque types.

odersky:
Overall the implementation adds just 50 lines of new code (partly because the scheme to find companion has gotten simpler). Anyway, that's a lot cheaper than value classes!

odersky:
test performance please

odersky:
I think this one is best reviewed commit by commit.

odersky:
If somemone has more tests it would be great to add them. I only picked those from the SIP and added some small tests that test various invalid code.

odersky:
test performance please

jvican:
I’ll have a look at this on Tuesday, it’s great to see progress on this area.

japgolly:
Just gave this a try:
```scala
opaque type Fix[F[_]] = F[Fix[F]]
```
```
4 |  opaque type Fix[F[_]] = F[Fix[F]]
  |              ^
  |illegal cyclic reference: alias [F <: [_$1] => Any] => F[opaquetypes.Fix[F]] of type Fix refers back to the type itself
```

Any plans to support recursive opaque types?

odersky:
test performance please

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -299,6 +300,7 @@ object TastyFormat {
   final val DEFAULTparameterized = 30
   final val STABLE = 31
   final val MACRO = 32
+  final val OPAQUE = 33
```
jvican: Does this deserve a version bump in the tasty format?

compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
```diff
@@ -212,8 +212,7 @@ private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder
 
     // Synthetic methods that are always present do not affect the API
     // and can therefore be ignored.
-    def alwaysPresent(s: Symbol) =
-      s.isCompanionMethod || (csym.is(ModuleClass) && s.isConstructor)
+    def alwaysPresent(s: Symbol) = csym.is(ModuleClass) && s.isConstructor
```
jvican: Why don't declarations in the API extractor include companion methods? What invariant has opaque types introduced to make the previous code obsolete?

compiler/src/dotty/tools/dotc/transform/FirstTransform.scala
```diff
@@ -109,10 +121,13 @@ class FirstTransform extends MiniPhase with InfoTransformer { thisPhase =>
     }
 
     def registerCompanion(name: TermName, forClass: Symbol): TermSymbol = {
-      val (modul, mcCompanion, classCompanion) = newCompanion(name, forClass)
+      val modul = newCompanion(name, forClass)
       if (ctx.owner.isClass) modul.enteredAfter(thisPhase)
-      mcCompanion.enteredAfter(thisPhase)
-      classCompanion.enteredAfter(thisPhase)
+      val modcls = modul.moduleClass
+      modcls.registerCompanion(forClass)
```
jvican: Should `modcls` be installed after this phase too?

odersky:
> Could you elaborate on why recursive opaque types are not supported in this implementation and how that affects existing language semantics? Our goal would be to support this in the scalac implementation.

Recursion is one of the toughest things to handle, generally. For instance, F-bounds seem useful, but have turned out the single biggest snake-pit for Scala's implementation. I can't count the number of times I wished Scala did not have them! In DOT we do have recursion via `this`, i.e. `rec(x: T)`. That's necessary to express a lot of what Scala does, including all objects and classes (and F-bounds as well), but it is also very tricky to formalize and prove sound. Adding another kind of recursion over _type_ variables for some (IMO rather marginal) use cases is completely the wrong tradeoff of complexity for convenience. This is a fundamental choice, not an implementation-dependent one.

Blaisorblade:
> Could you elaborate on why recursive opaque types are not supported in this implementation and how that affects existing language semantics? Our goal would be to support this in the scalac implementation.

@jvican To elaborate from the theoretical point of view: a recursive opaque type `T = F[T]` would amount to an *equirecursive type* as soon as the equality between `T` and `F[T]` is visible. And typechecking for equirecursive types are hard to implement *for simple types* (see Pierce's TaPL). If you add *just* type constructors you still get soundness, but typechecking lands between "barely decidable", "decidable with a complexity class involving multiple exponentials" and "undecidable" (see POPL paper at http://www-ps.informatik.uni-tuebingen.de/research/functors/, which doesn't even include subtyping or any other Scala feature or inference).
And that's if you make the recursion explicit and make sure to have a relevant node in your types

The only hope is that you want to typecheck programs before the equality is visible, but type-directed algorithms where the equality is visible would still risk needing to deal with recursion somehow.
But from the source program you can easily tell where you convert between `T` and `F[T]`: you can then insert a special coercion node that is erased by the backend. But even a sketch would take longer than a comment here.

jvican:
Good arguments, thanks for chiming in the discussion.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
odersky: You are right, it should be a minor version bump.

compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
odersky: Companion methods don't exist anymore. (They were the synthetic methods that pointed to the companion class or object to the other. We now use an explicit field in `ClassDenotation` for this).

odersky:
I integrated the remaining test cases from SIP 35, except for the fixpoint one, which is not supported.

odersky:
@nicolasstucki We have some FromTasty tests failing for opaque tests. My guess is that somehow gadt bounds do not get updated correctly. Can you take a look?

nicolasstucki:
I had a look at the from TASTY tests. It can be minimized to:
```scala
object opaquetypes {
  opaque type Logarithm = Double
  object Logarithm {
    class LogarithmOps(x: Logarithm) {
      def foo: Double = x + 1
    }
  }
}
```

Before pickling the tree is
```scala
package <empty> {
  final lazy module val opaquetypes: opaquetypes$ = new opaquetypes$()
  @scala.annotation.internal.SourceFile("tests/pos/opaque.scala") final module 
    class
   opaquetypes$() extends Object() { this: opaquetypes.type => 
    @<opaque>[Double{<companion>: opaquetypes.Logarithm.type}] opaque type Logarithm = Double
    final lazy module val Logarithm: opaquetypes.Logarithm$ = 
      new opaquetypes.Logarithm$()
    final module class Logarithm$() extends Object() { 
      this: opaquetypes.Logarithm.type =>
     
      class LogarithmOps(x: opaquetypes.Logarithm) extends Object() { 
        val x: opaquetypes.Logarithm
        def foo: Double = LogarithmOps.this.x.+(1)
      }
    }
  }
}
```

After unpickling the tree is 
```scala
package <empty> {
  final lazy module val opaquetypes: opaquetypes$ = new opaquetypes$()
  @scala.annotation.internal.SourceFile("tests/pos/opaque.scala") final module 
    class
   opaquetypes$() extends Object() { this: opaquetypes.type => 
    @<opaque>[Double] opaque type Logarithm = Double
    final lazy module val Logarithm: opaquetypes.Logarithm$ = 
      new opaquetypes.Logarithm$()
    final module class Logarithm$() extends Object() { 
      this: opaquetypes.Logarithm.type =>
     
      class LogarithmOps(x: opaquetypes.Logarithm) extends Object() { 
        val x: opaquetypes.Logarithm
        def foo: Double = LogarithmOps.this.x.+(1)
      }
    }
  }
}
```

Note that the `@<opaque>` annotation losses information. 
```scala
@<opaque>[Double{<companion>: opaquetypes.Logarithm.type}] opaque type Logarithm = Double
```
```scala
@<opaque>[Double] opaque type Logarithm = Double
```

nicolasstucki:
@odersky 
On the following code:
```scala
object opaquetypes {
  opaque type Logarithm = Double
  object Logarithm {
    class LogarithmOps(x: Logarithm) {
      def foo: Double = x + 1
    }
  }
}
```
When unpickling the selection of member `+` on `x` of type `Logarithm`, the tree unpickler tries to select the member of `x` ([here](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala#L942)), but it can not find it as it is not a member on `Logarithm` due to its opacity. We probably need to special case opaque types here and look at their underlying type.

odersky:
@nicolasstucki Fixes to FromTasty look good to me. Thanks!

TomasMikula:
It is unfortunate that in order to define a method that sees through multiple opaque types, we have to introduce artificial nesting. For example

```scala
object graphics {
  opaque type Image = ???
  object Image {
    opaque type Point = ???
    object Point {
      opaque type Color = ???
      object Color {
        def colorAt(img: Image, p: Point): Color = ???
      }
    }
  }
}
```

Why should `Color` be defined inside `Point` inside `Image`?

sjrd:
@TomasMikula That's not how I would structure it. I would make them flat, and each companion would expose a pair of `fromUnderlying`/`toUnderlying` that would be `private[graphics]`. Then `colorAt`, being defined in `graphics`, would use those explicit private conversions.

If explicit conversions are not enough in your opinion, could you provide a concrete use case?

TomasMikula:
That's indeed sufficient in this example and perhaps a better aesthetic trade-off than nesting. Still feels like more boilerplate than necessary, though.

More generally, instead of `fromUnderlying`/`toUnderlying` we would need to expose type equality (`=:=`) and (perhaps less frequently) subtype relation (`<:<`), and their higher-kinded versions (i.e. equality/subtyping of type constructors) which are not even part of the standard library. Uses of [`substitute`](https://scala-lang.org/files/archive/api/2.13.0-M3/scala/Predef$$$eq$colon$eq.html#substitute[F[_]](ff:F[From]):F[To]) can get quite hairy. That seems like a lot of bureaucracy and advanced concepts to achieve a no-op conversion.

sjrd:
Do you have a suggestion how we could address this issue?

Maybe we can view the underlying type as a form of *private* information of the type, which is why the companion object can see through it. Under that interpretation, we could generalize this privateness to qualified privateness:
```scala
object graphics {
  opaque[graphics] type Image = ImageImpl
}
```
would mean that the opaque type alias can be dealiased everywhere inside `graphics`, but not outside of it.

I haven't thought about the feasibility of this generalization in terms of implementation, but it *seems* to make sense from a design point of view ...

This should be discussed on the SIP discussion, though, not in this PR.

TomasMikula:
That seems like a good solution. Moving the discussion to [Pre-SIP: Unboxed wrapper types](https://contributors.scala-lang.org/t/pre-sip-unboxed-wrapper-types/987/71).

odersky:
I have a tendency to merge this now. I just rebased and it was quite painful to keep it up-to-date. I have the impression that opaque types will be introduced in some form or another. The main objection against is duplication with value classes, and there we have with extension methods #4114 a solution that is also generally liked.

So it seems it will come down to 

 - add opaque types
 - add extension methods a la #4114 
 - deprecate value classes and implicit classes

smarter:
> I have a tendency to merge this now. I just rebased and it was quite painful to keep it up-to-date. 

Please assign a reviewer first. I don't think master will change too much over the next week so we have some time.

> deprecate value classes

There's a lot of confusion around this and it needs to be discussed more. Whether or not opaque types are a replacement or an addition to value class should affect their design, but in https://contributors.scala-lang.org/t/pre-sip-unboxed-wrapper-types/987/64?u=smarter, @non said:
> One reason including this kind of syntax in this SIP might be difficult is that we’ve already received feedback that the first version (which included this kind of syntax) overlapped too much with value classes, and that the feature didn’t seem sufficiently novel.

Furthermore, I'm personally strongly against deprecating value classes because we'll need good two-way interop with Java value classes when they arrive, and because I think opaque types are an inferior solution for most usecases (they require users to reason about erasure, an advanced topic in Scala; they require a lot of boilerplate to imitate the behavior you'd want from a class).

jvican:
> Furthermore, I'm personally strongly against deprecating value classes because we'll need good two-way interop with Java value classes when they arrive, and because I think opaque types are an inferior solution for most usecases

Note that opaque types around primitive types will have full-blown value type support from the very day in which Valhalla is merged in the JDK sources. The runtime will detect when boxing of primitives happen and special case it so that there's no overhead.

If Java does get value types, opaque types will only need to be special cased in codegen so that we omit the right bits. The language support in Java for value types is still in flux, so we cannot make a technical assessment of what will be required, but the Java folks were proposing to have annotations to infer value types class descriptors so that they can tell the runtime to optimize them away.
  
In some sense, opaque types are already the interface of "value types" in the Scala language, but independent of the runtime semantics, focused on library design and acommodating other cases (yay, immutable arrays). This is a selling point.

> they require users to reason about erasure

I don't see how opaque types are dependent of erasure; I can give you a pitch for normal Scala developers that will not even mention the word performance. 
  
Opaque types are just wrapper types, a familiar concept for many programmers (e.g. C# programmers).
  
I agree that reasoning about erasure is an advanced erasure -- Scala programmers won't need to be bothered by opaque types more than they are now bothered by `int` wrapping to `java.lang.Integer`; e.g. only people that *really* care about performance reason about that runtime property.
  
> they require a lot of boilerplate to imitate the behavior you'd want from a class
  
Isn't that solved by @odersky's extension method proposal in Dotty?
  
Note that this "syntax boilerplate" will be alleviated if we have top-level definitions in Scala. Not clear if we will eventually get it, but it seems we're on the way.

smarter:
> If Java does get value types, opaque types will only need to be special cased in codegen so that we omit the right class file bits.

I don't understand what you mean, opaque types are not classes, they're type members that never reach the Java bytecode, so there's nothing to emit in classfiles.

> I don't see how opaque types are dependent of thing; I can give you a pitch for normal Scala developers that will not even mention the word performance.

Why would you use opaque types if you don't care about performance?

> Opaque types are just wrapper types, a familiar concept for many programmers (e.g. C# programmers).

When you pattern match on a type, you should be able to distinguish between a wrapper type and a wrapped type, you can do this with value classes (in Scala now and in Java in the future), but not with opaque types. Understanding this requires reasoning about erasure.

> > they require a lot of boilerplate to imitate the behavior you'd want from a class
> Isn't that solved by @odersky's extension method proposal in Dotty?

opaque type + extension method is a lot more boilerplate than just writing a class. Converting between a class and opaque type + extension method is much more work than adding/removing `extends AnyVal`.

jvican:
> I don't understand what you mean, opaque types are not classes, they're type members that never reach the Java bytecode, so there's nothing to emit in classfiles.
  
(I was referring to the example of an opaque type of a class extending `AnyRef`. In that case, the compiler can decide to make the underlying class a candidate for a value-type if deemed possible (in a transitive fashion). If converting that class to a value type is not possible, you always pay the price of boxing and unboxing, so there's nothing you can do.)
  
The case that opaque types handle the best performance-wise is wrapping around `AnyRef` classes, because all references to the wrapper type are gone after typing and only the underlying type remains (e.g. type tagging is a common example). For this case, we don't need special value type support.
  
The case that opaque types cannot handle unless specialization is enabled (wrapping around primitive types) is handled by Valhalla *out of the box* because the current value types proposal has special support for primitive types (the runtime will always try to remove boxing/unboxing on them).
  
Do you know of any example in which opaque types will need special support to benefit from value types?
  
> Why would you use opaque types if you don't care about performance?
  
You use opaque types to solve the (common) problem of wrapping types to provide more type safety. Why opaque types over value classes? They give you the solution with the best default performance you can get, in all cases.
  
As programming language designers, should we not favor features with good defaults?

Another revealing example is to constrain the API of an internal, mutable class by an immutable class (for example, `Array`).
  
> When you pattern match on a type, you should be able to distinguish between a wrapper type and a wrapped type
  
This assumption goes against the current proposal of value types in Java. It's not possible to distinguish between the two in a pattern match. You cannot have different `equals` or `hashCode` between the wrapped type and the underlying type, and John Rose confirms that in his proposal.
  
> opaque type + extension method is a lot more boilerplate than just writing a class. Converting between a class and opaque type + extension method is much more work than adding/removing extends AnyVal.
  
I believe the boilerplate problem is not as big as we think:
  
* You can write a rewrite for it.
* You can alleviate it with top-level definitions or similar improvements.
  
On top of that, I'm not sure until which extent this slight increase in boilerplate is bad. I believe opaque types help you think carefully about the API you expose in an idiomatic Scala way (users are aware they are not defining a class, they are defining a type and its companion). This key difference makes the feature more intuitive; developers cannot expect the same semantics of a class.
  
In Dotty land, people will use extension methods instead of `AnyVal`s. In Scala land, people will use `AnyVal` exclusively for extension methods (which is where they shine anyway).

sjrd:
> Why would you use opaque types if you don't care about performance?

The main reason I care about opaque types: good interoperability semantics. They will *always* be seen by the host language (Java/JavaScript) as the underlying type, no matter whether they are in a monomorphic context, or lifted in generic types.

As opposed to `AnyVal`s, which:

* In Scala.js, are always seen as their *boxed* variant (whether in a monomorphic or generic context),
* In Scala/JVM, *much worse*, are seen by Java as their underlying type if in a monomorphic context, but as their boxed variant in a generic context.

smarter:
> (I was referring to the example of an opaque type of a class extending AnyRef. In that case, the compiler can decide to make the underlying class a candidate for a value-type if deemed possible (in a transitive fashion).

This is impossible without a whole-program linker. I'm also having a hard time understanding what you're describing here. If I wrap String into PersonName, I want PersonName to be a value class, I don't want String to be a value class.

> I believe the boilerplate problem is not as big as we think:

If I thought boilerplate was not a big deal, I would program in Java :).

> The main reason I care about opaque types: good interoperability semantics. They will always be seen by the host language (Java/JavaScript) as the underlying type, no matter whether they are in a monomorphic context, or lifted in generic types.

This is a valid usecase, but it's also an extremely niche one.

smarter:
> You cannot have different equals or hashCode between the wrapped type and the underlying type, and John Rose confirms that in his proposal.

Not sure what you mean here, a value class is not == to the type it's wrapping (that wouldn't even be possible for a value class with more than one field).

sjrd:
> but it's also an extremely niche one.

Go tell that to all the Scala.js developers who need to understand that
```scala
sealed trait Weekday extends js.Any

object Weekday {
  val monday = "monday".asInstanceOf[Weekday]
  ...
  val sunday = "sunday".asInstanceOf[Weekday]
}
```
not only means something, but is *the* way to type a large number of "data types" in JavaScript.

Whereas it would be so simple and *meaningful* to write:
```scala
opaque type Weekday = String

object Weekday {
  val monday: Weekday = "monday"
  ...
}
```

smarter:
I would rather write `type Weekday = "monday" | "tuesday" | ...` (which is also the TypeScript way to do it I believe), although that requires fixing https://github.com/lampepfl/dotty/issues/1551 :).

smarter:
> Do you really believe that there is a significant difference in syntax boilerplate between the following two equal examples? 

Yes, because the existing version is not just boilerplate, if you remove `extends AnyVal`, it still has meaning. A class, a method in a class, an object, all of those are existing concepts in Scala that are well-known (and can be composed too, a value class can be made a case class, removing some of the boilerplate from your example). The opaque type proposal introduces several new concepts that are only used for opaque types, such as the notion that a type is or is not an alias of an underlying type depending on the current scope, and the notion of a companion object for types (I wouldn't have a problem with this if all types had companion objects, but only opaque types have them in the proposal).

> My point, and John Rose's proposal, is that they are both equal -- you cannot tell them apart because you cannot define a custom equals and hashCode for the wrapper type.

An Int is not == to a WrappedInt if WrappedInt is a Valhalla value class.

> "opaque types do have a good story to support value types in the future"

I disagree.
- Opaque types are not usable from Java.
- Opaque types don't help you when you have multiple fields.
- Java value classes cannot be interpreted as opaque types when used in Scala, they have different semantics.

jvican:
> A class, a method in a class, an object, all of those are existing concepts in Scala that are well-known (and can be composed too, a value class can be made a case class, removing some of the boilerplate from your example).

The fact that an opaque type doesn't have these properties is a feature, not a disadvantage. You don't want to ressemble a class when the runtime semantics are not those of a class. When we tried to make opaque types looked like a class in the first version of the spec, Erik and I got overwhelming feedback from the SIP Committee saying that ressembling a class was a bad idea.
  
In hindsight, I think this is the most critical mistake that the current encoding of value types made, even though it was elegant on paper. You want to be as intuitive as possible when it comes to semantics.
  
> An Int is not == to a WrappedInt if WrappedInt is a Valhalla value class.
  
You don't want to be doing this kind of equality, because you'll get bitten. From the "Heisenboxes" section in [the last section of the proposal in May 2017](http://cr.openjdk.java.net/~jrose/values/shady-values-0510.md):
  
>   * Reference comparision (Java operator `==`, or the `acmp`
    instruction) may report "true" on two equivalent boxed Q-type
    values, even if the references previously returned false, or
    "false" when they previously returned "true".  Such variation
    would of course be subject to the logic of substitutability, of the
    underlying Q-types.  Two boxes that were once detected as equal
    references would be permanently substitutable for each other.
  
It's not very useful to be able to tell that `WrapperType != UnderlyingType` if `WrapperType == WrapperType2` when they point to the same underlying type.
  
That is, at the moment you rely on wrapper type equality, all your type safety guarantees are gone. The approach of opaque types here is always safe: a wrapper type is equivalent to the underlying type because one erases to the other one (easy to reason about; just as if you were "inlining" the type).
  
If you try to pattern match on a wrapper type and the underlying type, you'll get a warning or even a compile error from `patmat`. Henc
  
> Opaque types are not usable from Java.
  
How many Scala features can be used from Java? OR types? AND types? Implicits? The list is long.
  
It seems reasonable to me opaque types cannot be used from Java; they are a property of Scala's typechecker. You could even say their existence is an implementation detail.
  
In reality, Scala language features are usually not easy (or even possible) to use from Java, so this is not something to be concerned about. If a user really wants to declare a wrapper type and use it from Java, they can use the language support for value types in Java (and then use it from the Scala side, as we've always done).
  
> Opaque types don't help you when you have multiple fields.
  
It's not yet clear wrapper types of multiple fields will be possible in the current proposal of value types. 

If they happen to be, you can represent multiple fields as tuples, and have them special cased as value-type-compatible classes afterwards in codegen. `opaque type Foo = (Int, Int)` does the trick. If that is not good enough, we can then think of adding special syntax to support this use case.
  
> Java value classes cannot be interpreted as opaque types when used in Scala, they have different semantics.
  
Indeed. But why does the Scala compiler need to interpret value types coming from Java or even know about them? They are useless for the compiler -- they are just another class; the semantics of value types don't affect typer.

<hr>
  
Lastly, I'd like to highlight the high level of uncertainty around Valhalla. Valhalla doesn't have yet a solid implementation that can prove that all the properties explained in the spec are achievable, and John Rose reserves the right to change them upon feedback from its implementation. So we shouldn't be too attached to what Valhalla promises to bring until there's clear evidence that it will be merged in the JDK sources and all the hypothesis are justified.
  
Also, Valhalla competes against Graal in many ways, and I think it's fair to say that some people at Oracle think they will never happen. I suggest we don't take for granted that value types will ever be production-ready. Even the proposal says that if there's a prototype it will be under experimental mode for a long time, and it may not even be production ready *ever*.
  
That being said, if they ever get merged, opaque types can benefit from value types just as much as `AnyVal`s can, with lots of upsides: support across different runtimes, easier to reason about, better API design and superior defaults performance-wise *now*.
  
If value types never make it, opaque types will still be the best we can do to solve the problem of wrapping a type in Scala.

smarter:
> It's not very useful to be able to tell that WrapperType != UnderlyingType if WrapperType == WrapperType2 when they point to the same underlying type.

I think you're misinterpreting the proposal here, it's talking about Java `==` which is Scala `eq`, it's saying that when switching from a class to a value class in Java, `a == b` can suddenly start returning true. Something like this would not be possible in Scala because `eq` does not exist on types that extend `AnyVal`. I don't think the proposal is saying that two different wrapper classes of the same underlying type are `==`, that would be crazy.

> How many Scala features can be used from Java? OR types? AND types? Implicits? The list is long.

This is not a good argument for making things worse. All of those things are not supported in Java, value classes will be supported in Java.

> It's not yet clear wrapper types of multiple fields will be possible in the current proposal of value types.

Everything I've seen discussed so far seems to assume that (e.g., the most common example is a Complex type that would wrap two ints). The latest proposals from February 2018 do not restrict the number of fields: http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2018-February/000556.html

> If they happen to be, you can represent multiple fields as tuples, and have them special cased as value-type-compatible classes afterwards in codegen.

I don't understand what kind of codegen you're talking about, a full example would be appreciated.

> opaque type Foo = (Int, Int) does the trick.

No, tuples members are unnamed, fields have names.

> that is not good enough, we can then think of adding special syntax to support tuples with parameter names.

Let's add even more special syntax, for something that would just be a class in Java?

> Indeed. But why does the Scala compiler need to interpret value types coming from Java or even know about them? They are useless for the compiler -- they are just another class; the semantics of value types don't affect typer.

At the very least they need to `extend AnyVal` so that you don't use them in illegal ways.

> Lastly, I'd like to highlight the high level of uncertainty around Valhalla. Valhalla doesn't have yet a solid implementation that can prove that all the properties explained in the spec are achievable, and John Rose reserves the right to change them upon feedback from its implementation. So we shouldn't be too attached to what Valhalla promises to bring until there's clear evidence that it will be merged in the JDK sources and all the hypothesis are justified.

Valhalla is being actively developed, see http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/. By the way, this mailing list is open to non-Oracle people so they can provide feedback, I think some Scala people are members although I'm not sure any of them are active currently. We could ask for more members of the Scala community to be included.

> Also, Valhalla competes against Graal in many ways, and I think it's fair to say that some people at Oracle think they will never happen. I suggest we don't take for granted that value types will ever be production-ready. Even the proposal says that if there's a prototype it will be under experimental mode for a long time, and it may not even be production ready ever.

I expect them to be complementary, not to compete. If Valhalla is abandoned of course we'll need to adapt for this, but until then we should definitely take it into account when planning for the feature.

> That being said, if they ever get merged, opaque types can benefit from value types just as much as AnyVals can, with lots of upsides: support across different runtimes, easier to reason about, better API design and superior defaults performance-wise now.

Again, I really don't understand how this is possible, opaque types are not classes, they cannot be represented by value classes.

jvican:
> This is not a good argument for making things worse. All of those things are not supported in Java, value classes will be supported in Java.
  
Well, you never know what Java will have next.
  
Will `enum`s from Dotty be usable from Java as if they were Java data classes, which were planned to come to Java pretty soon? The reasonable answer here is no, too. The same should apply to opaque types. Every language has its own language specific way of solving a problem, with independent semantics. Widening the scope of that and adding cross-language features only makes things more difficult.
  
I don't think this is an argument "that makes things worse". If "Java->Scala" compatibility is the property we use to judge Scala features, then we're doomed. 
  
> I don't understand what kind of codegen you're talking about, a full example would be appreciated.
  
If multi field value types are possible, then tuples can just be represented as value types (and for that you need to modify how they are described in the class file).
  
> At the very least they need to extend AnyVal so that you don't use them in illegal ways.

Hmmm, I'm not sure. I think one is not supposed to override wrapper types (I didnt find anything in the spec).
  
The current proposal doesn't describe any invariant that must not be violated in the use site. I would be very surprised if this was the case. The whole point of the proposal is value types are seen from the outside just as if they were a normal class. I still believe that there's no need for the Scala compiler to special case value types from Java.
  
> Valhalla is being actively developed, see http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/
  
Yeah, I'm subscribed and I've read all the latest mailing lists in 2018 (January, February and March). Still no further action on the implementation nor refinement of the value types spec... only brainstorming on template classes.
  
> I expect them to be complementary, not to compete.
  
Not to the best of my knowledge, they're very much in competition since they cannot be reconciled together -- they are forked VMs with different trade-offs and design goals.
  
> Again, I really don't understand how this is possible, opaque types are not classes, they cannot be represented by value classes.
  
Let me try to be clearer...
  
* Are you using opaque types to wrap classes extending any ref? Good, you get the best performance possible, no need for value types. :heavy_check_mark: 
* Are you using opaque types to wrap primitive values and you don't use specialization? Good, Valhalla special cases them to remove boxing/unboxing, so they benefit from the performance of value types without any specific compiler support. :heavy_check_mark: 
* Are you using opaque types to wrap product types, as if they were multi-field classes? Good, you can modify the compiler to make tuples value types, and add better syntax to have named parameters. :question: 
  
The two first cases, the most critical and popular ones, will benefit from opaque types from day 1, with no compiler support whatsoever. That won't be the case if you want to enrich `AnyVal` with value types, you need specific compiler support. The third case can be done if we deem it necessary later down the road. You don't need opaque types to be classes to benefit from value types.
  
Overall, I prefer a cross-runtime alternative that we can ship today and has many other upsides over waiting for an unspecified period of time (would 2 or 3 years be a good estimate?) for Valhalla to be fully implemented, refined, tested and production ready, and then get them compatible with `AnyVal`s.
  
I don't know... it feels fair to me to pay a 9% increase in boilerplate and the impossibility of using opaque types from Java to get a superior alternative for wrapper types in *all* regards.
  
The rest of the points you've made (value types support, multi-field arguments) can be acommodated if Valhalla is ready in the future. These are good points, I'm sure a collaborative effort (like the one we're having now) can find the best solution for them.

smarter:
> Will enums from Dotty be usable from Java as if they were Java data classes, which were planned to come to Java pretty soon?

We'd like Dotty enums to be Java enums if possible: https://github.com/lampepfl/dotty/issues/4088

> Not to the best of my knowledge, they're very much in competition since they cannot be reconciled together -- they are forked VMs with different trade-offs and design goals.

Graal is distributed with Java 10 now and can be enabled with `-XX:+UnlockExperimentalVMOptions -XX:+UseJVMCICompiler`, so it's not really a fork anymore.

> Are you using opaque types to wrap classes extending any ref? Good, you get the best performance possible, no need for value types.

Unless you want to use your wrapper from Java, or you want to pattern match on the type of your wrapper.

> Are you using opaque types to wrap product types, as if they were multi-field classes? Good, you can modify the compiler to make tuples value types, and add better syntax to have named parameters. 

Again, we shouldn't need new syntax for something as simple as a class with multiple fields, we already have syntax for that, it's `class Foo(a: X, b: Y)`, it cannot get better than this.

> Overall, I prefer a cross-runtime alternative that we can ship today and has many other upsides over waiting for an unspecified period of time (would 2 or 3 years be a good estimate?) for Valhalla to be fully implemented, refined, tested and production ready, and then get them compatible with AnyVals.

Shipping a new feature is a big deal. It may be the right thing to do here but we need to have a plan for how things will evolve in the future. We cannot just ship whatever is most convenient today and ignore what will happen in a few years.

> I don't know... it feels fair to me to pay a 9% increase in boilerplate and the impossibility of using opaque types from Java to get a superior alternative for wrapper types in all regards.

The price to pay is teaching everybody about this new concept (Rust RFC have a "How are we going to to teach this?" section, I believe Scala SIPs should have that too), porting code to use this new concept, and potentially porting it back to value classes in two years.

smarter:
> Yeah, I'm subscribed and I've read all the latest mailing lists in 2018 (January, February and March). Still no further action on the implementation nor refinement of the value types spec... only brainstorming on template classes.

Looking at http://mail.openjdk.java.net/pipermail/valhalla-dev/2018-March/thread.html I see quite a lot of commits actually, so I wouldn't say there's no action on the implementaiton. Concerning the spec, the latest version I can find comes from http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2018-February/000556.html from February 2018, so I wouldn't say it's not being refined either.

sjrd:
> I don't think this is an argument "that makes things worse". If "Java->Scala" compatibility is the property we use to judge Scala features, then we're doomed.

JavaScript->Scala.js and Scala.js->JavaScript compatibility are the properties that *we* use to judge Scala.js features. And I daresay we've been quite successful with that criterium!

odersky:
The first parts of this were already reviewed by @jvican. The commits starting with 6161baf still need to be checked. I leave @jvican as reviewer but everybody else can also chip in.

jvican:
> Graal is distributed with Java 10 now and can be enabled with -XX:+UnlockExperimentalVMOptions -XX:+UseJVMCICompiler, so it's not really a fork anymore.
  
That's not the point, though. They both were forks at some point. The fact that Graal is quickly getting adopted mainstream actually tells a lot about who seems to be the successor of the JVM as we know it now...
   
> Unless you want to use your wrapper from Java, or you want to pattern match on the type of your wrapper.
  
Those are different use cases that have nothing to do with the fact that opaque types give you the best performance for classes extending `AnyRef`, out of the box, without any compiler support.
  
I'm not sure if "pattern matching on the type of a wrapper" is something that users of value classes do. I'll try to run some code analysis on Olafur's codebase, but my guess is that it's rarely used and hence not a major blocker.
  
> Again, we shouldn't need new syntax for something as simple as a class with multiple fields, we already have syntax for that, it's class Foo(a: X, b: Y), it cannot get better than this.
  
I think it's important that something that is not a class should not be represented as a class. The fact that `AnyVal` classes are not classes is obvious from looking at what you cannot do with them:
  
* They cannot have constructors (no initialization logic).
* You can only extend universal traits (which add complexity to the language and are useless outside of the scope of `AnyVal`).
* They can only have one field.
* They cannot override `equals` or `hashCode`.
  
Spec-wise, `AnyVal`s are not JVM classes. Are you suggesting that we represent wrapper types as a class even if they are not, just because the syntax is nicer? I think we can do better. If we really care about syntax, I don't see why we cannot have:
  
```scala
opaque type Logarithm = Double
opaque type Int128 = (f: Int, f2: Int)
```
  
or even:

```scala
opaque type Logarithm(d: Double)
opaque type Int128(f: Int, f2: Int)
```
  
> Shipping a new feature is a big deal. It may be the right thing to do here but we need to have a plan for how things will evolve in the future. We cannot just ship whatever is most convenient today and ignore what will happen in a few years.

I wholeheartedly agree, that's why we're discussing now potential ways of evolving opaque types in the future should value types come to the JVM.
  
> The price to pay is teaching everybody about this new concept (Rust RFCs have a "How are we going to to teach this?" section, I believe Scala SIPs should have that too), porting code to use this new concept, and potentially porting it back to value classes in two years.
  
It's a concept they're already familiar with: type aliases. The familiarity between both is really high, except the visibility of the type. I don't think this is a big deal, Scala devs use `private[this]` or `private[mypackage]` all the time, they're already used to think from where a term or type is visible.
  
Erik and I are doing a complete rewrite of the proposal, so I'll make sure we add a simple pitch that answer this question. I think we can teach opaque types better than value classes + universal traits and all their weird limitations.
  
BTW, I really like your suggestion of having a "How are we going to teach this" section, I'm going to add it to the SIP template :+1: 

> Looking at http://mail.openjdk.java.net/pipermail/valhalla-dev/2018-March/thread.html I see quite a lot of commits actually, so I wouldn't say there's no action on the implementation.
  
Can you point me to the repository? That's a link to the mailing list. Last time I checked the activity was limited.
  
> Concerning the spec, the latest version I can find comes from http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2018-February/000556.html from February 2018, so I wouldn't say it's not being refined either.
  
I read those slides two weeks ago, but those don't describe the spec nor refine the design goals and non-goals of the original document proposed by John Rose. They are discussing what seem to be implementation details, so still, the best thing we can do to discuss about this specification is the document I linked to before from May 2017.
  
> JavaScript->Scala.js and Scala.js->JavaScript compatibility are the properties that we use to judge Scala.js features. And I daresay we've been quite successful with that criterium!
  
:thinking: Fair point. Just wanted to emphasize that Java->Scala compatibility is not in my book to evolve Scala, and I don't think it has been judging from the current design decisions we've taken during the past years.
  
Using the Scala collections from Java is hard at best, and there are many Scala features like `lazy val`s or by-name parameters that have no seamless Java support whatsoever. Why should we apply this rule for this concrete language feature, then?
  
The fact that opaque types are a feature that only the typechecker understands is a feature, not a limitation.
  
> The commits starting with 6161baf still need to be checked. I leave @jvican as reviewer but everybody else can also chip in. 
  
I'm having a look at it some time this afternoon!

odersky:
Value classes seem to have a number of functions

 1. A low-cost implementation of some functionality in terms of some other type
 2. A way to add extension methods
 3. (In the future, once we can support multiple parameters) A way to do structs

Let's look at them in turn. For (1), the question is: Should the implementation ever involve boxing or not? Value classes do box when going from monomorphic to generic context. This is a heavy price to pay, mainly because the consequences are not so easy to see. For instance, you need to know that arrays over value classes box their elements. It also means that depending on your application, value classes could even be a lot slower than normal classes, and it requires considerable expertise from programmers to figure out the causes when that happens. 

What do we gain for the boxing pains? Essentially, it's two things:

 - A nice `toString`, e.g. we can print `1.0sec` instead of just `1.0`.
 - A way to pattern match on value classes coming from `Any`. E.g. you can write

        (x: Any) match {
          case Second(x) => ...
          case Meter(y) => ...
        }

   I believe that's all. It seems too little in return for paying the price of boxing.

2. Extension methods are dealt with much more nicely in #4114 

3. Multi-parameter value classes are not possible yet. If the JVM supports them at some point, it looks indeed like a class-like syntax could be a nice way to express them. But there are a lot of ifs: What exactly will happen on the JVM? When will it happen? What should we do on other platforms? If my target was Scala.Native (so I would not have to second-guess what happens) there are two possible alternative designs. 

One design would be as stripped down classes:

    inline class Complex(x: Double, y: Double)

No parents or body allowed. Rely on extensions to add functionality.

Another possible design would be to introduce structs directly and generically. Where with #2199 tuple syntax is used as syntactic sugar for `HList`s, we could introduce tuples with named elements as syntactic sugar for `HMap`s. Then rely on opaque types to add nominality..

So in summary I believe (1) is better served with opaque types, (2) with extension methods, and for (3) the jury is still out. 

One possible alternative would be to use evolve value classes to a non-boxing future. But since that necessarily changes the runtime behavior of value classes, migration would be a huge headache. Also, at least to me opaque types feel like a more natural way of expressing that we create a new type that is implemented in terms of some hidden representation type.

smarter:
(I'm going to stop responding here after this, this discussion is more suitable for the contributors Discourse anyway)

> That's not the point, though. They both were forks at some point. The fact that Graal is quickly getting adopted mainstream actually tells a lot about who seems to be the successor of the JVM as we know it now...

What I see is that 1) Graal is part of the mainline JDK 2) Valhalla is still actively developed. From this I infer that they will work fine together, if you find a credible source saying this isn't true I would be interested.

> I think it's important that something that is not a class should not be represented as a class. The fact that AnyVal classes are not classes is obvious from looking at what you cannot do with them:

I strongly disagree. A class with restrictions is still a class, because in all the cases where you can use it, it will behave like a class. We already have classes with restrictions (case classes cannot extend case classes, ...) and we still call them classes, because you only need to understand the semantics of classes to use them, that's the big win: less cognitive overhead.

> opaque type Int128(f: Int, f2: Int)

Try to write out the details of this and you'll see that you're just reinventing classes, but worse.

> I wholeheartedly agree, that's why we're discussing now potential ways of evolving opaque types in the future should value types come to the JVM.

I hope the SIP will contain an elaborate discussion of this. So far, I'm really not convinced by what I've seen.


> It's a concept they're already familiar with: type aliases. The familiarity between both is really high, except the visibility of the type. I don't think this is a big deal, Scala devs use private[this] or private[mypackage] all the time, they're already used to think where a term or type is visible from.

I think 99% of people will not know how to write opaque types properly and will have to copy-paste code, because it's too different from everything they're used to. (Also things get even more complicated when you have an opaque type that wraps another opaque type).



>> Looking at http://mail.openjdk.java.net/pipermail/valhalla-dev/2018-March/thread.html I see quite a lot of commits actually, so I wouldn't say there's no action on the implementation.
> 
> Can you point me to the repository? That's a link to the mailing list. Last time I checked the activity was limited.

I'm linking to the mailing-list because every commit to the repository sends a mail to the mailing-list. The repository is at http://hg.openjdk.java.net/valhalla/valhalla but is harder to judge activity from because you also see commits not specific to valhalla here.



>> Concerning the spec, the latest version I can find comes from http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2018-February/000556.html from February 2018, so I wouldn't say it's not being refined either.
>
> I read those slides two weeks ago, but those don't describe the spec nor refine the design goals and non-goals of the original document proposed by John Rose. They are discussing what seem to be implementation details, so still, the best thing we can do to discuss about this specification is the document I linked to before from May 2017.

The mail I link to contains a link to a diff for the JVMS. The "S" in "JVMS" stands for specification.

> Fair point. Just wanted to emphasize that Java->Scala compatibility is not in my book to evolve Scala, and I don't think it has been judging from the design decisions we've taken during the past years.

Hence why we should try to not make things actively worse. We have a chance to influence the future of the JVM here, and instead we're developing our own limited custom solution instead of collaborating with the rest of the JVM ecosystem, I think that's too bad.

odersky:
So, I am trying to summarize here:

Why not keep value classes? 

 - The boxing model trips people up and can get in the way of high-performance code (in particular
    because value classes cannot be stored in arrays)
 - The boxing model has surprisingly nasty consequences for code generation, including issues
    that are still not fixed in any Scala compiler (#1905)
 - The limitations are hard to remember and a bit ad hoc.

What functionality do value classes provide?

 1. A low-cost implementation of some functionality in terms of some other type
 2. A way to add extension methods
 3. (In the future, once we can support multiple parameters) A way to do structs

Opaque types address (1), with some caveats

 - They can currently not be toplevel
 - They require sometimes a bit more code than value classes for the same functionality
 - They require some concept shifts that are non-trivial to implement (e.g. companions of opaque types, with special visibility)  

For (2) we have a separate extension method proposal which is syntactically nicer than implicit value classes. 

(3) is currently not possible on the JVM, non-sensical on JS, and addressed with @struct on Native. (3) might be possible with Valhalla in the future.

One idea we should explore before deciding is this: A lightweight way to define classes that, like opaque types, do not box, but that can take multiple parameters if the platform allows it. A possible syntax could be 

    inline class C(x: T) /* no parents allowed */ { 
       def ...
       def ...
       /* no other members allowed */
    }

`inline` classes defined that way are always represented as their underlying type. Their parent is `Any` but they cannot override any of its methods. The potential advantages that I can see relative to opaque types are:

 - sometimes less boilerplate
 - can do extension methods as well when combined with `implicit`. So it would then be a separate
    decision whether we want to replace `implicit class` with `extension`.
 - can express structs / multiparameter value classes.

fanf:
For reference to @odersky last comment, value type in Valhalla are described here (and seems to now be called Value Object): http://openjdk.java.net/jeps/169 (for full ref, Project Valhalla home page is here: http://openjdk.java.net/projects/valhalla/).

An other ongoing work by Brian Goetz is "records" which are more like scala case classes (where value type are more like opaque type): 
- "Data Classes for Java": http://cr.openjdk.java.net/~briangoetz/amber/datum.html
- and a recent "Records -- current status": http://mail.openjdk.java.net/pipermail/amber-spec-experts/2018-March/000400.html

I think it is extremelly interesting to see how Brian Goetz articulate the goals and boundaries of each proposal, and see how it correlates with regards to Scala needs.

odersky:
@briangoetz Thanks for the info! I am really looking forward to see what comes out of specialization! 
Regarding value types, a question: Are multiple return values part of the proposal? It seems that this is the only blocker for is why we could not implement fully general inline classes right now.

briangoetz:
Of course, you can implement multiple return or tuples in a JVM-based language today -- if you are willing to pay the boxing penalty.  Values let you implement these features without paying the boxing penalty, so that's great.  Once we have values, we don't see a need to implement multiple return at the JVM level, because the runtime cost of packaging scalars into a value to return them across a call boundary is effectively zero, after the JIT gets its hands on it.  (Even in the interpreter, values can be returned on the stack, rather than having to put them in the heap.)  Not sure if we will want to / need to expose it as a language feature, but of course then we'll have an easy and cheap way to do it if we want.  

You might enjoy my talk from JVMLS last year about implementing pattern matching -- which grapples with many of the same issues of multiple return.  The main challenge is that with neither value types nor multiple return nor out parameters nor mutating up-stack locals nor super-awesome box elimination at the JVM level, in order to avoid the boxing penalty, you have to turn yourself inside out.  

    https://www.youtube.com/watch?v=n3_8YcYKScw&t=1s&index=1&list=PLX8CzqL3ArzXJ2EGftrmz4SzS6NRr6p2n

Subjectively, it feels like I spent about half my time working around boxing.  I am really looking forward to getting that half of my life back!

odersky:
Yes, I meant multiple return values without boxing. So, JVM values would give us that, which is exactly what we need. Two more questions, just to be clear:

 1. If V is a JVM value class, is `x instanceof V` permitted? (I would assume not, because you'd need a header for that).
 2. If V is a JVM value class, can `V` override `toString` (I would assume not, since `V` is not a subclass of `Object`).

If it's indeed two nos here then I think the `inline class` proposal would fit exactly what value classes are.

odersky:
So, the answer is (still) "it depends". Without knowing all the details, it looks like inline classes would be a better fit with MVT. Can you give me a hint what were the problems you were running in then? Even with L world, it would probably mean that inline classes simply won't support some of the functionality that would be offered by the JVM.

briangoetz:
Many of them have to do with migration compatibility.  For example, if I migrate `Foo` from a reference class to a value class, then its signature changes from `LFoo;` to `QFoo;`.  That means that any method signatures that take or dispense `Foo` also changes.  (The irony here was that the L/Q divide was supposed to be a path to migration, since old bytecode that said LFoo would be interpreted as wanting a boxed QFoo, but it only got us part of the way in the migration story.  Bridge methods helped, but not enough.)  There was also no non-terrible story for how values could implement interfaces without boxing, which is no fun when you want to express things like `Comparable` on numerics.  Relatedly, there also was no way at the VM level to express "x is a Foo, which might be unboxed or boxed".  

The latter question led us down a path ("U types"), which had surprising commonality with SPJ's recent paper on Levity Polymorphism in Haskell.  

I think you'll find that you can model what you want in L-world reasonably well.  As you say, you don't have to expose all the features the VM gives you.

fanf:
@odersky there was a report talking about that here: "abandon all U-types, welcome to L-world (or, what I learned in Burlington)": http://mail.openjdk.java.net/pipermail/valhalla-spec-observers/2017-November/000388.html

odersky:
Migration is always a pain, isn't it? As to implementing interfaces I believe it is best to sidestep the issue and provide conversions from a value type to an interface instead (explicit or implicit). In Scala we tried to implement some interfaces (so called "universal traits") with value classes but it was a pain since it led to boxing anyway. In retrospect, if you must do boxing you might as well use a conversion.

Thanks for the references! I'll check them out.

fanf:
This cr from John Rose was recently published:
http://cr.openjdk.java.net/~jrose/values/value-type-hygiene.html

8<-----
Value Type Hygiene
May 2018 (v. 0.1)
John Rose and the Valhalla Expert Group

Briefly put, types in L-world are ambiguous, leading to unhygienic mixtures
of value operations with reference operations, and uncontrolled pollution
from nulls infecting value code.

This note explores a promising proposal for resolving the key ambiguity. It
is a cleaner design than the ad hoc mechanisms tried so far. The resulting
system would seem to allow more predictable and debuggable behavior, a
stronger backward compatibility story, and better optimization.
8<-----

It's a bit long and deals with compatibility and internal details I'm not
sure I wanted to know anything about, but I believe it's a great summary of
the state of the art on the subject.

odersky:
Rebased over add-transparent

panacekcz:
I was experimenting with this feature and came across what seems to be a bug in the implementation:
When the opaque type is defined by a type parameter, it is not erased to the actual type argument in method return type positions.
For example: 
```scala
class A
object OpaqueReturnA{
  opaque type T[X] = X
  def a(i:T[A]):T[A] = i
}
```
Results in:
```scala
def a(i: A): Object = i
```
Expected:
```scala
def a(i: A): A = i
```

If the type argument is a primitive, then this leads to boxing:
```scala
object OpaqueReturnInt{
  opaque type T[X] = X
  def a(i:T[Int]):T[Int] = i
}
```
Results in:
```scala
def a(i: Int): Object = scala.Int.box(i)
```

That is contrary to the expectation that opaque types will never box.

allanrenucci:
Thanks @panacekcz for the bug report. This is an issue with type aliases in general. I have opened #5090

panacekcz:
Another thing: the types are not opaque for overriding. That looks like a bug because it allows wrapping and unwrapping outside the companion object, like this:
```scala
object OpaqueEscape{
  opaque type Wrapped = Int
  abstract class EscaperBase {
    def unwrap(i:Wrapped):Int
    def wrap(i:Int):Wrapped
  }
  class Escaper extends EscaperBase{
    override def unwrap(i:Int):Int = i
    override def wrap(i:Int):Int = i
  }

  val e = new Escaper:EscaperBase
  val w:Wrapped = e.wrap(1)
  val u:Int = e.unwrap(w)
}
```
Of course one can imagine such wrapping or unwrapping happen unintentionally.

I'd expect this would be forbidden. However, the SIP does not mention overriding - perhaps it should be updated to explicitly state how that works.
Probably it makes sense to allow it when overriding Java methods.

jvican:
For all the SIP is concerned, overriding just follows the normal typechecking procedure. This is indeed a bug because the compiler should not typecheck the overriding unwrap.

odersky:
Rebased to latest master, with significant changes to make from-tasty compilation work. Comments were reorganized into logical steps, as opposed to history.

odersky:
Opaque types have very tricky interactions with the compiler's sophisticated caching strategy. The danger is that some cached information leaks between companion objects of opaque types and the outside world. 41791f7 tries to improve the situation. 

We also have to keep in mind that the same problems arise for all GADT constrained type variables, not just for opaque types. That means changing the opaque type spec, so that instead of GADT equalities we define (say) conversions between an opaque type and its alias gains us nothing, and in fact reduces test coverage.

odersky:
Reverting to wip status until #5254 is resolved.

odersky:
If we can't rely on GADTs to represent the equality between opaque type and its right hand side in some contexts, but not in others, we need a fallback scheme. The most straightforward one would be to make conversions available. I.e the compiler could provide in the companion object of an opaque type a pair of private conversion methods `reveal` from opaque type to its rhs and `inject` in the other direction. I.e. as a first approximation, in the `Logarithm` example, the compiler would
generate in `object Logarithm` the two methods
```scala
private def reveal(x: Logarithm): Double
private def inject(x: Double): Logarithm
```
This is still too cumbersome, as one could not easily map a `List[Double]` to a `List[Logarithm]`, nor an `Array[Logarithm]` to an `Array[Double]`. 

A more convenient solution treats `reveal` and `inject` calls as compiler magic. Each

 - requires its argument to be fully typed
 - obtains its result type by applying the map between opaque type and its rhs to all parts of its argument (using a `TypeMap`).

That would be straighforward to use and straightforward to implement. The only downside is that it is still a bit verbose.

johnynek:
You could have an `=:=[Opaque, Original]` private val in the object. It is a bit of a pain to use but it has substitution so it can deal with the List example you mention.

odersky:
> You could have an =:=[Opaque, Original] private val in the object. It is a bit of a pain to use but it has substitution so it can deal with the List example you mention.

Can you explain that in more detail? I am not sure how to use `=:=` for the purpose you have in mind.

johnynek:
so, in your example, you have:
```scala
opaque type ImutableArray[+A] = Array[A]
object ImmutableArray {
  def sum[A: Numeric](a: ImmutableArray[A]): A = ...
}

opaque type Log = Double

object Log {
  private implicit val reveal: Log =:= Double = ... // compiler generated, but could also be a cast of `refl`

  // a product in the original space, is sum in log space
  def product(l: ImmutableArray[Log]): Log = { 
    val p = ImmutableArray.sum(reveal.substituteCo(l))
    reveal.flip(p)
  }
}
```

johnynek:
https://github.com/scala/scala/blob/e57997a9cd1d7e73792cfa1a49eff6e4d7848932/src/library/scala/Predef.scala#L668

johnynek:
PS: immutable array is one I'm really excited about. We can have an `ImmutableArray[Log]` but get the same unboxed performance as `Array[Double]`.

odersky:
@johnynek Ah I see what you mean now. That works to some degree but I also think it can be a pain to use. 

I agree that immutable array is one the most exciting parts of this effort.

LPTK:
PPS: the above is for covariant and contravariant substitution. For substitution in invariant places, one just needs `type Ev >: A | B <: A & B` as a member of `=:=`.

Here is a [complete example (Scastie)](https://scastie.scala-lang.org/Sgj4ybeyQCC3Gl4anxjRAg) of an opaque type implemented using my strategy:

```scala
abstract class OpaqueAPI {
  type S  // abstract type
  protected implicit val ev: S =:= String  // not visible from outside
  
  def in (xs: Set[String]): Set[S] = xs:Set[ev.Ev]
  def out(xs: Set[S]): Set[String] = xs:Set[ev.Ev]
  
  val arr0: Array[String] = Array("a")
  val arr1: Array[S] = Main.foo(arr0)
}
object Main {
  def foo[A,B](arr: Array[A])(implicit sub: A =:= B): Array[B] = arr:Array[sub.Ev]
  
  val Opaque: OpaqueAPI = new OpaqueAPI {
    type S = String  // not visible from outside
    val ev: S =:= String = implicitly
  }
  import Opaque._
  
    val s0 = Set("a")
    val s1: Set[S] = in(s0)
    println(s1)
    val s2: Set[String] = out(s1)
    println(s2)
    
    println(s"Done.")
  }
}
```

...where we have:

```scala
abstract class <:< [-A,+B] { type Ev >: A <: B }
object <:< {
  implicit def ev[A<:B,B]: A <:< B = new { type Ev = A }
}
abstract class =:= [A,B] { type Ev >: A | B <: A & B }
object =:= {
  implicit def ev[A>:B<:B,B]: A =:= B = new { type Ev = A }
  implicit def flip[A,B](implicit ev: A =:= B): B =:= A = new { type Ev = ev.Ev }
}
```

odersky:
I revived the original approach since #5254 is now resolved. The main difference is that, in light of the closing comment of #5254, we no longer assume all the properties of a full type alias in the opaque type companion. So some tests have to be changed, as is shown in 207c380.

odersky:
We found a much better way to get opaque types without restrictions and without too many contortions in the compiler: #5300

PR #4036, opened 2018-02-26
Fix #4031: Check arguments of dependent methods for realizability

### LINKED ISSUE #4031: Good old unsoundness, CCE
http://io.livecode.ch/learn/namin/unsound/scala

```scala
object App {
  trait A { type L >: Any}
  def upcast(a: A, x: Any): a.L = x
  lazy val p: A { type L <: Nothing } = p
  def coerce(x: Any): Int = upcast(p, x)

    println(coerce("Uh oh!"))
  }
}
```
`lazy` above is not necessary

```scala
	at App$.coerce(HelloWorld.scala:5)
	at App$.main(HelloWorld.scala:8)
	at App.main(HelloWorld.scala)
```

odersky:
Also some additions to realizability tests


odersky:
I now think we should try to avoid this split into isStable (checked in Typer) and checkRealizable in PostTyper, at least as far as lazy vals are concerned. We should try to test for realizibility when checking whether a lazy val is stable. If this runs into cycles, we can always back out by being conservative, i.e. assume non-stable in case of cycles. I think that's at least worth a try. The alternatives look all even less appealing.

Blaisorblade:
Treating lazy vals as unstable here would already be an improvement. Is “running into cycles” insensitive to compilation order/separate compilation, or does it have a suitable conservative approximation?

Beyond “what we can implement” we should probably think of what we can specify and explain to an (advanced) user.

tests/neg/i1050c.scala
```diff
@@ -4,7 +4,7 @@ object Import {
     trait B { type L >: Any}
     trait U {
       lazy val p: B
-      locally { val x: p.L = ??? } // old-error: nonfinal lazy
+      locally { val x: p.L = ??? } // error: nonfinal lazy
       locally {
         import p._ // error: Import.B(U.this.p) is not a legal path
         val x: L = ??? // old-error: nonfinal lazy
```
Blaisorblade: Probably this should also be re-enabled, but at least the `import p._` fails.

tests/neg/i1050c.scala
```diff
@@ -4,7 +4,7 @@ object Import {
     trait B { type L >: Any}
     trait U {
       lazy val p: B
-      locally { val x: p.L = ??? } // old-error: nonfinal lazy
+      locally { val x: p.L = ??? } // error: nonfinal lazy
```
Blaisorblade: `p.L` has bad bounds, and this test was disabled (by mistake) by unrelated PR #1106 (probably because it had broken), so re-enabling this test.

Since it isn't about a method call it's not addressed by the current fix, but it's still related to realizability checks I guess.

odersky: No, it was because the tests in 1050c are done during typer and the non-final lazy is detected later in PostTyper, so not detected when there are typer errors. The comment was misleading.
There are tests like this in i1050.scala, which get reported correctly.

odersky:
test performance please

odersky:
Last commit should fix the broken build.

Blaisorblade:
Cherry-picked last commit in #4141.

compiler/src/dotty/tools/dotc/typer/TypeAssigner.scala
```diff
@@ -337,13 +338,31 @@ trait TypeAssigner {
...
-  def safeSubstParam(tp: Type, pref: ParamRef, argType: Type)(implicit ctx: Context) = {
+  def safeSubstParam(tp: Type, pref: ParamRef, argType: Type, initVariance: Int = 1)(implicit ctx: Context) = {
     val tp1 = tp.substParam(pref, argType)
-    if ((tp1 eq tp) || argType.isStable) tp1
-    else tp.substParam(pref, SkolemType(argType.widen))
+    if ((tp1 eq tp) || argType.isRealizable) tp1
+    else {
+      val widenedArgType = argType.widen
+      if (realizability(widenedArgType) == Realizable)
+        tp.substParam(pref, SkolemType(widenedArgType))
```
Blaisorblade: `widenedArgType` can be realizable when `argType` isn't if argType is a lazy/erased value.
If `argType` is a normal value it's realizable always, even if `widenedArgType` isn't.

If neither is realizable, one could try harder, with an even wider type, but we don't have a reliable way to get at it — there are wider types (such as `Any`) but it's unclear there's a _smallest_ wider type, since the class hierarchy is a DAG.

tests/pos/z1720.scala
```diff
@@ -3,7 +3,7 @@ package test
 class Thing {
   def info: Info[this.type] = InfoRepository.getInfo(this)
   def info2: Info[this.type] = {
-    def self: this.type = this
+    val self: this.type = this
```
Blaisorblade: So it seems that `this.type` should be as realizable as `this`, no?

odersky:
Based on #4108, because that way we could remove a puzzler in `derivedSelect`.

tests/pos/z1720.scala
odersky: Yes, the change to `val` is no longer needed.

odersky: Yes you are right. I added a further tweak to support this in 8389bea

Blaisorblade:
Should rebase on top of #4108 (now merged). Can try next week.

compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
```diff
@@ -72,16 +72,18 @@ class CheckRealizable(implicit ctx: Context) {
...
+        else {
+          val r =
+            if (!sym.isStable) NotStable
+            else if (!isLateInitialized(sym)) realizability(tp.prefix)
+            else if (!sym.isEffectivelyFinal) new NotFinal(sym)
+            else realizability(tp.info).mapError(r => new ProblemInUnderlying(tp.info, r))
+          if (r == Realizable) sym.setFlag(Stable)
+          r
+        }
+      if (r == Realizable || tp.info.isStableRealizable) Realizable else r
```
Blaisorblade: `r == Realizable || ` can be dropped without changing meaning. In fact, it's unclear why we should check `tp.info.isStableRealizable` at this point — if this is necessary, it seems to warrant a comment. Maybe this subtly depends on `setFlag` influencing `isStableRealizable`.

Should we check `tp.info.isStableRealizable` first and skip all the other expensive computations? That doesn't set the `sym.is(Stable)` cache, but it can be set on request the next time.*

Otherwise, should `tp.info.isStableRealizable` be used when computing `r`?

Blaisorblade: Playing with this myself.

Blaisorblade:
Status: I modified the checking algorithm (for small optimizations and to cache everything cacheable) but it should check the same "typing rules" — except for any undiscovered bugs due to internal mutations/cycles/...

I removed a code which seemed to repeat a call to `realizability(tp.info)`, but I'm not sure if anything could actually trigger it before.

The changes might make the code faster in some cases — but the reason I made them was to make the code IMHO more sensible: now symbols are marked as `is(Stable)` in _any_ of the conditions that show the symbol is `Stable`.

TODO: urgently rename `isStable` to something else.

Blaisorblade:
The code in this PR is as good as I can make it I think (the comments are far from perfect but hopefully not wrong, I'm thinking of moving them around — some info might belong to `TermRef` or `NamedType` really).

@smarter Should I get my changes reviewed by you for coding questions before going by Martin?

compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
```diff
@@ -68,17 +68,44 @@ class CheckRealizable(implicit ctx: Context) {
...
+          else
+            // 4. If the symbol is effectively final, and a lazy or erased val
+            // and has a realizable type. We require finality because overrides
+            // of realizable fields might not be realizable.
+
+            // Since patchRealizability checks realizability(tp.info) through
+            // isStableRealizable, using patchRealizability wouldn't make
+            // a difference, and calling it here again might introduce
+            // a slowdown exponential in the prefix length.
+            realizability(tp.info).mapError(r => new ProblemInUnderlying(tp.info, r))
```
Blaisorblade: Wondering why the prefix need not be checked here...

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -297,7 +297,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         thirdTry
       case tp1: TypeParamRef =>
         def flagNothingBound = {
-          if (!frozenConstraint && tp2.isRef(defn.NothingClass) && state.isGlobalCommittable) {
+          if (!frozenConstraint && tp2.isRef(NothingClass) && state.isGlobalCommittable) {
```
Blaisorblade: XXX This commit looks stale after the merge of kind-polymorphism, almost just a refactoring is left...

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -379,8 +379,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         if (cls2.isClass) {
           if (cls2.typeParams.isEmpty) {
             if (cls2 eq AnyKindClass) return true
-            if (tp1.isRef(defn.NothingClass)) return true
+            if (tp1.isRef(NothingClass)) return true
             if (tp1.isLambdaSub) return false
+            if (cls2 eq AnyClass) return true
```
Blaisorblade: Actual behavior change. But `AnyKindClass` was tested first, gotta check if this is still needed.

Blaisorblade:
Martin mentioned he will look at this once I assign him, so I should decide what to do about the small issues I wanted to fix.

Blaisorblade:
TODO: check this PR fixes https://twitter.com/propensive/status/1031419434320179200?s=21:
```scala
implicitly[(A with B)#Z =:= Int]
implicitly[(A with B)#Z =:= Char]
implicitly[(A with B)#Z =:= (B with A)#Z]
```

Blaisorblade:
Closed in favor of #5558 and #5559.

PR #4038, opened 2018-02-26
Fix #4023: Ignore healed implicit tag at level 0

### LINKED ISSUE #4023: Cannot parameterize inline macro
```
import scala.quoted._
object Macro {
  inline def ff[T](x: T): T = ~impl[T]('(x))
  def impl[T](x: Expr[T]): Expr[T] = x
}
```
fails with
```
3 |  inline def ff[T](x: T): T = ~impl[T]('(x))
  |                                    ^
  |              access to type T from wrong staging level:
  |               - the definition is at level 1,
  |               - but the access is at level 0.
  |              
  |               The access would be accepted with the right type tag, but
  |               no implicit argument of type scala.quoted.Type[T] was found
```

But if we add the implicit `quoted.Type[T]` it fails with 
```
3 |  inline def ff[T: Type](x: T): T = ~impl[T]('(x))
  |                                     ^
  |                      access to value evidence$1 from wrong staging level:
  |                       - the definition is at level 1,
  |                       - but the access is at level -1.
```

nicolasstucki:
Fixes #4023 by making phase inconsistent types references in level 0 (of a macro) phase consistent without. This type will be phase consistent at the inline site.

### DIFF
--- a/compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
+++ b/compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
@@ -69,7 +69,7 @@ class Interpreter(implicit ctx: Context) {
 
     tree match {
       case Quoted(quotedTree) =>
-        if (tree.isTerm) new scala.quoted.Exprs.TreeExpr(quotedTree)
+        if (quotedTree.isTerm) new scala.quoted.Exprs.TreeExpr(quotedTree)
         else new scala.quoted.Types.TreeType(quotedTree)
 
       case Literal(Constant(c)) => c.asInstanceOf[Object]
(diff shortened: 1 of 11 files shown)


compiler/src/dotty/tools/dotc/interpreter/Interpreter.scala
```diff
@@ -129,7 +128,7 @@ class Interpreter(implicit ctx: Context) {
       case _ =>
         // TODO Add more precise descriptions of why it could not be interpreted.
         // This should be done after the full interpreter is implemented.
-        throw new StopInterpretation(s"Could not interpret ${tree.show}. Consider extracting logic into a helper def.", tree.pos)
+        throw new StopInterpretation(s"Could not interpret ${tree.show}. Consider extracting logic into a helper def.\n$tree\n", tree.pos)
```
nicolasstucki: Need to revert this change

PR #4039, opened 2018-02-26
Rename isDependent

### LINKED ISSUE #4031: Good old unsoundness, CCE

odersky:
isDependent -> isResultDependent

so that `isDependent` now is implemented as

    isResultDependent || isParamDependent

We got bitten by this when trying to fix #4031.


Blaisorblade:
Anything against also adding the new isDependent you described? I also wanted to rename isStable to isApproxStable and write down what you explained, can I go ahead?

PR #4045, opened 2018-02-27
Fix #3324: add `isInstanceOf` check

### LINKED ISSUE #3324: Missing unchecked warning for isInstanceOf checks
This code snippet emits a warning with scalac but not with Dotty
```scala
class Foo {
  def foo(x: Any): Boolean =
    x.isInstanceOf[List[String]]
}
```
```scala
Foo.scala:3: warning: non-variable type argument String in type List[String] (the underlying of List[String]) is unchecked since it is eliminated by erasure
    x.isInstanceOf[List[String]]
                  ^
one warning found
```

liufengyun:
Fix #3324: add `isInstanceOf` check.

This specification is different from the [scalac compiler implementation](https://github.com/scala/scala/blob/94cee79eb63b4449d21dc74cb9175d6412cc8a55/src/compiler/scala/tools/nsc/typechecker/Checkable.scala#L12-L64). I hope it's more understandable.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
@@ -1,0 +1,84 @@
+package dotty.tools.dotc
+package transform
+
+import util.Positions._
+import MegaPhase.MiniPhase
+import core._
+import Contexts.Context, Types._, Decorators._, Symbols._, typer._
+import TypeUtils._, Flags._
+import config.Printers.{ transforms => debug }
+
+/** check runtime realizability of type test
+ */
+class IsInstanceOfChecker extends MiniPhase {
+
+  import ast.tpd._
+
+  val phaseName = "isInstanceOfChecker"
+
+  override def transformTypeApply(tree: TypeApply)(implicit ctx: Context): Tree = {
+    def ensureCheckable(qual: Tree, pt: Tree): Tree = {
+      if (!Checkable.checkable(qual.tpe, pt.tpe))
+         ctx.warning(
+           s"the type test for ${pt.show} cannot be checked at runtime",
+           tree.pos
+         )
+
+      tree
+    }
+
+    tree.fun match {
+      case fn: Select if fn.symbol == defn.Any_typeTest =>
+        ensureCheckable(fn.qualifier, tree.args.head)
+      case fn: Select if fn.symbol == defn.Any_isInstanceOf =>
+        ensureCheckable(fn.qualifier, tree.args.head)
+      case _ => tree
+    }
+  }
+}
+
+object Checkable {
+  import Inferencing._
+  import ProtoTypes._
+
+  /** Whether `(x:X).isInstanceOf[P]` can be checked at runtime?
+   *
+   *  The following cases are not checkable at runtime:
+   *
+   *  1. if `P` refers to an abstract type member
+   *  2. if `P` is `pre.F[Ts]` and `pre.F` refers to a class:
+   *     (a) replace `Ts` with fresh type variables `Xs`
+   *     (b) instantiate `Xs` with the constraint `pre.F[Xs] <:< X`
+   *     (c) `pre.F[Xs] <:< P` doesn't hold
+   *  3. if `P = T1 | T2` or `P = T1 & T2`, checkable(X, T1) && checkable(X, T2).
+   */
+  def checkable(X: Type, P: Type)(implicit ctx: Context): Boolean = {
+    def Psym = P.dealias.typeSymbol
+
+    def isAbstract = !Psym.isClass
+
+    def isClassDetermined(tpe: AppliedType) = {
+      val AppliedType(tycon, args) = tpe
+      val tvars = tycon.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
+      val P2 = tycon.appliedTo(tvars)
+
+      debug.println("P2 : " + P2)
+      debug.println("X : " + X)
+
+      !(P2 <:< X.widen) || {
+        val syms = maximizeType(P2, Psym.pos, fromScala2x = false)
+        val res = P2 <:< P
+        debug.println("P2: " + P2.show)
+        debug.println("P2 <:< P = " + res)
+        res
+      }
+    }
+
+    P match {
+      case tpe: AppliedType  => !isAbstract && isClassDetermined(tpe)
+      case AndType(tp1, tp2) => checkable(X, tp1) && checkable(X, tp2)
+      case OrType(tp1, tp2)  => checkable(X, tp1) && checkable(X, tp2)
+      case _                 => !isAbstract
+    }
+  }
+}
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
@@ -105,7 +105,9 @@ object Checkable {
         val bounds = ctx.typerState.constraint.entry(tvar.origin)
         if (bounds.loBound =:= bounds.hiBound)
           tvar.instantiateWith(bounds.loBound)
-        else {
+        else if (tycon.classSymbol.is(Final))  // 3324g.scala cannot happen because of final
+          instantiateSelected(P1, tvar :: Nil)
+        else {                                 // see 3324g.scala
           val wildCard = ctx.newSymbol(ctx.owner, WildcardParamName.fresh().toTypeName, Case, tvar.origin.underlying, coord = pos)
           tvar.instantiateWith(wildCard.typeRef)
         }
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
@@ -43,11 +43,18 @@ object Checkable {
 
   /** Whether `(x:X).isInstanceOf[P]` can be checked at runtime?
    *
-   *  Replace `T @unchecked` and pattern binder types (e.g., `_$1`) in P with WildcardType, then check:
+   *  First do the following substitution:
+   *  (a) replace `T @unchecked` and pattern binder types (e.g., `_$1`) in P with WildcardType
+   *  (b) replace pattern binder types (e.g., `_$1`) in X:
+   *      - variance = 1  : hiBound
+   *      - variance = -1 : loBound
+   *      - variance = 0  : OrType(Any, Nothing)
+   *
+   *  Then check:
    *
    *  1. if `X <:< P`, TRUE
    *  2. if `P` is a singleton type, TRUE
-   *  3. if `P` refers to an abstract type member or type parameter, `X <:< P`
+   *  3. if `P` refers to an abstract type member or type parameter, FALSE
    *  4. if `P = Array[T]`, checkable(E, T) where `E` is the element type of `X`, defaults to `Any`.
    *  5. if `P` is `pre.F[Ts]` and `pre.F` refers to a class which is not `Array`:
    *     (a) replace `Ts` with fresh type variables `Xs`
@@ -60,12 +67,23 @@ object Checkable {
   def checkable(X: Type, P: Type)(implicit ctx: Context): Boolean = {
     def isAbstract(P: Type) = !P.dealias.typeSymbol.isClass
 
-    def replaceBinderMap(implicit ctx: Context) = new TypeMap {
+    def replaceP(implicit ctx: Context) = new TypeMap {
       def apply(tp: Type) = tp match {
         case tref: TypeRef
         if !tref.typeSymbol.isClass && tref.symbol.is(Case) => WildcardType
         case AnnotatedType(_, annot)
         if annot.symbol == defn.UncheckedAnnot => WildcardType
+        case _ => mapOver(tp)
+      }
+    }
+
+    def replaceX(implicit ctx: Context) = new TypeMap {
+      def apply(tp: Type) = tp match {
+        case tref: TypeRef
+        if !tref.typeSymbol.isClass && tref.symbol.is(Case) =>
+          if (variance == 1) tref.info.hiBound
+          else if (variance == -1) tref.info.loBound
+          else OrType(defn.AnyType, defn.NothingType)
         case _ => mapOver(tp)
       }
     }
@@ -85,6 +103,7 @@ object Checkable {
       val res  = isFullyDefined(P1, ForceDegree.noBottom) && P1 <:< P
       debug.println("P1 : " + P1)
       debug.println("P1 <:< P = " + res)
+
       res
     }
 
@@ -105,7 +124,7 @@ object Checkable {
       case _                    => true
     })
 
-    val res = recur(X.widen, replaceBinderMap.apply(P))
+    val res = recur(replaceX.apply(X.widen), replaceP.apply(P))
 
     debug.println(i"checking  ${X.show} isInstanceOf ${P} = $res")
 
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
@@ -28,9 +28,8 @@ class IsInstanceOfChecker extends MiniPhase {
     }
 
     tree.fun match {
-      case fn: Select if fn.symbol == defn.Any_typeTest =>
-        ensureCheckable(fn.qualifier, tree.args.head)
-      case fn: Select if fn.symbol == defn.Any_isInstanceOf =>
+      case fn: Select
+      if fn.symbol == defn.Any_typeTest || fn.symbol == defn.Any_isInstanceOf =>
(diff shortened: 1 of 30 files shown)


tests/neg-custom-args/isInstanceOf/Result.scala
```diff
@@ -0,0 +1,20 @@
+import scala.util.control.NonFatal
+object p {
+
+  enum Result[+T, +E] {
+    case OK [T](x: T) extends Result[T, Nothing]
+    case Err[E](e: E) extends Result[Nothing, E]
+  }
```
liufengyun: Add a comment why this test?

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -152,7 +153,7 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
      *  def equals(that: Any): Boolean =
      *    (this eq that) || {
      *      that match {
-     *        case x$0 @ (_: C) => this.x == this$0.x && this.y == x$0.y
+     *        case x$0 @ (_: C @unchecked) => this.x == this$0.x && this.y == x$0.y
```
liufengyun: Add comment warning maintainers this code is fragile, since unchecked might hide type errors...

tests/neg-custom-args/isInstanceOf/enum-approx2.scala
```diff
@@ -0,0 +1,9 @@
+sealed trait Exp[T]
+case class Fun[A, B](f: Exp[A => B]) extends Exp[A => B]
+
+class Test {
+  def eval[T](e: Exp[T]) = e match {
+    case Fun(x: Fun[Int, Double]) => ???          // error
+    case Fun(x: Exp[Int => String]) => ???        // error
+  }
```
Blaisorblade: Looking at this with @liufengyun, we realize this complication has to do with compiling a match like:

```scala
def eval[T](e: Exp[T]) = e match {
  case f @ Fun(x: Fun[Int, Double]) => ???
}
```

The correct desugaring is a bit surprising to me, because we want to ensure `f: Fun[Int, Double]`, but this is not justified where `f` is bound but only after the inner match.

```scala
def eval[T](e: Exp[T]) = e match {
  case f1: Fun[a1, b1] =>
    f1.f match {
      case f2: Fun[Int, Double] =>
        //***Here*** we learn that a1 = Int and b1 = Double
        //So only ***here*** we can bind `f` with the correct type.
        val f: Fun[Int, Double] = f1
    }
}
```

We suspect the typechecker currently refines type variables `a1` and `b1` too soon, which has been fine until now, but strictly speaking means (I think) the current typechecker output is not well-typed in a sound typesystem.

smarter:
Haven't looked at what this PR does in details, but one thing to note is that right now the GADT bounds are stored separately in typer and not preserved in subsequent phases. There used to be code in PostTyper to do this that got removed (I don't know why), you may need to bring it back: https://github.com/lampepfl/dotty/commit/eb21d24043cc285238b7bb41518c777dc16af28c#diff-21d08d8a179119a483e67f9f0f994e99L62

liufengyun:
Thanks for the tip @smarter . Upon reflection, it seems to correctly handle GADTs in patterns, we need some _context-sensitive typing_, the type of a term name or type name could be refined in a context depending on type test. The refinement should be in sync with the type test.

In TypeScript, there's a related feature called [_type guards_](https://www.typescriptlang.org/docs/handbook/advanced-types.html), which can refine the type of a variable in a context:

```TypeScript
// Type is 'SpaceRepeatingPadder | StringPadder'
let padder: Padder = getRandomPadder();

if (padder instanceof SpaceRepeatingPadder) {
    padder; // type narrowed to 'SpaceRepeatingPadder'
}
if (padder instanceof StringPadder) {
    padder; // type narrowed to 'StringPadder'
}
```

I've no idea how this works in theory. In our case, we need to refine the type of types names (as @Blaisorblade points out in the example) instead of / (in addition to) term names.

liufengyun:
An example related to GADT typing of Dotty trees:

```Scala
class Test {
  trait Type
  class Tree[-T]
  case class Ident[-T](x: String) extends Tree[T]

  trait Base extends Tree[Type]
  class DummyIdent extends Ident[Null]("hello") with Base

  def foo(tree: Tree[Type]): Ident[Type] = tree match {
    case id @ Ident(x) => id                   // should we emit warnings here?
  }

  foo(new DummyIdent)
}
```

Blaisorblade:
@liufengyun `DummyIdent` should be forbidden by the PR for #3989 — as I said, I'd wait for that PR before trying to get right variant GADTs.

> I've no idea how this works in theory

TypeScript has flow-sensitive typing; in Haskell's internal language used for GADTs, IIUC, matches produce *evidence* that some type is equal or smaller than another (=:= and <:<). Doing that properly would require adapting the theory though, which isn't an immediate step.

Blaisorblade:
I just confirmed that `scalac`'s established behavior with `isInstanceOf` and `ClassTag` is likely incorrect:

https://gist.github.com/Blaisorblade/a0eebb6a4f35344e48c4c60dc2a14ce6

liufengyun:
@Blaisorblade If we always warn for a ClassTag-based test like `x: T`, it makes this feature not usable. (BTW, `ClassTag` will not generate `isInstanceOf`, but just a call to uanpply on the evidence)

I think soundness is not the only concern here. Giving that it happens rarely, I think we should optimize for usability as Scalac does.

smarter:
> @Blaisorblade If we always warn for a ClassTag-based test like x: T, it makes this feature not usable.

It's not great, but one can always use `@unchecked` to suppress the warning.

tests/neg-custom-args/isInstanceOf/3324g.scala
```diff
@@ -12,7 +12,7 @@ class Test {
   }
 
   def quux[T](a: A[T]): Unit = a match {
-    case _: B[T] => // error
+    case _: B[T] => // err-or: cannot handle this for now
```
Blaisorblade: maybe replace `err-or` with `should be an error` or such?

liufengyun:
For `(x: X).isInstanceOf[P]`, scalac emits an unchecked warning when:

>We designate another type, XR, which is essentially the intersection of X and |P|, where |P| is the erasure of P.  If XR <: P, then no warning is emitted.

ref: https://github.com/scala/scala/blob/94cee79eb63b4449d21dc74cb9175d6412cc8a55/src/compiler/scala/tools/nsc/typechecker/Checkable.scala#L12-L64


However, the above specification cannot explain why we should not produce a warning in the following case:

```Scala
sealed trait A[T]
class B[T] extends A[T]

def g(x: A[Int]) = x match { case _: B[Int] => }
```

The specification in this PR can be seen as making the obscurity in Scalac specification clear by resorting to type inference. Otherwise, the two are equivalent in spirit.

allanrenucci:
test performance please

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -0,0 +1,133 @@
...
+   *  7. if `P` is a refinement type, FALSE
+   *  8. otherwise, TRUE
+   */
+  def checkable(X: Type, P: Type, pos: Position)(implicit ctx: Context): Boolean = {
+    def isAbstract(P: Type) = !P.dealias.typeSymbol.isClass
+
+    def replaceP(implicit ctx: Context) = new TypeMap {
+      def apply(tp: Type) = tp match {
+        case tref: TypeRef
+        if !tref.typeSymbol.isClass && tref.symbol.is(Case) => WildcardType
```
allanrenucci: Can you explain what this test does?

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -0,0 +1,133 @@
...
+   *      - variance = 1  : hiBound
+   *      - variance = -1 : loBound
+   *      - variance = 0  : OrType(Any, Nothing) // TODO: use original type param bounds
+   *
+   *  Then check:
+   *
+   *  1. if `X <:< P`, TRUE
+   *  2. if `P` is a singleton type, TRUE
+   *  3. if `P` refers to an abstract type member or type parameter, FALSE
+   *  4. if `P = Array[T]`, checkable(E, T) where `E` is the element type of `X`, defaults to `Any`.
```
allanrenucci: Can you explain why there is a special treatment for Array?

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -0,0 +1,133 @@
...
+   *  5. if `P` is `pre.F[Ts]` and `pre.F` refers to a class which is not `Array`:
+   *     (a) replace `Ts` with fresh type variables `Xs`
+   *     (b) constrain `Xs` with `pre.F[Xs] <:< X`
+   *     (c) instantiate Xs and check `pre.F[Xs] <:< P`
+   *  6. if `P = T1 | T2` or `P = T1 & T2`, checkable(X, T1) && checkable(X, T2).
```
allanrenucci: Can you add a test for `&` types. For example:
```scala
trait Marker
def foo[T](x: T) = x match {
  case _: T & Marker => // no warning
  // case _: T with Marker => // scalac emits a warning
  case _ =>
}
```
Scalac emits a warning but I believe it is erroneous.
Use case:
```scala
def foo[T](xs: List[T]): List[T] = xs.collect { case x: T with Marker => x }
```

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
liufengyun: Thanks @allanrenucci for the helpful feedback, I've addressed them in the latest commit.

For `Array`,  we need a special treatment, because JVM carries type information of array elements, it's possible to test the element type at runtime `[1]`:

```Scala
scala> Array(1, 2, 3).getClass
res2: Class[_ <: Array[Int]] = class [I
// where the string "[I" is the run-time type signature for the class 
// object "array with component type int".
```

`[1]`: https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -0,0 +1,133 @@
...
+        case AnnotatedType(_, annot)
+        if annot.symbol == defn.UncheckedAnnot => WildcardType
+        case _ => mapOver(tp)
+      }
+    }
+
+    def replaceX(implicit ctx: Context) = new TypeMap {
+      def apply(tp: Type) = tp match {
+        case tref: TypeRef
+        if !tref.typeSymbol.isClass && tref.symbol.is(Case) =>
```
allanrenucci: Could be `isPatternTypeSymbol(tref.typeSymbol)`

tests/neg-custom-args/isInstanceOf/3324h.scala
```diff
@@ -0,0 +1,10 @@
+object Test {
+  trait Marker
+  def foo[T](x: T) = x match {
+    case _: (T & Marker)       => // no warning
+    // case _: T with Marker => // scalac emits a warning
```
allanrenucci: This should probably tested too

tests/neg-custom-args/isInstanceOf/3324h.scala
```diff
@@ -0,0 +1,10 @@
...
+    case _ =>
+  }
+
+  def bar(x: Any) = x.isInstanceOf[List[String]]  // error
```
allanrenucci: You could remove this line  and it `pos` test. This already tested in `i3324.scala`

tests/neg-custom-args/isInstanceOf/3324c.scala
```diff
@@ -0,0 +1,10 @@
+sealed trait A[T]
+class B[T] extends A[T]
+
+class Test {
+  def f(x: B[Int]) = x match { case _: A[Int] if true => }
+
+  def g(x: A[Int]) = x match { case _: B[Int] => }
+
+  def foo(x: Any) = x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/3324d.scala
```diff
@@ -0,0 +1,11 @@
+class Test {
+  val x: Any = ???
+
+  x match {
+    case _: List[Int @unchecked] => 5
+    case _: List[Int] @unchecked => 5
+  }
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/3324e.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  type F[X]
+
+  val z: F[T] = ???
+  z match {
+    case x: F[T] =>
+  }
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/Result.scala
```diff
@@ -0,0 +1,11 @@
+object p {
+
+  // test parametric case classes, which synthesis `canEqual` and `equals`
+  enum Result[+T, +E] {
+    case OK [T](x: T) extends Result[T, Nothing]
+    case Err[E](e: E) extends Result[Nothing, E]
+  }
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/Result.scala
```diff
@@ -0,0 +1,11 @@
+object p {
+
+  // test parametric case classes, which synthesis `canEqual` and `equals`
+  enum Result[+T, +E] {
+    case OK [T](x: T) extends Result[T, Nothing]
+    case Err[E](e: E) extends Result[Nothing, E]
```
allanrenucci: What is tested here? There is no pattern match

tests/neg-custom-args/isInstanceOf/classTag.scala
```diff
@@ -0,0 +1,13 @@
...
+object IsInstanceOfClassTag {
+  def safeCast[T: ClassTag](x: Any): Option[T] = {
+    x match {
+      case x: T => Some(x)
+      case _ => None
+    }
+  }
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/gadt.scala
```diff
@@ -0,0 +1,40 @@
...
+    case Plus(e1, e2) => eval(e1)(env) + eval(e2)(env)
+    case v: Var[T] => env(v)
+    case Lambda(x: Var[s], e) => ((y: s) => eval(e)(env + (x -> y)))
+    case App(f, e) => eval(f)(env)(eval(e)(env))
+  }
+
+  eval(exp)(Env.empty)
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/t2755.scala
```diff
@@ -0,0 +1,61 @@
...
+    println(f3(Array(1.0, -2.0, 3.0, 1.0)))
+    println(f3(Array(1.0f, 2.0f, 3.0f, -3.0f)))
+    println(f3((1 to 4).toArray map (_.toString)))
+    println(f3(new Array[Any](10)))    // should match as Array[AnyRef]
+    println(f3(Array(1L)))
+    println(f3(null))
+  }
+
+  def foo(x: Any): Boolean =
+    x.isInstanceOf[List[String]]  // error
```
allanrenucci: Remove this line. Tested in `i3324.scala`?

tests/neg-custom-args/isInstanceOf/3324h.scala
liufengyun: Thanks for the catch. This line is intended to make the test a neg test. Make it a pos test will not work, as we want to check warnings here, `-fatal-warning` is required. And it's nice to have all related tests of a feature under one folder.

tests/neg-custom-args/isInstanceOf/Result.scala
liufengyun: Here the synthesized code will generate pattern match in `canEqual` and `equals`.

liufengyun:
test performance please

liufengyun:
Merge now, so that we can fix bugs related to this feature.

PR #4049, opened 2018-02-28
Fix path to `dotty/out`

nicolasstucki:


nicolasstucki:
Tested locally, all outputs went to the correct directory.

PR #4056, opened 2018-03-01
handle binder type symbols in extractor patterns

liufengyun:
Handle binder type symbols in extractor patterns. Previously the following code will result in two warnings for `ConstNullClass` and `Identity` because of the interpolated type symbols `_$1, _$2` and `T, U` don't match:

```Scala
enum Fun[-T, +U >: Null] {
  def f: T => U = this match {
    case Identity(g) => g
    case ConstNull => (_ => null)
    case ConstNullClass(y) => (_ => null)
    case ConstNullSimple => null
  }

  case Identity[T, U >: Null](g: T => U) extends Fun[T, U]
  case ConstNull
  case ConstNullClass(x: T)
  case ConstNullSimple
}
```
This PR fixes this problem.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -6,8 +6,8 @@ import core._
 import Types._
 import Contexts._
 import Flags._
-import ast.Trees._
-import ast.tpd
+import ast._
+import Trees._
 import Decorators._
 import Symbols._
 import StdNames._
@@ -299,7 +299,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     if (res) Typ(and, true) else Empty
   }
 
-  /* Whether the extractor is irrefutable */
+  /** Whether the extractor is irrefutable */
   def irrefutable(unapp: Tree): Boolean = {
     // TODO: optionless patmat
     unapp.tpe.widen.finalResultType.isRef(scalaSomeClass) ||
@@ -307,8 +307,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       productArity(unapp.tpe.widen.finalResultType) > 0
   }
 
-  /** Return the space that represents the pattern `pat`
-   */
+  /** Return the space that represents the pattern `pat` */
   def project(pat: Tree): Space = pat match {
     case Literal(c) =>
       if (c.value.isInstanceOf[Symbol])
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -326,9 +325,9 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
         if (fun.symbol.owner == scalaSeqFactoryClass)
           projectSeq(pats)
         else
-          Prod(pat.tpe.stripAnnots, fun.tpe, fun.symbol, projectSeq(pats) :: Nil, irrefutable(fun))
+          Prod(erase(pat.tpe.stripAnnots), fun.tpe, fun.symbol, projectSeq(pats) :: Nil, irrefutable(fun))
       else
-        Prod(pat.tpe.stripAnnots, fun.tpe, fun.symbol, pats.map(project), irrefutable(fun))
+        Prod(erase(pat.tpe.stripAnnots), fun.tpe, fun.symbol, pats.map(project), irrefutable(fun))
```
Blaisorblade: To be sure: basically this throws away `_$1` and `_$2`, which is fine now but might be less fine if we had type arguments for extractor patterns (which at least I'd like eventually?)

liufengyun: If we check according to semantics, I think it's correct. I cannot come up with a counter-example. If some type cannot be checked at run-time, they get a warning instead.

Blaisorblade: Yes; since type arguments to classes/traits are erased at runtime, the only use case is for annotating what GADT inference should deduce or for binding type variables generated by GADT inference. For instance:

```scala
trait Exp[T]
case class[A, B])[f: Exp[A => B])
def eval[T](e: Exp[T])(env: Env): T = e match {
    case Fun[a, b](x, body) =>
      (v: a) => body.eval(env + (x -> v))
}
```

liufengyun: It seems the example doesn't compile at all, despite any repairing efforts.

Blaisorblade: I mentioned upfront the issue is with a potential future feature, so of course my example doesn't work yet:

>  if we *had* type arguments for extractor patterns (which at least I'd like eventually?)

PR #4057, opened 2018-03-01
Fix #2973: Check variances of class parents

### LINKED ISSUE #2973: Unsoundness due to missing variance checks for extends clause
This shouldn't compile:
```scala
class Foo[T]
class Bar[-T] extends Foo[T]
```
scalac realises that:
```scala
error: contravariant type T occurs in invariant position in type [-T]Foo[T] {
  def <init>(): Bar[T]
} of class Bar
class Bar[-T] extends Foo[T]
      ^
one error found
````

### PR DESCRIPTION
**Pull request by Blaisorblade**
I'm opening this PR to recover this "lost" fix to a soundness issue (#2973, #3856). Let's see its status.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala
+++ b/compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala
@@ -93,8 +93,8 @@ class VarianceChecker()(implicit ctx: Context) {
           this(status, tp.resultType) // params will be checked in their TypeDef or ValDef nodes.
         case AnnotatedType(_, annot) if annot.symbol == defn.UncheckedVarianceAnnot =>
           status
-        //case tp: ClassInfo =>
-        //  ???  not clear what to do here yet. presumably, it's all checked at local typedefs
+        case tp: ClassInfo =>
+          foldOver(status, tp.classParents)
         case _ =>
           foldOver(status, tp)
       }
@@ -132,14 +132,16 @@ class VarianceChecker()(implicit ctx: Context) {
           ctx.debuglog(s"Skipping variance check of ${sym.showDcl}")
         case tree: TypeDef =>
           checkVariance(sym, tree.pos)
+          tree.rhs match {
+            case rhs: Template => traverseChildren(rhs)
+            case _ =>
+          }
         case tree: ValDef =>
           checkVariance(sym, tree.pos)
         case DefDef(_, tparams, vparamss, _, _) =>
           checkVariance(sym, tree.pos)
           tparams foreach traverse
           vparamss foreach (_ foreach traverse)
-        case Template(_, _, _, body) =>
-          traverseChildren(tree)
         case _ =>
       }
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Blaisorblade** [comment]
All failures (in `test_legacy` and `test`) have to do with `tests/pos/Iterable.scala`, in particular because type param `CC` is covariant for `IterableCompanion` but invariant for `IterableImpls`, yet `IterableCompanion` extends `IterableImpls`. The extension is annotated with `@uncheckedVariance`, so I guess I must only adapt the fix to respect `@uncheckedVariance`.

> abstract class IterableCompanion[+CC[X] <: Iterable[X] with Collection[CC, X]] extends CollectionCompanion[CC] with IterableImpls[CC] @uncheckedVariance {

File: compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala [hunk 1]
```diff
@@ -132,14 +132,16 @@ class VarianceChecker()(implicit ctx: Context) {
           ctx.debuglog(s"Skipping variance check of ${sym.showDcl}")
         case tree: TypeDef =>
           checkVariance(sym, tree.pos)
+          tree.rhs match {
+            case rhs: Template => traverseChildren(rhs)
```
Blaisorblade: Here `traverseChildren` sees `parents` (in the form of a constructor call), but this `Traverser` ignores terms and only looks at definitions...

compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala
```diff
@@ -93,8 +93,8 @@ class VarianceChecker()(implicit ctx: Context) {
           this(status, tp.resultType) // params will be checked in their TypeDef or ValDef nodes.
         case AnnotatedType(_, annot) if annot.symbol == defn.UncheckedVarianceAnnot =>
           status
-        //case tp: ClassInfo =>
-        //  ???  not clear what to do here yet. presumably, it's all checked at local typedefs
+        case tp: ClassInfo =>
+          foldOver(status, tp.classParents)
```
Blaisorblade: This visit sees the parent classes, but stripped of any annotations!

compiler/src/dotty/tools/dotc/typer/VarianceChecker.scala
```diff
@@ -132,14 +132,16 @@ class VarianceChecker()(implicit ctx: Context) {
           ctx.debuglog(s"Skipping variance check of ${sym.showDcl}")
         case tree: TypeDef =>
           checkVariance(sym, tree.pos)
```
Blaisorblade: On `tests/pos/Iterable.scala` we get an error inside this call (going through `tp.classParents`), because `sym` contains the parens but without annotations!

Blaisorblade:
Closing since this is likely noise — I've a first guess why Martin got stuck on it, and it's not an issue for a beginner like me yet.

PR #4059, opened 2018-03-01
Fix #4032: Fix direction of instantiation.

### LINKED ISSUE #4032: Regression in Community Build: ScalaCheck
https://travis-ci.org/lampepfl/dotty-community-build/jobs/342277554

odersky:
Interpolation is a pretty delicate scenario. It's difficult to even say what is right and what
is wrong. In #4032 there's an unconstrained type variable that should be interpolated and it's a coin
flip whether we instantiate it to the lower or upper bound. A completely unrelated change in
#3981 meant that we instantiated the variable to the upper instead of the lower bound which
caused the program to fail. The failure was because the type variable appeared covariantly
in the lower bound of some other type variable. So maximizing the first type variable
overconstrained the second. We avoid this problem now by computing the variance of the type
variable that's eliminated in the rest of the constraint. We take this into account to
instantiate the variable so that it narrows the overall constraint the least, defaulting
again to lower bound if there is not best direction. Since the test is expensive (linear
in the constraint size), we do this only if the variable's lower bound is unconstrained.
We could do it for all non-occurring type variables, but have to see how that would affect
performance.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -313,6 +313,29 @@ object Inferencing {
 
     propagate(accu(SimpleIdentityMap.Empty, tp))
   }
+
+  private def varianceInContext(tvar: TypeVar)(implicit ctx: Context): FlagSet = {
+    object accu extends TypeAccumulator[FlagSet] {
+      def apply(fs: FlagSet, t: Type): FlagSet =
+        if (fs == EmptyFlags) fs
+        else if (t eq tvar)
+          if (variance > 0) fs &~ Contravariant
+          else if (variance < 0) fs &~ Covariant
+          else EmptyFlags
+        else foldOver(fs, t)
+    }
+    val constraint = ctx.typerState.constraint
+    val tparam = tvar.origin
+    (VarianceFlags /: constraint.uninstVars) { (fs, tv) =>
+      if ((tv `eq` tvar) || (fs == EmptyFlags)) fs
+      else {
+        val otherParam = tv.origin
+        val fs1 = if (constraint.isLess(tparam, otherParam)) fs &~ Covariant else fs
+        val fs2 = if (constraint.isLess(otherParam, tparam)) fs1 &~ Contravariant else fs1
+        accu(fs2, constraint.entry(otherParam))
+      }
+    }
+  }
 }
 
 trait Inferencing { this: Typer =>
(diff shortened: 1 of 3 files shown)


Blaisorblade:
Unclear from the description: Does this fix implement Scalac's behavior or just fixes the specific failure? In the latter case, this fix could cause regressions in other projects, right?

odersky:
@Blaisorblade scalac's inference is too different to be able to answer that question.

odersky:
test performance please

odersky:
I changed the algorithm to be less arbitrary than what was in the initial PR.

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -313,6 +313,29 @@ object Inferencing {
...
+        else foldOver(fs, t)
+    }
+    val constraint = ctx.typerState.constraint
+    val tparam = tvar.origin
+    (VarianceFlags /: constraint.uninstVars) { (fs, tv) =>
+      if ((tv `eq` tvar) || (fs == EmptyFlags)) fs
+      else {
+        val otherParam = tv.origin
+        val fs1 = if (constraint.isLess(tparam, otherParam)) fs &~ Covariant else fs
+        val fs2 = if (constraint.isLess(otherParam, tparam)) fs1 &~ Contravariant else fs1
```
Blaisorblade: @odersky We were wondering about

> We take this into account to instantiate the variable so that it narrows the overall constraint the least, defaulting again to lower bound if there is not best direction.

Are you trying to give the biggest type to the typing context? If so, do you know the variance of `tv`'s occurrences in the typing context? It could occur in either covariant or contravariant positions, and that affects the polarity of occurrences of `tvar` in `tv`'s bounds, but it doesn't look like `tvar`'s polarity is accounted for...

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -313,6 +313,29 @@ object Inferencing {
 
     propagate(accu(SimpleIdentityMap.Empty, tp))
   }
+
+  private def varianceInContext(tvar: TypeVar)(implicit ctx: Context): FlagSet = {
```
allanrenucci: Can you add documentation for this method

allanrenucci:
Merging this to fix the regression before the release.

Code in this PR will be removed in #4065, so no need to further document it.

PR #4061, opened 2018-03-02
Rename `unused` to `ghost`

nicolasstucki:


allanrenucci:
> Any chance to pick another name, maybe even before the release, since apparently nobody likes unused?

I wouldn't make such a claim. Only one person complained which for me is more a sign that the name is not that controversial.

Personally, I prefer `unused`. I would discuss this a little more before settling on a new name. Maybe create a new topic in https://contributors.scala-lang.org/

nicolasstucki:
I would change it to `ghost` as it is less controversial and does not clash whit any other concept we have in the language. It also highlights the link with phantom types. We can later create a topic on it to see which keyword should be used.

Blaisorblade:
I just merged this, also to unblock the release (let me know if I was too hasty).

For future discussion: the concept of "ghost" is [also used by Liquid Haskell](https://ucsd-progsys.github.io/liquidhaskell-blog/2013/01/28/bounding-vectors.lhs/) (and maybe @gsps's Liquid Scala), but for essentially the same thing: values that don't exist at runtime.

So maybe @gsps is also interested?

biboudis:
It is also related to F*'s Ghost monad and `erased` type. 

https://www.fstar-lang.org/papers/mumon/paper.pdf

joan38:
@allanrenucci Have you seen the poll: https://goo.gl/forms/UD44AfLUD3dkfnxs2 ?

Issue #597, opened 2015-05-22
Allow overloading a method with a signature which matches before erasure but not after


smarter:
This works in scala but not in dotty:
```scala
trait A
trait B

class Test {
  def foo(x: List[A]): Function1[A, A] = ???
  def foo(x: List[B]): Function2[B, B, B] = ???
}
```
It fails with:
```scala
[error] eras.scala:6: error: method foo is already defined as method foo: (x: scala.collection.immutable.List[A])A => A
[error]  (both definitions have the same erased type signature)
[error]   def foo(x: List[B]): Function2[B, B, B] = ???
[error]         ^
[error] one error found
j
```
Even though they don't have the same erased type signature.

The problem is that the following code in `Checking#checkNoDoubleDefs`:
```scala
if (decl.signature matches other.signature) {
```
Is run before erasure, and `matches` is defined as:
```scala
  /** The meaning of `matches` depends on the phase. If types are not erased,
   *  it means `sameParams`. Once types are erased, it means `==`, comparing parameter as
   *  well as result type parts.
   */
  final def matches(that: Signature)(implicit ctx: Context) =
    if (ctx.erasedTypes) equals(that) else sameParams(that)
```

Simply changing the code in `Checking#checkNoDoubleDefs` is probably not enough to fix that since the compiler complains:
```scala
warning: cannot merge (x: scala.collection.immutable.List[B])(B, B) => B with (x: scala.collection.immutable.List[A])A => A as members of one type; keeping only (x: scala.collection.immutable.List[B])(B, B) => B
warning: cannot merge (x: scala.collection.immutable.List[B])(B, B) => B with (x: scala.collection.immutable.List[A])A => A as members of one type; keeping only (x: scala.collection.immutable.List[B])(B, B) => B
```
And changing `Signature#matches` to always match the whole signature and not just the params breaks everything in a way I don't understand:
```scala
[info] applied type mismatch: TypeRef(TermRef(TermRef(TermRef(TermRef(TermRef(ThisType(TypeRef(NoPrefix,<root>)),scala),reflect),runtime),package),universe),TypeTag) List(TypeRef(NoPrefix,dotty$DottyPredef$typeTag$$T)), typeParams = List(), tsym = val <none>#1
[info] precomplete decls = 
[error] java.util.NoSuchElementException: head of empty list
[error]         at scala.collection.immutable.Nil$.head(List.scala:420)
[error]         at scala.collection.immutable.Nil$.head(List.scala:417)
[...]

PR #4064, opened 2018-03-02
Add error message about double definition (#1589)

jendrikw:
The old error message said "definition", but I think I should be called declaration. Please clearify this. I will update the code if needed.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2080,18 +2080,18 @@ object messages {
     val explanation = ""
   }
 
-  case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
+  case class DoubleDeclaration(decl: Symbol, previousDecl: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
     val kind = "Duplicate Symbol"
     val msg = {
-      val details = if (decl.isRealMethod && previousSymbol.isRealMethod) {
+      val details = if (decl.isRealMethod && previousDecl.isRealMethod) {
         // compare the signatures when both symbols represent methods
-        decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
+        decl.asTerm.signature.matchDegree(previousDecl.asTerm.signature) match {
           case Signature.NoMatch => ""
           case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
           case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
         }
       } else ""
-      hl"${decl.showLocated} is already defined as ${previousSymbol.showDcl} in line ${previousSymbol.pos.line + 1}." + details
+      hl"${decl.showLocated} is already defined as ${previousDecl.showDcl} in line ${previousDecl.pos.line + 1}." + details
     }
     val explanation = ""
   }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2087,11 +2087,11 @@ object messages {
         // compare the signatures when both symbols represent methods
         decl.signature.matchDegree(previousDecl.signature) match {
           /* case Signature.NoMatch => // can't happen because decl.matches(previousDecl) is checked before reporting this error */
-          case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
-          case _ /* Signature.FullMatch */ => "\nThe definitions have the same signature after erasure."
+          case Signature.ParamMatch => "\nOverloads with matching parameter types are not allowed."
+          case _ /* Signature.FullMatch */ => "\nThe definitions have matching type signatures after erasure."
         }
       } else ""
-      hl"${decl.showLocated} is already defined as ${previousDecl.showDcl} in line ${previousDecl.pos.line + 1}." + details
+      hl"${decl.showLocated} is already defined as ${previousDecl.showDcl} at line ${previousDecl.pos.line + 1}." + details
     }
     val explanation = ""
   }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2085,7 +2085,7 @@ object messages {
     val msg = {
       val details = if (decl.isRealMethod && previousDecl.isRealMethod) {
         // compare the signatures when both symbols represent methods
-        decl.asTerm.signature.matchDegree(previousDecl.asTerm.signature) match {
+        decl.signature.matchDegree(previousDecl.signature) match {
           /* case Signature.NoMatch => // can't happen because decl.matches(previousDecl) is checked before reporting this error */
           case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
           case _ /* Signature.FullMatch */ => "\nThe definitions have the same signature after erasure."
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2088,7 +2088,7 @@ object messages {
         case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
         case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
       }
-      hl"${decl.showLocated} is already defined as ${previousSymbol.showDcl} in line ${previousSymbol.pos.line}." + details
+      hl"${decl.showLocated} is already defined as ${previousSymbol.showDcl} in line ${previousSymbol.pos.line + 1}." + details
     }
     val explanation = ""
   }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2079,4 +2079,10 @@ object messages {
     }
     val explanation = ""
   }
+
+  case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
+    val kind = "Duplicate Symbol"
+    val msg = hl"$decl is already defined as $previousSymbol${previousSymbol.showExtendedLocation}"
+    val explanation = ""
+  }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2086,9 +2086,9 @@ object messages {
       val details = if (decl.isRealMethod && previousDecl.isRealMethod) {
         // compare the signatures when both symbols represent methods
         decl.asTerm.signature.matchDegree(previousDecl.asTerm.signature) match {
-          case Signature.NoMatch => ""
+          /* case Signature.NoMatch => // can't happen because decl.matches(previousDecl) is checked before reporting this error */
           case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
-          case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
+          case _ /* Signature.FullMatch */ => "\nThe definitions have the same signature after erasure."
         }
       } else ""
       hl"${decl.showLocated} is already defined as ${previousDecl.showDcl} in line ${previousDecl.pos.line + 1}." + details
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2082,7 +2082,14 @@ object messages {
 
   case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
     val kind = "Duplicate Symbol"
-    val msg = hl"$decl is already defined as $previousSymbol${previousSymbol.showExtendedLocation}"
+    val msg = {
+      val details = decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
+        case Signature.NoMatch => "" // matchDegree also returns NoMatch if one of the terms is not a method
+        case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
+        case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
+      }
+      hl"${decl.showLocated} is already defined as ${previousSymbol.showDcl} in line ${previousSymbol.pos.line}." + details
+    }
     val explanation = ""
   }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2083,11 +2083,14 @@ object messages {
   case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
     val kind = "Duplicate Symbol"
     val msg = {
-      val details = decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
-        case Signature.NoMatch => "" // matchDegree also returns NoMatch if one of the terms is not a method
-        case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
-        case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
-      }
+      val details = if (decl.isRealMethod && previousSymbol.isRealMethod) {
+        // compare the signatures when both symbols represent methods
+        decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
+          case Signature.NoMatch => ""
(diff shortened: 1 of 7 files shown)


jendrikw:
It looks like `tests/neg/singletonOrs.scala` fails. I think this has nothing to do with my changes. Since I'm not super familiar with the build pipeline, can anyone explain to me why the master branch builds ok?

Blaisorblade:
@jendrikw that tests contains double definition, and i suspect the error markers must be updated. I'm new here so I'm not sure why the error count changed—what errors do you get there? You might need to simply add extra `//error` markers for the extra errors. That these errors didn't appear before and appear now is likely a side effect simply triggered by your changes.

```
- Error: tests/neg/singletonOrs.scala:2:12 ------------------------------------
2 |   def foo: 1 | 2 = 1 // error // error
  |            ^
  |            Singleton type Int(1) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:2:16 ------------------------------------
2 |   def foo: 1 | 2 = 1 // error // error
  |                ^
  |                Singleton type Int(2) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:3:12 ------------------------------------
3 |   def bar: 3 | 4 = foo // error // error
  |            ^
  |            Singleton type Int(3) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:3:16 ------------------------------------
3 |   def bar: 3 | 4 = foo // error // error
  |                ^
  |                Singleton type Int(4) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:4:12 ------------------------------------
4 |   def foo: 1 | 2 = 1 // error // error
  |            ^
  |            Singleton type Int(1) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:4:16 ------------------------------------
4 |   def foo: 1 | 2 = 1 // error // error
  |                ^
  |                Singleton type Int(2) is not allowed in a union type
6 errors found
java.lang.RuntimeException: Nonzero exit code returned from runner: 1
	at scala.sys.package$.error(package.scala:27)
```

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -627,7 +627,7 @@ trait Checking {
             def explanation =
               if (!decl.isRealMethod) ""
               else "\n(the definitions have matching type signatures)"
-            ctx.error(em"$decl is already defined as $other$ofType$explanation", decl.pos)
+            ctx.error(DoubleDeclaration(decl, other), decl.pos)
```
Blaisorblade: maybe keep using `explanation` in the DoubleDeclaration class, or move `def explanation` there?

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -623,11 +623,7 @@ trait Checking {
         typr.println(i"conflict? $decl $other")
         if (decl.matches(other)) {
           def doubleDefError(decl: Symbol, other: Symbol): Unit = {
-            def ofType = if (decl.isType) "" else em": ${other.info}"
-            def explanation =
-              if (!decl.isRealMethod) ""
-              else "\n(the definitions have matching type signatures)"
```
Blaisorblade: You must not drop the info on matching type signatures. Either you preserve it or you figure out why it's there by walking `git blame` (which leads for instance to #597) and then realize you should preserve it.
Methods can be overloaded, but ths code reject attempts at overloading that define methods with matching signatures.

For instance, for this code from #597 we need to explain users that this sort of overload is not allowed:

```scala
trait A
trait B

class Test {
  def foo(x: List[A]): Function1[A, A] = ???
  def foo(x: List[B]): Function2[B, B, B] = ??? //error
}
```

I think this code would make a good testcase.

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -623,11 +623,7 @@ trait Checking {
...
-            ctx.error(em"$decl is already defined as $other$ofType$explanation", decl.pos)
+            ctx.error(DoubleDeclaration(decl, other), decl.pos)
```
Blaisorblade: You must not drop the info on matching type signatures. Either you preserve it or you figure out why it's there by walking `git blame` (which leads for instance to #597) and then realize you should preserve it.
Methods can be overloaded, but ths code reject attempts at overloading that define methods with matching signatures.

For instance, for this code from #597 we need to explain users that this sort of overload is not allowed:

```scala
trait A
trait B

class Test {
  def foo(x: List[A]): Function1[A, A] = ???
  def foo(x: List[B]): Function2[B, B, B] = ??? //error
}
```

I think this code would make a good testcase.

jendrikw: Thank you for your feedback. If I understand correctly, the code given in #597 should not compile. I added a test case in tests/neg/, but the tests failed because it compiles successfully.

Blaisorblade: Uh! I just wanted to say "please don't remove the message", but I should still have tried.

The behavior of `decl.matches(other)` probably changed since #597 was filed. I did more experiments and asked around to figure this out.

So, here is the current behavior:
```scala
scala> class Test {
           def foo(x: List[A]): Function1[A, A] = ???
           def foo(x: List[B]): Function1[B, B] = ??? // error
         }
3 |    def foo(x: List[B]): Function1[B, B] = ??? // error
  |        ^
  |        method foo is already defined as method foo: (x: List[A]): A => A
  |        (the definitions have matching type signatures)
```

That code fails because, while the overload is "acceptable" (you can tell at call site which one to call, I forget the right word), the two functions erase to the same "signature" (which here means JVM signature, according to @smarter).

The regression I was observing is that here "the definitions have matching type signatures" is important. I'll accept this PR as soon as that message is restored with the same check (I'm not sure why `isRealMethod` is tested, please still keep it).

Here or in further PRs, we might want some clearer message there. Scalac gives:
```scala
<console>:15: error: double definition:
def foo(x: List[A]): A => A at line 14 and
def foo(x: List[B]): B => B at line 15
have same type after erasure: (x: List)Function1
                  def foo(x: List[B]): Function1[B, B] = ??? // error
```

Instead, 
```scala
class Test {
  def foo(x: List[A]): Function1[A, A] = ???
  def foo(x: List[B]): Function2[B, B, B] = ??? //error
}
```
is supported because the overload is acceptable and they erase to different JVM signatures (because of the different return types).

Finally, this code is rejected because this overload isn't acceptable:
```scala
scala> class Test {
           def foo(x: List[A]): Function1[A, A] = ???
           def foo(x: List[A]): Function2[B, B, B] = ??? // error
         }
3 |    def foo(x: List[A]): Function2[B, B, B] = ??? // error
  |        ^
  |        method foo is already defined as method foo: (x: List[A]): A => A
  |        (the definitions have matching type signatures)
```
This should arguably give a different error:

```scala
scala> class Test {
     |            def foo(x: List[A]): Function1[A, A] = ???
     |            def foo(x: List[A]): Function2[B, B, B] = ??? // error
     |          }
<console>:15: error: method foo is defined twice;
  the conflicting method foo was defined at line 14:16
                  def foo(x: List[A]): Function2[B, B, B] = ??? // error
                      ^
```

jendrikw:
I have tried to improve the error messages by testing for the different cases. Should I put the details in `explanation`?

Also, in the error I get from running `dotc tests/neg/doubleDefinition.scala` in the sbt shell, the conflicting definitions are sais to be in line 14 and 26. These numbers come from `previousSymbol.pos.line`. Is that the right method? 15 and 27 would make more sense to me.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,17 @@ object messages {
     }
     val explanation = ""
   }
+
+  case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
+    val kind = "Duplicate Symbol"
+    val msg = {
+      val details = decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
+        case Signature.NoMatch => "" // matchDegree also returns NoMatch if one of the terms is not a method
```
Blaisorblade: This sounds plausible, but I'm not sure how it relates to `realMethod` and why that code is there — not within my "review budget" I fear. I really want to approve this PR but I'm not sure it's not introducing regressions. Finding another reviewer.

Blaisorblade: OTOH, not sure `isRealMethod` worked anyway: for `abstract class A(val a: Int) { def a: String }` on master I get:
```
scala> abstract class A(val a: Int) { def a: String }
1 |abstract class A(val a: Int) { def a: String }
  |                                   ^
  |                           method a is already defined as value a: Int
  |                           (the definitions have matching type signatures)
```

Blaisorblade:
> I have tried to improve the error messages by testing for the different cases. Should I put the details in explanation?

I want to check a few details but that looks much better! Not sure on `explanations`, I _guess/feel_ it's there to provide relevant Scala lessons, but it is optional.

The line numbers confuse me, but IIUC your PR is not changing those positions, it just didn't fix them yet? Because then I'd merge the PR; I'm not yet unqualified to help with positions, and I hear they aren't trivial in Scalac.

Blaisorblade:
For extra fun, positions in the REPL appear correct in `master`:

```scala
scala> class Test2 {
         def foo(x: List[A]): A => A = ???
         def foo(x: List[B]): Function1[B, B] = ???
         }
3 |  def foo(x: List[B]): Function1[B, B] = ???
  |      ^
  |      method foo is already defined as method foo: (x: List[A]): A => A
  |      (the definitions have matching type signatures)
```

but also 
```
pgiarrusso@tsf-436-wpa-1-015:~/git/dotty$ dotc -d out tmp/Foo.scala
-- Error: tmp/Foo.scala:28:4 -------------------------------------------
28 |	def foo(x: List[B]): Function1[B, B] = ??? // error: same jvm signature
   |	    ^
   |	    method foo is already defined as method foo: (x: List[A]): A => A
   |	    (the definitions have matching type signatures)
one error found
```

Blaisorblade:
Found it
```
  override def toString =
    if (source.exists) s"${source.file}:${line + 1}"

**Comment by Blaisorblade** [comment]
And you're right regarding declaration/definition, since turning definitions into declarations doesn't affect the errors — say:

```scala
abstract class Test3 {
	// overload with same argument type, but different return types
	def foo(x: List[A]): Function1[A, A]// = ???
	def foo(x: List[A]): Function2[B, B, B]// = ??? // error
}
```
I guess we want both in tests. And BTW we'd also want a test for when `matchDegree` returns `NoMatch`.

As a consequence, I'd say `decl` in `checkNoDoubleDefs` is named correctly, but `checkNoDoubleDefs` is misnamed.

Again, you might want to move concerns that feel too hard into separate issues/PRs, or ask for another reviewer (when they're available) — I'm new here, and learning the code with you (reviewing to help others).
And my requests are turning an "easy" PR into a tougher one (though the errors are becoming better).

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,20 @@ object messages {
     }
     val explanation = ""
   }
+
+  case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
+    val kind = "Duplicate Symbol"
+    val msg = {
+      val details = if (decl.isRealMethod && previousSymbol.isRealMethod) {
+        // compare the signatures when both symbols represent methods
+        decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
```
allanrenucci: I believe you don't need `asTerm`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,20 @@ object messages {
...
+          case Signature.NoMatch => ""
```
allanrenucci: This should never happen. I would refactor your code as:
```scala
match {
  case Signature.FullMatch => ...
  case _ /* Signature.ParamMatch */ => ...
}
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,20 @@ object messages {
...
+          case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
+          case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
```
allanrenucci: The definitions have ~~the same~~ **matching** type signatures after erasure.

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,20 @@ object messages {
...
   }
+
+  case class DoubleDeclaration(decl: Symbol, previousSymbol: Symbol)(implicit ctx: Context) extends Message(DoubleDeclarationID) {
+    val kind = "Duplicate Symbol"
+    val msg = {
+      val details = if (decl.isRealMethod && previousSymbol.isRealMethod) {
+        // compare the signatures when both symbols represent methods
+        decl.asTerm.signature.matchDegree(previousSymbol.asTerm.signature) match {
+          case Signature.NoMatch => ""
+          case Signature.ParamMatch => "\nOverloads with equal parameter types but different return types are not allowed."
```
allanrenucci: Overloads with matching parameter types are not allowed

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2079,4 +2079,20 @@ object messages {
...
+          case Signature.FullMatch => "\nThe definitions have the same signature after erasure."
+        }
+      } else ""
+      hl"${decl.showLocated} is already defined as ${previousSymbol.showDcl} in line ${previousSymbol.pos.line + 1}." + details
```
allanrenucci: ~in~ **at** line

tests/neg/singletonOrs.scala
```diff
@@ -1,6 +1,6 @@
 object Test {
    def foo: 1 | 2 = 1 // error // error
    def bar: 3 | 4 = foo // error // error
-   def foo: 1 | 2 = 1 // error // error
-   def bar: 1 = foo
+   def foo: 1 | 2 = 1 // error // error // error
```
allanrenucci: @smarter This is weird, the change seems unrelated. You might want to look into this

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
jendrikw: Do you know why this is not an enum or ADT? Was is because of performance reasons? Using Ints is very unidiomatic scala IMO.

tests/neg/singletonOrs.scala
jendrikw: I am not sure about the internals, but I'm guessing that the error about singleton types in unions are from the typer phase and the compiler exists afterwards. But that wouldn't explain the different behaviour in this branch.

Manually compiling `tests/neg/singletonOrs.scala` gives
```-- Error: tests/neg/singletonOrs.scala:2:12 ------------------------------------
2 |   def foo: 1 | 2 = 1 // error // error
  |            ^
  |            Singleton type Int(1) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:2:16 ------------------------------------
2 |   def foo: 1 | 2 = 1 // error // error
  |                ^
  |                Singleton type Int(2) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:3:12 ------------------------------------
3 |   def bar: 3 | 4 = foo // error // error
  |            ^
  |            Singleton type Int(3) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:3:16 ------------------------------------
3 |   def bar: 3 | 4 = foo // error // error
  |                ^
  |                Singleton type Int(4) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:4:12 ------------------------------------
4 |   def foo: 1 | 2 = 1 // error // error // error
  |            ^
  |            Singleton type Int(1) is not allowed in a union type
-- Error: tests/neg/singletonOrs.scala:4:16 ------------------------------------
4 |   def foo: 1 | 2 = 1 // error // error // error
  |                ^
  |                Singleton type Int(2) is not allowed in a union type
-- [E119] Duplicate Symbol Error: tests/neg/singletonOrs.scala:4:7 -------------
4 |   def foo: 1 | 2 = 1 // error // error // error
  |       ^
  |method foo in object Test is already defined as def foo: => 
  |  <error Singleton type Int(1) is not allowed in a union type> | 
  |    <error Singleton type Int(2) is not allowed in a union type> at line 2.
  |The definitions have matching type signatures after erasure.
-- [E119] Duplicate Symbol Error: tests/neg/singletonOrs.scala:5:7 -------------
5 |   def bar: 1 = foo // error
  |       ^
  |method bar in object Test is already defined as def bar: => 
  |  <error Singleton type Int(3) is not allowed in a union type> | 
  |    <error Singleton type Int(4) is not allowed in a union type> at line 3.
  |Overloads with matching parameter types are not allowed.
```

I think the last two error messages are very confusing. The previous error is displayed when `previousDecl.showDcl` is called. That seemed like a nice way to show the conflicting declaration. Is there an alternative?

Blaisorblade: (Until you get more competent answers) If `showDcl` is the proper API (as I'm guessing), my ignorant vote is to stick to `showDcl` and maybe open an issue to improve its output.

On the error count, I suspect it's some side effect of how errors are combined/filtered—but not sure.

tests/neg/singletonOrs.scala
```diff
@@ -1,6 +1,6 @@
...
+   def bar: 1 = foo // error
```
allanrenucci: Actually I think a better solution is to rename the last two methods. This test is not meant to test duplicate definition

Blaisorblade:
Waiting for @allanrenucci before merging.

PR #4065, opened 2018-03-03
[WIP] Changes to Interpolation

odersky:
This is an attempt to do rethink interpolation of type variables from the ground up. So far it's an experiment.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -591,7 +591,13 @@ object Types {
         ctx.typerState.constraint.entry(tp) match {
           case bounds: TypeBounds if bounds ne next =>
             ctx.typerState.ephemeral = true
-            go(bounds.hi)
+            val upper = go(bounds.hi)
+            if (upper.exists) upper
+            else {
+              val lower = go(bounds.lo)
+              if (!lower.exists) upper
+              else ctx.typeComparer.addMemberBound(tp, bounds.lo, lower)
+            }
           case _ =>
             go(next)
         }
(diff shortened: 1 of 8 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -591,7 +591,13 @@ object Types {
         ctx.typerState.constraint.entry(tp) match {
           case bounds: TypeBounds if bounds ne next =>
             ctx.typerState.ephemeral = true
-            go(bounds.hi)
```
Blaisorblade: We discussed this with Martin on a whiteboard, so a memo (and possible comment).

If `X >: List[T]` and `x: X` and we typecheck `x.filter(p)`, how should `X` be instantiated? Here setting `X` to its lower bound works, but might violate other constraints — ideally we'd just add some constraint `X <: { def filter... }`, but since we don't have those, looking up `List.filter`'s definition class (say `Traversable[T]`) and adding bound `X <: Traversable[T]` is a more general solution.

This solution is not yet complete because some other ancestor of `List[T]` which doesn't extend `Traversable[T]` might have another overload of `filter`. There might even be a realistic example where this happens, but I can't think of it.

odersky: Yes, I think it makes sense to add this. Unfortunately, it is not enough. We still need to interpolate TypeVar results in some way when selecting because otherwise we might miss an implicit conversion. E.g. 

      Array(1, 2, 3).map(_ + 1).deep

The map operation with current stdlib leaves us with result type `That` where

    That >: Array[Int]

But to typecheck `deep` we need to find an implicit conversion from the result to `ArrayOps`. There is none from `That`, but there is one from `Array[Int]`.

odersky:
Superseded by #4080.

PR #4069, opened 2018-03-04
Add test to close #3666, problems with typechecking GADTs

### LINKED ISSUE #3666: Problem with type checking GADTs
Here's an interpreter for simply typed lambda calculus:

```scala
sealed trait Exp[T]
case class Num(n: Int) extends Exp[Int]
case class Plus(e1: Exp[Int], e2: Exp[Int]) extends Exp[Int]
case class Var[T](name: String) extends Exp[T]
case class Lambda[T, U](x: Var[T], e: Exp[U]) extends Exp[T => U]
case class App[T, U](f: Exp[T => U], e: Exp[T]) extends Exp[U]

abstract class Env { outer =>
  def apply[T](x: Var[T]): T

  def + [T](xe: (Var[T], T)) = new Env {
    def apply[T](x: Var[T]): T =
      if (x == xe._1) xe._2.asInstanceOf[T]
      else outer(x)
  }
}

object Env {
  val empty = new Env {
    def apply[T](x: Var[T]): T = ???
  }
}

object Test {

  val exp = App(Lambda(Var[Int]("x"), Plus(Var[Int]("x"), Num(1))), Var[Int]("2"))

  def eval[T](e: Exp[T])(env: Env): T = e match {
    case Num(n) => n
    case Plus(e1, e2) => eval(e1)(env) + eval(e2)(env)
    case v: Var[T] => env(v)
    case Lambda(x: Var[s], e) => ((y: s) => eval(e)(env + (x -> y)))
    case App(f, e) => eval(f)(env)(eval(e)(env))
  }

  eval(exp)(Env.empty)
}
```

When we compile that, we get two mysterious warnings:

    -- Warning: STL.scala:32:19 ----------------------------------------------------
    32 |        case Lambda(x: Var[s], e) => ((y: s) => eval(e)(env + (x -> y)))
       |             ^^^^^^^^^^^^^^^^^^^^
       |         There is no best instantiation of pattern type Lambda[Any, Any]
       |         that makes it a subtype of selector type Exp[T].
       |         Non-variant type variable U cannot be uniquely instantiated.
       |         (This would be an error under strict mode)
    -- Warning: STL.scala:33:16 ----------------------------------------------------
    33 |        case App(f, e) => eval(f)(env)(eval(e)(env))
       |             ^^^^^^^^^
       |           There is no best instantiation of pattern type App[Any, T]
       |           that makes it a subtype of selector type Exp[T].
       |           Non-variant type variable T' cannot be uniquely instantiated.
       |           
       |           where:    T  is a type in method eval
       |                     T' is a type variable
       |           
       |           (This would be an error under strict mode)
    two warnings found

If we look under the hood we find that these stem from the fact that we somehow miss the right subtype constraints for the `Lambda` and `App` cases. I suspect something fundamental is wrong here. I don't have the time to look into it now, but maybe someone who is up-to-speed with GADT matching can?

Blaisorblade:
Currently fails some checks in local testing and on CI, to investigate.

### DIFF
--- a/tests/pos/i3666.scala
+++ b/tests/pos/i3666.scala
@@ -38,3 +38,29 @@ object i3666 {
     eval(exp)(Env.empty)
   }
 }
+// A HOAS well-typed interpreter
+object i3666Hoas {
+  sealed trait Exp[T]
+  case class IntLit(n: Int) extends Exp[Int]
+  case class BooleanLit(b: Boolean) extends Exp[Boolean]
+
+  case class GenLit[T](t: T) extends Exp[T]
+  case class Plus(e1: Exp[Int], e2: Exp[Int]) extends Exp[Int]
+  case class Fun[S, T](f: Exp[S] => Exp[T]) extends Exp[S => T]
+  case class App[T, U](f: Exp[T => U], e: Exp[T]) extends Exp[U]
+
+
+  def eval[T](e: Exp[T]): T = e match {
+    case IntLit(n) => n
+    case BooleanLit(b) => b
+    case GenLit(t) => t
+    case Plus(e1, e2) => eval(e1) + eval(e2)
+    case f: Fun[s, t]  =>
+      (v: s) => eval(f.f(GenLit(v)))
+    case App(f, e) => eval(f)(eval(e))
+  }
+
+  val exp = App(Fun[S = Int](x => Plus(x, IntLit(1))), IntLit(2))
+
+  eval(exp)
+}


Blaisorblade:
Log has a bunch of failures on `tests/pos/i3666.scala`:

```
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for Any <:< _, frozen = true
assertion failure for Any <:< T, frozen = true
assertion failure for T <:< T, frozen = false
assertion failure for T <:< T, frozen = false
```

smarter:
The issue can be reduced to something like:
```scala
sealed trait Exp[T]                                                                                                                                                                                                                          
case class Var[T](name: String) extends Exp[T]                                                                                                                                                                                               
                                                                                                                                                                                                                                             
object Test {                                                                                                                                                                                                                                
  def env[T](x: Var[T]): T = ???                                                                                                                                                                                                             
                                                                                                                                                                                                                                             
  def eval[S](e: Exp[S]) = e match {                                                                                                                                                                                                         
    case v: Var[foo] =>                                                                                                                                                                                                                      
      env(v)                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                          
}
```
This crashes when run through `-Yno-deep-subtypes` because of a loop in subtype checking:
```scala
==> isSubType Any <:< foo
  ==> isSubType Any <:< S
   ==> isSubType Any <:< foo
```
The problem becomes apparent when we turn on the `gadts` printer:
```scala
narrow gadt bound of type foo:  from below to S TypeRef(NoPrefix,type S) false
narrow gadt bound of type S:  from below to foo TypeRef(NoPrefix,type foo) false
```
We end up with GADT constraints `foo >:> S` and `S >:> foo` but we don't have a mechanism to unify GADT constraints, so we end up with a loop in subtyping checks. We could add a special case to handle simple unifications like this but it seems like to be completely general we really need to reuse all of `ConstraintHandling` to deal with GADT constraints too.

biboudis:
Maybe this example with polymorphic recursion is also related:

```scala
object Test {
 
  class Foo[A]
  case class Bar[A, B](f: B => Foo[A]) extends Foo[A]

  def meth[A](consumer: A => Unit, s: Foo[A]): Unit = {
    s match {
      case bar: Bar[a, bt] => {
        meth[bt](((e: bt) => meth(consumer, new Foo[a])), new Foo[bt])
      }
    }
  }
}
```

Again detected with `-Yno-deep-subtypes`.

```shell
assertion failure for Any <:< a, frozen = true
assertion failure for Any <:< A, frozen = true
assertion failure for Any <:< a, frozen = true
...
```

PR #4070, opened 2018-03-04
Fix #4058: reject sealed and lazy for class parameters in parser

### LINKED ISSUE #4058: Lazy in constructor parameter causes Nil$.head
```scala
object App {
    class T(lazy val a: Int)
    new T(1)
  }
}

////////////////////////////////////////////////////////

Exception in thread "main" java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:428)
	at scala.collection.immutable.Nil$.head(List.scala:425)
	at dotty.tools.dotc.transform.SymUtils$.loop$1(SymUtils.scala:83)
	at dotty.tools.dotc.transform.SymUtils$.subst$extension(SymUtils.scala:85)
	at dotty.tools.dotc.transform.Constructors.$anonfun$transformTemplate$3(Constructors.scala:230)
	at scala.collection.immutable.List.flatMap(List.scala:335)
	at dotty.tools.dotc.transform.Constructors.transformTemplate(Constructors.scala:222)
	at dotty.tools.dotc.transform.MegaPhase.goTemplate(MegaPhase.scala:906)
	at dotty.tools.dotc.transform.MegaPhase.transformUnnamed$1(MegaPhase.scala:318)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:365)
	at dotty.tools.dotc.transform.MegaPhase.transformNamed$1(MegaPhase.scala:234)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:364)
	at dotty.tools.dotc.transform.MegaPhase.transformStat$1(MegaPhase.scala:373)
	at dotty.tools.dotc.transform.MegaPhase.$anonfun$transformStats$2(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.transformStats(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.transformUnnamed$1(MegaPhase.scala:264)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:365)
	at dotty.tools.dotc.transform.MegaPhase.mapDefDef$1(MegaPhase.scala:228)
	at dotty.tools.dotc.transform.MegaPhase.transformNamed$1(MegaPhase.scala:231)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:364)
	at dotty.tools.dotc.transform.MegaPhase.transformStat$1(MegaPhase.scala:373)
	at dotty.tools.dotc.transform.MegaPhase.$anonfun$transformStats$2(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.transformStats(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.transformUnnamed$1(MegaPhase.scala:317)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:365)
	at dotty.tools.dotc.transform.MegaPhase.transformNamed$1(MegaPhase.scala:234)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:364)
	at dotty.tools.dotc.transform.MegaPhase.transformStat$1(MegaPhase.scala:373)
	at dotty.tools.dotc.transform.MegaPhase.$anonfun$transformStats$2(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.transformStats(MegaPhase.scala:378)
	at dotty.tools.dotc.transform.MegaPhase.mapPackage$1(MegaPhase.scala:334)
	at dotty.tools.dotc.transform.MegaPhase.transformUnnamed$1(MegaPhase.scala:337)
	at dotty.tools.dotc.transform.MegaPhase.transformTree(MegaPhase.scala:365)
	at dotty.tools.dotc.transform.MegaPhase.transformUnit(MegaPhase.scala:384)
	at dotty.tools.dotc.transform.MegaPhase.run(MegaPhase.scala:396)
	at dotty.tools.dotc.core.Phases$Phase.$anonfun$runOn$1(Phases.scala:293)
	at scala.collection.immutable.List.map(List.scala:283)
	at dotty.tools.dotc.core.Phases$Phase.runOn(Phases.scala:291)
	at dotty.tools.dotc.core.Phases$Phase.runOn$(Phases.scala:290)
	at dotty.tools.dotc.transform.MegaPhase.runOn(MegaPhase.scala:138)
	at dotty.tools.dotc.Run.$anonfun$compileUnits$3(Run.scala:171)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at dotty.tools.dotc.util.Stats$.trackTime(Stats.scala:47)
	at dotty.tools.dotc.Run.$anonfun$compileUnits$2(Run.scala:168)
	at dotty.tools.dotc.Run.$anonfun$compileUnits$2$adapted(Run.scala:166)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:32)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:29)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:191)
	at dotty.tools.dotc.Run.runPhases$1(Run.scala:166)
	at dotty.tools.dotc.Run.$anonfun$compileUnits$1(Run.scala:191)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at dotty.tools.dotc.util.Stats$.maybeMonitored(Stats.scala:88)
	at dotty.tools.dotc.Run.compileUnits(Run.scala:149)
	at dotty.tools.dotc.Run.compileSources(Run.scala:136)
	at dotty.tools.dotc.Run.compile(Run.scala:120)
	at dotty.tools.dotc.Driver.doCompile(Driver.scala:29)
	at dotty.tools.dotc.Driver.process(Driver.scala:127)
	at dotty.tools.dotc.Driver.process(Driver.scala:96)
	at dotty.tools.dotc.Driver.process(Driver.scala:108)
	at dotty.tools.dotc.Driver.main(Driver.scala:135)
	at dotty.tools.dotc.Main.main(Main.scala)
```

Blaisorblade:
Fix #4058.
Thanks to @allanrenucci for guidance. Rejecting `lazy` there follows Scalac, while `sealed` *must* be rejected there because of our consistency checks.

Intentionally mixing in small fixes/tests nearby.

Wondering about the strategy; I guess using a general `modifiers` parser and then rejecting most modifiers does make sense to offer better error messages, compared to only accepting the actually sensible modifiers.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1650,6 +1650,7 @@ object Parsers {
       case PROTECTED => Mod.Protected()
       case SEALED    => Mod.Sealed()
     }
+    private def flagsOfModToken(tok: Int): FlagSet = modOfToken(tok).flags
 
     /** Drop `private' modifier when followed by a qualifier.
      *  Contract `abstract' and `override' to ABSOVERRIDE
@@ -1705,7 +1706,14 @@ object Parsers {
         }
       } else mods
 
-    /** {Annotation} {Modifier}
+    /**
+     * Extend the start Modifiers by parsing modifier tokens in the allowed BitSet, appearing in any order.
+     * Ignore modifier tokens in the parsedSeparately BitSet, so that the caller can parse them after modifier returns
+     * (suitable to parse `AllowedModifiers* lazy`).
+     * Report modifier tokens that appear neither in allowed nor in parsedSeparately.
+     *
+     * Grammar:
+     * {Annotation} {Modifier}
      *  Modifiers      ::= {Modifier}
      *  LocalModifiers ::= {LocalModifier}
      *  AccessModifier ::= (private | protected) [AccessQualifier]
@@ -1714,7 +1722,7 @@ object Parsers {
      *                  |  override
      *  LocalModifier  ::= abstract | final | sealed | implicit | lazy
      */
-    def modifiers(allowed: BitSet = modifierTokens, start: Modifiers = Modifiers()): Modifiers = {
+    def modifiers(allowed: BitSet = modifierTokens, start: Modifiers = Modifiers(), parsedSeparately: BitSet = BitSet.empty): Modifiers = {
       @tailrec
       def loop(mods: Modifiers): Modifiers = {
         if (allowed contains in.token) {
@@ -1725,6 +1733,8 @@ object Parsers {
           in.nextToken()
           loop(mods)
         } else {
+          if ((modifierTokens contains in.token) && !(parsedSeparately contains in.token))
+            syntaxError(hl"Modifier `${flagsOfModToken(in.token)}' not allowed at this position")
           mods
         }
       }
@@ -2471,7 +2481,7 @@ object Parsers {
         else if (isDefIntro(localModifierTokens))
           if (in.token == IMPLICIT || in.token == ERASED) {
             val start = in.offset
-            var imods = modifiers(funArgMods)
+            var imods = modifiers(funArgMods, parsedSeparately = BitSet(LAZY))
             if (isBindingIntro) stats += implicitClosure(start, Location.InBlock, imods)
             else stats +++= localDef(start, imods)
           } else {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1839,7 +1839,16 @@ object Parsers {
         val start = in.offset
         var mods = annotsAsMods()
         if (owner.isTypeName) {
-          mods = modifiers(start = mods) | ParamAccessor
+          mods = modifiers(start = mods)
+          if (mods.is(Lazy))
+            syntaxError("`lazy' modifier not allowed here. Use call-by-name parameters instead")
+          mods =
+            if (mods.is(Sealed)) {
+              syntaxError("`sealed' modifier can be used only for classes")
+              mods // Adding ParamAccessor would crash
+            } else {
+              mods | ParamAccessor
+            }
           mods =
             atPos(start, in.offset) {
               if (in.token == VAL) {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1839,16 +1839,10 @@ object Parsers {
         val start = in.offset
         var mods = annotsAsMods()
         if (owner.isTypeName) {
-          mods = modifiers(start = mods)
+          // Adding ParamAccessor would crash
+          mods = modifiers(start = mods, allowed = modifierTokens - SEALED) | ParamAccessor
           if (mods.is(Lazy))
             syntaxError("`lazy' modifier not allowed here. Use call-by-name parameters instead")
-          mods =
-            if (mods.is(Sealed)) {
-              syntaxError("`sealed' modifier can be used only for classes")
-              mods // Adding ParamAccessor would crash
-            } else {
-              mods | ParamAccessor
-            }
           mods =
             atPos(start, in.offset) {
               if (in.token == VAL) {
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -2454,7 +2454,7 @@ object Parsers {
 
     /** BlockStatSeq ::= { BlockStat semi } [ResultExpr]
      *  BlockStat    ::= Import
-     *                 | Annotations [implicit] [lazy] Def
+     *                 | Annotations [implicit] [erased] [lazy] Def
      *                 | Annotations LocalModifiers TmplDef
      *                 | Expr1
      *                 |
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1888,7 +1888,7 @@ object Parsers {
       def paramClause(): List[ValDef] = inParens {
         if (in.token == RPAREN) Nil
         else {
-          def funArgMods(): Unit = {
+          @tailrec def funArgMods(): Unit = {
             if (in.token == IMPLICIT) {
               implicitOffset = in.offset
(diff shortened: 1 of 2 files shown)


tests/neg/i4058.scala
```diff
@@ -4,6 +4,6 @@ class C(abstract val a: Int) // error
 class D {
   def f(sealed a: Int) = 0 // error
   def g(lazy a: Int) = 0 // error
-  def g(ov erride a: Int) = 0 //error
+  def g(override a: Int) = 0 // error
```
Blaisorblade: Hadn't found `testCompilation i4058` yet, my bad for not reading http://dotty.epfl.ch/docs/contributing/testing.html 😅

allanrenucci:
> Wondering about the strategy; I guess using a general modifiers parser and then rejecting most modifiers does make sense to offer better error messages, compared to only accepting the actually sensible modifiers

The modifier parser can take a set of allowed modifier as argument. But I believe in the case of `lazy` it is worth keeping the custom error message. I would not special case `sealed` and simply exclude it from the set of the allowed modifiers for parameters.

Blaisorblade:
> I would not special case sealed and simply exclude it from the set of the allowed modifiers for parameters.

Did that, then improved reporting: when `modifiers` runs into modifiers not `allowed` it says so, instead of "expected ... found 'sealed'". Code got trickier, so I added docs.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1839,7 +1849,10 @@ object Parsers {
         val start = in.offset
         var mods = annotsAsMods()
         if (owner.isTypeName) {
-          mods = modifiers(start = mods) | ParamAccessor
+          // Adding ParamAccessor would crash
+          mods = modifiers(start = mods, allowed = modifierTokens &~ BitSet(SEALED)) | ParamAccessor
```
allanrenucci: Why not ` modifierTokens - SEALED`?

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2451,7 +2464,7 @@ object Parsers {
 
     /** BlockStatSeq ::= { BlockStat semi } [ResultExpr]
      *  BlockStat    ::= Import
-     *                 | Annotations [implicit] [lazy] Def
+     *                 | Annotations [implicit] [ghost] [lazy] Def
```
allanrenucci: s/ghost/erased

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1714,7 +1722,7 @@ object Parsers {
      *                  |  override
      *  LocalModifier  ::= abstract | final | sealed | implicit | lazy
      */
-    def modifiers(allowed: BitSet = modifierTokens, start: Modifiers = Modifiers()): Modifiers = {
+    def modifiers(allowed: BitSet = modifierTokens, start: Modifiers = Modifiers(), parsedSeparately: BitSet = BitSet.empty): Modifiers = {
```
allanrenucci: I am not convinced by this change of API. If the caller wants to parse some modifier separately, then he can exclude them from the allowed modifiers.

Blaisorblade: To get the nicer error I need to distinguish "not allowed" (which gives an error) from "parsed separately" (which doesn't). I had to use `parsedSeparately` only once.

Blaisorblade: If the API is too hard for the error, I can revert/separate these refactorings.

Blaisorblade:
Rebased and addressed most comments.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2468,7 +2481,7 @@ object Parsers {
         else if (isDefIntro(localModifierTokens))
           if (in.token == IMPLICIT || in.token == ERASED) {
             val start = in.offset
-            var imods = modifiers(funArgMods)
+            var imods = modifiers(funArgMods, parsedSeparately = BitSet(LAZY))
```
allanrenucci: I don't think we need this, the current error message is good:
```scala
72 |  lazy def bar = 1
   |           ^
   |           modifier `lazy` is not allowed for this definition
```

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
allanrenucci: I think a better API would something along the line of:
```scala
/** ...
 *  @param errorOn emit a syntax error when a parsed modifier is in errorOn
 */
def modifiers(allowed: BitSet, start: Modifiers, errorOn: BitSet) = {
  ...
  if (allowed.contains(in.token) { ...; loop() }
  else if (errorOn.contains(in.token) { syntaxError; loop() }
  else ...
}
```
You can keep parsing modifiers after encountering an invalid modifier.

But since it is used only once, I think we should no complicate this API.

odersky:
Flag checking is generally done by Namer using `checkWellformed`. I think that's where any new tests belong as well.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
Blaisorblade: That’s not the error message I’m fixing. Will separate the PR and add examples.

Blaisorblade:
SEALED must be rejected here, otherwise adding ParamAccessor will crash right away.
If you still want to move other checks later, I can try.

odersky:
I believe we need a more systematic fix here. The root of the problem is that there are term flags and type flags and we are adding a term flag to a type flag or vice versa, which is not supported. 

We should use in Parser a general flag combining method which detects and reports this problem. But don't change the parsed syntax! The syntax is given by `syntax.md` and should be parsed as such.

odersky:
It turned out that the parser did already what I suggested. There was just one instance where the correct combiner method was not called. #4126 fixes this.

PR #4071, opened 2018-03-04
Fix #4060 by marking ghost symbols as unstable

### LINKED ISSUE #4060: Unsoundness due to `ghost`
Related to #50 #4042 #4031.

```scala
object App {
  def coerce[U, V](u: U): V = {
    trait X { type R >: U }
    trait Y { type R = V }

    class T[A <: X](ghost val a: A)(val value: a.R)

    object O { lazy val x : Y & X = ??? }

    val a = new T[Y & X](O.x)(u)
    a.value
  }

    val x: Int = coerce[String, Int]("a")
    println(x + 1)
  }
}

//////////////////////////////////////////////////////////////

	at App$.main(HelloWorld.scala:15)
	at App.main(HelloWorld.scala)
```

Blaisorblade:
Closes #4060.


Blaisorblade:
test performance please

tests/pos/ghost-pathdep-1.scala
```diff
@@ -1,19 +0,0 @@
-object Test {
-
-  fun1(new Bar)
-  fun2(new Bar)
-  fun3(new Bar)
-
-  def fun1[F >: Bar <: Foo](ghost f: F): f.X = null.asInstanceOf[f.X]
-  def fun2[F >: Bar <: Foo](ghost f: F)(ghost bar: f.B): f.B = null.asInstanceOf[f.B]
```
nicolasstucki: Make them neg tests. It is always good to make sure we do not have a regression.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -600,7 +600,7 @@ object SymDenotations {
 
     /** Is this a denotation of a stable term (or an arbitrary type)? */
     final def isStable(implicit ctx: Context) =
-      isType || is(Stable) || !(is(UnstableValue) || info.isInstanceOf[ExprType])
+      isType || !is(Ghost) && (is(Stable) || !(is(UnstableValue) || info.isInstanceOf[ExprType]))
```
nicolasstucki: We should avoid marking the symbols as stable if they are ghost. We could assert that a symbol must at most one of those flags.

nicolasstucki: Because in some places we might ask if `sym.is(Stable)` which would say true for a symbol with Stable and Ghost flags.

nicolasstucki: Actually don't do this change

Blaisorblade:
@nicolasstucki Good for you?

PR #4072, opened 2018-03-05
Fix precedence in pretty-printing

### LINKED ISSUE #4068: Pretty-printing precedence issue
Type `(Int => Any) => List[Any]` is shown without the parens:
```scala
scala> :type List(1, 2).map[T = Int]
Int => Any => List[Any]
```
while parens are omitted correctly here:
```scala
scala> val x: Int => (Int => Int) = _ => _ => 1
val x: Int => Int => Int = Lambda$1619/864078397@23310248
```

that is, the pretty-printer misses associativity of `=>`.

Blaisorblade:
WIP. I consider this a hack, though I haven't found the adequate test to show it.
Fix #4068.

### DIFF
--- a/compiler/test-resources/type-printer/functions
+++ b/compiler/test-resources/type-printer/functions
@@ -1,2 +1,10 @@
 scala> val toInt: Any => Int = new { def apply(a: Any) = 1; override def toString() = "<func1>" }
 val toInt: Any => Int = <func1>
+scala> val hoFun: (Int => Int) => Int = new { def apply(a: Int => Int) = 1; override def toString() = "<func2>" }
+val hoFun: (Int => Int) => Int = <func2>
+scala> val curriedFun: Int => (Int => Int) = new { def apply(a: Int) = _ => 1; override def toString() = "<func3>" }
+val curriedFun: Int => Int => Int = <func3>
+scala> val tupFun: ((Int, Int)) => Int = new { def apply(a: (Int, Int)) = 1; override def toString() = "<func4>" }
+val tupFun: ((Int, Int)) => Int = <func4>
+scala> val binFun: (Int, Int) => Int = new { def apply(a: Int, b: Int) = 1; override def toString() = "<func5>" }
+val binFun: (Int, Int) => Int = <func5>
--- a/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
@@ -135,7 +135,9 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
     def toTextFunction(args: List[Type], isImplicit: Boolean, isGhost: Boolean): Text =
       changePrec(GlobalPrec) {
         val argStr: Text =
-          if (args.length == 2 && !defn.isTupleType(args.head))
+          if (args.length == 2 && defn.isFunctionType(args.head)) {
+            "(" ~ argText(args.head) ~ ")"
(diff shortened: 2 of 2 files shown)


OlivierBlanvillain:
Looks reasonable, what would you like to do better?

Blaisorblade:
I wanted to use `atPrec(...)` but failed. Feel free to merge if you think it improves things, I can still fix it better later.

Blaisorblade:
Turns out more invasive changes are needed, since apparently both parsing and pretty-printing of type operators are buggy, and in different ways.

The parser gives precedence to type operators depending on the character used. That seems an oversight, and directly contradicts both Scalac behavior and the [spec](https://www.scala-lang.org/files/archive/spec/2.13/03-types.html#infix-types), which assigns the same precedence to *all* custom infix type operators. Dotty's grammar treats `&` and `|` according to the same rule, tho it's clear somebody writing the code assumed they should have special and higher precedences. Honestly, it'd be cool to make `&` have the same precedence as `with`.

Meanwhile, the pretty-printer instead cites and (tries to) implements the spec, tho it still has multiple bugs.

So I'm changing both, and checking that parser and pretty-printer roundtrip on manual tests.

compiler/src/dotty/tools/dotc/parsing/package.scala
```diff
@@ -7,7 +7,7 @@ import core.NameOps._
 
 package object parsing {
 
-  def precedence(operator: Name): Int =
+  def precedence(operator: Name, isType: Boolean = false): Int =
```
allanrenucci: `isType` unused?

compiler/test-resources/type-printer/functions
```diff
@@ -1,2 +1,10 @@
 scala> val toInt: Any => Int = new { def apply(a: Any) = 1; override def toString() = "<func1>" }
 val toInt: Any => Int = <func1>
+scala> val hoFun: (Int => Int) => Int = new { def apply(a: Int => Int) = 1; override def toString() = "<func2>" }
+val hoFun: (Int => Int) => Int = <func2>
+scala> val curriedFun: Int => (Int => Int) = new { def apply(a: Int) = _ => 1; override def toString() = "<func3>" }
+val curriedFun: Int => Int => Int = <func3>
+scala> val tupFun: ((Int, Int)) => Int = new { def apply(a: (Int, Int)) = 1; override def toString() = "<func4>" }
+val tupFun: ((Int, Int)) => Int = <func4>
+scala> val binFun: (Int, Int) => Int = new { def apply(a: Int, b: Int) = 1; override def toString() = "<func5>" }
+val binFun: (Int, Int) => Int = <func5>
```
allanrenucci: Can we rewrite these tests as:
```diff
- scala> val toInt: Any => Int = new { def apply(a: Any) = 1; override def toString() = "<func1>" }                                                                                                               
- val toInt: Any => Int = <func1>
+ scala> def toInt: Any => Int = ???
+ def toInt: Any => Int
```
There is too much noise with vals

compiler/test-resources/type-printer/functions
```diff
@@ -1,2 +1,10 @@
-scala> val toInt: Any => Int = new { def apply(a: Any) = 1; override def toString() = "<func1>" }
-val toInt: Any => Int = <func1>
+scala> def toInt: Any => Int = new { def apply(a: Any) = 1; override def toString() = "<func1>" }
```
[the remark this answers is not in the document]
Blaisorblade: aaah sorry i wasn't thinking fixed 😅

PR #4077, opened 2018-03-06
Remove PreName add (almost) all other implicit conversions

OlivierBlanvillain:
This is followup to #3986 than removes a few virtual calls that the JIT complained about.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ compiler/after-pickling.txt
 *.dotty-ide-version
 
 *.decompiled.out
+bench/compile.txt
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -808,23 +808,25 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
       else Nil
 
     def annotations: List[Annotation] = toDenot(sym).annotations
-    def companionModuleMembers: List[Symbol] =  {
+
+    def companionModuleMembers: List[Symbol] = {
       // phase travel to exitingPickler: this makes sure that memberClassesOf only sees member classes,
       // not local classes of the companion module (E in the exmaple) that were lifted by lambdalift.
       if (linkedClass.isTopLevelModuleClass) /*exitingPickler*/ linkedClass.memberClasses
       else Nil
     }
+
     def fieldSymbols: List[Symbol] = {
       toDenot(sym).info.decls.filter(p => p.isTerm && !p.is(Flags.Method))
     }
+
     def methodSymbols: List[Symbol] =
       for (f <- toDenot(sym).info.decls.toList if f.isMethod && f.isTerm && !f.isModule) yield f
+
     def serialVUID: Option[Long] = None
 
-
-    def freshLocal(cunit: CompilationUnit, name: String, tpe: Type, pos: Position, flags: Flags): Symbol = {
-      ctx.newSymbol(sym, name.toTermName, FlagSet(flags), tpe, NoSymbol, pos)
-    }
+    def freshLocal(cunit: CompilationUnit, name: String, tpe: Type, pos: Position, flags: Flags): Symbol =
+      ctx.newSymbol(sym, name.toTermName, FlagSet(flags), tpe, NoSymbol, pos.toCoord)
 
     def getter(clz: Symbol): Symbol = decorateSymbol(sym).getter
     def setter(clz: Symbol): Symbol = decorateSymbol(sym).setter
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -498,7 +498,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     // unrelated change.
        ctx.base.settings.YnoGenericSig.value
     || sym.is(Flags.Artifact)
-    || sym.is(Flags.allOf(Flags.Method, Flags.Lifted))
+    || sym.isBoth(Flags.Method, and = Flags.Lifted)
     || sym.is(Flags.Bridge)
   )
 
@@ -679,7 +679,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def isConstructor: Boolean = toDenot(sym).isConstructor
     def isAnonymousFunction: Boolean = toDenot(sym).isAnonymousFunction
     def isMethod: Boolean = sym is Flags.Method
-    def isPublic: Boolean =  sym.flags.is(Flags.EmptyFlags, Flags.Private | Flags.Protected)
+    def isPublic: Boolean =  sym.flags.is(Flags.EmptyFlags, butNot=Flags.Private | Flags.Protected)
     def isSynthetic: Boolean = sym is Flags.Synthetic
     def isPackageClass: Boolean = sym is Flags.PackageClass
     def isModuleClass: Boolean = sym is Flags.ModuleClass
@@ -716,7 +716,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def isDeprecated: Boolean = false
     def isMutable: Boolean = sym is Flags.Mutable
     def hasAbstractFlag: Boolean =
-      (sym is Flags.Abstract) || (sym is Flags.JavaInterface) || (sym is Flags.Trait)
+      (sym is Flags.Abstract) || (sym is Flags.Trait)
     def hasModuleFlag: Boolean = sym is Flags.Module
     def isSynchronized: Boolean = sym is Flags.Synchronized
     def isNonBottomSubClass(other: Symbol): Boolean = sym.derivesFrom(other)
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -828,8 +828,8 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def freshLocal(cunit: CompilationUnit, name: String, tpe: Type, pos: Position, flags: Flags): Symbol =
       ctx.newSymbol(sym, name.toTermName, FlagSet(flags), tpe, NoSymbol, pos.toCoord)
 
-    def getter(clz: Symbol): Symbol = decorateSymbol(sym).getter
-    def setter(clz: Symbol): Symbol = decorateSymbol(sym).setter
+    def getter(clz: Symbol): Symbol = new SymUtilsOps(sym).getter
+    def setter(clz: Symbol): Symbol = new SymUtilsOps(sym).setter
 
     def moduleSuffix: String = "" // todo: validate that names already have $ suffix
     def outputDirectory: AbstractFile = DottyBackendInterface.this.outputDirectory
@@ -842,7 +842,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
      * Redundant interfaces are removed unless there is a super call to them.
      */
     def superInterfaces: List[Symbol] = {
-      val directlyInheritedTraits = decorateSymbol(sym).directlyInheritedTraits
+      val directlyInheritedTraits = new SymUtilsOps(sym).directlyInheritedTraits
       val directlyInheritedTraitsSet = directlyInheritedTraits.toSet
       val allBaseClasses = directlyInheritedTraits.iterator.flatMap(_.symbol.asClass.baseClasses.drop(1)).toSet
       val superCalls = superCallsMap.getOrElse(sym, Set.empty)
--- a/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
+++ b/compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
@@ -154,9 +154,9 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
   lazy val Predef_classOf: Symbol = defn.ScalaPredefModule.requiredMethod(nme.classOf)
 
   lazy val AnnotationRetentionAttr = ctx.requiredClass("java.lang.annotation.Retention")
-  lazy val AnnotationRetentionSourceAttr = ctx.requiredClass("java.lang.annotation.RetentionPolicy").linkedClass.requiredValue("SOURCE")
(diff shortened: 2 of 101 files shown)


OlivierBlanvillain:
test performance please

odersky:
It does not seem to get us any performance gain. And there is quite a bit of duplication, in particular by duplicating InheritedCache. So I am dubious this is actually an improvement. 

regarding PreName:  yes we need relatively few overloads if we drop it, but that is not the end of the story. The problem is that in the future there will be constant hassles that a method exists for Name but not for String or vice versa, which will require either annoying conversions or more ad-hoc overloads. So I would actually be in favor of keeping PreName in as well.

OlivierBlanvillain:
@odersky I remove the InheritedCache commit, I agree it was too much duplication.

Regarding the PreName commit I updated it to minimize of number of `String` method (instead of minimizing the diff size, which I did previously). The result has only 4 overloaded methods that have a overloaded String version, which corresponds to the once that are heavily used in Definitions.scala. For everything else, Strings should be explicitly converted at call site with `.toTermName`/`.toTypeName`.

I believe the PreName change is worth doing for readability alone. The change was actually motivated by a real life confusion: @gsps came to my office a few weeks asking "how do I convert a String to PreName?", we spent ~10 minutes pocking around the code trying to figure things out, and finally had to resort to compiling Definitions.scala with `-Xprint:typer` to understand what was actually going on there.

odersky:
Looking at the diffs I am still not convinced about `PreName`. I think the code reads worse now than before. Regarding discoverability, let's just put the `String` instance for `PreName` next to `PreName`, then it is never an issue.

OlivierBlanvillain:
The few additional `.toTypeName` and `.toTermName` are the price to pay to get ride of implicit conversions. If these changes are too much I guess we should abandon the idea of further deprecating them, as it would imply adding `import scala.language.implicitConversions` on the ~100 files touched by this PR.

I newly pushed commits remove all the other implicit conversions, with the exception of [these two](https://github.com/dotty-staging/dotty/blob/bdfe7406d0f520f116947123c48b6e5d9753b26e/compiler/src/dotty/tools/dotc/core/Symbols.scala#L720-L724) to use `Symbol` as if it was a `SymDenotation`.

.gitignore
```diff
@@ -73,3 +73,4 @@ compiler/after-pickling.txt
 *.dotty-ide-version
 
 *.decompiled.out
+bench/compile.txt
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -134,7 +134,7 @@ object Types {
     }
 
     /** Is this type different from NoType? */
-    def exists: Boolean = true
+    final def exists: Boolean = !this.eq(NoType)
```
nicolasstucki: Use `this.ne(NoType)`, otherwise this commit LGTM

compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -91,7 +91,7 @@ object Trees {
     /** The type of the tree. In case of an untyped tree,
      *   an UnAssignedTypeException is thrown. (Overridden by empty trees)
      */
-    def tpe: T @uncheckedVariance = {
+    final def tpe: T @uncheckedVariance = {
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/dotc/transform/TypeUtils.scala
```diff
@@ -12,23 +12,20 @@ import NameOps._
...
 
-  def ensureMethodic(implicit ctx: Context): Type = self match {
-    case self: MethodicType => self
-    case _ => if (ctx.erasedTypes) MethodType(Nil, self) else ExprType(self)
+    def ensureMethodic(implicit ctx: Context): Type = self match {
+      case self: MethodicType => self
+      case _ => if (ctx.erasedTypes) MethodType(Nil, self) else ExprType(self)
+    }
   }
 }
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/dotc/util/Positions.scala
```diff
@@ -166,8 +170,9 @@ object Positions {
   }
 
   /** An index coordinate */
-  implicit def indexCoord(n: Int): Coord = new Coord(n + 1)
-  implicit def positionCoord(pos: Position): Coord =
+  def indexCoord(n: Int): Coord = new Coord(n + 1)
+
+  def pos2coord(pos: Position): Coord =
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/repl/terminal/filters/UndoFilter.scala
```diff
@@ -147,8 +147,8 @@ object UndoState {
...
 
 object UndoFilter {
-  val undoMsg       = Ansi.Color.Blue(" ...undoing last action, `Alt -` or `Esc -` to redo")
-  val cannotUndoMsg = Ansi.Color.Blue(" ...no more actions to undo")
-  val redoMsg       = Ansi.Color.Blue(" ...redoing last action")
-  val cannotRedoMsg = Ansi.Color.Blue(" ...no more actions to redo")
+  val undoMsg       = Ansi.Color.Blue(Ansi.Str.parse(" ...undoing last action, `Alt -` or `Esc -` to redo"))
+  val cannotUndoMsg = Ansi.Color.Blue(Ansi.Str.parse(" ...no more actions to undo"))
+  val redoMsg       = Ansi.Color.Blue(Ansi.Str.parse(" ...redoing last action"))
+  val cannotRedoMsg = Ansi.Color.Blue(Ansi.Str.parse(" ...no more actions to redo"))
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -716,7 +716,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def isDeprecated: Boolean = false
     def isMutable: Boolean = sym is Flags.Mutable
     def hasAbstractFlag: Boolean =
-      (sym is Flags.Abstract) || (sym is Flags.JavaInterface) || (sym is Flags.Trait)
+      (sym is Flags.Abstract) || (sym is Flags.Trait)
```
nicolasstucki: Why?

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -48,22 +50,18 @@ object Flags {
...
-    /** Does this flag set have all of the flags in given flag conjunction?
-     *  Pre: The intersection of the typeflags of both sets must be non-empty.
-     */
-    def is(flags: FlagConjunction): Boolean = {
-      val fs = bits & flags.bits
-      (fs & KINDFLAGS) != 0 &&
-      (fs >>> TYPESHIFT) == (flags.bits >>> TYPESHIFT)
-    }
+    /** Does this flag set have both of the given flags? */
+    def isBoth(flags: FlagSet, and: FlagSet): Boolean =
```
nicolasstucki: This commit LGTM

compiler/src/dotty/tools/dotc/reporting/Reporter.scala
```diff
@@ -30,32 +32,38 @@ import Reporter._
...
   def reportWarning(warning: Warning): Unit =
     if (!this.settings.silentWarnings.value) {
       if (this.settings.XfatalWarnings.value) reporter.report(warning.toError)
       else  reporter.report(warning)
     }
 
-  def deprecationWarning(msg: => Message, pos: SourcePosition = NoSourcePosition): Unit =
-    reportWarning(new DeprecationWarning(msg, pos))
+  def deprecationWarning(msg: => Message, pos: SourcePosition): Unit = reportWarning(new DeprecationWarning(msg, pos))
+  def deprecationWarning(msg: => Message, pos: Position = NoPosition): Unit = deprecationWarning(msg, sourcePos(pos))
```
odersky: Why duplicate `deprecationWarning` (and the others like it) but not `inform` and `echo`?'

Overall I am skeptical about this commit. `SourcePosition` is the better type, whereas `Position` is a compromise to gain speed. So it looks better to define the API of reporters in terms of `SourcePosition`. Their usage becomes more flexible that way.  It seems we can get this only at the price of duplicating all methods. Not terrible but not very nice either. The implicit conversion avoids all that boilerplate.

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -498,7 +498,7 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     // unrelated change.
```
odersky: I am against merging this as is. I believe the original version is both clearer and more efficient. Computing `|` on the fly as a cost, that's why we created all these flag set. Doing several flag tests in a row also has a cost. On the other hand, testing against a FlagConjunction with `is` is fast and does not require boxing.

compiler/src/dotty/tools/repl/terminal/filters/UndoFilter.scala
odersky: Not sure about this one.

compiler/src/dotty/tools/dotc/ast/Positioned.scala
```diff
@@ -16,10 +16,12 @@ abstract class Positioned extends DotClass with Product {
 
   setPos(initialPos)
 
-  /** The item's position.
-   */
+  /** The item's position. */
   def pos: Position = curPos
 
+  /** The item's coord. */
+  def coord: Coord = curPos.toCoord
```
odersky: I think this is wrong. "Coord" is an optimization for Symbols, so that they can take alternatively a Position or an Index. It's wrong to pollute other data types with this distinction - they should have no idea what a `Coord` is nor how to covert to it.

compiler/src/dotty/tools/dotc/util/Positions.scala
odersky: Overall I am skeptical about merging this.

compiler/src/dotty/tools/dotc/core/OrderingConstraint.scala
```diff
@@ -601,7 +601,7 @@ class OrderingConstraint(private val boundsMap: ParamBounds,
         " := " ~ tp.toText(printer)
     }
```
odersky: Not an improvement, lots of nitty gritty boilerplate to add.

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -828,8 +828,8 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
     def freshLocal(cunit: CompilationUnit, name: String, tpe: Type, pos: Position, flags: Flags): Symbol =
```
odersky: `SymUtils` is not used much, to this commit looks smallish. But we should use it more often because it gives us a way to add methods without overloading SymDenotation. In that case I think it is essential that it works for `Symbol` and `SymDenotation` in the same way. So my vote is to leave this as is.

compiler/src/dotty/tools/dotc/core/TypeApplications.scala
```diff
@@ -154,10 +154,11 @@ object TypeApplications {
   }
 }
 
-import TypeApplications._
-
 /** A decorator that provides methods for modeling type application */
-class TypeApplications(val self: Type) extends AnyVal {
+trait TypeApplications extends Any {
+  def self: Type
```
odersky: That will cause a box for every method call. So don't think this is an improvement.

compiler/src/dotty/tools/dotc/core/StdNames.scala
```diff
@@ -47,17 +47,17 @@ object StdNames {
   }
 
   abstract class DefinedNames[N <: Name] {
```
odersky: I prefer the current version.

compiler/test/dotty/tools/repl/ReplTest.scala
```diff
@@ -44,12 +44,14 @@ class ReplTest extends ReplDriver(
   def storedOutput(): String =
     stripColor(out.asInstanceOf[StoringPrintStream].flushStored())
 
-  protected implicit def toParsed(expr: String)(implicit state: State): Parsed = {
```
allanrenucci: I think a better solution is to expose a `compile` method in `ReplTest` that takes a `String`

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
OlivierBlanvillain: Because is(JavaInterface) implies is(Trait): https://github.com/lampepfl/dotty/blob/84bf2fac1aac3f381f2f3f68cc8a7dde538aca95/compiler/src/dotty/tools/dotc/core/Flags.scala#L614

compiler/src/dotty/tools/backend/jvm/DottyBackendInterface.scala
```diff
@@ -154,9 +154,9 @@ class DottyBackendInterface(outputDirectory: AbstractFile, val superCallsMap: Ma
   lazy val Predef_classOf: Symbol = defn.ScalaPredefModule.requiredMethod(nme.classOf)
```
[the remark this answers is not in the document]
OlivierBlanvillain: That doesn't work as moving it the companion of PreName is not enough to make `.toTermName` and `.toTypeName` available on Strings. I guess we could have two implicit conversions, one in PreName's companion and one available by import, but I'm feel it would just make things worse...

Are you sure you don't like this commit as is? In my opinion completely removing an abstraction from the code base is a clear simplification. The diff might look large but in reality most of the line changes are about making type *more precise* by having the `TermName`/`TypeName` instead of `PreName`. The balance sheet in term of added boilerplate is also reasonable:

`toTermName` added: 22
`toTermName` removed: 12

`toTypeName` added: 12
`toTypeName` removed: 4

+4 overloaded methods with String & Name arguments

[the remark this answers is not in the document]
odersky: I still think we should leave PreName. I don't like to have to overload methods, and the problem is, if you are systematic about it (and you should be!) you will end up with a lot more overloaded methods. `PreName` is a single abstraction which fixes this quite nicely. For me it is really analogous to `IterableOnce` in collections. Before we had some overloaded methods that took `Iterable`s and `Iterator`s respectively. At first there were only a few. Then people complained that, by rights, some other methods should also be dual purpose, and then some others more, and so on. The introduction of `IterableOnce` stopped the bleeding. 

I realize it's a contentious issue, but then the maxime "when in doubt, don't change it" applies.

One other thing: We have to tell everyone working on the compiler that `Decorators` is a highly recommended (maybe mandatory?) input. I have seen so much awkward code because people did not know about it. Is there a good place where we can put it so people will read it?

PR #4079, opened 2018-03-06
Fix docs(swap type arguments)

tokkiyaa:


tokkiyaa:
Oh...
I have just signed CLA.

PR #4080, opened 2018-03-06
Rework type inference

odersky:
This is an attempt to put type inference on a more solid basis. The main two commits are

### Drop Ephemeral 90a661b

Changing the interpolation scheme uncovered several cache invalidation problems with

 - the asSeenFrom cache in Denotations
 - the superType cache in AppliedType
 - the lastDenotation cache in NamedType

The new interpolation scheme performed essentially the same operations as the old one, but sometimes in a different order. I am still not quite sure how the differences made the cache invalidations fail. On the other hand, it is quite plausible (obvious even, in retrospect) that the previous invalidation schemes are incomplete. So this commit replaces them with a common algorithm that does not rely on the previous global state represented by `ephemeral`.

### New interpolation scheme 417f0c8

Variable interpolation "improves" types by instantiating type variables that do not appear in the result of a type derivation, or that appear only co- or only contra-variantly. This is convenient because it keeps the constraints small. It is also necessary since some operations have more or better solutions once type variables are instantiated. For instance, the members of a type
variable

    X >: T <: U

are the members of `U`. But assuming `X` is covariant once it is instantiated to `T` we get more members. Similarly for implicit searches. 

The major changes from the previous interpolation scheme are the following:

 - We explicitly keep track in typer of which variables should and
   which should not be interpolated. This replaces searching trees
   for embedded variable definitions, which is fragile e.g. in the
   presence of eta expansion.
 - We compute variances starting with all variables found in the type,
   not just the qualifying ones. The previous scheme caused some
   variance information to be missed, which caused some variables
   to be mis-classified as non-occurring. i4032.scala is a test case.
   Unfortunately, fixing this caused several other tricky inference
   failures because which were previously hidden because some variables
   were already instantiated prematurely. Examples were hmap.scala,
   hmap-covariant.scala, and i2300.scala. In all these cases there was another
   problem which was masked by the fact that some type variables had
   already been instantiated where they should not have been.
 - We interpolate at the end of `typedUnadapted` instead of at the
   beginning of `adapt`. Tracking instantiatable variables turned out
   to be easier this way.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TyperState.scala
+++ b/compiler/src/dotty/tools/dotc/core/TyperState.scala
@@ -76,7 +76,7 @@ class TyperState(previous: TyperState /* | Null */) {
   /** The uninstantiated variables */
   def uninstVars = constraint.uninstVars
 
-  /** The set of uninstantiated type varibles which have this state as their owning state */
+  /** The set of uninstantiated type variables which have this state as their owning state */
   private[this] var myOwnedVars: TypeVars = SimpleIdentitySet.empty
   def ownedVars = myOwnedVars
   def ownedVars_=(vs: TypeVars): Unit = myOwnedVars = vs
--- a/compiler/src/dotty/tools/dotc/core/TyperState.scala
+++ b/compiler/src/dotty/tools/dotc/core/TyperState.scala
@@ -42,19 +42,6 @@ class TyperState(previous: TyperState /* | Null */) extends DotClass with Showab
   private val previousConstraint =
     if (previous == null) constraint else previous.constraint
 
-  private[this] var myEphemeral: Boolean =
-    if (previous == null) false else previous.ephemeral
-
-  /** The ephemeral flag is set as a side effect if an operation accesses
-   *  the underlying type of a type variable. The reason we need this flag is
-   *  that any such operation is not referentially transparent; it might logically change
-   *  its value at the moment the type variable is instantiated. Caching code needs to
-   *  check the ephemeral flag; If the flag is set during an operation, the result
-   *  of that operation should not be cached.
-   */
-  def ephemeral = myEphemeral
-  def ephemeral_=(x: Boolean): Unit = { myEphemeral = x }
-
   private[this] var myIsCommittable = true
 
   def isCommittable = myIsCommittable
@@ -159,7 +146,6 @@ class TyperState(previous: TyperState /* | Null */) extends DotClass with Showab
     constraint foreachTypeVar { tvar =>
       if (tvar.owningState.get eq this) tvar.owningState = new WeakReference(targetState)
     }
-    targetState.ephemeral |= ephemeral
     targetState.gc()
     reporter.flush()
     isCommitted = true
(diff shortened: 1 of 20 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -96,6 +96,41 @@ object Types {
...
+     */
+    def isProvisional(implicit ctx: Context) = mightBeProvisional && testProvisional
+
+    private def testProvisional(implicit ctx: Context) = {
+      val accu = new TypeAccumulator[Boolean] {
+        override def apply(x: Boolean, t: Type) =
+          x || t.mightBeProvisional && {
+            t.mightBeProvisional = t match {
+              case t: TypeVar =>
+                !t.inst.exists
```
smarter: An instantiated TypeVar could have its underlying type refer to another uninstantiated TypeVar and therefore still be provisional I think.

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
Rebased

odersky:
test performance please

compiler/src/dotty/tools/dotc/core/TyperState.scala
```diff
@@ -76,6 +76,11 @@ class TyperState(previous: TyperState /* | Null */) {
   /** The uninstantiated variables */
   def uninstVars = constraint.uninstVars
 
+  /** The set of uninstantiated type varibles which have this state as their owning state */
```
smarter: typo: varibles -> variables

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -675,14 +675,16 @@ object Erasure {
       super.typedStats(stats1, exprOwner).filter(!_.isEmpty)
     }
 
-    override def adapt(tree: Tree, pt: Type)(implicit ctx: Context): Tree =
+    override def adapt(tree: Tree, pt: Type, locked: TypeVars)(implicit ctx: Context): Tree =
       trace(i"adapting ${tree.showSummary}: ${tree.tpe} to $pt", show = true) {
-        assert(ctx.phase == ctx.erasurePhase.next, ctx.phase)
+        assert(ctx.phase == ctx.erasurePhase || ctx.phase == ctx.erasurePhase.next, ctx.phase)
```
smarter: Why did this change?

compiler/src/dotty/tools/dotc/typer/Inferencing.scala
```diff
@@ -313,71 +318,47 @@ object Inferencing {
...
+   *  non-variantly in the type are left untouched.
+   *
+   *  Note that even type variables that do not appear directly in a type, can occur with
+   *  some variance in the type, because of the constraints. E.g if `X` occurs co-variantly in `T`
+   *  and we have a constraint
+   *
+   *      Y <: X
+   *
+   *  Then `Y` also occurs co-variantly in `T` because it needs to be minimized in order to constrain
+   *  `T` teh least. See `variances` for more detail.
```
smarter: typo: teh -> the

compiler/src/dotty/tools/dotc/transform/Erasure.scala
odersky: Not sure why. Some transform ran at phase erasure and used the erasure typer to do it. Seems legit. Not sure why it did not run before.

PR #4081, opened 2018-03-07
Fix #4044: Change quote pickled representation

### LINKED ISSUE #4044: Nested splice/quote fail trying to find value 'apply'
The following code which should work
```
import scala.quoted._
class Foo {
  val e: Expr[Int] = '(3)
  '{ ~( '{ ~e } ) }
}
```

fails with 
```
value `apply` is not a member of scala.quoted.Expr[Int]
```

nicolasstucki:


nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -258,46 +279,116 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
-    /** If inside a quote, split `body` into a core and a list of embedded quotes
+    /** If inside a quote, split the body of the splice into a core and a list of embedded quotes
      *  and make a hole from these parts. Otherwise issue an error, unless we
      *  are in the body of an inline method.
      */
-    private def splice(body: Tree, splice: Tree)(implicit ctx: Context): Tree = {
-      if (inQuote) {
-        val (body1, quotes) = nested(isQuote = false).split(body)
-        makeHole(body1, quotes, splice.tpe)
+    private def splice(splice: Select)(implicit ctx: Context): Tree = {
```
biboudis: `splice` could handle the case of `TypeTree` before.

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -162,7 +184,6 @@ class ReifyQuotes extends MacroTransformWithImplicits {
     def tryHeal(tp: Type, pos: Position)(implicit ctx: Context): Option[String] = tp match {
       case tp: TypeRef =>
         if (level == 0) {
-          assert(ctx.owner.is(Macro))
```
odersky: Why was the assert dropped?

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -258,46 +279,116 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
         splice
       }
-    }.withPos(splice.pos)
+      else {
+        val (body1, quotes) = nested(isQuote = false).split(splice.qualifier)
+        makeHole(body1, quotes, splice.tpe).withPos(splice.pos)
+      }
+    }
+
+    /** Transforms the contents of a splice nested
```
odersky: "splice nested" -> "nested splice"?

odersky: Add here: Assuming

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -258,46 +279,116 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+     *     '{
+     *        val x = ???
+     *        val y = ???
+     *        { ... '{ ... x .. y ... } ... }.unary_~
+     *      }
+     *
```
odersky: Add: "Then the spliced subexpression"

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
nicolasstucki: In one case where the macro contained an anonymous function inside and the `tryHeal` saw that function as `ctx.owner`. Now the assert is
```scala
assert(ctx.owner.ownersIterator.exists(_.is(Macro)))
```

PR #4082, opened 2018-03-08
Don't force LazyRefs when computing `isProvisional`

odersky:
I saw errors of recursive evaluation of LazyRefs in the IDE which were caused by
the fact that `isProvisional` forced their computation. It should not do that.
It now treats all uncompleted LazyRefs as provisional.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -88,13 +88,13 @@ object Types {
 
 // ----- Tests -----------------------------------------------------
 
-    // debug only: a unique identifier for a type
-    val uniqId = {
-      nextId = nextId + 1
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -88,13 +88,13 @@ object Types {
 
 // ----- Tests -----------------------------------------------------
 
-    // debug only: a unique identifier for a type
-    val uniqId = {
-      nextId = nextId + 1
+//    // debug only: a unique identifier for a type
+//    val uniqId = {
```
smarter: Why did this get turned off?

odersky: It's a field in every type, and we don't need it - it's used nowhere.

PR #4084, opened 2018-03-08
Check that we don't leak uninstantiated type variables

smarter:
Type variable instantiation should only occur in Typer, TreeChecker and
during exhaustiveness checking, this means that no uninstantiated type
variable should exist outside of these phases. This was not enforced
until now, and was not always true before #4080.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/TreeChecker.scala
@@ -475,7 +475,9 @@ class TreeChecker extends Phase with SymTransformer {
 }
 
 object TreeChecker {
-  /** Check that TypeParamRefs and MethodParams refer to an enclosing type */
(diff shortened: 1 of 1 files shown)


smarter:
test performance please

odersky:
I don't think we need to test that explicitly since uninstantiated type variables cause anyway failures in both Pickler and -Ycheck. I found so far that this was sufficient. If we do want the more precise functionality for debugging I would suggest putting this under a Config option.

smarter:
> I don't think we need to test that explicitly since uninstantiated type variables cause anyway failures in both Pickler and -Ycheck

Where is this checked in Ycheck? The failure cases I saw before happened with type variables instantiated after Typer but before Pickler.

odersky:
If the problem was caused by Typer, I always got an error by turning -Ycheck:front on.

smarter:
I've replaced the content of this PR with a much simpler check in TreeChecker.

PR #4085, opened 2018-03-08
Add extension clauses

odersky:
This is an evolution of #4043. 

Prompted by the discussion there, I made three major changes

 - change syntax from `augment/extends` to `extend/implement` and in a later commit to `extend/:`
 - drop labels for extends clauses 
 - add scheme to generate stable names for extensions


compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -749,6 +758,130 @@ object desugar {
...
+  }
+
+  /**     extend <type-pattern> <params> implements <parents> { <body>} }
+   *   ->
+   *      implicit class <extension-name> <type-params> ($this: <extended>) <combined-params>
+   *      extends <parents> { <body1> }
+   *
+   *  where
+   *
+   *    <extension-name> = concatenation of all alphanumeric characters between and including `extend` and `{`,
```
adriaanm: Let's consider the impact on binary compatibility. Are there valid changes to the signature, which result in a new name,  but which should not affect binary compatibility? Generally, I think it's important to give the programmer the ability to specify a name explicitly.

sjrd: One easy example: exchange a fully-qualified for a relative name or vice versa.
```scala
augment List[String] {
```
would be binary-incompatible with
```scala
augment List[java.lang.String] {
```
under Martin's naming proposal.

Also, renaming a `type T` parameter into `type U`.

Plenty more like that.

DavidGregory084: Could use fully qualified names & a positional naming scheme for type params?

DavidGregory084: Sorry, I think what I meant to say was type patterns? The `type T` and `type U` that @sjrd mentioned above. I think you already have a double definition if you have an extension for `type T` and `type U` that share the same method name.

odersky: >  I think you already have a double definition if you have an extension for type T and type U that share the same method name.

No that's precisely it -  you want to enable multiple extends of the same type. So you should not get a double definition if you change the parameter type.

sjrd:
The blue sky sketch for extension-based typeclasses has at least two intrinsic issues:
* It fundamentally creates many more wrappers at run-time, which increases memory pressure and decreases performance, compared to the existing typeclass mechanism of Scala
* It does not address [@OlivierBlanvillain's example](https://github.com/lampepfl/dotty/pull/4043#issuecomment-371503170) with a single parameter having to honor two different typeclasses (`HasArea` and `HasShape` in that example).

and one extrinsic issue:
* The purely-functional ecosystem of Scala is too large now to do a 180° turn and go back to a completely different, incompatible way of encoding typeclasses in Scala.

docs/docs/reference/extend/instance-declarations.md
```diff
@@ -0,0 +1,78 @@
...
+
+For example, assume we have the following two traits, which define binary and unary (infix) equality tests:
+
+```scala
+trait Eql[T] {
+  def eql (x: T, y: T): Boolean
+}
+
+trait HasEql[T] {
+  def === (that: T)
```
japgolly: Missing `: Boolean`

docs/docs/reference/extend/extension-methods.md
```diff
@@ -0,0 +1,184 @@
...
+
+  private object WrappedResult {
+    def wrap[T](x: T): WrappedResult[T] = x
+    def unwrap[T](x: WrappedResult[T]): T = x
+  }
+
+  def result[T](implicit er: WrappedResult[T]): T = WrappedResult.unwrap(er)
+
+  extend (type T) {
+    def ensuring[U](condition: implicit WrappedResult[T] => Boolean): T = {
```
japgolly: `[U]` unused

mdedetrich:
I might be a bit late to the party here as well as missing some context, but what is the case for using `augment` rather than `implicit class`es?

I noticed that that proposal advocates for deprecating implicit classes in favor of "extension methods" (or augments as the PR says), what exactly are `augment`s gaining over `implicit class`es to warrant this?

What is the general impetus behind making a new augment feature, is it to replace typeclasses (in which case doesn't it make more sense to just make a language feature specifically for typeclass) or is it to make implicit classes nicer?

tests/pending/pos/blueSkyExtensions.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  extend Int : Monoid {
+    static def unit = 0
+    def + (that: Int) = this + that
+  }
+
+  extend String : Monoid {
+    static def unit = ""
+    def + (that: Int) = this ++ that
```
martijnhoekstra: `that: String` surely?

odersky:
Closed in favor of #4114

odersky:
@mdedetrich The motivation for this work is SIP 35 and #4028. To avoid feature duplication, it would be nice if we could remove value classes once we have opaque types. Value classes have proven to be somewhat a can of worms, because of complicated restrictions and a sometimes surprising boxing model. But extension methods rely on value classes. Hence, the question: Can we find a more direct way to express extension methods?

That's the first level. The other, somewhat independent question is whether we can or should extend the scheme to do extension methods to also do instance declarations in the sense of Rust. People's opinions are so far divided on this (which means, most are against, and I am holding out :-)

PR #4095, opened 2018-03-09
Mark PatternMatcherOld as disabled

nicolasstucki:


smarter:
Might as well delete it.

PR #4096, opened 2018-03-09
Cleanup SymUtils

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/SymUtils.scala
```diff
@@ -141,7 +138,7 @@ class SymUtils(val self: Symbol) extends AnyVal {
    *                  entered at this point).
    */
   def registerIfChild(late: Boolean = false)(implicit ctx: Context): Unit = {
-    def register(child: Symbol, parent: Type) = {
+    def register(child: Symbol, parent: Type): Unit = {
```
allanrenucci: Type annotation for inner method?

nicolasstucki: I could remove it. Should I?

nicolasstucki: Removed

PR #4101, opened 2018-03-11
Add error reporter for inlined quotes

nicolasstucki:


tests/neg/quote-reporter/Macro_1.scala
```diff
@@ -0,0 +1,8 @@
+import quoted._
+
+object Macro_1 {
+  inline def foo(inline b: Boolean): Unit = ~fooImpl(b)
+  def fooImpl(b: Boolean)(implicit reporter: Reporter): Expr[Unit] =
+    if (b) '(println("foo(true)"))
+    else reporter.error("foo cannot be called with false")
```
OlivierBlanvillain: What about using exceptions instead?

```scala
else throw new QuoteExpansionError("foo cannot be called with false")
```

The compiler could catch this particular one and forward the message to it's reporter. The downside compared to what you did in this PR it wouldn't cover warning, but it would also be easier to document.

nicolasstucki: That is the solution in the first commit (reverted in the third commit). It did not cover warnings of multiple errors. Though with exceptions it was simpler and more intuitive for the users IMO. On the other hand, the current implementation is more flexible.
WDYT? @OlivierBlanvillain, @biboudis @liufengyun

OlivierBlanvillain: I think having a single error and no distinction between errors and warnings shouldn't be a problem, `@implicitNotFound` and `@ambiguiousImplicits` works this way seem to be sufficient.

```scala
scala> import annotation.{implicitNotFound => inf}
import annotation.{implicitNotFound=>inf}

scala> @inf("A") trait A
defined trait A

scala> @inf("B") trait B
defined trait B

scala> def foo()(implicit a: A, b: B) = ()
foo: ()(implicit a: A, implicit b: B)Unit

scala> foo()
<console>:14: error: A
       foo()
          ^
```

odersky: I also tend to favor exceptions. Gives a simpler API for macros.

nicolasstucki: Then I will rework that version

nicolasstucki: Done

PR #4102, opened 2018-03-11
Improve implicit search involving SelectProtos over dependent methods

odersky:
Refine the isMatchedBy condition of SelectProto to take into account that sometimes we do not know yet what the members of a type could be. In these cases we now err on the side of generosity and assume that the type does match the selection.

This enables implicit searches for cases like

    a.f  ==>   implicitDeco(a).f

where `implicitDeco` is a dependent function, with a result type depending on `a` or an implicit argument of `implicitDeco`. So far neither dotty nor scalac supported this pattern.

Test case in typeclass-encoding.scala.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -97,8 +97,30 @@ object ProtoTypes {
   abstract case class SelectionProto(name: Name, memberProto: Type, compat: Compatibility, privateOK: Boolean)
   extends CachedProxyType with ProtoType with ValueTypeOrProto {
 
+    /** Is the set of members of this type unknown? This is the caes if:
+     *  1. The type has Nothing or Wildcard as a prefix or underlying type
+     *  2. The type has an uninstantiated TypeVar as a prefix or underlying type,
+     *  or as an upper bound of a prefix and underlying type.
+     */
+    private def hasUnknownMembers(tp: Type)(implicit ctx: Context): Boolean = tp match {
+      case tp: TypeVar => !tp.isInstantiated
+      case tp: WildcardType => true
+      case tp: TypeRef =>
+        val sym = tp.symbol
+        sym == defn.NothingClass ||
+        !sym.isStatic && {
+          hasUnknownMembers(tp.prefix) || {
+            val bound = tp.info.hiBound
+            bound.isProvisional && hasUnknownMembers(bound)
+          }
+        }
+      case tp: TypeProxy => hasUnknownMembers(tp.superType)
+      case _ => false
+    }
+
     override def isMatchedBy(tp1: Type)(implicit ctx: Context) = {
-      name == nme.WILDCARD || {
+      name == nme.WILDCARD || hasUnknownMembers(tp1) ||
+      {
         val mbr = if (privateOK) tp1.member(name) else tp1.nonPrivateMember(name)
         def qualifies(m: SingleDenotation) =
           memberProto.isRef(defn.UnitClass) ||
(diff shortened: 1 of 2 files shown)


odersky:
test performance please

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -97,8 +97,30 @@ object ProtoTypes {
   abstract case class SelectionProto(name: Name, memberProto: Type, compat: Compatibility, privateOK: Boolean)
   extends CachedProxyType with ProtoType with ValueTypeOrProto {
 
+    /** Is the set of members of this type unknown? This is the caes if:
```
smarter: typo: caes -> cases

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -97,8 +97,30 @@ object ProtoTypes {
...
+     *  1. The type has Nothing or Wildcard as a prefix or underlying type
+     *  2. The type has an uninstantiated TypeVar as a prefix or underlying type,
+     *  or as an upper bound of a prefix and underlying type.
```
smarter: `and` -> `or`, I think

tests/pos/typeclass-encoding.scala
```diff
@@ -0,0 +1,91 @@
...
+    def unit = ""
+    def inject($this: String): Monoid { type This = String } =
+      new Monoid {
+        type This = String
+        def add(that: This): This = $this ++ that
+      }
+  }
+
+  def sum[T](xs: List[T])(implicit $ev: Implementation[T] { type Implemented = Monoid } ) = {
+    (instance[T, Monoid].unit /: xs)((x, y) => inject(x) `add` y)
```
Blaisorblade: why backticks around `add`?

tests/pos/typeclass-encoding.scala
```diff
@@ -0,0 +1,91 @@
...
+    type Implemented <: TypeClass
+    def inject(x: From): Implemented { type This = From }
+  }
+
+  class CompanionOf[T] { type StaticPart[_] }
+
+  def instance[From, To <: TypeClass](
+      implicit ev1: Implementation[From] { type Implemented = To },
+      ev2: CompanionOf[To]): Implementation[From] { type Implemented = To } & ev2.StaticPart[From] =
+    ev1.asInstanceOf  // can we avoid the cast?
```
Blaisorblade: why not swap `ev1` and `ev2` and make `ev1` (now second argument) have type `Implementation[From] { type Implemented = To } & ev2.StaticPart[From]`? Let's ignore inference issues for a second.*

It seems all that could be replaced by replacing, e.g., `MonoidStatic[T] with Implementation[T]` with
```scala
class MonoidDict[T] extends Implementation[T] {
  def unit: T
   def inject($this: T) = new Monoid {
      type This = T
      def add(that: T): T
   }
}
```

In fact, at that point it's clearer that MonoidDict instances encode differently the same info as

```scala
trait MonoidDict[T] extends Implementation[T] {
  def unit: T
  def add($this: T, that: T): T
}
```

* If inference has problems with that, maybe we want to get curried implicits in first, but I'd first try simplifying the encoding further.

odersky: This is by no means the last word in encodings. It's a first attempt to see what it would take. We should experiment with lots of other schemes.

tests/pos/typeclass-encoding.scala
odersky: I am experimenting with the idiom to always require them around alphanumeric infix operators.  Trying to evaluate how natural that feels after some usage.

PR #4108, opened 2018-03-12
Add kind polymorphism

odersky:
This is an attempt to develop a minimal type safe form of (subtype-) kind polymorphism.  At present, it's intended as a basis for discussion. @milessabin @adriaanm, others: It would be great to get your opinion.

An overview of the proposed scheme is in kind-polymorphism.md


julienrf:
This is an interesting feature. It would be nice to see if it helps dealing with the kind-alignment issues we have in the new collections (between `Map` and `Iterable`). I remember having tried to play with typelevel-scala kind polymorphism support (which seems to have been the main source of inspiration here) and the result wasn’t better that *without* kind polymorphism: https://github.com/scala/scala-lang/pull/651#issuecomment-303650333

Also, I find it confusing to use an upper bound constraint to qualify the fact that a type parameter is poly-kinded. I think these are unrelated concepts. Maybe there is an analogy with the following at the value level? `def foo(bar: Any)`, where `bar` is “polymorphic” by virtue of the fact that `Any` is a supertype of all types. But I’m not sure that’s a solid reason… Maybe we should think of a special syntax?

odersky:
See also: https://github.com/scala/scala/pull/5538

docs/docs/reference/kind-polymorphism.md
```diff
@@ -0,0 +1,43 @@
...
+We call type parameters and abstract types with an `AnyKind` upper bound _any-kinded types_`.
+Since the actual kind of an any-kinded type is unknown, its usage must be heavily restricted: An any-kinded type
+can be neither the type of a value, nor can it be instantiated with type parameters. So about the only
+thing one can do with an any-kinded type is to pass it to another any-kinded type argument.
+Nevertheless, this is enough to achieve some interesting generalizations that work across kinds, typicially
+through advanced uses of implicits.
+
+(todo: insert good concise example)
+
+Some technical details: `AnyKind` is a synthesized class just like `Any`, but without any members. It extends no other class.
```
julienrf: So, no `equals` nor `hashCode`?

docs/docs/reference/kind-polymorphism.md
```diff
@@ -0,0 +1,43 @@
...
+
+A type can be used only as prescribed by its kind. Subtypes of `Any` cannot be applied to type arguments whereas subtypes of `[X] => Any`
+_must_ be applied to a type argument, unless they are passed to type parameters of the same kind.
+
+Sometimes we would like to have type parameters that can have more than one kind, for instance to define an implicit
+value that works for parameters of any kind. This is now possible through a form of (_subtype_) kind polymorphism.
+Kind polymorphism relies on the special type `scala.AnyKind` that can be used as an upper bound of a type.
+
+```scala
+def f[T <: AnyKind] = ...
```
jvican: So this seems to be true top parametric type that you proposed some time ago in Scala Contributors? :smile:

docs/docs/reference/kind-polymorphism.md
adriaanm: Since there's no way to create instances of `AnyKind`, I suppose it doesn't matter.

adriaanm:
Maybe add a few words on how this relates to Any/Nothing's implicit any-kindedness? At least in scalac, they are used whenever type inference hits its limits, regardless of the actual kind of the inferred type param. We can't use AnyKind instead of Any here, since it generally wouldn't meet the bounds.

milessabin:
This is great news! Happy to iterate on it here and on https://github.com/scala/scala/pull/5538.

odersky:
@adriaanm I am not sure `Any` can be used as a kind-polymorphic super. Here's what I tried:

    f: [X[_]]=> Unit

    scala> f[Any]

    scala> def f[X] = ()
    f: [X]=> Unit

    scala> f[Int]

    scala> f[List]
    <console>:14: error: type List takes type parameters
          f[List]
            ^

    scala> def f[X[_]] = ()
    f: [X[_]]=> Unit

    scala> f[Int]
    <console>:14: error: Int takes no type parameters, expected: one
          f[Int]
            ^

    scala> f[Any]
    // OK!

So, curiously, an `Any` type argument seems to be compatible with types of any variance, but it's not the other way round - an `Any` bounded type variable only accepts *-types.

In dotty, `Any` is only a supertype of *-types. That's a difference wrt scalc.

docs/docs/reference/kind-polymorphism.md
odersky: Yes, it's even topper than top since it's also above type constructors 😄

mandubian:
hi guys, happy to see this old draft PR reviving in dotty after I had left it in a draft state as nobody was giving me any return on it!
I had left it in a state where my 2 main issues were:
- between 2 compiling units, I couldn't reuse `AnyKind` information as I was erasing a bit harshly (as far as I can remember)
- to be able to write truly kind-polymorphic useful code, I would have really really liked to be able to identify a Kind & say "Type T and type U have same kind". Something like:

```
def f[T <: AnyKind[K], U <: AnyKind[K]] = ???
```

I don't know what would be `K` in this case but being able to "manipulate" the Kind of a type is the graal and maybe a step further than our original idea about minimalistic KP with @milessabin

milessabin:
I also came to the conclusion that to be useful and not cumbersome we needed a little more than the bare bones in the PoC.

The idea I'd like to experiment with is something along the lines of a `KindOf[T]` yielding a kind which can then be used as a bound. When I tweeted my New Year's "kind-polymorphism or bust" resolution Conor McBride responded: "The best way to achieve kind polymorphism in a language with type polymorphism is to abolish the distinction between types and kinds." "Haskell faced/faces the same problem. Going the long way round in small bursts has significant advantages. Any taxi driver will tell you.". I think this is good advice. Embedding kinds in the space of types fits quite nicely with a kind being a bound on a type variable.

docs/docs/reference/kind-polymorphism.md
soronpo: what is expected of `.asInstanceOf[AnyKind]`?

Daenyth: As "just as regular user" a ClassCastException seems perfectly reasonable.

TomasMikula:
I am very interested in kind polymorphism, but I see no mention of kind variables and "kind constructors" (`Kind = * | Kind -> Kind`) in this proposal. As a result, I don't see much use for the present limited form of kind polymorphism.

For example, how would I express that two type constructors are of the same kind?

```scala
// How to express that F and G are of the same kind?
trait Foo[F <: AnyKind, G <: AnyKind]
```

Or that one type constructor has the suitable kind to be passed as argument to another type constructor?

```scala
// How to express that F[G] is well-kinded?
trait Bar[F <: AnyKind, G <: AnyKind]
```

mandubian:
@TomasMikula Yes, I agree. I've rapidly encountered cases in which one wants to write:
```def foo[F[_ <: AnyKind[K]], G <: AnyKind[K]]```

File: docs/docs/reference/kind-polymorphism.md (hunk 1 above)
Comment (Blaisorblade) [reply docs/docs/reference/kind-polymorphism.md]: AsInstanceOf[AnyKind] should probably be a no-op, no?

Comment (odersky) [reply docs/docs/reference/kind-polymorphism.md]: -- Error: ../neg/anykind2.scala:19:17 ------------------------------------------
    19 |  1.asInstanceOf[AnyKind] // error
       |                 ^^^^^^^
       |                 missing type parameter(s) for AnyKind

**Comment by odersky** [comment]
How about delegating kind comparisons to implicits? I.e. define

```scala
trait SameKind[A <: AnyKind, B <: AnyKind]

implicit def same0[A, B]: SameKind[A, B] = new {}
implicit def same1[A[_], B[_]]: SameKind[A, B] = new {}
implicit def same2[A[_, _], B[_, _]]: SameKind[A, B] = new {}
```

Then one could write

```scala
def foo[F[X <: AnyKind], G <: AnyKind](implicit ev: SameKind[X, G])
```

Could that work?

odersky:
@TomasMikula I believe kind variables and kind constructors are way out of our complexity budget here.

milessabin:
@odersky I think that using implicits is the right way to go for the most part, but I think we need one or two more primitives, otherwise we end up with unacceptable amounts of boilerplate which would be better handled as a language primitive.

@mandubian and I pushed about as far as is possible with something equivalent to this PR combined with implicits of the sort you're suggesting and it wasn't quite enough.

odersky:
@milessabin I'll be interested to see what you come up with!

TomasMikula:
> How about delegating kind comparisons to implicits? I.e. define
> 
> ```scala
> trait SameKind[A <: AnyKind, B <: AnyKind]
> 
> implicit def same0[A, B]: SameKind[A, B] = new {}
> implicit def same1[A[_], B[_]]: SameKind[A, B] = new {}
> implicit def same2[A[_, _], B[_, _]]: SameKind[A, B] = new {}
> ...
> ```

I guess this would work, but is somewhat unsatisfactory 😕 

> ```scala
> def foo[F[X <: AnyKind], G <: AnyKind](implicit ev: SameKind[X, G])
> ```
> Could that work?

Not sure about this. Doesn't `X` in `F[X <: AnyKind]` go out of scope after that `]`?

> I believe kind variables and kind constructors are way out of our complexity budget here.

For now or for ever? If just for now, then I would rather wait. 😊

odersky:
> Not sure about this. Doesn't X in F[X <: AnyKind] go out of scope after that ]?

Indeed it does. Back to the drawing board, I guess.

> For now or for ever? If just for now, then I would rather wait. 😊

Well, the future is hard to predict... But I would say, for the foreseeable future it's out of budget.

mandubian:
If you want some crazier cases, I've refound these LoC proposed to me by Alexander Konovalov when he tried Kind-Poly and that couldn't work with current kind-poly.

The idea was to write kind-polymorphic Leibniz which would be extremely powerful for people writing libraries about abstract structures without rewriting 10 times the same code.

What was clearly missing was the possibility to manipulate a kind-polymorphic type constructors:

`* -> (T <: AnyKind)`  or `F[_  <: AnyKind]`

and write funny things like:

```  def lift[F[_ <: AnyKind, _ <: AnyKind], A <: AnyKind, B <: AnyKind]
    (ab: A === B): ({ type l[α <: AnyKind] = F[A, α] })#l === ({ type l[α <: AnyKind] = F[B, α] })#l```


```scala
import scala.language.higherKinds

object Test extends App {

  type F[A <: AnyKind] = Nothing
  type X = F[({type L[α <: AnyKind] = Unit})#L]

  sealed abstract class ===[A <: AnyKind, B <: AnyKind] {
    def subst[F[_<: AnyKind]](fa: F[A]): F[B]
  }
  /**
    * Given `A === B` we can prove that `F[A, ?] === F[B, ?]`.
    *
    * @see [[lift2]]
    * @see [[lift3]]
    */
  def lift[F[_ <: AnyKind, _ <: AnyKind], A <: AnyKind, B <: AnyKind]
    (ab: A === B): ({ type l[α <: AnyKind] = F[A, α] })#l === ({ type l[α <: AnyKind] = F[B, α] })#l = {
  // fails on the line above with:
  // type Λ$ takes type parameters
    type f[β <: AnyKind] = ({ type l[α <: AnyKind] = F[A, α] })#l === ({ type l[α <: AnyKind] = F[B, α] })#l
    ab.subst[f](refl[({ type l[α <: AnyKind] = F[A, α] })#l])
  }

  final case class Refl[A <: AnyKind]() extends ===[A, A] {
    def subst[F[_ <: AnyKind]](fa: F[A]): F[A] = fa
  }
  val anyRefl = Refl[AnyKind]()
  def unsafeForce[A <: AnyKind, B <: AnyKind]: A === B = anyRefl.asInstanceOf[A === B]
  def refl[A <: AnyKind]: A === A = unsafeForce[A, A]


  refl[List]

  trait Foo[F[_], A]
  val l: ({ type l[t] = Foo[List, t] })#l === ({ type l[t] = Foo[List, t] })#l = lift[Foo, List, List](refl[List])

  trait Bar[F[_], G[_]]
  val l2: ({ type l[t[_]] = Bar[List, t] })#l === ({ type l[t[_]] = Bar[List, t] })#l = lift[Bar, List, List](refl[List])


  type Ξ = scala.AnyKind

  trait Forall[F[_ <: Ξ]] {
    def apply[A <: Ξ]: F[A]
  }

  trait Exists[F[_ <: Ξ]] { fa =>
    type A <: Ξ
    def apply: F[A]
  }

  type ∀[F[_ <: Ξ]] = Forall[F]
  type ∃[F[_ <: Ξ]] = Exists[F]
  type ~>[A[_ <: Ξ], B[_ <: Ξ]] = FunctionK[A, B]
  type =~=[A[_ <: Ξ], B[_ <: Ξ]] = ===[A, B]

  abstract class FunctionK[F[_ <: Ξ], G[_ <: Ξ]] {
    def apply[A <: Ξ](fa: F[A]): G[A]
  }
  object FunctionK {
    def id[F[_ <: Ξ]]: FunctionK[F, F] = new FunctionK[F, F] {
      override def apply[A <: Ξ](fa: F[A]): F[A] = fa
    }
    def const[K](k: K): ∀[ ({ type λ[α[_ <: Ξ]] = α ~> ({ type λ[β <: Ξ] = K })#λ })#λ] = new ∀[({ type λ[α[_ <: Ξ]] = α ~> ({ type λ[β <: Ξ] = K })#λ })#λ] {
      override def apply[A[_ <: Ξ]]: A ~> ({ type λ[β <: Ξ] = K })#λ = new (A ~> ({ type λ[β <: Ξ] = K })#λ) {
        override def apply[B <: Ξ](fa: A[B]): K = k
      }
    }

    val t: List ~> ({ type λ[β <: Ξ] = Unit })#λ = const(()).apply[List]
  }

  // trait Foo2[F[_]]
  // new FunctionK[Foo2, List]{
  //   def apply[A[_]](fa: Foo2[A]): List[A] = ???
  // }

  // trait Category[->[_ <: Ξ, _ <: Ξ]] {
  //   def id[A <: Ξ]: A -> A
  // }
  // implicit val catzBaseFunctionKCategory: Category[FunctionK] = new Category[~>] {
  //   override def id[A <: Ξ]: A ~> A = FunctionK.id[A] // Impossible to define?
  // }

  sealed abstract class Liskov[-A <: Ξ, +B <: Ξ] { ab =>
    def subst[F[-_ <: Ξ]](fb: F[B]): F[A]
  }

  type <~<[A<:Ξ, B<:Ξ] = Liskov[A, B]

  private final case class Refl2[A <: Ξ]() extends Liskov[A, A] {
    def subst[F[-_ <: Ξ]](fa: F[A]): F[A] = fa
  }
  def id[A <: Ξ]: A <~< A = Refl2[A]()
  implicit def reify[A <: Ξ, B >: A <: Ξ]: A <~< B = id
  // all of these work:
  reify[Int, Any]
  reify[Int, AnyVal]
  reify[List, Any]

  reify[List, Int]
  reify[Right, Either]
}

```

odersky:
@mandubian that looks rather ... scary

mandubian:
@odersky yes it is scary XD, alex was testing advanced ideas... But, it's showing the superior limit where we can certainly not go so IMHO it's interesting...

I wrote wrong things above: ```F[_ <: AnyKind]``` is not ```* -> T <: AnyKind``` but more ```(* <: AnyKind) -> *```.

Ideally, I'd like to write such things in pseudo-scala-code:

```scala
trait Kinded[F <: AnyKind] {
   type WellFormed[_ <: AnyKind]
}

Kinded[Int] { type WellFormed = Int }
Kinded[List] { type WellFormed[A] = List[A] }
Kinded[Map] { type WellFormed[A, B] = Map[A, B] }
Kinded[Monad] { type K[F[_]] = Monad[F] }

trait SameKind[F <: AnyKind, G <: AnyKind]

abstract class FunctionK[F <: AnyKind, G <: AnyKind](implicit sk: SameKind[F, G]) {
  def apply[A <: AnyKind](fa: Kinded[F].WellFormed[A]): Kinded[G].WellFormed[A]
}
```

milessabin:
@odersky even as things stand there are immediate applications to things like `ClassTag`, `TypeTag`, shapeless's `Typeable` and the like.

odersky:
Test failed with:

    failed Vulpix meta: the output of sbt differs from the expected output
    expected : [info] Test dotty.tools.vulpix.VulpixMetaTests.compilePos started
    actual   : Testing tests/vulpix-tests/meta/pos/does-not-compile.scala

Who can help debug/fix this?

smarter:
> Who can help debug/fix this?

The meta-tests were added by @OlivierBlanvillain

odersky:
Seemed to have been a transient failure. After restart it passed. Let's watch it and disable if it occurs more often.

odersky:
I think this is ready to be reviewed. I would argue it is an addition with lots of upsides. With `AnyKind` we get the missing top-type over all kinds, which is the analogue of `Nothing`. So the result is more symmetric than before, and the added use cases are interesting. This does not preclude further refinements in later PRs, but for these we need more time to work them out and study them.

odersky:
test performance please

odersky:
Some thoughts about where one could take this further, without busting our complexity budget. One relatively simple way to make the scheme more precise would indeed be to parameterize `AnyKind` with a "kind". What is a kind? We know whether two types have the same kind, but how characterize the kind by itself? Keeping with the recommendation that kinds are types, we could take as a kind of type `T` the largest (wrt `<:<`) type that has the same kind as `T`. If we ignore bounds for the moment, this would give

    KindOf[Int] = Any
    KindOf[List]  =  [+X] => Any
    KindOf[Function1] = [-X,+Y] => Any
    KindOf[AnyKind] = AnyKind

With bounds it would be:

    KindOf[Int] = Any
    KindOf[List]  =  [+X >: Any <: Nothing] => Any
    KindOf[Function1] = [-X >: Any <: Nothing,+Y >: Any <: Nothing] => Any
    KindOf[AnyKind] = AnyKind

We could then change the subtype relation so that for any type `T`, `T <: AnyKind[KindOf[T]]`

`KindOf` could be given as a primitive, but it is also definable. Assuming type macros, one could write:

    type KindOf[T] = if (T <:< AnyKind[type K]) K else AnyKind

Without type macros one could still define it indirectly as a binary predicate:

    trait KindOf[T, K]
    implicit def kindOf[T <: AnyKind[K], K]: KindOf[T, K] = new {}

@milessabin Do you think that goes in the right direction?

mandubian:
@odersky I'm not @milessabin but I can tell you that it goes in the right direction IMHO (even if I need to think a bit more)...
I like the type representation of Kind with Types... I was looking at something like that without being able to formalize it.

odersky:
@mandubian Thanks for the feedback!

Blaisorblade:
The idea of turning kinds into supertypes is interesting, and reminds me of [pure subtype systems](http://lambda-the-ultimate.org/node/4835).

Given the amount of expressive power involved here, can this be marked as experimental as long as somebody with a clue looks at whether it's sound? I know this probably requires higherKinds, but this seems much more powerful.

In particular, is `[X <: AnyKind] T[X]` a first-class type constructor? Can it be used to instantiate a type parameter `Y <: AnyKind`? If both questions have a positive answer (unlike in other systems I'm familiar with), this seem to be _impredicative_ kind-polymorphism, and it's not obvious it can be safely desugared by kind-specializing definitions. Whether this can be exploited to introduce unsoundness is a trickier question, but after our discussions with Sandro I'm wary.

milessabin:
@odersky yes, that'd definitely going in the right direction.

TomasMikula:
@milessabin @odersky What's the motivation for `KindOf[T]`? Seems to me like doing things in reverse: instead of prescribing the kind for `T`, we try to deduce `T`'s kind using `KindOf`. Do you have a good use for `KindOf`?

I propose an alternative: define `Kind` (without bounds) as follows:

```scala
sealed trait Kind
case object * extends Kind
case class KindFun[K1, K2]() extends Kind
```

and then only have special support for

```scala
type AnyKind[K <: Kind]
```

namely that `T <: AnyKind[K]` means that `T` is of kind `K`.

odersky:
@TomasMikula Sure, that's an encoding of the standard definition of kinds. We know that one. But I don't think this will work in the Scala context.

 - you need to account for bounds and variances
 - it's definitely going to bust the complexity budget. We have so far terms and types. We don't want to deal with a separate category of kinds in either language specification or compiler.

milessabin:
I agree with @odersky that capturing bounds and variance is the big issue here. To model Scala kinds in the way you're suggesting you would need to go straight to something similar to GHCs ConstraintKinds.

TomasMikula:
Right, I accepted (for the moment) to represent kinds as types. I thought
my proposal would be a simplification: no magic KindOf, only magic AnyKind.
But I can't say certainly until I see KindOf in action.

alexknvl:
Tested some old code back from @mandubian's polykind fork, everything works so far
https://gist.github.com/alexknvl/deb31d4c23cd17521069740782aefb11

This, however, didn't:
```scala
object Foo {
  type R[K, z <: K] <: z
  type X = R[AnyKind, Option]
}

14 |  type X = R[AnyKind, Option]
   |                      ^^^^^^
   |           Type argument Option has not the same kind as its bound  <: K
```

alexknvl:
```scala
object Foo {
  type Z
    = [K <: AnyKind, z <: K, s[_ <: K] <: K] => z
  type S[N[K <: AnyKind, _ <: K, _[_ <: K] <: K] <: K]
    = [K <: AnyKind, z <: K, s[_ <: K] <: K] => s[N[K, z, s]]
  type +[M[K <: AnyKind, _ <: K, _[_ <: K] <: K] <: K,
         N[K <: AnyKind, _ <: K, _[_ <: K] <: K] <: K]
    = [K <: AnyKind, z <: K, s[_ <: K] <: K] => M[K, N[K, z, s], [T <: K] => s[T]]

  type N = S[S[Z]] + S[Z]
  // This COMPILES!
  type X = N[[x] => Any, Option, [f <: ([x] => Any)] => ([x] => Option[f[x]])]

  // These do NOT work
  // val x: X[Int] = x
  // val y: X[Int] = None
}
```

TomasMikula:
> I agree with @odersky that capturing bounds and variance is the big issue here.

I don't think we need any special treatment of bounds (but I could be missing something).

Accounting for variance seems simple, just add the following cases

```scala
case class CovariantKindFun[K1, K2]() extends KindFun[K1, K2]
case class ContravariantKindFun[K1, K2]() extends KindFun[K1, K2]
```

I'm just attempting to give an explicit inductive definition of kinds, that's all. I say it is simpler and preferable to having kinds defined only implicitly in the definition of `KindOf`. Of course, `AnyKind[K <: Kind]` will still need special treatment by the compiler.

It also seems to me that `KindOf` is unnecessary, but I will wait to see an example of how you intend to use it.

adriaanm:
You can't pass a bounded type constructor, say, `C[T <: Foo]` when an unbounded one, `C[T]`, is expected. When abstracting over the latter, you could apply `C` to a type argument that's not a subtype of `Foo`, which is "kind unsound" if the actual type constructor was bounded, just like it would be "type unsound" to consider a method's argument type to be in covariant position (that would result in run-time errors, whereas kind unsoundness is an undetected compile-time error). So, we should definitely track bounds (in addition to variance).

TomasMikula:
OK, I see. I think that a definition of `Kind` that accounts for bounds could still be written down as a Scala type (although it might itself be kind-polymorphic).

I'm arguing for this because I find it much better if the programmer can write down a kind explicitly as opposed to getting her hands on a kind only via `KindOf[T]`.

TomasMikula:
> "The best way to achieve kind polymorphism in a language with type polymorphism is to abolish the distinction between types and kinds."

I just want to note that this doesn't say that declaring a type parameter `T` to be of kind `K` has to be done via subtyping (`T <: K`), which we seem to have assumed so far. It might have been meant as types themselves having types. Then _"`T` is of kind `K`"_ would be written as `T: K` or `T.type <: K`.

milessabin:
@TomasMikula I agree. And Conor being Conor he most likely didn't have subtyping in mind. But this is Scala and we do have the subtype relation available ... why not use it?

OlivierBlanvillain:
I'm really not convinced by the complexity and boilerplate of the SemiGoup example.

How much do we gain compared to lifting everything to a high enough kind?

```scala
object playground {
  // Typeclass to decompose X into G[A, B]
  case class LiftKind[X, G[_, _], A, B](lift: X => G[A, B])
  trait  LiftKind0 {
    implicit def a[A]: LiftKind[A, [X, Y] => A, Any, Any]           = LiftKind(identity)
  }
  trait  LiftKind1 extends LiftKind0 {
    implicit def f[F[_], A]: LiftKind[F[A], [X, Y] => F[X], A, Any] = LiftKind(identity)
  }
  object LiftKind  extends LiftKind1 {
    implicit def g[G[_, _], A, B]: LiftKind[G[A, B], G, A, B]       = LiftKind(identity)
  }

  trait SemiGroup[G[_, _]] {
    def append[A, B](x: G[A, B], y: G[A, B]): G[A, B]
  }

  implicit object SemiGroupInt extends SemiGroup[[X, Y] => Int] {
    def append[A, B](x: Int, y: Int) = x + y
  }

  implicit object SemiGroupList extends SemiGroup[[X, Y] => List[X]] {
    def append[A, B](x: List[A], y: List[A]) = x ++ y
  }

  implicit object SemiGroupMap extends SemiGroup[Map] {
    def append[A, B](x: Map[A, B], y: Map[A, B]) = x ++ y
  }

  implicitly[SemiGroup[[X, Y] => Int]]    .append(5, 8)
  implicitly[SemiGroup[[X, Y] => List[X]]].append(List(1), List(3))
  implicitly[SemiGroup[Map]]              .append(Map("toto" -> 1L), Map("tata" -> 3L))

  def append[X, G[_, _], A, B](x: X, y: X)
    (implicit k: LiftKind[X, G, A, B], s: SemiGroup[G]): G[A, B] =
      s.append(k.lift(x), k.lift(y))

  val r1: Int               = append(5, 8)
  val r2: List[Int]         = append(List(1), List(3))
  val r3: Map[String, Long] = append(Map("toto" -> 1L), Map("tata" -> 3L))
}
```

milessabin:
@OlivierBlanvillain how high is "high enough"? For `ClassTag`, `Typeable`, `Generic` and the like there simply isn't a high enough kind.

OlivierBlanvillain:
@milessabin I would have to see applications of `AnyKind` to `ClassTag` / `Typeable` to have an opinion there, but I have a feeling that `* → *` is high enough for `Generic`. GHC.Generics uses that lifting trick to share their `Generic` typeclass and sum of product representation for both `*` and `* → *`. Sure, they can't deals with data-types of kind `* → * → *` and higher, but are there use cases for that?

milessabin:
@OlivierBlanvillain I disagree.

shapeless already has a `Generic1`, which is put to good use in [Kittens](https://github.com/milessabin/kittens). The only reason it doesn't go any higher is that I've balked at the pain of the endless boilerplate.

Apropos `ClassTag`, I believe Dotty has recently had issues due to `ClassTag[F[_]]` being used as if it were equivalent to `ClassTag[F]` /cc @Blaisorblade.

OlivierBlanvillain:
> The only reason it doesn't go any higher is that I've balked at the pain of the endless boilerplate.

I'm very sceptical `AnyKind` would help reduce boilerplate here, at least looking at the `SemiGroup`/`SemiGroupK` example...

alexknvl:
Types like [`===`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/Is.scala), [`<~<`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/As.scala) (and their bounded cousins [`Leibniz`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/Leibniz.scala) and [`Liskov`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/Liskov.scala)), [`=!=`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/WeakApart.scala), [`</<`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/StrictAs.scala), can be generalized to kind `k -> k -> *`. Presumably any type that depends on them will generalize as well ([`Constant :: k -> *`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/variance/Constant.scala), [`Covariant :: k -> *`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/variance/Covariant.scala), [`Injective :: k -> *`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/variance/Injective.scala)). I also have a thing called [`TypeId`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/TypeId.scala), which can be used used to recover type equality at runtime (given `TypeId[A]` and `TypeId[B]` you can recover `Option[A === B]`; and you can serialize them), which would work with other kinds. Types like `Exists` and `Forall` can be generalized as well.

odersky:
Putting on @Blaisorblade's pile to evaluate soundness a bit more before we commit.

Blaisorblade:
As a follow-up to my previous comment, Martin pointed out that even "experimental" features end up being as painful to remove as "supported" features.

As long as it's _impredicative_ kind-polymorphism the amount of extra expressive power is concerning. The only needed expressivity is _predicative_ kind-polymorphism; impredicativity arises because it appears much simpler to encode, but @milessabin do you see an easy way to enforce predicativity? We don't, but since predicative kind-polymorphism can be compiled away (by generating the needed kind-overloads of the kind-polymorphic definition), it clearly cannot add soundness issues that aren't already there.

Otherwise, I'll study the issues with impredicative kind-polymorphism next week.

milessabin:
@Blaisorblade just so I'm clear ... is your concern that a type could be directly self-applied? ie.,

```scala
class Rec[T <: AnyKind]
val rec: Rec[Rec] = ???
```

TomasMikula:
I am worried that saving some budget on design and implementation could bust everyone's complexity budget on the use site. Eating one's own dog food is a good sanity check, therefore I suggest to include a few kind-polymorphic data types in the standard library. Kind-polymorphic versions of the following would be good candidates:

 - polymorphic function (aka. natural transformation or `FunctionK`)
 - `=:=`
 - `<:<`

Blaisorblade:
> just so I'm clear ... is your concern that a type could be directly self-applied? ie.,

Yes, but more in general if kind-polymorphic types are just first-class types, _everything_ taking a type can be applied to them. In usual predicative settings, you can have (for instance) a distinction between "types" and "kind-polymorphic types" (like monotypes and polytypes in ML), or a hierarchy like in Agda (and probably Idris?) Set0: Set1 : Set2 : ...
Again, I'm not afraid of paradoxes but of type soundness issues.

But yes, that might be the simplest case where you can't just desugar the kind-polymorphic version by copy-pasted variants.

milessabin:
@Blaisorblade I think I need an example of something you think _might_ be problematic.

Pointing in the direction of impredicativity suggests a concern with something paradoxical, but I'm not seeing anything immediately troublesome here. Yes, `KindOf[T]` is itself a type, so I guess you could try something Russellian, like the kind of types which are not self-kinded, but I really don't see any problem with denying that there are such kinds ... I don't think we're under any obligation to accept that for every set of types there must be a corresponding kind.

milessabin:
I think the discussion in section 7 of [System FC with Explicit Kind Equality](http://www.seas.upenn.edu/~sweirich/papers/fckinds.pdf) (adding type-in-type to GHC) is relevant here,

> Blurring the distinction between types and kinds is convenient, but is it wise? It is well known that
type systems that include the Γ `ty ? : ? rule are inconsistent logics (Girard 1972). Does that cause trouble? For FC the answer is no—inconsistency here means that all kinds are inhabited, but even
without our extensions, all kinds are already inhabited.

compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala
```diff
@@ -356,8 +356,8 @@ private class ExtractAPICollector(implicit val ctx: Context) extends ThunkHolder
   private def computeType(tp: Type): api.Type = {
     // TODO: Never dealias. We currently have to dealias because
     // sbt main class discovery relies on the signature of the main
-    // method being fully dealiased. See https://github.com/sbt/zinc/issues/102
-    val tp2 = if (!tp.isHK) tp.dealiasKeepAnnots else tp
+    // method being fully dealiased. See https://github.com/sbt/zinc/isisLambdaSub/102
```
smarter: accidental replacement

Blaisorblade:
@milessabin I realize kind-polymorphism does not lead to type unsoundness in existing systems; I already explained why that's not sufficient evidence. While investigating, we're considering a `-Y` flag, also because this code is useful elsewhere (#4036).

compiler/src/dotty/tools/dotc/config/Config.scala
```diff
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = false
+  final val tracingEnabled = true
```
smarter: Need to be turned off.

Blaisorblade:
While I'll keep looking at soundness of impredicative kind-polymorphism, the PR should probably be merged because it simplifies the typechecker internally (which helps in #4036) and the feature under debate is under a flag. Assigning to @smarter for review of the code.

mandubian:
@Blaisorblade 3 questions:

- `if we decided that any type with an AnyKind-bounded type variable (or member) is not <: AnyKind` would mean `F[_ <: AnyKind] <: AnyKind` can't be written, right?

- is impredicative case something like `type Mu[F[_ <: AnyKind] <: AnyKind] <: AnyKind = F[Mu[F]]` ?

- in current implementation, would imprecative case be possible or not?

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -123,6 +123,9 @@ class ScalaSettings extends Settings.SettingGroup {
   val YprofileRunGcBetweenPhases = PhasesSetting("-Yprofile-run-gc", "Run a GC between phases - this allows heap size to be accurate at the expense of more time. Specify a list of phases, or *", "_")
       //.withPostSetHook( _ => YprofileEnabled.value = true )
 
+  // Extremely experimetnal language feature
+  val YkindPolymorphism = BooleanSetting("-Ykind-polymorphism", "Potentially unsound.")
```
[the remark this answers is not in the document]
ritschwumm: and the typo in "experimetnal"  should be fixed :p

docs/docs/reference/kind-polymorphism.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+We call type parameters and abstract types with an `AnyKind` upper bound _any-kinded types_`.
+Since the actual kind of an any-kinded type is unknown, its usage must be heavily restricted: An any-kinded type
+can be neither the type of a value, nor can it be instantiated with type parameters. So about the only
+thing one can do with an any-kinded type is to pass it to another any-kinded type argument.
+Nevertheless, this is enough to achieve some interesting generalizations that work across kinds, typicially
+through advanced uses of implicits.
+
+(todo: insert good concise example)
```
Blaisorblade: Afraid this should still be done. And we should update the docs to mark this as experimental.

Blaisorblade:
> I'd prefer to see the flag only cover the impredicative version rather both that and the predicative version. If we're agreed that the the restriction is safe then it'd be good to be able to move it forward in Scala.

Makes sense to me, but somebody would have to implement the restriction (and I don't have the needed skills yet). In fact one should spec the semantics — if you need to allow subtyping between `AnyKind`-polymorphic types, you end up with a separate subtyping lattice... I can try to think about the second part.

The best-known example of predicative polymorphism is type polymorphism in ML/Haskell98, where type arguments of polymorphic functions cannot be polymorphic types but must be monomorphic types.

> if we decided that any type with an AnyKind-bounded type variable (or member) is not <: AnyKind would mean F[_ <: AnyKind] <: AnyKind can't be written, right?

Correct.

> is impredicative case something like [...]

The example is simpler: consider `trait F[X <: AnyKind]; trait G[X <: AnyKind]`. Right now `F <: AnyKind` and `G <: AnyKind`, and that would be forbidden by predicative kind-polymorphism.
So, right now you can construct `F[F]` and also `G[F]`, and those applications would be forbidden as well.

> in current implementation, would imprecative case be possible or not?

It'd be possible: This PR implements (under `-Ykind-polymorphism`) impredicative kind-polymorphism.

smarter:
Merging this now because this PR contains useful subtyping fixes, but feel free to continue discussing here.

PR #4111, opened 2018-03-13
Run evaluation of inlined quotes in secured sandbox

nicolasstucki:


compiler/src/dotty/tools/dotc/util/Sandbox.scala
```diff
@@ -38,15 +38,18 @@ object Sandbox {
 
   object SandboxSecurityManager {
     @volatile private[this] var running: Int = 0
+    private def getRunning = running
+    private def incRunning() = running += 1
+    private def decRunning() = running -= 1
```
Blaisorblade: @nicolasstucki FWIW: if the object is shared and you want `incRunning` to be atomic, you do need `synchronized`, `volatile` is powerless to keep together the read and the write as a single atomic action.

nicolasstucki: I knlow, I stumbled with some other bug and I'm just trying to move things too see what fixes it.

nicolasstucki: BTW, the atomicity was given by the `synchronized` block in the only method that could access `incRunning`.

compiler/src/dotty/tools/dotc/util/Sandbox.scala
```diff
@@ -0,0 +1,104 @@
...
+            securityManager.enable() // Enable security manager on this thread
+            thunk
+          }
+        }
+      }
+      val thread = new SandboxThread
+      thread.start()
+      thread.join(timeout)
+      if (thread.isAlive) {
+        // TODO kill the thread?
```
Blaisorblade: Any chance to add a test with an infinite loop to test this case?

Because Dotty can be embedded in build tools VMs, this means we must kill the thread I think.

I'd use `thread.interrupt()` first, and specify/hope that the thread must not catch `InterruptedException`. If the thread doesn't handle `interrupt` by exiting, I'm not sure what's a sensible behavior — using `stop` might leave monitors held hence cause a deadlock, not using `stop` might leave a thread wasting CPU, so I suspect the thread should be `stop`ed and the compiler should not be reused.

Blaisorblade: Shared data that might be corrupted by `Thread.stop` includes
- shared static mutable data such as the name table (#1584), that is reused across all compiler instances in the same classloader. More such fields might exist.
- any compiler-specific data that is internally mutable and shared across threads. The second isn't a problem currently because compiler instances are currently not reused by build tools, though that might change in the future.

After [some discussion on Gitter and with Allan](https://gitter.im/scala/contributors?at=5ab10ebf6f8b4b99464a8dfa), I learned that build tools usually fork Scala compilers in the same JVM (though that's configurable) — they typically create fresh compiler instances (which is safer) though @smarter suggests it could be faster to reuse them sometimes (which is only safe for instances which didn't get corrupted).

OlivierBlanvillain:
test performance please

OlivierBlanvillain:
@nicolasstucki Why is this PR `on hold`?

nicolasstucki:
test performance please

nicolasstucki:

compiler/src/dotty/tools/dotc/util/Sandbox.scala
```diff
@@ -0,0 +1,104 @@
...
+    def enable(): Unit = {
+      enabledFlag.set(true)
+    }
+
+    override def checkPermission(permission: Permission): Unit = {
+      if (enabledFlag.get()) {
+        permission match {
+          case _: ReflectPermission =>
+            // allow reflection, needed for interpreter
+            // TODO restrict more?
```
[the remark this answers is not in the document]
nicolasstucki: Now that this code is not interpreted we can disallow all `ReflectPermission`

compiler/src/dotty/tools/dotc/util/Sandbox.scala
```diff
@@ -0,0 +1,104 @@
...
+
+import java.lang.reflect.{InvocationTargetException, ReflectPermission}
+import java.security.Permission
+
+import scala.quoted.QuoteError
+
+object Sandbox {
+
+  /** Timeout in milliseconds */
+  final val timeout = 3000 // TODO add a flag to allow custom timeouts
```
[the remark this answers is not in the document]
nicolasstucki: Adressed

PR #4113, opened 2018-03-13
vscode-dotty: Compile typescript with --strict

jendrikw:
Compiling typescript with the strict option is generally recommend to catch more type errors.

Also, I changed the update-all to a postinstall script, which will automatically execute after every `npm install`.

### DIFF
--- a/vscode-dotty/package.json
+++ b/vscode-dotty/package.json
@@ -49,8 +49,9 @@
     "tsc": "./node_modules/.bin/tsc",
     "vscode:prepublish": "npm run update-all && ./node_modules/.bin/tsc -p ./",
     "compile": "./node_modules/.bin/tsc -p ./",
-    "update-all": "npm install && node ./node_modules/vscode/bin/install && curl -L -o out/coursier https://github.com/coursier/coursier/raw/v1.0.0/coursier",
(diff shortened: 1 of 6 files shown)


vscode-dotty/package.json
```diff
@@ -49,8 +49,9 @@
     "tsc": "./node_modules/.bin/tsc",
     "vscode:prepublish": "npm run update-all && ./node_modules/.bin/tsc -p ./",
     "compile": "./node_modules/.bin/tsc -p ./",
-    "update-all": "npm install && node ./node_modules/vscode/bin/install && curl -L -o out/coursier https://github.com/coursier/coursier/raw/v1.0.0/coursier",
-    "test": "node ./node_modules/vscode/bin/test"
+    "update-all": "npm install",
```
smarter: I think I would keep this as it was before, it matches how other vscode extensions do it.

jendrikw: Do you know why it is done this way? This is exaclty the use case postinstall was made for.

smarter: I just checked and the default template does use postinstall instead of update-all now actually, so this change is fine but we should get rid of update-all while we're at it.

PR #4114, opened 2018-03-14
Add named extensions

odersky:
Following suggestions from @sjrd, a scheme to have named extensions, using
    
    extension <name> for <type> { ... }
    
syntax.

Based on #4028. The first commit of the new PR is 5dedefa. This one contains worked out docs but no implementation yet.

This is a third attempt, after #4043 and #4085.


odersky:
Rebased to profit from the FromTasty fixes for opaque.

odersky:
This PR is complete now. It depends on #4028 (opaque types) being accepted, since otherwise there's less motivation in having an extension mechanism separate from value classes.

japgolly:
Something to consider: as it currently stands with implicit value classes, there is a restriction that value classes can only reside in objects and not traits (nor classes if I remember correctly). This restriction can be bypassed by separating the class and the implicit conversion ([example](https://github.com/japgolly/scalajs-react/blob/master/scalaz-7.2/src/main/scala/japgolly/scalajs/react/internal/ScalazReactExt.scala#L58-L64)). This leads to two questions:

1. Will `extension`s be allowed to exist in traits?
2. Have you considered one would alias/export an `extension` that resides somewhere else?

If not, it might be as easy as
```scala
object SomewhereElse {
  extension ListOps[T] for List[T] {
    def second: T = tail.head
  }
}

trait MyExports {
  extension ListOps[T] = SomewhereElse.ListOps[T]
}
```

Note: Ideally we'd have a `export` keyword (dual to `import`) but unfortunately that never seems to get off the ground despite community enthusiasm :(

odersky:
@japgolly Good point about aliasing, but not sure about the syntax. Method extensions should be expanded to objects, so there it would be possible to write:

```scala
trait MyExports {
  val ListOps = SomewhereElse.ListOps
}
```

For instance declarations that's not so clear yet. Maybe having `extension` aliases is more natural, though. 

I am also quite positive about `export`, but did not get around to think about it deeply and have not seen other worked out proposals either. I would encourage a SIP, though.

japgolly:
> I am also quite positive about export, but did not get around to think about it deeply and have not seen other worked out proposals either. I would encourage a SIP, though.

Thanks, that's fantastic to know! It kind of makes me wonder if I should try my luck at creating a SIP...

PR #4116, opened 2018-03-15
Fix #3847: Handle correctly splicing of (possibly) nested type parameters

### LINKED ISSUE #3847: Missing hole in quote
```scala
import scala.quoted._
import scala.reflect.ClassTag
object Arrays {
  implicit def ArrayIsLiftable[T: Liftable](implicit t: Type[T], ct: Expr[ClassTag[T]]): Liftable[Array[T]] = (arr: Array[T]) => '{
    new Array[~t](~(arr.length: Expr[Int]))(~ct)
    // TODO add elements
  }
}
```
fails with
```scala
assertion failed: unresolved symbols: value t when pickling Foo.scala
	at dotty.tools.dotc.core.tasty.TreePickler.pickle(TreePickler.scala:635)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickle(PickledQuotes.scala:74)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickleQuote(PickledQuotes.scala:24)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.quotation(ReifyQuotes.scala:272)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.$anonfun$transform$3(ReifyQuotes.scala:320)
```

when trying to pickle the quote
```
dotty.runtime.Arrays.newGenericArray[t.unary_~]([[0| ]])([[1| ]])
```

Note that there should be a hole in place of `t.unary_~`.

biboudis:
Based on #4081

The proposed fix handle the case when a type splice e.g., `~t` appears in possibly nested positions as a type parameter, e.g., `List[~t]` and `List[Array[~t]]`. If the code included just `List[T]` and `List[Array[T]]` the types are extracted and added as tags using implicit search in the phase of `tryHeal`. This commit collects and handles spliced types as well, using the same mechanism as before.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
@@ -109,10 +109,15 @@ class ReifyQuotes extends MacroTransformWithImplicits {
     def inSplice = outer != null && !inQuote
 
     /** A map from type ref T to expressions of type `quoted.Type[T]`".
-     *  These will be turned into splices using `addTags`
+     *  These will be turned into splices using `addTags` and represent type variables
+     *  that can be possibly healed.
      */
     val importedTags = new mutable.LinkedHashMap[TypeRef, Tree]()
 
+    /** A map from type ref T to expressions of type `quoted.Type[T]`" like `importedTags`
+      * These will be turned into splices using `addTags` and represent types spliced
+      * explicitly.
+      */
     val explicitTags = new mutable.LinkedHashSet[TypeRef]()
 
     /** A stack of entered symbols, to be unwound after scope exit */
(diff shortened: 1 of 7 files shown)


nicolasstucki:
@biboudis you can rebase on master and remove commit 87d364c (which is in PR #4103).

nicolasstucki:
You probably can uncomment this https://github.com/lampepfl/dotty/blob/master/tests/run-with-compiler/quote-lib.scala#L123

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -415,6 +460,11 @@ class ReifyQuotes extends MacroTransformWithImplicits {
         tree match {
           case Quoted(quotedTree) =>
             quotation(quotedTree, tree)
+          case tree: TypeTree if tree.tpe.typeSymbol.isSplice =>
+            val splicedType = tree.tpe.asInstanceOf[TypeRef].prefix.termSymbol
+            splice(ref(splicedType).select(tpnme.UNARY_~))
+          case tree: TypeApply =>
+            super.transform(tree)
```
nicolasstucki: I think we do not need this branch

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -127,25 +134,60 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+        explicitTags.clear()
+
+        // Maps type splices to type references of tags e.g., ~t -> some type T$1
+        val map: Map[Type, Type] = tagsExplicitTypeDefsPairs.map(x => (x._1, x._2.symbol.typeRef)).toMap
+        val tMap = new TypeMap() {
+          override def apply(tp: Type): Type = {
+            if (map.contains(tp))
+              map.apply(tp)
+            else
+              mapOver(tp)
```
nicolasstucki: Shorter and more efficient:
```scala
override def apply(tp: Type): Type = map.getOrElse(tp, mapOver(tp))
```

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -127,25 +134,60 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+          }
+        }
+
+        Block(typeDefs ++ tagsExplicitTypeDefsPairs.map(_._2),
```
nicolasstucki: Isn't this just `Block(typeDefs ++ explicitTypeDefs,`?

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
biboudis: True! Not anymore!

biboudis:
Updated! Thanks for the review!

PR #4117, opened 2018-03-15
Do not pickle empty packages

smarter:
Unpickling them may lead to stub symbol errors


nicolasstucki:
We can probably also get rid of this now https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/FirstTransform.scala#L121

PR #4118, opened 2018-03-15
Do not allow named self in objects

smarter:
`object A { foo: Bla => }` was already prohibited, we extend this
restriction to `object A { foo => }`, the latter is not really useful
and it currently causes a MatchError in SymDenotation#sourceModule when
unpickling.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -645,7 +645,7 @@ object desugar {
         .withPos(mdef.pos.startPos)
       val ValDef(selfName, selfTpt, _) = impl.self
       val selfMods = impl.self.mods
-      if (!selfTpt.isEmpty) ctx.error(ObjectMayNotHaveSelfType(mdef), impl.self.pos)
+      if (!selfTpt.isEmpty || selfName != nme.WILDCARD) ctx.error(ObjectMayNotHaveSelfType(mdef), impl.self.pos)
       val clsSelf = ValDef(selfName, SingletonTypeTree(Ident(moduleName)), impl.self.rhs)
         .withMods(selfMods)
         .withPos(impl.self.pos orElse impl.pos.startPos)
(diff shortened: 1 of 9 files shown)


odersky:
I think we should keep allowing it. The main use for named self is that you don't have to remember the qualified this syntax to access an outer this. It's inconsistent to allow this for classes but not for modules.

smarter:
You can always refer to modules via their name so the named self doesn't seem to buy you much, but OK, I'll try to find another fix then.

odersky:
> You can always refer to modules via their name so the named self doesn't seem to buy you much, but OK, I'll try to find another fix then.

Good point! I withdraw my objection. We can drop the self type for modules then.

liufengyun:
The benchmarks for [stdlib](https://github.com/liufengyun/bench/tree/master/tests/scala-library) fail after this merge:

```Scala
# Warmup Iteration   1: -- [E013] Syntax Error: /data/workspace/bench/tests/scala-library/Traversable.scala:92:61
92 |object Traversable extends TraversableFactory[Traversable] { self =>
   |                                                             ^^^^
   |                                       objects must not have a self type

-- Error: /data/workspace/bench/tests/scala-library/immutable/Range.scala:156:4
156 |    validateMaxLength()
    |    ^^^^^^^^^^^^^^^^^
    |method validateMaxLength in class Range cannot be accessed as a member of Range(Range_this) from class CombinationsItr.
    | This location is in code that was inlined at /data/workspace/bench/tests/scala-library/SeqLike.scala:223
two errors found
```

Need to investigate more:

https://github.com/liufengyun/bench/blob/master/tests/scala-library/Traversable.scala#L92

smarter:
This was fixed in the scala2-library submodule, you can just cherry-pick the fix: https://github.com/lampepfl/scala/compare/e588767817e4985852f6547fc7cf211a8a7df3fa...723f22266db0a72e3ac19358586e3426757ac201

Blaisorblade:
So this is a new change for Scalafix to handle? I'm not sure removing such things is worth the trouble.

odersky:
After having played with it I now think we should keep it as it was. It _is_ annoying that one can abstract over self by name only in classes, but not in modules. And there's no hard reason why we should disallow it.

PR #4120, opened 2018-03-15
Fix tests on case-insensitive filesystems

nicolasstucki:


compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -190,19 +190,23 @@ class GenBCodePipeline(val entryPoints: List[Symbol], val int: DottyBackendInter
...
+              // Order is not deterministic so we enforce lexicographic order between the duplicates for error-reporting
+              val (cl1, cl2) =
+                if (claszSymbol.name.toString < dupClassSym.name.toString) (claszSymbol, dupClassSym)
+                else (dupClassSym, claszSymbol)
+              ctx.warning(s"Class ${cl1.name} differs only in case from ${cl2.name}. " +
+                          "Such classes will overwrite one another on case-insensitive filesystems.", cl1.pos)
+          }
         }
+        checkName(claszSymbol)
+        if (int.symHelper(claszSymbol).isTopLevelModuleClass)
```
smarter: I don't think this should be limited to top-level modules, the same problem exists for inner modules no?

allanrenucci:
@nicolasstucki Can you fix #4166 as part of this PR?

nicolasstucki:
#4166 is already fixed by this PR

tests/neg-custom-args/fatal-warnings/i2673c.scala
```diff
@@ -0,0 +1,6 @@
+package Foos
+
+object Outer {
+  class X // error
+  object x
```
allanrenucci: @nicolasstucki Why is this not allowed? It is only an issue when `class X` and `object x` are top level, right? Scalac does not complain

nicolasstucki: Those classes will emit `Outer$X.class` and `Outer$x.class` which may conflict.

allanrenucci: I don't think so. I think `Outer$x$.class` will be emitted. So there is no conflict. I think it is only an issue for top level objects

nicolasstucki: Yes, the object will generate `Outer$x$.class` but is the class needs a companion object it will generate `Outer$X$.class` which causes the conflict.

nicolasstucki: That would happen with
```scala
object Outer {
  case class X()
  object x
}
```
or any other class that generates the module class for it.

PR #4122, opened 2018-03-15
Fix #4098: Check that refinements have good bounds

### LINKED ISSUE #4098: Unsoundness due to projections from structural types
A new variant of #50 (re)surfacing, haven't checked yet among all the original examples, nor if this is fixed by the existing PRs (in particular, #4036). The issue is that `(({ type R = A }) & ({ type R = B }))#R` is allowed and shouldn't, so I adapted @alexknvl's code in #4060 to implement `coerce`.

```scala
object App {
  def coerce[U, V](u: U): V = {
    type X = { type R >: U }
    type Y = { type R = V }
    type Z = X & Y
    val u1: Z#R = u

    u1
  }

    val x: Int = coerce[String, Int]("a")
    println(x + 1)
  }
}
```

```scala
$ dotc -d out UnsoundProj.scala
$ dotr -classpath out App
        at App$.main(UnsoundProj.scala:11)
        at App.main(UnsoundProj.scala)
```

This was prompted by a question from @allanrenucci. He also pointed me to the documentation on type projections (http://dotty.epfl.ch/docs/reference/dropped/type-projection.html), which seems both unclear and misleading:

> Scala so far allowed general type projection T#A where T is an arbitrary type and A names a type member of T.
> Dotty disallows this if T is an abstract type (class types and type aliases are fine). This change was made because unrestricted type projection is unsound.

Projections should be illegal on lots of types `T` that aren't abstract types themselves — `T` must be a class type (or alias to such), and `A` should be a fully defined type member, but type operators like intersections, unions, structural refinements should be forbidden — probably all operators should (unless we have a very good reason for allowing some).

This way, type projections can only access global (type) constants:

```scala
class Defining {
  type TypeMember = ...
}
object Defining {
  val valueMember
}
// Here, Defining#T and Defining.valueMember are both references to known global constants, at value or type-level
```

odersky:
So far, we did the check only for member symbols. We have to do the same
for type refinements that do not refer to a member.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
+++ b/compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
@@ -126,6 +126,7 @@ class CheckRealizable(implicit ctx: Context) {
     val refinementProblems =
       for {
         name <- refinedNames(tp)
+        if (name.isTypeName)
         mbr <- tp.member(name).alternatives
         if !(mbr.info.loBound <:< mbr.info.hiBound)
       }
--- a/compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
+++ b/compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
@@ -30,8 +30,8 @@ object CheckRealizable {
   class NotFinal(sym: Symbol)(implicit ctx: Context)
   extends Realizability(i" refers to nonfinal $sym")
 
-  class HasProblemBounds(typ: SingleDenotation)(implicit ctx: Context)
-  extends Realizability(i" has a member $typ with possibly conflicting bounds ${typ.info.bounds.lo} <: ... <: ${typ.info.bounds.hi}")
+  class HasProblemBounds(name: Name, info: Type)(implicit ctx: Context)
+  extends Realizability(i" has a member $name with possibly conflicting bounds ${info.bounds.lo} <: ... <: ${info.bounds.hi}")
 
   class HasProblemBaseArg(typ: Type, argBounds: TypeBounds)(implicit ctx: Context)
   extends Realizability(i" has a base type $typ with possibly conflicting parameter bounds ${argBounds.lo} <: ... <: ${argBounds.hi}")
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
```diff
@@ -107,12 +115,21 @@ class CheckRealizable(implicit ctx: Context) {
    *       also lead to base types with bad bounds).
```
Blaisorblade: This comment should be updated.

compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
```diff
@@ -96,6 +96,14 @@ class CheckRealizable(implicit ctx: Context) {
       else boundsRealizability(tp).andAlso(memberRealizability(tp))
   }
 
+  private def refinedNames(tp: Type): Set[Name] = tp.dealias match {
+    case tp: RefinedType => refinedNames(tp.parent) + tp.refinedName
+    case tp: AndType => refinedNames(tp.tp1) ++ refinedNames(tp.tp2)
+    case tp: OrType  => refinedNames(tp.tp1) ++ refinedNames(tp.tp2)
```
Blaisorblade: It seems that `A | B` would need a test? Confused what happens there with disjoint ranges.

odersky: `refinedNames` only collects names that appear in a top-level refinement. So that's the same for & and |. The bounds checking is done later.

compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
odersky: Actually, the comment is now correct. The previous implementation did less than the comment implied.

compiler/src/dotty/tools/dotc/core/CheckRealizable.scala
Blaisorblade: I meant I don't see a testcase exercising this path.

Blaisorblade:
I think I can sort-of follow what happens and I can't see anything else wrong with this, but maybe it's good to take a look together with you or somebody who understands the codebase better.

odersky:
Going to merge now to escape extra work that will be cause by the SBT upgrade.

PR #4128, opened 2018-03-16
Various fixes for implicit/erased function types

smarter:


compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -392,14 +392,28 @@ object TastyFormat {
...
+  final val TYPELAMBDAtype = 169
+  final val LAMBDAtpt = 170
+  final val PARAMtype = 171
+  final val ANNOTATION = 172
+  final val TERMREFin = 173
+  final val TYPEREFin = 174
+
+  // In binary: 101100EI
+  // I = implicit method type
+  // E = erased method type
```
odersky: Missing test case? I did not see a test where a refinement had implicit or erased parameters. In fact, I am not sure we even want to support this. If we don't support it, we won't need to complicate the TASTY format. Compared to the rest of the format, which is rather lean, this does stick out like a sore thumb.

smarter: https://github.com/dotty-staging/dotty/blob/f66660a590cb488a4ea817dbcd0a5f8f49d68df9/tests/pos/implicit-dep.scala is a testcase for an implicit dependent function type, which ends up being desugared to a refinement with an implicit method

odersky: I see. Seems we have to support this then!

PR #4132, opened 2018-03-17
Fixes #3683 REPL: Incorrect pretty-printing of singleton types

### LINKED ISSUE #3683: REPL: Incorrect pretty-printing of singleton types
```scala
scala> val a = "foo" 
val a: String = "foo"
scala> val x: a.type = a 
val x: String(a) = "foo"
```

This should be `a.type`, not `String(a)`.

eklavya:


eklavya:
I will fix the tests if this fix is ok.

allanrenucci:
Closing. Will be fixed by #4537. Thanks @eklavya!

eklavya:
@allanrenucci I am not able to differentiate between ```x: a.type``` being singleton and ```final val x``` being singleton.

Blaisorblade:
TL;DR
Talking with @allanrenucci we agreed that for constants like `"foo"` or `1` (in code, `ConstantType`) we want to use the constant as a singleton type (no `.type` added, so that they're legal syntax):
```
scala> val a: "foo" = "foo"
val a: "foo" = "foo"
scala> val b: "foo" = a
val b: "foo" = "foo"
```
but for other singleton types (`SingletonType`) that refer to terms we need to follow the request in #3683:
```
scala> val c: a.type = a 
val c: a.type = "foo"
```
`ConstantType` extends `SingletonType`, so this requires adding a new branch.

(Also wondering what happens on the other subclasses of `SingletonType`, but I hope adding `.type` works there too).

## Details

@allanrenucci convinced me to reconsider. We can replace `String("a")` with `"a"`, since those are SIP-23 literal types, which are all primitives, hence it's reasonable to assume users can find their types from the literals. Also, writing `"a"` means we output legal syntax. However, `"a".type` is not valid syntax, so we shouldn't add `.type` there.

> I am not able to differentiate between x: a.type being singleton and final val x being singleton.

Sorry for being unclear, you don't need to find whether `final` was used; I just wrote `final val a = "foo"` as a shortcut for writing `val a: "foo" = "foo"`.

eklavya:
Adding `ConstantType` case breaks defs :(

In the test "methodDoesNotTakeMorePrameters" - 
`expected:<[((a: Int): Unit)(Scope.foo)]> but was:<[Scope.foo.type]>`

eklavya:
@allanrenucci pushed, I am messing around with the types of SingletonType to figure out a way to make it work without changing existing behaviour on defs, which I can't seem to find.

allanrenucci:
@eklavya I pushed a commit that fixes the error message test failures. For the other test failures you should update the tests

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -141,6 +141,10 @@ class PlainPrinter(_ctx: Context) extends Printer {
         ParamRefNameString(tp) ~ ".type"
       case tp: TypeParamRef =>
         ParamRefNameString(tp) ~ lambdaHash(tp.binder)
+      case tp: ConstantType =>
```
allanrenucci: Should be:
```scala
case ConstantType(value) =>
 toText(value)
```

compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -141,6 +141,10 @@ class PlainPrinter(_ctx: Context) extends Printer {
...
+        toTextRef(tp)
+      case tp: TermRef =>
+        toTextPrefix(tp.prefix) ~ selectionString(tp) ~ ".type"
```
allanrenucci: Why special case `TermRef`?

eklavya: This is for `a.type` singleton types.

```scala
scala> final val a = "foo"
val a: "foo" = foo

scala> val x: a.type = a
val x: a.type = foo
```
a.type is made by special casing TermRef. I am sorry if that's nonsensical 😛

Blaisorblade: That makes some sense, but there are other cases above for `tp: TermRef` which already produce `toTextRef(tp) ~ ".type"` (which is equivalent to the code here), do they not apply?

eklavya: I had missed that. No, those cases (in PlainPrinter) are not applied because of the guards.

OlivierBlanvillain:
@eklavya The PR is getting really old... If you plan to work again on it please reopen or make a new one, I will close it for now.

PR #4135, opened 2018-03-18
Move LiftExprOps to quoted package

nicolasstucki:
To use `import scala.quotes._` to import `toExpr` extension


odersky:
Should it not be in the companion object of `Expr`? That would correspond to the rule that an implicit conversion from `A` to `B` must be co-defined with either `A` or `B`. In that case, the conversion is available without needing an import.

nicolasstucki:
This one is just an extension method. The implicit conversion is still in the object `Expr`.

odersky:
Ah I misunderstood.

PR #4137, opened 2018-03-18
Several type system fixes

odersky:
Prompted by a study of possible typeclass encodings several small fixes to aspects of type checking.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -1967,6 +1967,7 @@ object Types {
     def derivedSelect(prefix: Type)(implicit ctx: Context): Type =
       if (prefix eq this.prefix) this
       else if (prefix.isBottomType) prefix
+      else if (prefix.isInstanceOf[WildcardType]) WildcardType
       else if (isType) {
         val res =
           if (currentSymbol.is(ClassTypeParam)) argForParam(prefix)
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -1967,6 +1967,7 @@ object Types {
     def derivedSelect(prefix: Type)(implicit ctx: Context): Type =
       if (prefix eq this.prefix) this
       else if (prefix.isBottomType) prefix
+      else if (prefix.isInstanceOf[WildcardType]) WildcardType
```
smarter: There's what look like an identical case at the end of this method (`case _: WildcardType => WildcardType`), can the two be merged?

tests/pos/typeclass-encoding2.scala
```diff
@@ -81,22 +81,22 @@ object semiGroups {
...
   }
 
-  implicit object IntOps extends Extension[Int, Monoid] with MonoidCommon { self =>
+  implicit object IntOps extends Extension[Int, Monoid] with MonoidCommon {
     type This = Int
     type Instance = Monoid
     def unit: Int = 0
     def inject($this: Int) = new Monoid {
-      val common: self.type = self
+      val common: IntOps.this.type = IntOps.this
```
smarter: I think this could be `: IntOps.type = IntOps`

odersky: That's the same, but less regular. I am after an encoding that works the same for classes and objects. Unfortunately, with self types for objects dropped, we only have qualified this to avoid an irregularity.

odersky:
test performance please

odersky:
test performance please

PR #4139, opened 2018-03-19
Fix owners of AsFunction applied expressions

nicolasstucki:


compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -143,7 +143,7 @@ object PickledQuotes {
             case tree: Ident if tree.symbol == paramSym => x1Ref().withPos(tree.pos)
             case _ => super.transform(tree)
           }
-        }.transform(ddef.rhs)
+        }.transform(ddef.rhs).changeOwner(ddef.symbol, ctx.owner)
```
smarter: To be more efficient, you can replace your TreeMap by a TreeTypeMap(oldOwners = ..., newOwners = ...), and avoid the need for changeOwner (which would create its own TreeTypeMap)

nicolasstucki: Will change it

nicolasstucki:
Tests are failing on maser :( 

@odersky told me he already has a fix for it.

Blaisorblade:
Cherry-picked likely fix in #4141.

Blaisorblade:
And new build at http://dotty-ci.epfl.ch/lampepfl/dotty/3824

PR #4141, opened 2018-03-19
Avoid spurious "double-def" error message involving erroneous definition

Blaisorblade:
Looks like this fixes master's failure.


Blaisorblade:
Specifically, it fixes tests/neg/i4058.scala which is the file mentioned in http://dotty-ci.epfl.ch/lampepfl/dotty/3814/6 and which is affected by https://github.com/lampepfl/dotty/commit/ad002ffe3e1f62196e3894f4cec669a8e91d0a1e (the commit which introduced the issue).

Blaisorblade:
BTW this suggests a deeper issue: already in #4064 we had seen that turning string errors into proper errors affected errors counts.

Blaisorblade:
Self-merging since this appears to indeed fix `master` and was reviewed. Thanks!

PR #4142, opened 2018-03-19
Make VulpixMetaTests more deterministic

Blaisorblade:
Triggered by debugging the failure in http://dotty-ci.epfl.ch/lampepfl/dotty/3812/4. It might not fix the issue but shouldn't hurt. Mostly due to @allanrenucci's suggestions.

### DIFF
--- a/compiler/test/dotty/tools/vulpix/VulpixMetaTests.scala
+++ b/compiler/test/dotty/tools/vulpix/VulpixMetaTests.scala
@@ -8,16 +8,18 @@ import dotty.Properties
 import TestConfiguration._
 
 /** Meta tests for the Vulpix test suite. This test follows the structure of
- *  CompilationTests.scala. It is ment to be called from bash to diff with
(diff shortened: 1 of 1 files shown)


compiler/test/dotty/tools/vulpix/VulpixMetaTests.scala
```diff
@@ -8,16 +8,18 @@ import dotty.Properties
...
 @Category(Array(classOf[dotty.VulpixMetaTests]))
 class VulpixMetaTests extends ParallelTesting {
   def maxDuration = 1.seconds
-  def numberOfSlaves = 5
-  def safeMode = Properties.testsSafeMode
-  def isInteractive = SummaryReport.isInteractive
-  def testFilter = Properties.testsFilter
+  // Ensure maximum reproducibility.
+  def numberOfSlaves = 1
+  def safeMode = false // Don't fork separate VMs
```
allanrenucci: A more accurate comment would be `Don't fork a new VM after each run test`. Run tests are still run in a separate VM

Issue #4151, opened 2018-03-21
Disallow inline macro with statements before the splice


nicolasstucki:
We should disallow statements before the call to the top level splice of the macro such as 
```
inline def foo: Int = {
  println("foo") // should be an error
  ~impl()
}
def impl(): Expr[Int] = '(1)
```
The correct way to inline the statements would be to put them in the quoted code itself
```
inline def foo: Int = ~impl()
def impl(): Expr[Int] = '{
  println("foo")
  1
}
```
or
```
inline def foo: Int = ~{ 
  '{ 
    println("foo")
    ~impl()
  }
}
def impl(): Expr[Int] = '(1)
```
The motivation for this is to always know statically all the arguments of the top level splice. These argouments are the arguments of the inline macro.

It is not enough to check if the rhs is a splice as the following code should work and may generate a block without statements
```scala
inline def foo: Int = {
  ~impl()
}
```

PR #4155, opened 2018-03-22
Fix #3853 and #4151: Compile code in the macro definition and remove interpreter

### LINKED ISSUE #3853: Spliced quote interpreter
We currently define macros with inline and quotes link this
```scala
object Macros {
  inline def foo(inline n: Int): Int = ~fooImpl(n)
  def fooImpl(n: Int): Expr[Int] = n + 1 // Lifted by quoted.Liftables
}
```

When a `foo` is inlined by the compiler, the code/tree for `fooImpl(n)` is interpreted. This assumes that `fooImpl` was already compiled and is static.

Now the question is what should we do with 
```scala
object Macros {
  inline def foo(inline n: Int): Int = ~{
    (n + 1): Expr[Int]
  }
}
```

In the current scheme, we would need to interpret `n + 1` (or `1 + 1` after inlining). This implies that we need to be able to interpret at least all primitive operations. Potentially ending implementing a full interpreter for Scala (which sounds fun but extremely hard).

One alternative would be to extract the contents of `~{...}` into a helper method which would receive all the parameters that `foo` receives. All inliened parameters are passed as they are and all non-inlined are passed as `Expr[T]` (i.e. as trees). Basically desugaring into the original implementation.
```scala
inline def foo(inline n: Int, m: Int): Int = ~fooImpl(n)
def foo$macro(n: Int) = (n + 1): Expr[Int] // Should be a static location
```

This comes with the following advantages
* The interpreter logic can be kept simple
* Performace of the evaluation of the `Expr[T]` and the only reflective call is to `foo$macro`.

nicolasstucki:
The only interpreted code remaining is a call to a static method
containing a lambda which can compute the resulting inlined quoted
expression. Arguments to the macro are trivially computed from the
inlined call and bindings.


nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
@liufengyun do you know what failed in the benchmark server?

liufengyun:
Sorry, I have no access to computer now, but you should be able to login
into lamppc37 with your account and check the log file. Or, I can check in
one hour.

Le ven. 23 mars 2018 17:22, Nicolas Stucki <notifications@github.com> a
écrit :

liufengyun:
Cause diagnosed: in #4149 , an empty file `Foo.scala` is incorrectly added to the repo.

That error should not cause the benchmarks to fail, but the following `find` command failed due to the existence of `Foo.scala` in current directory. If I remove `Foo.scala`, everything works correctly.

```bash
find /data/workspace/bench/tests/scala-library -name *.scala
```

nicolasstucki:
@liufengyun thanks :)

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -231,7 +231,9 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
           //     be duplicated
           //  2. To enable correct pickling (calls can share symbols with the inlined code, which
           //     would trigger an assertion when pickling).
-          val callTrace = Ident(call.symbol.topLevelClass.typeRef).withPos(call.pos)
+          val callTrace =
+            if (call.symbol.is(Macro)) call
```
odersky: Why make the exception for macros? Since there is a reasoning in the comment above why we rewrite calls to callTraces, we need to add to the comment a reasoning why the same does not apply to macro calls.

nicolasstucki: I will add a comment

nicolasstucki:
Removing the interpreter does not really limit what we could already write in terms of macros. @odersky do you have some precise examples where we would need interpretation?

This transformation can also be extended to support multiple spliced in a single macro by making the macro return a list of lambdas. We would only need to find the additional definitions in the body of the inline method (such as `y`) and also pass them to the lambdas.
```
def foo(inline x: Int): Int = {
  val y  = 3
  ~impl2(x) + ~impl3('(y))
}
```
This could also be performed by the interpreter, but is currently not supported as well.

One thing is not possible anymore with the changes in this PR. We cannot write inline macros as extension methods in implicit classes. Those were only working by an abuse of the inline bindings which were wrongly moved to be interpreted to allow this to work. The failed if the implicit class was instantiated explicitly. The correct way to would be using the new extension methods, where we might additionally need to require the `this` parameter to be an `inline` parameter of the resulting method (this would need a direct encoding of the extensions without value classes).

compiler/src/dotty/tools/dotc/transform/PostTyper.scala
nicolasstucki: Added the comment and making this optimization at `ReifyQuotes` when the `call` is processed.

PR #4156, opened 2018-03-22
Optimize findRef

odersky:
... by inlining two frequently used local methods

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -549,7 +549,7 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
         if (inst.exists) sigName(inst) else tpnme.Uninstantiated
       case tp: TypeProxy =>
         sigName(tp.underlying)
-      case _: ErrorType | WildcardType =>
+      case _: ErrorType | WildcardType | NoType =>
         tpnme.WILDCARD
       case tp: WildcardType =>
         sigName(tp.optBounds)
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -508,10 +508,20 @@ object Types {
             case tp1 => tp1
           })
         case tp: TypeRef =>
-          tp.denot.findMember(name, pre, excluded)
+          tp.denot match {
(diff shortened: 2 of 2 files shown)


odersky:
test performance please

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -549,7 +549,7 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
         if (inst.exists) sigName(inst) else tpnme.Uninstantiated
       case tp: TypeProxy =>
         sigName(tp.underlying)
-      case _: ErrorType | WildcardType =>
+      case _: ErrorType | WildcardType | NoType =>
         tpnme.WILDCARD
       case tp: WildcardType =>
         sigName(tp.optBounds)
```
Blaisorblade: Does this code match _twice_ against `WildcardType`? I don't get it. In fact, the latter matches appear to be erased away:

```
scala> def foo: Any => Any = { case _: String | Int => 1; case _: Int => 2 }
def foo: Any => Any
scala> foo(1)
val res0: Any = 2
scala> def foo: Any => Any = { case _: String | AnyRef => 1; case _: AnyRef => 2 }
def foo: Any => Any
scala> foo(new AnyRef())
val res1: Any = 2
```

smarter: It first matches against the object `WildcardType`, then the case below matches against any instance of the class `WildcardType`.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -508,10 +508,20 @@ object Types {
...
-        case tp: ThisType =>
+          tp.tycon match {
+            case tc: TypeRef if tc.symbol.isClass =>
+              go(tc)
+            case tc: HKTypeLambda =>
+              goApplied(tp, tc)
+            case _ =>
+              go(tp.superType)
+          }
+        case tp: ThisType => // ??? inline
```
smarter: Stray comment?

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
Blaisorblade: Yeah sorry, Allan explained me what's going on. But do we really know that the only instance of `WildcardType(NoType)` is the `WildcardType` object?

PR #4159, opened 2018-03-22
Fix #4157: Broken error message when trying to eta-expand

### LINKED ISSUE #4157: Broken error message when trying to eta-expand a by-name
```scala
object Test {
  def foo(x: => Int) ={
    val a = x _ // error
    val b: () => Int = x _ // error
  }
}
```

The first error message is correct, but the second leaks the ErrorType and thus looks confusing:
```scala
-- [E099] Syntax Error: try/byeta.scala:3:14 -----------------------------------
3 |    val a = x _ // error
  |            ^^^
  |            Not a function: => Int(x): cannot be followed by _

-- [E099] Syntax Error: try/byeta.scala:4:25 -----------------------------------
4 |    val b: () => Int = x _ // error
  |                       ^^^
  |                       Not a function: <error found:    Int
  |                       required: () => Int
  |                       
  |                       >: cannot be followed by _
```

smarter:
Avoid leaking error types in the printed error messages

### DIFF
--- a/compiler/test-resources/repl/errmsgs
+++ b/compiler/test-resources/repl/errmsgs
@@ -72,3 +72,11 @@ scala> val a: iDontExist = 1
 1 | val a: iDontExist = 1
   |        ^^^^^^^^^^
   |        not found: type iDontExist
+scala> def foo1(x: => Int) = x _
+1 | def foo1(x: => Int) = x _
+  |                       ^^^
+  |Only function types can be followed by _ but the current expression has type Int
+scala> def foo2(x: => Int): () => Int = x _
+1 | def foo2(x: => Int): () => Int = x _
+  |                                  ^^^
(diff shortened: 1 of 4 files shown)


compiler/test-resources/repl/errmsgs
```diff
@@ -72,3 +72,11 @@ scala> val a: iDontExist = 1
 1 | val a: iDontExist = 1
   |        ^^^^^^^^^^
   |        not found: type iDontExist
+scala> def foo1(x: => Int) = x _
```
allanrenucci: I would make it an error message test instead. They are less fragile

smarter: But I really want to know what the exact output is here :). Can you help me figure out why the test doesn't pass?

PR #4162, opened 2018-03-22
Fix issues to make new collections compile again

odersky:
1. Revert #4118 
 2. Better disambiguation to avoid MergeErrors

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -529,6 +529,9 @@ object Denotations {
               try infoMeet(info1, info2)
               catch {
                 case ex: MergeError =>
+                  // TODO: this picks one type over the other whereas it might be better
+                  // to return a MultiDenotation instead. But doing so would affect lots of
+                  // things, starting with the return type of this method.
                   if (preferSym(sym2, sym1)) info2
                   else if (preferSym(sym1, sym2)) info1
                   else if (pre.widen.classSymbol.is(Scala2x) || ctx.scala2Mode)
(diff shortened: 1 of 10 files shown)


compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -523,7 +529,9 @@ object Denotations {
               try infoMeet(info1, info2)
               catch {
                 case ex: MergeError =>
-                  if (pre.widen.classSymbol.is(Scala2x) || ctx.scala2Mode)
+                  if (preferSym(sym2, sym1)) info2
```
smarter: Would it be difficult to construct a test case to exercise this? Also, are we confident this is sound?

vscode-dotty/test/extension.test.ts
```diff
@@ -12,10 +12,10 @@ import * as vscode from 'vscode';
 import * as dotty from '../src/extension';
 
 // Defines a Mocha test suite to group tests of similar kind together
-suite("vscode-dotty tests", function () {
+//suite("vscode-dotty tests", function () {
```
smarter: Shouldn't be necessary anymore.

compiler/src/dotty/tools/dotc/core/Denotations.scala
odersky: I tried for a while but did not manage. The code that breaks is in class `Id` in `MapView` of the collection strawman.

vscode-dotty/test/extension.test.ts
odersky: That was added by accident. I removed it now.

compiler/src/dotty/tools/dotc/core/Denotations.scala
odersky: Regarding soundness: What happens is that we get two denotations that should be overloaded and we pick one and forget about the other. I see that as a possible problem for completeness, not sure how it would affect soundness.

smarter: Maybe add a comment to note that this may be worth revisiting at some point?

PR #4165, opened 2018-03-23
Avoid "missing parameter type" error when checking `isMatchedBy`

odersky:
When checking whether a FunProto is applicable to an argument, we should avoid
falling into "missing parameter type" errors. Better to assume a WildcardType
for the argument instead. This follows the general principle to err on the side
of generosity when checking `isMatchedBy`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -212,7 +212,7 @@ object ProtoTypes {
     private[this] var evalState: SimpleIdentityMap[untpd.Tree, (TyperState, Constraint)] = SimpleIdentityMap.Empty
 
     def isMatchedBy(tp: Type)(implicit ctx: Context) =
-      typer.isApplicable(tp, Nil, typedArgs, resultType)
+      typer.isApplicable(tp, Nil, unforcedTypedArgs, resultType)
 
     def derivedFunProto(args: List[untpd.Tree] = this.args, resultType: Type, typer: Typer = this.typer) =
       if ((args eq this.args) && (resultType eq this.resultType) && (typer eq this.typer)) this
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -240,33 +240,45 @@ object ProtoTypes {
...
-    private def cacheTypedArg(arg: untpd.Tree, typerFn: untpd.Tree => Tree)(implicit ctx: Context): Tree = {
+    private def cacheTypedArg(arg: untpd.Tree, typerFn: untpd.Tree => Tree, force: Boolean)(implicit ctx: Context): Tree = {
       var targ = myTypedArg(arg)
       if (targ == null) {
-        targ = typerFn(arg)
-        if (!ctx.reporter.hasPending) {
-          myTypedArg = myTypedArg.updated(arg, targ)
-          evalState = evalState.updated(arg, (ctx.typerState, ctx.typerState.constraint))
+        if (!force && untpd.functionWithUnknownParamType(arg).isDefined)
+          // If force = true, assume ? rather than reporting an error.
```
smarter: You mean, force = false ?

odersky: Oops, yes.

odersky:
I got:

Running Vulpix meta test
++ mktemp
+ tmp=/tmp/tmp.NBPEpn
+ ./project/scripts/sbt 'dotty-compiler/testOnly dotty.tools.vulpix.VulpixMetaTests'
++ mktemp
+ tmp1=/tmp/tmp.NaCJLh
+ sed '/Test run started/,$!d'
+ cat /tmp/tmp.NBPEpn
+ set +x
failed Vulpix meta: the output of sbt differs from the expected output
expected : 2 |  a // Does not compile
actual   : [error]

This after changing a simple doc comment. It looks like the Vulpix meta test is flakey. We need to fix it or disable it. (This is not the first time it happened to me).

PR #4174, opened 2018-03-24
new line after rule heading

robstoll:
The list is broken on http://dotty.epfl.ch/docs. I suppose it requires a new line after the rule heading (at least the list in workflow.md has a new line and this one is displayed correctly)

### DIFF
--- a/docs/docs/reference/erased-terms.md
+++ b/docs/docs/reference/erased-terms.md
@@ -147,11 +147,10 @@ object Test {
 Rules
 -----
 
-1) The `erased` modifier can appear:
-
-* At the start of a parameter block of a method, function or class
-* In a method definition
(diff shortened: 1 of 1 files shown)


docs/docs/reference/erased-terms.md
```diff
@@ -164,11 +165,15 @@ def h(x: erased Int => Int) = ...
 class K(erased x: Int) { ... }
 ```
 
+
```
Blaisorblade: Is the extra line _before_ also required?

robstoll: Nope, its not but I figured it reads better in the raw version . However, the fix @allanrenucci provides is better IMO

allanrenucci:
Merging! Thanks @robstoll 🎉

PR #4178, opened 2018-03-25
fix indent so numbered list is continuous

robstoll:
Seems like the last commit #4174 fixed the bullet points but messed up the numbered list. Since I have no preview on how it will look like at  http://dotty.epfl.ch/docs/ it's again a guess.

### DIFF
--- a/docs/docs/reference/erased-terms.md
+++ b/docs/docs/reference/erased-terms.md
@@ -181,6 +181,7 @@ Rules
 
 
 4. Eta expansion
+
    if `def f(erased x: T): U` then `f: (erased T) => U`.
 
 
@@ -192,6 +193,7 @@ Rules
 
 
 6. Overloading
+
    Method with `erased` parameters will follow the normal overloading constraints after erasure.
 
 
--- a/docs/docs/reference/erased-terms.md
+++ b/docs/docs/reference/erased-terms.md
@@ -152,17 +152,17 @@ Rules
    * In a method definition
    * In a `val` definition (but not `lazy val` or `var`)
 
-```scala
-erased val x = ...
-erased def f = ...
+    ```scala
+    erased val x = ...
+    erased def f = ...
 
-def g(erased x: Int) = ...
+    def g(erased x: Int) = ...
 
-(erased x: Int) => ...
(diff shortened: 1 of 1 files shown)


smarter:
>  Since I have no preview on how it will look like at http://dotty.epfl.ch/docs/ it's again a guess.

run `sbt genDocs` in the project, output in `docs/_site/index.html`

docs/docs/reference/erased-terms.md
```diff
@@ -177,12 +177,11 @@ class K(erased x: Int) { ... }
...
    *  ...
 
-Note that there is no subtype relation between `erased T => R` and `T => R` (or `implicit erased T => R` and `implicit T => R`)
+   Note that there is no subtype relation between `erased T => R` and `T => R` (or `implicit erased T => R` and `implicit T => R`)
 
 
 4. Eta expansion
-
-if `def f(erased x: T): U` then `f: (erased T) => U`.
+   if `def f(erased x: T): U` then `f: (erased T) => U`.
```
allanrenucci: Will rendered on the same line as the previous line. So we might as well rewrite it as:
```
4. Eta expansion: if `def f(erased x: T): U` then `f: (erased T) => U`.
```

docs/docs/reference/erased-terms.md
```diff
@@ -193,8 +192,7 @@ if `def f(erased x: T): U` then `f: (erased T) => U`.
 
 
 6. Overloading
-
-Method with `erased` parameters will follow the normal overloading constraints after erasure.
+   Method with `erased` parameters will follow the normal overloading constraints after erasure.
```
allanrenucci: I would also put it on the previous line:
```
6. Overloading: Method with `erased` parameters will follow the normal overloading constraints after erasure.
```

docs/docs/reference/erased-terms.md
robstoll: Was not my intention, I'll revert this particular change

docs/docs/reference/erased-terms.md
robstoll: same here

PR #4179, opened 2018-03-25
fix parameter name in general example description

robstoll:
I suppose it should be `x1` and `xN` instead of `e1` and `eN` but maybe I misunderstand something because I don't see why only an upper approximation is possible.


Blaisorblade:
Merging this, can't investigate on the rest this week.

PR #4181, opened 2018-03-25
Fix #4180: Can't compare Scala and Java Booleans

### LINKED ISSUE #4180: Can't compare Scala and Java Booleans
```scala
object Bar {
  true == (true: java.lang.Boolean)
}
```
gives

```scala
-- Error: /tmp/scastie843852333271942738/src/main/scala/main.scala:3:2 ---------
3 |  true == (true: java.lang.Boolean)
  |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |  Values of types Boolean and Boolean cannot be compared with == or !=
```

smarter:


tests/pos/i4180.scala
```diff
@@ -0,0 +1,6 @@
+object Test {
+  def foo(s: Boolean, j: java.lang.Boolean) = {
+    val a = s == j
+    val b = j == s
+  }
+}
```
Blaisorblade: Just confirmed those comparisons give the expected result, though I guess this might not warrant a test:
```scala
$ /usr/local/bin/dotr
Starting dotty REPL...
scala> val v: java.lang.Boolean = true
val v: Boolean = true
scala> val w: Boolean = true
val w: Boolean = true
scala> v == w
1 |v == w
  |^^^^^^
  |Values of types Boolean and Boolean cannot be compared with == or !=
scala>  implicit def eqSBoolJBool: Eq[Boolean, java.lang.Boolean] = Eq
implicit val eqSBoolJBool: Eq[Boolean, Boolean]
scala> v == w
1 |v == w
  |^^^^^^
  |Values of types Boolean and Boolean cannot be compared with == or !=
scala> w == v
val res0: Boolean = true
scala> w == false
val res1: Boolean = false
scala> implicit def eqJBoolSBool: Eq[java.lang.Boolean, Boolean] = Eq
implicit val eqJBoolSBool: Eq[Boolean, Boolean]
scala> v == w
val res2: Boolean = true
scala> v == false
val res3: Boolean = false
```

PR #4187, opened 2018-03-26
Fix #4185: Do not check names for modules that will not be emitted

### LINKED ISSUE #4185: False "differs only in case" warning
Given
```scala
object ord {
  class Ord
  object Ord
}
```
I get:

Class Ord differs only in case from ord. Such classes will overwrite one another on case-insensitive filesystems.

nicolasstucki:


allanrenucci:
I think a better  and simpler solution would be to do the check when we generate the class file. At this point we only need to check if a class file was not emitted before
https://github.com/dotty-staging/dotty/blob/9aa2915b7bb251066bcc20209978b1d5851c15ca/compiler/src/dotty/tools/backend/jvm/GenBCode.scala#L419

Issue #4192, opened 2018-03-26
IncompatibleClassChangeError on inner class java.lang.Class


nicolasstucki:
```scala
object Test {
  def main(args: Array[String]): Unit = {
    class Foo
    classOf[Foo].getDeclaringClass
  }
}
```
fails with
```scala
Exception in thread "main" java.lang.IncompatibleClassChangeError: Test and Test$Foo$1 disagree on InnerClasses attribute
	at java.lang.Class.getDeclaringClass0(Native Method)
	at java.lang.Class.getDeclaringClass(Class.java:1235)
	at Test$.main(Foo.scala:4)
	at Test.main(Foo.scala)
```

It works on `scalac`.

farquet:
For reference, this issue is a blocker for generating GraalVM native images for code generated by the dotty compiler. GraalVM static analysis relies on those `InnerClasses` attributes to be correct to be able to work.

The scala compiler had a similar issue which got fixed in scala 2.12.7. scala/bug/issues/10487

prolativ:
Not sure if this should be treated as a separate issue or as a part of this one but while trying to construct a more complex test case for this I found out that there is a discrepancy between scala 2 and 3 about how class files are generated.
In this case
```scala
object Test {
  class Quux

  def main(args: Array[String]): Unit = {
    class Foo {
      class Bar

      println(classOf[Bar].getDeclaringClass)

      def test() = {
        class Baz
        println(classOf[Baz].getDeclaringClass)
      }
    }

    println(classOf[Foo].getDeclaringClass)
    val foo = new Foo
    foo.test()
  }
}
```
as the result of transformations done in `lambdaLift` and `flatten` phases `Baz` turns into `Test$Foo$1$Baz$1` in scala 2 but into `Test$Baz$1` in scala 3. Is this expected/acceptable? Is there a particular reason why this got changed? Taking into account that in both cases `Quux` becomes `Test$Quux` and `Foo` becomes `Test$Foo$1`, the scheme used in scala 2 seems more regular although it produces class files with longer names which might cause problems in some extreme cases (e.g. when one uses a Linux with an encrypted disk partition).
@smarter @nicolasstucki any ideas?

smarter:
>  Is this expected/acceptable? 

Yes, those are just implementation details.

>  it produces class files with longer names which might cause problems in some extreme cases (e.g. when one uses a Linux with an encrypted disk partition).

[...]

PR #4193, opened 2018-03-26
Fix #3947: Support lifting of java.lang.Class

### LINKED ISSUE #3947: Lift a Class[T] to an Expr[Class[T]]
We need to be able to lift class tags to be able to lift `Array`.  A `ClassTag` can be constructed with a `java.lang.Class`, hence we should be able to lift a `java.lang.Class`.

nicolasstucki:


nicolasstucki:
Missing a test case for lifted anonymous classes. These have an unrelated bug #4192.

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -33,7 +34,11 @@ object PickledQuotes {
   /** Transform the expression into its fully spliced Tree */
   def quotedExprToTree(expr: quoted.Expr[_])(implicit ctx: Context): Tree = expr match {
     case expr: TastyExpr[_] => unpickleExpr(expr)
-    case expr: LiftedExpr[_] => Literal(Constant(expr.value))
+    case expr: LiftedExpr[_] =>
+      expr.value match {
+        case value: Class[_] => ref(defn.Predef_classOf).appliedToType(classToType(value))
+        case value => Literal(Constant(value))
```
smarter: Did you know that you can pass a `Type` to the constructor of `Constant` and if you do `Constant#tpe` on such a value you get an instance of `java.lang.Class` ?

smarter: The miniphase `ClassOf` rewrites call to `classOf` into `Literal(Constant(...))`

nicolasstucki: I will look at those.

nicolasstucki: I can use the representation with `Literal(Constant(...))` from `ClassOf` though this representation is only expected after this phase. It seems better to create the `classOf[T]` and let `ClassOf` rewrite all of them at that phase.

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -154,4 +159,25 @@ object PickledQuotes {
...
+    if (clazz == classOf[Boolean]) defn.BooleanType
+    else if (clazz == classOf[Byte]) defn.ByteType
+    else if (clazz == classOf[Char]) defn.CharType
+    else if (clazz == classOf[Short]) defn.ShortType
+    else if (clazz == classOf[Int]) defn.IntType
+    else if (clazz == classOf[Long]) defn.LongType
+    else if (clazz == classOf[Float]) defn.FloatType
+    else if (clazz == classOf[Double]) defn.DoubleType
+    else if (clazz == classOf[Unit]) defn.UnitType
+    else if (!clazz.isMemberClass) ctx.getClassIfDefined(clazz.getCanonicalName).typeRef
```
retronym: `Array`-s probably need special handling, too.

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -154,4 +159,25 @@ object PickledQuotes {
...
+  private def classToType(clazz: Class[_])(implicit ctx: Context): Type = {
+    if (clazz == classOf[Boolean]) defn.BooleanType
+    else if (clazz == classOf[Byte]) defn.ByteType
+    else if (clazz == classOf[Char]) defn.CharType
+    else if (clazz == classOf[Short]) defn.ShortType
+    else if (clazz == classOf[Int]) defn.IntType
+    else if (clazz == classOf[Long]) defn.LongType
+    else if (clazz == classOf[Float]) defn.FloatType
+    else if (clazz == classOf[Double]) defn.DoubleType
+    else if (clazz == classOf[Unit]) defn.UnitType
```
retronym: Could enclose the above in `if (clazz.isPrimitive)` for performance reasons.

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
nicolasstucki: Will look into it

nicolasstucki: Yes, they needed it. Thanks :)

tests/run-with-compiler/i3947.scala
```diff
@@ -0,0 +1,24 @@
+
+import scala.quoted._
+import dotty.tools.dotc.quoted.Toolbox._
+
+object Test {
+
+  def main(args: Array[String]): Unit = {
+
+    def test[T](clazz: java.lang.Class[T]): Unit = {
+      val lclazz = clazz.toExpr
```
biboudis: What's the purpose of `lclazz` in the test?

tests/run-with-compiler/i3947b.scala
```diff
@@ -0,0 +1,29 @@
+
+import scala.quoted._
+import dotty.tools.dotc.quoted.Toolbox._
+
+object Test {
+
+  def main(args: Array[String]): Unit = {
+
+    def test[T](clazz: java.lang.Class[T]): Unit = {
+      val lclazz = clazz.toExpr
```
biboudis: Same.

tests/run-with-compiler/i3947.scala
nicolasstucki: Oh, I see. I lifted it directly in the quote. I will remove it or use it instead.

nicolasstucki: Done

tests/run-with-compiler/i3947b.scala
nicolasstucki: Done

nicolasstucki:
Rebased

PR #4199, opened 2018-03-27
IDE: Support jump to definition in imports

Duhemm:


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -251,18 +251,40 @@ class DottyLanguageServer extends LanguageServer
...
-      defs.map(d => location(d.namePos)).asJava
-    }
+    val defs =
+      if (sym == NoSymbol) Nil
+      else enclTree match {
+        case imp: Import =>
+          def lookup(name: Name): Symbol = {
+            imp.expr.tpe.member(name).symbol
+          }
+          val importedSyms = imp.selectors.flatMap {
```
allanrenucci: My guess would be that there is no more than one selector at the cursor position. So you should be able to do `imp.selectors.find(_.pos.contains(pos.pos))`.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -251,18 +251,40 @@ class DottyLanguageServer extends LanguageServer
...
+            case id: Ident if id.pos.contains(pos.pos) =>
+              lookup(id.name) :: lookup(id.name.toTypeName) :: Nil
```
allanrenucci: `id.name.toTermName`? Even if it's no-op, I think it is more explicit

compiler/src/dotty/tools/dotc/interactive/SourceTree.scala
```diff
@@ -10,7 +10,7 @@ import Contexts._, NameOps._, Symbols._
 import util._, util.Positions._
 
 /** A typechecked named `tree` coming from `source` */
-case class SourceTree(tree: tpd.NameTree, source: SourceFile) {
+case class SourceTree(topLevelImports: List[tpd.Import], tree: tpd.NameTree, source: SourceFile) {
```
allanrenucci: Shouldn't  an import be a source tree instead of attaching all top level import to every source tree?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -251,18 +251,40 @@ class DottyLanguageServer extends LanguageServer
...
+          val importedSyms = imp.selectors.find(_.pos.contains(pos.pos)) match {
+            case Some(id: Ident) =>
+              lookup(id.name.toTermName) :: lookup(id.name.toTypeName) :: Nil
+            case Some(Thicket((id: Ident) :: (_: Ident) :: Nil)) =>
+              lookup(id.name.toTermName) :: lookup(id.name.toTypeName) :: Nil
+            case _ =>
+              Nil
+          }
+
+          importedSyms.flatMap { sym =>
```
allanrenucci: `sym` can be `NoSymbol` (`def lookup` returns `NoSymbol` if the member is not found). I would filter them out here

compiler/src/dotty/tools/dotc/interactive/SourceTree.scala
smarter: Yes I think this makes more sense, instead of storing topLevelImports + TypeDef of the class, we should just store the root PackageDef filtered to only contain the TypeDef corresponding to the class.

Duhemm:
Rebased on latest master

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -269,38 +269,69 @@ object Interactive {
...
+            traverse(tree)
+            renameTree.foreach(traverse)
+          }
+
+          traverse(expr)
+        }
         override def traverse(tree: untpd.Tree)(implicit ctx: Context) = {
-          tree match {
+        tree match {
+            case imp @ Import(uexpr, (id: untpd.Ident) :: Nil) =>
```
smarter: Looks like this only handle lines with one import at a time ?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -269,38 +269,69 @@ object Interactive {
...
               traverseChildren(tree)
           }
         }
-      }.traverse(topTree)
+      }
+    }
+
+    trees foreach {
+      case SourceNamedTree(topTree, source) =>
+        traverser(source).traverse(topTree)
```
smarter: Looks like the two cases do the same thing, why not `trees.foreach(tree => traverser(source).traverse(tree.tree))` ?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -343,15 +353,22 @@ class DottyLanguageServer extends LanguageServer
...
+    val enclTree = Interactive.enclosingTree(driver.openedTrees(uri), pos)
 
-    if (sym == NoSymbol) Nil.asJava
-    else {
-      val refs = Interactive.namedTrees(uriTrees, Include.references | Include.overriding, sym)
-      ( for (ref <- refs if ref.namePos.exists)
-        yield new DocumentHighlight(range(ref.namePos), DocumentHighlightKind.Read)
-      ).asJava
+    val syms =
+      enclTree match {
```
smarter: Having various places where we special case what to do based on whether `enclTree` is an import or not seems fragile. Can we instead generalize `Interactive.enclosingSourceSymbol` to work with imports (and return a list of symbols instead of a single symbol) ?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -413,4 +446,43 @@ object Interactive {
...
+    }
+
+    symbols.filter(_.exists)
+  }
+
+  /** Used to represent a renaming import `{foo => bar}`.
+   *  We need this because the name of the tree must be the new name, but the
+   *  denotation must be that of the importee.
+   */
+  private case class RenameTree(name: Name, underlying: Tree) extends NameTree {
```
smarter: Do we really need to create a new AST node if the only thing we do on it is traverse it once and not store it anywhere afterwards?

Duhemm: The goal is to have a `NameTree` (we need a `NameTree` because that is what `namedTrees` is looking for), such that:
 - Its `denot` is the denotation of the thing that is actually imported, so that comparing symbols works as expected
 - Its `name` is the name of the symbol after renaming, so that `namePos` is correct.

The tree may be stored and returned by `namedTrees`.

Should I try to make the doc comment clearer, or do you have a better idea to achieve the same result?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -59,54 +60,45 @@ object Interactive {
    *
    * @see sourceSymbol
    */
```
smarter: Documentation needs to be updated

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -59,54 +60,45 @@ object Interactive {
...
+  def enclosingSourceSymbols(path: List[Tree], pos: SourcePosition)(implicit ctx: Context): List[Symbol] = {
+    val syms = path match {
       // For a named arg, find the target `DefDef` and jump to the param
       case NamedArg(name, _) :: Apply(fn, _) :: _ =>
         val funSym = fn.symbol
         if (funSym.name == StdNames.nme.copy
           && funSym.is(Synthetic)
           && funSym.owner.is(CaseClass)) {
-            funSym.owner.info.member(name).symbol
+            funSym.owner.info.member(name).symbol :: Nil
```
smarter: Style preference: I'd rather use List(...) than ... :: Nil

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -492,4 +511,44 @@ object Interactive {
...
+   * @param imp The import statement to analyze
+   * @param selectorPredicate A test to find the selector to use.
+   * @return The symbols imported.
+   */
+   private def importedSymbols(imp: tpd.Import,
+                       selectorPredicate: untpd.Tree => Boolean = util.common.alwaysTrue)
+                      (implicit ctx: Context): List[Symbol] = {
+     def lookup0(name: Name): Symbol = imp.expr.tpe.member(name).symbol
+     def lookup(name: Name): List[Symbol] = {
+       lookup0(name.toTermName) ::
```
smarter: I would use List(...) here too

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -492,4 +511,44 @@ object Interactive {
...
+
+     symbols.map(sourceSymbol).filter(_.exists).distinct
+   }
+
+  /**
+   * Used to represent a renaming import `{foo => bar}`.
+   * We need this because the name of the tree must be the new name, but the
+   * denotation must be that of the importee.
+   */
+   private case class RenameTree(name: Name, underlying: Tree) extends NameTree {
```
smarter: To be removed as discussed.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -311,7 +312,7 @@ class DottyLanguageServer extends LanguageServer
     val syms = Interactive.enclosingSourceSymbols(path, pos)
     val newName = params.getNewName
     val includes =
-      Include.references | Include.definitions | Include.linkedClass | Include.overriding
+      Include.references | Include.definitions | Include.linkedClass | Include.overriding | Include.imports
```
smarter: What if I'm trying to rename  a tree whose name comes from a renamed import ?

Duhemm: It'll behave the same as IntelliJ: The trees that use the new name won't be changed, but the ones that use the original symbol will. Renaming the renaming (renaming `baz` to something else in `import foo.{bar => baz}`) is not supported. There's a test for that here: https://github.com/lampepfl/dotty/pull/4199/files#diff-658e0b824b11a8c704adb2d5d7b05a38R91

But it would be better to display a warning too...

smarter: But what prevents us from renaming the renaming technically ?

Duhemm: Just adding ` | Include.renamingImports` here wouldn't work because then we'd rename imports into `import foo.{bar => bar}`. I'm not sure how much more complicated it'll get, I will look into it.

Duhemm:
@smarter I've added support for renaming renamed symbols. `RenameTree` is gone, too.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -645,7 +645,7 @@ object Interactive {
           case pkg: PackageDef if immediatelyEnclosesRenaming(toName, pkg) =>
             EmptyTree
           case template: Template if immediatelyEnclosesRenaming(toName, template) =>
-            cpy.Template(template)(constr = DefDef(template.constr.symbol.asTerm), self = EmptyValDef, body = Nil)
+            cpy.Template(template)(constr = DefDef(template.constr.symbol.asTerm), self = template.self, body = Nil)
```
smarter: This is equivalent to not pass a `self =` argument all, no?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -566,10 +566,132 @@ object Interactive {
...
+        case tree: NameTree =>
+          SourceNamedTree(tree, source)
+        case tree =>
+          val valDef = tpd.SyntheticValDef(NameKinds.UniqueName.fresh(), tree)
+          SourceNamedTree(valDef, source)
+      }
+    }
+
+    val includes =
+      Include.references | Include.imports | Include.renamingImports
```
smarter: Is there any situation where we have `Include.imports` but not `Include.renamingImports` ? Otherwise we should get rid of `renamingImports`

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -566,10 +566,132 @@ object Interactive {
...
+   * @return True if this tree immediately encloses an import that renames a symbol to `toName`,
+   *         false otherwise.
+   */
+  def immediatelyEnclosesRenaming(toName: Name, inTree: Tree)(implicit ctx: Context): Boolean = {
+    def isImportRenaming(tree: Tree): Boolean = {
+      tree match {
+        case Import(_, selectors) =>
+          selectors.exists {
+            case Thicket(_ :: Ident(rename) :: Nil) =>
+              rename.stripModuleClassSuffix.toString == toName.stripModuleClassSuffix.toString
```
smarter: could use sameName

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -566,10 +566,132 @@ object Interactive {
...
+          case block @ Block(stats, expr) if immediatelyEnclosesRenaming(toName, block) =>
+            EmptyTree
+          case other =>
+            super.transform(other)
+        }
+      }.transform(inTree)
+    }
+
+    val trees = {
+      val trees = enclosing match {
```
smarter: Having two vals with the same name like this is a bit confusing

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -490,4 +521,177 @@ object Interactive {
...
+   *         it uses a renaming introduced by an import statement.
+   */
+  def isRenamed(tree: NameTree)(implicit ctx: Context): Boolean = {
+    val symbol = tree.symbol
+    symbol.exists && !sameName(tree.name, symbol.name)
+  }
+
+  /** Are the two names the same? */
+  def sameName(n0: Name, n1: Name): Boolean = {
+    n0.stripModuleClassSuffix.toString == n1.stripModuleClassSuffix.toString
```
allanrenucci: Should be more efficient:
```scala
n0.stripModuleClassSuffix.toTermName eq n1.stripModuleClassSuffix.toTermName
```

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
Duhemm: Yes, this happens when we do `rename` on something that has not been renamed.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -566,10 +566,132 @@ object Interactive {
...
+   */
+  def findTreesMatchingRenaming(toName: Name,
+                                enclosing: Option[Tree],
+                                syms: List[Symbol],
+                                allTrees: List[Tree],
+                                source: SourceFile
+                               )(implicit ctx: Context): List[SourceNamedTree] = {
+
+    /**
+     * Remove the blocks that immediately enclose a renaming to `toName` in `inTree`.
```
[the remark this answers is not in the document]
Duhemm: I've added a commit that removes this

Duhemm:
@smarter Can you take another look? I think I addressed all of your comments.

smarter:
> lso I'm not sure if there was a follow-up to #4199 (comment).

Sorry, wrong comment, I was referring to https://github.com/lampepfl/dotty/pull/4199#discussion_r229319898 where I forgot to ask whether something would go wrong if we just merged imports and renamedImports into one flag

Duhemm:
The two flags were equivalent, but I managed to get rid of it by adding an additional predicate to `findTreesMatching`.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -332,16 +349,17 @@ class DottyLanguageServer extends LanguageServer
...
+    syms.flatMap { sym =>
+      val refs = Interactive.findTreesMatching(uriTrees, includes, sym)
       (for {
-        ref <- refs if !ref.tree.symbol.isPrimaryConstructor
+        ref <- refs
         nameRange <- range(ref.namePos, positionMapperFor(ref.source))
-      } yield new DocumentHighlight(nameRange, DocumentHighlightKind.Read)).asJava
-    }
+      } yield new DocumentHighlight(nameRange, DocumentHighlightKind.Read))
+    }.distinct.asJava
```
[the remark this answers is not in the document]
Duhemm: In the UI, it makes no difference, but it does when we test. Consider this snippet:
```scala
object O { def foo = 0; type foo = Int }
import O.foo
```

If you highlight `foo` in `O.foo`, you'll get two symbols to match: `def foo` and `type foo`. When inspecting `import O.foo`, we'll create the two select nodes `O.foo` (term) and `O.foo` (type). Both nodes have the same position and both will match. I don't know how to completely avoid the duplicates in this particular situation and be able to remove the `.distinct`.

I've improved how we construct those select nodes to make sure that we get fewer duplicates, which should be slightly more performant.

compiler/src/dotty/tools/dotc/interactive/SourceTree.scala
```diff
@@ -9,10 +9,24 @@ import core._, core.Decorators.{sourcePos => _}
...
+  /** The source from which `tree` comes. */
+  def source: SourceFile
+
   /** The position of `tree` */
-  def pos(implicit ctx: Context): SourcePosition = source.atPos(tree.pos)
+  final def pos(implicit ctx: Context): SourcePosition = source.atPos(tree.pos)
+}
+
+/** An import coming from `source` */
+case class SourceImportTree(tree: tpd.Import, source: SourceFile) extends SourceTree
```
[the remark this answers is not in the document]
Duhemm: I've replaced it with
```scala
case class SourceTree(tree: tpd.Tree /** really: tpd.Import | tpd.NameTree */, ...)
```

There are already a few instances of this pattern in the compiler.

I'm happy to replace it with a marker trait if you prefer. Can you suggest a name for this trait?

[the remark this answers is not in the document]
smarter: good enough for me :)

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -302,32 +300,64 @@ object Interactive {
...
    *  @param includeReferences  If true, include references and not just definitions
    */
-  def namedTrees(trees: List[SourceTree], includeReferences: Boolean, treePredicate: NameTree => Boolean)
-    (implicit ctx: Context): List[SourceTree] = safely {
-    val buf = new mutable.ListBuffer[SourceTree]
-
-    trees foreach { case SourceTree(topTree, source) =>
+  def namedTrees(trees: List[SourceTree], include: Include.Set, treePredicate: NameTree => Boolean)
+    (implicit ctx: Context): List[SourceNamedTree] = safely {
+    val includeReferences = (include & Include.references) != 0
```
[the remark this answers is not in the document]
Duhemm: I followed your suggestion in the last commit

Duhemm:
@smarter This is ready for another round!

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -316,19 +316,25 @@ class DottyLanguageServer extends LanguageServer
       path match {
         // Selected a renaming in an import node
         case Thicket(_ :: (rename: Ident) :: Nil) :: (_: Import) :: rest if rename.pos.contains(pos.pos) =>
-          Interactive.findTreesMatchingRenaming(rename.name, syms, uriTrees)
+          val includes = Include.references | Include.linkedClass | Include.imports
```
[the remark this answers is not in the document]
Duhemm: Yes, I think using `Include.all` would be fine here 👍

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -541,4 +555,54 @@ object Interactive {
     }
   }
 
+  /**
+   * Is this tree using a renaming introduced by an import statement?
+   *
+   * @param tree The tree to inspect
+   * @return True, if this tree's name is different than its symbol's name, indicating that
+   *         it uses a renaming introduced by an import statement.
```
smarter: Aren't there cases where a tree's name is different from the symbol's name even without renaming ? e.g. I think when you do:
```scala
class A { self =>
 val x = self
 val y = this
}
```
Then the symbol of `self` is the same as the symbol of `this`.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -541,4 +555,54 @@ object Interactive {
...
+
+  /**
+   * Is this tree immediately enclosing an import that renames a symbol to `toName`?
+   *
+   * @param toName        The target name to check
+   * @param tree          The tree to check
+   * @return True if this tree immediately encloses an import that renames a symbol to `toName`,
+   *         false otherwise.
+   */
+  def immediatelyEnclosesRenaming(toName: Name, inTree: Tree)(implicit ctx: Context): Boolean = {
```
smarter: Looks like this method is dead code

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -1147,5 +1147,89 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
+        }
+      case Thicket((id: untpd.Ident) :: (newName: untpd.Ident) :: Nil) =>
+        importedSymbols(imp.expr, id.name).flatMap { sym =>
+          imported(sym, id, Some(newName))
+        }
+    }
+  }
+
+  /** Replaces all positions in `tree` with zero-extent positions */
+  private def focusPositions(tree: Tree)(implicit ctx: Context): Tree = {
```
smarter: Why is this needed ?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -374,16 +413,17 @@ class DottyLanguageServer extends LanguageServer
 
     val pos = sourcePosition(driver, uri, params.getPosition)
     val uriTrees = driver.openedTrees(uri)
-    val sym = Interactive.enclosingSourceSymbol(uriTrees, pos)
+    val path = Interactive.pathTo(uriTrees, pos)
+    val syms = Interactive.enclosingSourceSymbols(path, pos)
+    val includes = Include.definitions | Include.references | Include.imports
```
smarter: I think including overriding/overriden symbols in the highlight would be useful too, so this could probably be Include.all

compiler/src/dotty/tools/dotc/ast/tpd.scala
Duhemm: This is to prevent the qualifier of the `Select` nodes that we generate from being added to the results and avoid duplicates. The original qualifier (with the real position) will already be considered in `namedTrees` when we handle the `Import` nodes.

I added a comment in `def importSelections`.

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
Duhemm: Good idea, but I excluded the `linkedClass`.

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
Duhemm: Good catch. It looks like the definition of `self` has a symbol of its own, but references to `self` have the same symbol as `this` (that is, the underlying class).

I fixed the doc in `def isRenamed` to mention that and will open an issue about renaming self-types.

PR #4202, opened 2018-03-27
Do not set/reset flags that are part of FromStartFlags/AfterLoadFlags

smarter:


smarter:
@sjrd Do you remember the context around https://github.com/lampepfl/dotty/pull/1126/commits/0ebf36b394b00f3f432d8fdedeaa15e7d4df2b06 ? It's problematic because Mutable is currently a FromStartFlags so it should never be set/reset after creation. I assume this was used in the Scala.js backend but could be replaced by a more indirect check like `if (sym.isSetter and sym.is(Scala2x))`.

smarter:
Ah, but https://github.com/lampepfl/dotty/commit/0ebf36b394b00f3f432d8fdedeaa15e7d4df2b06 happens in a treetransformer, so whether or not Mutable is FromStartFlags does not make a difference here I think.

Here are the remaining flag mutations I believe are problematic but didn't know how to fix:
- [ ] https://github.com/lampepfl/dotty/blob/16ce7e46672857a07d46422cd6a1c33aa8dff5f9/compiler/src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala#L115
- [ ] https://github.com/lampepfl/dotty/blob/16ce7e46672857a07d46422cd6a1c33aa8dff5f9/compiler/src/dotty/tools/dotc/typer/Namer.scala#L754
- [ ] https://github.com/lampepfl/dotty/blob/16ce7e46672857a07d46422cd6a1c33aa8dff5f9/compiler/src/dotty/tools/dotc/typer/Namer.scala#L987


The following are also problematic but are only used to do error recovery so maybe ok:
- [ ] https://github.com/lampepfl/dotty/blob/16ce7e46672857a07d46422cd6a1c33aa8dff5f9/compiler/src/dotty/tools/dotc/typer/Namer.scala#L777
- [ ] https://github.com/lampepfl/dotty/blob/16ce7e46672857a07d46422cd6a1c33aa8dff5f9/compiler/src/dotty/tools/dotc/typer/Checking.scala#L387

sjrd:
> Do you remember the context around 0ebf36b ?

Yes, the context was indeed to simplify the work of the Scala.js back-end, because in the IR it is absolutely forbidden to mutate a `val` outside of a constructor. Note that it is also forbidden in Java 9's bytecode format, so you probably don't want to regress about this. scalac has somewhere an issue where they're trying to fix this going forward, I think.

Using an indirect test would be fine too, as long as it's reliable. But then you'll need to tell -Ycheck that mutating a mixed-in `val` from a Scala2 trait is OK too (or at least, I *hope* you need to tell that to -Ycheck, otherwise it would mean that -Ycheck is letting through invalid assignments).

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -454,17 +454,18 @@ object Flags {
   /** Flags representing access rights */
   final val AccessFlags = Private | Protected | Local
 
-  /** Flags guaranteed to be set upon symbol creation */
+  /** Flags that are not (re)set when completing the denotation */
```
odersky: I think both the old and the new comment are useful. Can we combine them?

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -454,17 +454,18 @@ object Flags {
...
     Module | Package | Deferred | MethodOrHKCommon | Param | ParamAccessor.toCommonFlags |
     Scala2ExistentialCommon | Mutable.toCommonFlags | Touched | JavaStatic |
     CovariantOrOuter | ContravariantOrLabel | CaseAccessor.toCommonFlags |
     NonMember | Erroneous | ImplicitCommon | Permanent | Synthetic |
     SuperAccessorOrScala2x | Inline
 
-  /** Flags guaranteed to be set upon symbol creation, or, if symbol is a top-level
-   *  class or object, when the class file defining the symbol is loaded (which
-   *  is generally before the symbol is completed
+  /** Flags that are not (re)set when completing the denotation, or, if symbol is
```
odersky: Same here

compiler/src/dotty/tools/dotc/core/Flags.scala
smarter: I think the old comment is misleading. These flags are not guaranteed to be set up on symbol creation, as can be witnessed by them appearing in setFlag/resetFlag calls in various places. I think that correctness only requires them to not appear in completers and InfoTransformers, but they're free to appear in other places such as TreeTransformers (e.g, the `setFlag(Mutable)` in Memoize I mentioned above).

smarter:
I've opened https://github.com/lampepfl/dotty/issues/4221 to follow-up on the remaining issues.

PR #4210, opened 2018-03-29
Optimize `'(~x)` to `x` to avoid pickling

nicolasstucki:


biboudis:
I wonder if the output of `quote-nested-5.scala` is right under this PR or not.
Under this PR the result of 
```
    val q = '{
      val a = '(4)
      ~('{
        '(~a)
      })

    }
```
is `a` and not `'(~a)` which is correct, no?

nicolasstucki:
Yes, it should be `a`.

nicolasstucki:
Rebased

PR #4213, opened 2018-03-29
Pretty print quotes

nicolasstucki:


compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -51,4 +51,16 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
       }
     }
   }
+
+  override protected def toTextCore[T >: Untyped](tree: Tree[T]): Text = {
```
allanrenucci: Why is not part of `RefinedPrinter`?

nicolasstucki: No, we do not want this representation when debugging.

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -51,4 +51,16 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
...
+    import untpd.{modsDeco => _, _}
+    tree match {
+      case TypeApply(fun, args) =>
```
allanrenucci: I think we should have a protected `def` for each type of tree and override the bits we are interested in.,

nicolasstucki: Added it

PR #4215, opened 2018-03-29
Add plugin support to sbt-dotty

liufengyun:
Add plugin support to sbt-dotty


sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/test
```diff
@@ -1,5 +1,4 @@
 > plugin/publishLocal
+> clean
+> compile
```
allanrenucci: I have been playing around with this and the plugin doesn't seems to be applied. I changed the plugin to emit errors instead of warnings. Then this compile should fail but it does not

liufengyun: Yes, exactly, thanks for looking into it. I was debugging the whole day,  sbt somehow ignored the setting. If I pass `-Xplugin` manually, it works without problem.

liufengyun:
SBT does check `scalac-plugin.xml`, while we are using `plugin.properties`. I'll try if we can find some  walkaround in `sbt-dotty`:


https://github.com/sbt/sbt/blob/v0.13.15/util/classpath/src/main/scala/sbt/classpath/ClasspathUtilities.scala#L100-L105

```Scala
  /** Returns all entries in 'classpath' that correspond to a compiler plugin.*/
  private[sbt] def compilerPlugins(classpath: Seq[File]): Iterable[File] =
    {
      import collection.JavaConversions._
      val loader = new URLClassLoader(Path.toURLs(classpath))
      loader.getResources("scalac-plugin.xml").toList.flatMap(asFile(true))
    }
```

https://github.com/sbt/sbt/blob/v0.13.15/main/src/main/scala/sbt/Defaults.scala#L1856-L1863

sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala
```diff
@@ -167,6 +201,15 @@ object DottyPlugin extends AutoPlugin {
...
 
+      scalacOptions := {
+        val options = scalacOptions.value
+        if (autoCompilerPlugins.value) {
+          val newPlugins = autoPlugins(update.value, internalPluginClasspath.value.files)
+          val existing = options.toSet
+          options ++ newPlugins.filterNot(existing)
+        } else options
+      },
+
```
liufengyun: @smarter : could you please advise if it's the right thing to do here?

smarter: The right thing to do is to fix sbt upstream to look for plugin.properties. Meanwhile, we can use whatever hack happens to work.

liufengyun: Thanks a lot, I'll make an upstream PR.

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/build.sbt
```diff
@@ -7,13 +7,14 @@ lazy val pluginSetting = Seq(
...
   )
 )
 
 lazy val plugin = (project in file("plugin")).settings(pluginSetting: _*)
 
-lazy val app = (project in file(".")).settings(
-  scalaVersion := dottyVersion,
-  libraryDependencies += compilerPlugin("ch.epfl.lamp" %% "dividezero" % "0.0.1")
-)
+scalaVersion := dottyVersion
```
smarter: There's already a scalaVersion set above.

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/test
```diff
@@ -0,0 +1,4 @@
+> plugin/publishLocal
```
smarter: Do we have tests that actually run the compiler with a compiler plugin enabled? It'd be good to have something like that in this example project and run them with `> test` here.

sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala
```diff
@@ -141,6 +143,38 @@ object DottyPlugin extends AutoPlugin {
   }
 
   override def projectSettings: Seq[Setting[_]] = {
+    // get plugins
```
smarter: Add comments explaining that this is temporary until we fix sbt, with link to the relevant sbt issue.

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/test
liufengyun: This project depends on the plugin, thus it's used in the compilation. The plugin will rewrite divide-zero in source code, so that there will be no run-time errors.

Previously, this test (like all sbt scripted tests) only run on nightly. But now I've added this particular one to normal CI.

sbt-dotty/sbt-test/sbt-dotty/compiler-plugin/build.sbt
liufengyun: The `scalaVersion` above is for a different sub-project.

smarter: oh I see yes, I suggest explicitly using `lazy val root = (project in file("."))` when there's more than one project, I find it less confusing.

liufengyun: Yes, good advice, that's less confusing.

PR #4220, opened 2018-03-29
Disable sbt scripted test for compiler plugin

allanrenucci:
Until we find a fix


allanrenucci:
Merging CI failures are unrelated

PR #4229, opened 2018-04-02
Emit feature warnings for implicit conversions.

odersky:
We now emit feature warnings when encountering an implicit conversion def,
just like Scala 2 does.

In addition, we also emit a feature warning when an implicit conversion is used,
unless the conversion is an implicit class, or otherwise co-defined with the type
to which it converts, or the conversion is predefined.

Predefined are

  - all members of `Predef`,
  - the `scala.reflect.Selectable.reflectiveSelect` conversion.

We might need to extend this to more conversions.

The motivation for going down this route is the following: Implicit conversions are easily the
most mis-used feature in Scala. If I should venture a guess I'd say that more than 90% of the
uses are mis-guided. The danger of implicit conversions is that they seem simple. It's easy
to grok what they are and in particular new users see the possibilities before realizing the
downsides. So far we put definitions of implicit conversions under a feature flag but this is
not enough. Library writers tend not to be deterred by it, they just add the language import.

(Or, worse, IntelliJ does it for them. Dear IntelliJ Scala team: please change things so that
language imports are not inserted automatically, They are required for a reason, and automatic
insertion does not help!)

Either way, it's the users of a library with implicit conversions that need to be warned
by the compiler, since it is them who will run into troubles if implicit conversions have
surprising behavior.

In what concerns the precise rules, I believe we will have to try this out in the wild for a
while to see how it shapes up and whether we should tweak the rules.


compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -576,6 +576,49 @@ trait Checking {
...
+   *    - it is defined in Predef
+   *    - it is the scala.reflect.Selectable.reflectiveSelectable conversion
+   */
+  def checkImplicitConversionUseOK(sym: Symbol, pos: Position)(implicit ctx: Context): Unit = {
+    val conversionOK =
+      !sym.exists ||
+      sym.is(Synthetic) ||
+      sym.info.finalResultType.classSymbols.exists(_.owner.isLinkedWith(sym.owner)) ||
+      defn.isPredefClass(sym.owner) ||
+      sym.name == nme.reflectiveSelectable && sym.maybeOwner.maybeOwner.maybeOwner == defn.ScalaPackageClass
```
allanrenucci: Why not compare `symbol` here?

Glavo:
I don't think there is no need to warn the user when an implicit conversion function is imported separately. In this case, the user clearly expresses that he needs to use implicit conversion..

PR #4235, opened 2018-04-03
Cache classloaders in Splicer

nicolasstucki:


nicolasstucki:
test performance please

nicolasstucki:
While inlining 1000 times `power(1000, 5.0)` from [power-macro/Macro1.scala](https://github.com/lampepfl/dotty/blob/master/tests/pos/power-macro/Macro_1.scala), the `ReifyQuotes` phase became 2X faster.

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/Splicer.scala
```diff
@@ -93,6 +93,8 @@ object Splicer {
     }
   }
 
+  private val classLoaders = scala.collection.mutable.HashMap.empty[String, URLClassLoader]
+
```
liufengyun: It means the cache is going to be shared by all compiler runs?  I'm not sure about the following:

- concurrency
- memory leak
- setting changes between runs

I'm not familiar with these areas, could you explain?

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/Splicer.scala
nicolasstucki: I changed the `Splicer` to a class and instantiate one per run to avoid the issues above.

compiler/src/dotty/tools/dotc/transform/Splicer.scala
```diff
@@ -21,9 +21,11 @@ import dotty.tools.dotc.util.Positions.Position
 import scala.reflect.ClassTag
 
 /** Utility class to splice quoted expressions */
-object Splicer {
+class Splicer {
```
smarter: It's weird to make the whole thing a class if the only bit of state is the classloader cache used in a single method, the cache could be passed as an argument instead, and the classloader cache created

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -88,7 +88,8 @@ import dotty.tools.dotc.core.quoted._
 class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
   import ast.tpd._
 
-  private lazy val splicer: Splicer = new Splicer
+  /** Classloader used for loading macros */
+  private var macroClassLoader: java.lang.ClassLoader = _
```
smarter: `private[this]` also usually we put the field and the getter next to each other, and we call the field `myXXX` and the getter `XXX`.

nicolasstucki:
test performance please

nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/Splicer.scala
retronym: It is not to share `s.c.m.HashMap` between threads without synchronization.

I'm thinking about using a timed cache in Scalac to deal with the leak problem (keep the loader around for X seconds after last use).

retronym: The other issue with caching classloaders is having a good plan for invalidating them when JARs and directories on disk change.

nicolasstucki: Now the cache is only valid during one phase of one run of the compiler. None of these should be an issue.

PR #4236, opened 2018-04-03
Fix #4226: allow infallible boolean extractors to return `true`

### LINKED ISSUE #4226: Allow infallible boolean extractors to return `true` type
The following code ([here in Scastie](https://scastie.scala-lang.org/llbo0DppTvu01lceEOW0DA)) does not currently work:

```scala
sealed abstract class Maybe[A]
final case class Just[A](a: A) extends Maybe[A]
class Empty[A] extends Maybe[A]
object Empty {
  def apply[A](): Maybe[A] = new Empty[A]
  // def unapply[A](e: Empty[A]): Boolean = true  // ok
  def unapply[A](e: Empty[A]): true = true
}

object Test {
  val a: Maybe[Int] = Just(2)
  def main(args: Array[String]): Unit = a match {
    case Just(2) => true
    case Empty() =>
    // ^ Boolean(true) is not a valid result type of an unapply method of an extractor()
  }
}
```

Though it should arguably compile and raise an exhaustivity warning (because it does not match `Just(n)` for `n != 2`).

liufengyun:
Fix #4226: allow infallible boolean extractors to return `true`


tests/pos/i4226.scala
```diff
@@ -0,0 +1,15 @@
+sealed abstract class Maybe[A]
```
smarter: Is it really useful to have the same test in pos and patmat? Aren't the patmat tests also fully compiled?

liufengyun: OK, let's remove it. I was thinking more in terms of features.

PR #4238, opened 2018-04-03
Change owners only once for nested unpickled quotes

nicolasstucki:
Based on #4235


nicolasstucki:
test performance please

nicolasstucki:
Yes, the current solution will break if there are nested definitions inside the quote. I will try accumulating the correct owner changes in a map and then apply them all at the end.

nicolasstucki:
A simpler and more performant fix is available in #4242

PR #4242, opened 2018-04-04
Unpickle quotes with the destination context

nicolasstucki:
This avoids exponential traversal of the tree due to owner changes
of the dummy valdefs for the contents of quote.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
+++ b/compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
@@ -59,7 +59,10 @@ object PickledQuotes {
     val unpickled = unpickle(tastyBytes, expr.args)
     unpickled match {
       case PackageDef(_, (vdef: ValDef) :: Nil) =>
-        vdef.rhs.changeOwner(vdef.symbol, ctx.owner)
+        // vdef.rhs was unpickled with the current context. Symbols in vdef.rhs with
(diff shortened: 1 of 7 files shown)


nicolasstucki:
Alternative fix for issue tackled in #4238. This version does not need any owner change after unpickling.

nicolasstucki:
test performance please

nicolasstucki:

nicolasstucki:
@smarter don't forget to review this old PR

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -59,7 +59,10 @@ object PickledQuotes {
     val unpickled = unpickle(tastyBytes, expr.args)
     unpickled match {
       case PackageDef(_, (vdef: ValDef) :: Nil) =>
-        vdef.rhs.changeOwner(vdef.symbol, ctx.owner)
+        // vdef.rhs was unpickled with the current context. Symbols in vdef.rhs with
+        // owner vdef.symbol are unpickled as if their owner was ctx.owner. This optimization
```
smarter: Would be good to give the name of the method that does this in TreeUnpickler.

nicolasstucki:
I will explore the possibility of pickling the tree directly in tasty.

nicolasstucki:
As suggested by @smarter we will instead pickle the expression directly in the tasty (no encapsulation). See #4538.

PR #4243, opened 2018-04-04
Remove substDealias

smarter:
Does not seem to be needed anymore.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Substituters.scala
+++ b/compiler/src/dotty/tools/dotc/core/Substituters.scala
@@ -73,35 +73,6 @@ trait Substituters { this: Context =>
     }
   }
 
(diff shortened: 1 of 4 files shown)


smarter:
test performance please

odersky:
I am not sure. SubstDealias was needed when we modeled parameters as refinements. But inprinciple the same configurations could arise through explicit refinements. How do we know 
this cannot lead to problems? I am worried about scenarios like

    [foo := bar]  P { type T = foo } # T

smarter:
Hmm, but then just dealiasing does not seem to be enough to handle all cases, e.g. what about:
```scala
val x: P { type T = foo }
type S <: x.T
[foo := bar] S
```

odersky:
@smarter Good point. Can we construct a test case out of this, or is there something else that means we don't substitute in situations like this? If we can't construct a test case I agree we should drop substDealias anyway in order to fail faster.

PR #4244, opened 2018-04-04
Cache definitions in quote driver

nicolasstucki:
This showed a 10X on warm runs of the test file.

### DIFF
--- a/compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
+++ b/compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
@@ -2,7 +2,7 @@ package dotty.tools.dotc.quoted
 
 import dotty.tools.dotc.ast.tpd
 import dotty.tools.dotc.Driver
-import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Contexts.{Context, ContextBase}
 import dotty.tools.io.{AbstractFile, Directory, PlainDirectory, VirtualDirectory}
 import dotty.tools.repl.AbstractFileClassLoader
 import dotty.tools.dotc.printing.DecompilerPrinter
@@ -11,7 +11,8 @@ import scala.quoted.{Expr, Type}
 
 import java.net.URLClassLoader
 
-import Toolbox.{Settings, Run, Show}
+import Toolbox.{Run, Settings, Show}
+import dotty.tools.dotc.core.Definitions
 
 class QuoteDriver extends Driver {
   import tpd._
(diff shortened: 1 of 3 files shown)


liufengyun:
test performance with quotes please

liufengyun:

allanrenucci:
Why revert the last commit? The CI failures look easely fixable: https://github.com/lampepfl/dotty/blob/91de02962767cef1f6221e2bd565367aca1f3efc/compiler/src/dotty/tools/dotc/quoted/Toolbox.scala#L42-L44
Do `ctx.fresh` and set the setting only when needed

nicolasstucki:
The `ctx.fresh` and removing the `ctx.settings.color.update("never")` have no efffect. If I share the full `ContextBase` there is some state that is kept. The `settings` are permanent in `ContextBase`.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -62,8 +62,12 @@ class QuoteDriver extends Driver {
     output.getOrElse(throw new Exception("Could not extract " + expr))
   }
 
+  private[this] val myDefinitions = new ThreadLocal[Definitions] {
```
allanrenucci: Any reasons not to us a synchronised value shared between the threads?

nicolasstucki: To be able to run several compiler instances in parallel. Different threads might try to initialize lazy val in Definitions at the same time.

smarter: I'm not convinced at all this is safe. `Definitions` contains state like the `ctx` field and the `_isInitialized` field, why is it correct to share this between every quote in the same JVM thread? Note that you may have more than one dotty compiler running on the same JVM thread.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -11,7 +11,8 @@ import scala.quoted.{Expr, Type}
 
 import java.net.URLClassLoader
 
-import Toolbox.{Settings, Run, Show}
+import Toolbox.{Run, Settings, Show}
+import dotty.tools.dotc.core.Definitions
 
 class QuoteDriver extends Driver {
```
smarter: Since you're always using QuoteDriver from an enclosing compiler with its own context, why not get the definitions from the enclosing ctx.definitions?

nicolasstucki: Enclosing compiler?

smarter: Quotes are expanded by some compiler, this compiler has a context with definitions.

Issue #4241, opened 2018-04-04
Compiler generated implicit conversion between function literal and PartialFunction


allanrenucci:
Scalac and Dotty both implicitly convert from function literals to `PartialFunction`. For example, one can write:
```scala
val foo: PartialFunction[Int, String] = x => x.toString
```
However this conversion is different in Dotty and Scalac. The compilers need to generate an implementation for `def isDefinedAt(x: A): Boolean`. Both compilers inspect the rhs of the function literal:
- Dotty: if the rhs is a `match`, implements `isDefinedAt` in term of the `case`s, otherwise returns `true`
- Scalac: if the rhs is a `match`, implements `isDefinedAt` in term of the `case`s, otherwise fails to compile

This leads to surprising behaviors:
```scala
val a: PartialFunction[Int, String] = x => x match { case 1 => x.toString }
a.isDefinedAt(2) // Dotty: false, Scalac: false

val b: PartialFunction[Int, String] = x => { x match { case 1 => x.toString } }
b.isDefinedAt(2) // Dotty: true, Scalac: false

val c: PartialFunction[Int, String] = x => { println("foo"); x match { case 1 => x.toString } }
c.isDefinedAt(2) // Dotty: true
// Scalac:
//   error: type mismatch;
//     found   : Int => String
//     required: PartialFunction[Int,String]
```
I would argue that a function literal is a partial function that is always defined and both Dotty and Scalac are wrong.

**Note**: It works like this in Dotty because the compiler desugars partial function literals into function literals:
```scala
{ case 1 => x.toString } // becomes x => x match { case 1 => x.toString }
```
Then when the compiler generates the `isDefined` method, it doesn't differentiate between a tree that was originally a partial function literal and one that was a function literal

Blaisorblade:
1. Dotty's clearly wrong (incompatible with Scalac) here, and in a way that could actually cause breakage for code that is pretty simple to write.

2. [The 2.12 spec (Sec. 6.23)](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#anonymous-functions) mandates that `x => x match { … }` can be treated as a partial function. Spec changes would have to go eventually through the usual review. I'd say Dotty-only spec. changes on small details are mostly technical debt — we should just fix Dotty now.

[...]

PR #4245, opened 2018-04-04
Fix #4177: Generate optimised applyOrElse implementation for partial function literals

### LINKED ISSUE #4177: Partial Function: Prevent double evaluation of pattern matchers and guards
The documentation of `PartialFunction::applyOrElse` says:
> For all partial function literals the compiler generates an `applyOrElse` implementation which avoids double evaluation of pattern matchers and guards.

However this is not currently the case in Dotty:
```scala
object Test {
  private[this] var count = 0

  def test(x: Int) = { count += 1; true }

  object Foo {
    def unapply(x: Int): Option[Int] = { count += 1; Some(x) }
  }

    val res = List(1, 2).collect { case x if test(x) => x }
    println(count) // prints 4 but should be 2

    count = 0
    val res2 = List(1, 2).collect { case Foo(x) => x }
    println(count) // prints 4 but should be 2
  }
}
```

allanrenucci:
Fix #4177.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/tpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/tpd.scala
@@ -286,8 +286,12 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
         coord = fns.map(_.pos).reduceLeft(_ union _))
     val constr = ctx.newConstructor(cls, Synthetic, Nil, Nil).entered
     def forwarder(fn: TermSymbol, name: TermName) = {
-      val fwdMeth = fn.copy(cls, name, Synthetic | Method).entered.asTerm
-      DefDef(fwdMeth, prefss => ref(fn).appliedToArgss(prefss))
+      var flags = Synthetic | Method
+      def isOverriden(denot: SingleDenotation) = fn.info.overrides(denot.info, matchLoosely = true)
+      val isOverride = parents.exists(_.member(name).hasAltWith(isOverriden))
+      if (isOverride) flags = flags | Override
+      val fwdMeth = fn.copy(cls, name, flags).entered.asTerm
+      polyDefDef(fwdMeth, tprefs => prefss => ref(fn).appliedToTypes(tprefs).appliedToArgss(prefss))
     }
     val forwarders = (fns, methNames).zipped.map(forwarder)
     val cdef = ClassDef(cls, DefDef(constr), forwarders)
--- a/compiler/src/dotty/tools/dotc/ast/tpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/tpd.scala
@@ -282,11 +282,11 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
     val parents1 =
       if (parents.head.classSymbol.is(Trait)) parents.head.parents.head :: parents
       else parents
-    val cls = ctx.newNormalizedClassSymbol(owner, tpnme.ANON_CLASS, Synthetic, parents1,
+    val cls = ctx.newNormalizedClassSymbol(owner, tpnme.ANON_CLASS, Synthetic | Final, parents1,
         coord = fns.map(_.pos).reduceLeft(_ union _))
     val constr = ctx.newConstructor(cls, Synthetic, Nil, Nil).entered
     def forwarder(fn: TermSymbol, name: TermName) = {
-      var flags = Synthetic | Method
+      var flags = Synthetic | Method | Final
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -585,6 +585,10 @@ class Definitions {
 
   lazy val PartialFunctionType: TypeRef         = ctx.requiredClassRef("scala.PartialFunction")
   def PartialFunctionClass(implicit ctx: Context) = PartialFunctionType.symbol.asClass
+
+    lazy val PartialFunction_applyOrElseR = PartialFunctionClass.requiredMethodRef(nme.applyOrElse)
```
Blaisorblade: This has the wrong indent size?

compiler/src/dotty/tools/dotc/transform/ExpandSAMs.scala
```diff
@@ -54,38 +54,70 @@ class ExpandSAMs extends MiniPhase {
     val Block(
           (applyDef @ DefDef(nme.ANON_FUN, Nil, List(List(param)), _, _)) :: Nil,
           Closure(_, _, tpt)) = tree
-    val applyRhs: Tree = applyDef.rhs
+    val applyRhs = applyDef.rhs
     val applyFn = applyDef.symbol.asTerm
 
-    val MethodTpe(paramNames, paramTypes, _) = applyFn.info
+  val MethodTpe(paramNames, paramTypes, _) = applyFn.info
```
Blaisorblade: indent issue here too?

compiler/src/dotty/tools/dotc/core/Definitions.scala
allanrenucci: This is the indentation scheme used across the file

allanrenucci:
test performance please

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -273,8 +273,11 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
         coord = fns.map(_.pos).reduceLeft(_ union _))
     val constr = ctx.newConstructor(cls, Synthetic, Nil, Nil).entered
     def forwarder(fn: TermSymbol, name: TermName) = {
-      val fwdMeth = fn.copy(cls, name, Synthetic | Method).entered.asTerm
-      DefDef(fwdMeth, prefss => ref(fn).appliedToArgss(prefss))
+      var flags = Synthetic | Method
+      val isOverride = parents.exists(_.member(name).hasAltWith(_.info == fn.info))
```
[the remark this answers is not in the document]
smarter: See `SymDenotation#allOverriddenSymbols`

liufengyun:
@allanrenucci can you try rebase? It seems this PR has some interaction with `isInstanceofCheck` in the case of partial functions.

liufengyun:
Thanks a lot @allanrenucci , I'm investigating it, type inference somehow failed to infer `a <: Int`, given `Some[a] <: X` and `X <: Some[Int]`.

allanrenucci:
Blocked by #4297

compiler/src/dotty/tools/dotc/transform/ExpandSAMs.scala
```diff
@@ -54,38 +54,72 @@ class ExpandSAMs extends MiniPhase {
...
+          ref(applyFn).appliedTo(paramRef)
+        }
+    }
+
     val isDefinedAtDef = transformFollowingDeep(DefDef(isDefinedAtFn, isDefinedAtRhs(_)))
-    val anonCls = AnonClass(tpt.tpe :: Nil, List(applyFn, isDefinedAtFn), List(nme.apply, nme.isDefinedAt))
-    cpy.Block(tree)(List(applyDef, isDefinedAtDef), anonCls)
+    val applyOrElseDef = transformFollowingDeep(DefDef(applyOrElseFn, applyOrElseRhs(_)))
+
+    val anonCls = AnonClass(tpt.tpe :: Nil, List(applyFn, isDefinedAtFn, applyOrElseFn), List(nme.apply, nme.isDefinedAt, nme.applyOrElse))
```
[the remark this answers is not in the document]
odersky: I don't think there's a reason. We could implement them directly.

Blaisorblade:
> if the RHS of the anonymous function is a Match then it is not total otherwise it is

I assume you include `x => { x match { case 1 => x.toString } }` as a PartialFunction, per #4241.

-------

(Edited description so the PR is linked in #4177, "Fix #NNNN" in the title doesn't work).

PR #4250, opened 2018-04-05
Decompile while loops

nicolasstucki:


compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -40,6 +49,12 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
     decl ~~ typeText(nameIdText(tree)) ~ withEnclosingDef(tree) { toTextTemplate(impl) } ~ ""
   }
 
+  override protected def toTextTemplate(impl: Template, ofNew: Boolean = false): Text = {
+    val Template(constr, parents, self, preBody) = impl
+    val impl1 = Template(constr, parents.filterNot(_.symbol.maybeOwner == defn.ObjectClass), self, preBody)
```
allanrenucci:
```scala
val impl1 = impl.copy(parents = impl1.parents.filterNot(_.symbol.maybeOwner == defn.ObjectClass))
```

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -29,8 +36,10 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
       case _ => toTextGlobal(stats, "\n")
     }
     val bodyText =
-      if (currentPrecedence == TopLevelPrec) "\n" ~ statsText else " {" ~ statsText ~ "}"
-    keywordStr("package ") ~ toTextPackageId(tree.pid) ~ bodyText
+      if (tree.pid.symbol.isEmptyPackage) statsText
```
allanrenucci: Isn't this case redundant since you discard it on line 42 if  `tree.pid.symbol.isEmptyPackage`?

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -29,8 +36,10 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
...
+      else if (currentPrecedence == TopLevelPrec) "\n" ~ statsText
+      else " {" ~ statsText ~ "}"
+    (keywordStr("package ") ~ toTextPackageId(tree.pid)).provided(!tree.pid.symbol.isEmptyPackage) ~ bodyText
```
allanrenucci: `tree.pid.symbol` same as `tree.symbol`?

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
nicolasstucki: The body is the only part that is not discarded

PR #4253, opened 2018-04-05
Fix #4225: always do reachability check

### LINKED ISSUE #4225: Missing "unreachable code" warning when adding case after unfalsifiable extractor
The following code produces no warning, even though the `case _` is unreachable since `Bar` returns an instance of `Some` (scalac correctly produces a warning here):
```scala
object Bar {
  def unapply(x: Int): Some[Int] =
    Some(0)
}

object Test {
  def test(x: Int) =
    x match {
      case Bar(a) => a
      case _ => x // should be unreachable
    }
}
```

Interestingly, if we replace `Int` by a sealed trait, we do produce a warning:
```scala
sealed trait Base
case class One() extends Base
object Foo {
  def unapply(x: Base): Some[Int] =
    Some(0)
}

object Test {
  def test(x: Base) =
    x match {
      case Foo(a) => a
      case _ => x
    }
}
```
```scala
-- [E030] Match case Unreachable Warning: try/some.scala:12:13 -----------------
12 |      case _ => x
   |             ^^^^
   |             unreachable code
```

liufengyun:
Fix #4225 

- always do reachability check
- handle null in patterns

Note that we only handle null in reachability check, while ignore null in exhaustivity check. Otherwise, the following pattern match will report a warning about `null`, which is annoying:

```Scala
    (Some(3): Option[Int]) match {
       case Some(x) =>
       case None    =>
    }
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -490,8 +490,6 @@ class PlainPrinter(_ctx: Context) extends Printer {
             Text()
 
         nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ (Str(node.pos.toString) provided ctx.settings.YprintPos.value)
-      case _ =>
-        tree.fallbackToText(this)
     }
   }.close // todo: override in refined printer
 
--- a/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
@@ -473,24 +473,21 @@ class PlainPrinter(_ctx: Context) extends Printer {
     ("Scope{" ~ dclsText(sc.toList) ~ "}").close
 
   def toText[T >: Untyped](tree: Tree[T]): Text = {
-    tree match {
-      case node: Positioned =>
-        def toTextElem(elem: Any): Text = elem match {
-          case elem: Showable => elem.toText(this)
-          case elem: List[_] => "List(" ~ Text(elem map toTextElem, ",") ~ ")"
-          case elem => elem.toString
-        }
-        val nodeName = node.productPrefix
-        val elems =
-          Text(node.productIterator.map(toTextElem).toList, ", ")
-        val tpSuffix =
-          if (ctx.settings.XprintTypes.value && tree.hasType)
-            " | " ~ toText(tree.typeOpt)
-          else
-            Text()
-
-        nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ (Str(node.pos.toString) provided ctx.settings.YprintPos.value)
+    def toTextElem(elem: Any): Text = elem match {
+      case elem: Showable => elem.toText(this)
+      case elem: List[_] => "List(" ~ Text(elem map toTextElem, ",") ~ ")"
+      case elem => elem.toString
     }
+    val nodeName = tree.productPrefix
+    val elems =
+      Text(tree.productIterator.map(toTextElem).toList, ", ")
+    val tpSuffix =
+      if (ctx.settings.XprintTypes.value && tree.hasType)
+        " | " ~ toText(tree.typeOpt)
+      else
+        Text()
+
+    nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ (Str(tree.pos.toString) provided ctx.settings.YprintPos.value)
   }.close // todo: override in refined printer
 
   def toText(result: SearchResult): Text = result match {
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -909,8 +909,15 @@ object messages {
 
   case class MatchCaseUnreachable()(implicit ctx: Context)
   extends Message(MatchCaseUnreachableID) {
-    val kind = s"""Match ${hl"case"} Unreachable"""
+    val kind = s"""Match case Unreachable"""
     val msg = "unreachable code"
+    val explanation = ""
+  }
+
+  case class MatchCaseOnlyNullWarning()(implicit ctx: Context)
+  extends Message(MatchCaseOnlyNullWarningID) {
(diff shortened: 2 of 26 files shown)


compiler/src/dotty/tools/dotc/printing/PlainPrinter.scala
```diff
@@ -490,8 +490,6 @@ class PlainPrinter(_ctx: Context) extends Printer {
             Text()
 
         nodeName ~ "(" ~ elems ~ tpSuffix ~ ")" ~ (Str(node.pos.toString) provided ctx.settings.YprintPos.value)
-      case _ =>
```
smarter: Tree extends Positioned so we don't even need a pattern match at all here.

compiler/src/dotty/tools/dotc/transform/Splicer.scala
```diff
@@ -89,7 +89,6 @@ object Splicer {
          """.stripMargin
         ctx.error(msg, pos)
         None
-      case ex: Throwable => throw ex
```
smarter: This looks wrong to me, a try/catch can always catch a `Throwable`.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -315,6 +315,8 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       else
         Typ(ConstantType(c), false)
     case _: BackquotedIdent => Typ(pat.tpe, false)
+    case Ident(nme.WILDCARD) =>
+      Or(Typ(pat.tpe.stripAnnots, false) :: Typ(ConstantType(Constant(null))) :: Nil)
```
smarter: Would be cleaner to have  `val nullType = ConstantType(Constant(null))` somewhere.

tests/patmat/i4225b.scala
```diff
@@ -0,0 +1,12 @@
+object Bar {
+  def unapply(x: String): Some[Int] =
+    Some(0)
+}
+
+object Test {
+  def test(x: String) =
+    x match {
+      case Bar(a) => a
+      case _ => x // this case is reachable, i.e. test(null)
```
smarter: Suggestion; maybe if the _only_ way to reach a case is through null, we should suggest the user writes `case null` instead of `case _` ?

compiler/src/dotty/tools/dotc/transform/Splicer.scala
liufengyun: Yes, we could keep it, and no warning will be reported. I removed it because the compiler will generate a default case anyway.

smarter: OK.

tests/patmat/i4225b.scala
liufengyun: This is a good idea, it's implemented in the latest commit.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -376,7 +379,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
   /** Is `tp1` a subtype of `tp2`?  */
   def isSubType(tp1: Type, tp2: Type): Boolean = {
-    val res = tp1 <:< tp2 && (tp1 != ConstantType(Constant(null)) || tp2 == ConstantType(Constant(null)))
+    val res = tp1 <:< tp2 && (tp1 != nullType || tp2 == nullType)
```
smarter: the nullType checks could be moved before the subtype checks since they're cheaper.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -936,7 +940,17 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
 
         if (isSubspace(covered, prevs)) {
-          ctx.warning(MatchCaseUnreachable(), cases(i).body.pos)
+          ctx.warning(MatchCaseUnreachable(), pat.pos)
+        }
+
+        // if last case is `_` and only matches `null`, produce a warning
+        if (i == cases.length - 1) {
+          simplify(minus(covered, prevs)) match {
+            case Typ(ConstantType(Constant(null)), _) =>
```
smarter: You can use `` `nullType` `` here

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -909,11 +909,18 @@ object messages {
 
   case class MatchCaseUnreachable()(implicit ctx: Context)
   extends Message(MatchCaseUnreachableID) {
-    val kind = s"""Match ${hl"case"} Unreachable"""
+    val kind = s"""Match case Unreachable"""
```
allanrenucci: interpolator not needed

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -909,11 +909,18 @@ object messages {
...
     val msg = "unreachable code"
     val explanation = ""
   }
 
+  case class MatchCaseOnlyNullWarning()(implicit ctx: Context)
+  extends Message(MatchCaseOnlyNullWarningID) {
+    val kind = s"""Only null matched"""
```
allanrenucci: interpolator not needed

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -909,11 +909,18 @@ object messages {
...
+    val msg = s"Only ${hl"null"} is matched. Consider use `case null =>` instead."
```
allanrenucci: s/use/using

liufengyun:
The performance regression on `exhaustivity S` and `exhaustivity T` is incurred by reporting a friendly error about replacing `case _` with `case null`, which requires subtraction of spaces (`minus` more expensive than `isSubSpace`).

`exhaustivity V` doesn't suffer from the problem, as the subtraction finishes with `Empty` quickly, as there are no overlapping between the cases.

PR #4254, opened 2018-04-05
Drop MethodOrHKCommon, ensure HigherKinded is properly labeled

Blaisorblade:


Blaisorblade:
This failed locally but not on CI, I suppose it's because I didn't clean before launching... but would cleaning hide some bug?

Blaisorblade:
Opened #4258 for the local issues.

PR #4255, opened 2018-04-05
Simplify decompiler printer for lambda

allanrenucci:


tests/run-with-compiler/quote-run-staged-interpreter.check
```diff
@@ -1,6 +1,4 @@
-{
-  (x: Int) => 2.+(x).+(4)
-}
+(x: Int) => 2.+(x).+(4)
```
Blaisorblade: Could the result appear inside a "template" (class/trait)? Because then you sometimes do need parentheses around the closure to avoid `(x: Int) =>` being parsed as a self-type declaration.

allanrenucci: Indeed, if one writes
```scala
class Test {
  ((x: Int) => x)
}
```
It will be pretty printed as
```scala
class Test() extends Object() { 
  (x: Int) => x
}
```
which will be parsed as a self type. But I do think the lightweight syntax is worth the newly introduced bug. One can fix the issue in a separate PR

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -15,9 +16,15 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
   override protected def filterModTextAnnots(annots: List[untpd.Tree]): List[untpd.Tree] =
     annots.filter(_.tpe != defn.SourceFileAnnotType)
 
-  override protected def blockText[T >: Untyped](trees: List[Trees.Tree[T]]): Text = {
-    super.blockText(trees.filterNot(_.isInstanceOf[Closure[_]]))
-  }
+  override protected def blockToText[T >: Untyped](block: Block[T]): Text =
+    block match {
+      case Block((meth @ DefDef(nme.ANON_FUN, _, _, _, _)) :: Nil, _: Closure[T]) =>
```
nicolasstucki: Extract the RHS from the `DefDef`(..., rhs)` directly.

allanrenucci: Should we extract the RHS using `unapply`? We have something like this:
```scala
case class DefDef(..., private var preRhs: LazyTree) {
  def rhs(implicit ctx) = forceIfLazy
}
```
I expect that the RHS should be accessed through `def rhs`

nicolasstucki: Right

PR #4259, opened 2018-04-05
Fix a typo that caused unnecessary cache invalidation

smarter:


smarter:
test performance please

smarter:
retest performance please

smarter:
There's many cases where we call `copySymDenotation` without calling `copyCaches`, but assuming that we have set changeMembers/changeParents correctly in all phase, couldn't we always copy the caches when copySymDenotation is called?

PR #4261, opened 2018-04-05
Optimize ReifyQuotes#mayChange

smarter:


compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -620,7 +620,8 @@ class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
     transform(tp)
   }
 
-  override protected def mayChange(sym: Symbol)(implicit ctx: Context): Boolean = sym.is(Macro)
+  override protected def mayChange(sym: Symbol)(implicit ctx: Context): Boolean =
+    ctx.compilationUnit.containsQuotesOrSplices && sym.isTerm && sym.is(Macro)
```
nicolasstucki: `ctx.compilationUnit.containsQuotesOrSplices` will always be true. We only run this phase if it is true
https://github.com/lampepfl/dotty/pull/4261/files#diff-7b199cbe62cc044a10a26459c0f0fa46R94

nicolasstucki: I was wrong, we do need this check.

nicolasstucki: We could also avoid accessing the symbol 
```scala
  override def transform(ref: SingleDenotation)(implicit ctx: Context): SingleDenotation =
    if (ctx.compilationUnit.containsQuotesOrSplices) super.transform(ref) else ref

  override protected def mayChange(sym: Symbol)(implicit ctx: Context): Boolean = sym.isTerm && sym.is(Macro)
```

smarter: Calling `Denotation#symbol` is just a field access, so I don't think that's worth it.

nicolasstucki: Ok

PR #4266, opened 2018-04-06
Test compilation of decompiled code

nicolasstucki:


nicolasstucki:
Currently it only tests compilation of decompiled code for which a `XYZ.decompiled` file exists.

nicolasstucki:
This PR also contains a partial fix for module object decompilation.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -631,7 +631,8 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
     val tparamsTxt = withEnclosingDef(constr) { tparamsText(tparams) }
     val primaryConstrs = if (constr.rhs.isEmpty) Nil else constr :: Nil
     val prefix: Text =
-      if (vparamss.isEmpty || primaryConstrs.nonEmpty) tparamsTxt
+      if (constr.symbol.owner.is(Module)) " "
```
allanrenucci: Shouldn't this be the empty string?

compiler/src/dotty/tools/dotc/decompiler/Main.scala
```diff
@@ -14,13 +14,14 @@ object Main extends dotc.Driver {
...
-      Files.deleteIfExists(Paths.get(outputDir + ".decompiled"))
+      Files.deleteIfExists(Paths.get(outputDir + "/decompiled.scala"))
     new TASTYDecompiler
   }
 
   override def setup(args0: Array[String], rootCtx: Context): (List[String], Context) = {
     var args = args0.filter(a => a != "-decompile")
-    args = if (args.contains("-from-tasty")) args else "-from-tasty" +: args
+    if (!args.contains("-from-tasty")) args = "-from-tasty" +: args
+    if (args.contains("-d")) args = "-color:never" +: args
```
allanrenucci: why?

compiler/src/dotty/tools/dotc/printing/DecompilerPrinter.scala
```diff
@@ -55,7 +55,11 @@ class DecompilerPrinter(_ctx: Context) extends RefinedPrinter(_ctx) {
   }
 
   override protected def toTextTemplate(impl: Template, ofNew: Boolean = false): Text = {
-    val impl1 = impl.copy(parents = impl.parents.filterNot(_.symbol.maybeOwner == defn.ObjectClass))
+    def filter(sym: Symbol): Boolean = {
```
allanrenucci: Can you give a better name (e.g. `isSyntheticParent`)

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -642,7 +643,7 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
     val selfText = {
       val selfName = if (self.name == nme.WILDCARD) keywordStr("this") else self.name.toString
       (selfName ~ optText(self.tpt)(": " ~ _) ~ " =>").close
-    } provided !self.isEmpty
+    } provided (!self.isEmpty && !constr.symbol.owner.is(Module))
```
allanrenucci: Not sure this should go into `RefinedPrinter`. You could reproduce this behavior in `DecompilerPrinter` by replacing the self type with an `EmptyTree`

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -1259,8 +1260,8 @@ trait ParallelTesting extends RunnerOrchestration { self =>
    *  Tests in the first part of the tuple must be executed before the second.
    *  Both testsRequires explicit delete().
    */
-  def compileTastyInDir(f: String, flags0: TestFlags, blacklist: Set[String] = Set.empty)(
-      implicit testGroup: TestGroup): (CompilationTest, CompilationTest, CompilationTest) = {
+  def compileTastyInDir(f: String, flags0: TestFlags, blacklist: Set[String], recompileBlacklist: Set[String])(
+      implicit testGroup: TestGroup): (CompilationTest, CompilationTest, CompilationTest, CompilationTest) = {
```
allanrenucci: You should probably create the class `TastyCompilationTest` and replace the returned `Tuple4`

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
nicolasstucki: To print `object Foo { .. }` instead of `object Foo{ ... }`.

compiler/src/dotty/tools/dotc/decompiler/Main.scala
nicolasstucki: If we print colors in the output file it is impossible to read (and probably compile) the file.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
allanrenucci: But in `puzzle.decompiled` I see two spaces after `object Test`

compiler/src/dotty/tools/dotc/decompiler/Main.scala
allanrenucci: Shouldn't we do this even if `args` does not contain `-d`?

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
nicolasstucki: Will move it to `DecompilerPrinter`

compiler/src/dotty/tools/dotc/decompiler/Main.scala
nicolasstucki: If args does not contain `-d` the output is printed on the terminal. There it is fine and nicer to have colors. Also copy-pasting from the terminal usually removes the colors.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
nicolasstucki: Finally fixed this one detail

nicolasstucki:
Changes by @allanrenucci LGTM

PR #4267, opened 2018-04-06
Small performance enhancements

nicolasstucki:


nicolasstucki:
test performance please

compiler/src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -44,11 +44,11 @@ class InterceptedMethods extends MiniPhase {
 
   override def phaseName: String = InterceptedMethods.name
 
-  private[this] var primitiveGetClassMethods: Set[Symbol] = _
+  private[this] var scalaValueClasses: scala.collection.Set[Symbol] = _
 
   /** perform context-dependant initialization */
   override def prepareForUnit(tree: Tree)(implicit ctx: Context) = {
-    primitiveGetClassMethods = Set[Symbol]() ++ defn.ScalaValueClasses().map(x => x.requiredMethod(nme.getClass_))
```
biboudis: So, the improvement is that before `prepareForUnit` returned all the symbols for `getClass` methods, by generating those stubsymbols that are needed, and now it doesn't?

compiler/src/dotty/tools/dotc/transform/InterceptedMethods.scala
```diff
@@ -105,7 +105,7 @@ class InterceptedMethods extends MiniPhase {
               List(qual, typer.resolveClassTag(tree.pos, qual.tpe.widen))))
           }*/
          */
-      case t if primitiveGetClassMethods.contains(t) =>
+      case t if t.name == nme.getClass_ && scalaValueClasses.contains(t.owner) =>
```
smarter: I think you could just do defn.ScalaValueClasses.contains, no need for the intermediate variable.

compiler/src/dotty/tools/dotc/transform/InterceptedMethods.scala
nicolasstucki: Yes. Mostly the cost was in creating the set.

nicolasstucki:
test performance please

nicolasstucki:
There is some performance regression on some of the exhaustivity benchmarks

nicolasstucki:
Actually the performance regression was on master.

PR #4269, opened 2018-04-06
Fix #4217: display warnings in repl

### LINKED ISSUE #4217: Warnings not shown in the REPL
The same code compiled by hand does show the warning:
```scala
scala> def foo(x: Option[Int]) = x match { case None => }
def foo(x: Option[Int]): Unit
scala>
```
Could you have a look @liufengyun ?

liufengyun:
Fix #4217: display warnings in repl

### DIFF
--- a/compiler/src/dotty/tools/repl/ReplDriver.scala
+++ b/compiler/src/dotty/tools/repl/ReplDriver.scala
@@ -235,8 +235,8 @@ class ReplDriver(settings: Array[String],
             val newImports = newState.imports ++ extractImports(parsed.trees)
             val newStateWithImports = newState.copy(imports = newImports)
 
-            implicit val ctx: Context = newState.run.runContext
-            displayErrors(newState.run.runContext.flushBufferedMessages())
+            // display warnings
+            displayErrors(newState.run.runContext.flushBufferedMessages())(newState)
 
             displayDefinitions(unit.tpdTree, newestWrapper)(newStateWithImports)
           }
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -235,6 +235,9 @@ class ReplDriver(settings: Array[String],
             val newImports = newState.imports ++ extractImports(parsed.trees)
             val newStateWithImports = newState.copy(imports = newImports)
 
+            implicit val ctx: Context = newState.run.runContext
```
allanrenucci: I don't think this implicit context is needed.

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -235,6 +235,9 @@ class ReplDriver(settings: Array[String],
...
+            displayErrors(newState.run.runContext.flushBufferedMessages())
```
allanrenucci: Can you add a comment that this line display warnings. Otherwise it confusing (the name of the function is misleading).

compiler/src/dotty/tools/repl/ReplDriver.scala
liufengyun: You are right, an implicit State is required.

compiler/src/dotty/tools/repl/ReplDriver.scala
Blaisorblade: Can that function get renamed a bit?

liufengyun: Sure, it's for another PR :)

PR #4277, opened 2018-04-09
Simplify Tasty Trees 

odersky:
Several simplifications and fixes for tasty trees.

Also: Avoid crash if prefix of selection is higher-kinded
The test case caused a crash in NamedType, because `Lambda` is higher-kinded.
We now report an error instead.


tests/pos/TastyADT.scala
```diff
@@ -121,6 +49,25 @@ object tasty {
...
 
+// ------ Definitions ---------------------------------
+
+  class Symbol {
+    def owner: Symbol = ???
+    def definition: Definition = ???
+  }
+  object NoSymbol extends Symbol
+
+  trait Definition {
```
nicolasstucki: Should extend `Statement`

tests/pos/TastyADT.scala
```diff
@@ -76,39 +71,39 @@ object tasty {
...
-    case Super(thiz: Term, mixin: Id | Empty)
-    case Typed(expr: Term, tpt: Term)
+    case TypeApply(fn: Term, args: List[TypeTree])
+    case Super(thiz: Term, mixin: Option[Id])
+    case Typed(expr: Term, tpt: TypeTree)
     case Assign(lhs: Term, rhs: Term)
     case Block(stats: List[Statement], expr: Term)
     case Inlined(call: Term, bindings: List[Definition], expr: Term)
-    case Lambda(method: Term, tpt: Term | Empty)
+    case Lambda(method: Term, tpt: Option[TypeTree])
```
smarter: Why is tpt optional here?

tests/pos/TastyADT.scala
```diff
@@ -202,6 +197,8 @@ object tasty {
     object ErasedImplicitMethodType extends SpecializedMethodTypeCompanion
 
     case class TypeBounds(loBound: Type, hiBound: Type)
+    case class NoPrefix()
+    object NoPrefix extends NoPrefix
```
smarter: Why both a class and an object for NoPrefix?

nicolasstucki: To allow `NoPrefix` to be used in `|` types and be a singleton at the same time.

tests/pos/TastyADT.scala
nicolasstucki: It should probably not be. https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala#L88

nicolasstucki: Actually, the doc is wrong. https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala#L1022-L1025

The docs should be updated to 
```
LAMBDA         Length meth_Term target_Type?
```

nicolasstucki: Created #4285 to fix this

PR #4280, opened 2018-04-09
Add some missing tailrec annotations for clarity

Blaisorblade:
Clarify these methods are tail-recursive (and ensure they are).


allanrenucci:
I will [quote](https://github.com/lampepfl/dotty/pull/2330#discussion_r114077756) Martin here:
>Add a @tailrec only if it is important that the method is in fact tail-recursive. Either it is super-performance critical or can potentially recurse deeply. Do not add @tailrec just because the method happens to be tail recursive.

Closing. Feel free to reopen if you disagree

PR #4282, opened 2018-04-09
Fix #4273: handle dependent parameters in case class

### LINKED ISSUE #4273: Path-dependent classes don't work
The following code doesn't compile:
```scala
trait Foo {
  def config: String

  case class Bar() {
    def doThings: String = config //Do whatever here
  }
}

case class Baz(foo: Foo, bar: foo.Bar) // fails
case class Baz(val foo: Foo, bar: foo.Bar) // still fails
case class Baz(foo: Foo)(val bar: foo.Bar) // works, but equality ignores bar.
```
This came up on [Gitter](https://gitter.im/typelevel/cats?at=5aca3d851130fe3d36b55042) because somebody wanted to write `case class Baz(bar: Foo.Bar)`.

liufengyun:
Fix #4273: handle dependent parameters in case class

Dotty supports the following:

```Scala
trait Foo {
  case class Bar()
}

class Baz(foo: Foo, bar: foo.Bar) // ok
```

But it rejects the code if we make `Baz` a case class:

```Scala
trait Foo {
  case class Bar()
}

case class Baz(foo: Foo, bar: foo.Bar) // fails
```

This PR make dependent params work for case classes. As part of the PR, it makes Dotty supports _dependent pattern matching_ (see tests).


liufengyun:
test performance please

liufengyun:
The use case is the following one:

```scala
trait Foo {
  def config: String

  case class Bar() {
    def doThings: String = config //Do whatever here
  }
}

case class Baz(foo: Foo, bar: foo.Bar) // fails
case class Baz(val foo: Foo, bar: foo.Bar) // still fails
case class Baz(foo: Foo)(val bar: foo.Bar) // works, but equality ignores bar.
```

Now I've changed the code to just use `AnyRef` instead of dependent function types.

liufengyun:
> > Now I've changed the code to just use AnyRef instead of dependent function types.

> I did not see that in the code? 

The original changes to types are removed, now it only uses a DerivedTree to generate `AnyRef` if the construtor is dependent.

odersky:
I still don't see what this is supposed to achieve. I ran the test case in master and got:

-- [E006] Unbound Identifier Error: test.scala:21:30 ---------------------------
21 |case class Baz(foo: Foo, bar: foo.Bar)
   |                              ^^^
   |                              not found: foo

-- [E006] Unbound Identifier Error: test.scala:22:35 ---------------------------
22 |case class Baz2(val foo: Foo, bar: foo.Bar)
   |                                   ^^^
   |                                   not found: foo

If I delete those two lines, the Baz3 case compiles and gives AnyRef as a parent of the companion object. 

So which part of this behavior should change?

odersky:
@liufengyun Ah, I get it now! But is it enough? I think we need a lot more tests that check whether pattern matching, copy, equality checks etc all work.

liufengyun:
Indeed, this is a much trickier issue, it doesn't play well with default parameters and pattern match:

```Scala
class Foo(val config: String) {
  case class Bar() {
    def doThings: String = config //Do whatever here
  }
}

case class Baz(val foo: Foo, bar: foo.Bar = foo.Bar())    // failed: default param for `apply` type error

object Test {
    val foo1 = new Foo("deep")
    val foo2 = new Foo("deep")
    val baz1 = Baz(foo1, foo1.Bar())
    val baz2 = Baz(foo2)

    // pattern matching
    baz1 match {
      case Baz(foo, bar) =>
        bar : foo.Bar                // failed
        assert(foo.Bar() == bar)
    }

    // copy
    baz1.copy(bar = foo1.Bar())     // failed, type doesn't check
    baz1.copy(bar = foo2.Bar())     // failed, type doesn't check

    // defaults
  }
}
```

liufengyun:
The main obstacle here is _dependent pattern matching_. But actually it works out of the box already in Dotty with _product patterns_ and _dependent methods_ (not in Scala as it doesn't support product patterns):

```Scala
class Foo(val config: String) {
  case class Bar()
}

// case class Baz(foo: Foo, bar: foo.Bar)
abstract class Baz {
  val foo: Foo
  val bar: foo.Bar
}

object Baz {
  def unapply(baz: Baz): (baz.foo.type, baz.foo.Bar) = (baz.foo, baz.bar)
}

object Test {
  val foo1 = new Foo("deep")
  val baz1 = new Baz {
    val foo = foo1
    val bar = new foo.Bar()
  }

  baz1 match {
    case Baz(foo, bar) =>      // dependent pattern match
      bar : foo.Bar            // ok
  }
}
```

The problem now is that it's hairy to synthesize the dependent extractor `Baz.unapply`.

liufengyun:
If it's acceptable in such cases to ask programmer to either (1) define the dependent extractor himself or (2) define `def _1: foo.type`, `def _2: foo.Bar` themselves.

In either case, we need a way tell desugaring to not synthesize code for `Baz.unapply` or for `Baz._1` and `Baz._2`.

The first idea is to not synthesize `unapply` for case classes, as they are always identity! This requires some change to the pattern match rules to use a default identity extractor as the last resort. But define an extractor in companion object is more verbose than the following approach.

The second idea is to introduce an annotation like `@dependentProduct`, so that desugaring will not synthesize `def _1: Foo` and `def _2: foo.Bar`. Then the programmer can define them as `def _1: foo.type` and `def _2: foo.Bar`.

WDYT @julienrf @OlivierBlanvillain ?

liufengyun:
@julienrf If we synthesize the type of `_1` to be the more refined type `this.foo`, the dependent pattern matching will work without changing any other stuff.

The problem with `def _1: this.foo = this.foo` is what to do if `foo` is private. If `foo` is private, the synthesized code will not compile in this case. I'm not sure if it's acceptable to end users. 

Otherwise, it's possible to synthesize a more specific type for `_N` if it's referred dependently in other params.

liufengyun:
Discussed during the meeting, if `foo` is private, then we synthesize the type of `_1` as `Foo`.

liufengyun:
test performance please

liufengyun:
This PR is finally green. To be consistent with the existing behavior (as shown in the following code snippet), the compiler produces an error instead of widening private references in the result type:

```Scala
class Foo(val config: String) {
  case class Bar()
}

class Qux(foo: Foo) {
  private val y: foo.Bar = ???
  val z = y              //  the compiler reports an error instead of widening the result type
}
```

liufengyun:
Need to consider exhaustivity check, which is currently missing from this PR.

odersky:
@liufengyun Let's discuss on Monday what to do with this.

liufengyun:
I incline to close this one for now, as it needs a hack in pattern matching `result bind`, and we do not have  a good story for exhaustivity check.

liufengyun:
Let's close it, it's better to do it later than doing it wrong.

PR #4283, opened 2018-04-09
[do not merge] Test upgrading build to 2.13.0-M3

smarter:


liufengyun:
@smarter The library subset is intentionally picked to be compilable for early versions of Dotty. After 2.13.0, we could create dummy traits to make it compile to minimize the difference in performance.

liufengyun:
I just did a hot fix to test a different set of files for `scala-library`, so other benchmarks will finish without problem, if they compile without problem with this PR.

smarter:
Benchmark results look neutral, this gives us a baseline to compare against 2.13.0-M4 with the new collections once it's released. Until then, closing this PR.

PR #4285, opened 2018-04-10
Update tasty format for lambda

nicolasstucki:


Blaisorblade:
The failure seems spurious (http://dotty-ci.epfl.ch/lampepfl/dotty/4365/4 has a timeout, while the same test on the parent succeeded http://dotty-ci.epfl.ch/lampepfl/dotty/4348/4), but 

```
[error] Test dotty.tools.dotc.CompilationTests.runAll failed: java.lang.AssertionError: Run test failed, but should not, reasons:
[error]   - test 'tests/run-with-compiler/i3947f.scala' timed out, took 219.395 sec
```

is either misleading or worth investigating (why would a test take 3 minutes to run? was the machine that overloaded?)

nicolasstucki:
Yes, that test fails from time to time. I have another PR that should improve this.

nicolasstucki:
Because I am only updating the documentation to match the current implementation

nicolasstucki:
See:
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TreePickler.scala#L411-L417
and
https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala#L1022-L1025

PR #4286, opened 2018-04-10
Evolving Tasty definitions

odersky:
Move to separate directory so that we can define other functionality around it.

There might be some more fixes as I try to write some code using these abstractions.

### DIFF
--- a/tests/pos/tasty/definitions.scala
+++ b/tests/pos/tasty/definitions.scala
@@ -1,4 +1,6 @@
-object tasty {
+package tasty
+
+object definitions {
 
 // ------ Names --------------------------------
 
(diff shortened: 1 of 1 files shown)


tests/pos/tasty/definitions.scala
```diff
@@ -8,16 +10,13 @@ object tasty {
...
+
     case DefaultGetter(methodName: TermName, idx: String)           // s"$methodName${"$default$"}${idx+1}"
     case Variant(underlying: TermName, covariant: Boolean)          // s"${if (covariant) "+" else "-"}$underlying"
     case SuperAccessor(underlying: TermName)                        // s"${"super$"}$underlying"
     case ProtectedAccessor(underlying: TermName)                    // s"${"protected$"}$underlying"
     case ProtectedSetter(underlying: TermName)                      // s"${"protected$set"}$underlying"
     case ObjectClass(underlying: TermName)                          // s"$underlying${"$"}"
-
-    case Expanded(prefix: TermName, selector: String)               // s"$prefix${"$$"}$name"  , used only for symbols coming from Scala 2
-    case ExpandedPrefix(prefix: TermName, selector: String)         // s"$prefix${"$"}$name"   , used only for symbols coming from Scala 2
```
nicolasstucki: Why can we remove these ones?

PR #4287, opened 2018-04-10
Fix #3653: generic Java signature of FunctionXXL

### LINKED ISSUE #3653: Java method parameter generic signature error for FunctionXXL
First found by @Glavo

Here:https://github.com/Glavo/FunctionXXL

The `FunctionXXL` 's singnature is Function{$arityNumbers},which will cause an ClassNotFound exception when call `Parameter#getParameterizedType` method.

```
scala> res13(0).getType
val res17: Class[_] = interface scala.FunctionXXL

scala> res13(0).getParameterizedType
java.lang.TypeNotPresentException: Type scala.Function26 not present
```

dotc with veriry-generic-signature will not found any error .

dotc version:0.5.rc1

cc @Glavo

Duhemm:
We were encoding the generic Java signature of `FunctionN`s' with more
than P parameters as `FunctionP<..params..>` (which doesn't exist), but
in the bytecode, FunctionXXL is used instead.

This commit fixes that, so that the generic signature matches
`FunctionXXL`.

Fixes #3653

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
+++ b/compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
@@ -127,6 +127,9 @@ object GenericSignatures {
         case ref @ TypeParamRef(_: PolyType, _) =>
           typeParamSig(ref.paramName.lastPart)
 
+        case RefOrAppliedType(sym, _, _) if defn.isXXLFunctionClass(sym) =>
+          jsig(defn.FunctionXXLType, toplevel, primitiveOK)
+
         case RefOrAppliedType(sym, pre, args) =>
           def argSig(tp: Type): Unit =
             tp match {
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/GenericSignatures.scala
```diff
@@ -127,6 +127,9 @@ object GenericSignatures {
         case ref @ TypeParamRef(_: PolyType, _) =>
           typeParamSig(ref.paramName.lastPart)
 
+        case RefOrAppliedType(sym, _, _) if defn.isXXLFunctionClass(sym) =>
```
allanrenucci: Shouldn't you add it as a new case in the already existing `if else` chain starting line 86

Duhemm: Indeed

PR #4288, opened 2018-04-10
Fuse the FirstTransform and CheckStatic group

smarter:


smarter:
test performance please

smarter:
Hmm, looks like this makes no difference in benchmarks. That's kind of surprising since I'm pretty sure I've seen performance gets worse when splitting a phase group into two before.

smarter:
test performance please

smarter:
Which part did you find more complex? If it's just https://github.com/lampepfl/dotty/commit/c94087082f4f4119acd5281b781fce7178dc9129I can get rid of it since it doesn't seem to affect performance.

smarter:
Fusing FirstTransform and the next group also means that RefChecks gets closer to the frontend. If we decide to run the compiler until RefChecks in the IDE to get better diagnostics, this may improve latency.

odersky:
This would have to be reworked considerably to be up-to-date with the latest phase orderings, and I still don't see a clear benefit. @smarter Let's discuss on Monday what to do about it.

PR #4290, opened 2018-04-10
WIP Relax realizability for erased members, treat them like lazy ones

Blaisorblade:
This WIP PR (missing tests) builds on top of #4036.

Partially address TODO on erased members by treating them like lazy values: they
do not exist at runtime, and their initialization might fail, but they're
realizable if their type is.

I do not treat erased members as _final_ lazy ones because erased definitions
are not automatically final, and it's not 100% obvious they should be (after
discussion with Nicholas Stucki). Current situation:

If `U <: T`, the current rules allow the following override

```scala
trait A {
  erased def foo: T = ...
}
trait B extends A {
  erased def foo: U = ...
}
```

And just like lazy values, `U` might be unrealizable even when `T` is, so it's
unsafe to treat `(a: A).foo` as realizable.

The above code might be used (questionably) through
```scala
def takeErased(erased a: T): Any
def takeBetterErased(erased a: U): String
def f(a: A) = a match {
  case b: B =>
    takeBetterErased(b.foo)
  case _ =>
    takeErased(a.foo)
}
```
It's very unclear whether that's desirable.

Alternatively, concrete erased definitions might be made automatically final,
and that would be picked up by `isEffectivelyFinal` without further changes.

/cc @nicolasstucki @odersky


Blaisorblade:
Hm, the fact that `tests/neg/i4060.scala` fails suggests that even this change might not be conservative enough.

Blaisorblade:
TODO look into this again; I was convinced that invalid lazy vals would always fail while erased ones don't, but recursive lazy vals can produce `null` instead of failing, so erased vals are just as unsafe. OTOH, `null` is currently a soundness hole...

PR #4292, opened 2018-04-11
Fix #4291: Copy sticky attachments in `TreeCopier`

### LINKED ISSUE #4291: `TreeCopier` doesn't preserve attachments
While working on embedding the doc in Tasty, I noticed that the `TreeCopier` subclasses don't copy over the attachments of the trees.

Is this on purpose?

Doc comments are stored as attachments on `MemberDef` trees, and cannot be recovered in the REPL for instance, because they're discarded when the trees are modified.

Duhemm:
Sticky attachments are attachments whose key is a subclass of
`StickyKey`. Those attachments should be copied when a tree is copied
using a `TreeCopier`.

Fixes #4291


Duhemm:
test performance please

PR #4298, opened 2018-04-11
Fix #4297: handle type param reference

### LINKED ISSUE #4297: Spurious is instanceOf warning
As reported by @allanrenucci  in #4245, the following code will get a warning saying `Some[Int]` cannot be checked at runtime.

```Scala
class Test {
  def test[X <: Option[Int]](x: X) = x.isInstanceOf[Some[Int]]
}
```

Initial diagnosis shows that type inference somehow failed to infer `a <: Int`, given `Some[a] <: X` and `X <: Option[Int]`.

liufengyun:
Fix #4297: handle type param reference

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
+++ b/compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
@@ -88,19 +88,33 @@ object Checkable {
       }
     }
 
+    def stripTypeParam(implicit ctx: Context) = new ApproximatingTypeMap {
+      def apply(tp: Type): Type = tp match {
+        case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
+          val lo = this(tp.info.loBound)
+          val hi = this(tp.info.hiBound)
+          range(lo, hi)
+        case _ =>
+          mapOver(tp)
+      }
+    }
+
     def isClassDetermined(X: Type, P: AppliedType)(implicit ctx: Context) = {
       val AppliedType(tycon, _) = P
       val typeLambda = tycon.ensureLambdaSub.asInstanceOf[TypeLambda]
       val tvars = constrained(typeLambda, untpd.EmptyTree, alwaysAddTypeVars = true)._2.map(_.tpe)
       val P1 = tycon.appliedTo(tvars)
 
-      debug.println("P : " + P.show)
-      debug.println("P1 : " + P1.show)
-      debug.println("X : " + X.show)
+      debug.println("P : " + P)
+      debug.println("P1 : " + P1)
+      debug.println("X : " + X)
 
-      P1 <:< X  // may fail, ignore
+      P1 <:< X       // constraint P1
 
-      val res = isFullyDefined(P1, ForceDegree.noBottom) &&  P1 <:< P
+      // use fromScala2x to avoid generating pattern bound symbols
+      maximizeType(P1, pos, fromScala2x = true)
+
+      val res = P1 <:< P
       debug.println("P1 : " + P1)
       debug.println("P1 <:< P = " + res)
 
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -88,19 +88,35 @@ object Checkable {
       }
     }
 
+    def maximize(implicit ctx: Context) = new ApproximatingTypeMap {
+      def apply(tp: Type): Type = tp match {
+        case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
+          val lo = this(tp.info.loBound)
+          val hi = this(tp.info.hiBound)
+          if (variance == 0) range(lo, hi)
```
smarter: The `range` method already does variance checks like this.

liufengyun: Thanks 👍

liufengyun:
test performance with #fast please

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -116,7 +130,9 @@ object Checkable {
           case defn.ArrayOf(tpE)   => recur(tpE, tpT)
           case _                   => recur(defn.AnyType, tpT)
         }
-      case tpe: AppliedType     => isClassDetermined(X, tpe)(ctx.fresh.setNewTyperState())
+      case tpe: AppliedType     =>
+        isClassDetermined(X, tpe)(ctx.fresh.setNewTyperState()) ||
```
allanrenucci: Here, you first try without stripping type parameters for performance reasons? Maybe add a comment

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -88,19 +88,33 @@ object Checkable {
       }
     }
 
+    def stripTypeParam(implicit ctx: Context) = new ApproximatingTypeMap {
+      def apply(tp: Type): Type = tp match {
+        case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
```
allanrenucci: Probably equivalent to `case tp: TypeRef if tp.symbol.isTypeParam`. Maybe clearer?

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -88,19 +88,33 @@ object Checkable {
       }
     }
 
+    def stripTypeParam(implicit ctx: Context) = new ApproximatingTypeMap {
```
allanrenucci: Maybe add a comment: `/** Approximate type parameters depending on variance */`

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
```diff
@@ -116,7 +130,9 @@ object Checkable {
           case defn.ArrayOf(tpE)   => recur(tpE, tpT)
           case _                   => recur(defn.AnyType, tpT)
         }
-      case tpe: AppliedType     => isClassDetermined(X, tpe)(ctx.fresh.setNewTyperState())
+      case tpe: AppliedType     =>
+        isClassDetermined(X, tpe)(ctx.fresh.setNewTyperState()) ||
+        isClassDetermined(stripTypeParam.apply(X), tpe)(ctx.fresh.setNewTyperState())
```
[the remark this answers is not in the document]
liufengyun: It seems scalac has problem with this syntax, however, I've changed the method so we don't need to call `apply` anymore.

compiler/src/dotty/tools/dotc/transform/IsInstanceOfChecker.scala
liufengyun: `tp.symbol.isTypeParam` doesn't cover the case of type members, so I think it's better to keep the original version here.

PR #4299, opened 2018-04-11
use ApproximatingTypeMap to simplify exhaustivity check code

liufengyun:
use ApproximatingTypeMap to simplify exhaustivity check code

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -605,32 +605,16 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
    *
    */
   def instantiate(tp1: NamedType, tp2: Type)(implicit ctx: Context): Type = {
-    // expose abstract type references to their bounds or tvars according to variance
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -605,32 +605,16 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-        else if (variance == 1)
-          if (maximize) hi else lo
-        else
-          if (maximize) lo else hi
+    // expose type param references to their bounds according to variance
+    class AbstractTypeMap(maximize: Boolean)(implicit ctx: Context) extends ApproximatingTypeMap {
+      variance = if (maximize) 1 else -1
 
       def apply(tp: Type): Type = tp match {
         case tp: TypeRef if tp.underlying.isInstanceOf[TypeBounds] =>
```
allanrenucci: While you're at it, maybe replace this by `tp.symbol.isTypeParam`

liufengyun: @allanrenucci : `tp.symbol.isTypeParam` will miss the case where the reference is a type member.

PR #4304, opened 2018-04-12
Let VSCode start the language server

Duhemm:
We used to start the language server with `launchIDE`. However, this
means that the language server is dependent of sbt being running and
configured, which is not a good user experience for newcomers.

For newcomers, we want the language server to start when they create a
new Scala file in an empty directory, for instance, which means that no
build is already configured.

~~This commits takes us closer to this experience by using [load-plugin](https://github.com/scalacenter/load-plugin)
to inject the Dotty plugin. This means that the Dotty plugin can now be
loaded even if there's no build configured, which happens if the user
opens VSCode in an empty directory, for instance.~~

We use the recently introduced `--addPluginSbtFile` to provide this experience.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -980,14 +980,13 @@ object Build {
         val defaultIDEConfig = baseDirectory.value / "out" / "default-dotty-ide-config"
         IO.write(defaultIDEConfig, dottyVersion)
         val dottyPluginSbtFile = baseDirectory.value / "out" / "dotty-plugin.sbt"
-        IO.write(dottyPluginSbtFile, s"""addSbtPlugin("$dottyOrganization" % $sbtDottyName % "$sbtDottyVersion")""")
+        IO.write(dottyPluginSbtFile, s"""addSbtPlugin("$dottyOrganization" % "$sbtDottyName" % "$sbtDottyVersion")""")
         Seq(defaultIDEConfig, dottyPluginSbtFile)
       },
       compile in Compile := Def.task {
         val workingDir = baseDirectory.value
         val coursier = workingDir / "out" / "coursier"
         val packageJson = workingDir / "package.json"
-        // val _ = (managedResources in Compile).value
         if (!coursier.exists || packageJson.lastModified > coursier.lastModified)
           runProcess(Seq("npm", "install"), wait = true, directory = workingDir)
         val tsc = workingDir / "node_modules" / ".bin" / "tsc"
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -53,6 +53,11 @@ object Build {
   }
   val dottyNonBootstrappedVersion = dottyVersion + "-nonbootstrapped"
 
+  val sbtDottyVersion = {
+    val base = "0.2.3"
+    if (isRelease) base else base + "-SNAPSHOT"
+  }
+
   val agentOptions = List(
     // "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
     // "-agentpath:/home/dark/opt/yjp-2013-build-13072/bin/linux-x86-64/libyjpagent.so"
@@ -930,11 +935,7 @@ object Build {
   lazy val `sbt-dotty` = project.in(file("sbt-dotty")).
     settings(commonSettings).
     settings(
-      version := {
-        val base = "0.2.3"
-        if (isRelease) base else base + "-SNAPSHOT"
-      },
-
+      version := sbtDottyVersion,
       // Keep in sync with inject-sbt-dotty.sbt
       libraryDependencies ++= Seq(
         Dependencies.`jackson-databind`,
@@ -973,10 +974,16 @@ object Build {
       publishArtifact := false,
       includeFilter in unmanagedSources := NothingFilter | "*.ts" | "**.json",
       watchSources in Global ++= (unmanagedSources in Compile).value,
-      compile in Compile := {
+      resourceGenerators in Compile += Def.task {
+        val out = baseDirectory.value / "out" / "default-dotty-ide-config"
+        IO.writeLines(out, Seq(dottyVersion, sbtDottyVersion))
+        Seq(out)
+      },
+      compile in Compile := Def.task {
         val workingDir = baseDirectory.value
-        val coursier = workingDir / "out/coursier"
+        val coursier = workingDir / "out" / "coursier"
         val packageJson = workingDir / "package.json"
+        // val _ = (managedResources in Compile).value
         if (!coursier.exists || packageJson.lastModified > coursier.lastModified)
           runProcess(Seq("npm", "install"), wait = true, directory = workingDir)
         val tsc = workingDir / "node_modules" / ".bin" / "tsc"
@@ -988,7 +995,7 @@ object Build {
         runProcess(codeCommand.value ++ Seq("--install-extension", "daltonjorge.scala"), wait = true)
 
         sbt.internal.inc.Analysis.Empty
-      },
+      }.dependsOn(managedResources in Compile).value,
       sbt.Keys.`package`:= {
         runProcess(Seq("vsce", "package"), wait = true, directory = baseDirectory.value)
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -53,6 +53,7 @@ object Build {
   }
   val dottyNonBootstrappedVersion = dottyVersion + "-nonbootstrapped"
 
+  val sbtDottyName = "sbt-dotty"
   val sbtDottyVersion = {
     val base = "0.2.3"
     if (isRelease) base else base + "-SNAPSHOT"
@@ -935,6 +936,7 @@ object Build {
   lazy val `sbt-dotty` = project.in(file("sbt-dotty")).
     settings(commonSettings).
     settings(
+      name := sbtDottyName,
       version := sbtDottyVersion,
       // Keep in sync with inject-sbt-dotty.sbt
       libraryDependencies ++= Seq(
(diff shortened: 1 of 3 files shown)


vscode-dotty/src/extension.ts
```diff
@@ -17,12 +17,9 @@ export function activate(context: ExtensionContext) {
   outputChannel = vscode.window.createOutputChannel('Dotty Language Client');
 
   const artifactFile = `${vscode.workspace.rootPath}/.dotty-ide-artifact`
+  const defaultArtifact = "ch.epfl.lamp:dotty-language-server_0.8:0.8.0-bin-SNAPSHOT"
```
Duhemm: If this get in, we'd need to find a way to keep these values up-to-date with the reality. When's the vscode extension published? Every 6 weeks, too?

vscode-dotty/src/extension.ts
```diff
@@ -77,15 +74,57 @@ function fetchAndRun(artifact: string) {
         throw new Error(msg)
       }
 
-      run({
-        command: "java",
-        args: ["-classpath", classPath, "dotty.tools.languageserver.Main", "-stdio"]
+      configureIDE().then((res) => {
+        run({
+          command: "java",
+          args: ["-classpath", classPath, "dotty.tools.languageserver.Main", "-stdio"]
```
Duhemm: Any reason why we extract the classpath with `coursier fetch` and then run this way rather than doing `coursier launch`?

vscode-dotty/package.json
```diff
@@ -24,6 +24,7 @@
   ],
   "main": "./out/src/extension",
   "activationEvents": [
+    "onLanguage:scala",
```
smarter: The problem with doing this is that we conflict with every other language server for Scala

vscode-dotty/src/extension.ts
smarter: You can get the current nightly base of dotty from http://dotty.epfl.ch/versions/latest-nightly-base, we could also have a file to get the current release version.

vscode-dotty/src/extension.ts
smarter: We display a progress notification while coursier fetch is running, if we use coursier launch we don't know when downloading has finished.

vscode-dotty/package.json
```diff
@@ -51,7 +51,7 @@
     "vscode:prepublish": "npm install && ./node_modules/.bin/tsc -p ./",
     "compile": "./node_modules/.bin/tsc -p ./",
     "test": "node ./node_modules/vscode/bin/test",
-    "postinstall": "node ./node_modules/vscode/bin/install && curl -L -o out/coursier https://github.com/coursier/coursier/raw/v1.0.0/coursier && curl -L -o out/load-plugin.jar https://github.com/scalacenter/load-plugin/releases/download/v0.1.0/load-plugin_2.12-0.1.0.jar"
+    "postinstall": "node ./node_modules/vscode/bin/install && curl -L -o out/coursier https://github.com/coursier/coursier/raw/v1.0.0/coursier && curl -L -o out/load-plugin.jar https://oss.sonatype.org/content/repositories/releases/ch/epfl/scala/load-plugin_2.12/0.1.0+2-496ac670/load-plugin_2.12-0.1.0+2-496ac670.jar"
```
smarter: Can't you use coursier to fetch the load-plugins jar?

vscode-dotty/src/extension.ts
```diff
@@ -77,15 +74,57 @@ function fetchAndRun(artifact: string) {
...
+          "org.scala-sbt:sbt-launch:1.1.2", "--",
+          applyLoadPlugin,
+          ifAbsentCommands,
+          "configureIDE"
+      ])
+    const sbtProc = sbtPromise.childProcess
+
+    sbtProc.on('close', (code: number) => {
+      if (code != 0) {
+        let msg = "Configuring the IDE failed."
```
allanrenucci: `const msg`?

laughedelic:
It's cool to see load-plugin in action 👏 But I expected that it happens in the server:
1. user opens a file in the client (editor)
2. client launches server (either reading from `.dotty-ide-artifact` or using some default)
3. server checks if `.dotty-ide.json` is in place, if not runs load-plugin and `configureIDE`

This way client implementation stays minimal: it only needs to launch server using coursier. And this is important IMO because other clients may or may not reimplement the same thing: users experience will vary. If server takes care of its own automatic setup, then clients are simple and users get same awesome experience regardless of their editor of choice.

vscode-dotty/package.json
Duhemm: What can we do instead? The goal of this PR is to be able to get the Dotty IDE working without any prior setup (so, nobody would have generated `.dotty-ide.json`).

Duhemm:
Hi @laughedelic!

I agree that this could be pushed to the server implementation rather than the client, but I think it has been done this way for two reasons:

- Make the client more interactive (for instance, show that it is downloading stuff). I'm not sure how this could be done if the server was the one in charge of downloading sbt, and load-plugin, but LSP may very well have something for that...
- Have fewer dependencies in the LSP server. I think that coursier comes with a ton of dependencies, unfortunately.

I'd be open to try the approach that you suggest. Having the smallest client possible is definitely a big bonus. I think that @smarter may be able to shed more light on why the client works the way it does.

Duhemm:
@smarter I think I've addressed your comments, apart from the one regarding the `activationEvents`, because I don't know what to do.

The goal is to be able to start the Dotty LSP and automatically configure it when the user opens a `.scala` file, so I don't see any way around the conflict with other LSP servers.

smarter:
First we need to actually test how language server interact in VSCode. What happens when I have both one of the Scala 2 LS and the Dotty LS with this PR enabled at the same time, in a Scala 2 project? The Dotty LS shouldn't be doing anything since the sbt plugin won't find any Dotty project, so it's possible that things "just work". If they don't, it might help to actually shut down the Dotty LS if no Dotty project has been found, that way only one LS would run at a given time for a given project.

Duhemm:
I changed it so that the extension will ask the user whether the IDE should be started when it detects that there's no configuration (no `.dotty-ide.json`, no `build.sbt`) :

screen shot 2018-06-11 at 14 27 35

vscode-dotty/package.json
Duhemm: Now, if there's no configuration for the IDE a message is shown asking the user whether she wants to start the IDE or not. If there's a conflict with another language server, she can just say no. If the IDE is configured, then it'll automatically start up.

Duhemm:
I updated the PR to use `--addPluginSbtFile`.

vscode-dotty/src/extension.ts
```diff
@@ -16,73 +16,132 @@ export function activate(context: ExtensionContext) {
...
+    })
+
+  } else {
+    // Check whether `.dotty-ide-artifact` exists. If it does, start the language server,
+    // otherwise, try propose to start it if there's no build.sbt
+    if (fs.existsSync(languageServerArtifactFile)) {
+      runLanguageServer(coursierPath, languageServerArtifactFile)
+    } else if (!fs.existsSync(buildSbtFile)) {
+      vscode.window.showInformationMessage(
+          "This looks like an unconfigured project. Would you like to start Dotty IDE?",
```
smarter: "an unconfigured Scala project", "start the Dotty IDE"

Also maybe remember the choice of the user by creating an empty file ".dotty-ide-disabled" ?

vscode-dotty/src/extension.ts
```diff
@@ -16,73 +16,132 @@ export function activate(context: ExtensionContext) {
...
-    }
-    const artifact = data.toString().trim()
-
-    if (process.env['DLS_DEV_MODE']) {
-      const portFile = `${vscode.workspace.rootPath}/.dotty-ide-dev-port`
-      fs.readFile(portFile, (err, port) => {
-        if (err) {
-          outputChannel.append(`Unable to parse ${portFile}`)
-          throw err
+  const sbtArtifact = "org.scala-sbt:sbt-launch:1.2.0"
```
smarter: Upgrade to 1.2.3

vscode-dotty/src/extension.ts
```diff
@@ -16,73 +16,132 @@ export function activate(context: ExtensionContext) {
...
+        }
+      })
+      return coursierPromise.then(() => { return classPath })
     })
+}
+
+function configureIDE(sbtClasspath: string, languageServerScalaVersion: string, dottyPluginSbtFile: string) {
+  return vscode.window.withProgress({
+    location: vscode.ProgressLocation.Window,
+    title: 'Configuring IDE...'
```
smarter: "Configuring the IDE for Dotty"

project/Build.scala
```diff
@@ -942,10 +945,18 @@ object Build {
...
+        val dottyPluginSbtFile = baseDirectory.value / "out" / "dotty-plugin.sbt"
+        IO.write(dottyPluginSbtFile, s"""addSbtPlugin("$dottyOrganization" % $sbtDottyName % "$sbtDottyVersion")""")
+        Seq(defaultIDEConfig, dottyPluginSbtFile)
+      },
+      compile in Compile := Def.task {
         val workingDir = baseDirectory.value
-        val coursier = workingDir / "out/coursier"
+        val coursier = workingDir / "out" / "coursier"
         val packageJson = workingDir / "package.json"
+        // val _ = (managedResources in Compile).value
```
smarter: Can be deleted.

Duhemm:
@smarter Comments addressed.

smarter:
OK, feel free to merge.

PR #4311, opened 2018-04-13
Remove Two Warts: Auto-Tupling and Multi-Parameter Infix Operations

odersky:
Prompted by the Scala contributors thread, this was initially an attempt to drop auto-tupling altogether. But it became quickly annoying. The main problem is that it's unnatural to add another pair of parentheses to infix operations. Compare:

    (x, y) == z
    z == ((x, y))    // !yuck

Same thing for cons:

    ((1, x)) :: ((2, y)) :: xs   // double yuck!

So at the very least we have to allow auto-tupling for arguments of infix operators. I went a little bit
further and also allowed auto-tupling if the expected type is already a (some kind of) product or tuple of the right arity. The result is this commit.

The language import `noAutoTupling` has been removed. The previous auto-tupling is still reported under Scala 2 mode, and there is -rewrite support to add the missing (...) automatically.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -151,7 +151,7 @@ class Typer extends Namer
     def qualifies(denot: Denotation): Boolean =
       reallyExists(denot) &&
         !(pt.isInstanceOf[UnapplySelectionProto] &&
-          (denot.symbol is (Method, butNot = Accessor))) &&
+          (denot.symbol.is(Method, butNot = Accessor))) &&
         !(denot.symbol is PackageClass)
 
     /** Find the denotation of enclosing `name` in given context `ctx`.
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2032,6 +2032,45 @@ class Typer extends Namer
     }
   }
 
+  def supportsAutoTupling(tp: Type, args: List[untpd.Tree])(implicit ctx: Context): Boolean = {
+    def test(tp: Type): Boolean = tp match {
+      case tp: TypeRef =>
+        val sym = tp.symbol
+        defn.isTupleClass(sym) && defn.arity(sym, str.Tuple) == args.length ||
+        defn.isProductClass(sym) && defn.arity(sym, str.Product) == args.length ||
+        !sym.isClass && test(tp.underlying)
+      case tp: TypeParamRef =>
+        val bounds = ctx.typeComparer.bounds(tp)
+        test(bounds.lo) || test(bounds.hi)
+      case tp: TypeProxy =>
+        test(tp.underlying)
+      case _ =>
+        false
+    }
+    test(tp) || {
+      val pos = Position(args.map(_.pos.start).min, args.map(_.pos.end).max)
+      ctx.testScala2Mode(
+          i"""auto-tupling is no longer supported in this case,
+              |arguments now need to be enclosed in parentheses (...).""",
+          pos, { patch(pos.startPos, "("); patch(pos.endPos, ")") })
+    }
+  }
+
+  def takesAutoTupledArgs(tp: Type, sym: Symbol, args: List[untpd.Tree])(implicit ctx: Context): Boolean = tp match {
+    case tp: MethodicType =>
+      tp.firstParamTypes match {
+        case ptype :: Nil =>
+          !ptype.isRepeatedParam && {
+            sym.name.isOpName || supportsAutoTupling(ptype, args)
+          }
+        case _ => false
+      }
+    case tp: TermRef =>
+      tp.denot.alternatives.forall(alt => takesAutoTupledArgs(alt.info, alt.symbol, args))
+    case _ =>
+      false
+  }
+
   /** Perform the following adaptations of expression, pattern or type `tree` wrt to
    *  given prototype `pt`:
    *  (1) Resolve overloading
@@ -2119,21 +2158,9 @@ class Typer extends Namer
         }
       }
 
-      def isUnary(tp: Type): Boolean = tp match {
-        case tp: MethodicType =>
-          tp.firstParamTypes match {
-            case ptype :: Nil => !ptype.isRepeatedParam
-            case _ => false
-          }
-        case tp: TermRef =>
-          tp.denot.alternatives.forall(alt => isUnary(alt.info))
-        case _ =>
-          false
-      }
-
       def adaptToArgs(wtp: Type, pt: FunProto): Tree = wtp match {
         case _: MethodOrPoly =>
-          if (pt.args.lengthCompare(1) > 0 && isUnary(wtp) && ctx.canAutoTuple)
+          if (pt.args.lengthCompare(1) > 0 && takesAutoTupledArgs(wtp, tree.symbol, pt.args))
             adapt(tree, pt.tupled, locked)
           else
             tree
@@ -2499,10 +2526,10 @@ class Typer extends Namer
         case tp: FlexType =>
           ensureReported(tp)
           tree
-        case ref: TermRef =>
+        case ref: TermRef => // this case can happen in case tree.tpe is overloaded
           pt match {
             case pt: FunProto
-            if pt.args.lengthCompare(1) > 0 && isUnary(ref) && ctx.canAutoTuple =>
+            if pt.args.lengthCompare(1) > 0 && takesAutoTupledArgs(ref, ref.symbol, pt.args) =>
               adapt(tree, pt.tupled, locked)
             case _ =>
               adaptOverloaded(ref)
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1709,7 +1709,7 @@ class Typer extends Namer
    */
   def typedInfixOp(tree: untpd.InfixOp, pt: Type)(implicit ctx: Context): Tree = {
     val untpd.InfixOp(l, op, r) = tree
-    val app = typedApply(desugar.binop(l, op, r), pt)
+    val app = typedApply(desugar.binop(l, op, r), pt, scala2InfixOp = ctx.scala2Mode)
     if (untpd.isLeftAssoc(op.name)) app
     else {
       val defs = new mutable.ListBuffer[Tree]
@@ -2037,22 +2037,13 @@ class Typer extends Namer
   def canAutoTuple(args: List[untpd.Tree])(implicit ctx: Context): Boolean = {
     val pos = Position(args.map(_.pos.start).min, args.map(_.pos.end).max)
     ctx.testScala2Mode(
-        i"""auto-tupling is no longer supported in this case,
-            |arguments now need to be enclosed in parentheses (...).""",
+        i"""auto-tupling is no longer supported,
+           |arguments now need to be enclosed in parentheses (...).""",
         pos, { patch(pos.startPos, "("); patch(pos.endPos, ")") })
   }
 
-  def takesAutoTupledArgs(tp: Type, sym: Symbol, args: List[untpd.Tree])(implicit ctx: Context): Boolean = tp match {
-    case tp: MethodicType =>
-      tp.firstParamTypes match {
-        case ptype :: Nil => !ptype.isRepeatedParam && (sym.name.isOpName || canAutoTuple(args))
-        case _ => false
-      }
-    case tp: TermRef =>
-      tp.denot.alternatives.forall(alt => takesAutoTupledArgs(alt.info, alt.symbol, args))
-    case _ =>
-      false
-  }
+  def takesAutoTupledArgs(tp: Type, pt: FunProto)(implicit ctx: Context): Boolean =
+    pt.args.lengthCompare(1) > 0 && isUnary(tp) && (pt.scala2InfixOp || canAutoTuple(pt.args))
 
   /** Perform the following adaptations of expression, pattern or type `tree` wrt to
    *  given prototype `pt`:
@@ -2143,10 +2134,8 @@ class Typer extends Namer
 
       def adaptToArgs(wtp: Type, pt: FunProto): Tree = wtp match {
         case _: MethodOrPoly =>
-          if (pt.args.lengthCompare(1) > 0 && takesAutoTupledArgs(wtp, tree.symbol, pt.args))
-            adapt(tree, pt.tupled, locked)
-          else
-            tree
+          if (takesAutoTupledArgs(wtp, pt)) adapt(tree, pt.tupled, locked)
+          else tree
         case _ => tryInsertApplyOrImplicit(tree, pt, locked) {
           errorTree(tree, MethodDoesNotTakeParameters(tree, methPart(tree).tpe)(err))
         }
@@ -2511,8 +2500,7 @@ class Typer extends Namer
           tree
         case ref: TermRef => // this case can happen in case tree.tpe is overloaded
           pt match {
-            case pt: FunProto
-            if pt.args.lengthCompare(1) > 0 && takesAutoTupledArgs(ref, ref.symbol, pt.args) =>
+            case pt: FunProto if takesAutoTupledArgs(ref, pt) =>
               adapt(tree, pt.tupled, locked)
             case _ =>
               adaptOverloaded(ref)
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1434,6 +1434,8 @@ class Typer extends Namer
       for (param <- tparams1 ::: vparamss1.flatten)
         checkRefsLegal(param, sym.owner, (name, sym) => sym.is(TypeParam), "secondary constructor")
 
+    checkSymbolicUnary(sym)
+
     assignType(cpy.DefDef(ddef)(name, tparams1, vparamss1, tpt1, rhs1), sym)
       //todo: make sure dependent method types do not depend on implicits or by-name params
   }
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2032,37 +2032,18 @@ class Typer extends Namer
     }
   }
 
-  def supportsAutoTupling(tp: Type, args: List[untpd.Tree])(implicit ctx: Context): Boolean = {
-    def test(tp: Type): Boolean = tp match {
-      case tp: TypeRef =>
-        val sym = tp.symbol
-        defn.isTupleClass(sym) && defn.arity(sym, str.Tuple) == args.length ||
-        defn.isProductClass(sym) && defn.arity(sym, str.Product) == args.length ||
-        !sym.isClass && test(tp.underlying)
-      case tp: TypeParamRef =>
-        val bounds = ctx.typeComparer.bounds(tp)
-        test(bounds.lo) || test(bounds.hi)
-      case tp: TypeProxy =>
-        test(tp.underlying)
-      case _ =>
-        false
-    }
-    test(tp) || {
-      val pos = Position(args.map(_.pos.start).min, args.map(_.pos.end).max)
-      ctx.testScala2Mode(
-          i"""auto-tupling is no longer supported in this case,
-              |arguments now need to be enclosed in parentheses (...).""",
-          pos, { patch(pos.startPos, "("); patch(pos.endPos, ")") })
-    }
+  def canAutoTuple(args: List[untpd.Tree])(implicit ctx: Context): Boolean = {
+    val pos = Position(args.map(_.pos.start).min, args.map(_.pos.end).max)
+    ctx.testScala2Mode(
+        i"""auto-tupling is no longer supported in this case,
+            |arguments now need to be enclosed in parentheses (...).""",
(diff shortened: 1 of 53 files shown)


odersky:
Comparing with the status quo: Auto-tupling of method arguments is supported if 

 1. There are at least two arguments
 2. The method has a single parameter (and in the case of overloaded methods, all alternatives
    have a single parameter.)

Both of these conditions are missing in `scalac`s auto-tupling. Arguably, all reported confusing situations are already eliminated by these two conditions. So it is not clear why we would want to go further in restricting auto-tupling. It would complicate the spec and implementation, and I don't see really a gain.

So my current position would be to simply drop the `noAutoTupling` language import and keep the status quo.

sjrd:
Note that the two `yuck`s are because we allow infix method application with multiple parameters ... which is never what I want. Is there any known good use case for that?

If infix application only admitted one parameter, then `z == (x, y)` would parse the same way as `z == ((x, y))` which we need now. Same for the `::` example.

sjrd:
```scala
buf += ("hello", "world")
```
Well, see, I don't even know whether that's supposed to add 2 elements to `buf` (`"hello"` and `"world"`) or 1 element (`("hello", "world")`) ...

An easy migration path, if a bit long:

1. Deprecation warning when multiple parameters are used in an infix method call (without changing any semantics). This forces people to use dot notation if they want it.
2. Compilation error in the same situation. Still does not change any existing semantics.
3. Apply the new interpretation, where the argument is necessarily unique, hence a tuple.

`-Xsource:n+1` can be used to get one step ahead in an opt-in way, as usual.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2032,6 +2032,45 @@ class Typer extends Namer
     }
   }
 
+  def supportsAutoTupling(tp: Type, args: List[untpd.Tree])(implicit ctx: Context): Boolean = {
+    def test(tp: Type): Boolean = tp match {
+      case tp: TypeRef =>
+        val sym = tp.symbol
+        defn.isTupleClass(sym) && defn.arity(sym, str.Tuple) == args.length ||
```
adriaanm: How about directly comparing to the class with name `s"Tuple${args.length}"`?

adriaanm:
Very happy to see this! Looks like the best compromise we can get for now. Will work towards this through deprecation in 2.13 (and make it available under -Xsource:2.14)

odersky:
I dropped the case where we auto-tuple if the expected type is a tuple. Only two lines in the whole codebase broke. So this is clearly a feature that does not pull its own weight. This means the only
issue is operators, and indeed, we should get rid of the variable number of arguments for an operator instead. 

One additional measure to take: Warn if a symbolic method takes multiple parameters (since then there is no longer a way to use the method infix.

It would be good to start with this already for Scala 2.13. This also means that the variadic `+=` operators in the collections should be deprecated or dropped. /cc @julienrf @szeiger @lrytz

szeiger:
I agree with always tupling the RHS of an operator. It never made much sense to me that a single arg could be passed without parens but if you added parens it was an argument list, not a tuple.

We'd have to deprecate the variadic version of `Growable.+=` now. 

Another instance is `LongMap.+=`:

```scala
  def +=(key: Long, value: V): this.type = { update(key, value); this }
```

Semantics and call syntax are the same as for the inherited `+=` method (which takes a `Tuple2` in a `Map`) but this overload takes two primitive `Long` values instead of boxing them and wrapping them in a tuple. No longer being able to call it as an operator would cause a performance regression.

sjrd:
For `LongMap`:
```scala
@inline override final def +=(t: (Long, V)): this.type = {
  update(t._1, t._2)
  this
}
```
and let the optimizer get rid of it. If not scalac's, then the JVM's.

Additional benefit: also works for `m += x -> v` in addition to `m += (x, v)`.

szeiger:
Nice. I didn't think of using inlining to let the optimizer remove the boxing overhead.

odersky:
This is the second part of the cleanup, which turned out to be quite a bit harder than the first. In particular:

 - An infix operation `x op (y, z)` now unconditionally interprets
   `(y, z)` as a tuple, unless in Scala-2 mode.
 - In Scala-2 mode, the pattern will give a migration warning with
   option to rewrite in most cases. Excluded from rewrite are
   operator assignments and right-associative infix operations.
 - Auto-tupling is only enabled in `Scala-2` mode, and will give
   a migration warning with option to rewrite in that mode.

This change broke a lot more code than the auto-tupling change. As I feared, the pattern of infix operations that take multiple arguments on their right hand side is quite common. But I also think that in most cases the idiom is confusing and benefits from a rewrite to method syntax.

The other problem is that these patterns now mean something different (i.e. tuple on the rhs), which will likely lead to type errors later on, but it's also conceivable that they change behavior. The best remedy is to compiler under -language:Scala2 first. This will issue migration warnings for all 
problematic cases and rewrite most of them automatically.

SethTisue:
> `buf += ("hello", "world")`
> Good or not, it's pretty common.

I don't think it's common. I barely ever see it. I'd be happy to lose this.

odersky:
> In what codebase? In the compiler codebase? If so, perhaps it's an outlier?

Compiler, collections, testing framework, and tests themselves. The most common usage is something like

     xs and (x, y, z)

because the author decided that looks better than

    xs.and(x, y, z)

ritschwumm:
ouch, "An infix operation x op (y, z) now unconditionally interprets (y, z) as a tuple, unless in Scala-2 mode." will need _a lot_ of my code to be changed. especially in longer method chains like `a foo b bar c quux (c,d)` which would have to be sth like `a.foo(b).bar(c).quux(c,d)` or `(a foo b bar c).quux(c,d)` then - both of which are imho more difficult to read.

odersky:
@ritschwumm I would argue that `a.foo(b).bar(c).quux(c,d)` is a lot easier to read than the other two. But I am also concerned that we require changes in a lot of places here.

odersky:
languageserver has build issues now. @allanrenucci, @nicolasstucki could you take a look what needs to change?

nafg:
@ritschwumm the change doesn't have to be so invasive. You can write `(a foo b bar c).quux(c,d)`

In IntelliJ you can just Alt-Enter on `quux` and say "Convert from infix expression" and it will put the necessary parentheses in the right place automatically.

@odersky is there any sense in distinguishing between symbolic infix methods and alphanumeric infix methods? For instance a less breaking change would be to only not interpret parentheses as an argument list for symbolic infix methods. Of course it would be more elegant not to distinguish.

Also there would be the equivalent of a deprecation cycle, as was done for procedure syntax. `-language:Scala2` counts, and @adriaanm already said there would be on the scalac side as well.

I just wonder if it makes any sense at all to deprecate parentheses as an argument list for symbolic infix methods earlier (or stricter in some way) than for alphanumeric infix methods.

Another point here is whether you can call it "auto-untupling" if the primary meaning of the syntax is a single tuple argument, but the method is defined to take separate parameters instead.

Duhemm:
This is caused by sbt-buildinfo which generates the configuration for the language server in tests. It generates code which looks like
```scala
override val toString: String = "..." format (v1, v2, v3)
```

Which compiles but will throw an exception at runtime. It would be great to have this kind of issues caught by the compiler though; this is going to be painful to migrate.

See sbt/sbt-buildinfo#124

nafg:
I guess it could warn if despite accepting a single tuple, n arguments
would be valid too.
Also, that code might be changed to use the f interpolater.

Atry:
I actually LOVE the auto tuple feature, though not the current Scala auto tuple behavior.

I wonder if it is possible to not distinguish tuple from multiple parameters in the language (e.g. have a `type Function2[A, B, R] = Function[Tuple2[A, B], R]` definition), then it would be magically optimized to double parameter function at runtime.

nafg:
The main problem you have to solve is with method calls, not functions.

IIRC Odersky wanted to do that since the beginning of Scala but it didn't
end up being feasible.

What do you mean that you love it but not the current behavior?

Atry:
I imagine a language whose typer supports only one parameter. All calls to multi-parameter methods creates an temporary tuple. It's the optimizer or backend's duty to eliminate the tuple. This optimization should be a little simpler than uncurrying because no partially applied function involved.

odersky:
@Atry Yes, I agree that would be nice. My fear is that it would mess up overloading resolution sufficiently to be unworkable. To start with, the Joda time example would still have the surprising behavior. But I suspect there would be many more surprising corner cases.

odersky:
@nafg I believe adding a distinction of behavior between symbolic and alphanumeric operators would make matters worse. Besides, in `(x, y) :: xs` we _want_ `(x, y)` to be treated as a pair, not as two parameters.

nafg:
Well, that's a symbolic method, so that's not in question. Infix syntax on
alpha methods is a greyer area whichever way you look at it. I agree that
having more distinctions is less elegant, even though we would be able to
break less code if the change wouldn't affect alpha methods.

Atry:
@odersky Another example is `shapeless.Poly`, which use `HList` as the parameter list. However, as a library, it lacks of the ability to optimize HList into multi-parameters.

nafg:
On the other hand, it eliminates half the point of tuples -- zero-ceremony
products.

odersky:
I believe there are two possible local optima here:

 - Keep things as they are for Dotty, except that we don't want to prompt auto-tupling if the expected type is `Unit` (this was an oversight).

 - Go full in, and drop both auto-tupling and multi-parameter infix operations. This is much cleaner but will have significant migration pain.

Should this be a SIP, or do we want to discuss it here. I don't have the bandwidth right now to prepare a formal SIP, but maybe somebody else can take over, if we decide to go down that route.

odersky:
The consensus with Adriaan is that we first want to deprecate multi-parameter infix operations in `scalac`. Suggested rewrite for a symbolic operator with more than one parameter is to take a tuple instead.  Once that is done, the migration to Scala 3 should be easy enough that we can do it.

odersky:
SIP meeting proposal: 

 - Deprecate definition of symbolic multi-parameter operators in 2.13
 - Deprecate use of multi-parameter infix operators in 2.14 and 3.0
 - Deprecate auto-tupling _except_ for multi-parameter infix operators in 3.0
 - Drop multi-parameter infix operators and auto-tupling in 3.1

Richtype:
@Ichoran I think what we really need is Structs for compound deep value classes. On Native they could be just passed in and out of methods as their atomic component parameters on the stack, on the Jvm and Js runtime, they could still be passed as atomic component parameters, they would just have to be returned as a heap object. I don't think that would be a big problem as far more values are passed as parameters than are returned, particularly in recursive code.

Deep value Structs don't cover all cases, sometimes you want a reference field, but I'm sure they cover the bulk of the performance loss from unnecessary heap object creation and they're great because they don't stress the garbage collector. I don't think we should wait for Valhalla.

Ichoran:
@Richtype - That's one mechanism to try to reach zero-overhead, but my point is that right now we have exactly one way to have zero overhead, namely avoid the constructor and take the components in the method, and I want the zero-overhead problem solved before removing the feature that currently is the only solution (without turning math into not-math).

Ichoran:
Here is a complete benchmark runnable with Thyme that illustrates the problem.

```
val th = new ichi.bench.Thyme
val a = Array.fill(2000)(scala.util.Random.nextInt(1000))
case class V(x: Int, y: Int) { 
  def *(p: (Int, Int)): Int = p._1*x + p._2*x
  def *(x: Int, y: Int) = x*this.x + y*this.y
}
th.pbenchOff(){ 
  val v = new V(2, 7)
  var s, i = 0
  while (i < a.length) { s += v * (a(i), a(i+1)); i += 2 }
  s 
}{ 
  val v = new V(2, 7)
  var s, i = 0
  while (i < a.length) { s += v * ((a(i), a(i+1))); i += 2 }
  s
}
```

Running on my machine:

```
Welcome to Scala 2.12.7 (OpenJDK 64-Bit Server VM, Java 1.8.0_191).
Type in expressions for evaluation. Or try :help.
...

Benchmark comparison (in 659.1 ms)
Significantly different (p ~= 0)
  Time ratio:    2.47109   95% CI 2.44881 - 2.49337   (n=20)
    First     533.5 ns   95% CI 531.0 ns - 536.0 ns
    Second    1.318 us   95% CI 1.308 us - 1.328 us
res43: Int = 2027128
```

There's no workaround that preserves performance.

So the tupling solution isn't much of a solution presently.  You'd be better off just rewriting everything in `v.*(x, y)` format and deal with the (obvious) ugliness than (non-obviously) introduce a bunch of potential performance hits.

smarter:
@Ichoran Can you point to a library that uses this pattern ?

Ichoran:
@smarter - `mutable.AnyRefMap` and `mutable.LongMap` use it so you can `m += (k, v)` at full speed (put there by me, because I knew about the difference in speed and benchmarked it to be significant).  I use it in my own code in a bunch of places, for instance https://github.com/Ichoran/kse/blob/master/src/main/scala/maths/Vc.scala

I don't know if it's used in other Scala libraries.  Java uses the pattern, albeit without symbolic operators because it doesn't have any.  E.g. in awt.Point: https://docs.oracle.com/javase/8/docs/api/index.html?java/awt/Point.html

smarter:
For mutable maps at least you can do `m(k) = v` instead, right ?

ritschwumm:
i'm still not happy with this change, mostly because one of the reasons i fell in love with scala was the "no dots, no parens" style of writing code which to me is a _huge_ win in readability. i'm afraid with this change, it would become impossible to use in any halfway consistent way.

can we get tuples as proposed here, and still somehow save this style? ideally in a way that improves on what was possible before?

having to add parens for more than one parameter always struck me as inconsistent, same as having to break out of a nice and clean "subject verb object verb object" chain whenever one of the methods did not have a parameter list. for the latter, i'd have wished for "subject verb .. verb object" syntax (.. meaning "no parameter list here"). maybe something similar or better would work for mutiple parameters, too?

SethTisue:
meanwhile over in scala/scala, https://github.com/scala/scala/pull/7795

Ichoran:
Was anyone going to comment on the performance concerns?

eed3si9n:
@Ichoran This looks kind of interesting - https://github.com/lampepfl/dotty/pull/5200

```scala
  inline def sum(inline tup: Tuple2[Int, Int]): Int = ~Macros.tup2(tup)
```

It would be some more work on the library authors, but I think the overall change to limit infix to one argument is good tradeoff since the parameter list and tuple syntax looks the same.

Ichoran:
@eed3si9n - If the deprecation is going in now, the performance concerns should also be addressed now.  How bad are they, how widely used is the high-performance alternative, and will there be a replacement before the deprecated behavior is removed (so you can at least keep the old behavior if you're willing to tolerate the deprecation messages)?

Ichoran:
I guess it _is_ highly relevant for _this_ PR, but not to the linked one from which it was suggested to discuss the issue here.  I think that is actually the wrong decision; this PR shouldn't be cluttered up with decisions about 2.13.

som-snytt:
Scala2 lints all multiarg usages and multiparam definitions (for operators only), since the effort to deprecate (linked above) stalled (probably around migration or coordinating with Dotty):

https://github.com/scala/scala/pull/8951

No attempt to handle `op=` specially, and `+=` on `LongMap` is deprecated in favor of wordly method:

https://github.com/scala/scala/pull/9003

The lint has deprecation ergonomics, such as not linting deprecated things:
```
scala> class C { def ++ (i: Int, j: Int) = println(i+j) }
                     ^
       warning: multiarg infix syntax looks like a tuple and will be deprecated
class C

scala> @deprecated("Uses multiarg infix", since="forever") class C { def ++ (i: Int, j: Int) = println(i+j) }
class C

scala>
```
On the old forum thread, I mentioned `"" + ()`, which is the subject of a recent fix, which I just tried out :tada:, with the note, "In the PR, I rely on the spec language about multi-arg infix to assert that it doesn’t apply to empty args." So multi-arg infix was also the context for unit-value infix. On that forum post, I also added, "Before someone says it’s fixed in Dotty,...", but I'm afraid I can't edit it, now that it is fixed.

soronpo:
In my library I syntax that looks as follows:
```scala
  val foo = DFUInt(32) <> IN init (0, 1, 3, 5, 7)
```
This limitation completely breaks this syntax and forces it to be:
```scala
  val foo = DFUInt(32).<>(IN).init(0, 1, 3, 5, 7)
```
I don't understand the logic of this limitation within this context. IMO, where the arguments are not allowed to tuples anyway, there is no point limiting. If there can be an ambiguity, then (and only then) issue an error.

SethTisue:
Scala 2.13.3, to be released shortly, includes https://github.com/scala/scala/pull/8951 (by @som-snytt), which I guess is why Oron is hitting this now. We'll find out soon how many other library authors object. (The PR description there now links here.)

soronpo:
> 
> 
> Scala 2.13.3, to be released shortly, includes [scala/scala#8951](https://github.com/scala/scala/pull/8951) (by @som-snytt), which I guess is why Oron is hitting this now. We'll find out soon how many other library authors object. (The PR description there now links here.)

I would have opted for a public discussion first, and then deprecation later.

soronpo:
> 
> 
> In my library I have syntax that looks as follows:
> 
> ```scala
>   val foo = DFUInt(32) <> IN init (0, 1, 3, 5, 7)
> ```
> 
> This limitation completely breaks this syntax and forces it to be:
> 
> ```scala
>   val foo = DFUInt(32).<>(IN).init(0, 1, 3, 5, 7)
> ```
> 
> I don't understand the logic of this limitation within this context. IMO, where the arguments are not allowed to be tuples anyway, there is no point limiting. If there can be an ambiguity, then (and only then) issue an error.

I would say though that there is an alternative course for me which is to accept a single argument which can be a tuple and then using macro-magic to handle the rest. Still, seems unnecessary where the context can be better understood otherwise.

som-snytt:
@soronpo it's "just a lint" on scala 2. `-Xlint:-multiarg-infix` to turn it off.

soronpo:
> 
> 
> @soronpo it's "just a lint" on scala 2. `-Xlint:-multiarg-infix` to turn it off.

Although better, now every user of my library would require this flag, since the warning is at the use site. It would be one of those "put this in and not ask questions" kind of documentation for all its users. That is a no-go for me.

julienrf:
@soronpo Taking a single tuple parameter instead of several parameters is not a satisfactory solution for you?

henricook:
I'm really enjoying refactoring lots of scala test syntax as a result of this in Scala 2.13 :-/

e.g. `deltas.toSet should contain oneOf(` to `deltas.toSet should contain.oneOf(`

soronpo:
> 
> 
> (Edit [By @soronpo]: Sorry, I didn't notice I was editing this post. Reversing)
> 
> @soronpo Taking a single tuple parameter instead of several parameters is not a satisfactory solution for you?
> 
> That being said, to be honest when I read the following:
> 
> ```
>       val foo = DFUInt(32) <> IN init (0, 1, 3, 5, 7)
> ```
> 
> I find it very hard to parse (it is confusing what is an operator and what is an operand). I prefer:
> 
> ```
>       val foo = (DFUInt(32) <> IN).init(0, 1, 3, 5, 7)
> ```
> 
> (YMMV)

I'm sorry, I just noticed I edited your post by mistake. I edited again to restore your post.
Here is what I wrote in reply:


I understand what you mean, but you're currently reading it in parts, out of context, and not as a single statement. There are several legal variations to it:
```scala
val foo = DFUInt(32)
or
val foo = DFUInt(32) init 0
or
val foo = DFUInt(32) init (0, 1, 2, 3)
or
val foo = DFUInt(32) <> IN
or 
val foo = DFUInt(32) <> IN init 0
or
val foo = DFUInt(32) <> IN init (0, 1, 2, 3)
or
val foo = DFUInt(32) <> IN init (0, 1, 2, 3) !! SomeTag !! SomeOtherTag
or
<more tagged combinations like above>
```
Adding bracket/dots/whatever to be explicit about associativity undermines the flow and consistency I expect.
It would be like forcing you to use brackets on `protected def`. I use Scala as a DSL. So are many other users. And many dedicated APIs are essentially DSLs. I understand the purpose of being more opinionated in Scala 3.0, but I feel we might be blocking too much here.

godenji:
My table definitions look like:
```scala
def * = id ~ first ~ last <> (User.apply _, User.unapply _)
```
Forgetting the current spew of spurious (to me) warnings, removing this feature will result in -- well I'm not sure what as I haven't tried to fix the warnings -- presumably something like like:
```scala
def * = ((id ~ first ~ last) <>)(User.apply _, User.unapply _)
```
 which is more confusing than the original syntax. Is allowing this syntax really that terrible?

julienrf:
@godenji I think you will have to write this:

```scala
def * = (id ~ first ~ last).<>(User.apply _, User.unapply _)
```

smarter:
It seems that people are using this issue to complain about a new warning introduced in Scala 2.13.3. This warning is not present in Dotty itself as far as I can tell, so I don't really know why Scala 2.13's behavior was changed to be stricter than Dotty's, but this isn't the correct bug tracker to discuss this.

smarter:
Ah, it looks like that warning has been moved behind a lint flag in the as-yet-unreleased 2.13.4: https://github.com/scala/scala/pull/9089

som-snytt:
I created a topic on the forum. I did not attempt to summarize opinions here, but I did contribute the coinage "autountupling."

https://contributors.scala-lang.org/t/multiarg-infix-application-considered-warty/4490

odersky:
To be revived for 3.1

SethTisue:
> To be revived for 3.1

New ticket: https://github.com/lampepfl/dotty/discussions/19255

PR #4317, opened 2018-04-13
Experimental Streams using quotes

biboudis:
Port of the [strymonas](https://github.com/strymonas) library as described in O. Kiselyov et al., Stream fusion, to completeness (POPL 2017) using Dotty's staging facility.

Ideas:
- design a cleaner API, possibly investigate if it could be part of collections
- experiment with alternative code generation schemes, e.g., generate fused code run wtih OpenCL/CUDA bindings
- multidimensional types


tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -315,8 +315,95 @@ object Test {
...
+            // _2: the current element,
+            // _3: the step of the inner most steam
+            type St = (Var[Boolean], Var[A], Var[Unit => Unit])
+            val card: Cardinality = Many
+
+            def init(k: St => Expr[Unit]): Expr[Unit] = {
+              producer.init(st => '{
+                Var('{ (_: Unit) => ()}){ advf => {
+                  Var('{ true }) { hasNext => {
+                    // Var('{ null.asInstanceOf[A] }) { curr => {
```
biboudis: I would like to generate the following line:

`var curr: A = null.asInstanceOf[A]` 

so I guess the only way is:

`Var('{ null.asInstanceOf[A] }) { curr => ...`.

The minimal example that I made working is:

```
class Foo[A] {
  def meth[A: Type](a: Type[A]) = Var[A, Unit]('(null.asInstanceOf[~a])) { curr => '() }
}
```
Why do I need evidence for `A`? I didn't need one for the example at: https://github.com/lampepfl/dotty/blob/master/tests/run-with-compiler/quote-type-tags.scala#L8

If that is the solution indeed, how to avoid the extra parameter in `makeAdvanceFunction`?

biboudis: Ok, it can be implicit. Now I have a way not to pollute the API.

biboudis:
```scala
  trait Bar { def meth(): Unit }
  class Foo[T: Type] {
    new Bar {
      def meth() = {
        Var('(null.asInstanceOf[T])) { curr => '() }
      }
    }
  }
```
```
Var('(null.asInstanceOf[T])) { curr => '() }
   |                                ^
   |                            access to Foo.this from wrong staging level:
   |                             - the definition is at level 0,
   |                             - but the access is at level 1.
```

biboudis:
Note to `self`: the call to `adv` is not generated correctly in the nested case. Have to investigate.

biboudis:
@nicolasstucki parallel tests failed. I will restart the build.

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+
+    /** zip **/
+    def zip[B: Type, C: Type](f: (Expr[A] => Expr[B] => Expr[C]), stream2: Stream[B]): Stream[C] = {
+      val Stream(stream_b) = stream2
+      Stream(mapRaw[(Expr[A], Expr[B]), Expr[C]]((t => k => '{ ~k(f(t._1)(t._2)) }), zipRaw[A, Expr[B]](stream, stream_b)))
+    }
+  }
+
+  object Stream {
+    def of[A: Type](arr: Expr[Array[A]]): Stream[A] = {
```
nicolasstucki: Try changing it to 
```scala
def of[A: Type: Liftable](arr0: Array[A]): Stream[A] = {
  val arr = arr0.toExpr
  ...
```

You will need some lifter for Arrays from the library.

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+              ~t._1.update('{~t._1.get - 1})
+              ~k(el)
+            }), addTerminationCondition(b => '{ ~t._1.get > 0 && ~b}, nested.nestedf(t._2)))
+          })
+        }
+      }
+     }
+
+    /** A stream containing the first `n` elements of this stream. */
+    def take(n: Expr[Int]): Stream[A] = Stream(takeRaw[Expr[A]](n, stream))
```
nicolasstucki:
```scala
def take(n: Int): Stream[A] = Stream(takeRaw[Expr[A]](n.toExpr, stream))
```

biboudis:
I have the changes you proposed ready but I was stuck in something during the refactoring. I will post here, maybe discuss and merge.

Thanks @nicolasstucki

tests/run-with-compiler-custom-args/staged-streams_1.scala
julienrf: @nicolasstucki what’s the advantage of your proposal?

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+      *          .map{ case (a, b) => a + b }
+      *          .fold(0)((a, b) => a + b)
+      * }}}
+      *
+      * -->
+      *
+      * {{{
+      *           /* initialization for stream 1 */
+      *
+      *           var curr = null.asInstanceOf[Int]  // keeps each element from reified stream
```
julienrf: This should probably be `0` instead of `null`. Also, it seems that `curr` is never updated.

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+            // _2: the current element,
+            // _3: the step of the inner most steam
+            type St = (Var[Boolean], Var[A], Var[Unit => Unit])
+            val card: Cardinality = Many
+
+            def init(k: St => Expr[Unit]): Expr[Unit] = {
+              producer.init(st =>
+                Var('{ (_: Unit) => ()}){ nadv => {
+                  Var('{ true }) { hasNext => {
+                     Var('{ null.asInstanceOf[A] }) { curr => '{
```
julienrf: I’m wondering if this can cause subtle bugs, since `null.asInstanceOf[Int]` produces `0`, which is undistinguishable from the value `0` (by contrast with references, where `null` can always be distinguished from an object reference).

Why don’t you use `Option[A]` instead?

tests/run-with-compiler-custom-args/staged-streams_1.scala
nicolasstucki: The aim is to hide the `Expr` from the end user.

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+      * @param z   the accumulator
+      * @param f   the zipping function
+      * @tparam W  the type of the accumulator
+      * @return
+      */
+     def fold[W: Type](z: Expr[W], f: ((Expr[W], Expr[A]) => Expr[W])): Expr[W] = {
+       Var(z) { s: Var[W] => '{
+           ~foldRaw[Expr[A]]((a: Expr[A]) => '{
+             ~s.update(f(s.get, a))
+           }, stream)
```
julienrf: Is there any difference between `'{ ~s.update(…) }` and `s.update(…)`?

tests/run-with-compiler-custom-args/staged-streams_1.scala
julienrf: Then do we want `fold` to have the following signature?

~~~ scala
def fold[W : Type](z: W, f: (W, A) => W): W
~~~

I’m not sure this is what we want?

nicolasstucki: You are right, it is not exactly what I want. We need to hide the Expr behind a macro. I will test my signature idea and then document it here if it works

tests/run-with-compiler-custom-args/staged-streams_1.scala
nicolasstucki: It is the same

nicolasstucki: Though I would not change it. This kind of small additions have been useful to find bugs in the implementation. I would keep it like this for as a regression test.

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,697 @@
...
+
+           ~s.get
+         }
+       }
+     }
+
+     private def foldRaw[A](consumer: A => Expr[Unit], stream: StagedStream[A]): Expr[Unit] = {
```
julienrf: It seems that we don’t need to take `stream` as a parameter since it is already a case class parameter. (btw name shadowing is dangerous!)

tests/run-with-compiler-custom-args/staged-streams_1.scala
biboudis: 👍 

I was "updating" it in comments. Fixing it!

tests/run-with-compiler-custom-args/staged-streams_1.scala
biboudis: However, nice catch @julienrf!

tests/run-with-compiler-custom-args/staged-streams_1.scala
```diff
@@ -0,0 +1,700 @@
...
+          new Producer[Expr[A]] {
+            // _1: if the stream has ended,
+            // _2: the current element,
+            // _3: the step of the inner most steam
+            type St = (Var[Boolean], Var[A], Var[Unit => Unit])
+            val card: Cardinality = Many
+
+            def init(k: St => Expr[Unit]): Expr[Unit] = {
+              producer.init(st =>
+                Var('{ (_: Unit) => ()}){ nadv => {
```
biboudis: @nicolasstucki this is the line that triggers the exception after the last rebase

nicolasstucki: Fix in #4592

LPTK:
@biboudis very cool! Has this implementation been benchmarked? Does it expand full stream pipelines into low-level loops, as in the paper?

biboudis:
Thanks @LPTK. Yes, by inspection, the code generates the expected imperative low-level loops. If needed we will benchmark it in the future. For now this is merely a big test case for us.

PR #4319, opened 2018-04-13
[Experiment, PR for benchmarking] Try to reduce cache misses

smarter:
- Name#derived is costly because it involves getting and setting value
  in a Map for caching, we don't need this caching for fresh names.
- NameOps#moduleClassName is costly because it calls Name#derived
- Tree#denot was megamorphic


smarter:
test performance please

smarter:
Doesn't seem to move the needle, I'll investigate cache misses more later.

PR #4326, opened 2018-04-16
Refine dependency extraction: Ignore packages

allanrenucci:
We can ignore packages when collecting dependencies for incremental
compilation. We make sure not to ignore dependencies on package object.

### DIFF
--- a/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
+++ b/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
@@ -315,7 +315,7 @@ private class ExtractDependenciesCollector extends tpd.TreeTraverser { thisTreeT
     !sym.exists ||
     sym.unforcedIsAbsent || // ignore dependencies that have a symbol but do not exist.
                             // e.g. java.lang.Object companion object
-    sym.is(PackageClass) ||
+    sym.is(Package) ||
     sym.isEffectiveRoot ||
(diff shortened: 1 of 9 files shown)


allanrenucci:
Zinc only ignores packages for class dependencies but not for used names. This PR also ignores packages as used names. Not sure it is safe

allanrenucci:
@smarter What's your take on this? Should we add used name for packages like Zinc does?

smarter:
I guess you need to because of package objects? If you import something from a package, and if the package object changes, then that may affect you

allanrenucci:
I have added a test case that is failing whether or not we add used names for packages. The test also fails in Zinc (see sbt/zinc#530)

compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala
```diff
@@ -295,8 +295,12 @@ private class ExtractDependenciesCollector extends tpd.TreeTraverser { thisTreeT
       val fromClass = resolveDependencySource
       if (fromClass.exists) { // can happen when visiting imports
         assert(fromClass.isClass)
-        _dependencies += ClassDependency(fromClass, enclOrModuleClass, DependencyByMemberRef)
+
         addUsedName(fromClass, mangledName(sym), UseScope.Default)
+        // packages have class symbol. Only record them as used names but bot dependency
```
smarter: typo: bot -> not

PR #4329, opened 2018-04-17
Fix comment in isAsSpecificValue type and add test

odersky:
Fix comment in isAsSpecificValue to match the code, drop a probably spurious case from
the code which referred to the old encoding of applied types in terms of refinements,
and add a test that exercises some of the behavior.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -1142,15 +1142,15 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
      *
      *    flip(T) <: flip(U)
      *
-     *  where `flip` changes top-level contravariant type aliases to covariant ones.
-     *  Intuitively `<:s` means subtyping `<:`, except that all top-level arguments
+     *  where `flip` changes covariant occurrences of contravariant type parameters to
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1167,7 +1167,6 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
       else {
         val flip = new TypeMap {
           def apply(t: Type) = t match {
-            case t: TypeBounds => t
```
smarter: Is this PR just about changing comments or also about changing behavior?

milessabin: Is the dropping of "top-level" intentional? I'm not sure I see that change reflected in the code.

odersky: @smarter @milessabin The PR essentially adapts the comment to what the code now does. Dropping the `TypeBounds` case should have no effect except for wildcard arguments. When we still used refinements for type parameters it did have the effect of restricting to top-level. But with the change to direct parameter encoding that effect was lost, so that change caused a change in behavior.

I am actually quite unsure what the right behavior would be. Toplevel or not? What does toplevel mean, precisely. I don't have a good testcase that would demonstrate why restricting to toplevel is important. So I opted for the solution that's simpler to specify.

PR #4333, opened 2018-04-17
 Fix #4314: revert changes in #4299

### LINKED ISSUE #4314: Unreachable case in two-level sealed hierarchy is not detected
Given the following code:
```scala
sealed trait Foo[A]
case class One[A]() extends Foo[A]
sealed trait Bar[A] extends Foo[A]
case class Two() extends Bar[String]

object Test {
  def test(x: Foo[Int]) = x match {
    case One() =>
  }
}
```
The exhaustiveness checker is smart enough to not warn, because no instance of `Bar` is a subtype of `Foo[Int]`.

However, if we add a case for `Bar()` like this:
```scala
sealed trait Foo[A]
case class One[A]() extends Foo[A]
sealed abstract case class Bar[A]() extends Foo[A]
class Two() extends Bar[String]

object Test {
  def test(x: Foo[Int]) = x match {
    case One() =>
    case Bar() =>
  }
}
```
We should get an unreachable code warning (and indeed, scalac produces one), but we don't.

liufengyun:
Fix #4314

1. revert #4299: ApproximatingTypeMap produces Nothing, which is not what is needed.
2. Avoid blind erasure which lose information about the pattern, see tests/patmat/i4314b.scala

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -356,7 +356,11 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
     val map = new TypeMap {
       def apply(tp: Type) = tp match {
-        case tref: TypeRef if isPatternTypeSymbol(tref.typeSymbol) => WildcardType(tref.underlying.bounds)
+        case tp @ AppliedType(tycon, args) if tycon.isRef(defn.ArrayClass) =>
+          // walkaround `Array[_] <:< Array[Array[_]]`, see tests/patmat/t2425.scala
+          tp.derivedAppliedType(tycon, args.map(mapOver))
+        case tref: TypeRef if isPatternTypeSymbol(tref.typeSymbol) =>
+          WildcardType(tref.underlying.bounds)
         case _ => mapOver(tp)
       }
     }
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -113,6 +113,7 @@ trait SpaceLogic {
     case Prod(tp, fun, sym, spaces, full) =>
       val sp = Prod(tp, fun, sym, spaces.map(simplify(_)), full)
       if (sp.params.contains(Empty)) Empty
+      else if (canDecompose(tp) && decompose(tp).isEmpty) Empty
       else sp
     case Or(spaces) =>
       val set = spaces.map(simplify(_)).flatMap {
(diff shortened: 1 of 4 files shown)


liufengyun:
test performance with #fast please

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -356,7 +356,11 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
     val map = new TypeMap {
       def apply(tp: Type) = tp match {
-        case tref: TypeRef if isPatternTypeSymbol(tref.typeSymbol) => WildcardType(tref.underlying.bounds)
+        case tp @ AppliedType(tycon, args) if tycon.isRef(defn.ArrayClass) =>
+          // walkaround `Array[_] <:< Array[Array[_]]`, see tests/patmat/t2425.scala
```
smarter: I don't think this comment is true, the following code doesn't compile:
```scala
def foo(x: Array[_]): Array[Array[_]] = x
```

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -349,18 +350,22 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-      tp.derivedRefinedType(erase(parent), refinedName, WildcardType)
-    case _ => tp
+  /* Erase pattern bound types with WildcardType */
+  def erase(tp: Type) = {
+    def isPatternTypeSymbol(sym: Symbol) = !sym.isClass && sym.is(Case)
+
+    val map = new TypeMap {
+      def apply(tp: Type) = tp match {
+        case tp @ AppliedType(tycon, args) if tycon.isRef(defn.ArrayClass) =>
+          // walkaround `Array[_] <:< Array[Array[_]]`, see tests/patmat/t2425.scala
```
Blaisorblade: `s/walkaround/work around/`

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
liufengyun: The underscore above expands to TypeBounds `Array[<: Any :> Nothing]` instead of Wildcards. 

Here is the subtyping log:

```
  ==> Array[(?)] <:< Array[Array[(?)]]
    ==> scala.type <:< scala.type
    <== scala.type <:< scala.type   = true
    ==> Array[(?)] <:< (?)
      ==> Any <:< Any LoApprox
      <== Any <:< Any LoApprox  = true
    <== Array[(?)] <:< (?)   = true
    ==> (?) <:< Array[(?)]
    <== (?) <:< Array[(?)]   = true
  <== Array[(?)] <:< Array[Array[(?)]]   = true
```

And the behavior can be confirmed from the code in TypeComparer.

But thanks for raising the question, as by using TypeBounds instead WildcardType, we avoid the problem.

smarter: Yes, `WildcardType(lo, hi) <:< X` is equivalent to `lo <:< X`, see TypeComparer

PR #4334, opened 2018-04-17
Fix #4315: handle local child from scalac

### LINKED ISSUE #4315: Spurious unreachable/only null warning
```scala
import scala.concurrent.duration.{Duration, FiniteDuration}

class Test {
  def test(d: Duration) = d match {
    case finite: FiniteDuration =>
    case d =>
  }
}
```
```scala
-- [E120] Only null matched Warning: tests/allan/Test.scala:7:9 ----------------
7 |    case d =>
  |         ^
  |         Only null is matched. Consider using `case null =>` instead.
one warning found
```

liufengyun:
Fix #4315: handle local child from scalac

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -497,6 +497,15 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   def refine(parent: Type, child: Symbol): Type = {
     if (child.isTerm && child.is(Case, butNot = Module)) return child.termRef // enum vals always match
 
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -497,6 +497,8 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   def refine(parent: Type, child: Symbol): Type = {
     if (child.isTerm && child.is(Case, butNot = Module)) return child.termRef // enum vals always match
 
+    if (child.name == tpnme.LOCAL_CHILD) return child.typeRef
```
allanrenucci: Can you add a comment explaning

PR #4335, opened 2018-04-18
Optimize scala2Mode

odersky:
Profile showed that `scala2Mode` takes about 1% of compute time when run on dotc/typer/*.scala.
We can reduce this by caching previously computed results.


nicolasstucki:
test performance please

PR #4337, opened 2018-04-18
Mark failing test as pending

allanrenucci:
The test used to pass on case insensitive file system. This also make sure that no file is overwritten.


smarter:
How is this test related to case sensitivity?

allanrenucci:
https://github.com/lampepfl/dotty/blob/776e4c4021cab4728ff58d82b6f995cff18b4a5e/sbt-dotty/sbt-test/source-dependencies/packageobject-and-traits/test#L2

This line used to overwrite `Foo.scala` on case insensitive file system, and would make the test pass by triggering recompilation. My understanding of the test is that `foo.scala` and `Foo.scala` were supposed to co-exist

allanrenucci:
Regression introduced in #4331 which failed on the CI but not on macOS

allanrenucci:
That's what I did in this PR. I changed the test to make it fail on case insensitive FS and marked it as pending.

PR #4344, opened 2018-04-19
Fix #4342: Adapt outer types when calling into Scala 2 classes

### LINKED ISSUE #4342: ExplicitOuter doesn't match Scala 2.12's outer types
Scala 2.12 appears to use the declared self type of the enclosing class as the outer type when transforming inner classes. Dotty doesn't match this behavior and uses the enclosing class itself, leading to `NoSuchMethodErrors` wrt. Scala 2.12 classes upon executing the Dotty-compiled code. Here's a minimal example:

Consider the following code, compiled with 2.12:
```scala
package foo

class A extends B

trait B { this: A =>
  class C
}

object a extends A
```
and the following code, compiled with Dotty:
```scala
object Scala2_12FromDotty {
  new foo.a.C  // runtime error: trying to call constructor of C with a B instead of an A
}
```
We can see the discrepancy in the generated code. That is, `javap -p 'target/scala-2.12/classes/foo/B$C.class'` yields:
```
Compiled from "Foo.scala"
public class foo.B$C {
  public final foo.A $outer;
  public foo.A foo$B$C$$$outer();
  public foo.B$C(foo.A);
}

```
whereas `javap -c Scala2_12FromDotty$.class` yields
```
Compiled from "Scala2_12FromDotty.scala"
public final class Scala2_12FromDotty$ {
  (...)
  public Scala2_12FromDotty$();
    Code:
      (...)
      12: getstatic     #25                 // Field foo/a$.MODULE$:Lfoo/a$;
      15: invokespecial #28                 // Method foo/B$C."<init>":(Lfoo/B;)V
      (...)
}
```

I tried to look into the related phases a bit, but I'm not sure whether this should be fixed in `ExplicitOuter` or by transforming infos, applications and defdefs later, for instance, as an extension of `LinkScala2Impls`.

gsps:
See #4342 for some context.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
@@ -141,6 +141,20 @@ object ExplicitOuter {
   private def newOuterAccessors(cls: ClassSymbol)(implicit ctx: Context) =
     newOuterAccessor(cls, cls) :: (if (cls is Trait) Nil else newOuterParamAccessor(cls) :: Nil)
 
+  /** Scala 2.x and Dotty don't always agree on what should be the type of the outer parameter,
+   *  so we replicate the old behavior when passing arguments to methods coming from Scala 2.x.
+   */
+  private def outerClass(cls: ClassSymbol)(implicit ctx: Context): Symbol = {
+    val encl = cls.owner.enclosingClass
+    if (cls.is(Scala2x))
+      encl.asClass.classInfo.selfInfo match {
+        case tp: TypeRef => tp.classSymbol
+        case self: Symbol => self
+        case _ => encl
+      }
+    else encl
+  }
+
   /** A new outer accessor or param accessor.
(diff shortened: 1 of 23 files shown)


compiler/src/dotty/tools/dotc/transform/ExplicitOuter.scala
```diff
@@ -336,12 +336,23 @@ object ExplicitOuter {
   class OuterOps(val ictx: Context) extends AnyVal {
     private implicit def ctx: Context = ictx
 
+    private def outerType(cls: ClassSymbol): Type = {
```
allanrenucci: Can you add a comment explaining why this is needed

gsps:
Sure, there you go!

smarter:
I think all these other changes come from commits which are already in master, the PR just needs a rebase.

odersky:
OK @gsps Can you do the rebase then we can get this in quickly.

smarter:
Rebased PR opened here: https://github.com/lampepfl/dotty/pull/5801

PR #4351, opened 2018-04-20
Fix #4323: enter class type parameters in GADT bounds

### LINKED ISSUE #4323: GADT patmat doesn't narrow class type parameters
```scala
scala> enum Expr[T] {
           case IExpr(value: Int) extends Expr[Int]
           case BExpr(value: Boolean) extends Expr[Boolean]

           def join(other: Expr[T]): Expr[T] = (this, other) match {
             case (IExpr(i1), IExpr(i2)) => IExpr(i1 + i2)
             case (BExpr(b1), BExpr(b2)) => BExpr(b1 & b2)
           }
         }
6 |      case (IExpr(i1), IExpr(i2)) => IExpr(i1 + i2)
  |                                     ^^^^^^^^^^^^^^
  |                                     found:    Expr[Int]
  |                                     required: Expr[T]
  |
7 |      case (BExpr(b1), BExpr(b2)) => BExpr(b1 & b2)
  |                                     ^^^^^^^^^^^^^^
  |                                     found:    Expr[Boolean]
  |                                     required: Expr[T]
  |
```

This is similar to an old issue where class/trait parameters aren't narrowed.

liufengyun:
Fix #4323: enter class type parameters in GADT bounds


liufengyun:
test performance please

tests/pos/i4323.scala
```diff
@@ -0,0 +1,9 @@
+enum Expr[T] {
+  case IExpr(value: Int) extends Expr[Int]
+  case BExpr(value: Boolean) extends Expr[Boolean]
+
+  def join(other: Expr[T]): Expr[T] = (this, other) match {
```
allanrenucci: Add a neg test where `this.T` is not `other.T`:
```scala
def join[A](other: Expr[A]): Expr[T] = ...
```

smarter:
It'd be good to figure out why a test had to be moved to pos-deep-subtype, this often masks real bugs.

Blaisorblade:
In fact the _standard library_ had to be moved to -Yno-deep-subtypes, which seems even more concerning? We already know GADT bounds can create subtyping loops (#4069), so I'm not sure how robust this can possibly be.

liufengyun:
Close for now, maybe @AleksanderBG can have a look at the GADTMap refactoring.

PR #4359, opened 2018-04-22
Some baby steps towards an effect system

odersky:
So far:

 - define effect capabilities as types with separate bottom type `Impure` and top type `Pure`.
 - define some global capabilities in an `Effect` object.
 - ensure that effects cannot be mixed with normal types in type bounds, unions, and intersections.
 - disallow capture of effect values in lambdas and nested objects.
 - eliminate all effect capabilities before code generation.

Still missing is the tricky part: how effects interact with functions


liufengyun:
A very nice PR 👍 

I see in the code there is a definition of `canBeImpure`, I'm wondering what is the concept of _purity_ under consideration.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -595,7 +595,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
               case _ => false
             }
             val sym1 = tp1.symbol
-            (sym1 eq NothingClass) && tp2.isValueTypeOrLambda ||
+            (sym1 eq NothingClass) && tp2.isValueTypeOrLambda && !tp2.isEffect ||
             (sym1 eq NullClass) && isNullable(tp2)
```
nicolasstucki: Probably missing `(sym1 eq ImpureClass) && tp2.isEffect`

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -256,6 +256,17 @@ object Types {
       case _ => false
     }
 
+    /** Is this an effect type? */
+    def isEffect(implicit ctx: Context): Boolean = this match {
+      case tp: TypeRef => tp.stableInRunSymbol.is(Effect)
+      case tp: TypeProxy => tp.superType.isEffect
+      case tp: MethodType => tp.resType.isEffect
+      case tp: PolyType => tp.resType.isEffect
+      case AndType(tp1, tp2) => tp1.isEffect && tp2.isEffect
```
nicolasstucki: Given that we check that both types are effects in an `And` type, we only need to ask `tp1.isEffect`. Same for `Or`.

tests/pos/effects-pure.scala
```diff
@@ -0,0 +1,69 @@
...
+      (ex: NSE) => default
+    }
+
+  def summon[E <: Pure](implicit eff: E): Unit = ()
+
+  type EFF <: Pure
+
+  { import Effect.{canThrow, canBoth}
+    implicit val eff: EFF = Effect.isImpure
+    summon[CanThrow[AssertionError] eff_& CanThrow[NullPointerException]]
```
nicolasstucki: Can't we write `summon[CanThrow[AssertionError] & CanThrow[NullPointerException]]` and define `canBoth` as
`def canBoth[A <: Pure, B <: Pure](implicit a: A, b: B): A & B = _`?

nicolasstucki:
Ported test cases from Phantom types. Some of the fail with the current implementation.

Blaisorblade:
Apart from other questions, why is this not using `erased`?

nicolasstucki:
The current implementation is doing the checks in lambdaLift which is performed after erasure.

nicolasstucki:
I propose a different approach to check captures that could be done before erasure. Which would allow us to reuse `erased`. This is based on a [prototype I made on phantom types](https://github.com/nicolasstucki/dotty/blob/implement-non-capturing-methods/compiler/src/dotty/tools/dotc/transform/CheckCaptures.scala).

The idea is to have a miniphase that will find all references of effect type. On each one of those we would traverse the `ctx.owner.ownersIterator` until we reach the owner of the reference or a package. If durring the traversal of the owners we traverse a closure or a class, we know that this reference has been captured.

arnolddevos:
Is mutation one of the effects to be represented?  If so, the obvious starting point is a capability to assign and reference var's and Array elements.  Say _CanMutate_.  

That divides the world into pure and impure w.r.t mutation. At least as far as type signatures are concerned.

But it would be nice to label other common cases with types. The essence is the extent of sharing of mutable data.   Consider a method or constructor that:-

1. Mutates local vars or local mutable objects excluding parameters. 

This is a pure method or constructor as seen from the caller using mutation internally.  A very common case, for example, in scala.collection.immutable.  The _CanMutate_ capability should not appear in this method's signature.

2. Mutates members var's or _unshared_ mutable members. 

Here unshared means that no other object or closure contains the same mutable object.  A difficult thing, I know. 

Nevertheless, a well behaved data structure that can be safely used in case (1) provides such methods. Examples are the scala.collection.mutable collections.

3. Mutates parameters. 

The object to which this method belongs, or the eta-expanded method itself, can be taken as a reified mutation. Since mutation is limited to the parameters a pure, associative combinator can be defined. (It is a Monad.)

It would be nice if capabilities based on phantom types could cover such cases. As a start, it might need something like a compiler-generated capability to distinguish when a local or a parameter is the target of a method. 

I previously worked out an idea for immutable types and found that these cases fell out naturally. https://arnolddevos.github.io/2018/03/26/Scala_Effects.html  In a nutshell, there is a marker trait for immutable types and rules to prevent immutable values incorporating mutable values.  (Similar to the rules about capturing phantom types.)

Blaisorblade:
@arnolddevos Good question. That'd be nice, but it's not easy. @liufengyun might have some ideas — and some of what's good below came up in chats with him (errors are mine).

Safely detecting when an object is unshared seems a big challenge. Currently, nothing forbids a constructor from saving a pointer to the object being built in some global variable, so even after `val foo = new Bar()` you don't know that `foo` is unshared (unless you inspect `Bar`'s constructor, but that violates separate compilation). If `Bar` has no external side effects, I suspect `foo` is unshared; but that is much more restrictive than needed.

IIUC (and I could be speaking *lots* of nonsense so corrections welcome), Rust's type system tracks aliasing and sharing, so it could ensure that an object is unshared. If that's correct, it seems we're still in trouble: the annotation overhead for programmers is not trivial, and Scala allocated its complexity budget elsewhere. Maybe this use case has simpler requirements and a simpler solution works, but that's far from obvious.

arnolddevos:
@Blaisorblade Thanks for relaying those thoughts.

odersky:
Leaving open since we will hopefully be able to get back to this at some point.

odersky:
To be revived at some later time

PR #4360, opened 2018-04-23
Clean up kindedness tests

odersky:
We currently mostly use isLambdaSub which detects unapplied
abstract and alias types but not unapplied generic classes.
We should use `hasHigherKind` more often, which includes
unapplied classes.

The first commit fixes a problem with `hasHigherKind` where
abstract types upper bounded by `AnyKind` where not classified
as higher-kinded.

It also changes checkSimpleKinded to use hasHigherKind instead
if isLambdaSub.

I thought it would be good to get this in before I proceed with effects, since
effects effectively introduce another kind.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PostTyper.scala
@@ -213,7 +213,10 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
           }
         case tree: TypeApply =>
           val tree1 @ TypeApply(fn, args) = normalizeTypeArgs(tree)
-          Checking.checkBounds(args, fn.tpe.widen.asInstanceOf[PolyType])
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/transform/PostTyper.scala
```diff
@@ -213,7 +213,10 @@ class PostTyper extends MacroTransform with IdentityDenotTransformer { thisPhase
           }
         case tree: TypeApply =>
           val tree1 @ TypeApply(fn, args) = normalizeTypeArgs(tree)
-          Checking.checkBounds(args, fn.tpe.widen.asInstanceOf[PolyType])
+          if (fn.symbol != defn.ChildAnnot.primaryConstructor) {
```
smarter: Instead of special-casing Child, the compiler could be changed to always use it with fully-applied types with wildcards, e.g. `Child[Cons[_]]`

PR #4363, opened 2018-04-24
Fix #4350: Do not check prefix of class type parameters

### LINKED ISSUE #4350: Wrong staging level when accessing type from class
When:
```
class F[T: Type] { '(null.asInstanceOf[T]) }
```

access to T from the class F is reported as a wrong level:
```
4 |  '(null.asInstanceOf[T])
  |                      ^
  |                      access to F.this from wrong staging level:
  |                       - the definition is at level 0,
  |                       - but the access is at level 1.
```

SImilarly:
```
class F {
  type T
  '(null.asInstanceOf[T])
}
```

nicolasstucki:
TODO
- [x] Fix infinite recursion in neg test
- [x] Fix substitution of type

### DIFF
--- a/compiler/src/dotty/tools/dotc/quoted/TreeCleaner.scala
+++ b/compiler/src/dotty/tools/dotc/quoted/TreeCleaner.scala
@@ -17,12 +17,14 @@ class TreeCleaner extends tpd.TreeMap {
   /** List of symbols and their types for type aliases `type T = U` */
   private[this] var aliasesSyms: List[Symbol] = Nil
   private[this] var aliasesTypes: List[Type] = Nil
+  private[this] val aliases = newMutableSymbolMap[Tree]
 
   override def transform(tree: Tree)(implicit ctx: Context): Tree = {
     val tree0 = tree match {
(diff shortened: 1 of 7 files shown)


biboudis:
@nicolasstucki it works for me (apart from the infinite recursion if someone omits the evidence). 

The other error we were getting was indeed a bug in the code.

compiler/src/dotty/tools/dotc/quoted/TreeCleaner.scala
```diff
@@ -17,12 +17,14 @@ class TreeCleaner extends tpd.TreeMap {
   /** List of symbols and their types for type aliases `type T = U` */
   private[this] var aliasesSyms: List[Symbol] = Nil
   private[this] var aliasesTypes: List[Type] = Nil
+  private[this] val aliases = newMutableSymbolMap[Tree]
```
biboudis: The changes in this file are not relevant to the fix right? BTW can we add a comment for the `aliases` var just for symmetry? 😛

PR #4379, opened 2018-04-25
Fix test reporting

allanrenucci:
When running the tests, we were reporting the number of errors
encountered instead of the number of failed tests. For neg tests, errors
are not test failures

### DIFF
--- a/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
+++ b/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
@@ -253,6 +253,9 @@ trait ParallelTesting extends RunnerOrchestration { self =>
     /** A set of the different failures */
     def failureReasons: Set[Failure] = _failures
 
+    /** Number of failed tests */
+    def failureCount: Int = _failures.size
+
     protected def logBuildInstructions(reporter: TestReporter, testSource: TestSource, err: Int, war: Int) = {
       val errorMsg = testSource.buildInstructions(reporter.errorCount, reporter.warningCount)
       addFailureInstruction(errorMsg)
@@ -295,7 +298,7 @@ trait ParallelTesting extends RunnerOrchestration { self =>
           "[" + ("=" * (math.max(progress - 1, 0))) +
             (if (progress > 0) ">" else "") +
             (" " * (39 - progress)) +
-            s"] completed ($tCompiled/$sourceCount, $errorCount failed, ${timestamp}s)\r"
+            s"] completed ($tCompiled/$sourceCount, $failureCount failed, ${timestamp}s)\r"
         )
 
         Thread.sleep(100)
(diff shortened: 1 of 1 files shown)


Blaisorblade:
`errorCount` seems a trap, and `reasonsForFailure` seems to fall into it _even if it's also called for neg tests by `checkExpectedErrors`_.
`Neg test shouldn't have failed, but did. Reasons:  - encountered ${test.errorCount} error(s)`.

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -1010,8 +1009,8 @@ trait ParallelTesting extends RunnerOrchestration { self =>
     /** Extract `Failure` set and render from `Test` */
     private[this] def reasonsForFailure(test: Test): String = {
       val errors =
-        if (test.errorCount == 0) ""
-        else s"\n  - encountered ${test.errorCount} error(s)"
+        if (test.failureCount == 0) ""
+        else s"\n  - encountered ${test.failureCount} error(s)"
```
allanrenucci: I would change `error(s)` for `failure(s)`

allanrenucci:
@Blaisorblade I am not sure about the second part of this PR. I think `errorCount` is useful is some cases. E.g. compiling a neg file and failing because the compiler emits 2 errors instead of 3.

I would suggest to open a new PR with your changes

Blaisorblade:
> I am not sure about the second part of this PR. I think errorCount is useful is some cases. E.g. compiling a neg file and failing because the compiler emits 2 errors instead of 3.

That still works, and it's still visible in the meta test, in the output line
```
Wrong number of errors encountered when compiling out/VulpixMetaTests/neg/missing-error-annotation, expected: 0, actual: 2
```

The code you talk about is not affected — https://github.com/lampepfl/dotty/pull/4379/commits/565a2e53a7b37cadf2c95f91c474ff698dda2eab changes some other behavior, the rest is purely meaning-preserving dead code elimination.

The behavior you describe involves _another_ field named `errorCount` — see https://github.com/dotty-staging/dotty/blob/29f47602ee0d94bc0c89deab126dad32950316c0/compiler/test/dotty/tools/vulpix/ParallelTesting.scala#L754-L759 for the code involving the other `errorCount`, and see https://github.com/dotty-staging/dotty/blob/29f47602ee0d94bc0c89deab126dad32950316c0/compiler/test/dotty/tools/vulpix/ParallelTesting.scala#L714 to see where `// error` appears to be actually handled.

FWIW, I don't have a problem with splitting the other commits _if there is plausible reason at all_, this just isn't one.

allanrenucci:
> FWIW, I don't have a problem with splitting the other commits if there is plausible reason at all, this just isn't one

The reason is that it's safe to merge the first commit and I'd like to experiment with the other commits to convince myself that it is indeed dead code

Blaisorblade:
The current PR was tested in http://dotty-ci.epfl.ch/lampepfl/dotty/4656, so merging.

Issue #4372, opened 2018-04-25
StackOverflow


alexknvl:
```scala
class X[A >: X[_ <: X[_]] <: X[A]]
```

```scala
Exception in thread "main" java.lang.StackOverflowError
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3940)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:156)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3978)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:156)
	at dotty.tools.dotc.core.Types$TypeMap.mapArgs$1(Types.scala:3959)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3968)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:156)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3980)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:156)
	at dotty.tools.dotc.core.Types$TypeMap.mapArgs$1(Types.scala:3959)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3968)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:156)
	at dotty.tools.dotc.core.Types$TypeMap.$anonfun$mapOver$5(Types.scala:3942)
	at dotty.tools.dotc.core.Types$VariantTraversal.atVariance(Types.scala:3888)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3942)
	at dotty.tools.dotc.core.TypeComparer$$anon$1.apply(TypeComparer.scala:154)
	at dotty.tools.dotc.core.TypeComparer.monitoredIsSubType$1(TypeComparer.scala:159)
	at dotty.tools.dotc.core.TypeComparer.$anonfun$recur$2(TypeComparer.scala:853)
	at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:12)
	at dotty.tools.dotc.reporting.trace$.op1$3(trace.scala:32)
	at dotty.tools.dotc.reporting.trace$.apply(trace.scala:33)
[...]

Issue #4371, opened 2018-04-25
StackOverflow in TypeApplications


alexknvl:
Probably related to https://github.com/lampepfl/dotty/issues/4368, but stack trace is different.

```scala
object App {
  class Foo { type A = Boo#B }
  class Boo { type B = Foo#A }
}
```

```scala
Exception in thread "main" java.lang.StackOverflowError
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:173)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:192)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:178)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:192)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:178)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:192)
	at dotty.tools.dotc.util.Stats$.track(Stats.scala:35)
	at dotty.tools.dotc.core.TypeApplications$.typeParams$extension(TypeApplications.scala:171)
	at dotty.tools.dotc.core.TypeApplications$.$anonfun$typeParams$extension$1(TypeApplications.scala:178)
[...]

Issue #4369, opened 2018-04-25
Wrong lower type bound is accepted


alexknvl:
```scala
trait X[R <: Z, Z >: X[R, R]] {
  // def foo(x: X[R, R]): Z = x
}
class Z extends X[Z, Z]
```
Somehow this compiles, even though `foo` doesn't. Note that `class Z` is not a super type of `X[Z, Z]`, which is required by the second type parameter of `trait X`.

```scala
trait X[R <: Z, Z >: X[R, R] <: X[R, R]] {
  // implicitly[Z =:= X[R, R]] // won't compile
}
class Z extends X[Z, Z]
```

alexknvl:
This might be related to https://github.com/lampepfl/dotty/issues/4368 since
```scala
trait X { self =>
  type R <: Z
  type Z >: X { type R = self.R; type Z = self.R }
}
class Foo extends X { type R = Foo; type Z = Foo }
```
causes stack overflow.

Issue #318, opened 2014-12-21
Infinite loop in subtyping check


samuelgruetter:
While looking for counterexamples to narrowing in DOT, I found this example:

``` scala
object subtypcycle {
  trait Y {
    type A <: { type T >: B }
    type B >: { type T >: A }
  }

  val y: Y = ???
  val a: y.A = ???
  val b: y.B = a
}
```

Which sends dotty into an infinite loop checking if `y.A <: y.B`.

`scalac` has the same problem, see [SI-9056](https://issues.scala-lang.org/browse/SI-9056).

newca12:
FYI this is not (or not anymore) an infinite loop
After a certain amount of time (< 1min) dotc terminate with a Type Mismatch Error

 ```
-- [E007] Type Mismatch Error: <console> -----------------------------
9 |  val b: y.B = a
  |               ^
  |               found:    subtypcycle.y.A(subtypcycle.a)
  |               required: subtypcycle.y.B
  |

 ```

newca12:
The fix for #2397 in PR #2402 is a regression an turn #318 in an infinite loop again.
This example found by @samuelgruetter is referenced as a pending test : https://github.com/lampepfl/dotty/blob/master/tests/pending/pos/subtypcycle.scala

felixmulder:
Hi @newca12 - thanks for reporting and pointing us to the pending file. We'll try to get to this during the week!

Cheers,
Felix

nicolasstucki:
The infinite loop is still there.

newca12:
Since dotty 0.6.0-RC1, this now fail immediately with a stack overflow.
<details>
 <summary>Stack trace</summary>
Exception in thread "main" java.lang.StackOverflowError
	at java.util.Arrays.copyOfRange(Arrays.java:3484)
	at java.util.Arrays.copyOfRange(Arrays.java:3441)
	at java.lang.invoke.MethodType.insertParameterTypes(MethodType.java:396)
	at java.lang.invoke.MemberName.getInvocationType(MemberName.java:168)
	at java.lang.invoke.DirectMethodHandle.preparedLambdaForm(DirectMethodHandle.java:167)
	at java.lang.invoke.DirectMethodHandle.make(DirectMethodHandle.java:81)
	at java.lang.invoke.MethodHandles$Lookup.getDirectMethodCommon(MethodHandles.java:1660)
[...]

PR #4385, opened 2018-04-26
Fix #4368: Avoid stackoverflows for some cyclic definitions

### LINKED ISSUE #4368: StackOverflow in typer
```scala
trait X {
  type A = B
  type B
}
trait Y {
  type A
  type B = A
}
trait Z extends X with Y
```

```scala
Exception in thread "main" java.lang.StackOverflowError
	at java.lang.invoke.MemberName.getInvocationType(MemberName.java:164)
	at java.lang.invoke.DirectMethodHandle.preparedLambdaForm(DirectMethodHandle.java:167)
	at java.lang.invoke.DirectMethodHandle.make(DirectMethodHandle.java:89)
	at java.lang.invoke.MethodHandles$Lookup.getDirectMethodCommon(MethodHandles.java:1660)
	at java.lang.invoke.MethodHandles$Lookup.getDirectMethodNoSecurityManager(MethodHandles.java:1617)
	at java.lang.invoke.MethodHandles$Lookup.getDirectMethodForConstant(MethodHandles.java:1802)
	at java.lang.invoke.MethodHandles$Lookup.linkMethodHandleConstant(MethodHandles.java:1751)
	at java.lang.invoke.MethodHandleNatives.linkMethodHandleConstant(MethodHandleNatives.java:477)
	at dotty.tools.dotc.core.Types$Type.findMember(Types.scala:661)
	at dotty.tools.dotc.core.Types$Type.memberExcluding(Types.scala:491)
	at dotty.tools.dotc.core.Types$Type.$anonfun$nonPrivateMember$1(Types.scala:482)
	at dotty.tools.dotc.core.Types$Type.nonPrivateMember(Types.scala:482)
	at dotty.tools.dotc.core.Types$NamedType.memberDenot(Types.scala:1777)
	at dotty.tools.dotc.core.Types$NamedType.reload$1(Types.scala:2025)
	at dotty.tools.dotc.core.Types$NamedType.withPrefix(Types.scala:2037)
	at dotty.tools.dotc.core.Types$NamedType.derivedSelect(Types.scala:1997)
	at dotty.tools.dotc.core.Types$TypeMap.derivedSelect(Types.scala:3902)
	at dotty.tools.dotc.core.Types$ApproximatingTypeMap.derivedSelect(Types.scala:4174)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.$anonfun$apply$4(TypeOps.scala:61)
	at dotty.tools.dotc.reporting.trace$.op1$1(trace.scala:18)
	at dotty.tools.dotc.reporting.trace$.conditionally(trace.scala:20)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.apply(TypeOps.scala:54)
	at dotty.tools.dotc.core.Types$TypeMap.$anonfun$mapOver$7(Types.scala:3974)
	at dotty.tools.dotc.core.Types$VariantTraversal.atVariance(Types.scala:3888)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3974)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.$anonfun$apply$4(TypeOps.scala:67)
	at dotty.tools.dotc.reporting.trace$.op1$1(trace.scala:18)
	at dotty.tools.dotc.reporting.trace$.conditionally(trace.scala:20)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.apply(TypeOps.scala:54)
	at dotty.tools.dotc.core.TypeOps.asSeenFrom(TypeOps.scala:28)
	at dotty.tools.dotc.core.TypeOps.asSeenFrom$(TypeOps.scala:27)
	at dotty.tools.dotc.core.Contexts$Context.asSeenFrom(Contexts.scala:71)
	at dotty.tools.dotc.core.Types$Type.$anonfun$asSeenFrom$1(Types.scala:767)
	at dotty.tools.dotc.core.Types$Type.asSeenFrom(Types.scala:766)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.computeAsSeenFrom(Denotations.scala:1075)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.computeAsSeenFrom(Denotations.scala:660)
	at dotty.tools.dotc.core.Denotations$PreDenotation.asSeenFrom(Denotations.scala:131)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.mapInherited(Denotations.scala:1062)
	at dotty.tools.dotc.core.Denotations$SingleDenotation.mapInherited(Denotations.scala:660)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.collect$1(SymDenotations.scala:1605)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.collect$1(SymDenotations.scala:1600)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.collect$1(SymDenotations.scala:1600)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.$anonfun$computeNPMembersNamed$1(SymDenotations.scala:1613)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.computeNPMembersNamed(SymDenotations.scala:1592)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.nonPrivateMembersNamed(SymDenotations.scala:1582)
	at dotty.tools.dotc.core.SymDenotations$ClassDenotation.findMember(SymDenotations.scala:1617)
	at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:509)
	at dotty.tools.dotc.core.Types$Type.goThis$1(Types.scala:613)
	at dotty.tools.dotc.core.Types$Type.go$1(Types.scala:522)
	at dotty.tools.dotc.core.Types$Type.findMember(Types.scala:658)
	at dotty.tools.dotc.core.Types$Type.memberExcluding(Types.scala:491)
	at dotty.tools.dotc.core.Types$Type.$anonfun$nonPrivateMember$1(Types.scala:482)
	at dotty.tools.dotc.core.Types$Type.nonPrivateMember(Types.scala:482)
	at dotty.tools.dotc.core.Types$NamedType.memberDenot(Types.scala:1777)
	at dotty.tools.dotc.core.Types$NamedType.reload$1(Types.scala:2025)
	at dotty.tools.dotc.core.Types$NamedType.withPrefix(Types.scala:2037)
	at dotty.tools.dotc.core.Types$NamedType.derivedSelect(Types.scala:1997)
	at dotty.tools.dotc.core.Types$TypeMap.derivedSelect(Types.scala:3902)
	at dotty.tools.dotc.core.Types$ApproximatingTypeMap.derivedSelect(Types.scala:4174)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.$anonfun$apply$4(TypeOps.scala:61)
	at dotty.tools.dotc.reporting.trace$.op1$1(trace.scala:18)
	at dotty.tools.dotc.reporting.trace$.conditionally(trace.scala:20)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.apply(TypeOps.scala:54)
	at dotty.tools.dotc.core.Types$TypeMap.$anonfun$mapOver$7(Types.scala:3974)
	at dotty.tools.dotc.core.Types$VariantTraversal.atVariance(Types.scala:3888)
	at dotty.tools.dotc.core.Types$TypeMap.mapOver(Types.scala:3974)
	at dotty.tools.dotc.core.TypeOps$AsSeenFromMap.$anonfun$apply$4(TypeOps.scala:67)
	at dotty.tools.dotc.reporting.trace$.op1$1(trace.scala:18)
	at dotty.tools.dotc.reporting.trace$.conditionally(trace.scala:20)
.......
```

odersky:
We use a two-pronged approach:

First, we now detect some cycles that span inherited or intersected classes or traits. Unfortunately, this operation forces the types of type members, which means it can lead to spurious cycles itself. So we can do this only after a lot of types are already computed, which means some cycles lead to stackoverflows before they are detected.

Hence the second strategy: give more useful information when deep recursions happen.


odersky:
test performance please

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -647,19 +645,14 @@ object Types {
...
         ctx.pendingMemberSearches = name :: ctx.pendingMemberSearches
-      }
       ctx.findMemberCount = recCount + 1
-      //assert(ctx.findMemberCount < 20)
       try go(this)
       catch {
         case ex: Throwable =>
-          core.println(i"findMember exception for $this member $name, pre = $pre")
-          throw ex // DEBUG
+          core.println(s"findMember exception for $this member $name, pre = $pre, recCount = $recCount")
```
Blaisorblade: Nit: why change formatter to `s`?

odersky:
Yes, #318 is handled, albeit crudely.

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: Avoid possible crashes when trying to show `pre`.

odersky:
test performance please

odersky:
So, this is a rather sneaky way to fix a whole class of problems that cause the compiler to stackoverflow in some way or another. One would hope there was a principled way to detect and repair all illegal cyclic references before they cause stackoverflows. Unfortunately, that looks impossible or at least impossibly hard: Scala's type system subsumes F_sub and is therefore undecidable. So some infinite loops/stackoverflows are unavoidable. What's more, checking for cycles early has the unfortunate consequence that we force types and thereby can cause spurious cyclic references! So the mere act of checking generates spurious errors. This is a real problem in practice, observed in both the compiler codebase and stdlib. Another possibility would be to attach a cycle detection component to every operation that might stackoverflow. But these tend to be also operations that are most time-critical, so doing the extra work is prohibitive.

The idea in this pull request is to act only if there _is_ a stackoverflow. In that case we unroll the stack until we hit a safepoint (typically: a point where we typecheck an AST node and therefore have an error position to record). While unrolling we keep a count of all important method frames that were unrolled. Important means: the method contains a try/catch that does the unrolling and counting. At the safepoint, we check which method appeared most often in the unrolled frames. We then compose a nice error message showing only calls info for that method and leaving out the middle of the stack if it is too large. A typical error message looks like this:

    |  val b: y.B = a 
    |         ^
    |         Recursion limit exceeded.
    |         Maybe there is an illegal cyclic reference?
    |         A recurring operation is (inner to outer):
    |         
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           ...
    |         
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B
    |           find-member y.A
    |           find-member y.B

Of course, it's still nicer to pinpoint the cycle where it arises, and we try to do this also, as long as it can be done safely and efficiently. But we have transformed the game from a bug that needs fixing, (a stackoverflow is a compiler crash, so it's wrong and embarrassing) to a quality of error message question. Such a question can be answered on a case-by-case basis, taking into account the frequency of the error (in my experience most of these cyclic reference errors are synthetic and don't tend to occur in practice) and the effort needed to get a more specific error message.

odersky:
test performance please

odersky:
This PR fixes #4368, #4369, #4370, #4371, #4372 and #318.

odersky:
test performance please

odersky:
I won't invest more time in this. So, ready to get this in?

Blaisorblade:
Hm. Especially since `RecursionOverflow` appears in the error, it seems worth renaming that (and friends) as `RecursionOverflowException`.

But otherwise, it seems reasonable to classify those as requests for enhancements. I'll have to look at the first more closely, because the output contradicts quite a few of my expectations and one of the two must be corrected.

smarter:
I think it'd be good to have a -Y setting to turn off the exception catching, so that the root cause of issues can be investigated more easily.

odersky:
@smarter The recommended way to find out more is to turn the `cyclic` printer on.

smarter:
@odersky The issue is that by turning everything into errors, @alexknvl fuzzer will never detect any stackoverflow again, even though some of them might be legitimate bugs we can fix.

Blaisorblade:
Status: I'd like to wait for #4511 before enabling this by default, lest this PR swallows errors. But that involves reviewing positions in quite a few scenarios.

Given the amount of code and potential for divergence, it could be best to make this conditional as planned, default to false and merge this PR, then flip the default in #4511.

Blaisorblade:
@smarter: can you take a look? This can now be disabled, and the remaining issues here (in comments above) don't seem blockers, so I think this should be mergeable.

Blaisorblade:
Or @odersky can you review my changes? Whoever's faster.

compiler/src/dotty/tools/dotc/config/ScalaSettings.scala
```diff
@@ -146,6 +146,7 @@ class ScalaSettings extends Settings.SettingGroup {
   val Xlink = BooleanSetting("-Xlink", "Recompile library code with the application.")
   val YoptPhases = PhasesSetting("-Yopt-phases", "Restrict the optimisation phases to execute under -optimise.")
   val YoptFuel = IntSetting("-Yopt-fuel", "Maximum number of optimisations performed under -optimise.", -1)
+  val YnoDecodeStacktraces = BooleanSetting("-Yno-decode-stacktraces", "Show operations that triggered StackOverflows instead of printing the raw stacktraces.")
```
allanrenucci: Comment is misleading. How I understand it: "if you pass the flag, you won't see original stack traces". I guess it is the opposite

Blaisorblade: Oh right, fixing it

PR #4386, opened 2018-04-26
Fix RCs publishing

allanrenucci:


smarter:
You'll need to make a new release of sbt-dotty too.

PR #4401, opened 2018-04-28
Fixes for fuzzing tests reported on Gitter

odersky:


odersky:
This was quite enjoyable, trying to fix fuzzing problems in real time, as they are reported. But I won't be able to spend more time on this. So, reviewers, if you notice something, it would be good if you could fix it yourself and push on top of this PR.

odersky:
This is best reviewed commit by commit.

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -1004,7 +1004,7 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   def sameTypes(trees: List[tpd.Tree], trees1: List[tpd.Tree]): Boolean = {
     if (trees.isEmpty) trees.isEmpty
     else if (trees1.isEmpty) trees.isEmpty
-    else (trees.head.tpe eq trees1.head.tpe) && sameTypes(trees.tail, trees1.tail)
+    else (trees.head.typeOpt eq trees1.head.typeOpt) && sameTypes(trees.tail, trees1.tail)
```
smarter: I think I have a better fix for that one: https://github.com/lampepfl/dotty/pull/4402

odersky: @smarter I agree. Can you push #4402 on top of this and revert my corresponding commit? Thanks!

odersky: Never mind, I can do it myself.

odersky: I see you have already started doing this, so please go ahead. Otherwise I could have also approved yours and done the rest myself.

odersky:
Some existing neg tests give more errors now that we check arguments of erroneous functions anyway. So these error counts need to be updated. Otherwise the last two commits LGTM.

PR #4402, opened 2018-04-28
Avoid constructing typed tree with untyped arguments

smarter:


smarter:
Closing since I'll push these changes in #4401 directly.

Issue #4389, opened 2018-04-26
A variety of bugs (42 or so).


alexknvl:
All files were generated using https://github.com/alexknvl/fuzzball

https://files.gitter.im/lampepfl/dotty/dWPj/exc-files.zip

There could be some duplicates. Most snippets have syntax errors in them. Feel free to create separate issues for individual bugs and reference them here, I'll update the list.

```
IN 3e54d491b47b3dc8034b4d16f4389554a09b94ba.scala
IN e7bb3c4598709966b31dbec096e11229922d4eb7.scala
IN b11c241ce88c83c0c6f949b3106d4e7ff9ef3d99.scala
IN 87b1e375168a7888470eefc1fb867d0c9f550865.scala
IN 0f12f42d878dcd4cb8b8ab62dbfb6b41861989b1.scala
scala.MatchError: This(Ident()) (of class dotty.tools.dotc.ast.Trees$This)
  at dotty.tools.dotc.typer.Namer$ClassCompleter.parentType$1(Namer.scala:910)
  at dotty.tools.dotc.typer.Namer$ClassCompleter.checkedParentType$1(Namer.scala:928)
  at dotty.tools.dotc.typer.Namer$ClassCompleter.$anonfun$completeInCreationContext$5(Namer.scala:981)
  at scala.collection.immutable.List.map(List.scala:283)
--
IN dea0cf7fd832a6f39963ddda6ffd89b336d18808.scala
IN 49fc58e3885b671a7569326d1ad6590f9668a1a2.scala
IN 44131e7b1cdc0192d39471ad01e2045991ec48b4.scala
IN eff46343d34261c03ec1223a254a3d39587f8d96.scala
java.lang.ClassCastException: dotty.tools.dotc.core.Types$CachedTypeRef cannot be cast to dotty.tools.dotc.core.Types$TermRef
  at dotty.tools.dotc.core.Types$TermRef$.apply(Types.scala:2150)
  at dotty.tools.dotc.core.SymDenotations$SymDenotation.termRef(SymDenotations.scala:1167)
  at dotty.tools.dotc.typer.TypeAssigner.assignType(TypeAssigner.scala:536)
  at dotty.tools.dotc.typer.TypeAssigner.assignType$(TypeAssigner.scala:535)
--
IN 37e5de7d6a0fa8cc9123fd570d9fd82d1fc5767f.scala
IN 36e924e6d37279c86a4420f3a80810e419bd31dd.scala
[...]

PR #4411, opened 2018-04-29
Fix more problems detected by fuzzing in #4389

odersky:
Fix more problems detected by fuzzing in #4389

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -88,9 +88,8 @@ object desugar {
             else {
               def msg =
                 s"no matching symbol for ${tp.symbol.showLocated} in ${defctx.owner} / ${defctx.effectiveScope.toList}"
-              if (ctx.reporter.errorsReported) ErrorType(msg)
-              else throw new java.lang.Error(msg)
-            }
+			  ErrorType(msg).assertingErrorsReported(msg)
+			}
           case _ =>
             mapOver(tp)
         }
--- a/compiler/src/dotty/tools/dotc/ast/Trees.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Trees.scala
@@ -769,28 +769,26 @@ object Trees {
   def genericEmptyTree[T >: Untyped]: Thicket[T]        = theEmptyTree.asInstanceOf[Thicket[T]]
 
   def flatten[T >: Untyped](trees: List[Tree[T]]): List[Tree[T]] = {
-    var buf: ListBuffer[Tree[T]] = null
-    var xs = trees
-    while (!xs.isEmpty) {
-      xs.head match {
-        case Thicket(elems) =>
-          if (buf == null) {
-            buf = new ListBuffer
-            var ys = trees
-            while (ys ne xs) {
-              buf += ys.head
-              ys = ys.tail
+    def recur(buf: ListBuffer[Tree[T]], remaining: List[Tree[T]]): ListBuffer[Tree[T]] =
+      remaining match {
+        case Thicket(elems) :: remaining1 =>
+          var buf1 = buf
+          if (buf1 == null) {
+            buf1 = new ListBuffer[Tree[T]]
+            var scanned = trees
+            while (scanned `ne` remaining) {
+              buf1 += scanned.head
+              scanned = scanned.tail
             }
           }
-          for (elem <- elems) {
-            assert(!elem.isInstanceOf[Thicket[_]])
-            buf += elem
-          }
-        case tree =>
+          recur(recur(buf1, elems), remaining1)
+        case tree :: remaining1 =>
           if (buf != null) buf += tree
+          recur(buf, remaining1)
+        case nil =>
+          buf
       }
-      xs = xs.tail
-    }
+    val buf = recur(null, trees)
     if (buf != null) buf.toList else trees
   }
 
--- a/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymDenotations.scala
@@ -1466,16 +1466,11 @@ object SymDenotations {
       if (classParents.isEmpty && !emptyParentsExpected)
         onBehalf.signalProvisional()
       val builder = new BaseDataBuilder
-      for (p <- classParents) {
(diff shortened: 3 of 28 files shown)


compiler/src/dotty/tools/dotc/ast/Trees.scala
```diff
@@ -769,28 +769,26 @@ object Trees {
...
-            while (ys ne xs) {
-              buf += ys.head
-              ys = ys.tail
+            val buf = new ListBuffer[Tree[T]]
+            var scanned = trees
+            while (scanned `ne` remaining) {
+              buf += scanned.head
+              scanned = scanned.tail
             }
+            recur(buf, remaining)
```
smarter: at this point `remaining = Thicket(elems) :: remaining1` so isn't this call equivalent to the one in the else branch below?

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1447,7 +1447,11 @@ class Typer extends Namer
     assignType(cpy.TypeDef(tdef)(name, rhs1), sym)
   }
 
-  def typedClassDef(cdef: untpd.TypeDef, cls: ClassSymbol)(implicit ctx: Context) = track("typedClassDef") {
+  def typedClassDef(cdef: untpd.TypeDef, cls: ClassSymbol)(implicit ctx: Context): Tree = track("typedClassDef") {
+    if (!cls.info.isInstanceOf[ClassInfo]) {
+      assert(ctx.reporter.errorsReported)
+      return cdef.withType(UnspecifiedErrorType)
```
smarter: I think ideally we should have a way to recursively set the type of the tree children to maintain the invariant that a typed tree only contains typed trees.

compiler/src/dotty/tools/dotc/ast/Trees.scala
odersky: Indeed it is.

compiler/src/dotty/tools/dotc/typer/Typer.scala
odersky: Good point, yes.

odersky: Actually, that's not so easy. Sometimes, there's simply no way to proceed. And we cannot just assign an ErrorType, because a lot of trees are not typed trees, so we'd have to replace them with something else.

odersky:
Rebased to master

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3478,7 +3478,7 @@ object Types {
       if (selfTypeCache == null)
         selfTypeCache = {
           val given = cls.givenSelfType
-          if (!given.exists) appliedRef
+          if (!given.isValueType) appliedRef
```
smarter: Might be worth explicitly checking that an error happened like other commits in this PR do, e.g.:
```scala
if (!given.isValueType) {
  assert(!given.exists || ctx.reporter.errorsReported)
  appliedRef
}
```

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
```diff
@@ -472,16 +472,18 @@ object ProtoTypes {
...
+  }
   /** The result type of `mt`, where all references to parameters of `mt` are
    *  replaced by either wildcards (if typevarsMissContext) or TypeParamRefs.
    */
   def resultTypeApprox(mt: MethodType)(implicit ctx: Context): Type =
     if (mt.isResultDependent) {
       def replacement(tp: Type) =
-        if (ctx.mode.is(Mode.TypevarsMissContext)) WildcardType else newDepTypeVar(tp)
+        if (ctx.mode.is(Mode.TypevarsMissContext) ||
+            !tp.widenExpr.isValueTypeOrWildcard) WildcardType
```
smarter: Here too, the error recovery case could be more explicit:
```scala
if (!tp.widenExpr.isValueTypeOrWildcard) {
  assert(ctx.reporter.errorsReported)
  WildcardType
}
else if (ctx.mode.is(Mode.TypevarsMissContext)
  WildcardType
else ...
```

compiler/src/dotty/tools/dotc/core/Types.scala
odersky: The given type might be higher-kinded (in fact that's what it was in the test) but this would be detected only later.

compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
odersky: Same response: I am not sure the error would always be detected beforehand.

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -1466,16 +1466,11 @@ object SymDenotations {
...
         onBehalf.signalProvisional()
       val builder = new BaseDataBuilder
-      for (p <- classParents) {
-      	var pcls = p.typeSymbol
-      	if (!pcls.isClass) pcls = p.underlyingClassRef(refinementOK = false).typeSymbol
-      		// This roundabout way is necessary for avoiding cyclic references.
-      		// A test case is CompilationTests.compileMixed
-      	pcls match {
+      for (p <- classParents)
+       	p.classSymbol match {
```
smarter: incorrect indentation.

compiler/src/dotty/tools/dotc/core/Types.scala
smarter: OK, in that case maybe just add a comment explaining that non-value types self can only happen with erroneous code? Otherwise readers of this code might be puzzled.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -88,9 +88,8 @@ object desugar {
             else {
               def msg =
                 s"no matching symbol for ${tp.symbol.showLocated} in ${defctx.owner} / ${defctx.effectiveScope.toList}"
-              if (ctx.reporter.errorsReported) ErrorType(msg)
-              else throw new java.lang.Error(msg)
-            }
+			  ErrorType(msg).assertingErrorsReported(msg)
```
smarter: Wrong indentation.

PR #4417, opened 2018-04-29
Fix #4390: Sort REPL autocomplete suggestions

### LINKED ISSUE #4390: REPL completion output is not sorted
image

ankitson:
This seems to be a quick fix for https://github.com/lampepfl/dotty/issues/4390. Its my first PR to dotty so I might be missing something.

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -219,7 +219,7 @@ object Interactive {
       case _  => getScopeCompletions(ctx)
     }
     interactiv.println(i"completion with pos = $pos, prefix = $prefix, termOnly = $termOnly, typeOnly = $typeOnly = ${completions.toList}%, %")
-    (completionPos, completions.toList)
+    (completionPos, completions.toList.sortBy(_.name.show))
   }
 
   /** Possible completions of members of `prefix` which are accessible when called inside `boundary` */
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -218,8 +218,10 @@ object Interactive {
       case (sel @ Select(qual, _)) :: _ => getMemberCompletions(qual)
       case _  => getScopeCompletions(ctx)
     }
-    interactiv.println(i"completion with pos = $pos, prefix = $prefix, termOnly = $termOnly, typeOnly = $typeOnly = ${completions.toList}%, %")
-    (completionPos, completions.toList.sortBy(_.name.toString))
+
+    val sortedCompletions = completions.toList.sortBy(_.name: Name)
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -219,7 +219,7 @@ object Interactive {
       case _  => getScopeCompletions(ctx)
     }
     interactiv.println(i"completion with pos = $pos, prefix = $prefix, termOnly = $termOnly, typeOnly = $typeOnly = ${completions.toList}%, %")
-    (completionPos, completions.toList)
+    (completionPos, completions.toList.sortBy(_.name.show))
```
allanrenucci: You can use `_.name.toString`

ankitson:
Done!

compiler/test/dotty/tools/repl/TabcompleteTests.scala
```diff
@@ -65,4 +65,14 @@ class TabcompleteTests extends ReplTest {
       List("\"", ")", "'", "¨", "£", ":", ",", ";", "@", "}", "[", "]")
         .foreach(src => assertTrue(tabComplete(src).suggestions.isEmpty))
     }
+
+  @Test def sortedCompletions: Unit =
+    fromInitialState { implicit state  =>
+      val src = """class Foo { def comp1 = 1; def compa = 3; def compA = 4 }"""
```
allanrenucci: Use simple quotes. Triple quotes are not needed here.

allanrenucci: I prefer the code snippet I gave you. I think the ordering is clearer:
```scala
class Foo { def comp3 = 3; def comp1 = 1; def comp2 = 2 }
```
```scala
val expected = List("comp1", "comp2", "comp3")
```

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -219,7 +219,7 @@ object Interactive {
       case _  => getScopeCompletions(ctx)
     }
     interactiv.println(i"completion with pos = $pos, prefix = $prefix, termOnly = $termOnly, typeOnly = $typeOnly = ${completions.toList}%, %")
-    (completionPos, completions.toList)
+    (completionPos, completions.toList.sortBy(_.name.toString))
```
allanrenucci: @smarter We have an ordering for `Name` that differs from the default `String` ordering. I am wondering if we should use it here? The Name ordering will not group type names and term names together

smarter: I think they should be grouped together, but we should also not display type completions if we're not in a type context

smarter: I think they should be grouped together, but we should also not display type completions if we're not in a type context

allanrenucci: Sure but that's probably out of the scope of this PR

PR #4422, opened 2018-04-30
Fix #4373: reject wildcard types in syntactically invalid positions

Blaisorblade:
The testcase is from examples in #4373, variations and tests for types with wildcards. I'm unsure for a couple cases.

The strategy I use comes from generalizing existing code.

A separate commit avoids cascade errors. That was the trickiest part of this PR, but pretty instructive. I now sometimes use `Ident(nme.ERROR)` instead of `scalaAny` to replace illegal wildcards in `extends` clauses.

Action items:
- [x] rebase
- [x] merge #4424 since it was approved
- [x] make `checkWildcard` private to Parsers again
- [x] go through #4389 and try all reported issues in Bugs.txt that contain a `_` somewhere.
- [x] test and fix `with` too (noticed from #4389)
- [x] add testcases with annotations

PR #4423, opened 2018-04-30
Fix #4382: Disallow wildcard arguments for higher-kinded types

### LINKED ISSUE #4382: Applications of abstract higher-kinded types to wildcard arguments aren't rejected
Carved out of #4375. The following code compiles, while I expect all the errors described—they're necessary if we want to _ensure_ people can't write existentials (and otherwise, not sure what the current check is for). Such an error is useful for people who might not realize their types mean something else now.

The check for unreducibility seems simply incomplete:
```scala
object App {
  type Id[A] >: A <: A
  def v1: Id[_] = ??? // should give error

  type HkL[A] >: A
  def v2: HkL[_] = ??? // should give error

  type HkU[A] <: A
  def v3: HkU[_] = ???  // should give error

  type HkAbs[A]
  def v4: HkAbs[_] = ??? // should give error
}
```

contrast with
```scala
scala> def v: _ = ???
1 |def v: _ = ???
  |       ^
  |       Unbound wildcard type
```

To demonstrate this is a problem, let's use a trait instead:

```scala
trait Foo {
  type HkP[A] >: (A, A) <: (A, A)
  def v1: HkP[_] = ??? // should give error

  type HkL[A] >: (A, A)
  def v2: HkL[_] = ??? // should give error

  type HkU[A] <: (A, A)
  def v3: HkU[_] = ???  // should give error

  type HkAbs[A]
  def v4: HkAbs[_] = ??? // should give error
}
object Bar extends Foo {
  type HkP[A] = (A, A)
  type HkL[A] = (A, A)
  type HkU[A] = (A, A)
  type HkAbs[A] = (A, A)
}
```

Then at least the types are approximates without crashes (here):

```scala
scala> def foo = Bar.v1
def foo: (_, _)
```

odersky:
This needed some follow-on work on how we encode `Eq` for enums, which could still be improved once we have typeclass derivation.


odersky:
@Blaisorblade Between #4423 and #4422 that makes co-ordinating swatting of wildcards. Great to have this!

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -366,6 +366,12 @@ object desugar {
       (if (args.isEmpty) tycon else AppliedTypeTree(tycon, args))
         .withPos(cdef.pos.startPos)
 
+    def isHK(tparam: Tree): Boolean = tparam match {
+      case TypeDef(_, LambdaTypeTree(tparams, body)) => true
```
smarter: I don't think that's enough in general, e.g. `enum Foo[F <: [X] => Any]`, but maybe we're ok with that since that's only a temporary solution.

odersky: Yes, I think the right solution cannot be done at Desugar, it needs full type information.

PR #4425, opened 2018-04-30
Replace Names with Strings in Macro API

odersky:
Semantic names are not used enough to be really useful for the macro API. And
they are an additional concept to learn. So it's probably better to just use strings.
We had to duplicate one enum case for Types, to distinguish between TermNameRefs and
TypeNameRefs.

### DIFF
--- a/tests/pos/tasty/definitions.scala
+++ b/tests/pos/tasty/definitions.scala
@@ -2,40 +2,6 @@ package tasty
 
 object definitions {
 
-// ====== Names ======================================
-
-  trait Name
-  trait PossiblySignedName
-
-  enum TermName extends Name with PossiblySignedName {
-    case Simple(str: String)
-    case Qualified(prefix: TermName, selector: String)              // s"$prefix.$name"
-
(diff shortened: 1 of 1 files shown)


smarter:
Does that mean users need to use "Foo$"` to refer to the object `Foo` ?

Blaisorblade:
But `TermName` was meant to hide the mangling of identifiers. Would users see mangled strings now instead?

odersky:
To the object's class, yes. Or we expose a constructor for the name.

odersky:
@Blaisorblade 

> But TermName was meant to hide the mangling of identifiers. Would users see mangled strings now instead?

Yes, but we expose very few mangled names anyway. Basically: module class name `<name>$` and super accessor name `super$<name>`. The other two were protected accessor and setter, but these will be generated after Tasty in the future. It does not seem worth to create a new abstraction for just these two.

smarter:
What about name encoding? If I want `Int#+` do I need to select `"+"` or `"$plus"` ?

PR #4426, opened 2018-04-30
Prune polymorphic implicit candidates early

smarter:
Idea and testcase stolen from https://github.com/scala/scala/pull/6582


smarter:
test performance please

smarter:
This is slightly different from https://github.com/scala/scala/pull/6582 which does not change how eligible candidates are computed but rather does a second pruning check right before calling `typedImplicit`. It'd be good if we could come up with more testcases to evaluate the performance of each approach. See discusison on Gitter: https://gitter.im/typelevel/general?at=5ae775d162316e0505176c8c

smarter:
(I've commented out most of inductive-implicits-bench.scala to get it to pass the CI with a normal stack size and -Yno-deep-subtypes enabled, but we should add the full thing to our benchmark suite)

smarter:
Oops, I just realized this patch doesn't have the effect I thought it did on inductive-implicits-bench.scala, closing until I figure things out properly.

PR #4429, opened 2018-05-01
Fix tree traversal in ElimByName

allanrenucci:
ElimByName prevented some trees to be traversed by the following phases
in the group.


allanrenucci:
test performance #fast please

compiler/src/dotty/tools/dotc/transform/ElimByName.scala
```diff
@@ -50,7 +50,11 @@ class ElimByName extends TransformByNameApply with InfoTransformer {
   /** Map `tree` to `tree.apply()` is `ftree` was of ExprType and becomes now a function */
   private def applyIfFunction(tree: Tree, ftree: Tree)(implicit ctx: Context) =
     if (isByNameRef(ftree))
-      ctx.atPhase(next) { implicit ctx => tree.select(defn.Function0_apply).appliedToNone }
+      ctx.atPhase(next) { implicit ctx =>
+        transformFollowing(tree)
```
smarter: The transformFollowing should be outside of the ctx.atPhase.

PR #4432, opened 2018-05-01
Prevent position errors on Ident(nme.ERROR)

Blaisorblade:
Stop-gap fix for
```
Exception in thread "main" java.lang.AssertionError: assertion failed: position error: position not set for Ident(<error>) # 24
	at dotty.DottyPredef$.assertFail(DottyPredef.scala:36)
	at dotty.tools.dotc.ast.Positioned.check$1(Positioned.scala:178)
	at dotty.tools.dotc.ast.Positioned.check$5$$anonfun$4(Positioned.scala:203)
```

which comes up often on the unminimized @alexknvl errors.

This fix looks wrong, and I've seen the comment in `atPos` on the logic I'm fighting against, but I don't yet understand the logic enough for a proper fix.


Blaisorblade:
@smarter I hear you're one of the few experts on positions bugs?

PR #4435, opened 2018-05-02
Fix bridge generation for object definitions

odersky:
We did not generate a bridge if the overriding definition was a static object.
This caused AbstractMethodError crashes at run time.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Flags.scala
+++ b/compiler/src/dotty/tools/dotc/core/Flags.scala
@@ -542,6 +542,9 @@ object Flags {
   /** Either method or lazy */
   final val MethodOrLazy = Method | Lazy
 
+  /** Either method or module */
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -542,6 +542,9 @@ object Flags {
   /** Either method or lazy */
   final val MethodOrLazy = Method | Lazy
 
+  /** Either method or module */
+  final val MethodOrModule = Method | Module
```
nicolasstucki: Do we really need to define a new one for only one use?

odersky: It's  takes slightly more efficient that way. The `|` computation takes time, it's not just a simple bitwise `|`.

PR #4436, opened 2018-05-02
Fix handling of special implicit searches for ClassTag, quoted.Type, Eq

odersky:
They did not work if the searched-for type was an uninstantiated TypeVar that
had the class as an upper bound. Getting them to work is surprisingly subtle.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Implicits.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Implicits.scala
@@ -652,7 +652,7 @@ trait Implicits { self: Typer =>
           arg
       case fail @ SearchFailure(failed) =>
         def trySpecialCase(cls: ClassSymbol, handler: Type => Tree, ifNot: => Tree) = {
-          val base = formalValue.contextualBaseType(cls)
+          val base = formalValue.baseType(cls)
           if (base <:< formalValue) {
             // With the subtype test we enforce that the searched type `formalValue` is of the right form
             handler(base).orElse(ifNot)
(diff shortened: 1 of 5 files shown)


odersky:
Based on #4435

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
So this ended up rewriting `baseTypeOf` to be context dependent, i.e. `baseTypeOf` now takes constraint bounds for TypeParams into account. The first attempt was a bit messy so I ended up with a rewrite of `baseTypeOf` that does not caching and normal logic. The result is easier to understand and slightly more efficient than before.

odersky:
@smarter, can we get this in? It would make the Tasty extractors somewhat nicer to write.

smarter:
Sure, I'll review this today.

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -650,17 +650,20 @@ trait Implicits { self: Typer =>
...
-        else if (formalValue.isRef(defn.QuotedTypeClass))
-          synthesizedTypeTag(formalValue).orElse(tree)
-        else if (formalValue.isRef(defn.EqClass))
-          synthesizedEq(formalValue).orElse(tree)
+      case fail @ SearchFailure(failed) =>
+        def trySpecialCase(cls: ClassSymbol, handler: Type => Tree, ifNot: => Tree) = {
+          val base = formalValue.baseType(cls)
+          if (base <:< formalValue) {
+            // With the subtype test we enforce that the searched type `formalValue` is of the right form
+            handler(base).orElse(ifNot)
```
smarter: In `synthesizedClassTag` there's a call to `fullyDefinedType`, I wonder if we should move this around here to do this consistently for all handlers.

Blaisorblade:
@nicolasstucki Enjoy ClassTags!

PR #4437, opened 2018-05-02
Fix #4415: remove show from repl

### LINKED ISSUE #4415: The function whose return type is Nothing cannot be called in the REPL top-level environment
```scala
scala> def exit(): Nothing = {System.exit(0); throw new AssertionError} 
def exit(): Nothing
scala> exit() 
1 |exit()
  |^
  |value `show` is not a member of Nothing
```

OlivierBlanvillain:
This PR removes the infrastructure around the `Show` type class in the repl. Instead we simply call .toString as done in scalac. It fixes #4415 and should make things more stable.

The goal of Show was to add quotes around strings to make the output of repl copy pastable. I think using a type class with a default for `Any` is the wrong approach as it gives inconsistent results depending on the type:

```scala
Starting dotty REPL...
scala> List("foo", "bar") 
val res0: List[String] = List("foo", "bar")

scala> Option(res0)           // So far so good...
val res2: Option[List[String]] = Some(List("foo", "bar"))

scala> Some(res0)             // No instance of Some[String], fallback to showAny!
val res1: Some[List[String]] = Some(List(foo, bar))
```


compiler/src/dotty/tools/repl/Rendering.scala
```diff
@@ -43,26 +43,29 @@ private[repl] class Rendering(compiler: ReplCompiler,
...
       resObj
-        .getDeclaredMethods.find(_.getName == sym.name.toString + "Show").get
-        .invoke(null).toString
-
+        .getDeclaredMethods.find(_.getName == sym.name.encode.toString).get
+        .invoke(null)
+    val string = value match {
+      case null => "null" // Calling .toString on null => NPE
+      case ""   => "\"\"" // Sepcial cased for empty string, following scalac
+      case x    => x.toString
```
allanrenucci: Another special case for `Array`?

compiler/src/dotty/tools/repl/Rendering.scala
```diff
@@ -43,26 +43,29 @@ private[repl] class Rendering(compiler: ReplCompiler,
...
+    val objectName = sym.owner.fullName.encode.toString.stripSuffix("$")
     val resObj: Class[_] = Class.forName(objectName, true, classLoader())
-
-    val res =
+    val value =
       resObj
-        .getDeclaredMethods.find(_.getName == sym.name.toString + "Show").get
-        .invoke(null).toString
-
+        .getDeclaredMethods.find(_.getName == sym.name.encode.toString).get
```
allanrenucci: Does this assume a getter? Do we generate a getter for case like this:
```scala
> private[this] val foo = 1
```

compiler/src/dotty/tools/repl/Rendering.scala
```diff
@@ -43,26 +43,29 @@ private[repl] class Rendering(compiler: ReplCompiler,
...
+      case null => "null" // Calling .toString on null => NPE
+      case ""   => "\"\"" // Sepcial cased for empty string, following scalac
+      case x    => x.toString
+    }
     if (!sym.is(Flags.Method) && sym.info == defn.UnitType)
       None
-    else if (res.startsWith(str.REPL_SESSION_LINE))
-      Some(res.drop(str.REPL_SESSION_LINE.length).dropWhile(c => c.isDigit || c == '$'))
+    else if (string.startsWith(str.REPL_SESSION_LINE))
+      Some(string.drop(str.REPL_SESSION_LINE.length).dropWhile(c => c.isDigit || c == '$'))
```
allanrenucci: I know the code above is not new, but I'm wondering why this is needed

compiler/src/dotty/tools/repl/Rendering.scala
```diff
@@ -43,26 +43,29 @@ private[repl] class Rendering(compiler: ReplCompiler,
...
+    val value =
       resObj
-        .getDeclaredMethods.find(_.getName == sym.name.toString + "Show").get
-        .invoke(null).toString
-
+        .getDeclaredMethods.find(_.getName == sym.name.encode.toString).get
+        .invoke(null)
+    val string = value match {
+      case null => "null" // Calling .toString on null => NPE
+      case ""   => "\"\"" // Sepcial cased for empty string, following scalac
```
allanrenucci: typo: **Special**

compiler/src/dotty/tools/repl/Rendering.scala
allanrenucci: Ok I get it. It is to pretty print a class. However it quickly leaks:
```scala
scala> class Foo 
// defined class Foo
scala> new Foo 
val res0: Foo = Foo@7baf1f5a
scala> println(res0.toString) 
rs$line$1$Foo@7baf1f5a
```
It is the same in `scalac`, so I guess that's fine. I imagine we could generate a default `toString` for classes in the REPL, but it is maybe not worth the trouble

compiler/src/dotty/tools/repl/Rendering.scala
OlivierBlanvillain: But I guess it doesn't hurt to avoid the .get...

allanrenucci: I was wondering if you could access the field instead

allanrenucci:
Since this fixes #4415, can you add a test for it:
```scala
scala> lazy val x = ???
val x: Nothing = <lazy>
```

smarter:
I'm not convinced having a few hardcoded cases is better than a Show typeclass. I agree that the Show typeclass isn't ideal if it's only used by the REPL, but if we agree that having a standard Show typeclass in the standard library of Scala is useful, then it makes a lot of sense to use it for the REPL.

felixmulder:
Obviously the repl should only be able to show things that have a Show instance 😜

But I'm not emotionally attached to it even though I agree with Guillaume's comment :)

OlivierBlanvillain:
A proper `Show` type class would clearly be better, but what we have right now is simply a fancy way to hardcoded a few cases with a fallback to `.toString`. Using implicit resolution makes things more fragile and somewhat less consistent (would you have guessed the output of my code snippet?). Given that we've already had a couple of issues related to `Show`, the proposal is to burn it with fire :)

felixmulder:
So to fix `Show`, you'd have to both fix implicit resolution and remove the default value (thus no Java show - how would we solve that anyway with a principles type class 😭)?

If that's correct then we'd still have to fix implicit resolution to accommodate contravariant typeclasses at some point?

smarter:
> we'd still have to fix implicit resolution to accommodate contravariant typeclasses at some point?

This is already "fixed" in Dotty compared to scalac, unless you have something else in mind than what Dotty does now? See https://github.com/lampepfl/dotty/pull/4329

felixmulder:
@smarter - that's what I thought as well. So I'm sort of curious what part is broken about our `Show` when it comes to "fickle implicits". 🙂

Maybe it's the base case for `Any` that needs to disappear?

OlivierBlanvillain:
There isn't anything to fix about implicits, they are just the wrong tool for the job in this case. In a REPL people expect to see a String produced for everything:

```scala
scala> class Foo
defined class Foo

scala> val a = new Foo
a: Foo = Foo@377008df
```

Unless we plan to change the above (and update the entire Scala ecosystem to provide instances of Show), then we shouldn't use a typeclass. For reference here is what you get in GHC if you forget the deriving Show:

```haskell
Prelude> data Shape = Circle Float Float Float | Square Float -- deriving Show
Prelude> Circle 10 20 5 

<interactive>:2:1: error:
    • No instance for (Show Shape) arising from a use of ‘print’
    • In a stmt of an interactive GHCi command: print it
```

felixmulder:
If you just want feature parity with Scala 2 and to close some minor bugs - then yes, you're  absolutely right.

I think we can do better with Dotty, but I leave the how up to you 😀

allanrenucci:
@Blaisorblade `Show` used to be contravariant but was changed in https://github.com/lampepfl/dotty/commit/cee2b5014c02c2e16d5279de035e1b0437626402#diff-4c776b5d5db3fc261f0abe59b7a1a803

we got ambiguous implicit error when `Nothing` is inferred:
```scala
val x = List() // error ambiguous implicit. found charShow and floatShow...
```

OlivierBlanvillain:
> wouldn't it affect any contravariant typeclass?

The pattern is called implicit spaghetti, people try to avoid it ;)

Seriously, if you need contravariance + an instance for `Nothing`, it's really not a typeclass...

alexknvl:
Why not? There is nothing wrong with instances for `Nothing`.
```scala
new Show[Nothing] {
  def show(n: Nothing): String = n
}
```
is a perfectly valid `Show`.

OlivierBlanvillain:
Right, I mixed the two, the non sense is a `Nothing` instance for a covariant typeclass and an `Any` instance for a contravarant one.

Blaisorblade:
@alexknvl Opened a separate issue on typeclass resolution.

lihaoyi:
Note that Ammonite made the same decision and moved it's pretty-printing (https://github.com/lihaoyi/PPrint) to `pprintln(x: Any)` and based on `asInstanceOf` rather than an typeclass. You can configure it by manually registering `Class[_] => pprint.Tree` handlers, rather than relying on implicit resolution.

While it would be nice to "fix" implicit resolution to make `Show` always work, I highly doubt you'll be able to get implicit resolution stable enough to work 100% of the time in the presence of all the various type-system tricks that people and libraries use. Getting it working 100% of the time is necessary for something as simple as showing the result of a value at the REPL; people do not like it when things do not compile for a feature they never asked for!

In practice, manually recursing over `Any` using `asInstanceOf` checks works well enough, is 100% stable (falling back to `.toString` for un-handled cases), and have the advantage of printing `Seq(1, 2, 3)` and `Seq(1, 2, 3): Any` the same way

Blaisorblade:
@lihaoyi Ah thanks for the correction, I didn't find that in http://www.lihaoyi.com/PPrint/#TPrint, and I might have been misled by https://github.com/lihaoyi/PPrint/blob/005a51a509bae088dc69c4427a27280249ce8cf6/pprint/shared/src/main/scala-2.10%2B/TPrintImpl.scala. Is that dead code?

> have the advantage of printing Seq(1, 2, 3) and Seq(1, 2, 3): Any the same way

👍 yes, I think that's crucial for a REPL, the typeclass issues are probably more relevant for other typeclasses.

lihaoyi:
@Blaisorblade oh TPrint is used just for printing the type signature, not the actual value. For some reason that hasn't caused any stability issues like the typeclass-based value pretty-printer has

allanrenucci:
Having something similar to PPrint for the Dotty REPL would be nice. This can be implemented in a separate PR.

dwijnand:
I have a dead PR against scala/scala to "Introduce ReplPrinter as a hook for alternatives to replStringOf" that might be useful inspiration: https://github.com/scala/scala/pull/5222.

PR #4441, opened 2018-05-02
Create simple syntax highlighting tests

allanrenucci:
This is an alternative test suite to the one in #3987. I think we should rely on simple and small tests for syntax highlighting and not files.

The implementation is adapted from https://github.com/lihaoyi/Ammonite/blob/master/amm/repl/src/test/scala/ammonite/unit/HighlightTests.scala

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
+++ b/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
@@ -10,6 +10,7 @@ import util.Chars
 /** This object provides functions for syntax highlighting in the REPL */
 object SyntaxHighlighting {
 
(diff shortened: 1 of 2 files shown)


smarter:
@benkobalog Looks like we could use this instead of your proposed SyntaxHighlightingTests, what do you think?

benkobalog:
@smarter I think it's a good idea to use this simpler solution.
Probably there won't be a huge amount of syntax highlighting tests, so simplicity/speed seems to be the better option over test case readability. Even if readability would be a problem it's easy to make a function which inverts the transformation and writes it to a file or prints errors in the tests.

I can modify my PR to use this method instead, after this is merged.

allanrenucci:
@benkobalog I am merging this then

PR #4445, opened 2018-05-02
Avoid classcast exception in IDE

odersky:
We assumed that the toplevel class is always a ClassSymbol, but that assumption was wrong
in some code I was editing, which caused a crash at the point where the fix is done.


Blaisorblade:
What else did you find and what sort of code caused it? And how would this be tested?
Of course it's good to fix the issue, but without tests maintaining this code will get harder later.

smarter:
I've sene this happen too, NoSymbol can pop up here in code with parse errors.

odersky:
Agree it would be good to have tests. Maybe comments as well. Or we appeal that people should use git blame before they optimize some code that looks deliberately defensive.

Issue #4440, opened 2018-05-02
Serializable objects should have a "readResolve" method synthesized


smarter:
See explanation in https://github.com/scala/scala/blob/a9275b50e425d29c0b7fd45c10317953d5bf7804/src/compiler/scala/tools/nsc/typechecker/SyntheticMethods.scala#L319-L324
In Dotty this should be done in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala

ingarabr:
I have started working on this one.

PR #4450, opened 2018-05-03
add readResolve to objects with serializable interface

ingarabr:
If the object implements the serializable interface and not includes the
readResolve method then we add it. Since it's an object we reference it
to the objects singleton instance.

Resolves issue #4440

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
+++ b/compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
@@ -2,18 +2,13 @@ package dotty.tools.dotc
 package transform
 
 import core._
-import Symbols._, Types._, Contexts._, Names._, StdNames._, Constants._, SymUtils._
-import scala.collection.{ mutable, immutable }
+import Symbols._, Types._, Contexts._, StdNames._, Constants._, SymUtils._
 import Flags._
-import MegaPhase._
 import DenotTransformers._
-import ast.Trees._
-import ast.untpd
 import Decorators._
 import NameOps._
 import Annotations.Annotation
 import ValueClasses.isDerivedValueClass
-import scala.collection.mutable.ListBuffer
 import scala.language.postfixOps
 
 /** Synthetic method implementations for case classes, case objects,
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -261,12 +261,30 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
      */
     def canEqualBody(that: Tree): Tree = that.isInstance(AnnotatedType(clazzType, Annotation(defn.UncheckedAnnot)))
 
-    symbolsToSynthesize flatMap syntheticDefIfMissing
+    val methods = symbolsToSynthesize flatMap syntheticDefIfMissing
+
+    def createReadResolveMethod: Tree = {
```
smarter: It'd be nice to extract the common code between this method and `syntheticDef` in a separate method somehow.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -261,12 +261,30 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
...
+     else
+      methods
   }
 
-  def addSyntheticMethods(impl: Template)(implicit ctx: Context) =
-    if (ctx.owner.is(Case) || isDerivedValueClass(ctx.owner))
-      cpy.Template(impl)(body = impl.body ++ syntheticMethods(ctx.owner.asClass))
+  def addSyntheticMethods(impl: Template)(implicit ctx: Context) = {
+    val isSerializableObject =
+      (ctx.owner.is(Module)
```
smarter: To match scalac behavior, we should only do this for static objects (that is, objects which are not inside classes or methods), this can be checked using `.isStatic`.

compiler/src/dotty/tools/dotc/transform/SyntheticMethods.scala
```diff
@@ -261,12 +261,30 @@ class SyntheticMethods(thisPhase: DenotTransformer) {
...
+        && ctx.owner.derivesFrom(defn.JavaSerializableClass)
+        && !ctx.owner.asClass.membersNamed(nme.readResolve).exists)
```
smarter: We should still emit our `readResolve` method if the user declared a `readResolve` method with a different signature (e.g. `def readResolve(x: Int): AnyRef`). We can filter for methods with `(): AnyRef` as signature by doing `membersNamed(nme.readResolve).suchThat(_.signature == Signature(defn.AnyRefType, isJava = false)`

smarter:
I'll have a look next week, I think there's still one thing I wanted to do here.

smarter:
I've opened a new PR adapted from this PR and https://github.com/scala/scala/pull/7297 at https://github.com/lampepfl/dotty/pull/5775, so I'll close this one, thanks for your work @ingarabr !

PR #4457, opened 2018-05-04
Fix #4456: Pickle quote before compiling

### LINKED ISSUE #4456: Crash while running quote inside macro
```scala
import scala.quoted._
import dotty.tools.dotc.quoted.Toolbox._
object Macros {
  inline def foo(i: Int): Int = ~{
    val y = ('(i)).run
    y.toExpr
  }
}
```

```scala
import Macros._
object Test {
    println(foo(1))
    println(foo(1  + 3)) // Crashes the compiler
  }
}
```

nicolasstucki:


compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -22,7 +22,26 @@ import scala.reflect.ClassTag
 object PickledQuotes {
   import tpd._
 
-  /** Pickle the quote into strings */
+  /** Pickle the tree of the a quoted.Expr */
```
biboudis: remove the `a` :P

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -22,7 +22,26 @@ import scala.reflect.ClassTag
...
+          throw new scala.quoted.FreeVariableError(tree.name.toString)
+        case tree: DefTree =>
+          definedHere += tree.symbol
+          traverseChildren(tree)
+        case _ =>
+          traverseChildren(tree)
+      }
+    }.traverse(tree)
+    val pickled = pickleQuote(tree)
+    scala.runtime.quoted.Unpickler.unpickleExpr(pickled, Nil)
```
biboudis: The `Nil` there makes me wonder what if this fix applies to non-inline functions.

library/src/scala/quoted/Expr.scala
```diff
@@ -37,7 +37,8 @@ object Exprs {
   }
 
   /** An Expr backed by a tree. Only the current compiler trees are allowed. */
```
biboudis: Clarify what we mean by "current compiler"?

biboudis:
I tried this and it raises an exception 🤔 

```scala
import scala.quoted._
import dotty.tools.dotc.quoted.Toolbox._

object Test {

    def foo(i: Expr[Int] => Expr[Int]): Int = {
      val x = '(1)

      val y = xx: Expr[Int] => (i(xx).run)

      y(x)
    }

    println(foo(x => '{~x + 1}))
  }
}
```

biboudis:
LGTM

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
biboudis: // ignore, I have just went through the second unit test and you do what I had in mind.

nicolasstucki:
@biboudis check if this added documentation is clear enough

biboudis:
The description of the exception is exceptional 👍 😄

PR #4458, opened 2018-05-04
Fix #4455: Lift inlined values when quoted

### LINKED ISSUE #4455: Error while compiling macro inline argument used in quote
```scala
import scala.quoted._
object Macros {
  inline def foo(inline i: Int): Int = ~bar('(i))
  def bar(x: Expr[Int]): Expr[Int] = ???
}
```

```
exception while typing Macros.bar(i$1) of class class dotty.tools.dotc.ast.Trees$Apply # 175
exception while typing {
  val i$1: Int = x$0.apply(0).asInstanceOf[Int]
  Macros.bar(i$1)
} of class class dotty.tools.dotc.ast.Trees$Block # 176
exception while typing def $anonfun$1(x$0: Seq[Any]): quoted.Expr[Int] = 
  {
    val i$1: Int = x$0.apply(0).asInstanceOf[Int]
    Macros.bar(i$1)
  } of class class dotty.tools.dotc.ast.Trees$DefDef # 180
exception while typing {
  def $anonfun$1(x$0: Seq[Any]): quoted.Expr[Int] = 
    {
      val i$1: Int = x$0.apply(0).asInstanceOf[Int]
      Macros.bar(i$1)
    }
  closure($anonfun$1)
}
```

We should probably add a `.toExpr` to inlined references that are inside of a quote https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala#L549

nicolasstucki:
This is part of the transformation that allows compiling
the contents of the inline macro. The alternative would be to
pass each inline parameter twice, once as a value and once as
it's tree.


tests/run/i4455/Macro_1.scala
```diff
@@ -0,0 +1,8 @@
+import scala.quoted._
+object Macros {
+  inline def foo(inline i: Int): Int = ~bar('(i))
+  def bar(x: Expr[Int]): Expr[Int] = x
+
+  inline def foo2(inline i: Int): Int = ~bar('(i + 1))
+  def bar2(x: Expr[Int]): Expr[Int] = x
```
OlivierBlanvillain: bar2 is unused

nicolasstucki: Oh, I only need one of them. Will remove it.

PR #4461, opened 2018-05-04
Add `Comments` section in TASTY

Duhemm:
The comments are pickled when `-Ykeep-comments` is set.

### DIFF
--- a/compiler/src/dotty/tools/dotc/Driver.scala
+++ b/compiler/src/dotty/tools/dotc/Driver.scala
@@ -52,13 +52,11 @@ class Driver extends DotClass {
   }
 
   def setup(args: Array[String], rootCtx: Context): (List[String], Context) = {
-    val ctx0 = rootCtx.fresh
-    val summary = CompilerCommand.distill(args)(ctx0)
-    ctx0.setSettings(summary.sstate)
+    val ctx = rootCtx.fresh
+    val summary = CompilerCommand.distill(args)(ctx)
+    ctx.setSettings(summary.sstate)
 
-    val ctx =
-      if (shouldAddDocContext(ctx0)) ctx0.setProperty(ContextDoc, new ContextDocstrings)
-      else ctx0
+    if (shouldAddDocContext(ctx)) ctx.setProperty(ContextDoc, new ContextDocstrings)
 
     val fileNames = CompilerCommand.checkUsage(summary, sourcesRequired)(ctx)
     (fileNames, ctx)
--- a/compiler/src/dotty/tools/dotc/Driver.scala
+++ b/compiler/src/dotty/tools/dotc/Driver.scala
@@ -42,21 +42,14 @@ class Driver extends DotClass {
 
   protected def sourcesRequired = true
 
-  /**
-   * Should the `ContextDocstrings` be set for this context? The `ContextDocstrings` is used
-   * to store doc comments unless `-Ydrop-comments` is set, or when TASTY is configured to
-   * unpickle the doc comments.
-   */
-  protected def shouldAddDocContext(implicit ctx: Context): Boolean = {
-    !ctx.settings.YdropComments.value || ctx.mode.is(Mode.ReadComments)
-  }
-
   def setup(args: Array[String], rootCtx: Context): (List[String], Context) = {
     val ctx = rootCtx.fresh
     val summary = CompilerCommand.distill(args)(ctx)
     ctx.setSettings(summary.sstate)
 
-    if (shouldAddDocContext(ctx)) ctx.setProperty(ContextDoc, new ContextDocstrings)
+    if (!ctx.settings.YdropComments.value(ctx) || ctx.mode.is(Mode.ReadComments)) {
+      ctx.setProperty(ContextDoc, new ContextDocstrings)
+    }
 
     val fileNames = CompilerCommand.checkUsage(summary, sourcesRequired)(ctx)
     (fileNames, ctx)
--- a/compiler/src/dotty/tools/dotc/Driver.scala
+++ b/compiler/src/dotty/tools/dotc/Driver.scala
@@ -44,11 +44,11 @@ class Driver extends DotClass {
 
   /**
    * Should the `ContextDocstrings` be set for this context? The `ContextDocstrings` is used
-   * to store doc comments when `-Ykeep-comments` is set, or when TASTY is configured to
+   * to store doc comments unless `-Ydrop-comments` is set, or when TASTY is configured to
    * unpickle the doc comments.
    */
   protected def shouldAddDocContext(implicit ctx: Context): Boolean = {
-    ctx.settings.YkeepComments.value || ctx.mode.is(Mode.ReadComments)
+    !ctx.settings.YdropComments.value || ctx.mode.is(Mode.ReadComments)
   }
 
   def setup(args: Array[String], rootCtx: Context): (List[String], Context) = {
--- a/compiler/src/dotty/tools/dotc/Driver.scala
+++ b/compiler/src/dotty/tools/dotc/Driver.scala
@@ -2,7 +2,9 @@ package dotty.tools.dotc
 
 import dotty.tools.FatalError
 import config.CompilerCommand
+import core.Comments.{ContextDoc, ContextDocstrings}
 import core.Contexts.{Context, ContextBase}
+import core.Mode
 import util.DotClass
 import reporting._
 import scala.util.control.NonFatal
@@ -40,10 +42,24 @@ class Driver extends DotClass {
 
   protected def sourcesRequired = true
 
+  /**
+   * Should the `ContextDocstrings` be set for this context? The `ContextDocstrings` is used
+   * to store doc comments when `-Ykeep-comments` is set, or when TASTY is configured to
+   * unpickle the doc comments.
+   */
+  protected def shouldAddDocContext(implicit ctx: Context): Boolean = {
+    ctx.settings.YkeepComments.value || ctx.mode.is(Mode.ReadComments)
+  }
+
   def setup(args: Array[String], rootCtx: Context): (List[String], Context) = {
-    val ctx = rootCtx.fresh
-    val summary = CompilerCommand.distill(args)(ctx)
-    ctx.setSettings(summary.sstate)
+    val ctx0 = rootCtx.fresh
+    val summary = CompilerCommand.distill(args)(ctx0)
+    ctx0.setSettings(summary.sstate)
+
+    val ctx =
+      if (shouldAddDocContext(ctx0)) ctx0.setProperty(ContextDoc, new ContextDocstrings)
+      else ctx0
+
     val fileNames = CompilerCommand.checkUsage(summary, sourcesRequired)(ctx)
     (fileNames, ctx)
   }
--- a/compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
@@ -4,10 +4,10 @@ import dotty.tools.dotc.core.tasty._
 import dotty.tools.dotc.core.tasty.TastyUnpickler.NameTable
 
 object TastyUnpickler {
-  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], splices: Seq[Any])
-    extends DottyUnpickler.TreeSectionUnpickler(posUnpickler) {
+  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], commentUnpickler: Option[CommentUnpickler], splices: Seq[Any])
+    extends DottyUnpickler.TreeSectionUnpickler(posUnpickler, commentUnpickler) {
     override def unpickle(reader: TastyReader, nameAtRef: NameTable) =
-      new TreeUnpickler(reader, nameAtRef, posUnpickler, splices)
+      new TreeUnpickler(reader, nameAtRef, posUnpickler, commentUnpickler, splices)
   }
 }
 
@@ -19,6 +19,6 @@ class TastyUnpickler(bytes: Array[Byte], splices: Seq[Any]) extends DottyUnpickl
   import DottyUnpickler._
   import TastyUnpickler._
 
-  protected override def treeSectionUnpickler(posUnpicklerOpt: Option[PositionUnpickler]): TreeSectionUnpickler =
-    new QuotedTreeSectionUnpickler(posUnpicklerOpt, splices)
+  protected override def treeSectionUnpickler(posUnpicklerOpt: Option[PositionUnpickler], commentUnpicklerOpt: Option[CommentUnpickler]): TreeSectionUnpickler =
+    new QuotedTreeSectionUnpickler(posUnpicklerOpt, commentUnpicklerOpt, splices)
 }
--- a/compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
@@ -4,14 +4,14 @@ import dotty.tools.dotc.core.tasty._
 import dotty.tools.dotc.core.tasty.TastyUnpickler.NameTable
 
 object TastyUnpickler {
-  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], commentUnpickler: Option[CommentUnpickler], splices: Seq[Any])
-    extends DottyUnpickler.TreeSectionUnpickler(posUnpickler, commentUnpickler) {
+  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], splices: Seq[Any])
+    extends DottyUnpickler.TreeSectionUnpickler(posUnpickler, None) {
     override def unpickle(reader: TastyReader, nameAtRef: NameTable) =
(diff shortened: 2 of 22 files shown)


compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
...
+import java.nio.file.{FileSystems, FileVisitOption, FileVisitResult, FileVisitor, Files, Path}
+import java.nio.file.attribute.BasicFileAttributes
+import java.util.EnumSet
+
+import scala.concurrent.duration.{Duration, DurationInt}
+
+import org.junit.Test
+import org.junit.Assert.{assertEquals, assertFalse, fail}
+
+class CommentPicklingTest extends ParallelTesting {
```
allanrenucci: Why does this extends `ParallelTesting`?

compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
...
+      allFiles.foreach(Files.delete(_))
+    }
+  }
+
+  private def getAll(base: Path,
+             pattern: String,
+             maxDepth: Int = Int.MaxValue): Seq[Path] = {
+    val out = collection.mutable.ListBuffer.empty[Path]
+    val matcher = FileSystems.getDefault.getPathMatcher(pattern)
+    val visitor = new FileVisitor[Path] {
```
allanrenucci: Something like?
```scala
val paths = Files.walk(base)
try paths.filter(matcher.matches).iterator().asScala.toList
finally paths.close()
```

compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
...
+    try fn(temp)
+    finally {
+      val allFiles = getAll(temp, "glob:**").sortBy(_.toAbsolutePath.toString).reverse
+      allFiles.foreach(Files.delete(_))
+    }
+  }
+
+  private def getAll(base: Path,
+             pattern: String,
+             maxDepth: Int = Int.MaxValue): Seq[Path] = {
```
allanrenucci: `maxDepth` is not used. Remove?

compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
...
+    val sources = "/** foo */ object A" :: Nil
+    compileAndCheckComment(sources, "A".toTermName, Some("/** foo */"))
+  }
+
+  @Test def commentOnlazyVal: Unit = {
+    val sources = "class A { /** foo */ lazy val buzz = 2 }" :: Nil
+    compileAndCheckComment(sources, "buzz".toTermName, Some("/** foo */"))
+  }
+
+  private def compileAndCheckComment(sources: Seq[String], treeName: Name, expectedComment: Option[String]): Unit = {
```
allanrenucci: I would use `List` instead of `Seq` across the file. I don't think you gain anything by using `Seq`

allanrenucci: Do we optimise the API for the most common use case where you want to check for the presence of a comment? `Option[String]` => `String`.

Also maybe add an overload for one source

compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
...
+          val comment = for { docCtx <- ctx.docCtx
+                              comment <- docCtx.docstring(symbol) } yield comment.raw
+          assertEquals(expectedComment, comment)
+        case other =>
+          fail(s"Unexpected: $other")
+      }
+    }
+  }
+
+  private def findTreeNamed(name: Name)(trees: Seq[tpd.Tree], ctx: Context): Option[tpd.NameTree] = {
```
allanrenucci: We have `find` on `Tree` in `tpd.TreeOps`

compiler/test/dotty/tools/dotc/CommentPicklingTest.scala
```diff
@@ -0,0 +1,178 @@
+package dotty.tools.dotc
```
allanrenucci: Move to tasty package?

OlivierBlanvillain:
What do you think about reversing the flag, `-Ydrop-comments`? Python has great success with carrying comments everywhere!

Duhemm:
@OlivierBlanvillain I've tried switching `-Ykeep-comments` to `-Ydrop-comments`, but unfortunately this doesn't work with `-Ycheck` at the moment. There are several reasons for that:

- The `@usecase` annotations present in the docstrings introduce new symbols. `-Ycheck` will complain that the enclosing class has a symbol for which it doesn't find a declaration. This one is easy to solve by special-casing `-Ycheck` for those symbols (there must be a better solution, though...) https://github.com/lampepfl/dotty/blob/91de02962767cef1f6221e2bd565367aca1f3efc/compiler/src/dotty/tools/dotc/typer/Docstrings.scala#L26
- The symbols introduced by the `@usecase` annotations have names that are not valid on the JVM. This could probably be changed, but I imagine that Dottydoc extracts information from that name? This sounds like a hack to me. 
https://github.com/lampepfl/dotty/blob/91de02962767cef1f6221e2bd565367aca1f3efc/compiler/src/dotty/tools/dotc/typer/Docstrings.scala#L26
- I got other errors with `-Ycheck` in the body of the `@usecase` definition; I didn't debug them further.

I'd prefer to leave the switching of the flag for a later PR, when it's clearer to me what needs to be fixed in order to keep the comments by default.

compiler/src/dotty/tools/dotc/core/tasty/CommentPickler.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.core.Comments.{Comment, CommentsContext}
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.tasty.TastyBuffer.Addr
+
+import java.nio.charset.Charset
+
+class CommentPickler(pickler: TastyPickler, addrOfTree: tpd.Tree => Option[Addr])(implicit ctx: Context) {
+  val buf = new TastyBuffer(5000)
```
nicolasstucki: `private[this]`

compiler/src/dotty/tools/dotc/core/tasty/TastyBuffer.scala
```diff
@@ -48,6 +48,10 @@ class TastyBuffer(initialSize: Int) {
 
   // -- Output routines --------------------------------------------
 
+  /** Write a boolean value. */
+  def writeBoolean(b: Boolean): Unit =
```
nicolasstucki: I would write the byte explicitly `pickleComment` and read the byte in `CommentUnpickler`. This abstraction is only used once and hides the underlying representation. It would make it too easy to write two calls to `writeBoolean` instead of one to `writeByte` (with bitset representation).

compiler/src/dotty/tools/dotc/core/tasty/CommentPickler.scala
```diff
@@ -0,0 +1,43 @@
...
+    new Traverser().traverse(root)
+
+  def pickleComment(addrOfTree: Option[Addr], comment: Option[Comment]): Unit = (addrOfTree, comment) match {
+    case (Some(addr), Some(cmt)) =>
+      val bytes = cmt.raw.getBytes(Charset.forName("UTF-8"))
+      val length = bytes.length
+      buf.writeAddr(addr)
+      buf.writeNat(length)
+      buf.writeBytes(bytes, length)
+      buf.writeBoolean(cmt.isExpanded)
```
nicolasstucki: This seems to be missing from the documentation.

compiler/src/dotty/tools/dotc/core/tasty/TastyReader.scala
```diff
@@ -41,6 +41,10 @@ class TastyReader(val bytes: Array[Byte], start: Int, end: Int, val base: Int =
   def subReader(start: Addr, end: Addr): TastyReader =
     new TastyReader(bytes, index(start), index(end), base)
 
+  /** Read a boolean value. */
+  def readBoolean(): Boolean =
```
nicolasstucki: Same

nicolasstucki:
The `TastyPrinter` should print something for this section as is done for [positions](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/tasty/TastyPrinter.scala#L110-L121). You can test the printer with `dotc -print-tasty XYZ` for class `XYZ` in your classpath. This can be done in another PR.

compiler/src/dotty/tools/dotc/core/tasty/CommentPickler.scala
Duhemm: What do you mean?

Duhemm: I got it now

Duhemm:
@nicolasstucki I've addressed your comments and modified the `TastyPrinter` to print the comments, too. Maybe it's too verbose to print the whole comments? I don't know how you usually use it, but it may be better to print only the first line of every comment?

nicolasstucki:
Maybe print the first `n` characters of the string and then add `...` and say on the side the actual length of the comment. `n` could be based on the current value of `-pagewidth`.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -820,6 +823,16 @@ class TreeUnpickler(reader: TastyReader,
         // Child annotations for local classes and enum values are not pickled, so
         // need to be re-established here.
         sym.registerIfChild(late = true)
+
+      if (ctx.mode.is(Mode.ReadComments)) {
+        for { docCtx <- ctx.docCtx
```
allanrenucci: If the mode `ReadComments`, then I suppose we should both `docCtx` and `commentUnpickler` should not be empty

compiler/src/dotty/tools/dotc/Driver.scala
```diff
@@ -40,10 +42,22 @@ class Driver extends DotClass {
 
   protected def sourcesRequired = true
 
+  /**
+   * Should the `ContextDocstrings` be set for this context? The `ContextDocstrings` is used
+   * to store doc comments unless `-Ydrop-comments` is set, or when TASTY is configured to
+   * unpickle the doc comments.
+   */
+  protected def shouldAddDocContext(implicit ctx: Context): Boolean = {
+    !ctx.settings.YdropComments.value || ctx.mode.is(Mode.ReadComments)
```
allanrenucci: Why do we have both a new mode and a compiler option?
Isn't `!ctx.settings.YdropComments <=> Mode.ReadComments`.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
Duhemm: The same goes for `posUnpicklerOpt`: we check if the mode is `ReadPositions` and whether `posUnpicklerOpt` is set.

The `TreeUnpickler` can be constructed with a certain configuration in mind (i.e. not defining `commentUnpicklerOpt`), and then used with a different configuration (i.e. `ctx.mode.is(Mode.ReadComments)`), which would lead to crashes if we just used `.get` on the `Option`.

Do we prefer crashing? Should we add a warning?

compiler/src/dotty/tools/dotc/Driver.scala
```diff
@@ -40,10 +42,22 @@ class Driver extends DotClass {
 
   protected def sourcesRequired = true
 
+  /**
+   * Should the `ContextDocstrings` be set for this context? The `ContextDocstrings` is used
+   * to store doc comments unless `-Ydrop-comments` is set, or when TASTY is configured to
+   * unpickle the doc comments.
+   */
+  protected def shouldAddDocContext(implicit ctx: Context): Boolean = {
```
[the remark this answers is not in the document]
allanrenucci: Likely or not, this is not currently the case. I would just inline it into `setup`. In the future, if there is a use case, one can always extract the wanted bits into a protected def

compiler/src/dotty/tools/dotc/Driver.scala
allanrenucci: Fair enough

compiler/src/dotty/tools/dotc/core/tasty/CommentPickler.scala
```diff
@@ -0,0 +1,43 @@
...
+      val length = bytes.length
+      buf.writeAddr(addr)
+      buf.writeNat(length)
+      buf.writeBytes(bytes, length)
+      buf.writeByte(if (cmt.isExpanded) 1 else 0)
+    case other =>
+      ()
+  }
+
+  private class Traverser extends tpd.TreeTraverser {
```
allanrenucci: Make the traverser take the `docCtx` and only create the traverser in `pickleComment` if it exists

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
allanrenucci: I really can't tell why the position unpickler is taken as an option.

I don't see a good reason to do the same. It only makes thing more complicated. I think the `TreeUnpickler` does not need an optional `CommentUnpickler`. I think we should just look at `Mode.ReadComments` to decide whether or not we unpickle comments. And in the case we do, we should assert that we have a `docCtx`.

Maybe @smarter or @nicolasstucki can explain why it was done like this for `PositionUnpickler`

compiler/test/dotty/tools/dotc/core/tasty/CommentPicklingTest.scala
```diff
@@ -0,0 +1,151 @@
...
+                              comment <- docCtx.docstring(symbol) } yield comment.raw
+          assertEquals(expectedComment, comment)
+        case other =>
+          fail(s"Unexpected: $other")
+      }
+    }
+  }
+
+  private def findTreeNamed(name: Name)(trees: List[tpd.Tree], ctx: Context): Option[tpd.NameTree] = {
+    val acc = new tpd.TreeAccumulator[Option[tpd.NameTree]] {
```
allanrenucci: why not use `TreeOps.find` from `tpd.scala`?

compiler/test/dotty/tools/dotc/core/tasty/CommentPicklingTest.scala
```diff
@@ -0,0 +1,151 @@
...
+          val path = tmp.resolve(s"Src$id.scala").toAbsolutePath
+          Files.write(path, src.getBytes("UTF-8"))
+          path.toString
+      }
+
+      val out = tmp.resolve("out")
+      Files.createDirectories(out)
+
+      val options = compileOptions.and("-d", out.toAbsolutePath.toString).and(sourceFiles: _*)
+      val driver = new Driver
```
allanrenucci: `dotty.tools.package.dotc.Main`?

compiler/test/dotty/tools/dotc/core/tasty/CommentPicklingTest.scala
```diff
@@ -0,0 +1,151 @@
...
+          case md: tpd.NameTree if md.name == name => Some(md)
+          case other => foldOver(None, other)
+        })
+      }
+    }
+    acc(None, trees)(ctx)
+  }
+
+  private def compileAndUnpickle[T](sources: List[String])(fn: (List[tpd.Tree], Context) => T) = {
+    inTempDirectory { tmp =>
```
allanrenucci: A nice to have (that shouldn't block this PR): a driver that compiles from virtual files. I think it would be fairly easy: override `doCompile` and call `compileSources(List[SourceFile])` instead of  `compile(List[String])`. Creating a virtual source file is straightforward: `new SourceFile("<fake name>", content.toCharArray)`

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
smarter: No idea, but what you said sounds reasonable

compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
```diff
@@ -4,10 +4,10 @@ import dotty.tools.dotc.core.tasty._
 import dotty.tools.dotc.core.tasty.TastyUnpickler.NameTable
 
 object TastyUnpickler {
-  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], splices: Seq[Any])
-    extends DottyUnpickler.TreeSectionUnpickler(posUnpickler) {
+  class QuotedTreeSectionUnpickler(posUnpickler: Option[PositionUnpickler], commentUnpickler: Option[CommentUnpickler], splices: Seq[Any])
```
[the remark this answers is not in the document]
nicolasstucki: I do not have a use case in mind. We should not add it.

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
Duhemm: It's actually valid for the `positionUnpickler` or `commentUnpickler` to be `None` if there are no `Positions` or `Comments` sections in the TASTY file from which we're unpickling.

PR #4467, opened 2018-05-05
Default to emitting .tasty files (not .hasTasty)

nicolasstucki:


compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -404,8 +404,11 @@ trait ParallelTesting extends RunnerOrchestration { self =>
       tastyOutput.mkdir()
       val flags = flags0 and ("-d", tastyOutput.getAbsolutePath) and "-from-tasty"
 
-      def hasTastyFileToClassName(f: JFile): String =
-        targetDir.toPath.relativize(f.toPath).toString.dropRight(".hasTasty".length).replace('/', '.')
+      def hasTastyFileToClassName(f: JFile): String = {
```
allanrenucci: name is now misleading

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -404,8 +404,11 @@ trait ParallelTesting extends RunnerOrchestration { self =>
...
+        val pathStr = targetDir.toPath.relativize(f.toPath).toString.replace('/', '.')
+        if (pathStr.endsWith(".hasTasty")) pathStr.dropRight(".hasTasty".length)
+        else pathStr.dropRight(".tasty".length)
```
allanrenucci: `.stripSuffix(".hasTasty").stripSuffix(".hasTasty")`?

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -700,7 +700,7 @@ object PatternMatcher {
           }
           val newArgs =
             for {
-              (rhs, actual) <- seenVars.toList
+              (rhs, actual) <- seenVars.toList.sortBy(_._2.name.toString)
```
allanrenucci: We have an ordering on `Name` (i.e. `Names.NameOrdering`), you shouldn't need to do `toString`

nicolasstucki: This is just to test if this is the source of the idempotency failure.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -798,8 +798,9 @@ class ClassfileParser(
               Array.empty[Byte]
             case Some(jar: ZipArchive) => // We are in a jar
               val jarFile = JarArchive.open(io.File(jar.jpath))
-              try readTastyForClass(jarFile.jpath.resolve(classfile.path))
-              finally jarFile.close()
+              readTastyForClass(jarFile.jpath.resolve(classfile.path))
+              // Do not close the file system as some else might use it later. Once closed it cannot be re-opened.
+              // TODO find a way to safly close the file system or ose some other abstraction
```
allanrenucci: Why is this an issue now but it was not before?

nicolasstucki: We never reused the same jar file in the same JVM to load some TASTY. With this change, we load the bootstrapped library from the `.tasty` files. During `dotty-bootstrapped/testCompilation` we do reuse this jar all the time.

compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -256,7 +256,7 @@ class CompilationTests extends ParallelTesting {
       // as well as bootstrapped compiler:
       defaultOutputDir + dotty1Group + "/dotty/:" +
       Jars.dottyInterfaces,
-      Array("-Ycheck-reentrant")
+      Array("-Ycheck-reentrant", "-Yemit-tasty-in-class")
```
allanrenucci: Why not the default?

compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -1369,5 +1371,5 @@ object ParallelTesting {
   }
 
   def isHasTastyFile(f: JFile): Boolean =
```
allanrenucci: Maybe rename `isTastyFile`

compiler/test/dotty/tools/dotc/CompilationTests.scala
nicolasstucki: It triggers an idempotency test failure. I have partially fixed it but still working on finding the second source of non-determinism.

nicolasstucki:
Yes, let's wait for `0.9.0-RC1`.

nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
Rebased

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
smarter: Does this leak memory? If so it definitely needs to be addressed before merging. Also on Windows keeping a file open means that it cannot be deleted, does this apply to this stuff too?

nicolasstucki: It probably leaks but I was not able to see it. I will have another look at this.

nicolasstucki: I changed the implementation and used a classloader to load the tasty files in the jar.

nicolasstucki:
@allanrenucci could you review the last  commit

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -784,24 +786,30 @@ class ClassfileParser(
...
+              Array.empty
             case Some(jar: ZipArchive) => // We are in a jar
-              val jarFile = JarArchive.open(io.File(jar.jpath))
-              try readTastyForClass(jarFile.jpath.resolve(classfile.path))
-              finally jarFile.close()
+              val cl = new URLClassLoader(Array(jar.jpath.toUri.toURL))
+              val path = classfile.path.stripSuffix(".class") + ".tasty"
+              val stream = cl.getResourceAsStream(path)
+              if (stream != null) {
+                val tasty = new Array[Byte](stream.available())
```
allanrenucci: From the [Javadoc](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#available())
> Note that while some implementations of InputStream will return the total number of bytes in the stream, many will not. It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -784,24 +786,30 @@ class ClassfileParser(
...
+          val tastyBytes: Array[Byte] = classfile.underlyingSource match { // TODO: simplify when #3552 is fixed
             case None =>
               ctx.error("Could not load TASTY from .tasty for virtual file " + classfile)
-              Array.empty[Byte]
+              Array.empty
             case Some(jar: ZipArchive) => // We are in a jar
-              val jarFile = JarArchive.open(io.File(jar.jpath))
-              try readTastyForClass(jarFile.jpath.resolve(classfile.path))
-              finally jarFile.close()
+              val cl = new URLClassLoader(Array(jar.jpath.toUri.toURL))
```
allanrenucci: Are you sure we won't have to deal with the same issues using a `URLClassLoader`?

nicolasstucki: I would say no because the classloader never instanciates an class with a reference to it. Hence the classloader should be GCed at some point. I will experiment a bit with the profiler to see if I can observe any leaks.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -784,25 +786,31 @@ class ClassfileParser(
...
-              readTastyForClass(jarFile.jpath.resolve(classfile.path))
-              // Do not close the file system as some else might use it later. Once closed it cannot be re-opened.
-              // TODO find a way to safly close the file system or ose some other abstraction
+              val cl = new URLClassLoader(Array(jar.jpath.toUri.toURL))
+              val path = classfile.path.stripSuffix(".class") + ".tasty"
+              val stream = cl.getResourceAsStream(path)
+              if (stream != null) {
+                val tasty = Array.newBuilder[Byte]
+                tasty.sizeHint(stream.available())
+                while (stream.available() > 0) // TODO improve performance
```
allanrenucci: I think the proper way to read from an `InputStream`, is to `read()` until `-1` is returned

nicolasstucki: Changed it

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
nicolasstucki: I tried it and it does not affect the class loader. I closed the file system in the middle of reading from the stream and some other places.

nicolasstucki:
Rebased

odersky:
Ready to merge?

nicolasstucki:
It is ready to merge. But after it is merged everyone needs to clean their `out/` directory.

nicolasstucki:
I ran the `sbt-dotty/scripted` tests locally and `source-dependencies / continuations` failed. Will test it on the CI

nicolasstucki:
`sbt-dotty/scripted` passed on the CI http://dotty-ci.epfl.ch/lampepfl/dotty/6302

PR #4468, opened 2018-05-05
Fix #4322: Avoid generating multiple inline accessors for the same ac…

### LINKED ISSUE #4322: When a private member is accessed from an inline method, a new public accessor is created for every use of the private member
This code should only create one accessor for `x`, but it ends up creating three:
```scala
object Foo {
  private[this] val x: Int = 1

  inline def foo: Int = x + x + x
}
```
```scala
  final lazy module val Foo: Foo$ = new Foo$()
  final module class Foo$() extends Object() { this: Foo.type => 
    private[this] val x: Int = 1
    @Foo.foo$_inlineAccessor_$1.+(Foo.foo$_inlineAccessor_$2).+(
      Foo.foo$_inlineAccessor_$3
    ) inline def foo: Int = 
      Foo.foo$_inlineAccessor_$1.+(Foo.foo$_inlineAccessor_$2).+(
        Foo.foo$_inlineAccessor_$3
      )
    def foo$_inlineAccessor_$1: Int = Foo.x
    def foo$_inlineAccessor_$2: Int = Foo.x
    def foo$_inlineAccessor_$3: Int = Foo.x
  }
```

odersky:
…cess

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inliner.scala
@@ -31,7 +31,7 @@ import util.Positions.Position
 object Inliner {
   import tpd._
 
-  object InlineAccessors extends AccessProxies {
+  class InlineAccessors extends AccessProxies {
     def getterName = InlineGetterName
     def setterName = InlineSetterName
 
@@ -95,7 +95,7 @@ object Inliner {
           sym.updateAnnotation(LazyBodyAnnotation { _ =>
             implicit val ctx = inlineCtx
             val body = treeExpr(ctx)
-            if (ctx.reporter.hasErrors) body else InlineAccessors.makeInlineable(body)
+            if (ctx.reporter.hasErrors) body else ctx.compilationUnit.inlineAccessors.makeInlineable(body)
           })
         }
     }
--- a/compiler/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inliner.scala
@@ -49,6 +49,10 @@ object Inliner {
     object addAccessors extends TreeMap {
       val accessors = new mutable.ListBuffer[MemberDef]
 
+      type AccessorMap = mutable.HashMap[Symbol, DefDef]
+      private val getters = new AccessorMap
+      private val setters = new AccessorMap
+
       /** A definition needs an accessor if it is private, protected, or qualified private
        *  and it is not part of the tree that gets inlined. The latter test is implemented
        *  by excluding all symbols properly contained in the inlined method.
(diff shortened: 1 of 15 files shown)


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -48,6 +48,10 @@ object Inliner {
     object addAccessors extends TreeMap {
       val accessors = new mutable.ListBuffer[MemberDef]
 
+      type AccessorMap = mutable.HashMap[Symbol, DefDef]
```
nicolasstucki: Use `MutableSymbolMap[T]`

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -48,6 +48,10 @@ object Inliner {
...
+      private val getters = new AccessorMap
```
nicolasstucki: Use `newMutableSymbolMap`

tests/pos/i4322.scala
```diff
@@ -0,0 +1,10 @@
+object Foo {
+  private[this] var x: Int = 1
+
+  inline def foo: Int = x + x + x
```
smarter: I suggest turning this into a run test and then checking that there's only one accessor method using `Class#getMethods` (see https://github.com/lampepfl/dotty/blob/master/tests/run/traitNoInit.scala for an example on how to do something like that)

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -83,71 +87,77 @@ object Inliner {
        *  @param rhs          A function that builds the right-hand side of the accessor,
        *                      given a reference to the accessed symbol and any type and
        *                      value arguments the need to be integrated.
+       *  @param seen         An map of already generated accessor methods of this kind (getter or setter)
```
smarter: typo: An map -> A map

smarter:
Even with this PR, we're still generating two accessors for `x` because it's accessed both by `foo` and `bar` in `i4322.scala`, I think ideally we should only generate one. To do this we'd need to change the naming scheme for these accessors to be `<nameOfAccessedMethod>$inlineAccessor` instead of `<nameOfInlineMethod>_inlineAccessor_$<n>`. Do you know of any reason to not do this?

odersky:
@smarter The new version generates only one accessor per symbol.

odersky:
@smarter @nicolasstucki Can one of you give this another review?

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -58,131 +57,64 @@ object Inliner {
...
-              case tp => MethodType(qualType :: Nil, tp)
+        val accessorInfo =
+          if (onLHS) MethodType(accessed.info :: Nil, defn.UnitType)
+          else accessed.info.ensureMethodic
+        val accessorName = nameKind(accessed.name)
+        val accessorSymbol =
+          accessed.owner.info.decl(accessorName).suchThat(refersToAccessed).symbol
+            .orElse {
+              val acc = newAccessorSymbol(accessed, accessorName, accessorInfo)
+              acc.addAnnotation(Annotation.Accessed(accessed))
```
nicolasstucki: This should probably go in `newAccessorSymbol`

PR #4479, opened 2018-05-08
Handle QuotedError when evaluating holes in quotes

nicolasstucki:
These holes are expanded while unpickling in PickledQuotes.quotedExprToTree


nicolasstucki:
I will outline it.

nicolasstucki:
@biboudis I updated it WDYT?

PR #4480, opened 2018-05-08
FIx source of performance regression in #3961

nicolasstucki:


nicolasstucki:

nicolasstucki:
test performance please

nicolasstucki:

nicolasstucki:

nicolasstucki:

nicolasstucki:

nicolasstucki:
test performance please

nicolasstucki:

nicolasstucki:

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -44,14 +44,12 @@ class StringInterpolatorOpt extends MiniPhase {
...
     def unapply(tree: Tree)(implicit ctx: Context): Option[(List[Literal], List[Tree])] = {
-      if (tree.symbol.eq(defn.StringContextRaw) || tree.symbol.eq(defn.StringContextS)) {
-        tree match {
-          case Apply(Select(Apply(StringContextApply(), List(Literals(strs))), _),
-          List(SeqLiteral(elems, _))) if elems.length == strs.length - 1 =>
-            Some(strs, elems)
-          case _ => None
-        }
-      } else None
+      tree match {
```
liufengyun: A minor issue: the object name `SOrRawInterpolator` doesn't strictly correspond to what is in the code.

nicolasstucki: That is not logic I made, I did not choose that name. Do you want me to change it to something else?

allanrenucci: It used to reflect what was in the code. Not anymore. I would just inline this extractor in `StringContextIntrinsic`

liufengyun: Before the change, it's. After the symbolic equality checks are removed, it wasn't. But as we are going to remove local optimisations, it's fine to keep it as it is now.

allanrenucci: @liufengyun This is not part of the local optimisations that will be removed. Scalac does something similar

PR #4481, opened 2018-05-08
Fix #4414: Use original type in generated unpickle{Expr|Type}

### LINKED ISSUE #4414: Assertion error when passing an Expr[A] at a polymorphic method with Type evidence for A
It is probably not an intended use if you require a type tag for an `Expr` using it in a single-stage manner but here is an exception I get with `-Ycheck: all` while typing the method `b[quoted.Expr[A]]`.

```scala
import scala.quoted._

object Test {

  def a[A: Type](): Unit = {
    b[Expr[A]]()
    a[A]()
  }

  def b[A: Type](): Unit = ???
}
```

with the following assertion error:
```
assertion failed: found:    scala.quoted.Type[quoted.Expr[T$1]]
required: quoted.Type[quoted.Expr[A]]
```

The problem disappears if I do one of these (mechanically):
- either remove `a[A]()` or 
- remove the `Type` evidence in `b` 
- or add a `()` return value in `a`
- or change the order of the method calls `a` and `b`

nicolasstucki:


biboudis:
I am wondering why this hasn't been manifested before?? It is not related to `Expr[T]` per se. In any case, it looks good. 👍 @nicolasstucki

PR #4483, opened 2018-05-08
Try to reduce allocations

odersky:
1. Cache DottyBackendInterface/SymbolHelpers


odersky:
test performance please

odersky:
Seems the profiler fooled us again. It highlighted SymHelper as the second most allocated data. But removing the allocations has no discernible effect. Maybe escape analysis is normally successful in avoiding the allocations and instrumenting for profiling actually causes the object allocations.

Issue #4459, opened 2018-05-04
Three bugs caught by FindBugs


alexknvl:
#### Using pointer equality to compare different types
```scala
val receiverIsSuper = (method.name eq sym) && enclosingClass.appliedRef.widen <:< recvWiden
```
on line 276 in `dotty.tools.dotc.transform.TailRec`

#### Self comparison of value with itself
```scala
  private class Set3[+Elem <: AnyRef](x0: AnyRef, x1: AnyRef, x2: AnyRef) extends SimpleIdentitySet[Elem] {
    def size = 3
    def + [E >: Elem <: AnyRef](x: E): SimpleIdentitySet[E] =
      if (contains(this)) this // <---- WAT
      else {
```
on line 62 of `dotty.tools.dotc.util.SimpleIdentitySet`

#### Method ignores return value
```scala
    val boundsText = domainLambdas
      " bounds = " + {
        val assocs =
          for (param <- domainParams)
          yield
            param.binder.paramNames(param.paramNum) + ": " + entryText(entry(param))
        assocs.mkString("\n")
    }
```
in `dotty.tools.dotc.core.OrderingConstraint`, line 639

allanrenucci:
```scala
val receiverIsSuper = (method.name eq sym) && enclosingClass.appliedRef.widen <:< recvWiden
```
It should be `method.name == sym.name`.

However the current implementation if very inconsistent. With the fix you would get one error for:
```scala
class Foo {
  def foo = 1
}

class Test extends Foo {
  @tailrec
  final override def foo = super.foo // error
}
```
and two errors for:
```scala
class Foo {
  def foo() = 1
}

class Test extends Foo {
  @tailrec
  final override def foo() = super.foo() // error // error
}
```

allanrenucci:
> in `dotty.tools.dotc.core.OrderingConstraint`, line 639

Should probably be:
```scala
    val boundsText =
      " bounds = " + {
        val assocs =
          for (param <- domainParams)
          yield
            param.binder.paramNames(param.paramNum) + ": " + entryText(entry(param))
        assocs.mkString("\n")
    }
```

smarter:
All three bugs have been fixed.

PR #4488, opened 2018-05-08
Fix typo in SimpleIdentitySet

Blaisorblade:
This might not affect correctness but will create a bigger set even if unneeded.

From #4459.


smarter:
Ideally we should have unit tests for this too.

Issue #4502, opened 2018-05-10
Unnecessary outer accessors are sometimes generated


smarter:
Consider:
```scala
class Foo
class Bar { self =>
  new Foo {
    println(self)
  }
}
```

The anonymous class generated by Dotty looks like:
```scala
private static class Bar$.anon
extends Foo {
    private final Bar $outer;

    public Bar$.anon(Bar $outer) {
        if ($outer == null) {
            throw new NullPointerException();
        }
        this.$outer = $outer;
        Predef$.MODULE$.println((Object)this.Bar$_$$anon$$$outer());
    }

    private Bar $outer() {
        return this.$outer;
    }

    public final Bar Bar$_$$anon$$$outer() {
        return this.$outer();
    }
}
```
Whereas with scalac we get:
```scala
public final class Bar$.anon
extends Foo {
    public Bar$.anon(Bar $outer) {
        Predef$.MODULE$.println((Object)$outer);
    }
}
```

This seems to be implemented by removing outer accessors that are never called in the constructors phase:  https://github.com/scala/scala/blob/a52482c7236a782adedf7523301a97659e4d720f/src/compiler/scala/tools/nsc/transform/Constructors.scala#L114-L117

DieBauer:
I tried reproducing the above, and can't seem to get the same output with current master. Given that the issue is almost 2 years old, I was wondering is this still an issue? If yes, which reproducer would trigger this?

smarter:
Thanks for checking, this was probably fixed by https://github.com/lampepfl/dotty/pull/8253

PR #4503, opened 2018-05-10
Avoid using anonymous classes in some situations

smarter:
- They make debugging harder since it's not always clear what
  `Symbols$$anon$3` refes to
- If they take an outer reference as parameter, we currently generate
  inefficient code for them due to #4502


smarter:
test performance please

PR #4504, opened 2018-05-10
Try to refine handling of SymDenotations

odersky:
It seems we get many cache misses when going from Symbols to their denotations. This is a very common operation: When compiling typer/*.scala, we generate ~80K symbols, ~125K SymDenotations, yet dereference .denot more than 15.8M times. Of these, 11.7M go to the symbol's initial denotation. 

So one thing I'd like to try is to store the initial denotation of a symbol in the symbol itself, by making Symbol inherit from SymDenotation. To get there, several obstacles need to be put out of the way first:

 - ClassDenotation needs to be a trait.
 - Eliminate the implicit conversions from Symbol to SymDenotation, replacing them with
   forwarders in Symbol.
 - Rename all SymDenotation methods so that they are different from Symbol methods (otherwise we'd get accidental overrides).

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
+++ b/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
@@ -61,11 +61,11 @@ object CollectEntryPoints{
     val StringType = d.StringType
     // The given class has a main method.
     def hasJavaMainMethod(sym: Symbol): Boolean =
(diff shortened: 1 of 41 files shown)


odersky:
Abandoned, since the following proved much too burdensome:

 - Rename all SymDenotation methods so that they are different from Symbol methods (otherwise we'd get accidental overrides).

This is literally many thousands of places where we need to invent new names.

PR #4507, opened 2018-05-10
Turn SymbolLoaders into an object

smarter:
The extra flexibility of a class isn't useful here and leads to worse
code being generated (e.g. it requires PackageLoader to carry an outer
accessor).

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
+++ b/compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
@@ -19,17 +19,14 @@ import parsing.Parsers.OutlineParser
 import reporting.trace
 
 object SymbolLoaders {
+  import ast.untpd._
+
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/core/SymbolLoaders.scala
```diff
@@ -19,15 +19,12 @@ import parsing.Parsers.OutlineParser
...
 object SymbolLoaders {
+  import ast.untpd._
+
   /** A marker trait for a completer that replaces the original
    *  Symbol loader for an unpickled root.
    */
   trait SecondCompleter
-}
-
-/** A base class for Symbol loaders with some overridable behavior  */
```
allanrenucci: It seems that `SymbolLoaders` was meant to be extended. Make the `protected` members `private`?

PR #4508, opened 2018-05-10
Micro-optimize SymbolLoader#complete

smarter:
Avoid checking a setting value in a hot path.


smarter:
Other uses of -YdebugTrace already required Config.tracingEnabled, it's not an option you want to use without checking what it does anyway.

PR #4511, opened 2018-05-11
WIP: Narrow typers error positions

Blaisorblade:
Suggested by @smarter. Gotta review and fix all the testcases. Needed to unblock #4385, which right now has too much potential to hide stack overflows.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
@@ -16,7 +16,7 @@ object ErrorReporting {
   import tpd._
 
   def errorTree(tree: untpd.Tree, msg: => Message)(implicit ctx: Context): tpd.Tree =
-    tree withType errorType(msg, tree.pos)
+    tree withType errorType(msg, tree.pos.focus)
 
   def errorType(msg: => Message, pos: Position)(implicit ctx: Context): ErrorType = {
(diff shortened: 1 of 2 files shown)


odersky:
@Blaisorblade Can we get this in now please? We are blocked on a number of issues reporting stackoverflows until this in.

compiler/src/dotty/tools/dotc/typer/ErrorReporting.scala
```diff
@@ -16,7 +16,7 @@ object ErrorReporting {
   import tpd._
 
   def errorTree(tree: untpd.Tree, msg: => Message)(implicit ctx: Context): tpd.Tree =
-    tree withType errorType(msg, tree.pos)
+    tree withType errorType(msg, tree.pos.focus)
```
smarter: I'm not so sure anymore that this is a good idea, since it can lead to worse error messages

Blaisorblade: We only desired this effect for one call site: https://github.com/lampepfl/dotty/pull/4385/commits/0f2d1a3aa1d5d2c54f7a8664de3063577e40e871. So closing this PR.

Medium-term, I think you'd want positions inside `TypeError`, but that shouldn't block #4385.

Blaisorblade:
Closing. The blocker is in fact #4385, which I just unblocked otherwise.

PR #4512, opened 2018-05-11
Try to refine handling of SymDenotations V2

odersky:
It seems we get many cache misses when going from Symbols to their denotations. This is a very common operation: When compiling typer/*.scala, we generate ~80K symbols, ~125K SymDenotations, yet dereference .denot more than 15.8M times. Of these, 11.7M go to the symbol's initial denotation.

After closing #4504, this PR tries something else: Cache the hottests fields of denotations
in symbols, so that we can get at them without a dereference. 

We don't need to introduce forwarders in Symbols to do this (although it might be a good idea for other reasons), and we don't need to have Symbol inherit from SymDenotation (which has proven to be a bad idea).

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
+++ b/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
@@ -61,11 +61,11 @@ object CollectEntryPoints{
     val StringType = d.StringType
     // The given class has a main method.
     def hasJavaMainMethod(sym: Symbol): Boolean =
-      (toDenot(sym).info member nme.main).alternatives exists(x => isJavaMainMethod(x.symbol))
+      (sym.info member nme.main).alternatives exists(x => isJavaMainMethod(x.symbol))
 
     def fail(msg: String, pos: Position = sym.pos) = {
       ctx.warning(          sym.name +
-        s" has a main method with parameter type Array[String], but ${toDenot(sym).fullName} will not be a runnable program.\n  Reason: $msg",
+        s" has a main method with parameter type Array[String], but ${sym.fullName} will not be a runnable program.\n  Reason: $msg",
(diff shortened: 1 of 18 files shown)


odersky:
test performance please

odersky:
Current status: Symbol.denot calls reduced from 15.8M to 6.4M when compiling typers/*.scala

smarter:
> The previous performance test have a 404. What could be the reason for this?

This happens randomly sometimes, I don't know if @liufengyun knows the root cause, but the workaround is to add index.html at the end: http://dotty-bench.epfl.ch/4512/index.html

odersky:
Rebased to master

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

odersky:
test performance please

allanrenucci:
There is probably some rebase breakage

odersky:
test performance please

liufengyun:
The performance failure is fixed, the result is here: http://dotty-bench.epfl.ch/4512

odersky:
So it seems there's a significant speedup for dotty itself (about 3% according to latest benchmark), and no speedups for the other tests. I'll run one more test, and then we can decide what to do with this.

odersky:
test performance please

allanrenucci:
I see a non negligible slow down for scalap and re2.

Isn't the speedup on Dotty legitimate? My guess is that the speedup comes from the changes in the Dotty code base and not from a faster compiler . By adding all the cached SymDenotation facades to `Symbol`, you significantly reduce the number of implicit conversions from `Symbol` to `SymDenotation`.

odersky:
No clear win, except for dotty. Let's cherry-pick any improvements that make sense.

PR #4516, opened 2018-05-11
Fix #4498: Go to definition in lifted expressions

### LINKED ISSUE #4498: IDE: Cannot navigate from left operand of a right-associative infix op
In `dotc.Compiler`, try to navigate from the phases list to the individual phases. E.g. goto definition on "ElimRepeated" gives "no definition found". However, if I copy the `ElimRepeated` identifier outside the list, navigation works as expected.

Duhemm:
Consider the following snippet:

```scala
val lst = new A :: Nil
```

This is rewritten as:

```scala
val lst = {
  val x$0 = new A
  Nil.::(x$0)
}
```

During lifting, we need to assign positions to the trees. We were
assigning the position of `new A` in the original code to `x$0` in
`Nil.::(x$0)`, and the original position of `new A` to `val x$0 = new A`
with a zero-extent.

Still considering this snippet, this means that when trying to go to the
definition on `new A`, the IDE would think that the user tried to go to
the definition of `x$0`, because the IDE would look for the matching
tree by position.

This commit changes this so that the lifted definition
(`val x$0 = new A`) gets the position of the original expression, and
the reference gets a synthetic, zero-extent position. With this change,
the IDE correctly finds the right tree when selecting `new A` in the
original code.

Fixes #4498

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/EtaExpansion.scala
+++ b/compiler/src/dotty/tools/dotc/typer/EtaExpansion.scala
@@ -49,8 +49,8 @@ abstract class Lifter {
       var liftedType = fullyDefinedType(expr.tpe.widen, "lifted expression", expr.pos)
       if (liftedFlags.is(Method)) liftedType = ExprType(liftedType)
       val lifted = ctx.newSymbol(ctx.owner, name, liftedFlags, liftedType, coord = positionCoord(expr.pos))
-      defs += liftedDef(lifted, expr).withPos(expr.pos.focus)
-      ref(lifted.termRef).withPos(expr.pos)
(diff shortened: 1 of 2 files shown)


Duhemm:
@smarter I think you're the positions expert, can you take a look at this patch, please?

smarter:
> Still considering this snippet, this means that when trying to go to the
definition on new A, the IDE would think that the user tried to go to
the definition of x$0

Interesting, I didn't think of this. We should check what happens when going to definition on accessors created by the Inliner because I used a similar pattern there: https://github.com/lampepfl/dotty/commit/87ffb886a21b96c674c6dc9a04376f21acfc7135

Issue #4514, opened 2018-05-11
Crash when pickling buggy macro implementation 


nicolasstucki:
```scala
object Foo {
  inline def foo[X](x: X): Unit = ~fooImpl('(x))
  def fooImpl[X: quoted.Type](x: X): quoted.Expr[Unit] = '()
}
```

Note that this code should have `(x: Expr[X])` in `fooImpl`.

It crashes with:
```
Exception in thread "main" java.lang.AssertionError: assertion failed: unresolved symbols: type X when pickling Foo.scala
	at scala.Predef$.assert(Predef.scala:219)
	at dotty.tools.dotc.core.tasty.TreePickler.pickle(TreePickler.scala:639)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickle(PickledQuotes.scala:98)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickleQuote(PickledQuotes.scala:31)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickleAsTasty$1(ReifyQuotes.scala:401)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickledQuote(ReifyQuotes.scala:416)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.quotation(ReifyQuotes.scala:389)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.$anonfun$transform$2(ReifyQuotes.scala:547)
	... 
```

nicolasstucki:
Possibly caused by #4493

PR #4518, opened 2018-05-11
Fix #4493 and #4514

### LINKED ISSUE #4493: Failure to search for implicit type tag inside ~ of macro
Both
```scala
class Index[K]
object Index {
  inline def succ[K]: Unit = ~{
    implicit val k = '[K]
    '(new Index[K])
  }
}
```

```scala
class Index[K]
object Index {
  inline def succ[K]: Unit = ~{
    '(new Index[K])
  }
}
```

fail with 
```
Exception in thread "main" java.lang.AssertionError: assertion failed: unresolved symbols: type K when pickling tests/run/tasty-indexed-map/quoted_1.scala
	at dotty.tools.dotc.core.tasty.TreePickler.pickle(TreePickler.scala:639)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickle(PickledQuotes.scala:98)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickleQuote(PickledQuotes.scala:31)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickleAsTasty$1(ReifyQuotes.scala:401)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickledQuote(ReifyQuotes.scala:404)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.quotation(ReifyQuotes.scala:389)
	....
```

nicolasstucki:
Use captured type tags for macro type parameters. Before this change we just kept the type and failed when we tied to pickle it.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
@@ -120,7 +120,7 @@ class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
      *
      *  See `isCaptured`
      */
-    val capturers = new mutable.HashMap[Symbol, RefTree => Tree]
+    val capturers = new mutable.HashMap[Symbol, Tree => Tree]
   }
 
   /** The main transformer class
@@ -143,6 +143,9 @@ class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
 
     /** We are in a `~(...)` context that is not shadowed by a nested `'(...)` */
     def inSplice = outer != null && !inQuote
+
+    /** We are not in a `~(...)` or a `'(...)` */
+    def isRoot = outer != null
 
     /** A map from type ref T to expressions of type `quoted.Type[T]`".
      *  These will be turned into splices using `addTags` and represent type variables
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -144,6 +144,9 @@ class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
     /** We are in a `~(...)` context that is not shadowed by a nested `'(...)` */
     def inSplice = outer != null && !inQuote
 
+    /** We are not in a `~(...)` or a `'(...)` */
+    def isRoot = outer != null
```
biboudis: Hmm. I am confused. Isn't this root if the outer *is* `null` according to line 128 above? 

(also a typo in the comment of the description of the parameter `outer` )

nicolasstucki: Actually it is a typo. It should be `isNotRoot`

nicolasstucki: Or I will simplify it to an actual `isRoot` and add the not at call site.

biboudis: Good to go, then. Thanks!

PR #4524, opened 2018-05-12
Perform substitutions lazily in TypeComparer

odersky:
This could be a win if many methodic types are compared in isAsGood.


odersky:
test performance please

odersky:
Did not seem to help.

Issue #4431, opened 2018-05-01
Compiler crash when quoting an inline argument of function type


TomasMikula:
```scala
$ dotty-0.8.0-RC1/bin/dotr
Starting dotty REPL...
scala> inline def h(inline f: Int => String): String = ~ '(f(42)) 

Exception in thread "main" java.lang.AssertionError: assertion failed: unresolved symbols: value f$1 when pickling rs$line$1
	at dotty.DottyPredef$.assertFail(DottyPredef.scala:36)
	at dotty.tools.dotc.core.tasty.TreePickler.pickle(TreePickler.scala:634)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickle(PickledQuotes.scala:98)
	at dotty.tools.dotc.core.quoted.PickledQuotes$.pickleQuote(PickledQuotes.scala:31)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickleAsTasty$1(ReifyQuotes.scala:394)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.pickledQuote(ReifyQuotes.scala:413)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.quotation(ReifyQuotes.scala:382)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.op1$1(ReifyQuotes.scala:547)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.transform(ReifyQuotes.scala:537)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.transformWithCapturer$2(ReifyQuotes.scala:472)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.body$4(ReifyQuotes.scala:499)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.$anonfun$4(ReifyQuotes.scala:505)
	at dotty.tools.dotc.ast.tpd$.polyDefDef(tpd.scala:229)
	at dotty.tools.dotc.ast.tpd$.DefDef(tpd.scala:191)
	at dotty.tools.dotc.ast.tpd$.Closure(tpd.scala:103)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.makeLambda(ReifyQuotes.scala:505)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.split(ReifyQuotes.scala:523)
[...]

PR #4527, opened 2018-05-14
Document meaning of constant expressions

Blaisorblade:
Clarification following #4431.

The error checking's done by `Checking.checkInlineConformant(tree, isFinal =
false, "argument to inline parameter")`, and `checkInlineConformant` uses
`ConstantType` which characterizes constant expressions.

@tomasmikula opinions? (Since you asked what "constant expression" means).

### DIFF
--- a/docs/docs/reference/inline.md
+++ b/docs/docs/reference/inline.md
@@ -29,7 +29,8 @@ inlined at the point of use. Example:
 The `Config` object contains a definition of an `inline` value
 `logging`. This means that `logging` is treated as a constant value,
 equivalent to its right-hand side `false`. The right-hand side of such
-an inline val must itself be a constant expression. Used in this way,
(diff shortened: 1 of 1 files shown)


docs/docs/reference/inline.md
```diff
@@ -124,6 +125,13 @@ it in backticks, i.e.
 
     @`inline` def ...
 
+### The definition of constant expression
+
+Right-hand sides of inline values and of arguments for inline parameters
+must be constant expressions in the sense defined by the [SLS §
+6.24](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#constant-expressions)),
```
nicolasstucki: You have one `)` to many.

TomasMikula:
Seems sufficient as a documentation.

Issue #4470, opened 2018-05-06
Compiler crash with repeated enum


Blaisorblade:
Source:
```scala
enum Maybe[A] {
  case Foo
}

enum Maybe[A] {
  case Foo
}
```
gives

```
-- Error: /Users/pgiarrusso/git/dotty-example-project/src/main/scala/playground/IDETest.scala:5:5
5 |enum Maybe[A] {
  |^
  |class Maybe has already been compiled once during this run
6 |  case Foo
7 |}
-- Error: /Users/pgiarrusso/git/dotty-example-project/src/main/scala/playground/IDETest.scala:2:7
2 |  case Foo
  |  ^^^^^^^^
  |  cannot determine type argument for enum parent class Maybe,
  |  type parameter type A is non variant
exception occurred while typechecking /Users/pgiarrusso/git/dotty-example-project/src/main/scala/playground/IDETest.scala
exception occurred while compiling /Users/pgiarrusso/git/dotty-example-project/src/main/scala/playground/IDETest.scala
Exception in thread "main" java.lang.AssertionError: NoDenotation.owner
	at dotty.tools.dotc.core.SymDenotations$NoDenotation$.owner(SymDenotations.scala:1899)
	at dotty.tools.dotc.core.SymDenotations$SymDenotation.typeRef(SymDenotations.scala:1164)
	at dotty.tools.dotc.ast.DesugarEnums$.enumScaffolding(DesugarEnums.scala:92)
	at dotty.tools.dotc.ast.DesugarEnums$.nextEnumTag(DesugarEnums.scala:182)
	at dotty.tools.dotc.ast.DesugarEnums$.expandSimpleEnumCase(DesugarEnums.scala:219)
	at dotty.tools.dotc.ast.DesugarEnums$.expandEnumModule(DesugarEnums.scala:199)
	at dotty.tools.dotc.ast.desugar$.moduleDef(Desugar.scala:654)
	at dotty.tools.dotc.ast.desugar$.defTree(Desugar.scala:773)
	at dotty.tools.dotc.typer.Namer.expand(Namer.scala:419)
	at dotty.tools.dotc.typer.Namer.$anonfun$index$18(Namer.scala:683)
	at dotty.tools.dotc.typer.Namer.$anonfun$index$18$adapted(Namer.scala:683)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at dotty.tools.dotc.typer.Namer.index(Namer.scala:683)
[...]

PR #4530, opened 2018-05-15
Fix compiler crash for repeated Enums

snadorp:
Fix: https://github.com/lampepfl/dotty/issues/4470


OlivierBlanvillain:
@smarter wdyt?

smarter:
- Doing the check in expandSimpleEnumCase is too late since it doesn't cover enum cases with parents:
```scala
enum Maybe[T] {
  case Foo extends Maybe[Int]
}

enum Maybe[T] {
  case Foo extends Maybe[Int]
}
```
- we probably should add an `assert(ctx.reporter.errorsReported)` since we should never get to this case unless the "class has already been compiled" error has been reported.

snadorp:
@smarter Sadly `assert(ctx.reporter.errorsReported)` does not work.
I've added your example as test case and fixed the crash.

compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
```diff
@@ -87,14 +87,17 @@ object DesugarEnums {
...
-    val privateValuesDef =
-      ValDef(nme.DOLLAR_VALUES, TypeTree(),
-             New(TypeTree(defn.EnumValuesType.appliedTo(enumClass.typeRef :: Nil)), ListOfNil))
-        .withFlags(Private)
-    val valueOfDef = enumDefDef("enumValue", "fromInt")
-    val withNameDef = enumDefDef("enumValueNamed", "fromName")
-    val valuesDef = enumDefDef("enumValues", "values")
-    List(privateValuesDef, valueOfDef, withNameDef, valuesDef)
+
+    if (enumClass.exists) {
```
smarter: Instead of duplicating the check, you should be able to have only one check done earlier, in expandEnumModule

tests/neg/enum-repeated-4470.scala
```diff
@@ -0,0 +1,10 @@
+object RepeatedEnum {
```
allanrenucci: Our convention is to name our regression test files like `i4470.scala`. If you have mutiple test cases: `i4470a.scala`, `i4470b.scala`

snadorp: Alrighty, will rename the files thanks for the hint.

snadorp:
@allanrenucci will look into it.

compiler/src/dotty/tools/dotc/ast/DesugarEnums.scala
```diff
@@ -64,7 +64,8 @@ object DesugarEnums {
   }
 
   /** A type tree referring to `enumClass` */
-  def enumClassRef(implicit ctx: Context) = TypeTree(enumClass.typeRef)
+  def enumClassRef(implicit ctx: Context) =
+    if (enumClass.exists) TypeTree(enumClass.typeRef) else TypeTree()
```
snadorp: By retruning an empty TypeTree here, calls to `enumClassRef` are made safe even from other classes.

snadorp:
From my perspective there's no simple way of getting rid of the 2 `if`s in `expandEnumModule` and `expandSimpleEnumCase` as they are both called from `Desugar` on different occasions.
Expanding only in cases when there's no compiler error breaks a couple of unit test and is therefore not the wanted behaviour.
Any other ideas?

allanrenucci:
Thanks @snadorp 🎉

PR #4531, opened 2018-05-15
Fix #4405 REPL highlight for StringLiteral

### LINKED ISSUE #4405: REPL highlights strings incorrectly when multiple non-null characters are in front of the string
image
image

fabianpage:
Fix #4405.

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
+++ b/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
@@ -72,7 +72,7 @@ object SyntaxHighlighting {
         // Interpolation prefixes are a superset of the keyword start chars
         val (prefix, after) = remaining.span(interpolationPrefixes.contains)
         if (after.startsWith("\"")) {
-          newBuf ++= (n +: prefix)
+          newBuf += n ++= prefix
           prev = prefix.lastOption.getOrElse(n)
           if (remaining.nonEmpty) takeChars(prefix.length + 1) // drop 1 for appendLiteral
           appendString('"', after.startsWith("\"\"\""), true)
--- a/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
+++ b/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
@@ -70,12 +70,12 @@ object SyntaxHighlighting {
       val n = takeChar()
       if (interpolationPrefixes.contains(n)) {
         // Interpolation prefixes are a superset of the keyword start chars
-        val next = remaining.take(3).mkString
-        if (next.startsWith("\"")) {
-          newBuf += n
-          prev = n
-          if (remaining.nonEmpty) takeChar() // drop 1 for appendLiteral
-          appendString('"', next == "\"\"\"")
+        val (prefix, after) = remaining.span(interpolationPrefixes.contains)
+        if (after.startsWith("\"")) {
+          newBuf ++= (n +: prefix)
+          prev = prefix.lastOption.getOrElse(n)
+          if (remaining.nonEmpty) takeChars(prefix.length + 1) // drop 1 for appendLiteral
+          appendString('"', after.startsWith("\"\"\""), true)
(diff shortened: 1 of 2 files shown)


fabianpage:
I sign the CLA

allanrenucci:
@fabianpage Can you please add test cases to `SyntaxHighlightingTests`. Some suggestions:
```scala
s"Hello"
s"Hello $name!"
raw"Hello"
```

fabianpage:
@allanrenucci I think these 3 tests are enough or do you think it need's some more?

compiler/test/dotty/tools/dotc/printing/SyntaxHighlightingTests.scala
```diff
@@ -50,6 +50,9 @@ class SyntaxHighlightingTests {
   def strings = {
     // For some reason we currently use literal color for string
     test("\"Hello\"", "<L|\"Hello\">")
+    test("s\"Hello\"", "s<L|\"Hello\">")
+    test("s\"Hello $name\"", "s<L|\"Hello <V|$name<L|\">")
+    test("raw\"Hello\"", "raw<L|\"Hello\">")
```
allanrenucci: Let's add one more with triple quotes: `s"""Hello"""`

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -66,16 +60,44 @@ object SyntaxHighlighting {
...
+
+    def opChar(c:Char) = (c: @switch) match {
+      case '~' | '!' | '@' | '#' | '%' |
+           '^' | '*' | '+' | '-' | '<' |
+           '>' | '?' | ':' | '=' | '&' |
+           '|' | '\\' | '/' => true
+      case _ => false
+    }
+
+    def tryInterpolatorPrefix(n: Char, remaining: Stream[Char]): Option[String] = {
```
allanrenucci: I don't fully understand what you are doing below. Could you explain in detail?

The previous solution didn't work because it assumed that the interpolator prefix was on length 1.  So we just need to consume characters as long as they are valid prefixes. Something like:
```scala
val (prefix, after) = remaining.span(interpolationPrefixes.contains)
if (after.startsWith("\"")) {
   ...
```

fabianpage: The problem was that it also didn't supported string literals with symbols in it. something like a_+""

fabianpage:
@allanrenucci back from Berlin. Should i continue working on this PR or will #4537 be merged soon and my change is not needed any more?

allanrenucci:
Not sure when #4537 will be merged. But yes, the fix in it is more generic and it will replace all the actual code that does syntax highlighting.

fabianpage:
@allanrenucci Then i will try to finish this PR. Should i try to change the existing code to only support longer interpreter suffix?

allanrenucci:
> Should i try to change the existing code to only support longer interpreter suffix?

Sounds good! Then we can get this PR merged before #4537

fabianpage:
@allanrenucci What do you think? I think this way there are only minimal changes.

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -70,12 +70,12 @@ object SyntaxHighlighting {
...
         // Interpolation prefixes are a superset of the keyword start chars
-        val next = remaining.take(3).mkString
-        if (next.startsWith("\"")) {
-          newBuf += n
-          prev = n
-          if (remaining.nonEmpty) takeChar() // drop 1 for appendLiteral
-          appendString('"', next == "\"\"\"")
+        val (prefix, after) = remaining.span(interpolationPrefixes.contains)
+        if (after.startsWith("\"")) {
+          newBuf ++= (n +: prefix)
```
allanrenucci: This will create an intermediate collection, prefer:
```scala
newBuf += n ++= prefix
```

PR #4532, opened 2018-05-15
Do not print class fields

dieutth:
* Partly fix i4526


nicolasstucki:
Thanks @dieutth!

dieutth:
Your guidance was really inspiring. Thanks a lot.
May the Fourth-square be with us :D

PR #4537, opened 2018-05-16
Reimplement SyntaxHighlighter using Scanner and Parser

slothspot:
**WORK IN PROGRESS**

Problems
- [x] annotations are not supported
- [ ] comments are not supported
- [ ] tests are broken
- [ ] list of keywords which should not be highlighted is not defined
- [ ] something else I forgot 

Nice parts:
- implementation is simpler than the current one
- #4405 was fixed here

Next steps:
fix tests and reach features parity with current highlighter implementation

@allanrenucci I would appreciate your feedback on the current state of this experiment and clarification on comments highlighting and list of keywords which shouldn't be highlighted


compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -355,4 +365,88 @@ object SyntaxHighlighting {
...
+    }
+
+    val sb = new mutable.StringBuilder()
+
+    while(s.token != EOF) {
+      val isKwd = isKeyword(s.token) && !ignoredKwds.contains(s.name)
+      val offsetStart = s.offset
+
+      if(s.token == IDENTIFIER && s.name == nme.???) {
+        highlightRange(Position(s.offset, s.offset + s.name.length), Console.RED_B)
```
allanrenucci: define `highlightRange` and `highlightPosition` as:
```scala
def highlightRange(from: Int, to: Int, color: String) = ...
def highlightPosition(pos: Postition, color: String) =
  if (pos.exists) highlightRange(pos.start, pos.end, color)

```

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -355,4 +365,88 @@ object SyntaxHighlighting {
...
+      if(idx == 0 || outputH(idx-1) != outputH(idx)){
+        sb.append(outputH(idx))
+      }
+      sb.append(in(idx))
+    }
+    if(outputH.last != NoColor) {
+      sb.append(NoColor)
+    }
+
+    sb.mkString
```
allanrenucci: Even if it does the same thing, we prefer calling `toString` on `StringBuilder`

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -355,4 +365,88 @@ object SyntaxHighlighting {
...
+  private val ignoredKwds = Seq(nme.ARROWkw, nme.EQ, nme.EQL, nme.COLONkw)
+
+  def highlight(in: String)(ctx0: Context): String = {
+    import dotty.tools.dotc.ast.untpd._
+
+    implicit val ctx: Context = ctx0.fresh.setReporter(new NoReporter)
+
+    val sf = new SourceFile("<highlighting>", in.toCharArray)
+    val p = new Parser(sf)
+    val s = new Scanner(sf)
```
allanrenucci: Move the scanner declaration closer to use site.

allanrenucci:
> 1/ How to match annotations? (Annotated node type doesn't work for me)

You should be able to match on `MemberDef`. Then you can do `.rawMods.annotations`. Another (more limited) possibility would be to use the scanner and highlight `Tokens.AT` as well as the following identifier.

> 2/ What to add to the list of excluded keywords?

In `Tokens.scala`, we have `keywords = alphaKeywords + symbolicKeywords`, we could decide to only highlight `alphaKeywords`. So instead of doing `isKeyword(scanner.token)`, you could do `alphaKeywords.contains(scanner.token)`

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -355,4 +365,89 @@ object SyntaxHighlighting {
...
+
+    val colorAt = Array.fill(in.length)(NoColor)
+
+    def highlightRange(from: Int, to: Int, color: String) = {
+      try {
+        for (i <- from until to)
+          colorAt(i) = color
+      } catch {
+        case _: IndexOutOfBoundsException =>
+          ctx.error("Encountered tree with invalid position, please open an issue with the code snippet that caused the error")
```
allanrenucci: `ctx.error` will not print anything since your are using a dummy reporter. Simply use `println`

allanrenucci:
@slothspot Let's not bother about annotations in this PR. This can be fixed in a separate PR.

I'd like to get this in and close all bugs related to syntax highlighting. Can you update the tests cases and add `@Ignore` on the tests for annotations and comments that are currently failing.

slothspot:
Preliminary support for annotations was added.
Have next problem:
this `@deprecated class Zoo` matches to `ValOrDefDef` with name `<init>` and `namePos` pointing to range [0..6] which results in first half of `@deprecated` highlighted with `ValDefColor`.

Questions:
1/ Shouldn't this be matched with `TypeDef`?
2/ Shouldn't `namePos` be `<no position>` in this case?

slothspot:
And due to your last request:
1/ I may mark annotations and comments related tests as ignored and create separate PR to handle these.
2/ Is it ok to use LiteralColor for strings?
3/ I'm not happy with interpolated strings:
for s"Some str with $v inside" current implementation gives:
`s<L|"Some str with> <V|$v> $<L|inside">`
but new one gives:
`s<L|"Some str with $>v <L|inside">`
May try to fix this with this PR or create separate one

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -355,4 +365,93 @@ object SyntaxHighlighting {
...
+      val offsetStart = scanner.offset
+
+      if (scanner.token == IDENTIFIER && scanner.name == nme.???) {
+        highlightRange(scanner.offset, scanner.offset + scanner.name.length, Console.RED_B)
+      }
+      scanner.nextToken()
+
+      if (isKwd) {
+        val offsetEnd = scanner.lastOffset
+        highlightPosition(Position(offsetStart, offsetEnd), KeywordColor)
```
allanrenucci: `highlightRange`?

compiler/test/dotty/tools/dotc/printing/SyntaxHighlightingTests.scala
```diff
@@ -50,11 +50,33 @@ class SyntaxHighlightingTests {
...
+
+    val expected =
+      """
+        |<T|@deprecated>
+        |<K|class> <T|Foo> {
+        |  <T|@inline> <K|val> <V|bar> = <L|42>
+        |}
+      """.stripMargin
+
+    test(source, expected)
```
allanrenucci: Prefer simple single line test case:
```scala
test("@inline val bar = 42", ...)
```

compiler/test/dotty/tools/dotc/printing/SyntaxHighlightingTests.scala
```diff
@@ -89,12 +111,12 @@ class SyntaxHighlightingTests {
 
   @Test
   def unionTypes = {
```
allanrenucci: Let remove this test and move it to `def types`. Also I don't think we need all these combinations. Let's keep:
```scala
test("type A = String|Int", "<K|type> <T|A> = <T|String|Int>")
test("type B = String & Int", "<K|type> <T|A> = <T|String & Int>")
test("fn[String](input)", "fn[<T|String>](input)")
```

compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala
```diff
@@ -2,10 +2,20 @@ package dotty.tools
...
+
 import scala.annotation.switch
 import scala.collection.mutable.StringBuilder
-import util.Chars
+import util.{Chars, SourceFile}
+
+import scala.collection.mutable
 
 /** This object provides functions for syntax highlighting in the REPL */
 object SyntaxHighlighting {
```
allanrenucci: Remove all the legacy code

allanrenucci:
> this @deprecated class Zoo matches to ValOrDefDef with name <init> and namePos pointing to range [0..6] which results in first half of @deprecated highlighted with ValDefColor.

You are matching the contstructor of the class. The position looks weird to me too. I will have to investigate. What you can try to do is not print synthetic positions: `if (pos.exists && pos.isSourceDerived) highlightRange(...)`

> I may mark annotations and comments related tests as ignored and create separate PR to handle these

Yes, let's handle annoations in a separate PR. Please remove the code for annotations. You can keep the tests and mark them as `Pending`.

> Is it ok to use LiteralColor for strings?

Yes, we can always change it later

> I'm not happy with interpolated strings:

You could special case the tree `InterpolatedString`. Let's do this in another PR

danielyli:
@smarter Added more tests that catch current bugs.

*  https://github.com/danielyli/dotty/commit/468942014ec82c21a903e0fba8ac1af1efbe308e
*  https://github.com/lampepfl/dotty/commit/d6444e37d739a5ee981352b4618764df89617b60

smarter:
@danielyli Thanks! Pushed to the PR.

allanrenucci:
I rebased the PR. The fix in d31f32ec6325c702cf6d265a82d4933ab677393d  is not correct. It breaks a lot of stuff. The compiler does not bootstrap anymore

allanrenucci:
Closed in favor of #5137. Thanks @slothspot for the original PR

PR #4538, opened 2018-05-17
Pickle quotes directly as expressions in TASTY

nicolasstucki:
* We avoid having to do owner changes after unpickling
* The pickled code is slightly smaller

Replaces #4242


nicolasstucki:

nicolasstucki:
And performance improvements all around :)

nicolasstucki:
And performance improvements all around :)

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -73,11 +73,18 @@ class TreeUnpickler(reader: TastyReader,
    */
   def enterTopLevel(roots: Set[SymDenotation])(implicit ctx: Context): Unit = {
     this.roots = roots
-    var rdr = new TreeReader(reader).fork
+    val rdr = new TreeReader(reader).fork
     ownerTree = new OwnerTree(NoAddr, 0, rdr.fork, reader.endAddr)
     rdr.indexStats(reader.endAddr)
   }
 
+  def unpickleExpr()(implicit ctx: Context): Tree = {
```
smarter: Document what this method does, how it differs from unpickle(), when it is used

smarter:
I wonder if we could always unpickle as an expression and get rid of the other way to unpickle.

nicolasstucki:
Do you mean get rid of `enterTopLevel`?

nicolasstucki:
I will merge the two unpicking ways in another PR

PR #4543, opened 2018-05-17
[WIP] Safe initialization for Scala

liufengyun:
This is a joint attempt with @biboudis .

Currently, zero change to typer and types.

```Scala
import scala.annotation.{ partial, init }


class Parent(x: Int) {
  var name: String = _
  var addr: String = _                       // warning: addr is not initialized

  val len = name.size                        // warning: name not initialized
  lazy val l1 = name.size                    // ok: l1 is lazy
  lazy val l2 = addr.size                    // warning: l2 is forced at L32 before `addr` is initialized

  val fun: Int => Int = n => n + list.size   // ok, fun is a partial value
  val bar = new Bar("bar", fun)              // ok, Bar accepts partial value
  bar.result                                 // warning: bar is a partial value

  val child = new Child(this)                // warning: `this` is partial, while full value expected
  List(5, 9).map(n => n + list.size)         // warning: partial value used as full value

  f(20)                                      // warning: list not initialized

  val list = List(1, 2, 3)

  if (x > 5) {
    name = "big"
    addr = "Lausanne"
  }
  else {
    name = "small"
  }

  val temp1 = l1                               // ok, name init
  val temp2 = l2                               // warning: addr not initialized

  // f is called during initialization, needs to be `private` or `final`.
  def f(m: Int) =
    m + list.size                              // warning: `f` is called at L19 before `list` is initialized

  List(1, 3, 5).map(n => n + list.size)        // ok, `this.list` already initialized
}


class Bar(val name: String, fun: (Int => Int) @partial) {
  def result = fun(20)
}

class Child(parent: Parent) {
  println(parent.name)
}
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/Driver.scala
+++ b/compiler/src/dotty/tools/dotc/Driver.scala
@@ -40,10 +40,13 @@ class Driver {
       }
     else ctx.reporter
 
+  @scala.annotation.init
   protected def initCtx: Context = (new ContextBase).initialCtx
 
+  @scala.annotation.init
   protected def sourcesRequired: Boolean = true
 
+  @scala.annotation.init
   def setup(args: Array[String], rootCtx: Context): (List[String], Context) = {
     val ctx = rootCtx.fresh
     val summary = CompilerCommand.distill(args)(ctx)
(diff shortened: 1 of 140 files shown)


liufengyun:
test performance please

liufengyun:
test performance please

liufengyun:
test performance please

liufengyun:
Close for now, more research is required to completely remove the annotation requirement.

PR #4545, opened 2018-05-17
Remove DotClass

smarter:


smarter:
test performance please

OlivierBlanvillain:
This does not seem to impact performance... What's the point of that DotClass? Should we merge anyway?

compiler/src/dotty/tools/package.scala
```diff
@@ -20,4 +20,8 @@ package object tools {
       }
     case _ => ys.isEmpty
   }
+
+  /** Throws an `UnsupportedOperationException` with the given method name. */
+  def unsupported(methodName: String): Nothing =
+    throw new UnsupportedOperationException(s"$getClass.$methodName")
```
allanrenucci: I guess `getClass` doesn't make much sense anymore

PR #4546, opened 2018-05-17
Join logic for unpickling top level and term tasty

nicolasstucki:


nicolasstucki:
Improvement proposed in #4538

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -71,24 +74,21 @@ class TreeUnpickler(reader: TastyReader,
...
     ownerTree = new OwnerTree(NoAddr, 0, rdr.fork, reader.endAddr)
-    rdr.indexStats(reader.endAddr)
-  }
-
-  def unpickleExpr()(implicit ctx: Context): Tree = {
-    this.roots = Set(ctx.owner)
-    val rdr = new TreeReader(reader).fork
-    ownerTree = new OwnerTree(NoAddr, 0, rdr.fork, reader.endAddr)
-    rdr.readTerm()
+    if (isTopLevel)
```
smarter: Can't we infer whether this is top level or not based on the tag of the first tree node?

compiler/src/dotty/tools/dotc/core/quoted/PickledQuotes.scala
```diff
@@ -110,7 +110,8 @@ object PickledQuotes {
   /** Unpickle TASTY bytes into it's tree */
   private def unpickle(bytes: Array[Byte], splices: Seq[Any])(implicit ctx: Context): Tree = {
     val unpickler = new TastyUnpickler(bytes, splices)
-    val tree = unpickler.unpickleExpr()
+    unpickler.enter(Set(ctx.owner))
```
smarter: Is entering ctx.owner actually useful here? Otherwise you could just do Set.empty

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -849,15 +845,15 @@ class TreeUnpickler(reader: TastyReader,
...
+        skipToplevel()
+      }
     }
 
+    def isTopLevel(implicit ctx: Context): Boolean =
+      nextByte == IMPORT || nextByte == PACKAGE
+
     def readTopLevel()(implicit ctx: Context): List[Tree] = {
       @tailrec def read(acc: ListBuffer[Tree]): List[Tree] = nextByte match {
         case IMPORT | PACKAGE =>
```
smarter: Could use isTopLevel here. Also I wonder if we couldn't simplify that to return a single Tree instead of a List[Tree] if we never pickle a list of trees.

nicolasstucki: We may have problems tasty files that have top-level imports

nicolasstucki: If I put a top level import I get

```scala
import scala.collection.mutable.TreeMap
...
```

```scala
     0: PACKAGE(80)
     2:   TERMREFpkg 1 [<empty>]
     4:   IMPORT(8)
     6:     SELECT 2 [mutable]
     8:       SELECT 3 [collection]
    10:         TERMREFpkg 4 [scala]
    12:     IMPORTED 5 [TreeMap]
    14:   PACKAGE(66)
    16:     TERMREFpkg 6 [foo]
    ...
```

same with 
```scala
package bar
import scala.collection.mutable.TreeMap
```

```scala
     0: PACKAGE(80)
     2:   TERMREFpkg 1 [bar]
     4:   IMPORT(8)
     6:     SELECT 2 [mutable]
     8:       SELECT 3 [collection]
    10:         TERMREFpkg 4 [scala]
    12:     IMPORTED 5 [TreeMap]
```

nicolasstucki: I think we may go further and do 

```scala
def isTopLevel(implicit ctx: Context): Boolean = 
  nextByte == PACKAGE
```

nicolasstucki:
Rebased

compiler/src/dotty/tools/dotc/core/quoted/TastyUnpickler.scala
```diff
@@ -15,10 +15,13 @@ object TastyUnpickler {
  *  @param bytes         the bytearray containing the Tasty file from which we unpickle
  *  @param splices       splices that will fill the holes in the quote
  */
-class TastyUnpickler(bytes: Array[Byte], splices: Seq[Any]) extends DottyUnpickler(bytes) {
+class TastyUnpickler(bytes: Array[Byte], splices: Seq[Any], isTypeTree: Boolean) extends DottyUnpickler(bytes) {
```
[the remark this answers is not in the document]
nicolasstucki: Yes. I will change it.

compiler/src/dotty/tools/dotc/core/tasty/DottyUnpickler.scala
```diff
@@ -40,14 +40,13 @@ class DottyUnpickler(bytes: Array[Byte]) extends ClassfileParser.Embedded with t
   private val posUnpicklerOpt = unpickler.unpickle(new PositionsSectionUnpickler)
   private val treeUnpickler = unpickler.unpickle(treeSectionUnpickler(posUnpicklerOpt)).get
 
+  protected def mode: TreeUnpickler.UnpickleMode = TreeUnpickler.UnpickleMode.TopLevel
```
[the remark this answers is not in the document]
nicolasstucki: Yes, it should

nicolasstucki:
Rebased

nicolasstucki:
@smarter don't forget to review this one.

compiler/src/dotty/tools/dotc/core/tasty/DottyUnpickler.scala
```diff
@@ -46,14 +46,13 @@ class DottyUnpickler(bytes: Array[Byte]) extends ClassfileParser.Embedded with t
   private val commentUnpicklerOpt = unpickler.unpickle(new CommentsSectionUnpickler)
   private val treeUnpickler = unpickler.unpickle(treeSectionUnpickler(posUnpicklerOpt, commentUnpicklerOpt)).get
 
+  protected val mode: TreeUnpickler.UnpickleMode = TreeUnpickler.UnpickleMode.TopLevel
```
smarter: Maybe instead of a field this could be a class parameter: `class DottyUnpickler(bytes: Array[Byte], mode: UnpickleMode = UnpickleMode.TopLevel)`

nicolasstucki: Good idea. Done.

compiler/src/dotty/tools/dotc/core/tasty/DottyUnpickler.scala
```diff
@@ -37,7 +38,7 @@ object DottyUnpickler {
 /** A class for unpickling Tasty trees and symbols.
  *  @param bytes         the bytearray containing the Tasty file from which we unpickle
  */
-class DottyUnpickler(bytes: Array[Byte]) extends ClassfileParser.Embedded with tpd.TreeProvider {
+class DottyUnpickler(bytes: Array[Byte], mode: UnpickleMode = UnpickleMode.TopLevel) extends ClassfileParser.Embedded with tpd.TreeProvider {
```
allanrenucci: Can you update the class documentation?

nicolasstucki: Added

PR #4547, opened 2018-05-19
Add regression test

nicolasstucki:
Same as we have to do when unpickling tasty into a compilation unit ([here](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/CompilationUnit.scala#L45))


nicolasstucki:

nicolasstucki:
The fix was in #4279. This PR add a regression test for it.

PR #4549, opened 2018-05-20
Drop requirement that implicit functions must be non-empty

odersky:
This removes an arbitrary restriction and is a first step towards unifying by-name types and implicit function types.

It's been trickier than anticipated since we needed a new mechanism to mark nullary implicit closures. Without parameters we cannot look at the parameter modifiers. Instead we produce an implicit function type as closure target type.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ExpandSAMs.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ExpandSAMs.scala
@@ -15,12 +15,14 @@ import dotty.tools.dotc.util.Positions.Position
  *  These fall into five categories
  *
  *   1. Partial function closures, we need to generate isDefinedAt and applyOrElse methods for these.
- *   2. Closures implementing non-trait classes.
+ *   2. Closures implementing non-trait classes
  *   3. Closures implementing classes that inherit from a class other than Object
  *      (a lambda cannot not be a run-time subtype of such a class)
  *   4. Closures that implement traits which run initialization code.
  *   5. Closures that get synthesized abstract methods in the transformation pipeline. These methods can be
  *      (1) superaccessors, (2) outer references, (3) accessors for fields.
+ *
+ *  However, implicit function types do not count as SAM types.
  */
 class ExpandSAMs extends MiniPhase {
   override def phaseName = "expandSAMs"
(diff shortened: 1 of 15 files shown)


compiler/src/dotty/tools/dotc/transform/ExpandSAMs.scala
```diff
@@ -32,9 +34,12 @@ class ExpandSAMs extends MiniPhase {
     ctx.platform.isSam(cls)
 
   override def transformBlock(tree: Block)(implicit ctx: Context): Tree = tree match {
-    case Block(stats @ (fn: DefDef) :: Nil, Closure(_, fnRef, tpt)) if fnRef.symbol == fn.symbol =>
+    case Block(stats @ (fn: DefDef) :: Nil, cl @ Closure(_, fnRef, tpt)) if fnRef.symbol == fn.symbol =>
```
nicolasstucki: `cl` is never used

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -761,14 +761,10 @@ class Typer extends Namer
...
-      case tree: untpd.NonEmptyFunction =>
-        if (args.nonEmpty) (tree.mods.is(Implicit), tree.mods.is(Erased))
-        else {
-          ctx.error(FunctionTypeNeedsNonEmptyParameterList(tree.mods.is(Implicit), tree.mods.is(Erased)), tree.pos)
-          (false, false)
-        }
+      case tree: untpd.FunctionWithMods => (tree.mods.is(Implicit), tree.mods.is(Erased))
       case _ => (false, false)
     }
+    if (isErased && args.isEmpty) ctx.error(em"empty function cannot not be erased", tree.pos)
```
nicolasstucki: This looks inconsistent. We do provide names for `ErasedFunction0` and `ErasedImplicitFunction0` and add them in scope, but we do not allow them to be used in function types.

* We could remove the start `ErasedFuctionN` and `ErasedImplicitFuctionN` at `N = 1`
* Or allow them, as a function type that is equivalent to `Funcion0` or `ImplicitFunction0`. We could even consider making them type aliases and warn the the `erased` keyword is useless.

odersky: @nicolasstucki Agree we should start ErasedFunctionN at 1. Can you do the change and push it on top of this PR? Thanks!

PR #4552, opened 2018-05-21
Closes #4300 adds better worded pattern match unreachability warning

### LINKED ISSUE #4300: Reachability warnings could be better worded
```scala
scala> def foo(a: Int) = a match { case x: Int => x } 
1 |def foo(a: Int) = a match { case x: Int => x }
  |                                 ^
  |this will always yield true, since `class Int` is a subclass of `class Int`
```
The reason given is a tautology, It would make more sense to say something like "this case will always be taken because the scrutinee `` `a` `` has type Int".
Also, the highlighted position is the position of `x`, but it would be more intuitive to highlight the position of `x: Int`.

zoltanelek:
- Used when isInstanceOf is called inside a match case
  - The new message is "this case will always be taken because the scrutinee has type `<type>`"
  - Highlights the position of `x: Int` instead of `x`

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2117,38 +2117,10 @@ object messages {
     val explanation: String = ""
   }
 
-  case class PatternMatchAlwaysSucceeds(foundCls: Symbol, testCls: Symbol)(implicit ctx: Context) extends Message(PatternMatchAlwaysSucceedsID) {
+  case class TypeTestAlwaysSucceeds(foundCls: Symbol, testCls: Symbol)(implicit ctx: Context) extends Message(TypeTestAlwaysSucceedsID) {
     val kind = "Syntax"
-    val msg: String =
-      hl"""
-         |The type check in the highlighted part of the pattern match case statement will always succeed,
-         |since `$foundCls` is a subclass of `$testCls`.
-       """
-    val codeExampleValidWarning = "def foo(a: Int) = a match { case x: Int => x }"
-    val codeExampleDisregardWarning =
-      hl"""
-        |object Foo {
-        |  def unapply(x: Int): Option[Int] = if (x % 2 == 0) Some(x) else None
-        |}
-        |
-        |class Test {
-        |  def foo(a: Int) = a match { case Foo(x: Int) => x }
-        |}
-      """
-    val explanation: String =
-      hl"""
-          |The highlighted typecheck will always match.
-          |For example:
-          |
-          |$codeExampleValidWarning
-          |
-          |The type of the scrutinee 'a' will always be 'Int', so a the 'case x: Int'
-          |match will always be successful. Maybe a guard clause was forgotten.
-          |
-          |In the following counter-example, because of the unapply method, the
-          |whole match will not be always successful, only the type check.
-          |$codeExampleDisregardWarning
-          |
-        """
+    val msg =
+      s"The highlighted type test will always succeed since the scrutinee type ($foundCls) is a subtype of ${testCls}"
+    val explanation = ""
   }
 }
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -2119,8 +2119,7 @@ object messages {
 
   case class TypeTestAlwaysSucceeds(foundCls: Symbol, testCls: Symbol)(implicit ctx: Context) extends Message(TypeTestAlwaysSucceedsID) {
     val kind = "Syntax"
-    val msg =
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/transform/TypeTestsCasts.scala
```diff
@@ -90,9 +90,14 @@ object TypeTestsCasts {
 
           if (expr.tpe <:< testType)
             if (expr.tpe.isNotNull) {
-              ctx.warning(
-                em"this will always yield true, since `$foundCls` is a subclass of `$testCls`",
-                expr.pos)
+              if (inMatch)
+                ctx.warning(
+                  em"this case will always be taken because the scrutinee has type `$testType`",
+                  tree.pos)
```
liufengyun: This error message doesn't make sense for the following test case:

```Scala
object Foo {
  def unapply(x: Int): Option[Int] = if (x % 2 == 0) Some(x) else None
}

class Test {
  def foo(a: Int) = a match { case Foo(x: Int) => x }
}
```

Blaisorblade: - To address both @zoltanelek and @liufengyun's concerns, I'd at least say "this pattern will always succeed", where "this" refers to `x: Int` (as shown by the position) — I'd prefer in fact "the highlighted pattern match".

I think both this message and the current one can still be a bit confusing because they might only refer to one pattern out of many nested ones, but maybe that can be clarified in the message explanation.

- I'd also please keep showing both `$foundCls` and `$testCls`, since in the presence of type aliases they might be nontrivial to deduce from the scrutinee type and the pattern match.

- Separately, "since `$foundCls` is a subclass of `$testCls`" says more than "because the scrutinee has type `$testType`": who wrote the code and fixes the warning might know the scrutinee's type, but what they probably missed is that `$foundCls` is a subclass of `$testCls`.

zoltanelek:
@Blaisorblade sure!

zoltanelek:
Hey guys! I went with 
 `the type check in the highlighted part of the pattern match case statement
will always succeed, since $foundCls is a subclass of $testCls`

I think this is much more meaningful when matching on tuples.
E.g. here: "the highlighted pattern match will always succeed" would be actually false, only the first parameter, x is matched
```
|def foo(a: (Int,Int)) = a match { case (x: Int, y: String) => x }
|                                        ^^^^^^
|the type check in the highlighted part of the pattern match case statement will always succeed, since `class Int` is a subclass of `class Int`
```

And 'the type check in the highlighted part' would not imply that there can't be an unapply behind the scenes, which will fail to match.
```
scala> object Foo {
           def unapply(x: Int): Option[Int] = if (x % 2 == 0) Some(x) else None
         } 
// defined object Foo

scala> class Test {
           def foo(a: Int) = a match { case Foo(x: Int) => x }
         } 
2 |    def foo(a: Int) = a match { case Foo(x: Int) => x }
  |                                         ^^^^^^
  |the type check in the highlighted part of the pattern match case statement will always succeed, since `class Int` is a subclass of `class Int`
// defined class Test
```

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2117,4 +2117,38 @@ object messages {
...
 
+  case class PatternMatchAlwaysSucceeds(foundCls: Symbol, testCls: Symbol)(implicit ctx: Context) extends Message(PatternMatchAlwaysSucceedsID) {
+    val kind = "Syntax"
+    val msg: String =
+      hl"""
+         |The type check in the highlighted part of the pattern match case statement will always succeed,
+         |since `$foundCls` is a subclass of `$testCls`.
+       """
+    val codeExampleValidWarning = "def foo(a: Int) = a match { case x: Int => x }"
+    val codeExampleDisregardWarning =
```
allanrenucci: The user should not disregard the warning. The warning is legitimate, he should have written `case Foo(x)` instead of `case Foo(x: Int)`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -2117,4 +2117,38 @@ object messages {
...
+      hl"""
+        |object Foo {
+        |  def unapply(x: Int): Option[Int] = if (x % 2 == 0) Some(x) else None
+        |}
+        |
+        |class Test {
+        |  def foo(a: Int) = a match { case Foo(x: Int) => x }
+        |}
+      """
+    val explanation: String =
```
allanrenucci: I would remove the current explanation. I don't think it gives you more information than the `msg`.

liufengyun:
@Blaisorblade @allanrenucci I think it looks good now, do you have more comments?

allanrenucci:
Thanks @zoltanelek!

zoltanelek:
Cool! Thanks for the help guys!

PR #4554, opened 2018-05-22
Fix #4540: Add hint of probable source of issue

### LINKED ISSUE #4540: Better error when macro call is used within the same compilation unit as macro definition.
When macro definition and macro call is written within the same compilation unit, compiler errors with:
`Could not find interpreted class Main in classpath`, see https://scastie.scala-lang.org/h8cEUZIER2mT6YBtl4wFlA as example. 
It'd be great to add a compiler hint as it was in scala.reflect macros: `error: macro implementation not found: printf (the most common reason for that is that
you cannot use macro implementations in the same compilation run that defines them)` and improve docs in http://dotty.epfl.ch/docs/reference/principled-meta-programming.html
See last paragraph of https://docs.scala-lang.org/overviews/macros/overview.html#a-complete-example for example

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/Splicer.scala
```diff
@@ -149,17 +149,19 @@ object Splicer {
       try clazz.getMethod(name.toString, paramClasses: _*)
       catch {
         case _: NoSuchMethodException =>
-          val msg = s"Could not find interpreted method ${clazz.getCanonicalName}.$name with parameters $paramClasses"
+          val msg = s"Could not find inline macro method ${clazz.getCanonicalName}.$name with parameters $paramClasses $extraMsg"
           throw new StopInterpretation(msg, pos)
       }
     }
 
+    private def extraMsg = "(the most common reason for that is that you cannot use inline macro implementations in the same compilation run that defines them)"
```
liufengyun: Maybe make this sentence shorter as it is in parenthesis. Or make it a separate sentence.

Something like "You defined and used macros in the same compilation run?"

nicolasstucki: Splitted into two sentences

nicolasstucki: I will keep this text as it is the same as for scala 2 macros andif they seach for it they might still get some useful insight on why they need to be in different compilation units.

PR #4555, opened 2018-05-22
Support `sourcesInBase` setting from sbt

Duhemm:
In sbt, the `sourcesInBase` setting configures whether the sources that are found directly in the base directory of a project are compiled within the `Compile` configuration. If this setting is set to true, Dotty IDE will add the `baseDirectory` to the list of source directories for the project. Note that sbt will just filter the sources that are immediate children of the base directory and add them to the `unmanagedSources`.

This means that we can now have source directories that are children of other source directories. We resolve conflicts here by selecting the longest matching directory.

@smarter I'm not sure this is a good idea. This solves an issue with multi-project builds and sources in the base directory, but still doesn't completely reproduce sbt's behavior (in sbt, `baseDirectory / somedirectory / Foo.scala` would NOT be considered as a source, but here it would). It may be better just to ignore this issue and wait for BSP to solve the problem for us in a more robust way.

### DIFF
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -104,16 +104,16 @@ class DottyLanguageServer extends LanguageServer
 
   /** The driver instance responsible for compiling `uri` */
   def driverFor(uri: URI): InteractiveDriver = {
-    val matchingConfig =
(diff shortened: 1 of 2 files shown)


smarter:
> @smarter I'm not sure this is a good idea

Same. I would rather not complicate the logic here and wait for a BSP implementation in sbt to use as a replacement for the stuff the plugin does.

PR #4556, opened 2018-05-22
Fix #4542: Enable Dotty LSP on `.scala` files only

### LINKED ISSUE #4542: Dotty IDE - sbt file get falsely red squiggly marks
The issue seems to be [reproducible](https://gitter.im/lampepfl/dotty?at=5afd88129ad18967b9f27329) with the sample project
`sbt new lampepfl/dotty.g8 `
followed the instructions at
http://dotty.epfl.ch/docs/usage/ide-support.html

vscode is yielding the folwing:

> expected class or object definition
> expected start of definition

referring to the sbt file.

Duhemm:
Previously, the Dotty language server would try to provide services for
all `.scala` and `.sbt` files, because both are associated to the Scala
language . The association from filename to language is provided by the
extension `daltonjorge.scala`, which associates all `.scala` and `.sbt`
files to the Scala language.

This commit explicitly states that the Dotty LSP should only be used for
`.scala` files only.

Fixes #4542.

### DIFF
--- a/vscode-dotty/src/extension.ts
+++ b/vscode-dotty/src/extension.ts
@@ -89,8 +89,8 @@ function fetchAndRun(artifact: string) {
 function run(serverOptions: ServerOptions) {
   const clientOptions: LanguageClientOptions = {
     documentSelector: [
-      { language: 'scala', scheme: 'file' },
-      { language: 'scala', scheme: 'untitled' }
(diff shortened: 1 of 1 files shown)


smarter:
@nicolasstucki This is probably the missing place where you need to add .tasty/.hasTasty.

nicolasstucki:
I will try it. Thanks.

nicolasstucki:
That did not help

PR #4567, opened 2018-05-22
Fix macros with array parameters

nicolasstucki:
We have to convert names in the form `scala.Int[]` to `[I`, `scala.Int[][]` to `[[I`, `java.lang.String[]` to `[Ljava.lang.String;` and so on.


nicolasstucki:
@biboudis you will need this fix to be able to write the `Stream.of[A](x: Array[A])` as a macro

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -1156,9 +1156,9 @@ class Definitions {
   /** The JVM tag for `tp` if it's a primitive, `java.lang.Object` otherwise. */
   def typeTag(tp: Type)(implicit ctx: Context): Name = typeTags(scalaClassName(tp))
 
-  /** The `Class[_]` of a primitive value type name */
-  def valueTypeNameToJavaType(name: TypeName)(implicit ctx: Context): Option[Class[_]] =
-    valueTypeNamesToJavaType.get(if (name.firstPart eq nme.scala_) name.lastPart.toTypeName else name)
+//  /** The `Class[_]` of a primitive value type name */
```
biboudis: I guess this should remain commented-out and not removed, right?

nicolasstucki: Yes, in case we need it for something else again.  It was really useful the first time.

PR #4568, opened 2018-05-23
Change protected accessors

odersky:
Generate protected accessors after pickling in a mini-phase.

Change rules so that accessors are also created for Scala-defined protected definitions, if
the access comes from outside the definitions visibility boundary and the class enclosing
the access is not a subclass. This prepares half of the way towards leaving such definitions
protected instead of widening their accessibility to public. The other half of the way would be
to forbid access from nested packages, but this will not be done in this PR.


odersky:
It turns out that the rules governing super accessors and protected accessors are more subtle and intertwined than it looks at first. The root cause of this is that traits extend classes in Scala but not in the JVM.

PR #4570, opened 2018-05-23
Sbt 1.1.5

allanrenucci:


allanrenucci:
The last commit is forward ported from Scalac

PR #4575, opened 2018-05-24
Enable `-Ycheck:all` in Dotty Doc tests

Duhemm:


compiler/src/dotty/tools/dotc/typer/Docstrings.scala
```diff
@@ -17,13 +17,14 @@ trait Docstrings { self: Typer =>
...
-      val relevantSyms = syms.filter(docbase.docstring(_).isDefined)
+      val relevantSyms = syms.filter(docbase.docstring(_).exists(!_.isExpanded))
       relevantSyms.foreach { sym =>
         expandParentDocs(sym)
         val usecases = docbase.docstring(sym).map(_.usecases).getOrElse(Nil)
 
         usecases.foreach { usecase =>
-          enterSymbol(createSymbol(usecase.untpdCode))
+          val createdSymbol = enterSymbol(createSymbol(usecase.untpdCode))
+          createdSymbol.setFlag(Flags.DocArtifact)
```
allanrenucci: Can you avoid creating a new flag and use the symbol name instead (i.e. is the name a `DocArtifactName`)?

Duhemm: I didn't know about `name.is`, thanks!

PR #4576, opened 2018-05-24
Do not allow running Expr that came as macro arguments

nicolasstucki:
It is too hard to ensure that the contents of the tree
can be pickled. It would be posible but the current ad-hoc
solution is too brittle and has no good usecases.


library/src/scala/quoted/Expr.scala
```diff
@@ -50,9 +50,8 @@ object Exprs {
    *
    *  May contain references to code defined outside this Expr instance.
    */
-  final class TreeExpr[Tree](val tree: Tree, pickle: => Expr[_]) extends quoted.Expr[Any] {
```
biboudis: Perhaps the comment need to be sync up that this refers to a Tasty Expr tree.

compiler/src/dotty/tools/dotc/quoted/Toolbox.scala
```diff
@@ -24,8 +24,8 @@ object Toolbox {
     def run(expr: Expr[T]): T = expr match {
       case expr: LiftedExpr[T] =>
         expr.value
-      case expr: TreeExpr[Tree] @unchecked =>
-        new QuoteDriver().run(expr.pickled, runSettings)
+      case expr: TastyTreeExpr[Tree] @unchecked =>
+        throw new scala.quoted.QuoteError("Can't run and scala.quoted.Expr coming from a macro argument")
```
biboudis: The comment as is has a typo, so a fix can be "Cannot run a type scala.quoted.Expr" ... 

However that error is thrown at the call site of an inline def. The `f.run` is invoked inside the macro body, so maybe we can rephrase into something the following maybe:

> Cannot call an inline definition that runs an Expr that comes from a macro argument.

PR #4577, opened 2018-05-24
Add tasty Show type class and us in Tasty.XYZ.show

nicolasstucki:
For XYZ: Tree, TypeOrBoundTree, TypeOrBound and Constant


library/src/scala/tasty/util/ShowExtractors.scala
```diff
@@ -2,22 +2,19 @@ package scala.tasty.util
 
 import scala.tasty.Tasty
 
-class TastyPrinter[T <: Tasty with Singleton](val tasty: T) {
+class ShowExtractors[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty0) {
```
liufengyun: I hope the name `ShowExtractors` can be improved later if possible.

nicolasstucki: Yes

compiler/src/dotty/tools/dotc/tasty/TastyImpl.scala
```diff
@@ -24,6 +25,15 @@ object TastyImpl extends scala.tasty.Tasty {
     def toTasty(implicit ctx: Context): TypeTree = PickledQuotes.quotedTypeToTree(x)
   }
 
+  // ===== Show =====================================================
+
+  def defaultShow: Show[this.type] = showExtractors
```
Blaisorblade: Using `def` here makes the member unstable problems, see c9dbecce233978048cedad6a0154e1df85d1f83a — I'm not using the API you used, but I'd expect both to work as they usually do.

PR #4578, opened 2018-05-24
Do not compute PackageDef members in extractor

nicolasstucki:


compiler/src/dotty/tools/dotc/tasty/TastyImpl.scala
```diff
@@ -194,17 +194,19 @@ object TastyImpl extends scala.tasty.Tasty {
...
-    def unapply(x: PackageDef)(implicit ctx: Context): Option[(String, List[Statement])] = x match {
+    def unapply(x: PackageDef)(implicit ctx: Context): Option[(String, PackageDef)] = x match {
       case x: PackageDefinition =>
-        // FIXME Do not do this eagerly as it forces everithing in the package to be loaded.
-        //       An alternative would be to add it as an extension method instead.
-        val definitions =
-          if (x.symbol.is(core.Flags.JavaDefined)) Nil // FIXME should also support java packages
-          else x.symbol.info.decls.iterator.map(FromSymbol.definition).toList
-        Some(x.symbol.name.toString, definitions)
+        Some((x.symbol.name.toString, FromSymbol.packageDef(x.symbol.owner)))
```
Blaisorblade: Pretty hard to guess here: the returned PackageDef is the owner. Lacking constructors that's hard to document.

nicolasstucki:
Rebased

nicolasstucki:
The confusion here is that in tasty reflet we call the PackageDef tree PackageClause. The PackageDef in tasty reflect represents the symbol of the package.

nicolasstucki:
@Blaisorblade any other comment or can we merge this one?

Blaisorblade:
Merging.

> The PackageDef in tasty reflect represents the symbol of the package.

Now I see what you mean, but I don't understand what the distinction is supposed to represent conceptually.

As far as I can tell, `tasty.PackageDef` has no corresponding node in tasty, and it's just used as a way to proxy package clauses without loading the whole package.

PR #4579, opened 2018-05-24
Implement decompiler on Tasty

nicolasstucki:
The implementation of the decompiler must:
* print compilable code
* print precise types 
* print resolved implicits
* resugar code that cannot be written in source

It does NOT need:
* pretty printing (line wrapping, small types, ...), this is something that can be done trough another tool or another custom printer


nicolasstucki:
Rebased

allanrenucci:
Should we remove `DecompilerPrinter`? We could simplify the implementation of `RefinedPrinter`

nicolasstucki:
The idea it to remove the `DecompilerPrinter` and merge the code with `ReplPrinter`

compiler/src/dotty/tools/dotc/decompiler/DecompilationPrinter.scala
```diff
@@ -35,16 +36,12 @@ class DecompilationPrinter extends Phase {
...
 
   private def printToOutput(out: PrintStream)(implicit ctx: Context): Unit = {
-    val unit = ctx.compilationUnit
-    val pageWidth = ctx.settings.pageWidth.value
-    val printLines = ctx.settings.printLines.value
-
     if (ctx.settings.printTasty.value) {
       new TastyPrinter(unit.pickled.head._2).printContents()
     } else {
+      val unit = ctx.compilationUnit
```
liufengyun: `unit` is no longer accessible in L40, the CI failed because of this.

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -0,0 +1,751 @@
...
+
+      case Type.MethodType(paramNames, paramTypes, resType) =>
+        ???
+
+      case Type.PolyType(paramNames, paramBounds, resType) =>
+        ???
+
+      case Type.TypeLambda(paramNames, paramBounds, resType) =>
+        ???
+    }
```
liufengyun: It seems `???` in the above means they should never be reached. Maybe use a "|" pattern to combine them in one case (the variable patterns are not used, can be underscore), and add a note.

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -0,0 +1,751 @@
...
+        printTree(elsep)
+
+      case Term.Match(selector, cases) =>
+        printTree(selector)
+        this += " match {"
+        indented {
+          this += lineBreak()
+          printCases(cases, lineBreak())
+        }
+        this += lineBreak() += "}"
```
liufengyun: Do we have a test for partial functions?

Another tricky issue is extractors with implicits. Maybe add a TODO to support later.

nicolasstucki: Not yet, I will add it later. The idea is to start testing all the files in `test/pos` and `test/run`.

library/src/scala/tasty/util/ShowSourceCode.scala
nicolasstucki: I will add a better way to handle unreachable patterns.

liufengyun:
LGTM

PR #4584, opened 2018-05-26
Fix decompilation of while loops

nicolasstucki:
Show that there is no semantic difference between inline functions and by name functions.

### DIFF
--- a/library/src/scala/tasty/util/ShowSourceCode.scala
+++ b/library/src/scala/tasty/util/ShowSourceCode.scala
@@ -145,6 +145,39 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
             this
         }
 
+      case While(cond, stats) =>
+        this += "while ("
+        printTree(cond)
+        this += ") "
+        stats match {
+          case stat :: Nil =>
+            printTree(stat)
+          case stats =>
+            this += "{"
+            indented {
+              this += lineBreak()
+              printTrees(stats, lineBreak())
+            }
+            this += lineBreak() += "}"
+        }
+
+      case DoWhile(stats, cond) =>
+        this += "do "
+        stats match {
+          case stat :: Nil =>
+            printTree(stat)
+          case stats =>
+            this += "{"
+            indented {
+              this += lineBreak()
+              printTrees(stats, lineBreak())
+            }
+            this += lineBreak() += "}"
+        }
+        this += " while ("
+        printTree(cond)
+        this += ")"
+
       case ddef@DefDef(name, targs, argss, tpt, rhs) =>
         val flags = ddef.flags
         if (flags.isOverride) sb.append("override ")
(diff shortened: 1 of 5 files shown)


odersky:
We are bout to disallow inline parameters of function type anyway. So should we close this commit?

nicolasstucki:
I will remove the inline function part and keep the by name functions as regression tests

nicolasstucki:
Found a regression

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -225,7 +245,14 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
         this += " = "
         printTree(rhs)
 
-      case Term.Block(stats, expr) =>
+      case Term.Block(stats0, expr) =>
+        def isLoopContinue(tree: Tree): Boolean = tree match {
```
allanrenucci: Rename is `isLoopEntryPoint`?

nicolasstucki:
@allanrenucci this needs a review as the code completely changed

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -235,31 +275,22 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
-
-          case Term.Apply(Term.Ident("doWhile$"), _) =>
-            val DefDef("doWhile$", _, _, _, Some(Term.Block(List(body), Term.If(cond, _, _)))) = stats.head
-            this += "do "
-            printTree(body)
-            this += " while ("
-            printTree(cond)
-            this += ")"
-
+          case expr if isLoopEntryPoint(expr) && stats.size == 1 =>
```
allanrenucci: I think it would be cleaner to have this case be part of the extractor:
```scala
def unapply(arg: Tree)(implicit ctx: Context): Option[(Term, List[Statement])] = arg match {
  case DefDef("while$", _, _, _, Some(Term.If(cond, Term.Block(bodyStats, _), _))) => Some((cond, bodyStats))
  case Block(List(tree), _) => unapply(tree)
  case _ => None
}
```

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -753,6 +748,7 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
   private object While {
     def unapply(arg: Tree)(implicit ctx: Context): Option[(Term, List[Statement])] = arg match {
       case DefDef("while$", _, _, _, Some(Term.If(cond, Term.Block(bodyStats, _), _))) => Some((cond, bodyStats))
+      case Term.Block(List(tree), _) => unapply(tree)
```
allanrenucci: I suppose you will need to add the same case for the `DoWhile` extractor

PR #4589, opened 2018-05-26
Towards Inlining Untyped Code

odersky:
We would like inline code that is only typed at the leaves. This seems to be the best way to have an inlining scheme that can generate new types. This ability to generate new types gives us some of the power of whitebox macros. It is needed for most typelevel programming idioms.

This PR is preparatory. It contains

 - Regularization of modifier structure: Mods no longer add any new information, are kept
   only as carriers of positions for syntactic analysis in IDE and elsewhere.
 - Improvements in debug and printing options.
 - Constant folding unary expressions which was forgotten in typer so far.
 - General infrastructure to have untyped splices in typed trees.
 - Code for pickling and unpickling such untyped splices.
 - Drop `inline` as a modifier for function parameters, use by-name instead.
 - Fixes to inliner to make the previous change work, and in general to do a better job
   reducing the amount of code that's generated.

The PR can be reviewed separately. It will be followed up with another PR, introducing `transparent` methods, which allow to inline untyped trees.


odersky:
test performance please

odersky:
I admit it would be better to split this into separate PRs. But my time-budget is already overdrawn as is, so have chosen to push ahead and try to get the subsequent steps done instead.

odersky:
Replaced by #4620 for now. Putting decision to pickle untyped trees on hold.

PR #4593, opened 2018-05-28
Fix signature caching in NamedType

smarter:


smarter:
test performance please

Issue #4590, opened 2018-05-28
Compiler crash because of inline accessors


allanrenucci:
Regression in community build introduced in #4468 

```scala
class ArrayDeque {
  inline def isResizeNecessary(len: Int) = len > ArrayDeque.StableSize
}

object ArrayDeque {
  private val StableSize = 256
}
```
```scala
class List {
  inline def foo(x: List.Cons): Unit = {
    x.next = this
  }
}
object List {
  case class Cons(head: Int, private[List] var next: List)
}
```
```
Exception in thread "main" java.lang.AssertionError: assertion failed: symbol (method inline_set$next) entered the scope of non-class owner val <none>
	at scala.Predef$.assert(Predef.scala:219)
	at dotty.tools.dotc.core.Symbols$Symbol.entered(Symbols.scala:533)
	at dotty.tools.dotc.transform.AccessProxies$Insert.newAccessorSymbol(AccessProxies.scala:61)
	at dotty.tools.dotc.transform.AccessProxies$Insert.newAccessorSymbol$(AccessProxies.scala:59)
	at dotty.tools.dotc.typer.Inliner$InlineAccessors$MakeInlineable.newAccessorSymbol(Inliner.scala:41)
	at dotty.tools.dotc.transform.AccessProxies$Insert.$anonfun$useAccessor$2(AccessProxies.scala:83)
	at dotty.tools.dotc.core.Symbols$Symbol.orElse(Symbols.scala:559)
	at dotty.tools.dotc.transform.AccessProxies$Insert.useAccessor(AccessProxies.scala:82)
	at dotty.tools.dotc.transform.AccessProxies$Insert.useAccessor$(AccessProxies.scala:69)
	at dotty.tools.dotc.typer.Inliner$InlineAccessors$MakeInlineable.useAccessor(Inliner.scala:41)
	at dotty.tools.dotc.transform.AccessProxies$Insert.accessorIfNeeded(AccessProxies.scala:104)
	at dotty.tools.dotc.transform.AccessProxies$Insert.accessorIfNeeded$(AccessProxies.scala:96)
	at dotty.tools.dotc.typer.Inliner$InlineAccessors$MakeInlineable.accessorIfNeeded(Inliner.scala:41)
	at dotty.tools.dotc.typer.Inliner$InlineAccessors$MakeInlineable.transform(Inliner.scala:58)
	at dotty.tools.dotc.ast.Trees$Instance$TreeMap.transform(Trees.scala:1154)
[...]

PR #4596, opened 2018-05-29
Re-enable collection-strawman tests

allanrenucci:
~Blocked on #4590~

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,5 +8,5 @@
 	branch = dotty-library2.12
 [submodule "collection-strawman"]
 	path = collection-strawman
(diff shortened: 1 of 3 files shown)


Blaisorblade:
The failing test is for a dormant feature, but it covers a compilation path that we care for and isn't tested otherwise on collection-strawman, so we should add a test.

In particular, failing test `LinkTests` relies on `-Xlink`, which we don't really support (it's part of dotty-linker). But the failure could be due to loading `collections-strawman` from Tasty: to confirm this, we should extend `FromTastyTests` to apply to `collections-strawman`.

allanrenucci:
Closing in favor of #4724

Issue #3509, opened 2017-11-20
Cannot pass jvm parameters to dotr


nicolasstucki:
When calling `dotr` (without launching the repl) we cannot pass JVM arguments such as `-Xmx512`. We should probably support the `-JXmx512` notation the `dotc` supports. All flags starting with `-J` should be passed to the JVM without the `J`.

jiminhsieh:
I can give it a shot. :)
If I understand correctly, it would be like how `dotc` handle `-JXmx512`. Thanks!

nicolasstucki:
Similarly. When we run `dotr -JXmx512 Foo` we should end up executing `java -Xms512 -classpath <dotty runtime ...> Foo` .

Or maybe it is enough to run  `dotr -Xmx512 Foo`? That might be the case.

nicolasstucki:
Maybe wait a day before starting, I will write down the exact way to handle arguments in dotr.

nicolasstucki:
```
dotr [-repl|-run] <jvmopt>* [<Main> <main-args>*]

<jvmopt> = -classpath <some-classpath>
 |  -J<jvm-option>
                 |  -J<jvm-option>
```

Run the REPL if `-repl` is set or if `-run` is not set and `[<Main> <main-args>*]` not is present.
Run the app if `-run` is set or `[<Main> <main-args>*]` is present.

nicolasstucki:
@jiminhsieh if you are still interested, now you can have a stab at it.

jiminhsieh:
@nicolasstucki If I don't misunderstand, this ticket should change scripts of `dotr` or `dotc`. If I'm correct, I have some questions about scripts of `dotr` and `dotc`. Should I discuss in gitter or in this ticket? Thanks! :)

nicolasstucki:
This should only change the `dotr` script. The reason for this special handling of is that when we run a dotty program we do not enter through one of our `Main`s which handle the args. We instead boot the JVM directly, hence we have to adapt the args that are passed to `java ...`. I belive it would be better to keep discussions on this issue in this GitHub issue to allow others to give their opinion.

jiminhsieh:
Basically, we introduce a new command line option `-run` for `dotr`. It seems `-run` is used to execute the app, `Main` directly. 

```bash
dotr [-repl|-run] <jvmopt>* [<Main> <main-args>*]

<jvmopt> = -classpath <some-classpath> | -J<jvm-option>
```
[...]

PR #4598, opened 2018-05-29
Fix #4066 `./bin/dotr -language:Scala2` fails

### LINKED ISSUE #4066: `./bin/dotr -language:Scala2` fails
It gives
```
warning: multiple classpaths are found, dotr only use the last one.
Unrecognized option: -language:Scala2
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
while `./bin/dotc -language:Scala2` is fine and `./bin/dotr -repl -language:Scala2` works as well.
Comparing `dotc` and `dotr` leads to the differences between `dist-bootstrapped/target/pack/bin/dotr` and `dist-bootstrapped/target/pack/bin/dotc`.

newca12:
- allow to pass compilation option to ./bin/dotr 
- restore standard notation for java options (using -J-Xmx512m instead of -JXmx512m introduce with https://github.com/lampepfl/dotty/issues/3509)
- allow to pass java options to ./bin/dotr -repl
- allow to pass java options to ./bin/dotc

I was a bit surprise to find dotr and dotc in such bad shape.
The code was and still is a bit fishy but as far as I can see now work in every situation.

### DIFF
--- a/dist/bin/dotc
+++ b/dist/bin/dotc
@@ -92,8 +92,8 @@ case "$1" in
    -no-colors) unset colors && shift ;;
   -with-compiler) jvm_cp_args="$PSEP$DOTTY_COMP" && shift ;;
 
-          # break out -D and -J options and add them to JAVA_OPTS as well
-          # so they reach the JVM in time to do some good. The -D options
+          # break out -D and -J options and add them to java_args so
+          # they reach the JVM in time to do some good. The -D options
           # will be available as system properties.
           -D*) addJava "$1" && shift ;;
           -J*) addJava "${1:2}" && shift ;;
--- a/dist/bin/dotc
+++ b/dist/bin/dotc
@@ -95,8 +95,8 @@ case "$1" in
           # break out -D and -J options and add them to JAVA_OPTS as well
           # so they reach the JVM in time to do some good. The -D options
           # will be available as system properties.
-          -D*) addJava "$1" && addScala "$1" && shift ;;
-          -J*) addJava "${1:2}" && addScala "$1" && shift ;;
+          -D*) addJava "$1" && shift ;;
+          -J*) addJava "${1:2}" && shift ;;
             *) addResidual "$1" && shift ;;
   esac
 done
--- a/dist/bin/dotr
+++ b/dist/bin/dotr
@@ -31,6 +31,10 @@ addJvmOptions () {
   jvm_options+=("$1")
 }
 
+addRawJvmOptions () {
+  java_options+=("$1")
+}
+
 source "$PROG_HOME/bin/common"
 
 declare -a residual_args
@@ -39,6 +43,11 @@ execute_run=false
 with_compiler=false
 class_path_count=0
 CLASS_PATH=""
+
+# Little hack to check if all arguments are options
+all_params="$*"
+truncated_params="${*#-}"
+not_only_options=$(( ${#all_params} - ${#truncated_params} - $# ))
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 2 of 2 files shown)


newca12:
I did sign the CLA.

dist/bin/dotc
```diff
@@ -95,8 +95,8 @@ case "$1" in
           # break out -D and -J options and add them to JAVA_OPTS as well
           # so they reach the JVM in time to do some good. The -D options
           # will be available as system properties.
```
Blaisorblade: This comment matches the old behavior ("add them to JAVA_OPTS as well [as to the Scala options]"), could you please update accordingly? I double-checked the code, but I agree with your change.

Blaisorblade:
[CLA signed](https://www.lightbend.com/contribute/cla/scala/check/newca12).

dist/bin/dotr
```diff
@@ -40,6 +44,11 @@ with_compiler=false
 class_path_count=0
 CLASS_PATH=""
 
+# Little hack to check if all arguments are options
+all_params="$*"
+truncated_params="${*#-}"
+not_only_options=$(( ${#all_params} - ${#truncated_params} - $# ))
```
liufengyun: Maybe rename `not_only_options` to `num_of_options`?

dist/bin/dotr
```diff
@@ -65,7 +74,8 @@ while [[ $# -gt 0 ]]; do
       shift
       ;;
     -J*)
-      addJvmOptions "-${1:2}"
+      addJvmOptions "${1:2}"
+      addRawJvmOptions "${1}"
```
liufengyun: Maybe rename to something like `addJvmOptionsRun/addJvmOptionsRepl` or `addJvmOptions/addDotcOptions`, which is more clear?

dist/bin/dotc
liufengyun: This code dates back to scalac script, maybe we should keep it:

```bash
  case "$1" in
    -D*)
      # pass to scala as well: otherwise we lose it sometimes when we
      # need it, e.g. communicating with a server compiler.
      java_args+=("$1")
      scala_args+=("$1")
      # respect user-supplied -Dscala.usejavacp
      case "$1" in -Dscala.usejavacp*) OVERRIDE_USEJAVACP="";; esac
      shift
      ;;
    -J*)
      # as with -D, pass to scala even though it will almost
      # never be used.
      java_args+=("${1:2}")
      scala_args+=("$1")
      shift
      ;;
```

newca12:
@Blaisorblade, the real signification of "as well" come from a 8 years old paulp commit https://github.com/scala/scala/commit/80b0d241349f6463e814a8c1eca34f827ce24518#diff-eb4553171a4a79a0be4ea69f883e75b2R72
@liufengyun I cannot find any evidence that scala_args is used in Scala. Sound like dead code for me.
This PR does not fix problems with Windows.

Blaisorblade:
> @liufengyun I cannot find any evidence that scala_args is used in Scala. Sound like dead code for me.

Right, that comment hints at `fsc`, which we don't have: the options for the fsc client had to be parsed and passed by hand to the existing fsc server.

liufengyun:
> > @liufengyun I cannot find any evidence that scala_args is used in Scala. Sound like dead code for me.

> Right, that comment hints at `fsc`, which we don't have: the options for the fsc client had to be parsed and passed by hand to the existing fsc server.

OK, I agree with you, thanks for the clarification.

Blaisorblade:
Assigning to @liufengyun for merge if he has no further comments.

liufengyun:
Thanks a lot for the contribution @newca12 🎉

PR #4601, opened 2018-05-30
Fix synthetic type bounds in Tasty reflect

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -538,7 +538,7 @@ object TastyImpl extends scala.tasty.Tasty {
 
     object Synthetic extends SyntheticExtractor {
       def unapply(x: TypeTree)(implicit ctx: Context): Boolean = x match {
-        case Trees.TypeTree() => true
+        case x @ Trees.TypeTree() => !x.tpe.isInstanceOf[Types.TypeBounds]
         case _ => false
       }
     }
```
liufengyun: Given the name `synthetic`, I cannot justify the change. Maybe a doc for `Synthetic`, I naively assume synthetic bounds are also synthetic.

library/src/scala/tasty/util/ShowExtractors.scala
```diff
@@ -188,7 +190,8 @@ class ShowExtractors[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
       case Type.PolyType(argNames, argBounds, resType) =>
         this += "Type.PolyType(" ++= argNames += ", " ++= argBounds += ", " += resType += ")"
       case Type.TypeLambda(argNames, argBounds, resType) =>
-        this += "Type.TypeLambda(" ++= argNames += ", " ++= argBounds += ", " += resType += ")"
+        // resType is not printed to avoid cycles
```
liufengyun: I'm curious what kind of cycles can happen here?

nicolasstucki: For definitions, we use the tree of the definition. Somewhere in the body of the type lambda there was a reference to itself and then tried to reprint itself in a loop.

compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
nicolasstucki: I will add some documentation.

nicolasstucki: Added

Issue #2573, opened 2017-05-29
Scala Wart: Convoluted de-sugaring of for-comprehensions


lihaoyi:
Opening this issue, as suggested by Martin, to provide a place to discuss the individual warts brought up in the blog post [Warts of the Scala Programming Language](http://www.lihaoyi.com/post/WartsoftheScalaProgrammingLanguage.html) and the possibility of mitigating/fixing them in Dotty (and perhaps later in Scala 2.x). These are based on Scala 2.x behavior, which I understand Dotty follows closely, apologies in advance if it has already been fixed

--------------------------------------------------------------------------------

Scala lets you write for-comprehensions, which are converted into a chain
of `flatMap`s an `map`s as shown below:

```scala
@ val (x, y, z) = (Some(1), Some(2), Some(3))
x: Some[Int] = Some(1)
y: Some[Int] = Some(2)
z: Some[Int] = Some(3)
@ for{
    i <- x
    j <- y
    k <- z
  } yield i + j + k
res40: Option[Int] = Some(6)
@ desugar{
    for{
      i <- x
      j <- y
      k <- z
    } yield i + j + k
  }
res41: Desugared = x.flatMap{ i => 
  y.flatMap{ j => 
    z.map{ k => 
      i + j + k
    }
  }
}
```

I have nicely formatted the desugared code for you, but you can try this 
yourself in the [Ammonite Scala REPL](http://www.lihaoyi.com/Ammonite/) to 
verify that this is what the for-comprehension gets transformed into. 

This is a convenient way of implementing nested loops over lists, and happily
works with things that aren't lists: `Option`s (as shown above), `Future`s, 
and many other things.

You can also assign local values within the for-comprehension, e.g.

```scala
@ for{
    i <- x
    j <- y
    foo = 5
    k <- z
  } yield i + j + k + foo
res42: Option[Int] = Some(11)
```

The syntax is a bit wonky (you don't need a `val`, you can't define `def`s or
`class`es or run imperative commands without `_ = println("debug")`) but for 
simple local assignments it works. You may expect the above code to be 
transformed into something like this

```scala
res43: Desugared = x.flatMap{ i => 
  y.flatMap{ j =>
    val foo = 5
    z.map{ k => 
      i + j + k
    }
  }
}
```

But it turns out it instead gets converted into something like this:

```scala
@ desugar{
    for{
      i <- x
      j <- y
      foo = 5
      k <- z
    } yield i + j + k + foo
  }
[...]

PR #4602, opened 2018-05-30
Stop rewriting withFilter to filter

Blaisorblade:
This was dropped for Scala 2 in https://github.com/scala/scala/pull/5252 for 2.12.0-RC1, as discussed for instance in https://github.com/scala/bug/issues/6455 and https://github.com/scala/bug/issues/8339.

Beyond this code, also remove `MaybeFilter` attachment and attending code
introduced in fcea3d5.

A caveat is that, unexpectedly, this code was added in 2016-08-24, after
scala/scala#5252 was merged, as if the decision was reversed; but #1461 says:

> Also, I went through tests in pending/pos and reclassified as far as possible.

So I assume this just happened because the Scala 2 PR (of course) didn't remove our pending test.


odersky:
Also related: #2573 

It would be great of we could drop `withFilter` and `filter` altogether and use a zero element instead. This would solve several problems where irrefutable patterns are recognized too late,  would interact better with `=` bindings in `for` and would generally be more efficient. Also, it would
get rid of the complexity of correctly modelling the type of `withFilter` in the collections.

PR #4603, opened 2018-05-31
More changes to protected accessors

odersky:
Some fixes in the AccessProxies scheme that were discovered when revising the inliner.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
+++ b/compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
@@ -58,12 +58,23 @@ abstract class AccessProxies {
     def needsAccessor(sym: Symbol)(implicit ctx: Context): Boolean
 
     /** A fresh accessor symbol */
-    def newAccessorSymbol(owner: Symbol, name: TermName, info: Type, pos: Position)(implicit ctx: Context): TermSymbol = {
+    private def newAccessorSymbol(owner: Symbol, name: TermName, info: Type, pos: Position)(implicit ctx: Context): TermSymbol = {
       val sym = ctx.newSymbol(owner, name, Synthetic | Method, info, coord = pos).entered
       if (sym.allOverriddenSymbols.exists(!_.is(Deferred))) sym.setFlag(Override)
       sym
     }
 
+    /** An accessor symbol, create a fresh one unless one exists already */
+    private def accessorSymbol(owner: Symbol, accessorName: TermName, accessorInfo: Type, accessed: Symbol)(implicit ctx: Context) = {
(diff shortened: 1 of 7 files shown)


odersky:
Hi Allan, I did some improvements to the protected accessor scheme. Can you use this one as a basis for the partial functions PR?

odersky:
Rebased on top of master which now contains #4597

compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
```diff
@@ -49,34 +47,60 @@ abstract class AccessProxies {
...
+        accessedBy(acc) = accessed
+        acc
+      }
+    }
+
+    /** Rewire reference to refer to `accessor` symbol */
+    private def rewire(reference: RefTree, accessor: Symbol)(implicit ctx: Context): Tree = {
+      reference match {
+        case Select(qual, _) => qual.select(accessor)
+        case Ident(name) => ref(accessor)
```
allanrenucci: Do you need to desugar the Identifier here like in `tpd.becomes`?

compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
```diff
@@ -49,34 +47,60 @@ abstract class AccessProxies {
...
-    def newAccessorSymbol(owner: Symbol, name: TermName, info: Type, pos: Position)(implicit ctx: Context): TermSymbol = {
+    private def newAccessorSymbol(owner: Symbol, name: TermName, info: Type, pos: Position)(implicit ctx: Context): TermSymbol = {
       val sym = ctx.newSymbol(owner, name, Synthetic | Method, info, coord = pos).entered
       if (sym.allOverriddenSymbols.exists(!_.is(Deferred))) sym.setFlag(Override)
       sym
     }
 
+    /** An accessor symbol, create a fresh one unless one exists already */
+    private def accessorSymbol(owner: Symbol, accessorName: TermName, accessorInfo: Type, accessed: Symbol)(implicit ctx: Context) = {
+      def refersToAccessed(sym: Symbol) = accessedBy.get(sym) == Some(accessed)
```
allanrenucci: `accessedBy.get(sym).contains(accessed)`?

compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
odersky: No, since we always to refer to given `accessor`.

compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
odersky: Ah, yes. I'll roll that into another PR.

PR #4604, opened 2018-05-31
Fix #4446: Inline implementation of PF methods into its anonymous class

### LINKED ISSUE #4446: Don't use forwarders in closure implementation
```scala
class Foo {
  (x: Int) => x + 1
}
```
A closure that is not a SAM will desugar to something like this in Dotty:
```scala
class Foo {
  def apply(x: Int) = x + 1

  class $anon extends Function1[Int, Int] {
    def apply(x: Int) = Foo.this.apply(x)
  }
  new $anon
}
```
and this in scalac:
```scala
class Foo {
  class $anon extends Function1[Int, Int] {
    def apply(x: Int) =  x + 1
  }
  new $anon
}
```
The issue with the Dotty desugaring, is that `$anon` will capture a reference to the outer class `Foo` that is not needed

allanrenucci:


allanrenucci:
Blocked on  #1895.

PR #4605, opened 2018-05-31
Fix printing extractors

nicolasstucki:
Extractors in pattern matches were printing `case <empty>.Bar.unapply(a) =>` instead of `case Bar(a) =>`.

Also added `.show` of `Tasty.Pattern` to make it easy to debug.


nicolasstucki:
Rebased

PR #4607, opened 2018-05-31
Add a way filter FromTasty tests

nicolasstucki:
In sbt use: `testFromTasty XYZ`

It does the same as `testCompilation XYZ` but on the tasty tests

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -65,6 +65,9 @@ object Build {
   // Run tests with filter through vulpix test suite
   lazy val testCompilation = inputKey[Unit]("runs integration test with the supplied filter")
 
+  // Run TASTY tests with filter through vulpix test suite
+  lazy val testFromTasty = inputKey[Unit]("runs tasty integration test with the supplied filter")
+
   // Spawns a repl with the correct classpath
   lazy val repl = inputKey[Unit]("run the REPL with correct classpath")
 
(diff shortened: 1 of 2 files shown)


project/Build.scala
```diff
@@ -583,6 +586,15 @@ object Build {
...
       }.evaluated,
 
+      testFromTasty := Def.inputTaskDyn {
+        val args: Seq[String] = spaceDelimited("<arg>").parsed
+        val cmd = " dotty.tools.dotc.FromTastyTests -- " + {
+          if (args.nonEmpty) " -Ddotty.tests.filter=" + args.mkString(" ")
+          else ""
+        }
+        (testOnly in Test).toTask(cmd)
+      }.evaluated,
```
allanrenucci: Let's abstract over the testOnly input task:
```scala
def testOnly(test: String, options: String) = Def.inputTaskDyn {
  val args = spaceDelimited("<arg>").parsed
  val cmd = s" $test -- $options" + {
    if (args.nonEmpty) " -Ddotty.tests.filter=" + args.mkString(" ")
    else ""
  }
  (testOnly in Test).toTask(cmd)
}
```

project/Build.scala
```diff
@@ -1237,4 +1234,13 @@ object Build {
       case BootstrappedOptimised => commonOptimisedSettings
     })
   }
+
+  def testOnlyFiltered(test: String, options: String) = Def.inputTaskDyn {
```
allanrenucci: Can you move it just above where is is used?

PR #4610, opened 2018-06-01
Add an implementation example of fully abstract interface

nicolasstucki:
* All type are defined as abstract type with subtyping if needed
* Abstract constructors and extractors are added to allow pattern matching
* Extension methods can be added to the abstract types to simulate methods
* Implicit class tags are added to allow correct type tests. If the classes
  of the implementation are not enough to distinguish the types we can implement 
  custom class tags the additionally check the state of the array. This is showed in
  the `ListImplementation` where everything is represented as a `List[Any]`

### DIFF
--- a/tests/run/fully-abstract-interface.check
+++ b/tests/run/fully-abstract-interface.check
@@ -1,0 +1,29 @@
+CaseClassImplementation
+underlying rep: class CaseClassImplementation$Const
+1
+test1 OK
+1 = 1
+test2 OK
+1 = 1
+
+underlying rep: class CaseClassImplementation$App
+7
+test3 OK
+AppliedOp(PlusOp, Const(1), App(MultOp,Const(2),Const(3))) = 7
+test4 OK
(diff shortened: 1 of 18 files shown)


Blaisorblade:
Ping for the test failures. Looks cool and scary at first sight, ping me again for a proper review once tests pass 😉

tests/run/fully-abstract-interface.check
```diff
@@ -0,0 +1,29 @@
...
+1
+test1 OK
+1 = 1
+test2 OK
+1 = 1
+
+underlying rep: class [Ljava.lang.Object;
+7
+test3 OK
+AppliedOp([Ljava.lang.Object;@5a39699c class [Ljava.lang.Object; class [Ljava.lang.Object;) = 7
```
Blaisorblade: Ah, the test fail because this address won't match (see http://dotty-ci.epfl.ch/lampepfl/dotty/5372/4 line 10323), so don't fully `toString` the thing I guess...

nicolasstucki: I replaced `Array` with `List` to have a clean `toString`

PR #4613, opened 2018-06-01
Fix #1895: Use proper validity period for extension methods

### LINKED ISSUE #1895: Bringing a symbol to a new run is broken

smarter:
When we create a symbol in a DenotationTransformer, we cannot safely
create it at phase `thisPhase.next` since this can lead to cycles. This
is why we created extension methods at phase `thisPhase`, but they're
not really valid at that point since they're only entered in their owner
once the transformer has been run. This lead to stale symbol errors
which were worked around in 51a458e but
the fix is incomplete and does not work with #4604 because that PR moves
the ExtensionMethods phase around.

This PR removes all the workarounds and uses a simpler solution:
create the symbols at phase `thisPhase` but set their validity
period to match the validity of the owner transformed denotation.


smarter:
test performance please

compiler/src/dotty/tools/dotc/core/DenotTransformers.scala
```diff
@@ -24,9 +24,9 @@ object DenotTransformers {
     /** The last phase during which the transformed denotations are valid */
     def lastPhaseId(implicit ctx: Context) = ctx.nextDenotTransformerId(id + 1)
 
-    /** The validity period of the transformer in the given context */
+    /** The validity period of the transformed denotations in the given context */
     def validFor(implicit ctx: Context): Period =
-      Period(ctx.runId, id, lastPhaseId)
+      Period(ctx.runId, id + 1, lastPhaseId)
```
odersky: At first I was scared, because it looks really hard to see all the ramifications of this change. But then I found out that the method is in fact not used at all! So do we still want to keep it?

PR #4614, opened 2018-06-02
Constant-fold unary operations in Typer

odersky:
This was previously forgotten, even though unary operations such as `!` and `~` were folded later
in FirstTransform. As a consequence, inlining simplifications involving constant expressions
using these operations were not done. This affected in particular the trace macro, which was always
expanded to an operation taking a closure argument.


odersky:
test performance please

nicolasstucki:
@odersky could you give some example of code that does not get optimized with `!`.

odersky:
@nicolasstucki No application of `trace` was optimized. That's how I found out.

PR #4617, opened 2018-06-04
Refactor "method does not take parameter" error message

allanrenucci:
A change I initially made as part of #4132. I decided to create a separate PR.

I changed the tests not to rely on pretty printing of singleton types and ended up refactoring the error message.

### DIFF
--- a/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
+++ b/compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
@@ -1371,27 +1371,25 @@ object messages {
            |"""
   }
 
-  case class MethodDoesNotTakeParameters(tree: tpd.Tree, methPartType: Types.Type)(err: typer.ErrorReporting.Errors)(implicit ctx: Context)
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1371,27 +1371,25 @@ object messages {
...
+    val explanation = {
+      val isNullary = methodSymbol.info.isInstanceOf[ExprType]
+      val addendum =
+        if (isNullary) "\nNullary methods may not be called with parenthesis"
+        else ""
 
-    val explanation =
-      s"""|You have specified more parameter lists as defined in the method definition(s).
-          |$noParameters""".stripMargin
+      "You have specified more parameter lists as defined in the method definition(s)." + addendum
```
Blaisorblade: @allanrenucci `s/as defined/than defined/`, no?

allanrenucci: Probably. I reuse the same wording as in the initial error message. Feel free to open a new PR 😄

PR #4618, opened 2018-06-04
WIP Provide abstract type tag

nicolasstucki:
Experiment on a general purpose type tag for abstract type tag that can be customized without extending ClassTag.


nicolasstucki:
Simple implementation, but conceptually broken :(

PR #4620, opened 2018-06-05
Various small improvements to prepare for transparent methods

odersky:
Supersedes #4589.

This PR contains some fixes accumulated while working on inlining untyped trees. It does not contain any code to inline, nor pickle untyped trees. We are currently considering an alternative design by @gsps and @OlivierBlanvillain which would not need to do this.

Main changes:

 - Regularization of modifier structure: Mods no longer add any new information, are kept
only as carriers of positions for syntactic analysis in IDE and elsewhere.
 - Improvements in debug and printing options.
 - Drop inline as a modifier for function parameters, use by-name instead.
 - Fixes to inliner to make the previous change work, and in general to do a better job
reducing the amount of code that's generated.


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -218,6 +218,31 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
...
+   *  to `bindingsBuf`.
+   *  @param name        the name of the parameter
+   *  @param paramtp     the type of the parameter
+   *  @param arg         the argument corresponding to the parameter
+   *  @param bindingsBuf the buffer to which the definition should be appended
+   */
+  private def paramBindingDef(name: Name, paramtp: Type, arg: Tree,
+                              bindingsBuf: mutable.ListBuffer[ValOrDefDef]): ValOrDefDef = {
+    val argtpe = arg.tpe.dealias
+    def isByName = paramtp.dealias.isInstanceOf[ExprType]
```
allanrenucci: Should be a `val`. It is always evaluated

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
```diff
@@ -612,18 +614,23 @@ class TreeUnpickler(reader: TastyReader,
...
+            annots = readAnnot(ctx) :: annots
           case tag =>
             assert(false, s"illegal modifier tag $tag at $currentAddr, end = $end")
         }
       }
-      (flags, annots.toList, privateWithin)
+      (flags, annots.reverse, privateWithin)
+    }
+
+    private val readAnnot: Context => Annotation = {
```
smarter: Why a val and not a def?

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -455,13 +466,99 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
...
-    override def typedApply(tree: untpd.Apply, pt: Type)(implicit ctx: Context) =
-      tree.asInstanceOf[tpd.Tree] match {
-        case Apply(Select(InlineableArg(closure(_, fn, _)), nme.apply), args) =>
-          inlining.println(i"reducing $tree with closure $fn")
-          typed(fn.appliedToArgs(args), pt)
-        case _ =>
-          super.typedApply(tree, pt)
+    override def typedApply(tree: untpd.Apply, pt: Type)(implicit ctx: Context) = {
+
+      def betaReduce(tree: Tree) = tree match {
```
smarter: Would be nice to document this method since it's non-trivial and the name is a bit vague.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -455,13 +466,99 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
...
+        case Apply(Select(cl @ closureDef(ddef), nme.apply), args) =>
```
smarter: If the closure is a SAM type then there is no guarantee that the implement method is called `apply`, in fact it's possible that `apply` is a completely different method that has nothing to do with `ddef` so this optimization is incorrect in general, e.g.:
```scala
trait Foo {
  def apply(x: Int): Int = 42
  def realApply(x: Int): Int
}
object Test {
    println(((x => x): Foo).apply(10))
  }
}
```

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -563,6 +563,9 @@ object Flags {
   /** An inline parameter */
   final val InlineParam = allOf(Inline, Param)
 
+  /** An enum case */
+  final val EnumCase = allOf(Enum, Case)
```
liufengyun: The documentation for `Enum` needs to be updated:

```Scala
   /** Symbol is a Java enum */
   final val Enum = commonFlag(40, "<enum>")
```

compiler/src/dotty/tools/dotc/core/tasty/TreeUnpickler.scala
odersky: We use it as a closure everytime we read a symbol. Making it a `val` avoids a lot of allocations.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
odersky: I thought beta-reduce is as specific as it gets? 😄

odersky: Just joking. I agree the details of this should be documented.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
odersky: Right. We need to restrict this to function closures.

compiler/src/dotty/tools/dotc/core/Flags.scala
odersky: That's done in the latest commit.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -455,13 +466,110 @@ class Inliner(call: tpd.Tree, rhs: tpd.Tree)(implicit ctx: Context) {
...
+       *
+       *  to
+       *
+       *    val/def x1 = e1; ...; val/def xn = en; b
+       *
+       *  where `def` is used for call-by-name parameters. However, we shortcut any NoPrefix
+       *  refs among the ei's directly without creating an intermediate binding.
+       */
+      def betaReduce(tree: Tree) = tree match {
+        case Apply(Select(cl @ closureDef(ddef), nme.apply), args) if defn.isFunctionType(cl.tpe) =>
```
smarter: > `defn.isFunctionType(cl.tpe)`

A more efficient way to check this is `cl.tpt.isEmpty`

smarter:
Too late for this PR now, but I think we really need more tests to make sure inline behave as we think it does (to not repeat issues such as https://github.com/lampepfl/dotty/pull/4614). We currently only have one test in https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/InlineBytecodeTests.scala but it's fairly easy to add more.

PR #4622, opened 2018-06-05
Add transparent methods

odersky:
0th step, building on #4620: Add transparent methods.


odersky:
@gsps I think this is the right basis to build on for step 1.

gsps:
Looking through this PR I noticed lots of unrelated commits and finally saw that there's another PR that includes the prefix up to "Addressing reviewer comments". I guess I'll only do my review from that point on.

odersky:
@gsps Yes, everything after #4620.

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -1113,7 +1113,7 @@ class Namer { typer: Typer =>
 
       // println(s"final inherited for $sym: ${inherited.toString}") !!!
       // println(s"owner = ${sym.owner}, decls = ${sym.owner.info.decls.show}")
-      def isInline = sym.is(FinalOrInline, butNot = Method | Mutable)
+      def isInline = sym.is(FinalOrInlineOrTransparent, butNot = Method | Mutable)
```
gsps: Not sure if it's worth adding `Transparent` to the exception here, since the handling of return types of transparent methods in `widenRhs` will have to be more different anyways (it should be allowed to infer a TermRef, for instance; think `transparent def unapply(x: Foo) = x`)

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -1128,7 +1128,8 @@ class Namer { typer: Typer =>
       // it would be erased to BoxedUnit.
       def dealiasIfUnit(tp: Type) = if (tp.isRef(defn.UnitClass)) defn.UnitType else tp
 
-      val rhsCtx = ctx.addMode(Mode.InferringReturnType)
+      var rhsCtx = ctx.addMode(Mode.InferringReturnType)
+      if (sym.isTransparentMethod) rhsCtx = rhsCtx.addMode(Mode.TransparentBody)
```
gsps: I'm not sure `TransparentBody` should be a Mode. If it is, shouldn't it also be set in the other tree traversing helpers besides typer, i.e., `TreeAccumulator` and `MegaPhase#transformTree`? All of those have `localContext` helpers (in case of typer its in `NamerContextOps`, the other two are local defs) which currently set the owner, among other things.

As an alternative, we could deduce whether we are in a transparent context by walking the context's owner chain and checking for `Transparent`. Ideally that query would be cached by the context (and its descendants).

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -104,14 +94,14 @@ object Inliner {
    *                     to have the inlined method as owner.
    */
   def registerInlineInfo(
-      sym: SymDenotation, treeExpr: Context => Tree)(implicit ctx: Context): Unit = {
-    sym.unforcedAnnotation(defn.BodyAnnot) match {
+      inlined: Symbol, treeExpr: Context => Tree)(implicit ctx: Context): Unit = {
```
gsps: Docs should be updated. Maybe call it `inlineSym`, since it's not inline**d** at this point?

tests/run/typelevel.scala
```diff
@@ -0,0 +1,31 @@
+object Test extends App {
```
gsps: Maybe give this file and its '.check' companion a more descriptive name like `transparent.scala` or `TransparentToNat.scala`?

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -53,23 +53,13 @@ object Inliner {
...
-                  super.transform(tree)
-              }
-            case _ =>
-              super.transform(tree)
-          }
-        case _ =>
-          super.transform(accessorIfNeeded(tree))
-      }
-
+      override def transform(tree: Tree)(implicit ctx: Context): Tree =
```
gsps: (I have no useful comment, but I will say that I don't understand Inliner well enough yet to really judge this change.)

compiler/src/dotty/tools/dotc/core/SymDenotations.scala
```diff
@@ -779,7 +779,13 @@ object SymDenotations {
 
     def isSkolem: Boolean = name == nme.SKOLEM
 
-    def isInlineMethod(implicit ctx: Context): Boolean = is(InlineMethod, butNot = Accessor)
+    def isInlinedMethod(implicit ctx: Context): Boolean =
```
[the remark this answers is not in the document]
odersky: I was thinking if we keep something like `inline` as a modifier we should rename it to `inlined` instead. Then we don't have a clash with the `@inline` annotation, and it's a bit cleaner. `transparent` and `inlined` are both adjectives, whereas it's a stretch to call `inline` an adjective.

compiler/src/dotty/tools/dotc/typer/Namer.scala
odersky: Yes, that was only done as a temporary measure because I could not decide yet whether Java final constants should be called `inline(d)` or `transparent`. Probably `transparent`, but for now it's `inline`.

compiler/src/dotty/tools/dotc/typer/Namer.scala
odersky: Good point. Since it's a single bit we cache, it could still go into the `Mode` structure (context fields are a precious resource). But it could be maintained fully automatically. I'll do that change.

tests/run/typelevel.scala
odersky: I planned to extend this test with more typelevel operations as we go along.

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -363,7 +363,10 @@ object Flags {
   /** Symbol is a Java default method */
   final val DefaultMethod = termFlag(38, "<defaultmethod>")
 
-  /** Symbol is a Java enum */
+  /** Labelled with `transparent` modifier */
+  final val Transparent = termFlag(39, "transparent")
```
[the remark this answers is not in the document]
odersky: Since there are no `transparent` types, it should not apply to them.

odersky:
@gsps Regarding the next step, I was wondering whether `TypeOf(tree, underlying)` can be modelled as as an annotated type `underlying @TypeOf(tree)` or `AnnotatedType(underyling, tpd.Apply(tpd.ref(defn.ToTypeAnnot), tree))` in internal representation? This would avoid having to add a new case to all typemaps etc. 

The downside is that currently annotations on types are forgotten in a large number of cases, which would not be correct in this case. On the other hand, there are a number of other possible annotations that conceptually should behave the same as `TypeOf`, i.e. they are generally seen as 
unknown subtypes of their underlying types. @olhotak worked on `mutable/immutable` classifiers that would fall into that category. Or units of measure. So maybe it's worthwhile to create a new trait of, say, `SubTypeAnnotation` which annotation classes can inherit. Types annotated with `SubTypeAnnotation` would keep their annotations in most contexts. We'd have to go with a fine comb through all the occurrences of `stripAnnotation` to make this work.

OlivierBlanvillain:
@odersky Regarding your last comment, I think it would be more work to change the handling annotated type rather than simply adding a new type. It should maybe be done as a separate generalization step? I feel than TypeOf is going to have a large impact on typer than other potential annotations like `mutable`/`immutable`...

odersky:
@OlivierBlanvillain I agree it would be easier to get started with a new form of type. But in the end, we will want to refactor to avoid the duplication.

odersky:
#4620 is in now. Rebased on top of current master.

odersky:
I believe if we can do it, it would be ultimately better to base this on `AnnotatedType`. The reason is that any work we do along that path benefits the whole. The potential downside is that it might detract us from the immediate purpose. I'll take a look how hard it would be to change annotations.

PR #4625, opened 2018-06-06
Add subtype annotations

odersky:
Based on #4622. We are trying to add a kind of annotation that can be used as a carrier for TypeOf types.


odersky:
Still to do: deal with dealias.

odersky:
Superseded by #4626

PR #4626, opened 2018-06-07
Add refining annotations

odersky:
Adds a new kind of annotation that conceptually refines the annotated type. That is, refining annotations yield subtypes. They are more "sticky" than normal annotations. In a number of situations they are not stripped away where normal annotations are.

It is intended that this kind of annotation is a good basis to define `TypeOf` types.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeComparer.scala
@@ -236,8 +236,8 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         compareWild
       case tp2: LazyRef =>
         !tp2.evaluating && recur(tp1, tp2.ref)
-      case tp2: AnnotatedType =>
-        recur(tp1, tp2.tpe) // todo: refine?
+      case tp2: AnnotatedType if !tp2.isRefining =>
+        recur(tp1, tp2.tpe)
       case tp2: ThisType =>
         def compareThis = {
           val cls2 = tp2.cls
@@ -345,7 +345,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
         // because that would cause an assertionError. Return false instead.
         // See i859.scala for an example where we hit this case.
         !tp1.evaluating && recur(tp1.ref, tp2)
-      case tp1: AnnotatedType =>
+      case tp1: AnnotatedType if !tp1.isRefining =>
         recur(tp1.tpe, tp2)
       case AndType(tp11, tp12) =>
         if (tp11.stripTypeVar eq tp12.stripTypeVar) recur(tp11, tp2)
@@ -567,6 +567,9 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
             false
         }
         compareTypeBounds
+      case tp2: AnnotatedType if tp2.isRefining =>
+        (tp1.derivesAnnotWith(tp2.annot.sameAnnotation) || defn.isBottomType(tp1)) &&
+        recur(tp1, tp2.tpe)
       case ClassInfo(pre2, cls2, _, _, _) =>
         def compareClassInfo = tp1 match {
           case ClassInfo(pre1, cls1, _, _, _) =>
@@ -661,6 +664,8 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
           case _ =>
         }
         either(recur(tp11, tp2), recur(tp12, tp2))
+      case tp1: AnnotatedType if tp1.isRefining =>
+        isNewSubType(tp1.tpe)
       case JavaArrayType(elem1) =>
         def compareJavaArray = tp2 match {
           case JavaArrayType(elem2) => isSubType(elem1, elem2)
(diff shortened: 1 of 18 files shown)


odersky:
@gsps I recommend to use this status as a basis for further work.

Next step could be to define an annotation like `TypeOf` in 

   library/src/scala/annotation/internal

library/src/scala/annotation/RefiningAnnotation.scala
```diff
@@ -0,0 +1,7 @@
+package scala.annotation
+
+/** A base trait for annotations that yield proper subtypes of the types they annotate.
+ *  Refining annotations are more "sticky" than normal ones. They are conceptually kept
+ *  around when normal refinements would also not be stripped away.
+ */
+trait RefiningAnnotation extends StaticAnnotation
```
OlivierBlanvillain: Why expose that in the stdlib, isn't this compiler internal only?

odersky: No, you might want to define your own annotations (say units of measure). Maybe they need a compiler plugin to  order to be checked, or maybe you can do it in the future with annotation macros. `RefiningAnnotation` just makes sure they propagate correctly.

odersky:
> Should ConstantType and TermRef be refactored into AnnotatedType with RefiningAnnotation?

In principle, we could consider this. But these two are a little bit different since they are also the main carriers of singleton types. Also the terms they capture are not very general, so it's not clear what we would win.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -700,7 +705,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
               }
             case tycon1: TypeVar =>
               isMatchingApply(tycon1.underlying)
-            case tycon1: AnnotatedType =>
+            case tycon1: AnnotatedType if !tycon1.isRefining =>
```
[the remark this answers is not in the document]
odersky: Yes, well spotted. Widening is always permitted.

compiler/src/dotty/tools/dotc/util/common.scala
```diff
@@ -8,6 +8,7 @@ import core.Types.WildcardType
 object common {
 
   val alwaysTrue = Function.const(true) _
+  val alwaysFalse = Function.const(false) _
```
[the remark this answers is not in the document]
odersky: No. I used it in an intermediate version. I kept it for duality.

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -283,6 +283,15 @@ object Types {
       case _ => false
     }
 
+    /** Does this type have a supertype with an annotation satisfying given predicate `p`? */
+    def derivesAnnotWith(p: Annotation => Boolean)(implicit ctx: Context): Boolean = this match {
+      case tp: AnnotatedType => p(tp.annot) || tp.tpe.derivesAnnotWith(p)
```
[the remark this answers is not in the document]
odersky: Good idea.

gsps:
LGTM.

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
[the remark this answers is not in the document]
smarter: Looks like this PR was merged without addressing this comment?

compiler/src/dotty/tools/dotc/core/TypeComparer.scala
```diff
@@ -1546,7 +1552,7 @@ class TypeComparer(initctx: Context) extends DotClass with ConstraintHandling {
       }
     case tp1: TypeVar if tp1.isInstantiated =>
       tp1.underlying & tp2
-    case tp1: AnnotatedType =>
+    case tp1: AnnotatedType if !tp1.isRefining =>
       tp1.underlying & tp2
```
smarter: Why not rewrap the annotation around the intersection if it's a refining annotation?

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3679,22 +3702,33 @@ object Types {
...
+      derivedAnnotatedType(parent.stripTypeVar, annot)
+
+    override def stripAnnots(implicit ctx: Context): Type = parent.stripAnnots
+
+    private[this] var isRefiningKnown = false
+    private[this] var isRefiningCache: Boolean = _
 
-    override def stripAnnots(implicit ctx: Context): Type = tpe.stripAnnots
+    def isRefining(implicit ctx: Context) = {
+      if (!isRefiningKnown) {
```
smarter: Shouldn't the cache be invalidated at the start of a new run?

smarter:
This PR would really really benefits from having tests.

compiler/src/dotty/tools/dotc/core/Types.scala
gsps: I was thinking about that too, but decided that it's probably fine given the internal nature of such annotations. Then again, we do expose RefiningAnnotation in the public `scala.annotation` package.

PR #4627, opened 2018-06-08
Add Tasty compilation unit source file

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -41,6 +41,12 @@ object TastyImpl extends scala.tasty.Tasty {
 
   def ContextDeco(ctx: Context): AbstractContext = new AbstractContext {
     def owner: Definition = FromSymbol.definition(ctx.owner)(ctx)
+
+    def source: String = {
+      val thisPath = java.nio.file.Paths.get("").toAbsolutePath
+      val path = ctx.compilationUnit.source.file.jpath.toAbsolutePath
+      thisPath.relativize(path).toString
```
olafurpg: Some thoughts

- Not sure if consistent cross-platform behavior is a goal, but this will produce different paths in Windows and Linux/macOS. If you want normalized paths you can use `URI.normalize`

```scala
thisPath.toUri.relativize(path.toUri).getPath
```


- there are use-cases where you only care about the filename `source.getFileName.toString`, in that case it's kind of annoying to have to parse the string to extract the filename
- this relies on an implicit working directory `java.nio.file.Paths.get("").toAbsolutePath`, what happens for example with virtual files?

allanrenucci:
@nicolasstucki This PR broke master. Can you have a look?

nicolasstucki:
@allanrenucci there is already a fix in #4708

PR #4628, opened 2018-06-08
Fix #4600: method could be inside block in argument implicit resolution

### LINKED ISSUE #4600: Implicit Argument / vararg with Dynamic Trait fails
Consider the following simple example:

```scala
class ImplicitExample() extends Dynamic {
  def someMethod()(implicit s: String = "World"): String = s
  def applyDynamic(name: String)(args: Any*)(implicit s: String = "World"): String = name + s
}
```

And its test:

```scala
class ImplicitTest {
  @Test def t1(): Unit = {
    assertEquals("World", new ImplicitExample().someMethod())
  }

  @Test def t2(): Unit = {
    implicit val s: String = "Hello"
    assertEquals("Hello", new ImplicitExample().someMethod())
  }

  @Test def t3(): Unit = {
    assertEquals("runWorld", new ImplicitExample().run())
  }

  @Test def t4(): Unit = {
    implicit val s: String = "Hello"
    assertEquals("runHello", new ImplicitExample().run())
  }
}
```

Test `t1` and `t2` always work, while `t3` fails with the latest Nightly build (`0.9.0-bin-20180529-be6ae0e-NIGHTLY`):

```scala
[error] -- Error: ... ImplicitTest.scala: ...
[error] 15 |    assertEquals("runWorld", new ImplicitExample().run())
[error]    |                                                      ^
[error]    |no implicit argument of type String was found for parameter s of method applyDynamic in class ImplicitExample
[error] one error found
[error] (Test / compileIncremental) Compilation failed
```

If you remove the vararg `args`, it compiles but the implicit parameter is not considered anymore and `t4` fails with:

```scala
[error] Test ImplicitTest.t4 failed: expected:<run[Hello]> but was:<run[World]>
```

Everything is fine with `0.8.0-RC1`.

liufengyun:
Fix #4600: method could be inside block in argument implicit resolution

Scalac incorrectly rejects `tests/pos/i4600b.scala`, with this fix Dotty accepts the code.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -2258,7 +2258,7 @@ class Typer extends Namer
 
           // If method has default params, fall back to regular application
           // where all inferred implicits are passed as named args.
-          if (tree.symbol.hasDefaultParams && !propFail.isInstanceOf[AmbiguousImplicits]) {
+          if (methPart(tree).symbol.hasDefaultParams && !propFail.isInstanceOf[AmbiguousImplicits]) {
             val namedArgs = (wtp.paramNames, args).zipped.flatMap { (pname, arg) =>
               if (arg.tpe.isError) Nil else untpd.NamedArg(pname, untpd.TypedSplice(arg)) :: Nil
             }
--- a/tests/run/i4600.scala
+++ b/tests/run/i4600.scala
@@ -1,0 +1,37 @@
+import scala.language.dynamics
+
+class ImplicitExample() extends Dynamic {
+  def someMethod()(implicit s: String = "World"): String = s
+  def applyDynamic(name: String)(args: Any*)(implicit s: String = "World"): String = name + s
+}
+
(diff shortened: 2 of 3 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2258,7 +2264,7 @@ class Typer extends Namer
 
           // If method has default params, fall back to regular application
           // where all inferred implicits are passed as named args.
-          if (tree.symbol.hasDefaultParams && !propFail.isInstanceOf[AmbiguousImplicits]) {
+          if (methSymbol.hasDefaultParams && !propFail.isInstanceOf[AmbiguousImplicits]) {
```
allanrenucci: You could use `methPart(tree).symbol`

tests/pos/i4600.scala
```diff
@@ -0,0 +1,26 @@
+import scala.language.dynamics
```
allanrenucci: I would make this a run test

PR #4633, opened 2018-06-09
Fix case class decompilation

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -721,6 +737,10 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
 
       case Type.ThisType(tp) =>
         printType(tp)
+        tp match {
+          case Type.SymRef(cdef @ ClassDef(_, _, _, _, _), _) if !cdef.flags.isObject => this += ".this"
+          case _ => this
+        }
```
liufengyun: Could you explain why special case object here?

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -112,8 +112,24 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
           }
         }
 
+        def keepDefinition(d: Definition): Boolean = {
+          val flags = d.flags
+          import flags._
```
liufengyun: I think `import flags._` doesn't help in terms of readability: it takes me some time to realize that `isParam` is a method on `d.flags`. In `isCaseUnOverridableMethod`, it still uses `d.flags`.

A short doc can help better understand the motivation for the method `isCaseUnOverridableMethod`. A naive question: why not just ignore all synthetic definitions?

library/src/scala/tasty/util/ShowSourceCode.scala
nicolasstucki: When refering to the this type of a module we just refere directly to the module itself. For example `scala.Predef` and not `scala.Predef.this`.

library/src/scala/tasty/util/ShowSourceCode.scala
nicolasstucki: Will improve code and add docs. 

I decided to print the synthetic methods to allow users to see what was generated. I will add some `/* synthetic */` comment on them later. And I'm considering adding a mode that those not print them. I started printing them because I can test more code, therefore more of the Tasty reflect API.

PR #4635, opened 2018-06-09
Harden IDE

odersky:
I observed crashes when trying to compute the docstring of trees that did not have a position.
It happened for edited code with lots of syntax errors.

### DIFF
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -328,7 +328,9 @@ class DottyLanguageServer extends LanguageServer
     if (sym == NoSymbol) Nil.asJava
     else {
       val refs = Interactive.namedTrees(uriTrees, Include.references | Include.overriding, sym)
-      refs.map(ref => new DocumentHighlight(range(ref.namePos), DocumentHighlightKind.Read)).asJava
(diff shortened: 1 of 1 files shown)


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -328,7 +328,9 @@ class DottyLanguageServer extends LanguageServer
     if (sym == NoSymbol) Nil.asJava
     else {
       val refs = Interactive.namedTrees(uriTrees, Include.references | Include.overriding, sym)
-      refs.map(ref => new DocumentHighlight(range(ref.namePos), DocumentHighlightKind.Read)).asJava
+      ( for (ref <- refs if ref.namePos.exists)
```
allanrenucci: I've hit on the same issue when doing syntax highlighting. The problem is that `namePos` is nonsensical for some special trees. For example trees named `<init>` or `<error>`. This probably doesn't fix the problem because you still get a position that exists but is wrong. I tried to fix it in #4548, but that's not the correct fix. `namePos` should probably return `NoPosition` for "synthetic" names. But I don't think we have such think as "synthetic" names.

allanrenucci:
> @allanrenucci Do you have a reproducible way to reproduce that?

Yes, if you try to parse `class`, you get a `TypeDef` tree with a nonsensical `namePos`

PR #4639, opened 2018-06-09
Fix #4430: Disallow access to Java bridge methods before erasure

### LINKED ISSUE #4430: Method calls can produce signatures referencing unaccessible methods in Dotty when Scalac doesn't
On [Gitter](https://gitter.im/lampepfl/dotty?at=5ae814b6dea1b95c10f276c0), @jiminhsieh reports that Dotty doesn't run on Eclipse OpenJ9 (because of its version of the standard library doesn't match in some internals ABI details), because Dotty outputs (for https://github.com/lampepfl/dotty/blob/0.7.0/compiler/src/dotty/tools/dotc/util/NameTransformer.scala#L98) a call to `java/lang/StringBuilder.append([CII)Ljava/lang/AbstractStringBuilder;` rather than `java/lang/StringBuilder.append:([CII)Ljava/lang/StringBuilder;`, which is what Scalac 2.12.5 produces.

`AbstractStringBuilder` is a package-accessible class that isn't part of the specified Java standard library, and `StringBuilder.append` is declared to return `StringBuilder` not `AbstractStringBuilder`; the signature that Scalac uses exists on the reference implementation, but just because that implementation also defines `public AbstractStringBuilder.append(...): AbstractStringBuilder` which is inherited by `StringBuffer`.

## Minimization

That can be verified by compiling this code on a standard JDK (I'm using Oracle JDK 1.8.0_162) and calling `javap -c *.class` on the result:
```scala
object Foo {
  def main() = {
    (new java.lang.StringBuilder()).append(Array[Char](), 0, 0)
  }
}
```

## Analysis and blame

Is Dotty or OpenJ9's library at fault?

Scalac's result would work here, and JLS overload resolution would forbid emitting that call from Java, because [§15.12.2](https://docs.oracle.com/javase/specs/jls/se10/html/jls-15.html#jls-15.12.2) says
> This step uses the name of the method and the argument expressions to locate methods that are both accessible and applicable, that is, declarations that can be correctly invoked on the given arguments.

and accessibility of a method is defined in [§6.6.1](https://docs.oracle.com/javase/specs/jls/se10/html/jls-6.html#jls-6.6.1) and requires the defining class to be accessible too:
> A member (class, interface, field, or method) of a reference type, or a constructor of a class type, is accessible only if the type is accessible and the member or constructor is declared to permit access:

OTOH, it's not clear from [the JLS binary compatibility spec](https://docs.oracle.com/javase/specs/jls/se10/html/jls-13.html#jls-13.2) that the OpenJ9 definition of `AbstractStringBuilder` is binary compatible with the original one, but I think that might be because the definition isn't specific enough.

EDIT: I meant `*StringBuilder` throughout.

odersky:
Scala bridge methods only come to existence during erasure, so are not visible
before. But Java bridge methods come with their class files. It first I thought
we should get parity, so have to ignore Java bridge methods before erasure. But
that did not work, because unlike for Scala, some Java bridges are generated
to allow access to inherited methods. The situation can be described using
StringBuilder. It inherits a class AbstractStringBuilder which is not accessible
from Scala. Here's a minimized version highlighting the problem:

    private class AbstractStringBuilder {
      def length: Int = ...
      def append(xs: Array[Char]): AbstractStringBuilder = ...
    }
    class StringBuilder extends AbstractStringBuilder {
      override def append(xs: Array[Char]): StringBuilder = ...
      <bridge> def length: Int = super.length
      <bridge> def append(xs: Array[Char]): AbstractStringBuilder = this.append(xs)
    }

User code

    val sb = new StringBuilder
    sb.length      // needs to access bridge in StringBuilder
    sb.append(arr) // should not access bridge in StringBuilder

In the first case, `sb.length` needs to access the bridge method, as
the inherited method is inaccssible. So we cannot simply hide bridge
methods before erasure, as we do for Scala bridges. But in the second
case, we should not access the bridge append, because it refers to an
inaccessible class in its result type.

The fix is simple, but it took me a long time to find it: When merging denotations
with the same signature (which is the case for the two appends, since results are
ignored), always prefer a non-bridge over a bridge as the symbol of a JointRefDenotation.

### DIFF
--- a/tests/run/i4430.scala
+++ b/tests/run/i4430.scala
@@ -1,0 +1,7 @@
+object Test {
+  def main(args: Array[String]): Unit = {
+    val sb = new java.lang.StringBuilder()
+    sb.append(Array[Char](), 0, 0)
+    sb.length
+  }
+}
--- a/compiler/src/dotty/tools/dotc/config/Config.scala
+++ b/compiler/src/dotty/tools/dotc/config/Config.scala
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = false
+  final val tracingEnabled = true
 
   /** Initial capacity of uniques HashMap.
    *  Note: This MUST BE a power of two to work with util.HashSet
--- a/compiler/src/dotty/tools/dotc/config/Config.scala
+++ b/compiler/src/dotty/tools/dotc/config/Config.scala
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = true
+  final val tracingEnabled = false
 
   /** Initial capacity of uniques HashMap.
    *  Note: This MUST BE a power of two to work with util.HashSet
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -488,6 +488,7 @@ object Denotations {
               (!sym2.isAsConcrete(sym1) ||
                 precedes(sym1.owner, sym2.owner) ||
                 accessBoundary(sym2).isProperlyContainedIn(accessBoundary(sym1)) ||
+                sym2.is(Bridge) && !sym1.is(Bridge) ||
                 sym1.is(Method) && !sym2.is(Method)) ||
               sym1.info.isErroneous)
 
--- a/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala
+++ b/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala
@@ -264,4 +264,27 @@ class TestBCode extends DottyBytecodeTest {
       }
     }
   }
+
+  // See #4430
+  @Test def javaBridgesAreNotVisible = {
+    val source =
+      """
+        |class Test {
+        |  def test = (new java.lang.StringBuilder()).append(Array[Char](), 0, 0)
+        |}
+      """.stripMargin
+
+    checkBCode(source) { dir =>
+      // We check the method call signature to make sure we don't call a Java bridge
+      val clsIn = dir.lookupName("Test.class", directory = false).input
+      val clsNode = loadClassNode(clsIn)
+      val testMethod = getMethod(clsNode, "test")
+      val instructions = instructionsFromMethod(testMethod)
+      val containsExpectedCall = instructions.exists {
+        case Invoke(_, "java/lang/StringBuilder", "append", "([CII)Ljava/lang/StringBuilder;", _) => true
+        case _ => false
+      }
+      assertTrue(containsExpectedCall)
+    }
+  }
 }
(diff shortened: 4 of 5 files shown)


tests/run/i4430.scala
```diff
@@ -0,0 +1,5 @@
+object Test {
+  def main(args: Array[String]): Unit = {
+    (new java.lang.StringBuilder()).append(Array[Char](), 0, 0)
```
smarter: This should be a [DottyBytecodeTests](https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala) that checks the call signature.

smarter:
@retronym do you know how scalac deals with Java bridges?

tests/run/i4430.scala
odersky: Can somebody else do the bytecode test? It's one of the few things I don't want to get into.

odersky:
Turns out, `accessibleFrom` was the wrong way to do it since the compiler sometimes selected a bridge and then complained it was not accessible. A more precise way is in the current commit 1bef624: Mark them as `private` before erasure, and drop the `private` flag as part of the erasure transform. That brings them more in line with normal bridge methods. Also, the issue is contained better, needing changes in just `ClassfileParser` and `Erasure`.

odersky:
Still wrong. This is a lot trickier than I thought. Turns out some Java bridges need to be accessible from the start.

smarter:
I think something else is going on here. If I run the added test case with -Xprint-types -Xprint-types:all, the type of append is initially correct (the result type is StringBuilder), but after Erasure this changes (the result type is now AbstractStringBuilder), so it looks like Erasure is redoing overloading resolution somehow and getting it wrong.

odersky:
@allanrenucci Can I ask you to add the bytecode test? Thanks!

odersky:
@smarter I believe it is `mergeDenot` that is the culprit. Erasure does not perform any overloading resolution to my knowledge.

retronym:
By default, `scalac` excludes members that had the `ACC_BRIDGE` flag from `Type.member` etc.

```
scala> typeOf[java.lang.String].memberBasedOnName(TermName("compareTo"), excludedFlags = 0L).alternatives.map(_.defString)
res11: List[String] = List(def compareTo(x$1: Any): Int, def compareTo(x$1: String): Int)

scala> typeOf[java.lang.String].memberBasedOnName(TermName("compareTo"), excludedFlags = reflect.internal.Flags.BridgeFlags).alternatives.map(_.defString)
res12: List[String] = List(def compareTo(x$1: String): Int)
```

odersky:
@retronym 

> By default, scalac excludes members that had the ACC_BRIDGE flag from Type.member etc.

That's what I tried as well, but then `new StringBuilder().length` resolves to the one in `AbstractStringBuilder`, which gives an illegal access error at runtime. How does `scalac` handle this?

odersky:
> The backend uses the qualifier type as the class of CONSTANT_Methodref in the INVOKEVIRTUAL, rather than the owner of the invoked method. The method descriptor does come directly from the symbol itself.

Ah, that would explain it. We have to check in the backend to see whether it does the right thing and to fix it if not.

odersky:
I believe the change in this PR is harmless: when in doubt it is better to select a non-bridge over a bridge. Ideally, the backend would compensate. But is somebody ready to do this right now? Or should we wait for the merged backend. In the latter case I propose to merge this PR, so that we have one important issue less to deal with.

PR #4642, opened 2018-06-10
Fix tasty reflect singletons

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -621,6 +622,7 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
 
       case TypeTree.Singleton(ref) =>
         printTree(ref)
+        this += ".type"
```
liufengyun: How Constant types are handled?

nicolasstucki: They are handled by `printConstant`.  I will check that they are printed correctly.

nicolasstucki: @liufengyun fixed the literal constant singleton types  to not print `".type"`

liufengyun:
LGTM

PR #4643, opened 2018-06-10
Cache context base base in quote driver

nicolasstucki:
This enables caching the compiler used of call `Expr.run`.


compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
```diff
@@ -2,20 +2,21 @@ package dotty.tools.dotc.quoted
...
 import java.net.URLClassLoader
 
-import Toolbox.{Run, Settings, Show}
 import dotty.tools.dotc.tastyreflect.TastyImpl
 
 class QuoteDriver extends Driver {
   import tpd._
 
-  def run[T](expr: Expr[T], settings: Settings[Run]): T = {
+  private[this] val contextBase: ContextBase = new ContextBase
```
smarter: I don't think this is thread-safe

nicolasstucki: It is not and not intended to be. Each thread should generate its own toolbox. I will add documentation and thick how to check that they do not use the same compiler in two threads.

nicolasstucki: Decided to make it thread-safe by synchronization instead. If someone wants parallelism they should create several instances of the `Toolbox`

nicolasstucki:
Rebased

nicolasstucki:
Rebased and reenabled test

biboudis:
Also had a look. Seems good.

compiler/src/dotty/tools/dotc/quoted/QuoteDriver.scala
smarter: > Decided to make it thread-safe by synchronization instead.

I don't see any synchronization logic in the code.

nicolasstucki: I must have lost that commit in one of the reading. I will add it again.

Blaisorblade: Searching for `synchronized` still finds nothing here?

nicolasstucki: It is in #4746

PR #4644, opened 2018-06-10
Escape chars when printing strings in decompiler

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -770,6 +772,19 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
     def +=(x: Char): this.type = { sb.append(x); this }
     def +=(x: String): this.type = { sb.append(x); this }
 
+    @switch private def escapedChar(ch: Char): String = ch match {
```
liufengyun: `(ch @switch) match` ?

nicolasstucki: I copy pasted it from the `PlainPrinter`. Will double check if it works or change it in both places.

nicolasstucki: As expected the check was not done. I change it along with the other I found.

PR #4648, opened 2018-06-12
Show docs in the IDE

Duhemm:
Builds on top of #4461.

Rendering could certainly be improved, especially for very long docstrings where one needs to scroll to the type info. Maybe the type info should be displayed above the docstring?

I don't know what's the best way to display it would be, but I'm very open to suggestions.

### DIFF
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -19,7 +19,7 @@ import scala.io.Codec
 import dotc._
 import ast.{Trees, tpd}
 import core._, core.Decorators.{sourcePos => _, _}
-import Comments.Comment, Contexts._, Flags._, Names._, NameOps._, Symbols._, SymDenotations._, Trees._, Types._
+import Comments._, Contexts._, Flags._, Names._, NameOps._, Symbols._, SymDenotations._, Trees._, Types._
 import classpath.ClassPathEntries
 import reporting._, reporting.diagnostic.MessageContainer
 import util._
@@ -346,7 +346,6 @@ class DottyLanguageServer extends LanguageServer
 
     if (tpw == NoType) new Hover
     else {
-      import dotty.tools.dotc.core.Comments._
       val symbol = Interactive.enclosingSourceSymbol(trees, pos)
       val docComment = ctx.docCtx.flatMap(_.docstring(symbol))
       val markedStrings = docMarkedStrings(docComment, tpw.show.toString)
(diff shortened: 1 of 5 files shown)


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -338,13 +338,17 @@ class DottyLanguageServer extends LanguageServer
...
 
     if (tpw == NoType) new Hover
     else {
-      val str = tpw.show.toString
-      new Hover(List(JEither.forLeft(str)).asJava, null)
+      import dotty.tools.dotc.core.Comments._
+      val symbol = Interactive.enclosingSourceSymbol(trees, pos)
+      val docComment = ctx.docCtx.flatMap(_.docstring(symbol))
+      val markedString = docMarkedString(docComment, tpw.show.toString)
+      new Hover(List(JEither.forRight(markedString)).asJava, null)
```
olafurpg: What if you return two marked strings? One for the docstring and one for the info. The `textDocument/hover` response can return `MarkedString[]` arrays https://microsoft.github.io/language-server-protocol/specification#textDocument_hover

Duhemm: Thanks for the suggestion! I've changed it to return 2 different `MarkedString`s (one with the doc, the other with the type info). Unfortunately, this doesn't improve much the layout on VSCode: it simply adds an horizontal delimiter between the two hovers, but you still need to scroll.

I've inverted the order of the two `MarkedString`.

smarter:
> Rendering could certainly be improved, especially for very long docstrings where one needs to scroll to the type info. Maybe the type info should be displayed above the docstring?

Yes, open a typescript file and see what happens when you hover over something, we should match this. The signature is above the documentation, the two are separate by a thin horizontal line, the documentation is shown as rendered markdown.

Duhemm:
Just rebased. Do we agree that the formatting of the comments can be improved in a separate PR?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -340,13 +340,17 @@ class DottyLanguageServer extends LanguageServer
...
 
     val pos = sourcePosition(driver, uri, params.getPosition)
-    val tp = Interactive.enclosingType(driver.openedTrees(uri), pos)
+    val trees = driver.openedTrees(uri)
+    val tp = Interactive.enclosingType(trees, pos)
     val tpw = tp.widenTermRefExpr
 
     if (tpw == NoType) new Hover
     else {
+      import dotty.tools.dotc.core.Comments._
```
smarter: I would put the import at the top of the file.

PR #4651, opened 2018-06-12
Inline fixes

allanrenucci:


allanrenucci:
@odersky Do the included test cases exercise the inline fixes?

compiler/src/dotty/tools/dotc/transform/AccessProxies.scala
```diff
@@ -70,6 +71,11 @@ abstract class AccessProxies {
     def accessorNameKind: ClassifiedNameKind
     def needsAccessor(sym: Symbol)(implicit ctx: Context): Boolean
 
+    def ifNoHost(reference: RefTree)(implicit ctx: Context): Tree = {
```
allanrenucci: This is overridden in both `ProtectedAccesors` and `Inliner`, shouldn't it be abstract?

odersky:
I think we should just merge #4490 instead. It improves what we have so far, and `Lst` is useless without it.

PR #4654, opened 2018-06-13
Cache initial context in REPL

allanrenucci:


nicolasstucki:
Maybe you should cache the context base like I'm doing in #4643.

PR #4658, opened 2018-06-14
add documentation for compiler plugin

liufengyun:
add documentation for compiler plugin


docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+`plugin.properties` in the root directory of the jar. The property file
+specifies the fully qualified plugin class name. The format of a property file
+looks like the following:
+
+```
+pluginClass=dividezero.DivideZero
+```
+
+The above is a change from Scalac, which depends on an XML file
+`scalac-plugin.xml`. Since SBT 1.1.5, SBT also supports Dotty compiler plugins:
```
allanrenucci: Starting with 1.1.5, sbt also...

tests/plugins/neg/divideZero-research/plugin_1.scala
```diff
@@ -10,13 +10,13 @@ import Symbols.Symbol
 import Constants.Constant
 import StdNames._
 
-class DivideZero extends MiniPhase with ResearchPlugin {
+class DivideZero extends PluginPhase with ResearchPlugin {
```
allanrenucci: Since the CI only run the sbt scripted tests on nightlies, can you confirm that you ran this test locally

liufengyun: This test is a neg test, not a scripted test. I just checked it works locally. I confused my self, for research plugins, we don't need the phases to be `PluginPhase`.

docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+- No more support for analyzer plugins
+- Added support for research plugins
+
+[Analyzer plugins][1] in Scalac are executed during type checking to change the
+normal type checking. This is a nice feature for doing research, but for
+production usage, a predictable and consistent type checker is more important.
+
+For experiments and researches that rely on analyzer plugins in Scalac,
+_research plugin_ can be used for the same purpose in Dotty. Research plugins
+are more powerful than Scalac analyzers as it enables plugin authors to
```
Duhemm: "Than Scalac analyzer plugins"

Duhemm: as they enable

docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+customize the whole compiler pipeline. That means, you can easily use your
+customized typer to replace the standard typer, or roll your own parser for
+your domain-specific language. Research plugins are only enabled for nightly or
+snaphot releases of Dotty.
+
+The common plugins that add new phases to the compiler pipeline are called
```
Duhemm: I'd switch the order of these 2 paragraphs, let's describe first the common case?

docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+_standard plugins_ in Dotty. In terms of features, they are similar to
+Scalac plugins, despite minor changes in the API.
+
+## Physical Interface
```
Duhemm: Physical interface reads like a USB port, I'd call this section "Using compiler plugins" and have a separate subsection for the command line and for sbt.

docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+The compiler will examine the jar provided, and look for a property file
+`plugin.properties` in the root directory of the jar. The property file
+specifies the fully qualified plugin class name. The format of a property file
+looks like the following:
+
+```
+pluginClass=dividezero.DivideZero
+```
```
Duhemm: Is this relevant to this section? Maybe this should go in a section called "Writing compiler plugins"?

docs/docs/reference/changed/compiler-plugins.md
```diff
@@ -0,0 +1,133 @@
...
+
+The above is a change from Scalac, which depends on an XML file
+`scalac-plugin.xml`. Starting from 1.1.5, SBT also supports Dotty compiler plugins:
+
+```Scala
+addCompilerPlugin("org.divbyzero" % "divbyzero" % "1.0")
+```
+
+## Standard Plugin
```
Duhemm: Maybe call this "Writing your first plugin", and move the content of `plugin.properties` here?

docs/docs/reference/changed/compiler-plugins.md
liufengyun: I agree, but it seems the 2nd paragraph follows more naturally after the 1st paragraph. If we switch the 3rd and the 2nd, we have a problem between 1st and 3rd.

docs/docs/reference/changed/compiler-plugins.md
liufengyun: That can be more friendly, but I feel the tutorial-like style doesn't fit the documentation style of other pages.

docs/docs/reference/changed/compiler-plugins.md
liufengyun: Maybe we can call it plugin artifact interface?

liufengyun:
Otherwise, LGTM. Thanks a lot @allanrenucci @Duhemm !

PR #4662, opened 2018-06-14
Avoid typing annotations twice

smarter:
`typedAnnotation` was used both in `Namer#addAnnotations` and
`Typer#completeAnnotations`, resulting in some annotations being typed
twice. Fixed by introducing a `typedAheadAnnotation` (and renaming an
existing method with the same name that doesn't actually type the full
annotation tree).

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -745,10 +745,11 @@ class Namer { typer: Typer =>
       while (c.owner != target) c = c.outer
       c
     }
+
     for (annotTree <- untpd.modsDeco(stat).mods.annotations) {
-      val cls = typedAheadAnnotation(annotTree)(annotCtx)
+      val cls = typedAheadAnnotationClass(annotTree)(annotCtx)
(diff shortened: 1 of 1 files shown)


smarter:
@odersky Both `Namer#addAnnotations` and `Typer#completeAnnotations` define a `lazy val annotCtx`, but the definitions are different and don't always produce the same result, compare:
https://github.com/lampepfl/dotty/blob/78a5e092a834b76c54b18d5e58cb0a593f4bb8a5/compiler/src/dotty/tools/dotc/typer/Namer.scala#L741-L747
https://github.com/lampepfl/dotty/blob/78a5e092a834b76c54b18d5e58cb0a593f4bb8a5/compiler/src/dotty/tools/dotc/typer/Typer.scala#L1345-L1355

Is one "more correct" than the other? Can we come up with a single definition that we can use in both?

smarter:
There's another issue which is that we're not actually using the annotCtx with Annotation.deferred:
https://github.com/lampepfl/dotty/blob/78a5e092a834b76c54b18d5e58cb0a593f4bb8a5/compiler/src/dotty/tools/dotc/typer/Namer.scala#L751
The ctx that will be passed here will just be the enclosing context.

PR #4664, opened 2018-06-14
Fix -Ythrough-tasty

smarter:
- dotty.tools.io.Path#extension always returns something lowercase for
  some reason, so comparing against "hasTasty" always returned false
- .tasty files were not supported

### DIFF
--- a/compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
+++ b/compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
@@ -35,11 +35,11 @@ object Debug {
 
     val fromTastyOut = Files.createDirectory(tmpOut.resolve("from-tasty"))
 
-    val ext = "hasTasty"
-    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && x.extension == ext).map { x =>
+    val extensions = List("tasty", "hasTasty").map(_.toLowerCase)
+    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && extensions.exists(_ == x.extension.toLowerCase)).map { x =>
       val source = x.toString
       // transform foo/bar/Baz.hasTasty into foo.bar.Baz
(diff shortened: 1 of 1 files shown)


compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -35,11 +35,11 @@ object Debug {
 
     val fromTastyOut = Files.createDirectory(tmpOut.resolve("from-tasty"))
 
-    val ext = "hasTasty"
-    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && x.extension == ext).map { x =>
+    val extensions = List("tasty", "hasTasty")
```
nicolasstucki: `List("tasty", "hasTasty").toLowerCase` intead of on each file

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -35,11 +35,11 @@ object Debug {
...
+    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && extensions.exists(_.toLowerCase == x.extension)).map { x =>
```
nicolasstucki: Is extension supposed to always return lower cases? If not, we should do `x.extension.toLowerCase` to make sure it is as we expected.

compiler/src/dotty/tools/dotc/fromtasty/Debug.scala
```diff
@@ -35,11 +35,11 @@ object Debug {
 
     val fromTastyOut = Files.createDirectory(tmpOut.resolve("from-tasty"))
 
-    val ext = "hasTasty"
-    val classes = Directory(fromSourcesOut).walk.filter(x => x.isFile && x.extension == ext).map { x =>
+    val extensions = List("tasty", "hasTasty").toLowerCase
```
nicolasstucki: `List("tasty", "hasTasty").map(_.toLowerCase)`, my bad

PR #4665, opened 2018-06-14
Fix position of Inline#expansion

smarter:
Using `call.pos` here is very wrong since the positions of nodes insides
Inline#expansion should be positions inside the file in which the inline
method is defined, but `call.pos` is a position inside the file where
the inline method is called. See `TreeCopier#inlineContext`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/untpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/untpd.scala
@@ -280,7 +280,14 @@ object untpd extends Trees.Instance[Untyped] with UntypedTreeInfo {
   def Try(expr: Tree, cases: List[CaseDef], finalizer: Tree): Try = new Try(expr, cases, finalizer)
   def SeqLiteral(elems: List[Tree], elemtpt: Tree): SeqLiteral = new SeqLiteral(elems, elemtpt)
   def JavaSeqLiteral(elems: List[Tree], elemtpt: Tree): JavaSeqLiteral = new JavaSeqLiteral(elems, elemtpt)
-  def Inlined(call: tpd.Tree, bindings: List[MemberDef], expansion: Tree): Inlined = new Inlined(call, bindings, expansion)
+  def Inlined(call: tpd.Tree, bindings: List[MemberDef], expansion: Tree): Inlined = {
+    val tree = new Inlined(call, bindings, expansion)
+    // `expansion.pos` is a position inside the file in which the inline method
+    // is defined (see `TreeCopier#inlineContext` and `Decorators#sourcePos`),
+    // so it must not be taken into account when setting a position for the
+    // Inlined tree.
+    tree.withPos(call.pos)
+  }
   def TypeTree() = new TypeTree()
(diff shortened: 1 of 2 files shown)


smarter:
Hmm, this is trickier than I thought. `tests/neg/i1992.scala` illustrates an interesting issue, given the code:
```scala
object Test {
  def identity[T](x: T): T = x
  def test = {
    val z: Int = identity(z)
  }
}
```
The error we display is:
```scala
-- [E039] Reference Error: try/inline-error.scala:4:26 -------------------------
4 |    val z: Int = identity(z)
  |                          ^
  |           `z` is a forward reference extending over the definition of `z`
```
But If make `identity` an `inline def` then instead I get:
```scala
-- [E039] Reference Error: try/inline-error.scala:2:36 -------------------------
2 |  inline def identity[T](x: T): T = x
  |                                    ^
  |           `z` is a forward reference extending over the definition of `z`
```
Because `z` is now inside an inline expansion, so its position is inside the inlined body.

Here it seems that we've reached the limit of our current scheme for positions, where every tree has a `pos` field that specifies a position relative to some source file, and `Decorators#sourcePos` converts that into a SourcePosition, using the InlinedCalls property to keep track of expansions. The problem is that within an expansion we have references to arguments of the inline call that we'd expect to keep the original position of the argument. To be able to represent this it seems that having a field of type `Position` in each tree is not enough, we'd need a `SourcePosition` or something that can be translated into a `SourcePosition` (or maybe keep the position in an annotation? Something like that could be useful for source code generation too).

Thoughts?

Blaisorblade:
> The problem is that within an expansion we have references to arguments of the inline call that we'd expect to keep the original position of the argument.

Storing `SourcePosition` seems simpler but more expensive, but it seems you could just record *two* source-file changes. The inlined code comes from the function definition, but inside there the arguments come from the source, so you need to switch containing file twice, but it should be possible to do it the same way. Since the correct compiler context for the argument is the call site, could you use `inlineContext` for arguments too, if other code were adapted accordingly?

odersky:
Using full source positions is almost certainly too expensive. I wonder how widespread the problem is though. Normally, arguments of inline calls are right hand sides in bindings, and those bindings can have their original position. So it seems we get the problem only if we avoid creating a binding because the argument is a stable identifier. Maybe an attachment recording this fact would do the trick?

But as far as this PR goes, LGTM.

smarter:
> So it seems we get the problem only if we avoid creating a binding because the argument is a stable identifier.

Oh right, I totally missed that distinction. 

> Maybe an attachment recording this fact would do the trick?

Yes, but wouldn't it be simpler and more regular to create bindings even for stable identifier?

odersky:
> Yes, but wouldn't it be simpler and more regular to create bindings even for stable identifier?

You are right. I was worried about path equality but we can fix that by giving these bindings a singleton type.

bishabosha:
@smarter is this still relevant?

smarter:
I don't remember at all what this did but this seems obsolete /cc @nicolasstucki

Issue #4670, opened 2018-06-15
Make MethodTypes first-class


Blaisorblade:
Recently in #2500, @milessabin @LukaJCB and @LPTK advocated turning method types (with the existing syntax shown into method types) into value types. From my summary:

> With MethodTypes, [X]F[X] becomes internally Function0[F[_]] and then after erasure Function0 (and this is hidden from the typechecker). That is, method types are just functions, like in Haskell. This is just the standard and well-understood universal quantification in System F, and also should work fine with effects. It'd still be good to formalize this carefully.

Since such MethodTypes always need to be applied at runtime, they should avoid issues with polymorphic references that appeared in #2500.
This variant appears to avoid lots of other issues discussed there, and it seems what @odersky considered in https://github.com/lampepfl/dotty/pull/1560#issuecomment-252549931. He still said this would require a SIP tho, so that should be part of this issue.

sstucki:
Quoting @LPTK's question from #2500, as I am wondering about the same thing...
 
>> With MethodTypes, `[X]F[X]` becomes internally `Function0[F[_]]`
>
> Isn't the latter equivalent to `Function0[F[X] forSome { type X }]`, so very different from the former?

LukaJCB:
I'm confused, if `val foo: [X]F[X]` is equivalent to `def foo[X]: F[X]`, then it should be a universal quantifier, not an existential one, no?

Blaisorblade:
I probably should just write that MethodTypes just always erase to FunctionTypes, even if they take no value arguments; that's less confusing.

I only wrote that (informally) to represent the first erasure step — I want to say that `[X]F[X]` erases to `Function0[F]` and then to `Function0`, but `Function0[F]` isn't a real thing and erasure doesn't happen step-by-step.

fommil:
Not to state the obvious, but this will have implications for tooling that expects the old form (well, whatever tools that can be migrated to scala 3 that is).

LPTK:
One reason I was [showing](https://github.com/lampepfl/dotty/issues/2500#issuecomment-397300984) there is a pretty direct encoding of polymorphic methods using implicit dependent function types is that I think they should behave pretty much the same, at least for users.
I.e., it should be possible to omit type abstraction when defining a method value, just like one can omit the implicit parameters of a lambda when they can be inferred from the expected type:

```scala
// Example:
def m[A](x: A): List[A] = a :: Nil
// made first-class:
[...]

Issue #2500, opened 2017-05-22
Language request: Add forAll quantifier / Make MethodTypes first-class


japgolly:
We already have `forSome`, how about adding a `forAll` quantifier?
Haskell, for instance, has this.

odersky:
`forSome` has been deprecated 😄

notxcain:
@odersky anyway, why not add some means of universal quantification?

odersky:
@notxain Somebody has to study the theory and make sure it does not add an unsoundness hole as forSome did.

Blaisorblade:
@odersky That's generally good, but unsoundness shouldn't be a danger here, because we already know how to encode  first-class polymorphic values into a sound DOT—if we keep introduction and elimination explicit (unlike `forSome`), sth. like `[A] => t` (this is **just abstract syntax**). And we should keep intro and elims explicit (at least at first) because we know inference is very hard.

You presented in WadlerFest DOT the encoding from System F_<: to System D_<: and proved it type-preserving. I don't see an embedding of F_{<:>} into D_{<:>}, or a proof of meaning preservation—is that the problem?

For reference, the interesting rules for the type encoding are
```
translate(X) = X.Type
translate(forall X >: L <: U. T)` gives to `(x: {type Type >: translate(L) <: translate(U)}) => translate(T)
```

Also, you also suggested this ticket yourself here:

https://github.com/lampepfl/dotty/pull/1560#issuecomment-252549931

Blaisorblade:
For others: In non-PL speak, I'm just saying that, it seems, we can already translate `forAll X. T[X]` into Scala with a standard encoding, where `forAll X >: L <: U. T[X]` becomes `Forall[L, U, T]` , with `Forall` defined as

```scala
trait Forall[L, U >: L, T[X >: L <: U]] {
  def apply[X >: L <: U]: T[X]
}
```
and a value `[X]. t` become `new Forall[T] {def apply[X]: T[X] = t}`.

BTW, in `Forall` the compiler seems to require eager checking against bad bounds (which is why I needed to bind `U >: L` instead of just `U`). Is this motivated by theory? I understand details (including soundness of what's *already* in the compiler) are ongoing work.

Forgot to mention: I *can* see some details with unclear foundations, but they seem to be very limited.

japgolly:
Awesome @Blaisorblade, I was thinking the same thing!

> I can see some details with unclear foundations, but they seem to be very limited.

japgolly:
[...]

PR #4672, opened 2018-06-15
[Proof of concept] Polymorphic function types

smarter:
This is a sketch of how polymorphic function types could be implemented in Dotty, this PR adds types that look like:
```scala
[T <: AnyVal] -> List[T] => List[(T, T)]
```
and a way of writing values of these types:
```scala
val f = [T <: AnyVal] -> (x: List[T]) => x.map(e => (e, e))
```
Just like `P => R` has a method `def apply(x: A): B`, `[T] => P => R` has a method `def apply[T](x: P): R` (in fact, it doesn't have any other method). Behind the scene, this is erased to a regular `scala.FunctionN`, see the commit messages for more details.

This PR isn't complete, this is just a quick experiment done in a few hours. In particular, I haven't tried to work out how this interact with implicit/erased/dependent function types and SAM types. I also haven't tried to implement polymorphic eta-expansion yet, e.g. given:
```scala
def foo(x: [T] -> T => Option[T]): Unit
```
we'd like to be able to write:
```scala
foo(Some)
```
but for now you'll have to write:
```scala
foo([T] -> (x: T) => Some(x))
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1430,22 +1430,35 @@ object desugar {
     }
 
     val desugared = tree match {
-      case PolyFunction(targs, body) if (ctx.mode.is(Mode.Type)) =>
-        // Desugar [T_1, ..., T_M] -> (P_1, ..., P_N) => R
-        // Into    scala.PolyFunction { def apply[T_1, ..., T_M](x$1: P_1, ..., x$N: P_N): R }
-        val Function(vargs, resType) = body
+      case PolyFunction(targs, body) =>
+        val Function(vargs, res) = body
         // TODO: Figure out if we need a `PolyFunctionWithMods` instead.
         val mods = body match {
           case body: FunctionWithMods => body.mods
           case _ => untpd.EmptyModifiers
         }
+        val polyFunctionTpt = ref(defn.PolyFunctionType)
         val applyTParams = targs.asInstanceOf[List[TypeDef]]
-        val applyVParams = vargs.zipWithIndex.map { case (p, n) =>
-          makeSyntheticParameter(n + 1, p).withAddedFlags(mods.flags)
+        if (ctx.mode.is(Mode.Type)) {
+          // Desugar [T_1, ..., T_M] -> (P_1, ..., P_N) => R
+          // Into    scala.PolyFunction { def apply[T_1, ..., T_M](x$1: P_1, ..., x$N: P_N): R }
+
+          val applyVParams = vargs.zipWithIndex.map { case (p, n) =>
+            makeSyntheticParameter(n + 1, p).withAddedFlags(mods.flags)
+          }
+          RefinedTypeTree(polyFunctionTpt, List(
+            DefDef(nme.apply, applyTParams, List(applyVParams), res, EmptyTree)
+          ))
+        } else {
+          // Desugar [T_1, ..., T_M] -> (x_1: P_1, ..., x_N: P_N) => body
+          // Into    new scala.PolyFunction { def apply[T_1, ..., T_M](x_1: P_1, ..., x_N: P_N) = body }
+
+          val applyVParams = vargs.asInstanceOf[List[ValDef]]
+            .map(varg => varg.withAddedFlags(mods.flags | Param))
+          New(Template(emptyConstructor, List(polyFunctionTpt), Nil, EmptyValDef,
+            List(DefDef(nme.apply, applyTParams, List(applyVParams), TypeTree(), res))
+          ))
         }
-        RefinedTypeTree(ref(defn.PolyFunctionType), List(
-          DefDef(nme.apply, applyTParams, List(applyVParams), resType, EmptyTree)
-        ))
       case SymbolLit(str) =>
         Literal(Constant(scala.Symbol(str)))
       case InterpolatedString(id, segments) =>
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1448,7 +1448,7 @@ object desugar {
               ))
         }
       case _ =>
-        EmptyTree //may happen for erroneous input
+        EmptyTree // may happen for erroneous input. An error will already have been reported.
     }
 
     // begin desugar
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1418,20 +1418,8 @@ object desugar {
       }
     }
 
-    // begin desugar
-
-    // Special case for `Parens` desugaring: unlike all the desugarings below,
-    // its output is not a new tree but an existing one whose position should
-    // be preserved, so we shouldn't call `withPos` on it.
-    tree match {
-      case Parens(t) =>
-        return t
-      case _ =>
-    }
-
-    val desugared = tree match {
-      case PolyFunction(targs, body) =>
-        val Function(vargs, res) = body
+    def makePolyFunction(targs: List[Tree], body: Tree): Tree = body match {
+      case  Function(vargs, res) =>
         // TODO: Figure out if we need a `PolyFunctionWithMods` instead.
         val mods = body match {
           case body: FunctionWithMods => body.mods
@@ -1455,10 +1443,28 @@ object desugar {
 
           val applyVParams = vargs.asInstanceOf[List[ValDef]]
             .map(varg => varg.withAddedFlags(mods.flags | Param))
-          New(Template(emptyConstructor, List(polyFunctionTpt), Nil, EmptyValDef,
-            List(DefDef(nme.apply, applyTParams, List(applyVParams), TypeTree(), res))
-          ))
+            New(Template(emptyConstructor, List(polyFunctionTpt), Nil, EmptyValDef,
+              List(DefDef(nme.apply, applyTParams, List(applyVParams), TypeTree(), res))
+              ))
         }
+      case _ =>
+        EmptyTree //may happen for erroneous input
+    }
+
+    // begin desugar
+
+    // Special case for `Parens` desugaring: unlike all the desugarings below,
+    // its output is not a new tree but an existing one whose position should
+    // be preserved, so we shouldn't call `withPos` on it.
+    tree match {
+      case Parens(t) =>
+        return t
+      case _ =>
+    }
+
+    val desugared = tree match {
+      case PolyFunction(targs, body) =>
+        makePolyFunction(targs, body) orElse tree
       case SymbolLit(str) =>
         Literal(Constant(scala.Symbol(str)))
       case InterpolatedString(id, segments) =>
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1448,7 +1448,9 @@ object desugar {
               ))
         }
       case _ =>
-        EmptyTree // may happen for erroneous input. An error will already have been reported.
+        // may happen for erroneous input. An error will already have been reported.
+        assert(ctx.reporter.errorsReported)
+        EmptyTree
     }
 
     // begin desugar
--- a/compiler/src/dotty/tools/dotc/ast/Desugar.scala
+++ b/compiler/src/dotty/tools/dotc/ast/Desugar.scala
@@ -1430,6 +1430,22 @@ object desugar {
     }
 
     val desugared = tree match {
+      case PolyFunction(targs, body) if (ctx.mode.is(Mode.Type)) =>
+        // Desugar [T_1, ..., T_M] -> (P_1, ..., P_N) => R
+        // Into    scala.PolyFunction { def apply[T_1, ..., T_M](x$1: P_1, ..., x$N: P_N): R }
+        val Function(vargs, resType) = body
+        // TODO: Figure out if we need a `PolyFunctionWithMods` instead.
+        val mods = body match {
+          case body: FunctionWithMods => body.mods
+          case _ => untpd.EmptyModifiers
+        }
+        val applyTParams = targs.asInstanceOf[List[TypeDef]]
+        val applyVParams = vargs.zipWithIndex.map { case (p, n) =>
+          makeSyntheticParameter(n + 1, p).withAddedFlags(mods.flags)
+        }
+        RefinedTypeTree(ref(defn.PolyFunctionType), List(
+          DefDef(nme.apply, applyTParams, List(applyVParams), resType, EmptyTree)
+        ))
       case SymbolLit(str) =>
         Literal(Constant(scala.Symbol(str)))
       case InterpolatedString(id, segments) =>
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -1038,8 +1038,6 @@ class Definitions {
   lazy val PolyFunctionClass = ctx.requiredClass("scala.PolyFunction")
   def PolyFunctionType = PolyFunctionClass.typeRef
 
-  private lazy val TupleTypes: Set[TypeRef] = TupleType.toSet
-
   /** If `cls` is a class in the scala package, its name, otherwise EmptyTypeName */
   def scalaClassName(cls: Symbol)(implicit ctx: Context): TypeName =
     if (cls.isClass && cls.owner == ScalaPackageClass) cls.asClass.name else EmptyTypeName
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -1035,6 +1035,11 @@ class Definitions {
     if (n <= MaxImplementedFunctionArity && (!isContextual || ctx.erasedTypes) && !isErased) ImplementedFunctionType(n)
     else FunctionClass(n, isContextual, isErased).typeRef
 
+  lazy val PolyFunctionClass = ctx.requiredClass("scala.PolyFunction")
+  def PolyFunctionType = PolyFunctionClass.typeRef
+
+  private lazy val TupleTypes: Set[TypeRef] = TupleType.toSet
+
   /** If `cls` is a class in the scala package, its name, otherwise EmptyTypeName */
   def scalaClassName(cls: Symbol)(implicit ctx: Context): TypeName =
     if (cls.isClass && cls.owner == ScalaPackageClass) cls.asClass.name else EmptyTypeName
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -383,6 +383,7 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
    *      - otherwise, if T is a type parameter coming from Java, []Object
    *      - otherwise, Object
    *   - For a term ref p.x, the type <noprefix> # x.
+   *   - For a refined type scala.PolyFunction { def apply[...](x_1, ..., x_N): R }, scala.FunctionN
    *   - For a typeref scala.Any, scala.AnyVal, scala.Singleton, scala.Tuple, or scala.*: : |java.lang.Object|
    *   - For a typeref scala.Unit, |scala.runtime.BoxedUnit|.
    *   - For a typeref scala.FunctionN, where N > MaxImplementedFunctionArity, scala.FunctionXXL
@@ -429,6 +430,12 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       SuperType(this(thistpe), this(supertpe))
     case ExprType(rt) =>
       defn.FunctionType(0)
+    case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
+      assert(refinedInfo.isInstanceOf[PolyType])
+      val res = refinedInfo.resultType
+      val paramss = res.paramNamess
+      assert(paramss.length == 1)
+      this(defn.FunctionType(paramss.head.length, isContextual = res.isImplicitMethod, isErased = res.isErasedMethod))
     case tp: TypeProxy =>
       this(tp.underlying)
     case AndType(tp1, tp2) =>
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -605,6 +605,8 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       case tp: TypeVar =>
         val inst = tp.instanceOpt
         if (inst.exists) sigName(inst) else tpnme.Uninstantiated
+      case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
+        sigName(defn.FunctionType(refinedInfo.resultType.paramNamess.head.length))
       case tp: TypeProxy =>
         sigName(tp.underlying)
       case _: ErrorType | WildcardType | NoType =>
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -196,10 +196,24 @@ object TypeErasure {
           MethodType(Nil, defn.BoxedUnitType)
         else if (sym.isAnonymousFunction && einfo.paramInfos.length > MaxImplementedFunctionArity)
           MethodType(nme.ALLARGS :: Nil, JavaArrayType(defn.ObjectType) :: Nil, einfo.resultType)
+        else if (sym.name == nme.apply && sym.owner.derivesFrom(defn.PolyFunctionClass)) {
+          // The erasure of `apply` in subclasses of PolyFunction has to match
+          // the erasure of FunctionN#apply, since after `ElimPolyFunction` we replace
+          // a `PolyFunction` parent by a `FunctionN` parent.
+          einfo.derivedLambdaType(
+            paramInfos = einfo.paramInfos.map(_ => defn.ObjectType),
+            resType = defn.ObjectType
+          )
+        }
         else
           einfo
       case einfo =>
-        einfo
+        // Erase the parameters of `apply` in subclasses of PolyFunction
+        if (sym.is(TermParam) && sym.owner.name == nme.apply
+            && sym.owner.owner.derivesFrom(defn.PolyFunctionClass))
+          defn.ObjectType
+        else
+          einfo
     }
   }
 
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -605,8 +605,11 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       case tp: TypeVar =>
         val inst = tp.instanceOpt
         if (inst.exists) sigName(inst) else tpnme.Uninstantiated
-      case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
-        sigName(defn.FunctionType(refinedInfo.resultType.paramNamess.head.length))
+      case tp @ RefinedType(parent, nme.apply, _) if parent.typeSymbol eq defn.PolyFunctionClass => 
+        // we need this case rather than falling through to the default
+        // because RefinedTypes <: TypeProxy and it would be caught by
+        // the case immediately below
+        sigName(this(tp))
       case tp: TypeProxy =>
         sigName(tp.underlying)
       case _: ErrorType | WildcardType | NoType =>
--- a/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala
@@ -209,10 +209,13 @@ object TypeErasure {
           einfo
       case einfo =>
         // Erase the parameters of `apply` in subclasses of PolyFunction
+        // Preserve PolyFunction argument types to support PolyFunctions with
+        // PolyFunction arguments
         if (sym.is(TermParam) && sym.owner.name == nme.apply
-            && sym.owner.owner.derivesFrom(defn.PolyFunctionClass))
+            && sym.owner.owner.derivesFrom(defn.PolyFunctionClass)
+            && !(tp <:< defn.PolyFunctionType)) {
           defn.ObjectType
-        else
+        } else
           einfo
     }
   }
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3196,15 +3196,7 @@ object Types {
 
 
     def computeSignature(implicit ctx: Context): Signature = {
-      def polyFunctionSignature(tp: Type): Type = tp match {
-        case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
-          val res = refinedInfo.resultType
-          val paramss = res.paramNamess
-          defn.FunctionType(paramss.head.length)
-        case _ => tp
-      }
-
-      val params = if (isErasedMethod) Nil else paramInfos.mapConserve(polyFunctionSignature)
+      val params = if (isErasedMethod) Nil else paramInfos
       resultSignature.prepend(params, isJavaMethod)
     }
 
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3194,8 +3194,17 @@ object Types {
       companion.eq(ContextualMethodType) ||
       companion.eq(ErasedContextualMethodType)
 
+
     def computeSignature(implicit ctx: Context): Signature = {
-      val params = if (isErasedMethod) Nil else paramInfos
+      def polyFunctionSignature(tp: Type): Type = tp match {
+        case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
+          val res = refinedInfo.resultType
+          val paramss = res.paramNamess
+          defn.FunctionType(paramss.head.length)
+        case _ => tp
+      }
+
+      val params = if (isErasedMethod) Nil else paramInfos.mapConserve(polyFunctionSignature)
       resultSignature.prepend(params, isJavaMethod)
     }
 
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -850,9 +850,13 @@ object Parsers {
      */
     def toplevelTyp(): Tree = rejectWildcardType(typ())
 
-    /** Type        ::=  FunTypeMods FunArgTypes `=>' Type
-     *                |  HkTypeParamClause `=>>' Type
+    /** Type        ::=  FunType
+     *                |  HkTypeParamClause ‘=>>’ Type
+     *                |  MatchType
      *                |  InfixType
+     *  FunType     ::=  { 'erased' | 'given' } (MonoFunType | PolyFunType)
+     *  MonoFunType ::=  FunArgTypes ‘=>’ Type
+     *  PolyFunType ::=  HKTypeParamClause '=>' Type
      *  FunArgTypes ::=  InfixType
      *                |  `(' [ FunArgType {`,' FunArgType } ] `)'
      *                |  '(' TypedFunParam {',' TypedFunParam } ')'
@@ -1234,6 +1238,7 @@ object Parsers {
      *                      |  `throw' Expr
      *                      |  `return' [Expr]
      *                      |  ForExpr
+     *                      |  HkTypeParamClause ‘=>’ Expr
      *                      |  [SimpleExpr `.'] id `=' Expr
      *                      |  SimpleExpr1 ArgumentExprs `=' Expr
      *                      |  Expr2
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -1325,6 +1325,11 @@ object Parsers {
         atSpan(in.skipToken()) { Return(if (isExprIntro) expr() else EmptyTree, EmptyTree) }
       case FOR =>
         forExpr()
+      case LBRACKET =>
+        val start = in.offset
+        val tparams = typeParamClause(ParamOwner.TypeParam)
+        assert(isIdent && in.name.toString == "->", "Expected `->`")
+        atSpan(start, in.skipToken())(PolyFunction(tparams, expr()))
       case _ =>
         if (isIdent(nme.inline) && !in.inModifierPosition() && in.lookaheadIn(canStartExpressionTokens)) {
           val start = in.skipToken()
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -924,9 +924,18 @@ object Parsers {
           val tparams = typeParamClause(ParamOwner.TypeParam)
           if (in.token == TLARROW)
             atSpan(start, in.skipToken())(LambdaTypeTree(tparams, toplevelTyp()))
-          else if (isIdent && in.name.toString == "->")
-            atSpan(start, in.skipToken())(PolyFunction(tparams, toplevelTyp()))
-          else { accept(TLARROW); typ() }
+          else if (in.token == ARROW) {
+            val arrowOffset = in.skipToken()
+            val body = toplevelTyp()
+            atSpan(start, arrowOffset) {
+              body match {
+                case _: Function => PolyFunction(tparams, body)
+                case _ =>
+                  syntaxError("Implementation restriction: polymorphic function types must have a value parameter", arrowOffset)
+                  Ident(nme.ERROR.toTypeName)
+              }
+            }
+          } else { accept(TLARROW); typ() }
         }
         else infixType()
 
@@ -1328,8 +1337,16 @@ object Parsers {
       case LBRACKET =>
         val start = in.offset
         val tparams = typeParamClause(ParamOwner.TypeParam)
-        assert(isIdent && in.name.toString == "->", "Expected `->`")
-        atSpan(start, in.skipToken())(PolyFunction(tparams, expr()))
+        val arrowOffset = accept(ARROW)
+        val body = expr()
+        atSpan(start, arrowOffset) {
+          body match {
+            case _: Function => PolyFunction(tparams, body)
+            case _ =>
+              syntaxError("Implementation restriction: polymorphic function literals must have a value parameter", arrowOffset)
+              errorTermTree
+          }
+        }
       case _ =>
         if (isIdent(nme.inline) && !in.inModifierPosition() && in.lookaheadIn(canStartExpressionTokens)) {
           val start = in.skipToken()
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -924,6 +924,8 @@ object Parsers {
           val tparams = typeParamClause(ParamOwner.TypeParam)
           if (in.token == TLARROW)
             atSpan(start, in.skipToken())(LambdaTypeTree(tparams, toplevelTyp()))
+          else if (isIdent && in.name.toString == "->")
+            atSpan(start, in.skipToken())(PolyFunction(tparams, toplevelTyp()))
           else { accept(TLARROW); typ() }
         }
         else infixType()
--- a/compiler/src/dotty/tools/dotc/transform/ElimErasedValueType.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ElimErasedValueType.scala
@@ -83,14 +83,19 @@ class ElimErasedValueType extends MiniPhase with InfoTransformer { thisPhase =>
       override def matches(sym1: Symbol, sym2: Symbol) =
         sym1.signature == sym2.signature
     }
+
     def checkNoConflict(sym1: Symbol, sym2: Symbol, info: Type)(implicit ctx: Context): Unit = {
       val site = root.thisType
       val info1 = site.memberInfo(sym1)
       val info2 = site.memberInfo(sym2)
-      if (!info1.matchesLoosely(info2) &&
-          !(sym1.name == nme.apply &&
-            (sym1.owner.derivesFrom(defn.PolyFunctionClass) ||
-             sym2.owner.derivesFrom(defn.PolyFunctionClass))))
+      // PolyFunction apply methods will be eliminated later during
+      // ElimPolyFunction, so we let them pass here.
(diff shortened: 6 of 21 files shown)


smarter:
@SystemFw has a pretty compelling usecase he presents in https://gist.github.com/SystemFw/256205f51bf0135e4c6fd95dee4590fc, I've verified that it can be implemented with this PR: https://gist.github.com/smarter/9a28e39dfaf0235e9d0a16b09ff8a9a4 (would be nice to minimize this to a zero-dependency file that we can add to our test suite).

Glavo:
Why use `->` instead of `=>`? I think that `->` is just an ordinary identifier and is not suitable for use here.

smarter:
Yeah I'm not sure about the syntax, this is just a placeholder that works. Currently we can't use `=>` because it conflicts with type lambdas (like `[X] => List[X]`).

LukaJCB:
Another use case is for abstracting over tagless final programs and interpreters.
As soon as you try doing that currently, you’ll realize it’s impossible to get the type for a tagless final program without wrapping.
With implicit function types and rank-N-types we can now do it like this:

```scala
type Program[Alg[_[_]], A] = [F[_]: Applicative] -> Alg[F] => F[A]
```

This is super useful for e.g. optimizing tagless final programs. It means that now I can write this general function:

```scala
def optimize[Alg[_[_]], F[_]: Applicative, A, M: Monoid]
  (program: Program[Alg, A])
  (extract: Alg[Const[M, ?]])
  (rebuild: (M, Alg[F]) =>  F[Alg[F]]): Alg[F] => F[A] = { (interpreter: Alg[F]) =>
    val m: M = p(extract).getConst

    rebuild(m, interpreter).flatMap(newInterp => program(newInterp))
}
```

And that is basically what I did in [sphynx](https://github.com/LukaJCB/sphynx), though there I still have to wrap every program in an extra wrapper. So this would really help immensely in that regard!

LPTK:
@smarter would you agree that using `->` for type functions and leaving `=>` for value functions (for both type and value parameters) would be more homogeneous?
I've always found the `[T] => ...` type lambda syntax rather confusing, as to me `=>` is the visual marker of normal function types.

smarter:
@LPTK we're still hoping to get some sort of effect system in Dotty, and we were planning to use `A -> B` to mean "pure function" (whereas `A => B` would be retrofitted to mean "function with an arbitrary effect". Maybe we can find another syntax but we're running out of arrows.

LPTK:
@smarter I see.

BTW, why not use type lambdas to represent polymorphic function types (or any polymorphic type really)? So we'd allow values to have higher-kinded types. Even if that wouldn't work with the current compiler's infrastructure, does the user need to know they are different? What would happen if we sued the same syntax for both, letting the compiler distinguish merely on the context where the type appears?

smarter:
I think using the same syntax is likely to confuse users, it's also not clear that we can always distinguait between them based on the context. We could maybe look at the kind of the expected type of we have one, but what if there's no expected type? If we enable kind polymorphism in the compiler things would get even more muddled.

LPTK:
> using the same syntax is likely to confuse users

Are you sure about that? After all, aren't they conceptually the same? – and if so, I think having several syntaxes for the same thing would be the more confusing approach.

At least in dependently-typed systems, there is no distinction. So _in principle_ they could indeed be unified. Now, the question is: would that really work for Scala?

About giving values higher-kinded types, according to you is there a fundamental difference between these two?

```scala
trait A0 { val a: [T] -> a[T] ; type a[T] }
// ^ using the syntax of your PR

trait A1 { val a: a ; type a[T] }
// ^ using the unified approach
```

smarter:
The main difference is in that in this PR, the type `[T] -> A => B` is a subtype of `Any`, whereas with real polymorphic values it would be a subtype of `[T] -> Any`. This means that abstracting over these things would require [kind-polymorphism](https://github.com/lampepfl/dotty/pull/4108), it's a much bigger and conceptual leap, and likely to be far-reaching.

milessabin:
@smarter seeing as kind polymorphism _is_ on the table, I think we should be seeing if we can take advantage of it here.

smarter:
> Will types like [A] -> A be supported?

To be decided, `[A] -> () => A` is already supported and makes it clearer that there's a method call involved. On the other hand it's a bit verbose.

> seeing as kind polymorphism is on the table, I think we should be seeing if we can take advantage of it here.

Can you think of examples that shows things we could only do, or do better, by taking advantage of kind polymorphism here?

milessabin:
> Can you think of examples that shows things we could only do, or do better, by taking advantage of kind polymorphism here?

@smarter is was responding to your comment,

> The main difference is in that in this PR, the type `[T] -> A => B` is a subtype of `Any`, whereas with real polymorphic values it would be a subtype of `[T] => Any`. This means that abstracting over these things would require kind polymorphism, it's a much bigger conceptual leap, and likely to have far-reaching consequences.

smarter:
@milessabin Yes, I got that. I mean that if we were to go down that road, it would have to be because it presents significant practical advantages, not just because it's conceptually elegant, since it's likely to be extremely complex to implement and specify.

Blaisorblade:
> BTW, why not use type lambdas to represent polymorphic function types (or any polymorphic types really)? So we'd allow values to have higher-kinded types.

> Are you sure about that? After all, aren't they conceptually the same? – and if so, I think having several syntaxes for the same thing would be the more confusing approach.

No they aren't, even tho the difference is subtle (I've been confused for years), but they're clearly distinct things in Haskell Agda Coq Idris... The only system merging them doesn't look compatible with kind polymorphism — not because of a soundness hole, but because it's not clear at all what `List` would mean there. And what about `List => List`? Is that `([A] -> List[A]) => ([A] -> List[A])` or `[A] -> List[A] => List[A]`?

This is not an impossibility result, but absent some extraordinarily compelling argument, I strongly believe we should follow ~50 years of PL theory instead of doing something different here.

So, the correct question is "please find a pretty compelling reason for merging these very different things". And I encourage you to find fault with the existing distinction in languages that have it. In a separate issue, please.

==

Polymorphic function types (which I'll write for now with the existing MethodType syntax `[A]List[A]`) and type lambdas ([A] => List[A]) are two very different things in theory and practice. The only thing in common is that they bind a type variable `A` in some type `T`. But one is a value type (a type of functions), the other only *maps* value types to value types, that is, it becomes a value type after you apply it to some argument.

There's basically *one* [pretty hairy paper](http://www.macs.hw.ac.uk/~fairouz/forest/papers/journals-publications/jfp05.pdf) that uses the same type `[A]List[A]` for both things, but then, each time you use such a type, you must check if you use it at kind `*` or `* -> *` to know if you actually meant "polymorphic function type" or "type lambda". Since kind polymorphism lets you upcast `[A]List[A]` to kind `AnyKind` forgetting that bit of info, I'd say that identifying polymorphic functions and type functions is *incompatible* with kind polymorphism.

> The main difference is in that in this PR, the type [T] -> A => B is a subtype of Any, whereas with real polymorphic values it would be a subtype of [T] => Any

IIUC, the subtypes of `[T] => Any` are type functions, not "real polymorphic values" — I'd love to be corrected if I'm missing something here. And in fact, arguably `List` would be a subtype of *both* `Any` and `[T] => Any` if it can be used in both roles, which doesn't look any pleasant.

smarter:
> No they aren't, even tho the difference is subtle (I've been confused for years)

Same here, so you should disregard everything I said and listen to @Blaisorblade :)

LPTK:
@Blaisorblade thanks for the answer. You're right, I think I was confused too.

When looking at [the JFP'05 paper](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/418409138B4452969AC0736DB0A2C238/S095679681300018Xa.pdf/idris_a_generalpurpose_dependently_typed_programming_language_design_and_implementation.pdf) on Idris, in the syntax definition on page 560, we can see there is indeed a syntax for _functions_ (lambda abstractions such as `\x => x+1` and `\tp => List tp`) and _function spaces_ for writing the types of such expressions (respectively `Int -> Int` and `Type -> Type`). In Scala's type language, the first corresponds to `[A] => T` and the second to `(x: A) => T` (i.e., just `A => T` if `x` is not free in `T`).

Now, the `[A] -> T` form introduced in this PR can be understood as `(A : *) => T` in a hypothetical syntax where `*` is the type of types; so it's an instance of the second form, _not_ the first –– it's really not a (type) lambda.

Perhaps a more Scala-ish way of writing `(A : *) => T` would be `(type A) => T` or `[type A] => T`.

alexknvl:
Why not 
```scala
type Foo = [T <: AnyVal] List[T] => List[(T, T)]
val f = [T <: AnyVal] (x: List[T]) => x.map(e => (e, e))
foo([T] (x: T) => Some(x))
// or
foo(Some)
```
`->` looks confusing when mixed with `=>` (precedence?) IMHO

Blaisorblade:
@alexknvl Concrete syntax-wise, `[A] T` would be consistent with MethodTypes as pretty-printed today by Scala 2 and 3 — though that might be confusing if MethodTypes are still around as a distinct thing (maybe we could then change *their* syntax but that affects current error messages).

> // Polymorphic *values* would be AMAZING:

Yes but they'd turn the design into *two* research problems. I'd love polymorphic functions, and probably the way to have them is to *not* make them depend on research-level problems (a bit like the SI-2712).

1. *Pure* polymorphic values would be cool, but "impure" ones are unsound (like polymorphic references in ML), how do you exclude them?
-> Unless we managed to adapt ML's value restriction and use it for *all* polymorphic values, which I'm not sure we have considered — but that would interfere with nullary functions meant to be functions...
```scala
class Box[T] { var x: List[T] = Nil }
def foo[X]: Box[X] = new Box[X] // a normal polymorphic method
val fooVal1: [X]Box[X] = foo // valid polymorphic value? what are the semantics? probably should be rejected somehow
val fooVal2: [X]Box[X] = new Box[X] // should also be rejected
def higherOrder[M[_]: Monad] = {
  val hoFooVal: [X]M[List[X]] = Nil.pure // should the compiler accept this? does `M[X]` contain say some `var bar: X`? How do we tell, should we refine M's kind to track mutability via some effect system?
}
```

(See https://github.com/lampepfl/dotty/issues/2500#issuecomment-330977833, one of the reasons that lead to #4670). So, that seems to involve an effect system, for which we don't have a good design yet.

2. A "natural" design for polymorphic values would arguably have yet other properties (such as `forall X. F[X] <: F[T]` for any `T`, proposed by @TomasMikula and also discussed in #2500), but that's even more invasive in terms of required changes to the metatheory.

smarter:
> // With different syntax it is readable

This is tricky to get right. I'd like to add `[T: Foo] -> A => B` as syntactic sugar for `[T] -> implicit Foo[T] => A => B`, but that won't help in the [fmap example](https://gist.github.com/smarter/e9c67df0dd089019b823fcfde91ec0e8), because `[F[_]: Functor] -> F[A] => F[B]` will be expanded to:
```scala
[F[_]] -> implicit Functor[F] => F[A] => F[B]
```
instead of:
```scala
[F[_]] -> F[A] => implicit Functor[F] => F[B]
```
which means the implicit search will be done _before_ the type variable `F` has been constrained and won't succeed.

We could change the desugaring to be "add the implicit parameter just before the last occurence of =>" but that's not very satisfying since it means that in `[T: Foo] -> A => B` we will add the implicit parameter just before `B`, even if `B` is a type alias. Alternatively, the desugaring could use type information to decide where to put the implicit but that's likely to be extremely confusing for readers of the code.

smarter:
> which means the implicit search will be done before the type variable F has been constrained and won't succeed.

Alternatively, it may or may not be possible to delay the implicit search, I haven't tried to work out what that would entail.

smarter:
Actually no, we can't delay the implicit search even if none of the type variables it references have been constrained, because sometimes implicit search is legitimately used to infer type variables (as in https://milessabin.com/blog/2011/07/16/fundeps-in-scala/)

LPTK:
@smarter at this point, why not just add a syntax much closer to method types, as @Blaisorblade proposed? We already have:
```scala
def foo(x: Int) => t
foo: (x: Int) => T
```
So why not just go the extra mile and allow the same syntax/syntax sugar as in method definitions with respect to type parameters, currying and implicits? Let `[A,B:T](a: A)(b: B) => C` mean the same as the current `[A] -> [B] -> (a: A) => (b: B) => implicit T[B] => C`, which is arguably much less clear (as remarked by @alexander-myltsev).

This would also be a good time for changing the way method types are pretty-printed, as they are currently pretty inscrutable IMHO. `(a: A)B` does not look like a function-like type to me, and it got me confused when I was seeing this in compiler output as a beginner. It's also pretty hard to read when the `A` and `B` types get larger. I think `(a: A) => B` is much better, especially since it's a type users can write themselves.

smarter:
Completely revamping the syntactic sugar for function types is out of scope for this PR, this should be a separate discussion. (In general I wonder if we shouldn't always have syntax discussions related to a proposal in a separate thread from the main discussion, since it tends to expand until there's no breathing room for anything else).

Jasper-M:
> // Polymorphic *values* would be AMAZING:
val f: [T] List[T] = Nil

`val f: [T] => List[T] = Nil` should be possible though. It's the same thing as `def f[T]: List[T] = Nil`. It's also a point in favor of the `->`-less syntax, because `val f: [T] -> => List[T] = Nil` doesn't work very well. 
Also, apologies for bringing up syntax again 😅

Blaisorblade:
@Jasper-M You mean `val f: [T] => List[T] = Nil` should be basically a `Function0` which runs the body at each invocation? Yes that's fine.

compiler/src/dotty/tools/dotc/transform/ElimPolyFunction.scala
```diff
@@ -0,0 +1,68 @@
...
+
+/** This phase rewrite PolyFunction subclasses to FunctionN subclasses
+ *
+ *      class Foo extends PolyFunction {
+ *          def apply(x_1: P_1, ..., x_N: P_N): R = rhs
+ *      }
+ *   becomes:
+ *      class Foo extends FunctionN {
+ *          def apply(x_1: P_1, ..., x_N: P_N): R = rhs
+ *      }
```
Blaisorblade: I wonder if we should allow writing such a class `Foo`, rather than just allowing closures — we don't for implicit function types and nobody seems to mind the restriction, and this restriction enables transformations such as `ShortcutImplicits`. https://github.com/lampepfl/dotty/pull/1775/files#diff-71350811180f41d868e7fb3258fd774dR18

Jasper-M:
@Blaisorblade Yes it would. Though this is actually by-name syntax which is currently only available on method parameters. Making that generally available is probably a separate issue, and possibly more complicated than meets the eye. But we should probably at least allow polymorphic by-name parameters.

compiler/src/dotty/tools/dotc/transform/ElimPolyFunction.scala
LPTK: Note that it can be really useful being able to extend/implement a polymorphic function type; it's one of the use cases I mention in https://github.com/lampepfl/dotty/issues/4670#issuecomment-397819801 – making polymorphic type classes extend the corresponding polymorphic function type.

Blaisorblade: Ah right, tho `s/type classes/case classes/`. But if we follow the approach for implicit function types, `val b: Int => B[Int] = B` could still work by eta-expansion. In fact, it's not clear why eta-expansion doesn't handle that case today by turning `B` into `B _` or `B.apply` (and I'm not going to try which ones do work, I'd just ask they all work unless backward compatibility gets in the way).

Blaisorblade:
@Jasper-M Not quite, as discussed above it's essential to have `[X] List[X]` or `[X] -> List[X]` or however we write it mean a `Function0`, lest we need sound polymorphic values.

Jasper-M:
@Blaisorblade by-name values are essentially `Function0` with some syntactic sugar, so I'm not sure that's an issue here.

milessabin:
> Completely revamping the syntactic sugar for function types is out of scope for this PR, this should be a separate discussion.

Well, yeah, Wadler's Law etc.

However, the essence of this PR is aligning method and function types, and I think that does immediately raise the syntax question. If it turns out that there are methods which can't be easily mapped to function types syntactically, then that should raise alarm bells that maybe the semantics haven't been fully aligned either.

compiler/src/dotty/tools/dotc/transform/ElimPolyFunction.scala
LPTK: @Blaisorblade AFAIK eta expansion never inserts `.apply` calls on things that are not functions. It's true that adding this behavior could be an alternative solution to the stated polymorphic case class problem.

ivanopagano:
Since no syntax-specific issue is available yet, what would you say of ?
```
type Foo[T <: AnyVal] = List[T] => List[(T, T)]
val f[T <: AnyVal] = (x: List[T]) => x.map(e => (e, e))
foo((x: _) => Some(x))
// or
foo(Some)

// Polymorphic *values* would be AMAZING:
val f[T]: List[T] = Nil
def runST[A](st[S]: ST[S, A]): A
```

Just throwing new ideas in the pool

odersky:
I have written up my thoughts regarding syntax in #4712.

jdegoes:
@odersky This would be hugely useful to the functional programming community, which currently has implemented various plug-ins and hacks to workaround the verbosity of encoding polymorphic functions in Scala.

Indeed, I'd say the current pain is so high, functional programmers have a tendency to shy away from polymorphic functions even in cases where they would greatly improve the code, and if not for the boilerplate, make the code much more concise and readable.

I think inference would be extremely important to mainstream success of the feature.

Please consider rank n types and polymorphic values. If you are interested in examples, I can show you how these features could greatly simplify current Scala.

smarter:
> I think inference would be extremely important to mainstream success of the feature.

Agreed. It's on my list of things to do ;).

> Please consider rank n types and polymorphic values.

Polymorphic values that are not functions are out of scope here, see https://github.com/lampepfl/dotty/pull/4672#issuecomment-398950818

> If you are interested in examples, I can show you how these features could greatly simplify current Scala.

Yes please, more examples would be great, especially if they're self-contained and can be added to our test suite.

milessabin:
Rebased and added some missing cases.

compiler/src/dotty/tools/dotc/transform/Mixin.scala
```diff
@@ -169,8 +169,10 @@ class Mixin extends MiniPhase with SymTransformer { thisPhase =>
       case Block(stats, expr) =>
         val (scall, inits) = transformConstructor(expr)
         (cpy.Block(tree)(stats, scall), inits)
-      case _ =>
-        val Apply(sel @ Select(New(_), nme.CONSTRUCTOR), args) = tree
+      case Apply(sel @ Select(New(_), nme.CONSTRUCTOR), args) =>
+        val (callArgs, initArgs) = if (tree.symbol.owner.is(Trait)) (Nil, args) else (args, Nil)
+        (superRef(tree.symbol, tree.pos).appliedToArgs(callArgs), initArgs)
+      case Apply(sel @ Select(TypeApply(Select(New(_), aio), List(tpt)), nme.CONSTRUCTOR), args) if aio == defn.Any_asInstanceOf.name =>
```
smarter: How is this change related to this PR ?

milessabin: Good question. If you remove that case then,
```
testOnly dotty.tools.dotc.CompilationTests -- *tastyBootstrap
```
blows up with a `MatchError` there, the first failing file being `library/src-bootstrapped/scala/Tuple.scala` when compiling `NonEmptyTuple`. There's clearly an interaction with inlining here, but I didn't have time to get to the bottom of it.

milessabin:
Squashed and rebased.

milessabin:
Rebased again ... finger's X's this will sort out the community build failure.

milessabin:
Reader: it did ...

milessabin:
@smarter @odersky I've rebased this umpteen times now ... I'd really like to get this merged so that I don't have to keep doing it.

Could we put it behind a `-Y` flag for now? Failing that what needs to be done to get this into a merge worthy state?

smarter:
I'll have a look later this week.

milessabin:
@smarter feel free to assign any gruntwork to me :-)

milessabin:
Rebased again.

milessabin:
And again.

milessabin:
Rebased.

milessabin:
And again ...

milessabin:
Rerebased.

jdegoes:
@smarter @odersky I hereby volunteer to review this pull request if no one else has time. 😃

smarter:
@jdegoes  It's more that we need to add some missing stuff (type inference) and think through whether this is the best encoding. In particular since last week I've been investigating whether the encoding used here (magic refinement on a marker trait) could be generalized to encode all types of functions (regular, implicit, erased, dependent, ...), thereby making polymorphic functions almost trivial to implement (modulo figuring out type inference). So far it seems promising but there's a bunch of details to figure out, so I may try to prioritize getting polymorphic functions as implemented in this PR merged first since they're already providing values. But I'm not going to give an ETA for that yet because I've already missed enough self-imposed deadlines :).

jdegoes:
@smarter All right, that makes sense. Thanks for the detailed update. I do think unifying the representation of all functions could simplify a lot of code...

smarter:
Our current plan is to get this merged relatively soon to make experimentations easier, with the understanding that the syntax/semantics/implementation details are all likely to change widely in the near future.

odersky:
We need to come to an overall decision on the syntax of type lambdas before merging this. I don't think `->` for PFTs and `=>` for type lambdas is the right choice.

 - right now, they make type lambdas look like normal functions, and PFTs like something else. It should be the reverse. PFTs are like normal function types; type lambdas are on a different level.

 - we want to reserve `->` to mean "pure function" in the future. This means we will need `[X] -> Y` for pure polymorphic functions and `[X] => Y` for impure ones. This also means we can't use `=>` for type lambdas. If we use `->` for polymorphic functions now that will imply they are pure.

Last year when we discussed this we settled on

 - `[X] => Y`    for (impure) polymorphic functions
 - `[X] -> Y`     for pure polymorphic functions
 - `[X] #> Y`    for type lambdas

 Barring a better solution, I think we should go ahead with this. This means we need first a PR that changes the syntax of type lambdas. Everybody agreed with this? /cc @milessabin

sjrd:
At some point there was the following proposal:

* `[X](...params) => Y` for PFTs.
* `[X] => Y` for type lambdas.

Note that a PFT always requires a term parameter list, the same way that all function types require a term parameter list (the is no function type corresponding to `()`-less methods).

Was that idea abandoned?

odersky:
> Note that a PFT always requires a term parameter list, the same way that all function types require a term parameter list (the is no function type corresponding to ()-less methods).

If that is true that would be workable. But is it?

sjrd:
It seems so. But this PR has virtually no test, so I can't tell...

odersky:
The parser seems to accept arbitrary types after the `->` not just function types. From the diffs:
```
          else if (isIdent && in.name.toString == "->")
            atSpan(start, in.skipToken())(PolyFunction(tparams, toplevelTyp()))
```

sjrd:
Yes, but at the same time, `Desugar` clearly assumes that the `body` of a `PolyFunction` is a `Function`,  so ...

milessabin:
@odersky I like the proposal that @sjrd is reminding us of.

LPTK:
I think @sjrd's proposal makes the language look inconsistent.

We can currently write `=> A` to express the type of a by-name function (function with no parameter lists). So it's natural to expect that if the type of polymorphic function follows the pattern `[T]‹parameter-lists› => C`, we could also write `[T] => A` to express the type of polymorphic by-name functions (i.e., the type of `{def apply[T]: A}`).

Type lambdas are a very different thing than type functions, so they should get their own sufficiently-differentiated syntax.
For having used type lambdas in their current form (including in [some quite complex examples](https://contributors.scala-lang.org/t/proposal-for-explicit-syntax-for-type-lambdas/2949/8?u=lptk)), I find that **they blend with normal function types way too easily**, making the code hard to decipher.

That said, I find `#>` very, very weird-looking.

I still think that sticking the `type` keyword in the syntax of type lambdas could be a good way of easily distinguishing type lambdas. As in, `Functor[type A -> List[A]]` meaning `Functor[List]`.

milessabin:
`=> A` isn't the type of a by-name function, it's a _pseudo type_ for a by-name method argument of type `A` and can't be used in any other context.

I agree that `#>` is odd looking and for me that favours @sjrd's proposal.

A syntax like `type A -> ...` doesn't support multiple type parameters smoothly.

Jasper-M:
At this very moment type lambda's are pretty printed by scalac as `[x]Either[String, x]`. Would that conflict with other syntax?

nafg:
#> is used pretty heavily in Lift for its CSS selector DSL (the way you
populate data into templates).

nafg:
Oh if it's only for type lambdas it wouldn't conflict. It also is reminiscent of # (type projection) which is sort of also a way to go from one type to another.

You can add a character in the middle to get more options, like =#>, =:>, etc. Maybe PartialFunction could get =?>...

odersky:
> Yes, but at the same time, Desugar clearly assumes that the body of a PolyFunction is a Function, so ...

      case PolyFunction(targs, body) =>
        val Function(vargs, res) = body
        ...

It seems we need to do quite a bit of hardening before we can merge this! @smarter, @milessabin: So what's the deal? Is `[X] -> X` a legal type of the function resulting from eta expanding
```
  def f[X]: X = implicitly[X]
```
?

odersky:
> At this very moment type lambdas are pretty printed by scalac as [x]Either[String, x]. Would that conflict with other syntax?

No that would work as well. But last time we discussed this, people objected.

smarter:
>  Is [X] -> X a legal type of the function resulting from eta expanding

It's not supposed to be legal no, it would effectively allow "polymorphic by-names" which I don't think are a great idea. Indeed, the parser needs to be more strict, I just didn't implement anything more than the minimal changes I needed until the details were settled.

odersky:
> I still think that sticking the type keyword in the syntax of type lambdas could be a good way of easily distinguishing type lambdas

How do you handle bounds in this case?

odersky:
> It's not supposed to be legal no, it would effectively allow "polymorphic by-names" which I don't think are a great idea. 

Why not? Note that you could not write something like 
```
  [X] -> PartialFunction[F[X], X]
```
with this restriction.

LPTK:
@milessabin 
> A syntax like `type A -> ...` doesn't support multiple type parameters smoothly.

@odersky 
> How do you handle bounds in this case?

A better syntax may be `type [‹type-params›] -> ‹result›`, which naturally allows for several parameters and bounds.

```scala
type ListF = Functor[type [A] -> List[A]]
// i.e.,
type ListF = Functor[List]

type Pair = type [A] -> (A,A)
// i.e.,
type Pair[A] = (A,A)
```

This way, `[A] -> e` is a lambda that takes a type parameter, while `type [A] -> T` is a type lambda.

The `type` prefix to the parameter, which modifies the parameter's role, is akin to the `given` prefix in implicit function types.

milessabin:
I think that @sjrd's observation that there is no monomorphic function type corresponding to by-name values is critical here.

We have a strategy for dealing with that in the monomorphic case: convert the term to a function by adding `() =>` on the LHS, so by-name `T` becomes `() => T`.

We should do exactly the same in the polymorphic case, eg. for @odersky's example and @sjrd's proposed syntax we would write `[T] => () => PartialFunction[F[T], T]`.

It might make sense for us to address lazy/by-name values in general, both poly- and monomorphic, but I think that's orthogonal to this issue.

LPTK:
@milessabin 
> It might make sense for us to address lazy/by-name values in general, both poly- and monomorphic, but I think that's orthogonal to this issue.

It's not orthogonal with the choice of syntax: if we go by @sjrd's proposed syntax, we lose the future ability to naturally express the type of polymorphic by-name values!

In any case, making `[A] => T` denote a type lambda seems inconsistent and confusing (looks like a function type too much, as I explained above).

> there is no monomorphic function type corresponding to by-name values

I'd argue that `=> A` _is_ the type of monomorphic by-name values (whether we call it a function type or not is just a question of terminology at this point).
Really, what IMHO _is_ orthogonal to the issue at hand is the fact that by-name types are not first-class/not valid in all contexts. We could very well one day remove this limitation, or we could add this same limitation to polymorphic by-name values, or we could even live with different rules for these two types of values. So what?


> `=> A` isn't the type of a by-name function, it's a _pseudo type_ for a by-name method argument of type `A` and can't be used in any other context.

If we're going to be pedantic, this is not even true. By-names can live in positions that do not correspond to method arguments, such as in the type `(=> A) => B`.

milessabin:
> We could very well one day remove this limitation.

We could. I suggest we do it by introducing type constructors to represent evaluation order. So instead of `=> T` (which we can keep as syntactic sugar) we have a first class type `ByName[T]`. We can also have `Lazy[T]` (cp. Idris and shapeless) potentially giving us a quick fix for https://github.com/lampepfl/dotty/issues/3005 without the need for any new syntax. It also gives us the possibility of exploring [evaluation-order polymorphism](https://arxiv.org/pdf/1504.07680.pdf) further down the road.

This is all completely consistent with the syntax @sjrd proposed.

Blaisorblade:
> > We could very well one day remove this limitation.
> 
> We could. I suggest we do it by introducing type constructors to represent evaluation order. So instead of `=> T` (which we can keep as syntactic sugar) we have a first class type `ByName[T]`.

As long as we do that consistently everywhere, yes — but supporting `=> T` as sugar everywhere (for compatibility) and not here looks somewhat inconsistent.

> Note that a PFT always requires a term parameter list, the same way that all function types require a term parameter list (the is no function type corresponding to `()`-less methods).

That seems accidental — since we forbid polymorphic values to retain soundness, the only types that can be made polymorphic should not be function types but method types (which are *not* value types), and nullary method types exist (https://www.scala-lang.org/files/archive/spec/2.13/03-types.html#method-types), tho the spec treats them as separate from by-name param types. Then, the consistent choice would be to support polymorphic by-names. Is there some reason to be inconsistent?

odersky:
> We could. I suggest we do it by introducing type constructors to represent evaluation order. So instead of => T (which we can keep as syntactic sugar) we have a first class type ByName[T]. We can also have Lazy[T] (cp. Idris and shapeless) potentially giving us a quick fix for #3005 without the need for any new syntax. It also gives us the possibility of exploring evaluation-order polymorphism further down the road.

> This is all completely consistent with the syntax @sjrd proposed.

I claim it isn't. @sjrd's proposal is that a type parameter list _must_ be followed by `(...)`. So, you could not write `[T] -> ByName[T]`, it _must be_ `[T]() -> ByName[T]`, which defeats the purpose.

odersky:
> We should do exactly the same in the polymorphic case, eg. for @odersky's example and @sjrd's proposed syntax we would write [T] => () => PartialFunction[F[T], T].

I think it would be `[T]() => PartialFunction[F[T], T]`.

milessabin:
> So, you could not write [T] -> ByName[T], it must be [T]() -> ByName[T], which defeats the purpose.

Yes, my mistake on the syntax.

But no, this _doesn't_ defeat the point. Using a type constructor for by-name we might have something like, `[T] ByName[T] => T`. I'm suggesting the corresponding method would be `def foo[T](t: ByName[T]): T` which would be equivalent to the existing `def foo[T](t: => T): T` ... which is pretty much how things already work under the hood in both Dotty and Scala 2.

odersky:
Somehow we seem to be losing `()`'s. Surely it would be 
```
[T] () => ByName[T] => T
```
not 
```
[T] ByName[T] => T
```
(under the @sjrd proposal, I mean).

milessabin:
> Somehow we seem to be losing ()'s.

I think we're talking at cross purposes ... my fault.

If we're talking about encoding by-name polymorphic values, the syntax under @sjrd's proposal would be `[T] () => ByName[T]` encoding a polymorphic `T`. I agree that _isn't_ any better than `[T] () => T`.

OTOH, a first class equivalent of `def foo[T](t: => T): T` would be `[T] ByName[T] => T`.

odersky:
I think the most important issue in this discussion is not the syntax of polymorphic function types, but the syntax of type lambdas. @milessabin I must admit I got confused several times when I saw your code in your typeclass derivation strawman, which used definitions like this:
```
type MirroredType  =  [t] => Opt[t]
```
Knowing that your code was based on the current branch, I read this as a polymorphic function instead of a type lambda. Part of the reason was probably that I am used to higher-kinded type parameters on the left
```
type MirroredType[t] = Opt[t]
```
so did not "click" immediately that this was the same definition. And I say this as the person who designed and implemented the type lambda feature in its current form... This means there's a real risk that other people will get confused as well.

To play it safe, we should investigate whether we can find an alternate syntax of type lambdas with less potential for confusion. The original proposal was
```
[X] #> Opt[X]
```
which is admittedly ugly. @LPTK proposes
```
type [X] => Opt[X]
```
That risks inflationary use of `type` though. I.e. the original `MirroredType` definition would read like this:
```
type MirroredType = type [t] => Opt[t]
```
Does not look so appealing either, IMO. Another possibility would be to repeat the `>`, i.e.
```
[X] =>> Opt[X]
```
That one looks like an arrow, and the double `>` could indicate that we "lift" the function to the type level. 

I am not sure how likely this operator is to conflict with existing usage. If existing usage of `=>>` is widespread we could make `=>>` a soft keyword, that applies only when preceded by `[...]`.

Note that is we would take this decision to change the syntax for type lambdas, the syntax for polymorphic function types is obvious. It should be `[X] => Y`, mirroring normal and dependent function types.

ivanopagano:
> 
> To play it safe, we should investigate whether we can find an alternate syntax of type lambdas with less potential for confusion. The original proposal was
> 
> ```

Crazy brain at work for some creative proposals just for you:
```scala
type MirroredType = =[t]> Opt[t]

type MirroredType = =[t]=> Opt[t]

type MirroredType = >[t]> Opt[t]

type MirroredType = [t]> Opt[t]
```

milessabin:
> Knowing that your code was based on the current branch, I read this as a polymorphic function instead of a type lambda.

Fair enough.

I don't think that `=>>` is actually all that widely used, so I would be in favour of using that for type lambdas and `=>` for poly functions.

jdegoes:
`forAll [A]. A => A`

If it ain't broke, don't fix it. 😄 

Also suggests a possible path to existential types via the recently deprecated `forSome` (e.g. `forSome X. X => X`).

If Scala supported type blocks, where the type of the block is the final type in the block (similar to value blocks, where the value of the block is the final value in the block), you wouldn't need any new syntax, although it would be quite awkward and require permitting fully unapplied type aliases:

```scala
{ type Id[A] = A => A; Id }
```

odersky:
> forAll [A]. A => A

I don't want to create a new syntactic role for `.`. If we replace it with `=>` we get `forAll [A] => A => A`
and at that point we might as well drop the `forall`.

LPTK:
@jdegoes The problem is that in academia and academically-inspired languages, `forall` is generally used for polymorphic types, not for type lambdas. See Haskell for example, or System F.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -1322,6 +1324,11 @@ object Parsers {
         atSpan(in.skipToken()) { Return(if (isExprIntro) expr() else EmptyTree, EmptyTree) }
       case FOR =>
         forExpr()
+      case LBRACKET =>
+        val start = in.offset
+        val tparams = typeParamClause(ParamOwner.TypeParam)
+        assert(isIdent && in.name.toString == "->", "Expected `->`")
```
odersky: This should be checked syntax, not an assert.

compiler/src/dotty/tools/dotc/transform/ElimErasedValueType.scala
```diff
@@ -87,7 +87,10 @@ class ElimErasedValueType extends MiniPhase with InfoTransformer { thisPhase =>
       val site = root.thisType
       val info1 = site.memberInfo(sym1)
       val info2 = site.memberInfo(sym2)
-      if (!info1.matchesLoosely(info2))
+      if (!info1.matchesLoosely(info2) &&
+          !(sym1.name == nme.apply &&
```
odersky: I did not understand why this additional condition was needed.To make it clearer, define it as an auxiliary def and add a doc comment?

compiler/src/dotty/tools/dotc/transform/Erasure.scala
```diff
@@ -416,21 +416,28 @@ object Erasure {
...
 
+      val qual1 = typed(tree.qualifier, AnySelectionProto)
+
       def mapOwner(sym: Symbol): Symbol = {
-        def recur(owner: Symbol): Symbol =
-          if (defn.specialErasure.contains(owner)) {
+        def recur(owner: Symbol): Symbol = {
+          val owner = sym.maybeOwner
+          if (!owner.exists) {
+            // Hack for PolyFunction#apply
```
odersky: Explain more, please

odersky:
@milessabin Assigning to you to get this over the finish line.

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -1032,6 +1032,11 @@ class Definitions {
     if (n <= MaxImplementedFunctionArity && (!isContextual || ctx.erasedTypes) && !isErased) ImplementedFunctionType(n)
     else FunctionClass(n, isContextual, isErased).typeRef
 
+  lazy val PolyFunctionClass = ctx.requiredClass("scala.PolyFunction")
```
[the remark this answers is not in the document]
milessabin: Looking nearby, the pattern seems to be the same as here: the symbol is defined as a lazy val and the typeref as a def. Can you point me to an example which is arranged the way you want it?

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -1032,6 +1032,11 @@ class Definitions {
...
+  def PolyFunctionType = PolyFunctionClass.typeRef
+
+  private lazy val TupleTypes: Set[TypeRef] = TupleType.toSet
```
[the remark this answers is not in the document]
milessabin: Removed.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -923,6 +923,8 @@ object Parsers {
           val tparams = typeParamClause(ParamOwner.TypeParam)
           if (in.token == ARROW)
             atSpan(start, in.skipToken())(LambdaTypeTree(tparams, toplevelTyp()))
+          else if (isIdent && in.name.toString == "->")
```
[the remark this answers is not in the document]
milessabin: Done.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
milessabin: Done.

compiler/src/dotty/tools/dotc/transform/ElimErasedValueType.scala
milessabin: Done.

compiler/src/dotty/tools/dotc/transform/Erasure.scala
milessabin: Done.

milessabin:
* Rebased
* Responded to @odersky's review comments
* Switch to `=>` syntax
* Term-parameter-less poly functions excluded as an implementation restriction
* Poly function arguments of poly functions supported
* Fixed overloading for methods with poly function arguments
* Added tests

milessabin:
@smarter this is your PR so I can't request you as a reviewer via the UI :-)

milessabin:
@odersky I think this is ready to go.

milessabin:
Looks like the CLA bot needs a poke ...

smarter:
CLA bot breaks down when a commit has more than one author, you can ignore it.

compiler/src/dotty/tools/dotc/core/Definitions.scala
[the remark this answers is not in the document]
odersky: E.g.
```
  lazy val StringBuilderType: TypeRef      = ctx.requiredClassRef("scala.collection.mutable.StringBuilder")
  def StringBuilderClass(implicit ctx: Context): ClassSymbol = StringBuilderType.symbol.asClass

```
But I meant to go over `Definitions` anyway, trying to avoid the duplication and make it safe by design. The problem with the lazy val pattern as you wrote it is that it would not work in interactive mode if `PolyFunction` was edited. Then 
the system would hang on to the first version computed instead of the edited ones. I agree that's a rather esoteric use case. So we can leave it for now.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -924,9 +924,18 @@ object Parsers {
           val tparams = typeParamClause(ParamOwner.TypeParam)
           if (in.token == TLARROW)
             atSpan(start, in.skipToken())(LambdaTypeTree(tparams, toplevelTyp()))
-          else if (isIdent && in.name.toString == "->")
-            atSpan(start, in.skipToken())(PolyFunction(tparams, toplevelTyp()))
-          else { accept(TLARROW); typ() }
+          else if (in.token == ARROW) {
```
odersky: We should update the syntax in the doc comment to account for this, and do the same in file `syntax.md`. Generally, we try parser code, parser comments and syntax file in sync.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1455,10 +1443,28 @@ object desugar {
...
         }
+      case _ =>
+        EmptyTree //may happen for erroneous input
+    }
+
+    // begin desugar
+
+    // Special case for `Parens` desugaring: unlike all the desugarings below,
+    // its output is not a new tree but an existing one whose position should
+    // be preserved, so we shouldn't call `withPos` on it.
```
odersky: Nice catch.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1455,10 +1443,28 @@ object desugar {
...
             .map(varg => varg.withAddedFlags(mods.flags | Param))
-          New(Template(emptyConstructor, List(polyFunctionTpt), Nil, EmptyValDef,
-            List(DefDef(nme.apply, applyTParams, List(applyVParams), TypeTree(), res))
-          ))
+            New(Template(emptyConstructor, List(polyFunctionTpt), Nil, EmptyValDef,
+              List(DefDef(nme.apply, applyTParams, List(applyVParams), TypeTree(), res))
+              ))
         }
+      case _ =>
+        EmptyTree //may happen for erroneous input
```
smarter: If the input is erroneous, shouldn't we report an error here ? (also leave a space between `//` and the first character in a comment)

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3194,8 +3194,17 @@ object Types {
...
+      def polyFunctionSignature(tp: Type): Type = tp match {
+        case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
+          val res = refinedInfo.resultType
+          val paramss = res.paramNamess
+          defn.FunctionType(paramss.head.length)
+        case _ => tp
+      }
+
+      val params = if (isErasedMethod) Nil else paramInfos.mapConserve(polyFunctionSignature)
       resultSignature.prepend(params, isJavaMethod)
```
smarter: `Signature#prepend` ends up calling `TypeErasure#sigName`, changing that method to handle PolyFunction should avoid the need for a special case here.

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
milessabin: Done.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
milessabin: An error will already have been reported, this is to prevent `PolyFunction` trees from being passed along and blowing up downstream. I've added a comment to that effect (and fixed the whitespace).

compiler/src/dotty/tools/dotc/core/Types.scala
milessabin: Done.

milessabin:
Feedback addressed ... I'll merge when it goes green unless anyone objects.

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1448,7 +1448,7 @@ object desugar {
               ))
         }
       case _ =>
-        EmptyTree //may happen for erroneous input
+        EmptyTree // may happen for erroneous input. An error will already have been reported.
```
smarter: Instead of leaving it as a comment, you can enforce this with `assert(ctx.reporter.errorsReported)` (we do this in several places already)

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
```diff
@@ -605,6 +605,8 @@ class TypeErasure(isJava: Boolean, semiEraseVCs: Boolean, isConstructor: Boolean
       case tp: TypeVar =>
         val inst = tp.instanceOpt
         if (inst.exists) sigName(inst) else tpnme.Uninstantiated
+      case RefinedType(parent, nme.apply, refinedInfo) if parent.typeSymbol eq defn.PolyFunctionClass =>
+        sigName(defn.FunctionType(refinedInfo.resultType.paramNamess.head.length))
```
smarter: It still seems weird that an explicit case is needed here, given that the default case below does `val erasedTp = this(tp) ... sigName(erasedTp)` which should end up going through https://github.com/lampepfl/dotty/blob/93d3c2a719a4d59c5c6e24028fc52146097e5b1b/compiler/src/dotty/tools/dotc/core/TypeErasure.scala#L450

milessabin:
Green it is ... :tada:

compiler/src/dotty/tools/dotc/ast/Desugar.scala
milessabin: Done.

compiler/src/dotty/tools/dotc/core/TypeErasure.scala
milessabin: It doesn't get to the default case because It gets caught by the `TypeProxy` case immediately below.

smarter: Aha, I see. But then you should be able to do something like `sigName(this(tp))` in this case, and add a comment mentioning this is only needed to avoid going through the TypeProxy case below.

milessabin: Done.

milessabin:
Green now ... :tada:

biboudis:
We lack documentation in the Reference > New Types section.

aappddeevv:
Did any documentation on this ever get written? I did not see any on the dotty site or in github.

smarter:
Nope, documentation will follow once the implementation is complete.

aappddeevv:
Thanks. I had a specific and I think obvious need for this and AnyKind (if I understand these concepts correctly) working together. I'll hold off.

SethTisue:
> documentation will follow once the implementation is complete

is it time, now? This came up over at https://github.com/scala/scala-lang/pull/1186#discussion_r533445150 (Vincenzo's blog post on tuples)

smarter:
I still need to do type inference. Anyway I agree documentation would be nice, but I don't have time to deal with that right now.

nafg:
Sorry for mixing in, and if things need more time we need to give them more
time, but developers, myself included, don't give documentation enough
importance.

In some ways, having good documentation is more important than anything
else. What is the purpose of the whole Dotty effort? One objective is to
make life easier for existing Scala developers but I think the most
important goal is to make Scala accessible and speaking to a wider
audience. From that perspective, good documentation helps achieve that with
or without Dotty. Dotty should help tremendously but the extent to which it
does again depends on quality of documentation.

Honestly I'm not sure what specific thing is being discussed here (I get
emails for all repo events) so doing take this as referring to anything or
insinuating anything. Just wanted to write that.

If it takes hiring people specifically to do documentation properly (even
though it's already quite good) I hope they can.

PR #4675, opened 2018-06-18
Fix Tasty reflect patterns with select trees

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -474,7 +474,7 @@ object TastyImpl extends scala.tasty.Tasty {
     object Value extends ValueExtractor {
       def unapply(x: Pattern)(implicit ctx: Context): Option[Term] = x match {
         case lit: tpd.Literal @unchecked => Some(lit)
-        case ident: tpd.Ident @unchecked if ident.isTerm => Some(ident)
+        case ref: tpd.RefTree @unchecked if ref.isTerm => Some(ref)
```
liufengyun: When used in a traverser or accumulator, a method-select may qualify. Also, it may also return true for implicit params of extractors. Maybe add a doc for potential misuse.

PR #4676, opened 2018-06-18
Print annotation in decompiler

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -220,8 +227,16 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
         }
         this
 
-      case tree@Term.Ident(name) =>
-        printType(tree.tpe)
+      case tree @ Term.Ident(name) =>
+        tree.tpe match {
+          case Type.SymRef(_, Types.EmptyPrefix()) | Type.TermRef(_, Types.EmptyPrefix()) => this += name
```
liufengyun: It seems `SymRef` subsumes `TermRef`?

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -768,13 +800,35 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
     }
 
     def printDefinitionName(sym: Definition): Buffer = sym match {
-      case ValDef(name, _, _) => this += name
+      case ValDef(name, _, _) => this += name += ".type"
```
liufengyun: why this change?

nicolasstucki: I'm trying to figure out how to print `(x: @annot)`, the issue is that it gets expanded to something like `(x: x @annot)`. I was trying to workaround it by printing `(x: x.type @annot)` instead. I will now try to print it as `(x: @annot)` as it was originally.

nicolasstucki:
@liufengyun could you review the changes in the last commit

nicolasstucki:
@liufengyun last 2 commits

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -654,11 +654,20 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
+            printAnnotation(annot)
+          case tpe @ Type.TypeRef(name, _) if name.endsWith("$") =>
+            printType(tpe)
+            this += ".type"
+          case tpe @ Type.SymRef(ClassDef(name, _, _, _, _), _) if name.endsWith("$") =>
+            printType(tpe)
+            this += ".type"
+          case tpe => printType(tpe)
+        }
+        printTypeAndAnnots(tree.tpe)
```
liufengyun: `tests/run-with-compiler/i3947c.scala` failed due to the change, I cannot figure out why.

nicolasstucki: It was because `scala.runtime.Null$` and `scala.runtime.Nothing$` where handled as modules. Those are now special cases.

PR #4682, opened 2018-06-19
Fix type of Tasty.Type.TypeSelect.prefix

nicolasstucki:
Previously we did not support type trees as prefixes


liufengyun:
`tests/run/tasty-extractors-2` failed.

nicolasstucki:
`tests/run/tasty-extractors-2` should be fixed now

compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -554,9 +554,16 @@ object TastyImpl extends scala.tasty.Tasty {
...
+        case x: tpd.Select @unchecked if x.isType && x.qualifier.isTerm => Some(x.qualifier, x.name.toString)
+        case _ => None
+      }
+    }
+
+    object TypeSelect extends TypeSelectExtractor {
+      def unapply(x: TypeTree)(implicit ctx: Context): Option[(TypeTree, String)] = x match {
+        case x: tpd.Select @unchecked if x.isType && x.qualifier.isType => Some(x.qualifier, x.name.toString)
         case _ => None
       }
```
liufengyun: I find the name a little misleading, what about `TypeProject`? Then we can keep using `TypeSelect` as before.

nicolasstucki: Users would write the extractors as `Type.TypeSelect(_, _)` and `Type.TermSelect(_, _)`. We should have a pass on all names later to make sure all names are good.

PR #4686, opened 2018-06-19
Fix printing parents in decompiler

nicolasstucki:
* Fixed multiple argument list constructors
* Added missing `with` in between the parents


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -89,33 +89,39 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
-            case parent@TypeTree() =>
-              sb.append(" ")
+            case parent @ TypeTree() =>
               printTypeTree(parent)
+
+            case parent @ Term() => throw new MatchError(parent.show)
+          }
+
+          def printSeparated(list: List[Parent]): Unit = list match {
+            case Nil =>
```
liufengyun: unreachable case, throw an exception here if reached?

nicolasstucki: I moved the `nonEmpty` check, now this is reachable.

Issue #4583, opened 2018-05-26
Singleton bound and type projections don't play nice with each other


Blaisorblade:
If a value member `tasty` has abstract type `T` and `T` is marked as being a singleton type (that is, currently, `T <: Singleton`), I’d expect `tasty.type = T`, because both types are singleton that contain `tasty`. This currently fails in Scalac and Dotty (the Scala report is scala/bug#10905 — this is a copy, with Scalac output swapped with Dotty output).

A minimized example:

```scala
scala> trait A { type T; val prod: T; val cons: T => Int = _ => 0 }; trait Foo { type U <: Singleton with A; val tasty: U }
// defined trait A
// defined trait Foo

scala> val v: A = new A { type T = Int; val prod: T = 1}; val u: Foo = new Foo { type U = v.type; val tasty: U = v }
val v: A = anon$1@5416f8db
val u: Foo = anon$2@64e1377c

scala> u.tasty.cons(u.tasty.prod: u.U#T)
1 |u.tasty.cons(u.tasty.prod: u.U#T)
  |             ^^^^^^^^^^^^^^^^^^^
  |             found:    u.U#T
  |             required: u.tasty.T
  |
```

/cc @milessabin 

### Background

I ran into this while prototyping a heavily path-dependent API in Dotty (an early version of Tasty reflection), in relation to https://github.com/lampepfl/dotty/pull/4577/files/b315dbfa9f2d206cc26da75779388fca98faaabd#r191016341; the workaround involved turning defs into vals, which I suspect is acceptable here but potentially annoying when trying to save memory.

### Alternatives

For the example above, enabling the definition of idempotent methods might be more general, if we ever get there.

### Soundness

Not formally investigated, since nobody yet added to DOT either singleton types, type projections or singleton type variables.

milessabin:
Is there a clear statement of the intended semantics of type projections, in Dotty or in DOT?

My intuitions in Scala were always that `T#U` _ought_ to mean something like `t.U forAll { val t: T }` if there were such a thing as a `forAll` construct. But the implementation in Scala never aligned with that.

Blaisorblade:
> Is there a clear statement of the intended semantics of type projections, in Dotty or in DOT?

Nothing fully satisfactory, but something exists — the 2003 nuObj paper is a good starting point and describes the _intended_ semantics, but some of its rules break down in the presence of lower bounds. What's below are pretty educated guesses.

[...]

PR #4688, opened 2018-06-19
Add type refinement for abstract type bindings

nicolasstucki:


smarter:
If I can write any type on the rhs of `Refined`, isn't this completely unsound?

nicolasstucki:
There are still some constraints that must be added.

Blaisorblade:
But do we have a way to add constraints that makes this sound? I thought the plan was to warn the user that they use an escape hatch, just like `asInstanceOf`. Maybe call the member `UnsafeRefined` to make this more visible?

nicolasstucki:
@Blaisorblade the current implementation in this PR is a partial implementation of the new scheme.

nicolasstucki:
The unboxed result of `RefinedScrutinee` has the expected performance. Here is a small experiment https://gist.github.com/nicolasstucki/96347efecaa5f6fb71d7146635c079c3

Blaisorblade:
FWIW: last I tried to write value class decorators with a single member, I ran into #4583, which seems hard to avoid.

nicolasstucki:
Will reopen it when I'll have time to work on it

PR #4689, opened 2018-06-20
Print `;` at the end of statements to avoid ambiguity

nicolasstucki:
For example, the first snippet is `foo.apply(...)` and the second is
eval `foo` and then eval the block.

```scala
foo
{
 ...
}
```

```scala
foo;
{
 ...
}
```


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -400,6 +392,14 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
     }
 
+    def printStats(stats: List[Tree]): Unit = {
+      if (!stats.isEmpty) {
+        this += lineBreak()
+        printTrees(stats, ";" + lineBreak())
+        this += ";"
+      }
+    }
+
```
liufengyun: Can we add a blank line instead of `;`?

Also, it seems it's only needed if the previous tree may cause ambiguity. This can be an improvement.

nicolasstucki: It can be implemented but it would require knowing which is the next tree is printed and how it will be printed. This solution is much simpler, homogenous and less bug-prone. I will try to remove them in the future once I have a complete test suite.

Using by default an additional blank line add too many lines. I would consider if we only print it when needed.

allanrenucci: Should we add `;` all the time? Looking at the diff, I find it a bit weird that sometime a statement ends with `;` and sometime not

nicolasstucki: I added it at the end of all statements. The expression at the end of the block does not need it. Look at it as a separator.

Still later I will work on making the format nicer to look at. For now the main goal is corecness.

nicolasstucki:
@liufengyun could you re-review this PR

PR #4693, opened 2018-06-20
Fix #4674: Add regression test

### LINKED ISSUE #4674: specifying a type in pattern matching results in a spurious warning
```
val x: String = "something"
x.foreach {
  case 's' => println("s")
  case c: Char => println(c)
} 
```

Results in:
```
[warn] 154 |        case c: Char                   => println(c)
[warn]     |             ^
[warn]     |this will always yield true, since `class Char` is a subclass of `class Char`
```

While it's true that I can just omit the `: Char`, the intent here was to increase readability by specifying the type. Feel free to close if this is not a bug but an intended change.

janrock-ylb:
Tested based on http://dotty.epfl.ch/docs/contributing/testing.html#integration-tests.

```
scala> object Test {
           val x: String = "test"
           x.foreach {
             case 's' => println("s")
             case c: Char => println(c)
           }
         } 
7 |      case c: Char => println(c)
  |           ^^^^^^^
  |The highlighted type test will always succeed since the scrutinee type (class Char) is the same as the tested type
// defined object Test
```


tests/neg-custom-args/fatal-warnings/i4674.scala
```diff
@@ -0,0 +1,9 @@
+object i4674 {
+
+  val x: String = "test"
+
+  x.foreach {
+    case 's' => println("s")
+    case c: Char => println(c)
```
allanrenucci: You will need to add `// error: type test always succeed` at the end of the line to make the test succeed

PR #4694, opened 2018-06-21
Fix printing `catch` clause

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -383,8 +383,12 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
         printTree(body)
         if (cases.nonEmpty) {
-          this += " catch "
-          printCases(cases, lineBreak())
+          this += " catch {"
+          indented {
+            this += lineBreak()
+            printCases(cases, lineBreak())
+          }
+          this += lineBreak() += "}"
```
liufengyun: I think it's worthy to create another verb like `braced` which could be used in several places to simplify the code.

nicolasstucki: Nice idea. I will add it in another PR.

PR #4697, opened 2018-06-21
Print this id and filter object lazy val

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -317,6 +323,7 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
       case Term.Block(stats0, expr) =>
         def isLoopEntryPoint(tree: Tree): Boolean = tree match {
           case Term.Apply(Term.Ident("while$" | "doWhile$"), _) => true
+          case tree @ ValDef(_, _, _) => tree.flags.isObject
           case _ => false
         }
```
liufengyun: The name `isLoopEntryPoint` and the change don't align.

nicolasstucki: Done

PR #4704, opened 2018-06-22
Fix printing binded alternatives

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -677,7 +677,9 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
         this += ")"
 
       case Pattern.Alternative(trees) =>
+        this += "("
         printPatterns(trees, " | ")
+        this += ")"
```
liufengyun: Introduce a verb `inParens { ... }`?

PR #4705, opened 2018-06-22
Print `val` for class parameters

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -621,6 +621,17 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
+        case DefDef("<init>", _, _, _, _) =>
+          val ClassDef(_, _, _, _, body) = arg.owner.owner
+          body.collectFirst {
+            case vdef @ ValDef(`name`, _, _) if vdef.flags.isParamAccessor =>
+              if (!vdef.flags.isLocal && !vdef.flags.isCaseAcessor)
+                this += "val " // TODO `var`s
+          }
+        case _ =>
+      }
+
```
liufengyun: Rename `printArgDef` to `printParam`? My impression is that `param` is used at definition site, `arg` is used usually used at calling site.

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -621,6 +621,17 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
 
     def printArgDef(arg: ValDef): Unit = {
       val ValDef(name, tpt, rhs) = arg
+      arg.owner match {
+        case DefDef("<init>", _, _, _, _) =>
+          val ClassDef(_, _, _, _, body) = arg.owner.owner
+          body.collectFirst {
+            case vdef @ ValDef(`name`, _, _) if vdef.flags.isParamAccessor =>
+              if (!vdef.flags.isLocal && !vdef.flags.isCaseAcessor)
+                this += "val " // TODO `var`s
```
liufengyun: More todos: access modifiers, default parameters, implicits.

nicolasstucki:
@liufengyun could you review the second commit

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -1053,6 +1059,48 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
+            printWithin(within)
+            this += "] "
+            printedPrefix = true
+          case _ =>
+        }
+      }
+      printedPrefix
+    }
+
+    def printFullClassName(tp: TypeOrBounds): Unit = {
```
liufengyun: `TypeOrBounds` -> `Type`? The same for the next line.

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -619,15 +623,24 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
+              if (!vdef.flags.isLocal) {
+                var printedPrefix = false
+                if (vdef.flags.isOverride) {
+                  this += "override "
+                  printedPrefix = true
+                }
+                printedPrefix  |= printProtectedOrPrivate(vdef)
+                if (vdef.flags.isMutable) this += "var "
+                else if (!printedPrefix && vdef.flags.isCaseAcessor) this // val not explicitly needed
+                else this += "val "
```
liufengyun: I would negate the second `else if` and put `this += "val "` as the body.

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -619,15 +623,24 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
...
           val ClassDef(_, _, _, _, body) = arg.owner.owner
           body.collectFirst {
             case vdef @ ValDef(`name`, _, _) if vdef.flags.isParamAccessor =>
-              if (!vdef.flags.isLocal && !vdef.flags.isCaseAcessor)
-                this += "val " // TODO `var`s
+              if (!vdef.flags.isLocal) {
+                var printedPrefix = false
+                if (vdef.flags.isOverride) {
+                  this += "override "
+                  printedPrefix = true
```
liufengyun: Why skip `val ` (line 641) if we have `override`?

library/src/scala/tasty/util/ShowSourceCode.scala
liufengyun: The name `printedPrefix` is not informative enough to understand the logic here.

library/src/scala/tasty/util/ShowSourceCode.scala
nicolasstucki: We skip `val` if we have NOT `override`

PR #4706, opened 2018-06-22
Fix printing package objects and nested packages

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -74,7 +83,11 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
         if (flags.isFinal && !flags.isObject) this += "final "
         if (flags.isCase) this += "case "
 
-        if (flags.isObject) this += "object " += name.stripSuffix("$")
+        if (name == "package$") {
+          this += "package object "
+          printDefinitionName(cdef.owner)
+        }
```
liufengyun: Name-checking may potentially cause problems, though we expect no one will write the following code:

```Scala
class `package$`  {
  def `<init>` = ()
}
```

nicolasstucki: `$` is reserved for compiler use only by spec.

nicolasstucki: Or in this case the `$` might be encoded

PR #4710, opened 2018-06-23
Some Rudimentary Tracking of Class Purity

odersky:
This PR breaks out the commits in #4616 that let us establish a rudimentary notation of class purity. A class does not need an initalizer (indicated by the `NoInits` flag) if it only contains method and type definitions or abstract declarations of any kind and if it does not pass arguments to its parent class. A class is pure if it and all its baseclasses are `NoInits`. Constructors of pure classes are marked `stable`, which means that a `new` expression can now be pure. Currently purity is only indicated for Dotty classes and a few classes in `Definitions`. 

This scheme is admittedly quite rudimentary but it allows us to classify typical case classes that define data types as pure. This in turn is necessary to reduce projections of such data classes occurring in transparent methods.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
+++ b/compiler/src/dotty/tools/dotc/ast/TreeInfo.scala
@@ -243,10 +243,13 @@ trait TreeInfo[T >: Untyped <: Type] { self: Trees.Instance[T] =>
   /**  The largest subset of {NoInits, PureInterface} that a
    *   trait enclosing this statement can have as flags.
    */
-  def defKind(tree: Tree): FlagSet = unsplice(tree) match {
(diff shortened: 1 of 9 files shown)


odersky:
Tests are currently part of #4616, in particularly `neg/typelevel.scala`. We verify purity by observing that inlining took place.

odersky:
Hi Georg, can you have a look at this and see how it relates to your notion of stability?

odersky:
@gsps Can you give this a quick review please? I believe it would simplify things if we had this merged now.

Blaisorblade:
I'm focused on the other deadline, but I'll try to review this right after.

odersky:
Merging now, since not doing so would complicate things going forward.

PR #4711, opened 2018-06-23
Fix #4674: Add regression test

### LINKED ISSUE #4674: specifying a type in pattern matching results in a spurious warning

janrock-ylb:
Test: Passed

### DIFF
--- a/tests/neg-custom-args/fatal-warnings/i4674.scala
+++ b/tests/neg-custom-args/fatal-warnings/i4674.scala
@@ -1,0 +1,9 @@
+object Testi4674 {
+  def T(x: String) = {
+    x.foreach {
+      case 's' => println("s")
+      case c: Char => println(c) // error
+    }
+  }
+  T("test")
+}
--- a/tests/neg-custom-args/fatal-warnings/i4674.scala
+++ b/tests/neg-custom-args/fatal-warnings/i4674.scala
@@ -1,9 +1,8 @@
-object Testi4674 {
-  def T(x: String) = {
+class Test {
+  def test(x: String) = {
     x.foreach {
       case 's' => println("s")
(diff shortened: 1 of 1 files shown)


tests/neg-custom-args/fatal-warnings/i4674.scala
```diff
@@ -0,0 +1,9 @@
+object Testi4674 {
+  def T(x: String) = {
+    x.foreach {
+      case 's' => println("s")
+      case c: Char => println(c) // error
```
allanrenucci: Can you write `// error: type test always succeed` instead. Otherwise we don't what is the error without running the test

tests/neg-custom-args/fatal-warnings/i4674.scala
```diff
@@ -0,0 +1,9 @@
...
+    }
+  }
+  T("test")
```
allanrenucci: This is not needed. I would remove it. We try to keep tests as small as possible

janrock-ylb:
Done.

Thank you!

janrock-ylb:
CLA signed off. Confirmed from The Scala Team.

allanrenucci:
CLA [signed](https://www.lightbend.com/contribute/cla/scala/check/janrock-ylb). Thanks @janrock-ylb  for the contribution

PR #4715, opened 2018-06-24
Fix printed references to module types

nicolasstucki:


library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -904,7 +902,15 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
           case Type.SymRef(cdef @ ClassDef(_, _, _, _, _), _) if !cdef.flags.isObject =>
             printFullClassName(tp)
             this += ".this"
-          case _ => printType(tp)
+          case Type.TypeRef(name, prefix) if name.endsWith("$") =>
+            prefix match {
+              case Types.EmptyPrefix() =>
+              case _ =>
+                printTypeOrBound(prefix) += "."
```
liufengyun: I would change this line into two lines: 

``` Scala
printTypeOrBound(prefix)
this += "."
```

library/src/scala/tasty/util/ShowSourceCode.scala
```diff
@@ -750,9 +750,6 @@ class ShowSourceCode[T <: Tasty with Singleton](tasty0: T) extends Show[T](tasty
             printTypeAndAnnots(tp)
             this += " "
             printAnnotation(annot)
-          case tpe @ Type.TypeRef(name, _) if name.endsWith("$") =>
-            printType(tpe)
-            this += ".type"
```
liufengyun: Why the two cases below are not handled in TYPE printing as well ? Can we move the whole method `printTypeAndAnnots` to TYPE printing part, instead of being a local method inside TREE printing?

nicolasstucki:
There is no relationship between ShowSourceCode and PlainPrinter and RefinedPrinter. They are not intended to be unified.

PR #4719, opened 2018-06-26
Mention submodules in getting started doc

adriaanm:


docs/docs/contributing/getting-started.md
```diff
@@ -20,8 +20,11 @@ $ git clone --recursive --single-branch https://github.com/lampepfl/dotty.git
 $ cd dotty
 $ sbt managedSources # Needed for IDE import to succeed
 ```
+
 Pass `--single-branch` to clone only the master branch, otherwise cloning will be *much* slower (details in [issue #3236](https://github.com/lampepfl/dotty/issues/3236)).
 
+The repository uses submodules, which may need updating/initializing using `git submodule update --init`. The `managedSources` sbt task will alert you if this is needed.
```
allanrenucci: If the repository was cloned using the instruction above (i.e. using `--recursive`), it shouldn't be needed. I guess it is still useful if the repo was not cloned properly.

However, sbt will log an error at startup time. It is not the `managedSources` task. I would clarify this.

Glavo: The `--recursive` option has been deprecated and we'd better update it in the documentation with the `--recurse-submodules` option

PR #4621, opened 2018-06-05
Fix #4611: Change implicit function types rewriting rules

allanrenucci:
An abstract method definition

    def m(xs: Ts): IF

is expanded to:

    def m(xs: Ts): IF = implicit (ys: Us) => m$direct(xs)(ys)
    def m$direct(xs: Ts)(ys: Us): R

An overriding method definition

    override def m(xs: Ts): IF = implicit (ys: Us) => E

is expanded to:

    def m$direct(xs: Ts)(ys: Us): R = E

PR #4723, opened 2018-06-26
Fix #4611

### LINKED ISSUE #4611: Compiler crash: Implicit Function types + SAM
```scala
trait Responder[T] {
  def responseFor(value: T): implicit Request => T
}

object Responder {
  val responseResponder: Responder[Response] =
    response => response
}

class Response
class Request
```
```
Exception in thread "main" scala.MatchError: ArrayBuffer(<SingleDenotation of type MethodType(List(value), List(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Response)), AppliedType(TypeRef(ThisType(TypeRef(NoPrefix,module class scala)),class ImplicitFunction1),List(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Request), TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Response))))>, <SingleDenotation of type MethodType(List(value), List(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Response)), MethodType(List(x$0), List(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Request)), TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class Response)))>) (of class scala.collection.mutable.ArrayBuffer)
	at dotty.tools.dotc.transform.ExpandSAMs.transformBlock(ExpandSAMs.scala:46)
	at dotty.tools.dotc.transform.MegaPhase.goBlock(MegaPhase.scala:668)
        ...
```
Initially reported by @jeffmay on Gitter

allanrenucci:
class and trait with a single abstract method which result type is an
implicit function type are not considered SAM types anymore. The reason
is because the abstract member desugar to mutiple ones. E.g.

```scala
trait Foo {
  def foo(): implicit Int => Int
}

// becomes
trait Foo {
  def foo(): implicit Int => Int
  def foo$direct(implicit x: Int): Int
}
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Types.scala
+++ b/compiler/src/dotty/tools/dotc/core/Types.scala
@@ -3868,7 +3868,7 @@ object Types {
    *  A type is a SAM type if it is a reference to a class or trait, which
    *
    *   - has a single abstract method with a method type (ExprType
-   *     and PolyType not allowed!)
+   *     and PolyType not allowed!) whose result type is not an implicit function type
    *   - can be instantiated without arguments or with just () as argument.
    *
    *  The pattern `SAMType(sam)` matches a SAM type, where `sam` is the
@@ -3910,7 +3910,8 @@ object Types {
         // println(s"absMems: ${absMems map (_.show) mkString ", "}")
         if (absMems.size == 1)
           absMems.head.info match {
-            case mt: MethodType if !mt.isParamDependent =>
+            case mt: MethodType if !mt.isParamDependent &&
+                !defn.isImplicitFunctionType(mt.resultType) =>
               val cls = tp.classSymbol
 
               // Given a SAM type such as:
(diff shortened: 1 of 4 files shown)


allanrenucci:
An alternative fix which did not introduce this restriction on SAM types was tried in #4621.

For now this fixes the compiler crash. In the future, if one wants to try again and lift the restriction, `implicitFuns2.scala` is a very good test case

allanrenucci:
Blocked on #4753

compiler/src/dotty/tools/dotc/core/Types.scala
```diff
@@ -3837,7 +3837,7 @@ object Types {
    *  A type is a SAM type if it is a reference to a class or trait, which
    *
    *   - has a single abstract method with a method type (ExprType
-   *     and PolyType not allowed!)
+   *     and PolyType not allowed!) which result type is not an implicit function type
```
smarter: which -> whose

PR #4724, opened 2018-06-26
Add collections-strawman to our tests

allanrenucci:


allanrenucci:
Before we merge, let me add it as a TASTY test as well

allanrenucci:
Merging now. We can add it as a TASTY test in the future

PR #4726, opened 2018-06-26
Implement Shonan HMM

nicolasstucki:


biboudis:
Maybe add the link to the paper in both files before merging. Just for reference.

Looks good!

PR #4728, opened 2018-06-27
Fix Tasty reflect extractor for flatten closure blocks

nicolasstucki:


nicolasstucki:
Bug appeared in #4726 and solution tested in #4726

PR #4729, opened 2018-06-27
Fix #4709: report if local implicits need type annotation

### LINKED ISSUE #4709: Cyclic reference involving implicit value
```scala
class Context
class ContextBase { def settings = 1 }

class Test {
  implicit def toBase(ctx: Context): ContextBase = ???

  def test(ctx0: Context) = {
    implicit val ctx = { ctx0.settings; ??? }
  }
}
```
This code snippet above fails to compile:
```
-- [E047] Syntax Error: tests/allan/Test.scala:8:25 ----------------------------
8 |    implicit val ctx = { ctx0.settings; ??? }
  |                         ^
  |                         cyclic reference involving implicit value ctx
```
Adding an explicit type annotation to `val ctx` solves the issue

Blaisorblade:
Fix #4709 and #3253, and revisit fixes to #2001.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
@@ -109,6 +109,12 @@ class CyclicReference private (val denot: SymDenotation) extends TypeError {
   override def toMessage(implicit ctx: Context): Message = {
     val cycleSym = denot.symbol
 
+    // cycleSym.flags would try completing denot and would fail, but here we can use flagsUNSAFE to detect flags
+    // set by the parser.
+    val unsafeFlags = cycleSym.flagsUNSAFE
+    val isMethod = unsafeFlags.is(Method)
+    val isVal = !isMethod && cycleSym.isTerm
+
     /* This CyclicReference might have arisen from asking for `m`'s type while trying to infer it.
      * To try to diagnose this, walk the context chain searching for context in
      * Mode.InferringReturnType for the innermost member without type
@@ -117,13 +123,15 @@ class CyclicReference private (val denot: SymDenotation) extends TypeError {
     def errorMsg(cx: Context): Message = {
       if (cx.mode is Mode.InferringReturnType) {
         cx.tree match {
-          case tree: untpd.ValOrDefDef if inImplicitSearch && !tree.tpt.typeOpt.exists =>
-            // Can happen in implicit defs (#4709) or outside (#3253).
-            TermMemberNeedsResultTypeForImplicitSearch(cycleSym)
-          case tree: untpd.DefDef if !tree.tpt.typeOpt.exists =>
-            OverloadedOrRecursiveMethodNeedsResultType(tree.name)
-          case tree: untpd.ValDef if !tree.tpt.typeOpt.exists =>
-            RecursiveValueNeedsResultType(tree.name)
+          case tree: untpd.ValOrDefDef if !tree.tpt.typeOpt.exists =>
+            if (inImplicitSearch)
+              TermMemberNeedsResultTypeForImplicitSearch(tree.name, cycleSym)
+            else if (isMethod)
+              OverloadedOrRecursiveMethodNeedsResultType(tree.name, cycleSym)
+            else if (isVal)
+              RecursiveValueNeedsResultType(tree.name, cycleSym)
+            else
+              errorMsg(cx.outer)
           case _ =>
             errorMsg(cx.outer)
         }
--- a/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
@@ -125,11 +125,11 @@ class CyclicReference private (val denot: SymDenotation) extends TypeError {
         cx.tree match {
           case tree: untpd.ValOrDefDef if !tree.tpt.typeOpt.exists =>
             if (inImplicitSearch)
-              TermMemberNeedsResultTypeForImplicitSearch(tree.name, cycleSym)
+              TermMemberNeedsResultTypeForImplicitSearch(cycleSym)
             else if (isMethod)
-              OverloadedOrRecursiveMethodNeedsResultType(tree.name, cycleSym)
+              OverloadedOrRecursiveMethodNeedsResultType(cycleSym)
             else if (isVal)
-              RecursiveValueNeedsResultType(tree.name, cycleSym)
+              RecursiveValueNeedsResultType(cycleSym)
             else
               errorMsg(cx.outer)
           case _ =>
--- a/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeErrors.scala
@@ -98,13 +98,28 @@ object handleRecursive {
   }
 }
 
+/**
+ * This TypeError signals that completing denot encountered a cycle: it asked for denot.info (or similar),
+ * so it requires knowing denot already.
+ * @param denot
+ */
 class CyclicReference private (val denot: SymDenotation) extends TypeError {
+  var inImplicitSearch: Boolean = false
 
-  override def toMessage(implicit ctx: Context) = {
+  override def toMessage(implicit ctx: Context): Message = {
+    val cycleSym = denot.symbol
 
-    def errorMsg(cx: Context): Message =
+    /* This CyclicReference might have arisen from asking for `m`'s type while trying to infer it.
+     * To try to diagnose this, walk the context chain searching for context in
+     * Mode.InferringReturnType for the innermost member without type
+     * annotations (!tree.tpt.typeOpt.exists).
+     */
+    def errorMsg(cx: Context): Message = {
       if (cx.mode is Mode.InferringReturnType) {
         cx.tree match {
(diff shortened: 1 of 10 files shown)


tests/neg/i4709.scala
```diff
@@ -0,0 +1,10 @@
+class Context
+class ContextBase { def settings = 1 }
+
+class Test {
+  implicit def toBase(ctx: Context): ContextBase = ???
+
+  def test(ctx0: Context) = {
+    implicit val ctx = { ctx0.settings; ??? } // error
+  }
+}
```
allanrenucci: I would make it an error message test as well

Blaisorblade:
Memo: both here and in #3253, implicit search fails because it tries to force the type of the member we're typechecking — whether the method is actually implicit or not.  In #3253, we produce `OverloadedOrRecursiveMethodNeedsResultType` without checking if there's actually any overload/recursion. Ditto for `RecursiveValueNeedsResultType`.

It seems we should check, instead, if the failure is in implicit resolution or not (not sure how, `ctx.searchHistory.searchDepth > 0` only finds *nested* implicit searches; I might have to record more info), maybe what tree we're looking at. If we're not doing implicit search, then we can use the current errors.

Blaisorblade:
I figured how to do the basics and tell if I'm in implicit search, now I'm adding tests on mutual recursion on which method should be reported — the unannotated method we find (in `tree`) is not the one which triggers the cyclic error (`cycleSym`), but the current errors have specific assumptions on whether the mentioned identifier is a method or a value.
I could try to extract that through flags, or maybe steal from `SymDenotation.companionNamed` code such as `tpd.definingStats(symbol).iterator.map(tpd.definedSym)`...

Blaisorblade:
Memo for future explorers: if you really really want, it might be possible to find all the symbols involved in a cycle, but at least for now we decided not to try.

This would work following the idea of #4385: we'd recording more information on the `CyclicReference` as it traverses the stack outward — one would have to catch it and rethrow it in `SymDenotation.completeFrom`.

compiler/src/dotty/tools/dotc/core/NameOps.scala
```diff
@@ -49,7 +49,8 @@ object NameOps {
     }
   }
 
-  implicit class NameDecorator[N <: Name](val name: N) extends AnyVal {
+  implicit class NameDecorator[N <: Name](val name_ : N) extends AnyVal {
```
allanrenucci: I think you can make it private:
```diff
-  implicit class NameDecorator[N <: Name](val name: N) extends AnyVal {
+  implicit class NameDecorator[N <: Name](private val name: N) extends AnyVal {
     import nme._
 
     def testSimple(f: SimpleName => Boolean): Boolean = name match {
@@ -233,12 +233,12 @@ object NameOps {
       case nme.clone_ => nme.clone_
     }
 
-    def specializedFor(classTargs: List[Types.Type], classTargsNames: List[Name], methodTargs: List[Types.Type], methodTarsNames: List[Name])(implicit ctx: Context): name.ThisName = {
+    def specializedFor(classTargs: List[Types.Type], classTargsNames: List[Name], methodTargs: List[Types.Type], methodTarsNames: List[Name])(implicit ctx: Context): N = {
 
       val methodTags: Seq[Name] = (methodTargs zip methodTarsNames).sortBy(_._2).map(x => defn.typeTag(x._1))
       val classTags: Seq[Name] = (classTargs zip classTargsNames).sortBy(_._2).map(x => defn.typeTag(x._1))
 
-      name.likeSpaced(name ++ nme.specializedTypeNames.prefix ++
+      likeSpaced(name ++ nme.specializedTypeNames.prefix ++
         methodTags.fold(nme.EMPTY)(_ ++ _) ++ nme.specializedTypeNames.separator ++
         classTags.fold(nme.EMPTY)(_ ++ _) ++ nme.specializedTypeNames.suffix)
```

compiler/src/dotty/tools/dotc/core/TypeErrors.scala
```diff
@@ -98,13 +98,28 @@ object handleRecursive {
   }
 }
 
+/**
+ * This TypeError signals that completing denot encountered a cycle: it asked for denot.info (or similar),
+ * so it requires knowing denot already.
+ * @param denot
+ */
 class CyclicReference private (val denot: SymDenotation) extends TypeError {
+  var inImplicitSearch: Boolean = false
```
allanrenucci: Can you make this a parameter of the class and then do something like:
```scala
def withinImplicitSearch =
  if (inImplicitSearch) this
  else new CyclicReference(denot, inImplicitSearch = true)

...

} catch {
  case ce: CyclicReference =>
    throw ce.withinImplicitSearch
}
```
I'm not the best to come up with method name

compiler/src/dotty/tools/dotc/typer/Implicits.scala
```diff
@@ -840,36 +840,42 @@ trait Implicits { self: Typer =>
...
-          }
-          else result
-        case _ =>
-          result
+            else result
+          case _ =>
+            result
+        }
       }
+    } catch {
```
allanrenucci: Does this need to wrap the whole method? Can't you do:
```scala
val result =
  try new ImplicitSearch(pt, argument, pos).bestImplicit(contextual = true)
  catch {
    case ce: CyclicReference =>
      ...
  }
```

compiler/src/dotty/tools/dotc/core/TypeErrors.scala
```diff
@@ -109,6 +109,13 @@ class CyclicReference private (val denot: SymDenotation) extends TypeError {
   override def toMessage(implicit ctx: Context): Message = {
     val cycleSym = denot.symbol
 
+    // cycleSym.flags would try completing denot and would fail, but here we can use flagsUNSAFE to detect flags
+    // set by the parser.
+    val unsafeFlags = cycleSym.flagsUNSAFE
+    val isMethod = unsafeFlags.is(Method)
+    val isVal = !isMethod && cycleSym.isTerm
+    val isImplicit = unsafeFlags.is(Implicit)
```
allanrenucci: unused?

compiler/src/dotty/tools/dotc/core/TypeErrors.scala
Blaisorblade: The new Exception has a different stack trace and other state. I could try to copy the various relevant fields (I spent some time investigating), but that seems more fragile than having one mutable field.

Blaisorblade:
Hey @allanrenucci can I push a rebase on top of #4834? I'd like to squash the review fixes in the respective commits as well...

compiler/src/dotty/tools/dotc/core/TypeErrors.scala
```diff
@@ -98,28 +98,52 @@ object handleRecursive {
...
-            RecursiveValueNeedsResultType(tree.name)
+          case tree: untpd.ValOrDefDef if !tree.tpt.typeOpt.exists =>
+            if (inImplicitSearch)
+              TermMemberNeedsResultTypeForImplicitSearch(tree.name, cycleSym)
+            else if (isMethod)
+              OverloadedOrRecursiveMethodNeedsResultType(tree.name, cycleSym)
+            else if (isVal)
+              RecursiveValueNeedsResultType(tree.name, cycleSym)
+            else
+              errorMsg(cx.outer)
```
allanrenucci: Yes I think we should remove all of this if `tree.name` is unused and we only use the `cycleSym`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1260,26 +1260,27 @@ object messages {
...
 
-  case class RecursiveValueNeedsResultType(value: Names.TermName)(implicit ctx: Context)
+  case class RecursiveValueNeedsResultType(valName: Names.TermName, cycleSym: Symbol)(implicit ctx: Context)
   extends Message(RecursiveValueNeedsResultTypeID) {
     val kind = "Syntax"
-    val msg = hl"""recursive value ${value} needs type"""
+    val msg = hl"""recursive ${cycleSym} needs type"""
     val explanation =
-      hl"""The definition of `${value}` is recursive and you need to specify its type.
+      hl"""The definition of `${cycleSym}` is recursive and you need to specify its type.
```
allanrenucci: Nitpick: here and in some of the error messages above s/`${cycleSym}`/`$cycleSym`

compiler/src/dotty/tools/dotc/reporting/diagnostic/messages.scala
```diff
@@ -1260,26 +1260,27 @@ object messages {
            |""".stripMargin
   }
 
-  case class OverloadedOrRecursiveMethodNeedsResultType(method: Names.TermName)(implicit ctx: Context)
+  case class OverloadedOrRecursiveMethodNeedsResultType(methodName: Names.TermName, cycleSym: Symbol)(implicit ctx: Context)
   extends Message(OverloadedOrRecursiveMethodNeedsResultTypeID) {
     val kind = "Syntax"
```
allanrenucci: Oh, and we should probably classify `OverloadedOrRecursiveMethodNeedsResultType`, `RecursiveValueNeedsResultType`, `TermMemberNeedsResultTypeForImplicitSearch`, `CyclicReferenceInvolvingImplicit` and `CyclicReferenceInvolving` as `Cyclic` errors not `Syntax`

Blaisorblade:
All comments addressed, so merging.

PR #4731, opened 2018-06-27
Fix printing of package object symbol

allanrenucci:


compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -745,8 +745,12 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
         case info: ImportType => return s"import $info.expr.show"
         case _ =>
       }
-    if (sym.is(ModuleClass))
-      kindString(sym) ~~ (nameString(sym.name.stripModuleClassSuffix) + idString(sym))
+    if (sym.is(ModuleClass)) {
+      val name =
+        if (sym.isPackageObject) sym.owner.name
+        else sym.name.stripModuleClassSuffix
```
allanrenucci: `sym.name` is `package$` for package object

Blaisorblade: Makes sense to me, but is this testable without too much effort? Otherwise I'll need to figure out an example to try this out... (that is, write a test.)

allanrenucci: I've added a test case

PR #4733, opened 2018-06-28
Add quote automatic lifting

nicolasstucki:
Allow `(x: X) => '(... x ...)` if there is an implicit `Liftable[X]`.

This approach is the same used in OCaml.

It differs with our first attempt to something similar by giving an implicit conversion from `X` to `Expr[X]` if the there was an implicit `Liftable[X]`. In that version it was to easy to list the result of a block instead of the block of code. It also requited writing `(x: X) => '(... ~x ...)`.


nicolasstucki:
This is an interesting experiment, but it makes us diverge from the theory and simplicity of the system. The syntactic overhead is acceptable and helps to understand what is happening in the code.

nicolasstucki:
The current version uses `(x: X) => '(... ~x.toExpr ...)  `.

PR #4736, opened 2018-06-28
Add tests example of a partially unrolled loop

nicolasstucki:


tests/run/quote-unrolled-foreach/quoted_2.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.quoted._
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    val arr = Array.tabulate[Int](21)(x => x)
+    Macro.unrolledForeach(3, arr) { (x: Int) =>
+      System.out.println(2 * x)
+    }
+
+    /* unrooled code:
```
smarter: You can add a test to https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/InlineBytecodeTests.scala to check that the code produced by the compiler actually is what you think it is.

nicolasstucki: I don't thinks  `InlineBytecodeTests.scala` supports separate compilation. Not sure how much work it would be to make it support it. It is probably better to just test the code directly.

smarter: What do you mean by "test the code directly"? You're doing code generation here, if you don't test that the code you generate looks like the way you think it should look like, I can guarantee you that you will have hard-to-find bugs.

nicolasstucki: I mean test the high-level code generated instead of the final bytecode.

PR #4743, opened 2018-06-29
Resolve SeqType based on the alias defined in scala package object

allanrenucci:
The goal of this PR is to be able to build the 2.13 standard library while still supporting scala 2.12


allanrenucci:
Blocked on #4758

allanrenucci:
Superseded by #4797. I couldn't solve the cyclic reference errors so I decided to go with a compiler option instead

PR #4744, opened 2018-06-29
Simplify output directory handling

smarter:
Store the outputDir setting as an AbstractFile instead of a String, this
way we can use a virtual output directory in the REPL just by changing
the setting instead of having to subclass GenBCode.

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/GenBCode.scala
+++ b/compiler/src/dotty/tools/backend/jvm/GenBCode.scala
@@ -41,13 +41,9 @@ class GenBCode extends Phase {
 
   private[this] var myOutput: AbstractFile = _
 
-  protected def outputDir(implicit ctx: Context): AbstractFile = {
-    if (myOutput eq null) {
-      val path = Directory(ctx.settings.outputDir.value)
-      myOutput =
-        if (path.extension == "jar") JarArchive.create(path)
-        else new PlainDirectory(path)
-    }
+  private def outputDir(implicit ctx: Context): AbstractFile = {
+    if (myOutput eq null)
+      myOutput = ctx.settings.outputDir.value
     myOutput
   }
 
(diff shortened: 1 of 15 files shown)


compiler/src/dotty/tools/repl/Rendering.scala
```diff
@@ -37,7 +37,7 @@ private[repl] class Rendering(compiler: ReplCompiler,
         new java.net.URLClassLoader(compilerClasspath.toArray, classOf[ReplDriver].getClassLoader)
       }
 
-      myClassLoader = new AbstractFileClassLoader(compiler.directory, parent)
```
allanrenucci: `compiler` is now unused

compiler/src/dotty/tools/backend/jvm/GenBCode.scala
```diff
@@ -42,12 +42,8 @@ class GenBCode extends Phase {
   private[this] var myOutput: AbstractFile = _
 
   protected def outputDir(implicit ctx: Context): AbstractFile = {
```
allanrenucci: Can this be `private` now?

compiler/src/dotty/tools/dotc/config/Settings.scala
```diff
@@ -140,14 +141,14 @@ object Settings {
...
-              update(arg, args)
-            case _ =>
-              fail(s"'$arg' does not exist or is not a directory", args)
+        case (OutputTag, arg :: args) =>
+          val path = Directory(arg)
+          val isJar = path.extension == "jar"
+          if (!isJar && !path.isDirectory)
+            fail(s"'$arg' does not exist or is not a directory or .jar file", args)
+          else {
+            val output = if (isJar) JarArchive.create(path) else new PlainDirectory(path)
```
allanrenucci: We will now hold on to an open jar during the whole compiler run instead of just backend. We didn't even open the jar before if we stopped before backend

smarter: It's an empty open jar, so not a big deal I think (it cannot be deleted on Windows while it's open, but who cares?)

PR #4750, opened 2018-07-01
Only allow access to inline args at level 1

nicolasstucki:
If it is needed at level 1 it can be explicitly lifted.
This simplifies the rules that users need to know and
the implementation of the phase.


Blaisorblade:
That's the change we discussed time ago right? 👍 from me.

nicolasstucki:
Not exactly what we discussed, but it is related. I still want to do the one we discussed but it is a discussion for another PR.

nicolasstucki:
Thinking about it, the simplification I wanted to make only works for `inline` parameters, not for  `inline val`s. Though this experiment showed me some vulnerability in the implementation of `liftValue`. I will transfor this PR into a fix for that.

PR #4755, opened 2018-07-02
Introduce `@forceInline` annotation

### LINKED ISSUE #4737: Unable to inline tailrec methods
```scala
import annotation.tailrec
class Test {
  @inline @tailrec def rec(x: Int): Int = {
    if (x > 0) rec(x - 1)
    else 0
  }

  def bar = rec(100)
}
```

fails to compile with Dotty
```scala
-- Error: tests/allan/Test.scala:25:18 -----------------------------------------
25 |    if (x > 0) rec(x - 1)
   |               ^^^^^^^^^^
   |               Maximal number of successive inlines (32) exceeded,
   |               Maybe this is caused by a recursive inline method?
   |               You can use -Xmax:inlines to change the limit.
   | This location is in code that was inlined at tests/allan/Test.scala:25
```
Scalac does not have this limitation because inlining happens after the tailrec optimisation.

I believe this change of behavior might break a lot of code, and we should turn the error into a warning.

allanrenucci:
An annotation on methods that is equivalent to Dotty `inline` modifier.
The annotation should be used instead of the `inline` modifier in code
that needs to cross compile between Scala 2 and Dotty.

Fix #4737

### DIFF
--- a/library/src/scala/forceInline.scala
+++ b/library/src/scala/forceInline.scala
@@ -1,0 +1,16 @@
+package scala
+
+/** An annotation on methods that is equivalent to Dotty `inline` modifier.
+ *
+ *  The annotation should be used instead of the `inline` modifier in code
+ *  that needs to cross compile between Scala 2 and Dotty.
+ *
+ *  Note that Scala 2 ignores the `@forceInLine` annotation, and one must use
+ *  both the `@inline` and `@forceInline` annotation to inline across the
+ *  two compilers. E.g.
+ *
+ *  ```scala
+ *  @inline @forceInline def foo = ...
+ *  ```
+ */
+class forceInline extends scala.annotation.StaticAnnotation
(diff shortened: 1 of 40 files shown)


smarter:
On second thoughts, maybe we should keep this in the scala package like the inline annotation for now to avoid confusion. Eventually we'll remove this annotation anyway so it doesn't matter much.

library/src/scala/forceInline.scala
```diff
@@ -0,0 +1,8 @@
+package scala
+
+/** An annotation on methods that is equivalent to Dotty `inline` modifier.
+ *
+ *  The annotation should be used instead of the `inline` modifier in code
+ *  that needs to cross compile between Scala 2 and Dotty.
```
nicolasstucki: You should also mention that `@forceInline` will be ignored by Scala 2. Maybe hint at the need of `@forceInline @inline` for cross compilation with the optimization in both versions.

PR #4759, opened 2018-07-04
Improve usability of erased values

odersky:
Erased values were not very useful so far:

 - we could not instantiate them with anything interesting without triggering a warning.
 - we could not pull out a type from them without getting an error.

Both impediments are fixed in this PR. This is an alternate fix for #4060.


Blaisorblade:
👍 from me (a posteriori). I thought I had found an escape, but it seems this check is even more robust than expected. However, the fact that `cast1` is unsafe still makes this pretty fragile to work with.

```scala
abstract class <:<[A, B] {
  type T >: A <: B
}
def cast1[A, B](x: A)(erased y: A <:< B): B = x.asInstanceOf[B] //should be safe, but isn't
def cast2[A, B](x: A)(erased y: A <:< B): B = x: y.T // error

cast1[Int, String](1)(???) // crashes
cast2[Int, String](1)(???) // error
```

PR #4760, opened 2018-07-04
Fix #4754: Don't generate inline accessor for constants

### LINKED ISSUE #4754: Crash: Inlining private definition from companion object
```scala
object Foo {
  private final val C = 1
}

class Foo {
  import Foo._
  inline def foo(x: Int): Boolean = x == C
}
```
This code snippet crashes the compiler

```scala
dotc -d out tests/allan/Foo.scala
exception occurred while compiling Foo.scala
Exception in thread "main" java.lang.AssertionError: assertion failed: unresolved symbols: method inline$C in object Foo when pickling Foo.scala
	at dotty.tools.dotc.core.tasty.TreePickler.pickle(TreePickler.scala:642)
	at dotty.tools.dotc.transform.Pickler.$anonfun$run$3(Pickler.scala:56)
	at dotty.tools.dotc.transform.Pickler.$anonfun$run$3$adapted(Pickler.scala:49)
	at dotty.tools.dotc.transform.Pickler.$anonfun$run$2(Pickler.scala:49)
	at dotty.tools.dotc.transform.Pickler.$anonfun$run$2$adapted(Pickler.scala:48)
	at dotty.tools.dotc.transform.Pickler.run(Pickler.scala:48)
	at dotty.tools.dotc.core.Phases$Phase.$anonfun$runOn$1(Phases.scala:298)
	at scala.collection.immutable.List.map(List.scala:283)
	at dotty.tools.dotc.core.Phases$Phase.runOn(Phases.scala:296)
	at dotty.tools.dotc.core.Phases$Phase.runOn$(Phases.scala:295)
	at dotty.tools.dotc.transform.Pickler.runOn(Pickler.scala:83)
	at dotty.tools.dotc.Run.$anonfun$compileUnits$3(Run.scala:174)
...
```

allanrenucci:


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -48,13 +48,16 @@ object Inliner {
...
+       *  by excluding all symbols properly contained in the inlined method.
+       *
+       *  Constants don't need accessors since they are inlined in FirstTransform.
+       */
       def needsAccessor(sym: Symbol)(implicit ctx: Context) =
         sym.isTerm &&
         (sym.is(AccessFlags) || sym.privateWithin.exists) &&
-        !sym.isContainedIn(inlineSym)
+        !sym.isContainedIn(inlineSym) &&
+        !sym.info.widenTermRefExpr.isInstanceOf[ConstantType]
```
smarter: Tou also need to check `isIdempotentExpr` like we do in `TreeInfo#constToLiteral` which is called from `FirstTransform`. Alternatively you could just check `is(Final)` I think.

tests/pos/i4754.scala
```diff
@@ -0,0 +1,12 @@
...
+  private final val C = "Constant"
+}
+
+class Foo {
+  import Foo._
+  inline def foo(x: Int): Boolean = x == C.length
+}
+
+class Test {
+  (new Foo).foo(1)
```
smarter: This should be a run test that checks that no accessor method is present in the class using java reflection.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
smarter: Right so I'd check is(Final)

allanrenucci: Can you elaborate why `is(Final)`?
```scala
private final val x: Int = 1 // won't be inlined by FirstTransform, so need accessor
private final def y: Int = 1 // won't be inlined either, need accesor
```

odersky: I think the right test is `
```scala
!(sym.isStable && sym.info.widenTermRefExpr.isInstanceOf[ConstantType])
```
That excludes `defs` with constant result. These have to be called anyway for their side effects.

allanrenucci:
This PR doesn't solve the underlying issue: inline accessor are not generated for private definition in companion objects.

However once the issue is fixed, it could be added as an optimisation since accessors should not be needed for constant vals

PR #4761, opened 2018-07-04
repl: don't blink the opening paren after typing a closing paren

smarter:


compiler/src/dotty/tools/repl/JLineTerminal.scala
```diff
@@ -50,11 +50,12 @@ final class JLineTerminal extends java.io.Closeable {
...
       .parser(new Parser)
-      .variable(SECONDARY_PROMPT_PATTERN, "%M") // A short word explaining what is "missing".
-                                                // This is supplied from the EOFError.getMissing() method
-      .variable(LIST_MAX, 400)       // ask user when number of completions exceed this limit (default is 100)
-      .option(INSERT_TAB, true)      // at the beginning of the line, insert tab instead of completing
-      .option(AUTO_FRESH_LINE, true) // if not at start of line before prompt, move to new line
+      .variable(SECONDARY_PROMPT_PATTERN, "%M") // A short word explaining what is "missing",
+                                                // this is supplied from the EOFError.getMissing() method
+      .variable(LIST_MAX, 400)                  // Ask user when number of completions exceed this limit (default is 100).
+      .variable(BLINK_MATCHING_PAREN, 0)        // Don't blink the opening paren after typing a closing paren.
```
allanrenucci: Looking at the [implementation](https://github.com/jline/jline3/blob/b5a643de48f701a8088bb0e1b0cfaf4a52dde068/reader/src/main/java/org/jline/reader/impl/LineReaderImpl.java#L1947), it should be a `Long`

Blaisorblade:
Why? Just tried this out, looks potentially useful to me.

smarter:
I find it distracting, it's also a change from the existing behavior. We may want to expose the JLine configuration variables in a configuration file somewhere to let people customize whatever they want.

smarter:
A better solution I think would be to do what text editors usually do and highlight both parens when the cursor is on one of the paren, we should be able to do that based on the information we have after parsing.

PR #4766, opened 2018-07-05
Fix #4765, evaluate objects we inline from if not pure

### LINKED ISSUE #4765: Inliner can remove side effects from objects
Noticed while discussing #4763, [haven't found in the tracker](https://github.com/lampepfl/dotty/issues?utf8=%E2%9C%93&q=inliner+side+effects+). Might be a known bug (related to what @allanrenucci noticed recently with constant folding accepting idempotent rather than pure methods), but we should have an issue for this.

```
sbt:dotty> repl -Xprint:frontend
object A {
  println("object init");
  inline def x: 1 = { println("x init") ; 1 } //`inline val` requires a val on the rhs
}
scala> object A {
           println("object init");
           inline def x: 1 = { println("x init") ; 1 } //`inline val` requires a val on the rhs
         }

result of rs$line$1 after frontend:
package <empty> {
  final lazy module val rs$line$1: rs$line$1$ = new rs$line$1$()
  final module class rs$line$1$() extends Object() { this: rs$line$1.type =>
    final lazy module val A: A$ = new A$()
    final module class A$() extends Object() { this: A.type =>
      println("object init")
      inline def x: 1.type =
        {
          println("x init")
          1
        }
    }
  }
}
// defined object A

scala>

scala> object B {
           val y = A.x
         }
result of rs$line$2 after frontend:
package <empty> {
  final lazy module val rs$line$2: rs$line$2$ = new rs$line$2$()
  final module class rs$line$2$() extends Object() { this: rs$line$2.type =>
    final lazy module val B: B$ = new B$()
    final module class B$() extends Object() { this: B.type =>
      val y: Int =
        /* inlined from A.x*/
          {
            {
              println("x init")
              1
            }
          }
    }
  }
}
// defined object B
```

Blaisorblade:
Fix #4765: when inlining `A.x`, produce `A` and then `x`'s body.
Missing: tests — but first let's figure out if we want to do this change.

Here's the "repl-test" from #4765 — it turns `A.x` into `{ val A$_this: A.type(A) = A; { println("x init"); 1 } }` rather than `{ println("x init"); 1 }`.

```scala
sbt:dotty> repl -Xprint:frontend
[...]
[info] Running (fork) dotty.tools.repl.Main -classpath /Users/pgiarrusso/git/dotty/library/target/scala-2.12/dotty-library_2.12-0.9.0-bin-SNAPSHOT-nonbootstrapped.jar -Xprint:frontend
scala> object A {
     |   println("object init");
     |   inline def x: 1 = { println("x init") ; 1 } //`inline val` requires a val on the rhs
     | }
result of rs$line$1 after frontend:
package <empty> {
  final lazy module val rs$line$1: rs$line$1$ = new rs$line$1$()
  final module class rs$line$1$() extends Object() { this: rs$line$1.type =>
    final lazy module val A: A$ = new A$()
    final module class A$() extends Object() { this: A.type =>
      println("object init")
      inline def x: 1.type =
        {
          println("x init")
          1
        }
    }
  }
}
// defined object A

scala> object B {
     |            val y = A.x
     |          }
result of rs$line$2 after frontend:
package <empty> {
  final lazy module val rs$line$2: rs$line$2$ = new rs$line$2$()
  final module class rs$line$2$() extends Object() { this: rs$line$2.type =>
    final lazy module val B: B$ = new B$()
    final module class B$() extends Object() { this: B.type =>
      val y: Int =
        /* inlined from A.x*/
          {
            val A$_this: A.type(A) = A
            {
              println("x init")
              1
            }
          }
    }
  }
}
// defined object B
```

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Inliner.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inliner.scala
@@ -420,7 +420,7 @@ class Inliner(call: tpd.Tree, rhsToInline: tpd.Tree)(implicit ctx: Context) {
   /** The Inlined node representing the inlined call */
   def inlined(pt: Type) = {
     // make sure prefix is executed if it is impure
-    if (!isIdempotentExpr(prefix)) registerType(meth.owner.thisType)
+    if (!isPureExpr(prefix)) registerType(meth.owner.thisType)
 
     // Register types of all leaves of inlined body so that the `paramProxy` and `thisProxy` maps are defined.
     rhsToInline.foreachSubTree(registerLeaf)


odersky:
I think we should leave this as is for the moment. The problem is that, without a good purity analysis of objects we end up referring to way too many objects. The inliner's job is to produce fast code, and it can't do that if we fix this issue. 

I believe the correct approach is to come up with a better purity analysis of objects. #4710 is a start. If we have that, we can switch to `isImpure` but not before.

So my suggestion: 

 1. Get #4710 in.
 2. Think about possible ways to make the analysis more precise.
 3. Flip the switch to elide prefixes only if pure (instead of idempotent) while at the same
     time carefully evaluating how it affects performance. We should do both benchmarking
     and perform counts how many prefixes are then newly evaluated and get an idea how
     many false negatives we get in practice.

odersky:
If it's stat:revisit (which is the right label, I agree), then we should close for now.

PR #4769, opened 2018-07-06
fix cygwin which has problem with unix pipes

liufengyun:
fix cygwin which has problem with unix pipes

tested in https://github.com/lampepfl/packtest/pull/16

### DIFF
--- a/dist/bin/common
+++ b/dist/bin/common
@@ -119,7 +119,9 @@ SCALA_ASM=$(find_lib "*scala-asm*")
 SCALA_LIB=$(find_lib "*scala-library*")
 SCALA_XML=$(find_lib "*scala-xml*")
 SBT_INTF=$(find_lib "*compiler-interface*")
-JLINE=$(find_lib "*jline*")
+JLINE=$(find_lib "*jline-3*")
+JLINE_TERMINAL=$(find_lib "*jline-terminal-3*")
+JLINE_TERMINAL_JNA=$(find_lib "*jline-terminal-jna*")
 
 # debug
 DEBUG_STR=-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005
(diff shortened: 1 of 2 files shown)


dist/bin/dotc
```diff
@@ -65,9 +65,10 @@ classpathArgs () {
   toolchain+="$DOTTY_LIB$PSEP"
   toolchain+="$DOTTY_COMP$PSEP"
 
-  # JLine is mutiple jars
-  local all_jline=$(echo "$JLINE" | tr " " "$PSEP")
-  toolchain+="$all_jline"
+  # jine
+  toolchain+="$JLINE$PSEP"
+  toolchain+="$JLINE_TERMINAL$PSEP"
+  toolchain+="$JLINE_TERMINAL_JNA$PSEP"
```
allanrenucci: You don't need `PSEP` here at the end

dist/bin/common
```diff
@@ -119,7 +119,9 @@ SCALA_ASM=$(find_lib "*scala-asm*")
 SCALA_LIB=$(find_lib "*scala-library*")
 SCALA_XML=$(find_lib "*scala-xml*")
 SBT_INTF=$(find_lib "*compiler-interface*")
-JLINE=$(find_lib "*jline*")
+JLINE=$(find_lib "*jline-3*")
+JLINE_TERMINAL=$(find_lib "*jline-terminal-3*")
+JLINE_TERMINAL_JNA=$(find_lib "*jline-terminal-jna*")
```
allanrenucci: Let's add `-3` here at the end for consistency (even if not needed)

liufengyun:
@Blaisorblade Yes, it's a conjecture that the problem is caused by the usage of pipes. However, debugging windows is painful, so we think it's better to just avoid the problem.

PR #4772, opened 2018-07-06
Ignore implicit shortcuts in RefChecks

odersky:
The Note in the doc comment for class ImplicitShortcuts explains why
this is necessary.

Also, need to special case bridge generation to handle implicit shortcuts correctly.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
@@ -2,7 +2,7 @@ package dotty.tools.dotc
 package transform
 
 import MegaPhase._
-import core.DenotTransformers.IdentityDenotTransformer
+import core.DenotTransformers.{DenotTransformer, IdentityDenotTransformer}
 import core.Symbols._
 import core.Contexts._
 import core.Types._
@@ -44,8 +44,20 @@ import collection.mutable
  *  (2) A reference `qual.apply` where `qual` has implicit function type and
  *  `qual` refers to a method `m` is rewritten to a reference to `m$direct`,
  *  keeping the same type and value arguments as they are found in `qual`.
+ *
+ *  Note: The phase adds direct methods for all defined methods with IFT results,
+ *        as well as for all methods that are referenced. It does NOT do an
+ *        info transformer that adds these methods everywhere where an IFT returning
+ *        method exists.
+ *        Adding such an info transformer is impractical because it would mean
+ *        that we have to force the types of all members of classes that are referenced.
+ *        But not adding an info transformer can lead to inconsistencies in RefChecks.
+ *        We solve that by ignoring direct methods in Refhecks.
+ *        Another, related issue is bridge generation, where we also generate
+ *        shortcut methods on the fly.
  */
 class ShortcutImplicits extends MiniPhase with IdentityDenotTransformer { thisPhase =>
+  import ShortcutImplicits._
   import tpd._
 
   override def phaseName: String = ShortcutImplicits.name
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
```diff
@@ -61,82 +73,42 @@ class ShortcutImplicits extends MiniPhase with IdentityDenotTransformer { thisPh
...
-        shouldBeSpecialized(tree.qualifier.symbol)) {
+        needsImplicitShortcut(tree.qualifier.symbol)) {
       def directQual(tree: Tree): Tree = tree match {
         case Apply(fn, args)     => cpy.Apply(tree)(directQual(fn), args)
         case TypeApply(fn, args) => cpy.TypeApply(tree)(directQual(fn), args)
         case Block(stats, expr)  => cpy.Block(tree)(stats, directQual(expr))
         case tree: RefTree =>
+          def rewire(tp: Type, sym: Symbol) = tp match {
+            case tp: NamedType => tp.prefix.select(sym)
+            case _ => sym.termRef
```
liufengyun: Could you explain when the last case is reachable?

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -446,9 +447,12 @@ object RefChecks {
           }
 
       def ignoreDeferred(member: SingleDenotation) =
-        member.isType ||
-          member.symbol.isSuperAccessor || // not yet synthesized
-          member.symbol.is(JavaDefined) && hasJavaErasedOverriding(member.symbol)
+        member.isType || {
+          val mbr = member.symbol
+          mbr.isSuperAccessor || // not yet synthesized
+          ShortcutImplicits.isImplicitShortcut(mbr) || // only synthesized when referenced
```
liufengyun: The comment "only synthesized when referenced" is confusing, if I understand correctly direct method _trees_  are always synthesized.

compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
```diff
@@ -44,8 +44,20 @@ import collection.mutable
  *  (2) A reference `qual.apply` where `qual` has implicit function type and
  *  `qual` refers to a method `m` is rewritten to a reference to `m$direct`,
  *  keeping the same type and value arguments as they are found in `qual`.
+ *
+ *  Note: The phase adds direct methods for all defined methods with IFT results,
+ *        as well as for all methods that are referenced. It does NOT do an
+ *        info transformer that adds these methods everywhere where an IFT returning
+ *        method exists.
+ *        Adding such an info transformer is impractical because it would mean
+ *        that we have to force the types of all members of classes that are referenced.
```
liufengyun: I don't quite understand how info transformer works. Maybe a future compiler session on the topic, I know several in the team are interested in the topic.

compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
```diff
@@ -44,8 +44,20 @@ import collection.mutable
  *  (2) A reference `qual.apply` where `qual` has implicit function type and
  *  `qual` refers to a method `m` is rewritten to a reference to `m$direct`,
  *  keeping the same type and value arguments as they are found in `qual`.
+ *
+ *  Note: The phase adds direct methods for all defined methods with IFT results,
+ *        as well as for all methods that are referenced. It does NOT do an
```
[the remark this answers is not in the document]
odersky: I'll try to express this clearer.

compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
odersky: I'll try to eliminate the last case. It's better to fail fast if we misunderstand something.

PR #4775, opened 2018-07-07
Fix #4774: Skip TypeVars and skip if level is equal

### LINKED ISSUE #4774: Type splices containing TypeVars not handled
```scala
object Test {
  def loop[T](x: Expr[T])(implicit t: Type[T]): Expr[T] =
    '{ val y: ~t = ~x; ~loop('(y)) }
}
```
crashes with 
```
Exception in thread "main" java.lang.ClassCastException: dotty.tools.dotc.core.Types$TypeVar cannot be cast to dotty.tools.dotc.core.Types$TypeRef
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.$anonfun$transform$2(ReifyQuotes.scala:570)
	at dotty.tools.dotc.reporting.trace$.apply(trace.scala:40)
	at dotty.tools.dotc.reporting.trace$.apply(trace.scala:48)
	at dotty.tools.dotc.transform.ReifyQuotes$Reifier.transform(ReifyQuotes.scala:558)
	at dotty.tools.dotc.ast.Trees$Instance$TreeMap.$anonfun$transform$6(Trees.scala:1256)
	at dotty.tools.dotc.ast.Trees$Instance$TreeMap.transform(Trees.scala:1256)
	at dotty.tools.dotc.ast.Trees$Instance$TreeMap.transform(Trees.scala:1170)
	at dotty.tools.dotc.transform.MacroTransform$Transformer.transform(MacroTransform.scala:63)
	at dotty.tools.dotc.transform.MacroTransformWithImplicits$ImplicitsTransformer.transform(MacroTransformWithImplicits.scala:86)
...
```

nicolasstucki:
If we have ~t for `t: Type[T]` used at the same level it is an alias for T.
We disallow it if it is explicit because of PCP but if it is infered it is fine.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
@@ -567,8 +567,9 @@ class ReifyQuotes extends MacroTransformWithImplicits with InfoTransformer {
           case Quoted(quotedTree) =>
             quotation(quotedTree, tree)
           case tree: TypeTree if tree.tpe.typeSymbol.isSplice =>
-            val splicedType = tree.tpe.asInstanceOf[TypeRef].prefix.termSymbol
-            splice(ref(splicedType).select(tpnme.UNARY_~))
+            val splicedType = tree.tpe.stripTypeVar.asInstanceOf[TypeRef].prefix.termSymbol
+            if (levelOf.get(splicedType).contains(level)) tree
+            else splice(ref(splicedType).select(tpnme.UNARY_~))
           case tree: Select if tree.symbol.isSplice =>
             splice(tree)
           case tree: RefTree if isCaptured(tree.symbol, level) =>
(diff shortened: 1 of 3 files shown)


nicolasstucki:
Oh, in this context it can be written by hand. The user can write `loop(...)` or `loop[T](...)`. In the example, the former gets inferred to `loop[~t](...)` where `~t` is an alias of `T`, which is fine, we can assume it is just a `T`. We disallow the `~t` as it would not be syntactically PCP.

Blaisorblade:
@nicolasstucki As I mentioned earlier, I don't believe the PCP actually works until you can explain how it should apply to types without needing all these exceptions.

nicolasstucki:
This is just an implementation detail that I missed.

smarter:
> ~t is an alias of T, which is fine, we can assume it is just a T

I see, that seems OK, but it might be problematic if the `~t` leaks into error messages shown to the user

smarter:
It could also leak when decompiling from Tasty, or in some other ways (e.g. if we had a mode to the IDE to desugar/type-infer an expression).

nicolasstucki:
Good point. Then the type should be dealiased and replaced.

PR #4776, opened 2018-07-07
Fix #4758: Revert "New phase for entering annotations"

### LINKED ISSUE #4758: Cyclic error involving annotation
In a file `Test1.scala`:
```scala
package foo.bar

class Bar extends annotation.StaticAnnotation
```
In a file `Test2.scala`:
```scala
package foo

import foo.bar.Bar

@Bar
class Foo extends annotation.StaticAnnotation
```
In a file `Test3.scala`:
```scala
package object foo {
  @Foo class Hello
}
```
```scala
dotc Test1.scala Test2.scala Test3.scala
-- Warning: Test2.scala:5:1 ----------------------------------------
5 |@Bar
  | ^^^
  | cyclic import foo.bar.Bar, ignored
-- [E006] Unbound Identifier Error: Test2.scala:5:1 ----------------
5 |@Bar
  | ^^^
  | not found: type Bar
```

odersky:
This reverts commit 2b12868070be50fb70c687bcd8a415acbe398e3e.

Since we do not plan to run annotation macros during Namer, there's no need
to annotate before completing. This allows again to define an annotation in the
same scope as an annotated definition. So i3702.scala goes from neg to pos.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Namer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Namer.scala
@@ -687,90 +687,14 @@ class Namer { typer: Typer =>
     ctxWithStats
   }
 
-  /** Add all annotations of definitions in `stats` to the defined symbols */
-  def annotate(stats: List[Tree])(implicit ctx: Context): Unit = {
-    def recur(stat: Tree): Unit = stat match {
-      case pcl: PackageDef =>
-        annotate(pcl.stats)
-      case stat: untpd.MemberDef =>
-        stat.getAttachment(SymOfTree) match {
-          case Some(sym) =>
-            sym.infoOrCompleter match {
-              case info: Completer if !defn.isPredefClass(sym.owner) =>
-                // Annotate Predef methods only when they are completed;
-                // This is necessary to break a cyclic dependence between `Predef`
-                // and `deprecated` in test `compileStdLib`.
-                addAnnotations(sym, stat)(info.creationContext)
-              case _ =>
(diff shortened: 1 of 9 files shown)


compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -822,6 +746,22 @@ class Namer { typer: Typer =>
       else completeInCreationContext(denot)
     }
 
+    protected def addAnnotations(sym: Symbol): Unit = original match {
+      case original: untpd.MemberDef =>
+        var hasInlineAnnot = false
```
smarter: Dead variable.

compiler/src/dotty/tools/dotc/typer/Namer.scala
```diff
@@ -822,6 +746,22 @@ class Namer { typer: Typer =>
...
+        lazy val annotCtx =
```
smarter: While you're at it, you may want to make an annotCtx def that can be used both here and for https://github.com/lampepfl/dotty/blob/64bd0fe5210cacd7c17ab285c80712944157c212/compiler/src/dotty/tools/dotc/typer/Typer.scala#L1345 :)

allanrenucci:
Should we also revert #1249?

odersky:
> I think this needs to be implicit ctx => typedAnnotation(annotTree)(annotContext(original, sym))

The last commit caused double definitions of the Annotations class, so the context was too far out. I don't have the time to dig deeper now, so for the moment I am reverting.

PR #4777, opened 2018-07-07
Fix inline parameters lifted for 0 to 1

nicolasstucki:
In theory the inline parameters are available in level 1,
but we can access it "unlifted" as a value in level 0.
To only pass the parameter once it is passed as a value and lifted
using the standard lifter for the values.

Unsing lifter is just an implementation detail and should not be observable
by the user. The issue was that the lifter used could be replaced or unimported
which yield unexpected error messages.


nicolasstucki:
I originally implemented it with the implicit search thinking it would simplify the implementation loginc. Clearly not the case.

PR #4779, opened 2018-07-08
Fix #4732: Root the desugared call to StringContext

### LINKED ISSUE #4732: Failing to resolve to correct StringContext
```scala
object StringContext {
  s"Hello"
}
```
fails to compile with Dotty and Scalac:
```scala
-- [E050] Reference Error: tests/allan/Test.scala:49:3 -------------------------
49 |  s"Hello"
   |   ^
   |   object StringContext does not take parameters
```

This is because the code snippet desugars to
```scala
StringContext("Hello").s()
```

Is this on purpose? An easy fix would be to desugar to `scala.StringContext("Hello").s()`

lloydmeta:
Closes #4732

* Make desugared `s""` call `_root_.scala.StringContext` instead of just
  `StringContext`
* Adds a test


compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1109,8 +1109,13 @@ object desugar {
...
         }
-
-        Apply(Select(Apply(Ident(nme.StringContext), strs), id), elems)
+        // This is a deliberate departure from scalac, where StringContext is not rooted (See #4732)
+        val rootedStringContext = {
+          val rootPkg = Ident(nme.ROOTPKG)
+          val scalaPkg = Select(rootPkg, nme.scala_)
+          Select(scalaPkg, nme.StringContext)
+        }
+        Apply(Select(Apply(rootedStringContext, strs), id), elems)
```
nicolasstucki: Maybe `rootedStringContext` can be replaced with `ref(defn.StringContextModule)`

allanrenucci: No, Desugar produces untyped trees

nicolasstucki: Oh, I missed that. Indore my comment

compiler/src/dotty/tools/dotc/ast/Desugar.scala
```diff
@@ -1109,8 +1109,13 @@ object desugar {
           case Block(Nil, expr) => expr // important for interpolated string as patterns, see i1773.scala
           case t => t
         }
-
-        Apply(Select(Apply(Ident(nme.StringContext), strs), id), elems)
+        // This is a deliberate departure from scalac, where StringContext is not rooted (See #4732)
+        val rootedStringContext = {
+          val rootPkg = Ident(nme.ROOTPKG)
+          val scalaPkg = Select(rootPkg, nme.scala_)
+          Select(scalaPkg, nme.StringContext)
```
smarter: As a shortcut, you can use scalaDot(nme.StringContext)

compiler/src/dotty/tools/dotc/ast/Desugar.scala
allanrenucci: As suggested by @smarter, you can rewrite as
```scala
Apply(Select(Apply(scalaDot(nme.StringContext), strs), id), elems)
```

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
```diff
@@ -36,7 +36,7 @@ class StringInterpolatorOpt extends MiniPhase {
     def unapply(tree: Select)(implicit ctx: Context): Boolean = {
       tree.symbol.eq(defn.StringContextModule_apply) && {
         val qualifier = tree.qualifier
-        qualifier.isInstanceOf[Ident] && qualifier.symbol.eq(defn.StringContextModule)
+        qualifier.isInstanceOf[Select] && qualifier.symbol.eq(defn.StringContextModule)
```
allanrenucci: Thinking about it, the first test might not be needed. The method could be rewritten as:
```scala
def unapply(tree: Select)(implicit ctx: Context): Boolean =
  tree.symbol.eq(defn.StringContextModule_apply) &&
  tree.qualifier.symbol.eq(defn.StringContextModule)
```

tests/run/rooted_stringcontext.scala
```diff
@@ -0,0 +1,10 @@
+object StringContext {
+  val i = 42
+  val s = s"Answer: $i"
+}
+
+object Test {
+  def main(args: Array[String]): Unit = {
+    println(StringContext.s)
```
allanrenucci: I would remove the check file and simply do `assert(StringContext.s == "Answer: 42")`

compiler/src/dotty/tools/dotc/ast/Desugar.scala
lloydmeta: Good idea. Done.

compiler/src/dotty/tools/dotc/transform/localopt/StringInterpolatorOpt.scala
lloydmeta: Indeed, it seemed a bit fragile. Removed.

tests/run/rooted_stringcontext.scala
lloydmeta: Done.

lloydmeta:
@allanrenucci Just addressed all your feedback + @smarter 's suggestions; please have another look when you can 😄

PR #4780, opened 2018-07-09
Disallow inferred splices and set the correct position

nicolasstucki:


nicolasstucki:
Addressing comments in #4775

tests/neg/i4774a.scala
```diff
@@ -0,0 +1,11 @@
+
+import scala.quoted._
+
+object Test {
+  def loop[T](x: Expr[T])(implicit t: Type[T]): Expr[T] = '{
+    val y: ~t = ~x
+    ~loop( // error: inferred loop[~t] where T should be used
```
smarter: Instead of giving an error, couldn't the typechecker be smart enough to dealias?

nicolasstucki: The difficulty is that this depends on the quotation level and we do not have it at typer time.

nicolasstucki: I wouldn't worry to much about this case and encourage the use of `T` instead of `~t`.

biboudis:
Looks good!

PR #4782, opened 2018-07-10
Fix #4771: Relax overriding check for polymorphic methods

### LINKED ISSUE #4771: Error Overriding: incompatible types
```scala
sealed abstract class <:<[-From, +To] {
  def substituteCo[F[+_]](ff: F[From]): F[To]
}

sealed abstract class =:=[From, To] extends (From <:< To) {
  override def substituteCo[F[_]](ff: F[From]): F[To] = ???
}
```
```scala
dotc -d out tests/allan/Test.scala
-- Error: tests/allan/Test.scala:32:15 -----------------------------------------
32 |  override def substituteCo[F[_]](ff: F[From]): F[To] = ???
   |               ^
   |error overriding method substituteCo in class <:< of type [F <: [+_$1] => Any](ff: F[From]): F[To];
   |  method substituteCo of type [F <: [_$2] => Any](ff: F[From]): F[To] has incompatible type
one error found
```

smarter:
Just like in Scala 2, we now allow an override to widen the bounds of a
polymorphic type parameter (this is sound because parameters can vary
contravariantly). (And just like Scala 2, we continue not allowing an
overriding to widen the type of a term parameter).

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -390,33 +390,48 @@ object Denotations {
               case tp2: TypeBounds if tp2 contains tp1 => tp1
               case _ => mergeConflict(tp1, tp2, that)
             }
-          case tp1: MethodOrPoly =>
+
+          // Two remedial strategies:
+          //
+          //  1. Prefer method types over poly types. This is necessary to handle
+          //     overloaded definitions like the following
+          //
+          //        def ++ [B >: A](xs: C[B]): D[B]
+          //        def ++ (xs: C[A]): D[A]
+          //
+          //     (Code like this is found in the collection strawman)
+          //
+          // 2. In the case of two method types or two polytypes with matching
+          //    parameters and implicit status, merge corresponding parameter
+          //    and result types.
+          case tp1: MethodType =>
             tp2 match {
-              case tp2: MethodOrPoly =>
-                // Two remedial strategies:
-                //
-                //  1. Prefer method types over poly types. This is necessary to handle
-                //     overloaded definitions like the following
-                //
-                //        def ++ [B >: A](xs: C[B]): D[B]
-                //        def ++ (xs: C[A]): D[A]
-                //
-                //     (Code like this is found in the collection strawman)
-                //
-                // 2. In the case of two method types or two polytypes with matching
-                //    parameters and implicit status, merge corresppnding parameter
-                //    and result types.
-                if (tp1.isInstanceOf[PolyType] && tp2.isInstanceOf[MethodType]) tp2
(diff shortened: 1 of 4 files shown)


compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -391,29 +391,38 @@ object Denotations {
...
+            //  1. Prefer method types over poly types. This is necessary to handle
+            //     overloaded definitions like the following
+            //
+            //        def ++ [B >: A](xs: C[B]): D[B]
+            //        def ++ (xs: C[A]): D[A]
+            //
+            //     (Code like this is found in the collection strawman)
+            //
+            // 2. In the case of two method types or two polytypes with matching
+            //    parameters and implicit status, merge corresppnding parameter
```
odersky: Typo: corresppnding

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -391,29 +391,38 @@ object Denotations {
...
+            //    and result types.
+            (tp1, tp2) match {
```
odersky: This is bound to be very inefficient since we do not optimize tupled pattern matches yet. I suggest to use nested matches instead, similar to how it is done in infoJoin.

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -391,29 +391,38 @@ object Denotations {
...
+                     tp1.isImplicitMethod == tp2.isImplicitMethod =>
+                tp1.derivedLambdaType(
+                  mergeParamNames(tp1, tp2),
+                  tp1.paramInfos,
+                  infoMeet(tp1.resultType, tp2.resultType.subst(tp2, tp1)))
+              case (tp1: PolyType, tp2: PolyType)
+                  if ctx.typeComparer.matchingPolyParams(tp1, tp2) =>
+                tp1.derivedLambdaType(
+                  mergeParamNames(tp1, tp2),
+                  (tp1.paramInfos, tp2.paramInfos).zipped
```
odersky: Shorter and more efficient: use zipWithConserve (in Decorators.scala)

compiler/src/dotty/tools/dotc/core/Denotations.scala
```diff
@@ -566,13 +575,26 @@ object Denotations {
...
+            case _ =>
+              mergeConflict(tp1, tp2, that)
+          }
+        case tp1: PolyType =>
+          tp2 match {
+            case tp2: PolyType
+            if ctx.typeComparer.matchingPolyParams(tp1, tp2) =>
+              tp1.derivedLambdaType(
+                mergeParamNames(tp1, tp2),
+                (tp1.paramInfos, tp2.paramInfos).zipped
```
odersky: zipWithConserve again

PR #4787, opened 2018-07-12
Fix #4785: change typing rule for _* args

### LINKED ISSUE #4785: Arrays and Repeated arguments in 2.13 and future Dotty version
Starting with Scala 2.13 the default `Seq` type is `immutable.Seq`. This change apply to repeated arguments:
```scala
scala> def foo(x: Int*) = 1
foo: (x: Int*)Int

scala> foo(immutable.Seq(1, 2): _*)
res0: Int = 1

scala> foo(mutable.Seq(1, 2): _*)
                      ^
       error: type mismatch;
        found   : Seq[Int] (in scala.collection.mutable)
        required: Seq[Int] (in scala.collection.immutable)

scala> foo(Array(1, 2): _*)
res2: Int = 1
```
However, it is still possible to pass an Array where repeated arguments are expected. This is a requirement for java interoperability.

In Dotty, given `foo: _*`, `foo` must be a subtype of `Seq`. If `foo` is an array, then an implicit conversion is applied (i.e. `Predef.wrapXXXArray`). We don't need to special case arrays in typer but latter (in [ElimRepeated](https://github.com/lampepfl/dotty/blob/385082b2add7aabd5c2877892d78b6d40de025c2/compiler/src/dotty/tools/dotc/transform/ElimRepeated.scala#L101-L102)) we have to revert the conversion if the array is used as a repeated arguments of a java defined method. E.g.
```scala
// xs: Array[String]

java.nio.file.Paths.get("foo", xs: _*)

// >>>>>> Typer
java.nio.file.Paths.get("foo", Predef.wrapRefArray(xs): String*)

// >>>>>> ElimRepeated
java.nio.file.Paths.get("foo", xs)
```

The current implicit conversions defined in `Predef` returns a `mutable.Seq`.

How do we solve this problem in the future when Dotty switch its default `Seq` type to `immutable.Seq` instead of `collection.Seq`? Here are my thoughts, either:

1. Introduce new implicit conversions in `DottyPredef` (or `Predef`?) from `Array` to `immutable.Seq`.
2. Change typing rules for repeated arguments. In expression position:
    ```
    foo <:< Array[A] or foo <:< immutable.Seq[A]
    ---------------------------------------------
    foo: _* ⊢ Repeated[A]
    ```

I think `scalac` does 2). @adriaanm, any thoughts on this?

allanrenucci:
In expression position, if a tree is of the form `expr: _*`, try to type
`expr` as an `Array` and as a `Seq` if there are errors

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Definitions.scala
+++ b/compiler/src/dotty/tools/dotc/core/Definitions.scala
@@ -345,11 +345,6 @@ class Definitions {
     def Predef_classOf(implicit ctx: Context) = Predef_classOfR.symbol
     lazy val Predef_undefinedR = ScalaPredefModule.requiredMethodRef("???")
     def Predef_undefined(implicit ctx: Context) = Predef_undefinedR.symbol
-    // The set of all wrap{X, Ref}Array methods, where X is a value type
-    val Predef_wrapArray = new PerRun[collection.Set[Symbol]]({ implicit ctx =>
-      val methodNames = ScalaValueTypes.map(TreeGen.wrapArrayMethodName) + nme.wrapRefArray
-      methodNames.map(ScalaPredefModule.requiredMethodRef(_).symbol)
-    })
 
   lazy val ScalaRuntimeModuleRef = ctx.requiredModuleRef("scala.runtime.ScalaRunTime")
   def ScalaRuntimeModule(implicit ctx: Context) = ScalaRuntimeModuleRef.symbol
--- a/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
+++ b/compiler/src/dotty/tools/dotc/core/TypeApplications.scala
@@ -492,7 +492,7 @@ class TypeApplications(val self: Type) extends AnyVal {
   }
 
   /** The element type of a sequence or array */
-  def elemType(implicit ctx: Context): Type = self match {
+  def elemType(implicit ctx: Context): Type = self.widenDealias match {
     case defn.ArrayOf(elemtp) => elemtp
     case JavaArrayType(elemtp) => elemtp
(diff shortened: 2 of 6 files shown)


smarter:
FWIW, the following code used to find the `seq` implicit but now finds the `array` implicit (it doesn't compile at all with 2.12):
```scala
object Test {
  implicit def array: Array[Int] = Array(0)
  implicit def seq: Seq[Int] = Seq(0)

  def foo(x: Int*) = x
  def get[T](implicit x: T) = x

    println(foo(implicitly: _*))
  }
}
```
Any reason to try the Array prototype first? Most of the time I expect users would pass a Seq, so that would be a lot of unnecessary retyping. I also wonder if you could use a union of Array or Seq as the prototype.

smarter:
Wait, doesn't https://github.com/lampepfl/dotty/issues/4785 imply that this implicit conversion won't exist with 2.13, thus solving the problem?

allanrenucci:
No, still [there](https://github.com/scala/scala/blob/2465fce45bcc719d156fcded097db91607da5944/src/library/scala/Predef.scala#L779-L802).

allanrenucci:
Anyway, relying on an implicit conversion defined in the standard library to typecheck calls to Java methods with repeated arguments seems like a very bad idea to me. What if I compile with `-Yno-imports`?

odersky:
On further discussion, I think it's preferable to have automatic conversions for varargs. We already have a conversion from Seq to Array when passing a Seq argument to a Java vararg. We should have a dual conversion from Array to Seq when passing an Array argument to a Scala vararg. I believe the Array-> Seq conversion should make a defensive copy of the array in order to prevent mutations leaking into the vararg.

allanrenucci:
Superseded by #5577

PR #4789, opened 2018-07-13
Support `-from-tasty` in Dottydoc

Duhemm:
It's not yet integrated with the `doc` task in sbt, but I think this can be reviewed independently first.


compiler/test/dotty/tools/IOUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.collection.JavaConverters.asScalaIteratorConverter
+
+object IOUtils {
+
+  def inTempDirectory[T](fn: Path => T): T = {
+    val temp = Files.createTempDirectory("temp")
+    try fn(temp)
+    finally {
+      val allFiles = getAll(temp, "glob:**").sortBy(_.toAbsolutePath.toString).reverse
+      allFiles.foreach(Files.delete(_))
```
nicolasstucki: Use `dotty.tools.io.Path.deleteRecursively`, it does it more efficiently.

compiler/test/dotty/tools/IOUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }
+
+  def getAll(base: Path, pattern: String): List[Path] = {
+    val paths = Files.walk(base)
+    val matcher = FileSystems.getDefault.getPathMatcher(pattern)
+    try paths.filter(matcher.matches).iterator().asScala.toList
+    finally paths.close()
```
nicolasstucki: `dotty.tools.io.Directory.walkFilter` does something similar.

nicolasstucki: We may not need `IOUtils` at all

doc-tool/src/dotty/tools/dottydoc/DocFrontEnd.scala
```diff
@@ -1,17 +1,44 @@
...
+ *  from TASTY. The comments will be cooked after being unpickled.
+ *
  *  It currently still throws away Java sources by overriding
  *  `discardAfterTyper`.
  */
 class DocFrontEnd extends FrontEnd {
+
+  override def runOn(units: List[CompilationUnit])(implicit ctx: Context): List[CompilationUnit] = {
+    if (ctx.settings.fromTasty.value) {
+      val fromTastyFrontend = new ReadTastyTreesFromClasses
```
nicolasstucki: A front end should probably not be in another frontend.  `DocCompiler` should choose between `DocFrontEnd` or `ReadTastyTreesFromClasses` based on `ctx.settings.fromTasty.value`

doc-tool/src/dotty/tools/dottydoc/DocFrontEnd.scala
```diff
@@ -1,17 +1,44 @@
...
+      val unpickledUnits = fromTastyFrontend.runOn(units)
+
+      val typer = new Typer()
+      if (ctx.settings.YcookComments.value) {
+        ctx.docbase.docstrings.keys.foreach { sym =>
```
nicolasstucki: Similarly, this should probably be in it's own phase. Which would be added conditionally based on `ctx.settings.YcookComments.value`.

doc-tool/src/dotty/tools/dottydoc/DocFrontEnd.scala
```diff
@@ -1,17 +1,44 @@
...
  *  `discardAfterTyper`.
  */
 class DocFrontEnd extends FrontEnd {
+
+  override def runOn(units: List[CompilationUnit])(implicit ctx: Context): List[CompilationUnit] = {
+    if (ctx.settings.fromTasty.value) {
+      val fromTastyFrontend = new ReadTastyTreesFromClasses
+      val unpickledUnits = fromTastyFrontend.runOn(units)
+
+      val typer = new Typer()
```
nicolasstucki: No need to initialize `Typer` if `ctx.settings.YcookComments.value` is false

doc-tool/src/dotty/tools/dottydoc/core/DocASTPhase.scala
```diff
@@ -27,20 +24,20 @@ class DocASTPhase extends Phase {
   def phaseName = "docASTPhase"
 
   /** Build documentation hierarchy from existing tree */
-  def collect(tree: Tree)(implicit ctx: Context): Entity = {
+  def collect(tree: Tree)(implicit ctx: Context): List[Entity] = {
```
nicolasstucki: Why did you change `Entity` to `List[Entity]`?

nicolasstucki: As far as I can see there are only lists of size 0 or 1.

Duhemm: That's to support multiple `@usecase`. See the commit messages of https://github.com/lampepfl/dotty/pull/4789/commits/c528d6ae0a552df29ce2aa131114375d0d60079b and https://github.com/lampepfl/dotty/pull/4789/commits/f1da1e155f58818615e412ea9330148ee3f3b0bc

This is where longer lists may be created: https://github.com/lampepfl/dotty/pull/4789/files#diff-5a9874880800484ed4b082c8d776c661R30

doc-tool/src/dotty/tools/dottydoc/DocFrontEnd.scala
Duhemm: I also wanted to decouple everything, but I ran into issues when cooking the comments. I'll investigate a bit more.

nicolasstucki:
test performance please

Duhemm:
@nicolasstucki I think it's ready for another pass! Thanks for taking a look.

Duhemm:
@nicolasstucki Rebased and ready for another round 😄

compiler/src/dotty/tools/repl/ReplCompiler.scala
```diff
@@ -196,13 +194,16 @@ class ReplCompiler extends Compiler {
         val stat = stats.last.asInstanceOf[tpd.Tree]
         if (stat.tpe.isError) stat.tpe.show
         else {
-          val docCtx = ctx.docCtx.get
```
allanrenucci: `docCtx` is now optional in the REPL?

compiler/src/dotty/tools/dotc/core/tasty/CommentPickler.scala
```diff
@@ -23,7 +23,7 @@ class CommentPickler(pickler: TastyPickler, addrOfTree: tpd.Tree => Option[Addr]
       buf.writeAddr(addr)
       buf.writeNat(length)
       buf.writeBytes(bytes, length)
-      buf.writeByte(if (cmt.isExpanded) 1 else 0)
+      buf.writeLongInt(cmt.pos.coords)
```
allanrenucci: @smarter TASTY version bump?

PR #4792, opened 2018-07-13
Fix #4753: Direct method should keep `Override` if overriden nonempty

### LINKED ISSUE #4753: ShortcutImplicits conflicts with RefChecks
```scala
class A

trait Foo {
  def foo: implicit A => Int
}

class Test {
  new FooI{}
}

class FooI extends Foo {
  def foo: implicit A => Int = 3
}
```
fails to compile:
```scala
> dotc Test.scala
-- Error: tests/allan/Test.scala:8:2 -------------------------------------------
8 |  new FooI{}
  |  ^
  |object creation impossible, since def foo$direct(implicit x$0: A): Int is not defined 
```
The issue is that at the time `RefChecks` transforms ` new FooI{}`, `class FooI` hasn't gone through `ShortcutImplicits` which creates the member `def foo$direct(implicit x$0: A): Int` but `trait Foo` has

liufengyun:
Fix #4753: Direct method should keep `Override` if overriden nonempty.

It is the case if the overriden symbol is in the same compilation unit.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala
@@ -181,11 +181,14 @@ object ShortcutImplicits {
   }
 
   /** A new `m$direct` method to accompany the given method `m` */
-  private def newShortcutMethod(sym: Symbol)(implicit ctx: Context): Symbol =
-    sym.copy(
+  private def newShortcutMethod(sym: Symbol)(implicit ctx: Context): Symbol = {
+    val direct = sym.copy(
       name = DirectMethodName(sym.name.asTermName).asInstanceOf[sym.ThisName],
-      flags = sym.flags &~ Override | Synthetic,
+      flags = sym.flags | Synthetic,
       info = directInfo(sym.info))
+    if (direct.allOverriddenSymbols.isEmpty) direct.resetFlag(Override)
+    direct
+  }
 
   def shortcutMethod(sym: Symbol, phase: DenotTransformer)(implicit ctx: Context) =
     sym.owner.info.decl(DirectMethodName(sym.name.asTermName))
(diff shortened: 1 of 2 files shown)


allanrenucci:
It seems to work but there is something fishy. If you compile the class in the opposite order:
```scala
class Foo2 extends Foo1 {
  override def foo: implicit String => Int = 2
}

class Foo1 {
  def foo: implicit String => Int = 1
}
```
After `ShortcutImplicits` you get:
```scala
class Foo2 extends Foo1 { 
  override def foo: implicit String => Int = {
      def $anonfun(implicit evidence$1: String): Int = 
        this.foo$direct(evidence$1)
      closure($anonfun)
    }
  def foo$direct(implicit x$0: String): Int = 2
}

class Foo1 { 
  def foo: implicit String => Int = {
    def $anonfun(implicit evidence$2: String): Int = 
      this.foo$direct(evidence$2)
    closure($anonfun)
  }
  def foo$direct(implicit x$0: String): Int = 1
}
```
You can see that `foo$direct` in `Foo2` is missing the `override` modifier, but `RefChecks` doesn't complain...

This issue derives from the fact that `ShortcutImplicits` doesn't have an `InfoTransformer`. [Here](https://github.com/lampepfl/dotty/blob/2e5368896239673be3c4aeaf273fc998871ad725/compiler/src/dotty/tools/dotc/transform/ShortcutImplicits.scala#L48-L58) @odersky explains why. Maybe the correct fix is to add more special cases to `RefChecks`.

Alternatively, we could try to move `ShortcutImplicits` after `RefChecks`, but we would need to move `TailRec` as well

liufengyun:
Thanks a lot for looking into it @allanrenucci . I think the behavior is expected as when direct method in `Foo2` is synthesized, the direct method in `Foo1` doesn't exist yet. The fix actually just restores the original logic, it avoids changes to overriding check in RefChecks.

Blaisorblade:
After discussion in the meeting: *If* we rely on the `Override` flag to not mean anything after a certain phase (`RefChecks`?), which seemed the case, let's please mention this in `Override`'s Scaladocs.

allanrenucci:
@odersky Why was this closed?

PR #4793, opened 2018-07-15
Remove debug code for stdlib whitelist

nicolasstucki:
Now that our blacklist is a single file this code does not serve any purpose.
It was added back when the whitelist was the main file as a way to copy files
from the blacklist to the whitelist.


nicolasstucki:
@smarter this PR need a quick look at the last commit

PR #4795, opened 2018-07-16
Remove Macro flag from ReifyQuotes

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/MacrosSplitter.scala
```diff
@@ -0,0 +1,188 @@
...
+ *  Transforms an inline macro definitions we assume that we have a single ~ directly as the RHS.
+ *  We will transform the definition from
+ *    ```
+ *    inline def foo[T1, ...](inline x1: X, ..., y1: Y, ....): Z = ~{ ... x1 ... '{ ... T1 ... x1 ... '(y) ... } }
+ *    ```
+ *  to
+ *    ```
+ *    inline def foo[T1, ...](inline x1: X, ..., y1: Y, ....): Z = ~foo$splice$1[T1, ...]('[T1], ..., x1, ..., '(y1), ...)
+ *
+ *    def foo$splice$1[T1, ...](T1$1: Type[T1], ..., x1$1: X, ..., y1: Expr[Y], ....): Z =
```
odersky: What if the original splice is already of the form 

    ~bar(...)

In that case, we can simply forward to `bar`, no? Would it not be simpler to _require_ that all toplevel splices are of this form? 

If we do have a requirement like this, what is the added benefit of the phase?

compiler/src/dotty/tools/dotc/transform/MacrosSplitter.scala
```diff
@@ -0,0 +1,188 @@
...
+   *    val y1$1 = args(1).asInstanceOf[Expr[Y]]
+   *    ...
+   *    { ... x1$1 .... '{ ... T1$1.unary_~ ... x1$1.toExpr.unary_~ ... y1$1.unary_~ ... } ... }
+   *  }
+   *  ```
+   *
+   *  into
+   *
+   *  ```
+   *  def foo$splice$1[T1, ...](T1$1: Type[T1], ..., x1$1: X, ..., y1: Expr[Y], ....): Z =
```
odersky: Where does `foo` come from. It does not appear in the input.

odersky:
We have two concerns here:

 1. Where can the `inline` method be placed? I'd say, best are no restrictions at all.
 2. What is a legal right hand side? Ideally, anything, but here I am willing to compromise to
     say it must be a call of the form `~f[...](....)` where `f` is a static method.

If we do implement that restriction, the phase becomes redundant. 

What is the price for allowing more flexibility in the right hand side? We must find a suitable location for the static method, with suitable parameters and so on. This _looks pretty hard_! It will require at least a call to FulllParameterization to account for all possible combinations of enclosing parameters and so on. But the problem is even harder if we also want to support local inline methods that can refer to type parameters of enclosing classes. 

So, since we do not want to compromise (1) over (2), and (2) does not look really importany anyway, I'd be in favor to impose the restriction and close the PR.

PR #4797, opened 2018-07-17
Add compiler option to use `immutable.Seq` as the default `Seq`.

allanrenucci:
This PR also factors out some shared code


allanrenucci:
Superseded by #4807

PR #4799, opened 2018-07-17
Remove local optimizations

OlivierBlanvillain:
This PR removes the local optimizations, as discussed in the last meeting. To summarize:

- The performance improvements are too small (and inconsistent) compared to the additional compilation cost

- With the plan to reuse the scalac backend in dotty, it makes little sense to have similar optimizations earlier in the pipeline

- We are currently spending 1/3 of the CI time on testing this code


allanrenucci:
@liufengyun Can you update the benchmarks once this is merged?

PR #4804, opened 2018-07-18
Fix #4404: In LambdaLift, modify flags properly

### LINKED ISSUE #4404: Anonymous classes are not final.
It seems like the flags of Anonymous classes are somehow lost during compilation. Despite these lines in Typer.scala:
```scala
val x = tpnme.ANON_CLASS
val clsDef = TypeDef(x, templ1).withFlags(Final)
```
Anonymous classes are compiled without the `public final`.

Compiling this:
```scala
class Car
final class Volvo extends Car

object Car {
    val car = new Car {}
}
```
with dotc yields this:
```
$ dotc Anon.scala 
$ javap Car Volvo 'Car$$anon$1'
Compiled from "Anon.scala"
public class Car {
  public static Car car();
  public Car();
}
Compiled from "Anon.scala"
public final class Volvo extends Car {
  public Volvo();
}
Compiled from "Anon.scala"
class Car$$anon$1 extends Car {
  public Car$$anon$1();
}
```

scalac instead yields this:
```
$ scalac Anon.scala 
$ javap Car Volvo 'Car$$anon$1'
Compiled from "Anon.scala"
public class Car {
  public static Car car();
  public Car();
}
Compiled from "Anon.scala"
public final class Volvo extends Car {
  public Volvo();
}
Compiled from "Anon.scala"
public final class Car$$anon$1 extends Car {
  public Car$$anon$1();
}
```

Medowhill:


compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -338,11 +338,16 @@ object LambdaLift {
             else (topClass, JavaStatic)
           }
           else (lOwner, EmptyFlags)
+        // drop Module because class is no longer a singleton in the lifted context.
+        val nFlags = local.flags &~ Module | Lifted | maybeStatic
+        // keep Public when it is a local class
```
smarter: Why do local classes need to be public and classes lifted into traits non-final?

Medowhill: In the issue #4404, @mbloms said:
> Anonymous classes are compiled without the `public final`.

Therefore, I tried to keep anonymous classes as `public` after lifting. However, I did not consider non-anonymous local classes. I think non-anonymous local classes should not be `public`. I'll correct the code to distinguish anonymous and non-anonymous classes.

At the first, I changed the code simply to keep `final` always. Then, test t4565_1 failed. It was compiled but during the execution, there was an exception:
`Exception in thread "main" java.lang.ClassFormatError: Method A$1 in class T has illegal modifiers: 0x1A`
According to the JVM specification, a method declared inside interface cannot have the final flag, so that I thought that `final` should be dropped when a method is lifted into a trait.

smarter:
Otherwise, this looks great but is missing tests. You could either add a file in `tests/run` that uses Java reflection to check flags, or add tests in https://github.com/lampepfl/dotty/blob/master/compiler/test/dotty/tools/backend/jvm/DottyBytecodeTests.scala to check the flags of the emitted bytecode. If you're not sure how to do that, come to http://gitter.im/lampepfl/dotty and we'll help!

Medowhill:
I changed the code to add the `private` flag always for both methods and classes during the lifting.
I added comments about the JVM specification with its link.
Three tests using the Java reflection are added. The first and second tests are from issue #4404 and the last one is a test to check whether the final flag of a method is dropped correctly when it is lifted into a trait.

allanrenucci:
@Medowhill Can you rebase the PR on master and remove the merge commits?

Medowhill:
I removed the merge.

compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
```diff
@@ -338,11 +338,17 @@ object LambdaLift {
             else (topClass, JavaStatic)
           }
           else (lOwner, EmptyFlags)
+        // Drop Module because class is no longer a singleton in the lifted context.
+        val preInitFlags = local.flags &~ Module | Private | Lifted | maybeStatic
+        // Drop Final when a method is lifted into a trait.
+        // According to the JVM specification, a method declared inside interface cannot have the final flag.
+        // "Methods of interfaces may have any of the flags in Table 4.6-A set except ACC_PROTECTED, ACC_FINAL, ..."
+        // (https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.6)
+        val initFlags = if (!local.isClass && (newOwner.flags is Trait)) preInitFlags &~ Final else preInitFlags
```
allanrenucci: Why don't you test `local.is(Method)` instead of `!local.isClass`?

allanrenucci: Personal test: I think it is clearer to use a `var` rather than coming up with new unique names. Ie.
```scala
var initFlags = ...
if (...) initFlags = initFlags &~ Final
```

tests/run/i4404c.scala
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+class C {
+  def g = {
+    object P
+  }
+}
+
+object Test {
```
allanrenucci: Can you explain what you are testing here

compiler/src/dotty/tools/dotc/transform/LambdaLift.scala
Medowhill: I think your suggestion is better than my original code. I'll fix it. Thank you!

tests/run/i4404c.scala
Medowhill:
```
trait T {
  def f = {
    object O
  }
}
```
When we compile the above code, the compiler will generate `final lazy <accessor> module def O(): O$` inside method `f`. I tried to check whether the `final` flag of method `O` is dropped correctly during the `LambdaLift` phase through the test.

allanrenucci: If you want to test that methods lose their `final` modifier, I would replace the object by a normal `def`. I think it is clearer than using an object relying on the fact that it desugars to a lazy val which itself desugars to a method

Medowhill: That would be clearer. I'll change it. Thank you!

tests/run/i4404c.scala
```diff
@@ -0,0 +1,23 @@
+import java.lang.reflect.Modifier
+
+trait T {
+  def f = {
+    final def f0 = ()
```
allanrenucci: Can you remove the `final` modifier since I believe being able to write it in the first place is a bug?

Medowhill: Then, do I need to change the test to use `object` inside a method again?

allanrenucci: No I think the test is fine as is. We want to make sure that we don't add the `final` modifier in this case

PR #4809, opened 2018-07-19
Check tasty file hash

nicolasstucki:
Based on #4467


nicolasstucki:
This is new. A new PR based on an old approved PR is automatically approved.

PR #4810, opened 2018-07-19
add doc for Override flag hack in ShortcutImplicits

liufengyun:
add doc for Override flag hack in ShortcutImplicits

Address review in #4792 .


allanrenucci:
Thanks @liufengyun

Issue #4520, opened 2018-05-11
Documentation and error messages for synthesized `_${n}` selectors on case classes


gsps:
The new behavior in synthesizing selectors _1, _2, etc. to a case classes' fields caught me by surprise, especially since the error messages may turn out to be quite useless:
```scala
trait EdgeLike[T] {
  def _1: T
  def _2: T
}

case class LabeledEdge[T, U](_1: T, label: U, _2: T) extends EdgeLike[T]
```
leads to
```scala
-- [E007] Type Mismatch Error: try/CaseClassSelectors.scala:6:51 ---------------
6 |case class LabeledEdge[T, U](_1: T, label: U, _2: T) extends EdgeLike[T]
  |                                                   ^
  |                                       found:    U(LabeledEdge.this.label)
  |                                       required: T
  |                                       
-- [E119] Duplicate Symbol Error: try/CaseClassSelectors.scala:6:46 ------------
6 |case class LabeledEdge[T, U](_1: T, label: U, _2: T) extends EdgeLike[T]
  |                                              ^
  |value _2 in class LabeledEdge is already defined as def _2: => T at line 6.

```

Retrospectively, I probably also ran into a very annoying interaction of this and implicits: I was expecting an implicit class with members _1 and _2 to be found, but instead I presumably got an implicit conversion to a case class that now happens to have members _1 and _2.

At a minimum, we should document the new synthesis behavior on case classes.

Blaisorblade:
Since it took me a while to remember: Scala 2 generates no such accessors (arguably, a wart), Dotty does.

It seems we should define a semantics for `_${i}`-named case class fields — possibly, reject them with a specific error message in Scala 3 mode and warn about them in Scala 2 mode. The desugaring is in `Desugar.classDef`, around `caseClassMeths`. BTW, `caseClassMeths` ignores scala 2 mode altogether, which seems an oversight.

Blaisorblade:
Thanks, I'll take that as the needed spec for Scala 3 mode, we still need something for Scala2.

PR #4813, opened 2018-07-19
Fix #4564: Invalidate clashing case class methods

### LINKED ISSUE #4564: Overriding `copy` of case class results in unhelpful error messages
```scala
case class A(x: Int) {
  def copy(x: Int = x) = A(x)
}
```
results in merge errors and friends:
```scala
-- Error: try/CaseClassCopyMergeError.scala:2:20 -------------------------------
2 |  def copy(x: Int = this.x) = A(x)
  |                    ^^^^^^
  |             cannot merge
  |               method copy: (x: Int): A  and
  |               method copy: (x: Int): A;
  |             they are both defined in class A but have matching signatures
  |               (x: Int): A and
  |               (x: Int): A
  |             when seen as members of A(A.this)
-- Error: try/CaseClassCopyMergeError.scala:1:19 -------------------------------
1 |case class A(x: Int) {
  |                   ^
  |             cannot merge
  |               method copy: (x: Int): A  and
  |               method copy: (x: Int): A;
  |             they are both defined in class A but have matching signatures
  |               (x: Int): A and
  |               (x: Int): A
  |             when seen as members of A(A.this)
-- [E119] Duplicate Symbol Error: try/CaseClassCopyMergeError.scala:2:6 --------
2 |  def copy(x: Int = this.x) = A(x)
  |      ^
  |method copy in class A is already defined as def copy(x: Int): A at line 1.
  |The definitions have matching type signatures after erasure.
-- Error: try/CaseClassCopyMergeError.scala:1:0 --------------------------------
1 |case class A(x: Int) {
  |^
  |two or more overloaded variants of method copy have default arguments
```

We have a use case for this in inox (where we override `copy` to also also copy over some metadata), but I see how one would like to enforce some semantics for `copy` on case classes. In any case, this should produce a more intelligible error message.

odersky:
Invalidate a compiler generated case class method that clashes
with a user-defined method in the same scope with a matching type.

Invalidation is done by overwriting its info on completion with `NoType` and
eliminating the method altogether during type checking.

Methods potentially affected are `apply`, `unapply`, `copy`, and `copy`'s default getters.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1993,7 +1993,6 @@ class Typer extends Namer
    *                (but do this at most once per tree).
    *
    *  After that, two strategies are tried, and the first that is successful is picked.
-   *  If neither of the strategies are successful, continues with`fallBack`.
    *
    *  1st strategy: Try to insert `.apply` so that the result conforms to prototype `pt`.
    *                This strategy is not tried if the prototype represents already
@@ -2002,6 +2001,10 @@ class Typer extends Namer
    *  2nd strategy: If tree is a select `qual.name`, try to insert an implicit conversion
    *    around the qualifier part `qual` so that the result conforms to the expected type
    *    with wildcard result type.
+   *
+   *  If neither of the strategies are successful, continues with the `apply` result
+   *  if an apply insertion was tried and `tree` has an `apply` method, or continues
+   *  with `fallBack` otherwise. `fallBack` is supposed to always give an error.
    */
   def tryInsertApplyOrImplicit(tree: Tree, pt: ProtoType, locked: TypeVars)(fallBack: => Tree)(implicit ctx: Context): Tree = {
 
@@ -2023,7 +2026,7 @@ class Typer extends Namer
       else try adapt(simplify(sel, pt, locked), pt, locked) finally sel.removeAttachment(InsertedApply)
     }
 
-    def tryImplicit =
+    def tryImplicit(fallBack: => Tree) =
       tryInsertImplicitOnQualifier(tree, pt, locked).getOrElse(fallBack)
 
     pt match {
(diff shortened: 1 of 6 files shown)


odersky:
See also scala/scala #5816.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1413,7 +1413,12 @@ class Typer extends Namer
     }
   }
 
-  def typedDefDef(ddef: untpd.DefDef, sym: Symbol)(implicit ctx: Context) = track("typedDefDef") {
+  def typedDefDef(ddef: untpd.DefDef, sym: Symbol)(implicit ctx: Context): Tree = track("typedDefDef") {
+    if (!sym.info.exists) { // it's a discarded synthetic case class method, drop it
+      assert(sym.is(Synthetic) && desugar.isDesugaredCaseClassMethodName(sym.name))
+      sym.owner.info.decls.openForMutations.unlink(sym)
```
adriaanm: Neat! We should consider dropping here, instead of in the completer, for scalac too.

tests/pos/i4564.scala
```diff
@@ -0,0 +1,63 @@
+case class A(x: Int) {
```
adriaanm: There's also a negative version -- would be good to include.

odersky:
@smarter I believe it's probably better if we exclude selectors from the scheme altogether.

adriaanm:
PS: A pleasure to read the code!

compiler/src/dotty/tools/dotc/typer/Typer.scala
odersky: It works because `dotc` is quite good at deciding that symbols with `NoType` as info don't really exist. (c.f. `reallyExsists` and its references). I am not sure what the state for scalac is here.

odersky:
Added the neg test. This required some more changes in Typer to give better error messages similar to what's done in scala/scala.

Blaisorblade:
If we keep this scheme for `_1`, and so on, we could also try taking a look at #4520 (about the error messages we get, and the expected behavior in Scala 2 and Scala 3 modes). If we still generate `_1` in Scala2 mode, should 2.14 deprecate such custom accessors?

PR #4817, opened 2018-07-20
Fix #4451: Set flag in the correct order

### LINKED ISSUE #4451: Lazy vals initialisation can be retried after exception
```scala
class Foo(x: Int) {
  private[this] var init = false;
  lazy val foo = {
    if (!init) {
      init = true
      assert(false)
    }
    x
  }
}

object Test {
    val foo = new Foo(42)
    try foo.foo
    catch {
      case _: AssertionError =>
        println(foo.foo) // prints 0, should be 42
    }
  }
}
```

Medowhill:


compiler/src/dotty/tools/dotc/transform/LazyVals.scala
```diff
@@ -219,9 +219,9 @@ class LazyVals extends MiniPhase with IdentityDenotTransformer {
     */
   def mkNonThreadSafeDef(target: Tree, flag: Tree, rhs: Tree, nullables: List[Symbol])(implicit ctx: Context) = {
     val setFlag = flag.becomes(Literal(Constants.Constant(true)))
-    val setNullables = nullOut(nullables)
-    val setTargetAndNullable = if (isWildcardArg(rhs)) setNullables else target.becomes(rhs) :: setNullables
-    val init = Block(setFlag :: setTargetAndNullable, target.ensureApplied)
+    var setTargetFlagNullables = setFlag :: nullOut(nullables)
+    if (!isWildcardArg(rhs)) setTargetFlagNullables = target.becomes(rhs) :: setTargetFlagNullables
+    val init = Block(setTargetFlagNullables, target.ensureApplied)
```
allanrenucci: I would rewrite as:
```scala
val stats = new mutable.ListBuffer[Tree]
if (!isWildcardArg(rhs)) stats += target.becomes(rhs)
stats += setFlag += nullOut(nullables)
val init = Block(stats.toList, target.ensureApplied)
```

PR #4821, opened 2018-07-20
Fix #4364: Try SAM type when no candidates found

### LINKED ISSUE #4364: SAM types + Overloading + Eta-expansion
The following code snippet fails to compile with Dotty. (Works with scalac)
```scala
class Test {
  def println(): Unit = ()
  def println(x: String): Unit = ()

  def foo(x: java.util.function.Consumer[String]) = 1
  foo(println(_))      // ok
  foo(x => println(x)) // ok
  foo(println)         // error
}
```
```scala
-- [E007] Type Mismatch Error: tests/allan/Test.scala:26:6 ---------------------
26 |  foo(println)
   |      ^^^^^^^
   |      found:    Unit
   |      required: java.util.function.Consumer[String]
   |      
one error found
```

Medowhill:


compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1427,7 +1427,17 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
 
       case pt =>
-        alts filter (normalizedCompatible(_, pt))
+        val noSam = alts filter (normalizedCompatible(_, pt))
+        if (noSam.isEmpty) {
+          pt match {
+            case SAMType(mtp) =>
+              val sam = narrowByTypes(alts, mtp.paramInfos, mtp.resultType)
+              if (sam.nonEmpty && !pt.classSymbol.hasAnnotation(defn.FunctionalInterfaceAnnot))
+                ctx.warning(ex"$pt does not have the @FunctionalInterface annotation.", ctx.tree.pos)
```
smarter: I still don't think this is worth adding a warning for. Even if it is, there should be a simple way to turn off this warning.

allanrenucci: The simple way is to do explicit eta-expension, no?
```scala
new ObjectOutputStream(println) // warning: are you sure this what you want to do?
new ObjectOutputStream(() => println()) // Ok
new ObjectOutputStream(println _)  // Ok
```

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1427,7 +1427,17 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
         narrowByTypes(alts, args, resultType)
 
       case pt =>
-        alts filter (normalizedCompatible(_, pt))
+        val noSam = alts filter (normalizedCompatible(_, pt))
+        if (noSam.isEmpty) {
+          pt match {
+            case SAMType(mtp) =>
```
allanrenucci: I would move this case to the enclosing match

compiler/src/dotty/tools/dotc/typer/Applications.scala
allanrenucci: `ctx.tree.pos` here will not be very precise. Ideally it should be the position of the tree being expended:
```scala
-- Warning: tests/allan/Test.scala:5:6 -----------------------------------------
5 |  val out2 = new ObjectOutputStream(println)
  |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Should be:
```scala
-- Warning: tests/allan/Test.scala:5:6 -----------------------------------------
5 |  val out2 = new ObjectOutputStream(println)
  |                                    ^^^^^^^
```
```

File: tests/run/i4364a.scala [hunk 3]
```diff
@@ -0,0 +1,13 @@
+object Test {
+  var flag = false
```
allanrenucci: I suppose you can remove this

tests/run/i4364a.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  def f(): Unit = assert(false)
+  def f(x: Int): Unit = assert(false)
+  def f(x: String): Unit = flag = true
```
allanrenucci:
```scala
def f(x: String): Unit = ()
```

tests/run/i4364b.scala
```diff
@@ -0,0 +1,23 @@
+object Test {
```
allanrenucci: It is not clear what you are testing here that is not tested in `i4364a.scala`. I suggest you add a test to `tests/neg-custom-args/fatal-warnings` to test the warnings:
```scala
import java.io._

class Test {
  def foo(c: java.util.function.Consumer[String]) = ()

  val out1 = new ObjectOutputStream(x => println(x))
  val out2 = new ObjectOutputStream(println(_))
  val out3 = new ObjectOutputStream(println) // error: OutputStream is not @FunctionalInterface

  def f(x: String) = ()
  foo(f) // Ok: Consumer is @FunctionalInterface
}
```

Medowhill:
```scala
def f(x: Int): Unit = assert(false)
def f(x: String): Unit = assert(false)
def f: java.io.OutputStream = new java.io.OutputStream { def write(x: Int) = () }

val oosF = new java.io.ObjectOutputStream(f)
oosF.write(0)
```
Through the test, I tried to check that whether method overloading is correctly resolved when candidates exist without considering SAM type. In this case, the second `f` is valid choice when it is eta-expanded and `OutputStream` is considered as SAM type. However, the third `f` should be choosed. Therefore, I think this test needs to remain.

As you suggested, it would be better to check a warning in a seperate test. I'll remove
```scala
val oosG = new java.io.ObjectOutputStream(g)
oosG.write(0)
oosG.close()
```
from test b and add a test about a warning in `tests/neg-custom-args/fatal-warnings`.

allanrenucci: Good point. I suggest a simpler test though:
```scala
import java.util.function.Consumer

object Test {
  def foo(c: Consumer[String]) = c.accept("")

  def bar(x: String): Unit = ???
  def bar: Consumer[String] = new Consumer {  def accept(s: String) = () }

    foo(bar)
  }
}
```

Medowhill: I agree that this one is simpler and clearer. I'll fix it. Thank you!

compiler/src/dotty/tools/dotc/typer/Applications.scala
```diff
@@ -1421,13 +1421,31 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
...
+           * For example, the second f should be chosen for the following code:
+           *   def f(x: String): Unit = ???
+           *   def f: java.io.OutputStream = ???
+           *   new java.io.ObjectOutputStream(f)
+           */
+          pt match {
+            case SAMType(mtp) =>
+              val sam = narrowByTypes(alts, mtp.paramInfos, mtp.resultType)
+              if (sam.nonEmpty && !pt.classSymbol.hasAnnotation(defn.FunctionalInterfaceAnnot))
+                ctx.warning(ex"${sam.head.designator} is eta-expanded even though $pt does not have the @FunctionalInterface annotation.", pos)
```
allanrenucci: I am not 100% sure this is the right place to emit the warning. This assumes that this overload will be chosen, and its argument will be eta-expanded . Although I couldn't come up with an example where the warning is emitted but no eta-expansion happen.

I would emit the warning where the eta-expansion happens

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2321,9 +2321,11 @@ class Typer extends Namer
       if (arity >= 0 &&
           !tree.symbol.isConstructor &&
           !ctx.mode.is(Mode.Pattern) &&
-          !(isSyntheticApply(tree) && !isExpandableApply))
+          !(isSyntheticApply(tree) && !isExpandableApply)) {
+        if (!pt.classSymbol.hasAnnotation(defn.FunctionalInterfaceAnnot))
```
allanrenucci: You need to exclude function classes (i.e. `defn.isFunctionClass`)

Medowhill:
I found the following comment from the 1331st line of `Applications.scala`.
> This private version of `resolveOverloaded` does the bulk of the work of overloading resolution, but does not do result adaptation. It might be called twice from the public `resolveOverloaded` method, once with implicits and SAM conversions enabled, and once without.

In the current version of the public `resolveOverloaded` method, the private `resolveOverloaded` method is called without `Mode.ImplicitsEnabled` at the first time. If there are no candidates, it is called again with `Mode.ImplicitsEnabled`. I think that currently there is no mode corresponding to SAM conversion.

How do you think about the following fix?

Adding
```scala
val SAMConversionEnabled = newMode(...)
```
to `Mode.scala`.

Changing public `resolveOverloaded` to
```scala
var found = resolveOverloaded(alts, pt, Nil)(ctx.retractMode(Mode.ImplicitsEnabled).retractMode(Mode.SAMConversionEnabled)
if (found.isEmpty && (ctx.mode.is(Mode.ImplicitsEnabled) || ctx.mode.is(Mode.SAMConversionEnabled)))
  found = resolveOverloaded(alts, pt, Nil)
```

Changing private `resolveOverloaded` to
```scala
val candidates = pt match {
  ...
  case SAMType(mtp) if ctx.mode.is(Mode.SAMConversionEnabled) =>
    narrowByTypes(alts, mtp.paramInfos, mtp.resultType)
  case pt =>
    alts filter (normalizedCompatible(_, pt))
}
```

Medowhill:
I thought again about the above fix. It will result unnecessary redundant calculations. Would it be better to fix the comment to explain the current code correctly and keep the current code?

allanrenucci:
> I thought again about the above fix. It will result unnecessary redundant calculations. Would it be better to fix the comment to explain the current code correctly and keep the current code?

Your proposed fix looked reasonable to me (except that you can reuse `Mode.ImplicitsEnabled` and don't need to introduce a new mode). I tried it out and was able to come up with a test that would not compile:
```scala
import java.util.function.Consumer

class Test {
  def foo(c: Consumer[String]) = 1
  def f(x: Int): Unit = ()
  def f: Int = 1
  implicit def bar(x: Int): Consumer[String] = ???

  foo(f) // error: none of the overloaded alternatives of method f match expected type Consumer[String]
}
```
So I would say, we leave it as is for now. #4831 is a related issue I found while looking at it

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2353,9 +2353,13 @@ class Typer extends Namer
           !tree.symbol.isConstructor &&
           !tree.symbol.is(TransparentMethod) &&
           !ctx.mode.is(Mode.Pattern) &&
-          !(isSyntheticApply(tree) && !isExpandableApply))
+          !(isSyntheticApply(tree) && !isExpandableApply)) {
+        if (!pt.classSymbol.hasAnnotation(defn.FunctionalInterfaceAnnot) &&
+            !defn.isFunctionType(pt) &&
+            SAMType.unapply(pt).isDefined)
+          ctx.warning(ex"${tree.symbol} is eta-expanded even though ${pt.classSymbol} does not have the @FunctionalInterface annotation.", tree.pos)
```
odersky: Small nitpick: I'd change the condition to

    if (!defn.isFunctionType(pt))
      pt match {
        case SAMType(_) if !pt.classSymbol.hasAnnotation(defn.FunctionalInterfaceAnnot) =>
          ...
        case _ =>
     }

adriaanm:
I think it's a good idea to warn. Since Java users are encouraged (though not required) to express their intent with the `@functionalinterface` annotation it's likely that they will. So, the omission of the annotation is a good indication we shouldn't be treating the interface like a SAM. I'd be happy to have this improvement backported to Scala 2!

odersky:
Superseded by #5717.

PR #4822, opened 2018-07-21
Change top-level ~ evaluation scheme

nicolasstucki:
The new scheme consists in only allowing simple interpretable expressions in top-level splices. Then, the inlined splice can be evaluated directly.

Changes:
* top-level ~ must contain a call to a static method
  * arguments must be quoted
  * inline arguments can be passed directly
* removed dependency on the original inlined code (removed in PostTyper)
* ReifyQuotes is no longer an InfoTransformer
* Splicer implements both checking and evaluation of splices in a single abstraction
* Addapted some tests
* Fix #4773
* Fix #4735

PR followed by #4823

PR #4823, opened 2018-07-23
Enable macros in any transparent def

nicolasstucki:
Based on #4822.

Allow top-level ~ in and non static transparent method:
* Class methods (including inner and anonymous classes)
* Methods in def/val/var

Followed by #4826.


nicolasstucki:
Rebased on master

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -103,14 +103,16 @@ class ReifyQuotes extends MacroTransformWithImplicits {
   /** The main transformer class
    *  @param  inQuote    we are within a `'(...)` context that is not shadowed by a nested `~(...)`
    *  @param  outer      the next outer reifier, null is this is the topmost transformer
-   *  @param  level      the current level, where quotes add one and splices subtract one level
+   *  @param  level      the current level, where quotes add one and splices subtract one level.
+   *                     The initial level is 0, a level `l` where `l > 0` implies code has been quotes `l` times
+   *                     and `l == -1` is code inside a top level splice (in an transparent method).
```
biboudis: "in a"

PR #4824, opened 2018-07-23
Add missing SuperType in pattern match

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -571,6 +571,7 @@ class TastyImpl(val rootContext: Contexts.Context) extends scala.tasty.Tasty { s
       def unapply(x: Pattern)(implicit ctx: Context): Option[Term] = x match {
         case lit: tpd.Literal @unchecked => Some(lit)
         case ref: tpd.RefTree @unchecked if ref.isTerm => Some(ref)
+        case ths: tpd.This @unchecked => Some(ths)
```
Blaisorblade: Do we have tests for this code?

nicolasstucki: In #4645, but it will take some time to be merged.

PR #4827, opened 2018-07-23
Fix #4582: use untpd.New in maybeCall

### LINKED ISSUE #4582: Extending trait with context bounds requires unnecessary parenthesis
Given:
```scala
trait T[A: Numeric]
class TX[A: Numeric] extends T[A]
```
compiler yields:
```
class TX[A: Numeric] extends T[A]
                             ^^^^
                      constructor T in trait T does not take parameters
```
It works if you do `extends T[A]()`

Medowhill:


tests/pos/i4582.scala
```diff
@@ -0,0 +1,2 @@
+trait T[A: Numeric]
+class TX[A: Numeric] extends T[A]
```
allanrenucci: This doesn't exercice the code you changed but I would add a test with an argument list. E.g.
```scala
trait T2[A: Numeric](x: Int)
class T2X[A: Numeric] extends T[A](1)
```

Medowhill: I added it.

compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1475,8 +1475,7 @@ class Typer extends Namer
      */
     def maybeCall(ref: Tree, psym: Symbol, cinfo: Type): Tree = cinfo.stripPoly match {
       case cinfo @ MethodType(Nil) if cinfo.resultType.isImplicitMethod =>
-        val icall = New(ref).select(nme.CONSTRUCTOR).appliedToNone
-        typedExpr(untpd.TypedSplice(icall))(superCtx)
+        typedExpr(untpd.New(ref, Nil))(superCtx)
```
Blaisorblade: I love when people fix bugs by removing code! ❤️

Blaisorblade:
@Medowhill Could we also test the interaction with #4798? That is, try adding variants of those tests with context bounds. Hopefully that already just works!

Medowhill:
I think this issue is orthogonal to #4798 because the following code is already compiled successfully at the master branch.
```scala
trait A[X]
object O {
  type T[X] = A[X]
  class B[X] extends T[X]
}
```
`class B[X] extedns T[X]` is transformed into `class B[X] extends Object with A[X]`.

However, adding the a can be meaningful. Do you think it would be better to add the following test?
```scala
trait A[X: Numeric]
object O {
  type T[X] = A[X]
  class B[X: Numeric] extends T[X]
}
```

Blaisorblade:
Yep, I'm talking about adding tests — I only wondered because this code calls `untpd.New`, which you're changing. To test the interaction I was thinking of something like the following (which fails on master and works on your branch).

```
class A[X: Numeric]
type T[X, Y] = A[X]
type Foo[X, Y] = T[X, Y]
class B[X: Numeric] extends Foo[X, Int]
```

Medowhill:
I found a new issue while writing tests.
```scala
trait T[X]
type Foo[X, Y] = T[X]
class D[X] extends Foo[X, Unit]
```
```scala
trait T[X: Numeric]
type Foo[X] = T[X]
class D[X: Numeric] extends Foo[X]
```
Two above examples were compiled.
```scala
trait T[X: Numeric]
type Foo[X, Y] = T[X]
class D[X: Numeric] extends Foo[X, Unit]
```
However, this one resulted the following crash:
```scala
Exception in thread "main" java.lang.AssertionError: assertion failed: missing parameters for trait T from ($outer: C, implicit evidence$2: scala.math.Numeric) extends Object() with T {
  private implicit val evidence$2: scala.math.Numeric
  private val $outer: C
  private <accessor> def $outer(): C = this.$outer
  final def C$D$$$outer(): C = D.this.$outer()
} should have been caught in typer
	at dotty.tools.dotc.transform.Mixin.nextArgument$1(Mixin.scala:209)
	at dotty.tools.dotc.transform.Mixin.$anonfun$transformTemplate$12(Mixin.scala:235)
	at scala.collection.TraversableLike$WithFilter.$anonfun$map$2(TraversableLike.scala:739)
	at scala.collection.TraversableLike$WithFilter.map(TraversableLike.scala:738)
	at dotty.tools.dotc.transform.Mixin.traitInits$1(Mixin.scala:217)
	at dotty.tools.dotc.transform.Mixin.$anonfun$transformTemplate$17(Mixin.scala:263)
	at scala.collection.immutable.List.flatMap(List.scala:335)
	at dotty.tools.dotc.transform.Mixin.transformTemplate(Mixin.scala:262)
	at dotty.tools.dotc.transform.Mixin.transformTemplate(Mixin.scala:98)
	at dotty.tools.dotc.transform.MegaPhase.goTemplate(MegaPhase.scala:922)
```

allanrenucci:
> I found a new issue while writing tests.

Please open an issue

Medowhill:
I've just opened it: #4837.

PR #4828, opened 2018-07-23
Upgrade IDE dependencies, release vscode-dotty 0.1.5

smarter:
On the server side, uses of lsp4j.MarkedString were replaced by
lsp4j.MarkupContent since the former is deprecated (and seems to cause
some rendering bugs with the updated dependencies).

### DIFF
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -348,7 +348,7 @@ class DottyLanguageServer extends LanguageServer
     else {
       val symbol = Interactive.enclosingSourceSymbol(trees, pos)
       val docComment = ctx.docCtx.flatMap(_.docstring(symbol))
-      val content = hoverContent(tpw.show.toString, docComment)
+      val content = hoverContent(tpw.show, docComment)
       new Hover(content, null)
     }
   }
@@ -459,9 +459,9 @@ object DottyLanguageServer {
         CIK.Field
     }
 
-    val label = sym.name.show.toString
+    val label = sym.name.show
     val item = new lsp4j.CompletionItem(label)
-    item.setDetail(sym.info.widenTermRefExpr.show.toString)
+    item.setDetail(sym.info.widenTermRefExpr.show)
     item.setKind(completionItemKind(sym))
     item
   }
@@ -503,10 +503,10 @@ object DottyLanguageServer {
         SK.Field
     }
 
-    val name = sym.name.show.toString
+    val name = sym.name.show
     val containerName =
       if (sym.owner.exists && !sym.owner.isEmptyPackage)
-        sym.owner.name.show.toString
+        sym.owner.name.show
       else
         null
 
--- a/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
+++ b/language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
@@ -222,7 +222,7 @@ class DottyLanguageServer extends LanguageServer
 
 
   // FIXME: share code with messages.NotAMember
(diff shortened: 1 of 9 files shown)


language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -466,19 +466,17 @@ object DottyLanguageServer {
...
-
-    typeInfoHover :: docHover.toList
+  private def hoverContent(typeInfo: String, comment: Option[Comment]): lsp4j.MarkupContent = {
+    val docMarkup = comment.map(_.raw).getOrElse("")
+    val markup = new lsp4j.MarkupContent
+    markup.setKind("markdown")
+    markup.setValue(
+      s"""```scala
+         |$typeInfo
+         |$docMarkup```""".stripMargin.stripLineEnd)
```
allanrenucci: Why `stripLineEnd`? Also shouldn't the ``` be on a new line?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -344,12 +344,12 @@ class DottyLanguageServer extends LanguageServer
...
     val tpw = tp.widenTermRefExpr
 
-    if (tpw == NoType) new Hover
+    if (tpw == NoType) null // null here indicates that no response should be sent
     else {
       val symbol = Interactive.enclosingSourceSymbol(trees, pos)
       val docComment = ctx.docCtx.flatMap(_.docstring(symbol))
-      val markedStrings = docMarkedStrings(docComment, tpw.show.toString)
-      new Hover(markedStrings.map(JEither.forRight(_)).asJava, null)
+      val content = hoverContent(tpw.show.toString, docComment)
```
allanrenucci: `toString` redundant after `show`?

language-server/src/dotty/tools/languageserver/DottyLanguageServer.scala
```diff
@@ -459,26 +459,31 @@ object DottyLanguageServer {
...
+  private def hoverContent(typeInfo: String, comment: Option[Comment]): lsp4j.MarkupContent = {
+    val markup = new lsp4j.MarkupContent
+    markup.setKind("markdown")
+    markup.setValue((
+      comment.map(_.raw) match {
+        case Some(comment) =>
+          s"""```scala
+             |$typeInfo
+             |$comment
+             |```"""
```
Duhemm: There used to be a thin line that separated the type info from the comment. It would be nice to re-introduce it, since it makes the hover information easier to read:

```scala
case Some(comment) =>
  s"""```scala
     |$typeInfo
     |```
     |---
     |```scala
     |$comment
     |```"""
```

smarter: I tried that but that's not equivalent to the thin line we had before: it adds a really thick line which I find distracting

smarter: There's an issue for this: https://github.com/Microsoft/language-server-protocol/issues/518

smarter:
I'm going to merge this, however I won't release 0.1.5 until the next Dotty release: it looks like recent version of vscode-languageclient do not agree with LSP4J on how to read the content of MarkedString in hover responses, the result is that when using vscode-dotty 0.1.5 with Dotty 0.9.0-RC1, the hover messages just displays `[object Object]`. We could try to figure out if this is a bug in vscode-languageclient but it won't matter once we release a version of the Dotty Language Server with this PR in, since I removed usages of MarkedString, so it doesn't seem worth the effort. We could release 0.9.0-RC2 with this PR backported but that also doesn't seem worth it.

Issue #4819, opened 2018-07-20
Method in trait not a legal implementation in inheriting class


allanrenucci:
```scala
trait Iterable[+A]

trait IterableOps[+A, +CC[_], +C] {
  def concat[B >: A](suffix: Iterable[B]): CC[B] = ???
}

trait SetOps[A, +CC[X], +C <: SetOps[A, CC, C]]  {
  def concat(that: Iterable[A]): C = ???
}

class ChampHashSet[A] extends SetOps[A, ChampHashSet, ChampHashSet[A]] with
    IterableOps[A, ChampHashSet, ChampHashSet[A]]

```
```
-- Error: tests/allan/Test.scala:15:6 ------------------------------------------
15 |class ChampHashSet[A] extends SetOps[A, ChampHashSet, ChampHashSet[A]] with
   |      ^
   |method concat in trait IterableOps is not a legal implementation of `concat' in class ChampHashSet
   |  its type             [B >: A](suffix: Iterable[B]): ChampHashSet[B]
   |  does not conform to  (that: Iterable[A]): ChampHashSet[A]
one error found
```

Blaisorblade:
FWIW: I've taken a look and it makes sense to accept this (like Scalac does) — even ML-family languages allow implementations to be more polymorphic than interfaces.

smarter:
@Blaisorblade Maybe, but the following has never been allowed in Scala:
```scala
trait Foo {
  def foo(x: String): Unit
}
class Bar extends Foo {
  def foo[B](x: String): Unit = {}
}
```

smarter:
```scala
trait IterableOps[+A, +CC[_], +C] {
  def concat[B >: A](suffix: Iterable[B]): CC[B] = ???
}
```
ends up with a `concat` method whose erased type is:
```scala
def concat(suffix: Iterable): Object
```
whereas with:
```scala
trait SetOps[A, +CC[X], +C <: SetOps[A, CC, C]]  {
  def concat(that: Iterable[A]): C = ???
}
```
the erasure of `concat` is:
```scala
def concat(that: Iterable): SetOps
```
So in `ChampHashSet` we should be able to inherit from both of them and get two overloads for `concat`. This is what happens in scalac but in Dotty this is prevented by RefChecks for some reason.

Blaisorblade:
Sorry I was too quick and hopeful and misunderstood this completely...

allanrenucci:
[...]

PR #4829, opened 2018-07-23
Workaround #4819: Avoid creating incorrect JointRefDenotations

smarter:
When merging a denotation with a PolyType and a denotation with a
MethodType, `infoMeet` will return the MethodType info, but the symbol
that is then used to create the `JointRefDenotation` can come from
either denotation, if it comes from the PolyType one, the resulting
denotation is inconsistent, this resulted in RefChecks errors.

We fix this by making `preferSym` consistent with `infoMeet`. This is
only a workaround and not a proper fix for #4819 as it doesn't allow
tests/neg/i4819.scala to compile, see the comments in the file.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Denotations.scala
+++ b/compiler/src/dotty/tools/dotc/core/Denotations.scala
@@ -490,7 +490,9 @@ object Denotations {
          *   4. The access boundary of sym2 is properly contained in the access
          *      boundary of sym1. For protected access, we count the enclosing
          *      package as access boundary.
-         *   5. sym1 a method but sym2 is not.
+         *   5. sym1 is a method but sym2 is not.
+         *   6. sym1 is a non-polymorphic method but sym2 is a polymorphic method.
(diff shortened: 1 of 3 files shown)


allanrenucci:
With this patch, while compiling the standard library I get:
```scala
[error] -- Error: /Users/renucci/workspace/scala/src/library/scala/collection/immutable/TreeMap.scala:28:12 
[error] 28 |final class TreeMap[K, +V] private (tree: RB.Tree[K, V])(implicit val ordering: Ordering[K])
[error]    |            ^
[error]    |method collect in trait StrictOptimizedMapOps is not a legal implementation of `collect` in class TreeMap
[error]    |  its type             [K2, V2](pf: PartialFunction[(K, V), (K2, V2)]): scala.collection.Map[K2, V2]
[error]    |  does not conform to  [K2, V2](pf: PartialFunction[(K, V), (K2, V2)]): Map[K2, V2]
```
I'll try to minimise

PR #4832, opened 2018-07-24
Fix the reproduction script for testFromTasty tests

nicolasstucki:
For example a failure in a `testFromTasty` for `tests/run/eff-dependent.scala`  will output the following report:
```
================================================================================
Test Report
================================================================================

1 suites passed, 1 failed, 2 total
    tests/run/eff-dependent.scala failed


Test 'tests/run/eff-dependent.scala' compiled with 0 error(s) and 0 warning(s),
the test can be reproduced by running:

  sbt "testFromTasty tests/run/eff-dependent.scala"
```

### DIFF
--- a/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
+++ b/compiler/test/dotty/tools/vulpix/ParallelTesting.scala
@@ -1284,11 +1284,15 @@ trait ParallelTesting extends RunnerOrchestration { self =>
     ) extends JointCompilationSource(name, Array(file), flags, outDir, fromTasty, decompilation) {
 
       override def buildInstructions(errors: Int, warnings: Int): String = {
+        val runOrPos = if (file.getPath.startsWith("tests/run/")) "run" else "pos"
+        val listName = if (fromTasty) "from-tasty" else "decompilation"
         s"""|
             |Test '$title' compiled with $errors error(s) and $warnings warning(s),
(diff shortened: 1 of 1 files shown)


compiler/test/dotty/tools/vulpix/ParallelTesting.scala
```diff
@@ -1274,9 +1274,30 @@ trait ParallelTesting extends RunnerOrchestration { self =>
...
+       fromTasty: Boolean = false,
+       decompilation: Boolean = false
+    ) extends JointCompilationSource(name, Array(file), flags, outDir, fromTasty, decompilation) {
+
+      override def buildInstructions(errors: Int, warnings: Int): String = {
+        s"""|
+            |Test '$title' compiled with $errors error(s) and $warnings warning(s),
+            |the test can be reproduced by running:
+            |
+            |  sbt "testFromTasty $file"
```
allanrenucci: This is the easiest way to reproduce but is it helpful for debugging? My experience is that when you run a test with vulpix and the compiler crashes then the exception is swallowed and I either need to look at the test log or run the compiler manually.

I often want to run the compiler manually outside vulpix and I would rather have the manual instructions. E.g:
```
> dotc -d tmp Foo.scala -some-options
> dotc Foo -classpath tmp -from-tasty -some-options
```

PR #4833, opened 2018-07-24
Always run the compiler with a bootstrapped dotty-library

smarter:
The non-bootstrapped dotty-library may not be binary compatible with the
compiler, this hasn't affected us so far because the non-bootstrapped
dotty-library has always been compiled with Scala 2, but in the future
it will be compiled by an old version of Dotty itself.

To enforce this, I added a new flag -Yscala2-unpickler and set it so
that scala-library is the only jar from which we're allowed to unpickle
Scala 2 symbols in both the build and the tests.


project/Build.scala
```diff
@@ -679,20 +686,26 @@ object Build {
...
+      }
+    }.value
   )
 
   lazy val bootstrapedDottyCompilerSettings = commonDottyCompilerSettings ++ Seq(
     packageAll := {
       packageAll.in(`dotty-compiler`).value ++ Seq(
-        "dotty-compiler" -> packageBin.in(Compile).value.getAbsolutePath,
-        "dotty-library"  -> packageBin.in(`dotty-library-bootstrapped`, Compile).value.getAbsolutePath
+        "dotty-compiler" -> packageBin.in(Compile).value.getAbsolutePath
```
nicolasstucki: Why is the library not packaged in this case?

project/Build.scala
```diff
@@ -782,6 +782,21 @@ object Build {
...
+        assert(i != -1)
+        val oldValue = oldOptions(i + 1)
+
+        val attList = (dependencyClasspath in Test).value
+        val sbtIo = findLib(attList, "org.scala-sbt/io")
+        val zincApiInfo = findLib(attList, "zinc-apiinfo")
+
+        oldOptions.updated(i + 1, s"$sbtIo:$zincApiInfo:$oldValue")
+      }
+    )
```
nicolasstucki: Is there any way to add a test for this?

PR #4834, opened 2018-07-24
Refactorings on Names: forbid `.name` on Name and cleanups on the way

Blaisorblade:
Did you know you could call `.name` on a `Name`? No more! 😄

Extracted from #4729 after we saw more with @allanrenucci , assigned to him.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/Names.scala
+++ b/compiler/src/dotty/tools/dotc/core/Names.scala
@@ -98,7 +98,7 @@ object Names {
     /** A name in the same (term or type) namespace as this name and
      *  with same characters as given `name`.
      */
-    def likeSpaced(name: PreName): ThisName
+    def likeSpaced(name: Name): ThisName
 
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/core/Names.scala
```diff
@@ -98,7 +98,7 @@ object Names {
     /** A name in the same (term or type) namespace as this name and
      *  with same characters as given `name`.
      */
-    def likeSpaced(name: Name): ThisName
+    def likeSpaced(name: PreName): ThisName
```
allanrenucci: Currently, there don't seems to be usages for `likeSpaced(PreName)` so I would keep the API as is. If you want to share the implementation with the one in `NameOps`, it seems you can actually replace `PreName` by `Name`

Blaisorblade:
> Can you squash your commits into a single one?

Do you mind if I do a GitHub squash merge instead (once tests pass)? These are intentionally separate.

PR #4845, opened 2018-07-25
Add new test to the blacklist

nicolasstucki:


nicolasstucki:
The reason was due to conflicts between two concurrent merges. One added the blacklist and started testing all other tests, while the other PR added a test that should have been blacklisted.

PR #4853, opened 2018-07-27
Fix #4466: same ctx for tryInsertApplyOrImplicit

### LINKED ISSUE #4466: Calling case class constructor without apply fails
```scala
object Foo {
  case class Bar(map: Map[String, String]) 

  object Bar {
    def apply(str: String): Bar = ???
  }

  Bar(Map("A" -> "B"))
}
```

gives
```scala
-- [E050] Reference Error: /tmp/scastie7649626593689591865/src/main/scala/main.scala:8:2 
8 |  Bar(Map("A" -> "B"))
  |  ^^^
  |  object Bar in object Foo does not take parameters
```

works if using apply
```scala
  Bar.apply(Map("A" -> "B"))
```

works without companion object
```scala
object Foo {
  case class Bar(map: Map[String, String]) 

  Bar(Map("A" -> "B"))
}
```

Medowhill:
The below code generates the same error.
```scala
class Foo {
  def apply(bar: Bar[Int]): Unit = ()
  def apply(i: Int): Unit = ()
}

class Bar[X](x: X)

class Main {
  val foo = new Foo
  foo(new Bar(0)) // should be compiled, but fails
  foo.apply(new Bar(0)) // works
  foo(new Bar[Int](0)) // works
}
```
When `foo(new Bar(0))` is typed, since there is no method with name `foo`, `tryInsertApplyOrImplicit` is used to try `foo.apply(new Bar(0))`. In `tryInsertApplyOrImplicit`, `tryEither` is invoked, so that a new context with a new typer state is created. However, `pt` of `FunProto` type uses the original context. Therefore, the original typer state contains constraints about `X` but the new typer state does not. Finally, overloading resolution fails because neither `(Bar[Int])Unit` nor `(Int)Unit` is a proper candidate for `(Bar[X])_`. The last two lines work because `tryEither` is not used and `X` does not need to be inferred respectively.
To make the same typer state be used, I added `tryEitherWithSameContext`, which changes only `reporter`, and use it.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -37,7 +37,7 @@ import config.Printers.{gadts, typr}
 import rewrite.Rewrites.patch
 import NavigateAST._
 import transform.SymUtils._
-import reporting.{StoreReporter, trace}
+import reporting.trace
 import config.Config
 
 import language.implicitConversions
@@ -1978,15 +1978,6 @@ class Typer extends Namer
   def typedPattern(tree: untpd.Tree, selType: Type = WildcardType)(implicit ctx: Context): Tree =
     typed(tree, selType)(ctx addMode Mode.Pattern)
 
-  def tryEitherWithSameContext[T](op: Context => T)(fallBack: T => T)(implicit ctx: Context) = {
-    val reporter = ctx.reporter
-    val reporter1 = new StoreReporter(reporter)
-    ctx.typerState.setReporter(reporter1)
-    val result = op(ctx)
-    ctx.typerState.setReporter(reporter)
-    if (reporter1.hasErrors) fallBack(result) else result
-  }
-
   def tryEither[T](op: Context => T)(fallBack: (T, TyperState) => T)(implicit ctx: Context) = {
     val nestedCtx = ctx.fresh.setNewTyperState()
     val result = op(nestedCtx)
@@ -2070,10 +2061,14 @@ class Typer extends Namer
         tree
       case _ =>
         if (isApplyProto(pt) || isMethod(tree) || isSyntheticApply(tree)) tryImplicit(fallBack)
-        else tryEitherWithSameContext(tryApply(_)) { app =>
+        else tryEither(tryApply(_)) { (app, appState) =>
           tryImplicit {
-            // issue the error about the apply, since it is likely more informative than the fallback
-            if (tree.tpe.member(nme.apply).exists) app else fallBack
+            if (tree.tpe.member(nme.apply).exists) {
+              // issue the error about the apply, since it is likely more informative than the fallback
+              appState.commit()
+              app
+            }
+            else fallBack
           }
         }
      }
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -37,7 +37,7 @@ import config.Printers.{gadts, typr}
 import rewrite.Rewrites.patch
 import NavigateAST._
 import transform.SymUtils._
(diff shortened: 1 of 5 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -2061,14 +2070,10 @@ class Typer extends Namer
         tree
       case _ =>
         if (isApplyProto(pt) || isMethod(tree) || isSyntheticApply(tree)) tryImplicit(fallBack)
-        else tryEither(tryApply(_)) { (app, appState) =>
+        else tryEitherWithSameContext(tryApply(_)) { app =>
```
allanrenucci: What about `tryAlternatively`?

allanrenucci:
> However, pt of FunProto type uses the original context. Therefore, the original typer state contains constraints about X but the new typer state does not

I don't think this is true. `setNewTyperState` create a copy of the current typer state. See https://github.com/lampepfl/dotty/blob/36c2e34bc33a39a41ec383e997185613bf88c64c/compiler/src/dotty/tools/dotc/core/TyperState.scala#L72-L73

Medowhill:
Yes, it does create a copy. However, I think that the follwing things happen:
FunProto `pt` has `ctx0` -> `typerState0` -> `constraint0`.
A new context created: `ctx1` -> `typerState1` -> `constraint0`.
`pt.typedArgs` invoked
--  `constraint1`, which constrain `X`, is created: `ctx0` -> `typerState0` -> `constraint1`.
`foo.apply` is typed with `ctx1` -> `typerState1` -> `constraint0`

Medowhill:
I reverted `typer.scala` to the original version to use `tryEither`. Instead of using the same context, now it copies constraints from `FunProto` before it narrows candidates.

allanrenucci:
You can achieve something similar more easily:
```diff
diff --git a/compiler/src/dotty/tools/dotc/typer/Applications.scala b/compiler/src/dotty/tools/dotc/typer/Applications.scala
index c74c0a1d0..93ef48f59 100644
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -1418,7 +1418,9 @@ trait Applications extends Compatibility { self: Typer with Dynamic =>
           if (isDetermined(alts2)) alts2
           else {
             pretypeArgs(alts2, pt)
-            narrowByTrees(alts2, pt.typedArgs, resultType)
+            val targs = pt.typedArgs
+            pt.ctx.typerState.commit()
+            narrowByTrees(alts2, targs, resultType)
           }
         }
 
diff --git a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
index e592d3b25..86fcb27c2 100644
--- a/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
+++ b/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala
@@ -214,7 +214,7 @@ object ProtoTypes {
    *
    *  [](args): resultType
    */
-  case class FunProto(args: List[untpd.Tree], resType: Type, typer: Typer)(implicit ctx: Context)
+  case class FunProto(args: List[untpd.Tree], resType: Type, typer: Typer)(implicit val ctx: Context)
   extends UncachedGroundType with ApplyingProto {
     private[this] var myTypedArgs: List[Tree] = Nil
```
But to me it looks like we are treating the symptom rather than the cause. Every time `FunProto::typedArgs` is called you potentially need to commit the `FunProto` typer state to the current context.

allanrenucci:
For reference in #2922 Guillaume changed `FunProto::typedArgs`  to  take an implicit `Context`

PR #4856, opened 2018-07-27
Fix #4846: Allow transparent parameters at any stage 

### LINKED ISSUE #4846: Allow transparent parameters at any staging level
Inline parameters are currently only allowed at level 0 and 1. But they should be available at any level.
```scala
transparent def foo(transparent x: Int): Int = ~fooImpl(x, '(x), '( '(x) ), '( '( '(x) ) ))
def fooImpl(a: Int, b: Expr[Int], c: Expr[Expr[Int]], d: Expr[Expr[Expr[Int]]]): Expr[Int] = ???
```
This makes inline values behave the same way as transparent members do.

In general, the PCP rule should be updated to say that inline values are not free variables of the quote.

nicolasstucki:
Based on #4823

* Stop checking and using capture transformation for transparent parameters
* Keep transparent parameters as they are. Allow them to be inlined and replaced directly.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
+++ b/compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
@@ -225,7 +225,7 @@ class ReifyQuotes extends MacroTransformWithImplicits {
     def levelOK(sym: Symbol)(implicit ctx: Context): Boolean = levelOf.get(sym) match {
       case Some(l) =>
         l == level ||
-        l == 0 && level == -1 && isStageNegOneValue(sym)
+        level == -1 && sym == defn.TastyTopLevelSplice_tastyContext
       case None =>
         !sym.is(Param) || levelOK(sym.owner)
     }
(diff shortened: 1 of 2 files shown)


nicolasstucki:
Rebased on master

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -359,18 +359,12 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
-            // in '{ x } to '{ x$1.unary_~ } and go directly to `x$1`
-            capturers(body.symbol)(body)
-          }
+          // Optimization: avoid the full conversion when capturing `x`
+          // in '{ x } to '{ x$1.unary_~ } and go directly to `x$1`
+          capturers(body.symbol)(body)
         case _=>
           val (body1, splices) = nested(isQuote = true).split(body)
-          if (level >= 0) pickledQuote(body1, splices, body.tpe, isType).withPos(quote.pos)
+          if (level == 0 && !ctx.inTransparentMethod) pickledQuote(body1, splices, body.tpe, isType).withPos(quote.pos)
```
biboudis: Small question. We don't split if we are in a transparent method but the level is above or equal zero?
If yes, we can move splitting inside the `if` now.

nicolasstucki: We do not split, but we still need to perform the checks that split does. We should be able to improve the performance by calling transform directly in the other branch. I will try that later, nice catch :)

PR #4857, opened 2018-07-27
Fix inline scripted

allanrenucci:
With the introduction of transparent, an implicit conversion from Double
to Int kicked in. We change the type to String to prevent this


smarter:
Setting prio:blocker because this is preventing the nightly builds from being published.

nicolasstucki:
`source-dependencies/value-class/test` can be minimized to 

```
$ copy-file changes/A0.scala src/main/scala/A.scala
$ copy-file changes/B0.scala src/main/scala/B.scala
$ copy-file changes/C0.scala src/main/scala/C.scala

> compile

$ copy-file changes/A1.scala src/main/scala/A.scala

-> compile

$ copy-file changes/A0.scala src/main/scala/A.scala
$ copy-file changes/B1.scala src/main/scala/B.scala
$ copy-file changes/C1.scala src/main/scala/C.scala

> compile
```

The issue seams to be that in the third `compile` we try to load `A` from the classfile of the first compilation and the tasty of the second compilation. It looks like after the failure to compile in the second  `compile` the classfile is reverted but the tasty file is kept.

allanrenucci:
Scripted tests pass at http://dotty-ci.epfl.ch/lampepfl/dotty/6640

allanrenucci:
I'll look into the newly failing test `source-dependencies/restore-classes`. But I think we can merge this to unblock nightly releases

PR #4860, opened 2018-07-28
Decouple transparent and erased properties

odersky:
It is at times useful to have `transparent` methods that are not
`erased`. For instance, a transparent `unapply` might enable inline
reductions, but we want to also keep it at runtime, because the same
type might be matched at compile-time and at run-time.

This commit decouples `transparent` and `erased` again. `transparent`
does not imply `erased`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/config/Config.scala
+++ b/compiler/src/dotty/tools/dotc/config/Config.scala
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = false
+  final val tracingEnabled = true
 
   /** Initial capacity of uniques HashMap.
    *  Note: This MUST BE a power of two to work with util.HashSet
--- a/compiler/src/dotty/tools/dotc/config/Config.scala
+++ b/compiler/src/dotty/tools/dotc/config/Config.scala
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = true
+  final val tracingEnabled = false
 
   /** Initial capacity of uniques HashMap.
    *  Note: This MUST BE a power of two to work with util.HashSet
--- a/compiler/src/dotty/tools/dotc/core/Flags.scala
+++ b/compiler/src/dotty/tools/dotc/core/Flags.scala
@@ -548,8 +548,8 @@ object Flags {
   /** Assumed to be pure */
   final val StableOrErased = Stable | Erased
 
-  /** Labeled `private`, `final`, or `transparent` */
-  final val EffectivelyFinal = Private | Final | Transparent
+  /** Labeled `private`, `final`, `transparent`, or `erasedd` */
+  final val EffectivelyFinal = Private | Final | Transparent | Erased
 
   /** A private method */
   final val PrivateMethod = allOf(Private, Method)
@@ -562,6 +562,9 @@ object Flags {
 
   /** A transparent parameter */
   final val TransparentParam = allOf(Transparent, Param)
+
+  /** A transparent erased method */
+  final val TransparentErasedMethod = allOf(Transparent, Erased, Method)
 
   /** An enum case */
   final val EnumCase = allOf(Enum, Case)
--- a/compiler/src/dotty/tools/dotc/core/Flags.scala
+++ b/compiler/src/dotty/tools/dotc/core/Flags.scala
@@ -548,7 +548,7 @@ object Flags {
   /** Assumed to be pure */
   final val StableOrErased = Stable | Erased
 
-  /** Labeled `private`, `final`, `transparent`, or `erasedd` */
+  /** Labeled `private`, `final`, `transparent`, or `erased` */
   final val EffectivelyFinal = Private | Final | Transparent | Erased
 
   /** A private method */
--- a/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
+++ b/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
@@ -761,7 +761,7 @@ class ClassfileParser(
           assert(cond,
             s"Unpickling ${classRoot.symbol.showLocated} from ${classRoot.symbol.associatedFile} is not allowed with -Yscala2-unpickler $allowed")
 
-        if (false && allowed != "always") {
+        if (allowed != "always") {
           failUnless(allowed != "never")
           val allowedList = allowed.split(":").toList
           val file = classRoot.symbol.associatedFile
--- a/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
+++ b/compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
@@ -761,7 +761,7 @@ class ClassfileParser(
           assert(cond,
             s"Unpickling ${classRoot.symbol.showLocated} from ${classRoot.symbol.associatedFile} is not allowed with -Yscala2-unpickler $allowed")
 
-        if (allowed != "always") {
+        if (false && allowed != "always") {
           failUnless(allowed != "never")
           val allowedList = allowed.split(":").toList
           val file = classRoot.symbol.associatedFile
--- a/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
+++ b/compiler/src/dotty/tools/dotc/typer/RefChecks.scala
@@ -143,16 +143,14 @@ object RefChecks {
    *    1.8.1  M's type is a subtype of O's type, or
    *    1.8.2  M is of type []S, O is of type ()T and S <: T, or
    *    1.8.3  M is of type ()S, O is of type []T and S <: T, or
-   *    1.9    M must not be an erased definition
-   *    1.10.  If M is a 2.x macro def, O cannot be deferred unless there's a concrete method overriding O.
-   *    1.11.  If M is not a macro def, O cannot be a macro def.
+   *    1.9    If M is an erased definition, it must override at least one concrete member
    *  2. Check that only abstract classes have deferred members
    *  3. Check that concrete classes do not have deferred definitions
    *     that are not implemented in a subclass.
    *  4. Check that every member with an `override` modifier
    *     overrides some other member.
    *  TODO check that classes are not overridden
-   *  TODO This still needs to be cleaned up; the current version is a staright port of what was there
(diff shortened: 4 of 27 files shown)


odersky:
I am still not sure about this. It seems we need two kinds transparent methods:

**Compile-time only**, i.e. `transparent` and `erased`. These methods

   - are inlined
   - don't generate code
   - allow implicit matches
   - allow macro splices
   - don't inline in their body before they are expanded themselves (this means recursion can be expressed safely)
   - cannot override other methods or they have a special semantics for it

**Compile-time + Run-time**, i.e. `transparent` but not `erased`. These methods:

   - are also inlined, but
   - do generate code
   - don't allow implicit matches or macro splices
   - do inline in their body before they are expanded themselves - this means recursion is basically impossible, since at the definition site there is usually not enough knowledge to prune a recursive call, which means a recursive method will expand until the limit is exceeded.
   - can override other methods

Examples of the first kind of method are all the typelevel functions discussed in #4616. An example of the second kind of method is a transparent `unapply` that needs to be inlined so that a toplevel match can "see through it" but that also should be retained at runtime to allow normal matching.

In this PR, the first kind of method is labelled `erased transparent`, the second just `transparent`. But `erased transparent` is a mouthful. Other possibilities include:

 - `macro` for the first kind, `transparent` for the second
 - `transparent` for the first kind, `translucent` for the second

Any other ideas?

odersky:
I discovered a tweak which, even though it is very small, might change the balance of power. The tweak is:

When processing the body of a transparent, unerased method we inline only transparent, erased methods. Transparent, unerased methods are not inlined, since they do have a runtime representation. That means we essentially have two modes: 

 1. If a transparent, unerased method is inlined, all calls to transparent methods are inlined in turn.
 2. But if it is left to be executed at runtime, we inline only the minimum (i.e. the erased methods) whereas all calls to other tranparent unerased methods are run-time calls.

This means that now we don't have to choose: `transparent` methods are fine even if they are not `erased`. They can do everything `transparent erased` methods can do, as long as they do not contain implicit matches or macro splices. So there is a clear criterion when to go to `transparent erased`, and, furthermore, it's a criterion that won't trigger too often. 

Let's redo the characterisation with this change:

**transparent erased methods**

   - are inlined
   - don't generate code
   - allow implicit matches
   - allow macro splices
   - don't inline in their body before they are expanded themselves 
   - cannot override other methods or they have a special semantics for it

**transparent methods**:

   - are also inlined, but
   - do generate code
   - don't allow implicit matches or macro splices
   - do inline erased transparents in their body before they are expanded themselves
   - can override other methods

odersky:
@nicolasstucki I believe the whole `Splicer.canbeSpliced` logic can be eliminated, including the overhead of defining an abstract interpreter. A transparent method is marked as a macro (now already in Namer/Typer) if it contains a toplevel splice. Macros process their right hand side, which is known to be a splice. Other splices outside quotes are errors. That should be it. We don't need a separate testing logic here.

nicolasstucki:
To allow the use of `erased transparent` in a `transparent` we also should not allow the `transparent` (un-erased) method to have `transparent` parameters. Otherwise, when inlining the `erased transparent` in the `transparent` method body the parameter will not be a literal constant. This would cause problems for macros.

nicolasstucki:
> @nicolasstucki I believe the whole Splicer.canbeSpliced logic can be eliminated, including the overhead of defining an abstract interpreter. A transparent method is marked as a macro (now already in Namer/Typer) if it contains a toplevel splice. Macros process their right hand side, which is known to be a splice. Other splices outside quotes are errors. That should be it. We don't need a separate testing logic here.

`Splicer.canbeSpliced` does not check if the top-level splice is a valid top-level splice. Instead it tests if the contents of a valid top level splice will be interpretable at inline site. Though this new check will help remove some other logic that checks the top-level splices.

compiler/src/dotty/tools/dotc/config/Config.scala
```diff
@@ -160,7 +160,7 @@ object Config {
   final val showCompletions = false
 
   /** If set, enables tracing */
-  final val tracingEnabled = false
+  final val tracingEnabled = true
```
nicolasstucki: Was this enabled on purpose?

compiler/src/dotty/tools/dotc/core/Flags.scala
```diff
@@ -548,8 +548,8 @@ object Flags {
   /** Assumed to be pure */
   final val StableOrErased = Stable | Erased
 
-  /** Labeled `private`, `final`, or `transparent` */
-  final val EffectivelyFinal = Private | Final | Transparent
+  /** Labeled `private`, `final`, `transparent`, or `erasedd` */
```
nicolasstucki: Typo `erasedd` -> `erased`.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
```diff
@@ -761,7 +761,7 @@ class ClassfileParser(
           assert(cond,
             s"Unpickling ${classRoot.symbol.showLocated} from ${classRoot.symbol.associatedFile} is not allowed with -Yscala2-unpickler $allowed")
 
-        if (allowed != "always") {
+        if (false && allowed != "always") {
```
nicolasstucki: Should be removed

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -376,8 +376,8 @@ object RefChecks {
         overrideError("may not override a non-lazy value")
       } else if (other.is(Lazy) && !other.isRealMethod && !member.is(Lazy)) {
         overrideError("must be declared lazy to override a lazy value")
-      } else if (member.is(Erased) && member.allOverriddenSymbols.forall(_.is(Deferred))) { // (1.9)
-        overrideError("is an erased method, may not override only deferred methods")
+      } else if (member.is(Erased)) { // (1.9)
+        overrideError("is an erased method, may not override anything")
       } else if (member.is(Macro, butNot = Scala2x)) { // (1.9)
         overrideError("is a macro, may not override anything")
```
nicolasstucki: Unreachable code. This case is covered be the previous one.

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
```diff
@@ -143,7 +143,7 @@ object RefChecks {
    *    1.8.1  M's type is a subtype of O's type, or
    *    1.8.2  M is of type []S, O is of type ()T and S <: T, or
    *    1.8.3  M is of type ()S, O is of type []T and S <: T, or
-   *    1.9    M must not be a typelevel def or a Dotty macro def
+   *    1.9    M must not be an erased definition
    *    1.10.  If M is a 2.x macro def, O cannot be deferred unless there's a concrete method overriding O.
    *    1.11.  If M is not a macro def, O cannot be a macro def.
```
nicolasstucki: The `1.11` rule should probably be for `erased` in general.

odersky:
> To allow the use of erased transparent in a transparent we also should not allow the transparent (un-erased) method to have transparent parameters. Otherwise, when inlining the erased transparent in the transparent method body the parameter will not be a literal constant. This would cause problems for macros.

But macros are always implicitly erased. And I don't think there is a problem for normal transparent functions.

compiler/src/dotty/tools/dotc/typer/RefChecks.scala
odersky: In fact, 1.11 is not needed at all, since macros are erased, hence effectively final.

compiler/src/dotty/tools/dotc/core/classfile/ClassfileParser.scala
odersky: Ah, I needed that because my local setup does not run otherwise. I have to inquire why.

compiler/src/dotty/tools/dotc/config/Config.scala
odersky: No, I mistakenly added two files to the commit that should not be there.

odersky:
The latest commit brings back the rule that erased methods may override other methods, as long as at least one of the overridden methods must be concrete. Scala 2 has the same rule for macros. I believe it's a useful rule to have, since otherwise we can never re-implement an inherited method with macro-generated code. We can think about tightening the rule later.

nicolasstucki:
>> To allow the use of erased transparent in a transparent we also should not allow the transparent (un-erased) method to have transparent parameters. Otherwise, when inlining the erased transparent in the transparent method body the parameter will not be a literal constant. This would cause problems for macros.

> But macros are always implicitly erased. And I don't think there is a problem for normal transparent functions.

The issue arises with the combination of both

```scala
transparent def foo(transparent i: Int): Int = bar(i)
erased transparent def bar(transparent i: Int): Int = ~baz(i)
def baz(i: Int): Int = i
```

In this case, for the body of `foo` (retained at runtime) we will inline `bar` getting
```scala
transparent def foo(transparent i: Int): Int = ~baz(i)
```
but now we cannot evaluate `~baz(i)` as `i` is still unknown.

nicolasstucki:
Test `tests/run/i4803d` is failing because the macro `power` is not marked as erased and the splice survives `pruneErasedDefs`. 

Then there is a second problem, in the body of `power2` the method `power` is not inlined. In general, we should probably inline `erased transparent` in the RHS of `transparent` methods.
This issue is not only present due to the nesting, it can also be reproduced if the methods are defined as:
```scala
transparent def power2(x: Double) = power(x, 2)
erased transparent def power(x: Double, transparent n: Long) = ~PowerMacro.powerCode('(x), n)
```

We will probably have the same issues for type level matches in `transparent` methods.

nicolasstucki:
Deciding if a `transparent` is erased or not based on the contents of the right-hand side of the method is a bit ad-hoc. I would propose trying another approach that would keep transparent methods based on the `override` keyword. The idea would be that by default all `transparent` methods are `erased`. `transparent` method would not be `erased` only if it is a `override transparent` which implies that it implements a member (concrete or abstract) and is needed at runtime.

The complete rules would be
* All `transparent` are `final`
* All local `transparent` methods are erased (as they do not override any method)
* `transparent` methods are erased
  * may have implicit matches
  * may have top-level `~`
  * all calls must be inlined, (including calls in RHS of `override transparent`)
  * does not implement/override any other method
  * internally has the `erased` flag
* `override transparent` are not erased
  * cannot have implicit matches
  * cannot have top-level `~`
  * all calls to the macro should be inlined
  *  implements/overrides other method

odersky:
Closed for now, in favor of #4927.

PR #4861, opened 2018-07-29
Fix #4306: Implement -scansource for Java files

### LINKED ISSUE #4306: -scansource does not handle Java files correctly
Currently, -scansource uses OutlineParser which extends the regular Scala parser, so we get parser errors when it encounters a Java file. We should either skip Java files or make a Java version of OutlineParser.

Blaisorblade:
Fix #4306. Also document `OutlineParser` with what I needed to discover.

### DIFF
--- a/compiler/src/dotty/tools/dotc/parsing/JavaParsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/JavaParsers.scala
@@ -916,7 +916,7 @@ object JavaParsers {
    */
   class OutlineJavaParser(source: SourceFile)(implicit ctx: Context) extends JavaParser(source) {
 
-    def skipBraces[T](body: T): T = {
+    def skipBraces(): Unit = {
       accept(LBRACE)
       var openBraces = 1
       while (in.token != EOF && openBraces > 0) {
@@ -924,10 +924,11 @@ object JavaParsers {
         else if (in.token == RBRACE) openBraces -= 1
         in.nextToken()
       }
-      body
     }
 
-    override def typeBody(leadingToken: Int, parentName: Name, parentTParams: List[TypeDef]): (List[Tree], List[Tree]) =
-      skipBraces((List(EmptyValDef), List(EmptyTree)))
+    override def typeBody(leadingToken: Int, parentName: Name, parentTParams: List[TypeDef]): (List[Tree], List[Tree]) = {
+      skipBraces()
+      (List(EmptyValDef), List(EmptyTree)))
+    }
   }
 }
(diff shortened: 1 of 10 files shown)


compiler/src/dotty/tools/dotc/parsing/JavaParsers.scala
```diff
@@ -907,4 +907,28 @@ object JavaParsers {
...
+
+  /**
+   * OutlineJavaParser parses top-level declarations in `source` to find declared classes, ignoring their bodies (which
+   * must only have balanced braces). This is used to map class names to defining sources.
+   * This is necessary even for Java, because the filename defining a non-public classes cannot be determined from the
+   * classname alone.
+   */
+  class OutlineJavaParser(source: SourceFile)(implicit ctx: Context) extends JavaParser(source) {
+
+    def skipBraces[T](body: T): T = {
```
allanrenucci: I realise it was taken from the Scala `OutlineParser `, but I find it a bit weird that this method takes an argument and just returns it. I would define it and use as:
```scala
private def skipBraces(): Unit = ...
override def typeBody(leadingToken: Int, parentName: Name, parentTParams: List[TypeDef]) = {
  skipBraces()
  (List(EmptyValDef), List(EmptyTree))
}
  
```
Maybe change the implementation in the Scala OutlineParser as well.

compiler/src/dotty/tools/dotc/parsing/JavaParsers.scala
```diff
@@ -907,4 +907,28 @@ object JavaParsers {
...
   }
+
+
+  /**
+   * OutlineJavaParser parses top-level declarations in `source` to find declared classes, ignoring their bodies (which
+   * must only have balanced braces). This is used to map class names to defining sources.
+   * This is necessary even for Java, because the filename defining a non-public classes cannot be determined from the
+   * classname alone.
+   */
+  class OutlineJavaParser(source: SourceFile)(implicit ctx: Context) extends JavaParser(source) {
```
allanrenucci: Nitpick: the documentation style we (try to) use across the code base is:
```scala
/** First line
 *
 *  More lines...
 *  @param ...
 */
def foo = 
```
We start on the first line and all lines are alligned

compiler/src/dotty/tools/dotc/parsing/JavaParsers.scala
smarter: This is identical to the Scala version except for the lack of XMLSTART handling, can't we just use the Scala version for Java too ?

Blaisorblade:
All comments addressed, without too much complication. @allanrenucci last commit is a refactoring that might be worth reviewing for style, the new code is a bit more convoluted.

PR #4865, opened 2018-07-29
Fix tasty hash

nicolasstucki:
There were two bugs
* The hash of the `nameBuffer` was not computed
* The computed hash was 32-bit instead of 64-bit

The combination of the two bugs made the 128-bit UUID have `0`s in the first 96 bits.

### DIFF
--- a/compiler/src/dotty/tools/dotc/core/tasty/TastyPickler.scala
+++ b/compiler/src/dotty/tools/dotc/core/tasty/TastyPickler.scala
@@ -26,8 +26,8 @@ class TastyPickler(val rootCls: ClassSymbol) {
     nameBuffer.assemble()
     sections.foreach(_._2.assemble())
 
-    val nameBufferHash = pjwHash64(nameBuffer.bytes)
-    val treeSectionHash +: otherSectionHashes = sections.map(x => pjwHash64(x._2.bytes))
+    val nameBufferHash = TastyHash.pjwHash64(nameBuffer.bytes)
+    val treeSectionHash +: otherSectionHashes = sections.map(x => TastyHash.pjwHash64(x._2.bytes))
 
     // Hash of name table and tree
     val uuidLow: Long = nameBufferHash ^ treeSectionHash
@@ -78,21 +78,4 @@ class TastyPickler(val rootCls: ClassSymbol) {
 
   val treePkl = new TreePickler(this)
 
-  /** Returns a non-cryptographic 64-bit hash of the array.
-   *
-   *  from https://en.wikipedia.org/wiki/PJW_hash_function#Algorithm
-   */
-  private def pjwHash64(data: Array[Byte]): Long = {
-    var h = 0L
-    var i = 0
-    while (i < data.length) {
-      val d = data(i) & 0xFFL // Interpret byte as unsigned byte
-      h = (h << 8) + d
-      val high = h & 0xFF00000000000000L
-      h ^= high >> 48L
-      h &= ~high
-      i += 1
-    }
-    h
-  }
 }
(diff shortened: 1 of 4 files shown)


smarter:
So how come this didn't cause any test to fail so far (except the scripted test I guess)? And can we add a test for this?

nicolasstucki:
This has nothing to do with the scripted tests. I just noted that there were too many zeros in the hash.

nicolasstucki:
> Xoring something with 0 does not change the value so I don't see the point of the if
True, I'm trying to figure out why this implementation pjwHash is only returning `Int`s

compiler/src/dotty/tools/dotc/core/tasty/TastyPickler.scala
```diff
@@ -74,17 +80,19 @@ class TastyPickler(val rootCls: ClassSymbol) {
...
     var i = 0
     while (i < data.length) {
-      h = (h << 4) + data(i)
-      high = h & 0xF0000000L
-      h ^= high >> 24
-      h &= ~high
+      val d = data(i) & 0xFFL // Interpret byte as unsigned byte
+      h = (h << 8) + d
+      val high = h & 0xFF00000000000000L
+      if (high != 0) {
```
smarter: Remove the if, it's not necessary and branches should be avoided in tight loops.

nicolasstucki: Done

nicolasstucki:
@smarter don't forget to finish the review on this PR

Blaisorblade:
The review on the algorithm can be separated to offload @smarter.
Can we test the input on reference data against some reference implementation?

I've taken two difference reference C implementations, ported them to 64 bits following Wikipedia and its DrDobbs source, and gotten 0x34545c16020230 as hash for
`"abcdefghijklmnopqrstuvwxyz1234567890"`. With the code (I'll upload it), we can test more input data. Let's.

https://github.com/lampepfl/dotty/pull/2272#issuecomment-299182451
> The context for using Elf hashing instead of CityHash was because the implementation is trivial and it is simpler to be sure that it is correctly implemented.

tests/run-with-compiler/tasty-hash.scala
```diff
@@ -0,0 +1,25 @@
+import dotty.tools.dotc.core.tasty.TastyHash.pjwHash64
+
+object Test {
```
smarter: I would write a junit test instead, they're faster to run.

tests/run-with-compiler/tasty-hash.scala
```diff
@@ -0,0 +1,25 @@
...
+    testHash(0X55aa01fe55ab54ffL, Array.tabulate(1024)(_.toByte))
+    testHash(0x34545c16020230L, "abcdefghijklmnopqrstuvwxyz1234567890".getBytes)
+  }
+
+  def testHash(expected: Long, arr: Array[Byte]): Unit = {
+    val res = pjwHash64(arr)
+    assert(res == expected, s"Exprected 0x${expected.toHexString}L but got 0X${res.toHexString}L")
+  }
+
+}
```
Blaisorblade: It seems simpler and more appropriate to use a unit test here, rather than only doing this after bootstrap.

Blaisorblade:
FWIW I fixed a hash bug after @smarter's review — but the hashing details can be reviewed by non-Tasty experts.

tests/run-with-compiler/tasty-hash.scala
nicolasstucki: Done

nicolasstucki:
@Blaisorblade fix LGTM

PR #4869, opened 2018-07-30
Fix #4661: Missing unreachable case warnings

### LINKED ISSUE #4661: Missing unreachable case warnings
Given:
```scala
trait Foo
class One extends Foo
class Two extends Foo
class Three extends Foo

object Test {
  def test(f: Foo) = f match {
    case f: Foo =>
    case f: One =>
    case f: Two =>
    case f: Three =>
  }
}
```
We should get three warnings since the three last cases are not reachable, but we only get one for the last case:
```scala
-- [E120] Only null matched Warning: try/caseun.scala:11:9 ---------------------
11 |    case f: Three =>
   |         ^^^^^^^^
   |         Only null is matched. Consider using `case null =>` instead.
```
(The warning itself is a bit confusing but that's a separate issue: https://github.com/lampepfl/dotty/issues/4660)

abeln:


abeln:
@liufengyun this is the fix we discussed in #4661 but it introduces a failure in https://github.com/lampepfl/dotty/blob/ba2eafcf9e7f4d166fe3721d81506a0ab0fb3adf/tests/patmat/file.scala

The failure is that we now issue a warning that only null is matched for the last case:
```
-- [E120] Only null matched Warning: tests/patmat/file.scala:13:9 --------------
13 |    case _ =>
   |         ^
   |         Only null is matched. Consider using `case null =>` instead.
```

However, looking at the code it's clear that the wildcard case matches more than `null` (to start with, `AbstractFile` is not sealed).

If we look at that last wildcard case, we can print what's covered before and what's covered by the wildcard:

```
prev: _: ZipArchive.Entry, _: PlainFile
covered: Or(List(Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class AbstractFile),false), Typ(ConstantType(Constant(null)),true)))
```

which seems correct. However, the problem appears when we substract what's already covered from what the `_` adds:

```
minus: a = Or(List(Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class AbstractFile),false), Typ(ConstantType(Constant(null)),true))) b = Or(List(Typ(TypeRef(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class ZipArchive),class Entry),true), Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class PlainFile),true)))

minus: a = Or(List(Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class AbstractFile),false), Typ(ConstantType(Constant(null)),true))) b = Typ(TypeRef(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class ZipArchive),class Entry),true)

minus: a = Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class AbstractFile),false) b = Typ(TypeRef(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class ZipArchive),class Entry),true)
```

We can see that the `Or`s are unrolled from both sides, until we're subtracting `Typ(TypeRef(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class ZipArchive),class Entry)` from `Typ(TypeRef(ThisType(TypeRef(NoPrefix,module class <empty>)),class AbstractFile),false)`. 

The result is empty,  which is incorrect.

In particular, when we try to decompose the `ZipArchive#Entry` (because it's sealed), we get
```
candidates for ZipArchive#Entry : [class DirEntry]
class DirEntry instantiated ------> WildcardType(NoType)
```

The first line seems correct, but I don't understand the second line, and I think it's incorrect and caused by a bug in `refine`/`instantiate`. However, I don't understand well enough either of those functions to debug the issue.

In summary, I think this change uncovered an already-existing bug. Thoughts?

abeln:
I'd propose that I patch the failing test and file a separate bug to investigate `instantiate`/`refine`. I'd be happy to look into it with some guidance.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -294,16 +294,21 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
 
   override def intersectUnrelatedAtomicTypes(tp1: Type, tp2: Type) = {
     val and = AndType(tp1, tp2)
```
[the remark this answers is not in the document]
abeln: Done

Blaisorblade:
My 2 cents: even if this uncovers an existing bug (and I agree with you that it seems to be the case), that bug seems serious enough (or at least, worse than the one fixed here) that we might want to block this fix, at least until we understand what is going on. After all, the failing test does not look *that* esoteric.

liufengyun:
@abeln I confirm that there is a problem with the handling of `ZipArchive#Entry` in `tests/patmat/file.scala`.  I think as long as we open an issue for `tests/patmat/file.scala`, it's fine to disable the test in this PR.

Thanks a lot.

abeln:
Filed https://github.com/lampepfl/dotty/issues/4880 for the related bug.

abeln:
@liufengyun disabled the problematic part of the failing test.

abeln:
There are a bunch of warnings (that I guess are converted into errors) from `TastyImpl.scala` (https://github.com/lampepfl/dotty/blob/e72b80b85d2f939cb24707058fbb0c6ce657398f/compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala#L64)

For example
```
  type Id = untpd.Ident
...
  object Id extends IdExtractor {
    def unapply(x: Id): Option[String] = x match {
      case x: untpd.Ident => Some(x.name.toString) // TODO how to make sure it is not a Ident or TypeIdent? Check x.tpe?
      case _ => None
    }
}
```

Why is the code written in this style (i.e. the type alias and then the pattern matching)? What's the goal of the wildcard `case`? Is it really to match `null`? Looping in @nicolasstucki who I think wrote that code.

allanrenucci:
> Why is the code written in this style (i.e. the type alias and then the pattern matching)? What's the goal of the wildcard case? Is it really to match null? Looping in @nicolasstucki who I think wrote that code.

This seems to be  from historical reasons. You can update the code

nicolasstucki:
You can also remove the comment.

abeln:
@liufengyun 
I fixed all the warnings in `TastyImpl`, but the one in `ExtractDependencies` line 132 is a "real life" version of the test I previously disabled: `file.scala`: https://github.com/lampepfl/dotty/blob/36c2e34bc33a39a41ec383e997185613bf88c64c/compiler/src/dotty/tools/dotc/sbt/ExtractDependencies.scala#L132

I re-enabled the test.

So I think this PR is blocked by https://github.com/lampepfl/dotty/issues/4880

nicolasstucki:
The changes in `TastyImpl.scala` look good.

liufengyun:
@abeln can you try rebase against #4897 ?

nicolasstucki:
@abeln #4904 will change the code in `TastyImpl.scala` a large part of the changes you did in that file might not be needed.

abeln:
@liufengyun after rebasing over #4897 both the failing test and the bootstrapped compilation now succeed. So I'll wait for #4897 to get merged before updating the PR.

Blaisorblade:
@abeln Just merged #4897, and #4904 was merged few days ago.

Might be good to also test this keeps working when types are defined under a prefix, for instance
```scala
trait Prefix {
  trait Foo
  class One extends Foo
  class Two extends Foo
  class Three extends Foo
} 
class Test(p: Prefix) {
  import p._
  def test(f: Foo) = f match {
    case f: Foo =>
    case f: One =>
    case f: Two =>
    case f: Three =>
  }
}
```
Matching against `Prefix#Foo` and/or `Prefix#One` (and combinations thereof) is also interesting, since that stresses out the subtle logic in #4897.

tests/patmat/i4880a.check
```diff
@@ -0,0 +1 @@
+14: Match case Unreachable
```
abeln: This is new, and related to #4880, but seems correct: https://github.com/lampepfl/dotty/blob/master/tests/patmat/i4880a.scala#L14

tests/patmat/t4661.scala
```diff
@@ -0,0 +1,61 @@
+trait Foo
```
abeln: @Blaisorblade I can break this up into smaller tests, if you want. Not sure want the convention is.

compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -292,18 +292,23 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
-
-    if (!res) Empty
-    else if (tp1.isSingleton) Typ(tp1, true)
-    else if (tp2.isSingleton) Typ(tp2, true)
-    else Typ(and, true)
+  override def intersectUnrelatedAtomicTypes(tp1: Type, tp2: Type) = {
+    // Precondition: !isSubType(tp1, tp2) && !isSubType(tp2, tp1)
+    if (tp1 == nullType || tp2 == nullType) {
+      // Since projections of types don't include null, intersection with null is empty.
+      Empty
```
liufengyun: A minor comment: it seems more readable if we just `return Empty` for the exceptional case, fewer braces and indentations.

liufengyun:
LGTM, thanks a lot @abeln !

PR #4870, opened 2018-07-30
Add some post conditions to ReifyQuotes

nicolasstucki:


compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -74,6 +74,21 @@ class ReifyQuotes extends MacroTransformWithImplicits {
 
   override def phaseName: String = "reifyQuotes"
 
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = {
+    if (ctx.owner.ownersIterator.exists(_.isTransparentMethod)) {
```
allanrenucci: Does it make sense to do these tests if `ctx.compilationUnit.containsQuotesOrSplices` is false?

nicolasstucki: Yes, some other phase could inset by mistake one of those calls.

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -74,6 +74,21 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+        case tree: Select if tree.symbol == defn.QuotedExpr_~ =>
+          assert(Splicer.canBeSpliced(tree.qualifier))
+        case _ =>
+      }
+    } else if (!tree.isInstanceOf[DefDef]) {
+      assert(tree.symbol != defn.QuotedExpr_~)
+      // assert(tree.symbol != defn.QuotedType_~) // TODO widen ~ type references at stage 0?
+      assert(tree.symbol != defn.QuotedExpr_apply)
+      assert(tree.symbol != defn.QuotedType_apply)
+    }
```
allanrenucci: Could this be rewritten as?
```scala
tree match {
  case tree: RefTree if !ctx.inTransparentMethod =>
    assert(!tree.symbol.isQuote)
    // assert(!tree.symbol.isSplice) // TODO widen ~ type references at stage 0?
    assert(tree.symbol != defn.QuotedExpr_~)
  case tree: Select if tree.symbol == defn.QuotedExpr_~ =>
    assert(Splicer.canBeSpliced(tree.qualifier))
  case _ =>
}
```

`ctx.inTransparentMethod` is more expensive than a type test

nicolasstucki: Changed

compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -74,6 +74,19 @@ class ReifyQuotes extends MacroTransformWithImplicits {
 
   override def phaseName: String = "reifyQuotes"
 
+  override def checkPostCondition(tree: Tree)(implicit ctx: Context): Unit = {
+    val inTransparentMethod = ctx.owner.ownersIterator.exists(_.isTransparentMethod)
+    tree match {
+      case tree: RefTree if !inTransparentMethod =>
```
allanrenucci: Why not use `ctx.inTransparentMethod` here?

PR #4877, opened 2018-07-31
Fix Tasty reflect with varargs in unapply

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -587,10 +587,14 @@ class TastyImpl(val rootContext: Contexts.Context) extends scala.tasty.Tasty { s
...
       def unapply(x: Pattern)(implicit ctx: Context): Option[(Term, List[Term], List[Pattern])] = x match {
-        case Trees.UnApply(fun, implicits, patterns) => Some((fun, implicits, patterns))
-        case Trees.Typed(Trees.UnApply(fun, implicits, patterns), _) => Some((fun, implicits, patterns))
+        case Trees.UnApply(fun, implicits, patterns) => Some((fun, implicits, effectivePatterns(patterns)))
+        case Trees.Typed(Trees.UnApply(fun, implicits, patterns), _) => Some((fun, implicits, effectivePatterns(patterns)))
         case _ => None
       }
+      private def effectivePatterns(patterns: List[Pattern]): List[Pattern] = patterns match {
+        case patterns0 :+ Trees.SeqLiteral(elems, _) => patterns0 ::: elems
+        case _ => patterns
```
liufengyun: Is it possible that `patterns` only contain `SeqLiteral`?

nicolasstucki: Only the last one can one can have the `SeqLiteral` as only the last argument can be a varargs.

liufengyun: 1. It's nice to have a link to test case here, as it's really difficult to figure out the test case
2. The code fails to handle the following case:

```Scala
case class D(b: B[_]*)

class Test {
  def f2(d: D) = d match {
    case D(a, b1, b2) => ()
  }
}
```

PR #4887, opened 2018-08-02
Add <:< and =:= to testy reflect Type

nicolasstucki:


library/src/scala/tasty/Tasty.scala
```diff
@@ -591,6 +591,12 @@ abstract class Tasty { tasty =>
 
   type Type <: TypeOrBounds
 
+  trait TypeAPI {
+    def =:=(other: Type)(implicit ctx: Context): Boolean
+    def <:<(other: Type)(implicit ctx: Context): Boolean
+  }
+  implicit def TypeDeco(tpe: Type): TypeAPI
```
Blaisorblade: Not for this PR, but we were thinking about making the implicit defs inline — did we forget?

nicolasstucki: I will do a larger refactor later. Now we don't have a standard way to inline both on bootstrap and non bootstraped

PR #4888, opened 2018-08-03
 Fix scripted/source-dependencies/restore-classes

allanrenucci:
TastyFileManager should delete TASTY files and not class files


allanrenucci:
Scripted tests pass a http://dotty-ci.epfl.ch/lampepfl/dotty/6655

PR #4894, opened 2018-08-04
Fix #4891: Make scala.quoted.Expr covariant

### LINKED ISSUE #4891: Pikled expression failing to Ycheck
```scala
def foo: Expr[Option[String]] = '(None)
```
fails Ycheck with
```scala
checking xyz.scala after phase reifyQuotes
exception while typing def foo: quoted.Expr[Option[String]] = 
  scala.runtime.quoted.Unpickler.unpickleExpr[None](
    scala.collection.immutable.Nil.::[String](
      
        "\\¡«\037\0212\0200\00\00\00\00\00\00\00\00\00\00\00\00\03â\06°\0223\01\0204ASTs\01\0204None\01\0205scala\0200\0204s\02016\0202"
      
    )
  , scala.collection.immutable.Nil) of class class dotty.tools.dotc.ast.Trees$DefDef # 673
```

Where `unpickleExpr[None]` returns an `Expr[None]`. Probably the `T` in `Expr` should be covariant.

nicolasstucki:


Blaisorblade:
LGTM and merged then, since we already have the other test.

library/src/scala/quoted/Expr.scala
```diff
@@ -2,7 +2,7 @@ package scala.quoted
 
 import scala.runtime.quoted.Unpickler.Pickled
 
-sealed abstract class Expr[T] {
+sealed abstract class Expr[+T] {
```
smarter: I think it'd make sense to make the subclasses covariant too for consistency.

nicolasstucki: Yes, I will add apt them

nicolasstucki: Done in #4896

PR #4895, opened 2018-08-05
Add standard definitions on Tasty reflect

nicolasstucki:
Based on scala.reflect


compiler/src/dotty/tools/dotc/tastyreflect/FromSymbol.scala
```diff
@@ -1,37 +1,40 @@
...
 
 import dotty.tools.dotc.ast.tpd
-import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.core.Flags._
+import dotty.tools.dotc.core.StdNames._
 import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Types._
 
-object FromSymbol {
+trait FromSymbol { tasty: TastyImpl =>
```
liufengyun: It seems `tasty: TastyImpl` is not used in the implementation? What's the motivation to change from `object` to `trait`?

library/src/scala/tasty/reflect/StandardDefinitions.scala
```diff
@@ -0,0 +1,208 @@
...
+      */
+    def ScalaNumericValueClasses: List[ClassDef]
+  }
+
+  /** Defines standard types.
+    *  @group Definitions
+    */
+  trait StandardTypes {
+    /** The type of primitive type `Unit`. */
+    def UnitTpe: Type
```
liufengyun: For readability, I prefer `UnitType` instead of `UnitTpe`. The same for the other names.

compiler/src/dotty/tools/dotc/tastyreflect/FromSymbol.scala
nicolasstucki: This change is not needed, I will revert it

library/src/scala/tasty/reflect/StandardDefinitions.scala
nicolasstucki: Changed

compiler/src/dotty/tools/dotc/tastyreflect/FromSymbol.scala
nicolasstucki: Reverted

PR #4897, opened 2018-08-05
Fix #4880: always instantiate prefix tvar

### LINKED ISSUE #4880: Incorrect Space difference in pattern matching exhaustivity
Details here: https://github.com/lampepfl/dotty/pull/4869#issuecomment-408863786

liufengyun:
Fix #4880: always instantiate prefix tvar

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -292,7 +292,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   private val nullType             = ConstantType(Constant(null))
   private val nullSpace            = Typ(nullType)
 
-  override def intersectUnrelatedAtomicTypes(tp1: Type, tp2: Type) = {
+  override def intersectUnrelatedAtomicTypes(tp1: Type, tp2: Type): Space = {
     val and = AndType(tp1, tp2)
     // Precondition: !(tp1 <:< tp2) && !(tp2 <:< tp1)
     // Then, no leaf of the and-type tree `and` is a subtype of `and`.
@@ -689,7 +689,8 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       tvar =>
         !(ctx.typerState.constraint.entry(tvar.origin) `eq` tvar.origin.underlying) ||
         (tvar `eq` removeThisType.prefixTVar),
-      minimizeAll = false
+      minimizeAll = false,
+      allowBottom = false
     )
 
     // If parent contains a reference to an abstract type, then we should
--- a/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -656,18 +656,18 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
     // Fix subtype checking for child instantiation,
     // such that `Foo(Test.this.foo) <:< Foo(Foo.this)`
     // See tests/patmat/i3938.scala
-    def removeThisType(implicit ctx: Context) = new TypeMap {
-      // is in tvarBounds? Don't create new tvars if true
-      private var tvarBounds: Boolean = false
+    class RemoveThisMap extends TypeMap {
+      var prefixTVar: Type = null
       def apply(tp: Type): Type = tp match {
         case ThisType(tref: TypeRef) if !tref.symbol.isStaticOwner =>
           if (tref.symbol.is(Module))
             TermRef(this(tref.prefix), tref.symbol.sourceModule)
-          else if (tvarBounds)
+          else if (prefixTVar != null)
             this(tref)
           else {
-            tvarBounds = true
-            newTypeVar(TypeBounds.upper(this(tref)))
+            prefixTVar = WildcardType  // prevent recursive call from assigning it
+            prefixTVar = newTypeVar(TypeBounds.upper(this(tref)))
+            prefixTVar
           }
         case tp => mapOver(tp)
       }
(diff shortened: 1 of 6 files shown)


compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
```diff
@@ -681,14 +681,17 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
...
 
+    val removeThisType = new RemoveThisMap
+    val tvars = tp1.typeParams.map { tparam => newTypeVar(tparam.paramInfo.bounds) }
+    val protoTp1 = removeThisType.apply(tp1).appliedTo(tvars)
+
     val force = new ForceDegree.Value(
-      tvar => !(ctx.typerState.constraint.entry(tvar.origin) eq tvar.origin.underlying),
+      tvar =>
+        !(ctx.typerState.constraint.entry(tvar.origin) eq tvar.origin.underlying) ||
+        (tvar `eq` removeThisType.prefixTVar),
```
Blaisorblade: Why a backquoted `eq` here (or elsewhere for that matter)?

liufengyun: I should use it in both cases, just following a coding style experimented by Martin in Dotty.

Blaisorblade:
So I've pushed a couple of examples, and a patch I'm not happy about for them. Below's why.

Enabling `Printers.exhaustivity` reveals that `ZipArchive#Entry decomposes to []`, which is not correct. Varying the example a bit demonstrates an incorrect warning:

```scala
sealed abstract class AbstractFile
class PlainFile(path: String) extends AbstractFile
class VirtualFile(name: String) extends AbstractFile
abstract class ZipArchive(path: String) extends AbstractFile {
  sealed abstract class Entry(name: String) extends VirtualFile(name)
  class DirEntry(path: String) extends Entry(path)
}

object Test {
  def foo(file: AbstractFile) =  file match {
    case a: PlainFile =>
    case b: ZipArchive =>
    case c1: ZipArchive#Entry => // marked as unreachable, incorrectly!
    case c1: ZipArchive#DirEntry => // accepted, correctly, even tho this is the instance of ZipArchive#Entry
    case c: VirtualFile =>
  }
}
```

## Analysis

Once instantiate finds that `?X#DirEntry` can be an instance of `ZipArchive#Entry`, it must pick `?X`. The new code maximizes it to `Nothing`, but then we get `Nothing#DirEntry inhabited?  false`. The old code turns `?X#DirEntry` into `WildcardType`. Neither's correct.

I can add a special case for `inhabited(TypeRef(Nothing, member)) = true`, and that passes tests, but I'm not very happy with it — we'd risk running again into 3574 if, say, intersections were smarter.

I suspect we should produce something else instead of `Nothing#DirEntry`: I'll look into that next.

```
Space.instantiate: protoTp1((param)9#DirEntry) <:< tp2(ZipArchive#Entry) = true
Space.instantiate: isFullyDefined((param)9#DirEntry) = false
`Nothing#DirEntry inhabited?  false`
```

liufengyun:
@Blaisorblade  The following change works for me locally:

```Scala
+++ b/compiler/src/dotty/tools/dotc/transform/patmat/Space.scala
@@ -292,7 +292,7 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
   private val nullType             = ConstantType(Constant(null))
   private val nullSpace            = Typ(nullType)

   override def intersectUnrelatedAtomicTypes(tp1: Type, tp2: Type) = {
     val and = AndType(tp1, tp2)
     // Precondition: !(tp1 <:< tp2) && !(tp2 <:< tp1)
     // Then, no leaf of the and-type tree `and` is a subtype of `and`.
@@ -689,7 +689,8 @@ class SpaceEngine(implicit ctx: Context) extends SpaceLogic {
       tvar =>
         !(ctx.typerState.constraint.entry(tvar.origin) `eq` tvar.origin.underlying) ||
         (tvar `eq` removeThisType.prefixTVar),
-      minimizeAll = false
+      minimizeAll = false,
+      allowBottom = false
     )
--- a/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Inferencing.scala
@@ -115,7 +115,7 @@ object Inferencing {
             val minimize =
               force.minimizeAll ||
               variance >= 0 && !(
-                force == ForceDegree.noBottom &&
+                !force.allowBottom &&
                 defn.isBottomType(ctx.typeComparer.approximation(tvar.origin, fromBelow = true)))
             if (minimize) instantiate(tvar, fromBelow = true)
             else toMaximize = true
@@ -466,9 +466,9 @@ trait Inferencing { this: Typer =>

 /** An enumeration controlling the degree of forcing in "is-dully-defined" checks. */
 @sharable object ForceDegree {
-  class Value(val appliesTo: TypeVar => Boolean, val minimizeAll: Boolean)
+  class Value(val appliesTo: TypeVar => Boolean, val minimizeAll: Boolean, val allowBottom: Boolean = true)
   val none = new Value(_ => false, minimizeAll = false)
   val all = new Value(_ => true, minimizeAll = false)
-  val noBottom = new Value(_ => true, minimizeAll = false)
+  val noBottom = new Value(_ => true, minimizeAll = false, allowBottom = false)
 }
```

If you play with reachability, you may want to incorperate the changes in #4869.

Blaisorblade:
LGTM! This looks reasonable to me, and I'm pretty confident the changes to `Inferencing` are just a refactor for all other users, so they're safe.

Plus, my idea about `baseTypes` doesn't work as-is for two reasons: first, we need to find children and not parents, and second because (as @liufengyun pointed out) there might be no prefix (just like when GADT pattern match uses unification, it might find no solution).

PR #4899, opened 2018-08-06
Fix #4515: Synthesize implicit type tags in transparent methods

### LINKED ISSUE #4515: Synthesize class tags in top level splices.
```scala
object Foo {
  inline def foo[X](x: X): Unit = ~fooImpl('(x))
  def fooImpl[X: quoted.Type](x: quoted.Expr[X]): quoted.Expr[Unit] = '()
}
```

currently fails with 
```scala
3 |  inline def foo[X](x: X): Unit = ~fooImpl('(x))
  |                                                ^
  |no implicit argument of type quoted.Type[X] was found for parameter evidence$1 of method fooImpl in object Foo
```

We need to write instead
```scala
  inline def foo[X](x: X): Unit = ~fooImpl('(x))('[X])
```

We should be able to synthesize classtags inside top level splices to reduce this syntactic overhead.

This would be a change in `Typer` and not in `ReifyQuotes`.

nicolasstucki:


nicolasstucki:
Passed all tests but CLA failed.

nicolasstucki:
Rebased

nicolasstucki:
Rebased

PR #4900, opened 2018-08-06
Fix #4884: asSeenFrom on incorrect prefix in type applications

### LINKED ISSUE #4884: Abstract type bound of type member is forgotten?
```Scala
object Test {
  trait A
  trait TestConstructor1 { type F[_ <: A] }
  trait TestConstructor2[D] { type F[_ <: D] }

  val v1: TestConstructor1 => Unit = { f =>
    type P[a <: A] = f.F[a] // OK
  }

  val v2: TestConstructor2[A] => Unit = { f =>
    type P[a <: A] = f.F[a] // Error! Type argument a does not conform to upper bound D
  }
}
```

This compiles in Scala 2.12.1 and does not compile in Dotty 0.8.0. Just reporting this but I might have overlooked some dropped features.

Blaisorblade:
WIP fix #4884. PR for testing. Currently causing a stackoverflow on a benchmark, but only with the non-bootstrapped compiler; what's worse is that the stackoverflow doesn't go away when increasing stack size.

### Approach

In this code we fetch the type parameter bounds from the definition, so we must
substitute their definition using `asSeenFrom` with the correct prefix.

The correct prefix is however hard to find: `TypeApplications.typeParams` finds
a `self: TypeRefs` (after however many steps), the correct prefix is
`self.prefix`.

In fact, a compositional typing rule would obtain the bounds from the type (or,
well, kind) of the type constructor; that would avoid the need for using
`asSeenFrom`. We only have a way to obtain such a kind for `NamedType` (where we
can take the `info`). That might be enough, but there might be another way to
fix this.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -65,10 +65,14 @@ object Checking {
    *     cannot handle those.
    */
   def checkAppliedType(tree: AppliedTypeTree, boundsCheck: Boolean)(implicit ctx: Context) = {
+    def kind(tp: Type): Type = tp match {
+      case nTp: NamedType => nTp.info
+      case _ => tp
+    }
     val AppliedTypeTree(tycon, args) = tree
     // If `args` is a list of named arguments, return corresponding type parameters,
     // otherwise return type parameters unchanged
-    val tparams = tycon.tpe.typeParams
+    val tparams = kind(tycon.tpe).typeParams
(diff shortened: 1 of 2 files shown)


odersky:
I came across this problem independently when working on transparent functions. It's quite a subtle problem. Afters several false starts I found a fix which looks right in #4902.

Blaisorblade:
Fair enough — my original fix works for `Checking`, but not for earlier, and your fix also uses `info.typeParams` on `TypeRef`.

Issue #2862, opened 2017-07-13
Should all SymDenotations have trees?


odersky:
We now have two different ways to access a tree from a SymDenotation.

 - The IDE associates trees with toplevel classes and stores them in a `myTree` field of ClassDenotation.
 - Inline methods store trees in special funny attributes.

It would simplify things if we could simply store the tree representing the definition of a symbol in the symbol itself. (it should probably be the symbol, not the denotation, because we are probably not interested in transformed trees). The only downside I can see is increased memory consumption, but I don't think it will be a big concern.

smarter:
> The IDE associates trees with toplevel classes and stores them in a myTree field of ClassDenotation.

It's actually in ClassSymbol, not ClassDenotation

PR #4901, opened 2018-08-07
Tasty reflect load complete definition trees

nicolasstucki:
The goal is to have a way to load the tree from a symbol. 

The current scheme keeps any definition tree created while unpickling in the symbol directly. For symbols defined in classes that are being compiled are set during pickling.

Alternatively we could try to recover the tree by traversing the tree of the top level class, but this would not be performant. It could be implemented with some cache of some sort.

This is related to #2862


nicolasstucki:
test performance please

nicolasstucki:
test performance please

nicolasstucki:
test performance please

PR #4907, opened 2018-08-08
Remove -Xlink

nicolasstucki:
This mode has not been tested for a while and is not going to be used
in the future. It was the basis on which -from-tasty was created
and tested the loading loading of TASTy trees before -from-tasty existed.

It intended to use the linker call graph to know which classes to reload.
Now that this is not going to be supported this infrastructure is useless.


nicolasstucki:
@smarter I still had to fix a couple of files in the plugin tests.

PR #4914, opened 2018-08-08
Fix #4172: Add regression test

### LINKED ISSUE #4172: inline def is not inlined when it has two parameter lists, one with a repeated parameter and one with a default parameter
This is a very weird combination that I haven't been able to reduce further:
```scala
object Test {
  inline def foo(first: Int*)(second: String = "") = {}

  def test = {
    foo(1)()
  }
}
```

The body of test after typer is:
```scala
        {
          val first$1: Int* = [1 : Int]: Int*
          Test.foo(first$1)(Test.foo$default$2(first$1))
        }
```
Notice that `foo` hasn't been inlined away.

nicolasstucki:


nicolasstucki:
@smarter I tried that but it is impossible to write by hand the equivalent access to the default pareameter getter.

nicolasstucki:
In this case it is a call to a final method in the same class, it will not be an invokeDynamic.

smarter:
Invokedynamic is only used for lambdas, I don't see what it's doing here

Blaisorblade:
> Invokedynamic is only used for lambdas, I don't see what it's doing here

Is that tested? It seems easier to test for InvokeDynamic than to argue that can't possibly affect the test results, that's what we test for. Anyway, I dropped my commits since you agree.

PR #4915, opened 2018-08-08
Fix #4230: Handle import in the REPL correctly

### LINKED ISSUE #4230: Member defined in REPL should shadow previously imported member
Below an (un)reproducible transcript of the REPL not handling shadowing correctly. After both importing and defining `f`, only the imported one is in scope.

```scala
> repl
[...]
scala> object A { def f = 1 }
// defined object A
scala> import A._
scala> def f = "hi"
def f: String
scala> f
val res0: Int = 1
```

allanrenucci:
We used to collect imports from the parsed tree and insert them as
parsed trees at the top of the REPL wrapper object. This caused members
shadowing issues.

We now introduce a phase in the REPL compiler that collects imports
after type checking and store them as typed tree. We can then create a
context with its imports set in the correct order and use it to compile
future expressions.

### DIFF
--- a/compiler/src/dotty/tools/repl/CollectTopLevelImports.scala
+++ b/compiler/src/dotty/tools/repl/CollectTopLevelImports.scala
@@ -1,0 +1,30 @@
+package dotty.tools.repl
+
+import dotty.tools.dotc.ast.Trees._
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Phases.Phase
+
+/** A phase that collects user defined top level imports.
+ *
+ *  These imports must be collected as typed trees and therefore
+ *  after Typer.
+ */
+class CollectTopLevelImports extends Phase {
+  import tpd._
+
+  def phaseName = "collectTopLevelImports"
+
+  private[this] var myImports: List[Import] = _
+  def imports = myImports
+
+  def run(implicit ctx: Context): Unit = {
+    def topLevelImports(tree: Tree) = {
+      val PackageDef(_, _ :: TypeDef(_, rhs: Template) :: Nil) = tree
+      rhs.body.collect { case tree: Import => tree }
+    }
+
+    val tree = ctx.compilationUnit.tpdTree
+    myImports = topLevelImports(tree)
+  }
+}
--- a/compiler/src/dotty/tools/repl/ReplCompiler.scala
+++ b/compiler/src/dotty/tools/repl/ReplCompiler.scala
@@ -14,6 +14,7 @@ import dotty.tools.dotc.core.Phases.Phase
 import dotty.tools.dotc.core.StdNames._
 import dotty.tools.dotc.core.Symbols._
 import dotty.tools.dotc.reporting.diagnostic.messages
+import dotty.tools.dotc.transform.PostTyper
 import dotty.tools.dotc.typer.{FrontEnd, ImportInfo}
 import dotty.tools.dotc.util.Positions._
 import dotty.tools.dotc.util.SourceFile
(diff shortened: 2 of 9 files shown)


compiler/src/dotty/tools/repl/CollectTopLevelImports.scala
```diff
@@ -0,0 +1,30 @@
+package dotty.tools.repl
```
smarter: maybe put it in repl.transform ?

compiler/src/dotty/tools/repl/ReplCompiler.scala
```diff
@@ -26,39 +27,43 @@ import scala.collection.mutable
...
  *
- *  Specifically it replaces the front end with `REPLFrontEnd`, and adds a
- *  custom subclass of `GenBCode`. The custom `GenBCode`, `REPLGenBCode`, works
- *  in conjunction with a specialized class loader in order to load virtual
- *  classfiles.
+ *  Specifically it replaces the front end with `REPLFrontEnd`.
  */
 class ReplCompiler extends Compiler {
-  override protected def frontendPhases: List[List[Phase]] =
-    Phases.replace(classOf[FrontEnd], _ => new REPLFrontEnd :: Nil, super.frontendPhases)
```
smarter: I think keeping Phases.replace would be cleaner, it just needs to be passed a list of list of phases as a replacement.

allanrenucci: That would be trickier. I need to replace `FrontEnd` by `ReplFrontend`, remove `sbt.ExtractDependencies` and `sbt.ExtractAPI` and finally add `CollectTopLevelImports` which needs to run after `ReplFrontEnd` but in its own group

smarter: Ah indeed, not worth it then

allanrenucci:
I have a pickling issue:
```
class dotty.tools.dotc.reporting.diagnostic.messages$Error at ?: pickling difference for class ReplCompiler in dotty/compiler/src/dotty/tools/repl/ReplCompiler.scala
```
```
> diff before-pickling.txt after-pickling.txt
2857c2857,2861
<                           :Either[dotty.tools.repl.results.Errors, dotty.tools.dotc.ast.tpd.ValDef]>@<9296..9440>
---
>                           :
>                             Either[dotty.tools.repl.results.Errors, 
>                               dotty.tools.dotc.ast.Trees.ValDef[dotty.tools.dotc.core.Types.Type]
>                             ]
>                           >@<9296..9440>
2859c2863,2867
<                       :Either[dotty.tools.repl.results.Errors, dotty.tools.dotc.ast.tpd.ValDef]>@<9181..9440>
---
>                       :
>                         Either[dotty.tools.repl.results.Errors, 
>                           dotty.tools.dotc.ast.Trees.ValDef[dotty.tools.dotc.core.Types.Type]
>                         ]
>                       >@<9181..9440>
```
I am not sure how to debug that. cc/ @smarter @nicolasstucki

nicolasstucki:
I'll have a look tomorrow

nicolasstucki:
I think I found the place where it happens in 3402a29660ebc6ba632d122bfaa5e1e08dcbd93a (5cbb512db3309f9ef8cb677abf13ec27d6a81195 needs to be reverted, sorry). To debug it you should try writing https://github.com/lampepfl/dotty/pull/4915/commits/3402a29660ebc6ba632d122bfaa5e1e08dcbd93a#diff-c7dacf9ef97cabd025fede32386ffdecR269 in some other smaller class, then you should be able to reproduce it with `dotc` with the `TestConfiguration.picklingOptions` flags.

nicolasstucki:
@allanrenucci in a01be643165703a12df139ddc995135fe00f1457 I added a test that reproduces the issue. You can minimize it from that test file.

allanrenucci:
@nicolasstucki Thanks!

compiler/src/dotty/tools/repl/CollectTopLevelImports.scala
allanrenucci: Even if that's the only one?

allanrenucci:
With this new I still cannot handle:
```scala
scala> object A { def f = 1 }                                                                                                                                                                                   
// defined object A

scala> import A._; def f = 2
def f: Int

scala> f                                                                                                                                                                                                        
val res0: Int = 1 // should probably be 2
```
The scalac REPL handles it properly, but they use a complex object nesting scheme that we don't want to replicate

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -329,7 +329,7 @@ object Interactive {
 
   def contextOfPath(path: List[Tree])(implicit ctx: Context): Context = path match {
     case Nil | _ :: Nil =>
-      ctx.run.runContext.fresh.setCompilationUnit(ctx.compilationUnit)
```
allanrenucci: @odersky Was this intentional? If so, what's the rational for using the run context?

compiler/src/dotty/tools/repl/CollectTopLevelImports.scala
```diff
@@ -0,0 +1,30 @@
...
+
+/** A phase that collects user defined top level imports.
+ *
+ *  These imports must be collected as typed trees and therefore
+ *  after Typer.
+ */
+class CollectTopLevelImports extends Phase {
+  import tpd._
+
+  def phaseName = "collecttoplevelimports"
```
smarter: we usually use camelCase.

compiler/src/dotty/tools/repl/ReplCompiler.scala
```diff
@@ -26,39 +27,43 @@ import scala.collection.mutable
 /** This subclass of `Compiler` replaces the appropriate phases in order to
  *  facilitate the REPL
  *
- *  Specifically it replaces the front end with `REPLFrontEnd`, and adds a
- *  custom subclass of `GenBCode`. The custom `GenBCode`, `REPLGenBCode`, works
- *  in conjunction with a specialized class loader in order to load virtual
- *  classfiles.
+ *  Specifically it replaces the front end with `REPLFrontEnd`.
  */
```
smarter: I would just delete this whole comment, it doesn't really say anything interesting. (it could be replaced by an actual description of what REPLCompiler does though :))

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
smarter: Pretty sure that's intentional. `contextOfPath` constructs a context with the correct scoping information. If you start from a random Context then you'll have too much stuff in the scope.

smarter:
> With this new scheme I still cannot handle:

Can you open an issue for that?

compiler/src/dotty/tools/dotc/interactive/Interactive.scala
allanrenucci: I could patch the `rootContext` of the new run as we used to do:
```scala
class ReplCompiler extends Compiler {
  def newRun(initCtx: Context, state: State) = new Run(this, initCtx) {
    override protected[this] def rootContext(implicit ctx: Context) =
      addMagicImports(super.rootContext)
   
    private def addMagicImports(initCtx: Context): Context = ...
  }
}
```
This way I wouldn't need to modify this line

allanrenucci:
@smarter comments addressed

allanrenucci:
> Can you open an issue for that?

Will do once this is merged

PR #4916, opened 2018-08-09
Fixes to transparent to enable generic tuples

odersky:
Several fixes to transparent methods needed to make the `TupleAbstract.scala` test work. This one is a PoC for an implementation of generic tuples with the following properties:

 - no limit of 22
 - generic operations are possible
 - fully backwards compatible with Scala-2 tuples
 - specializes for small tuple sizes, yielding optimized code


odersky:
test performance please

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -655,8 +655,32 @@ class Inliner(call: tpd.Tree, rhsToInline: tpd.Tree)(implicit ctx: Context) {
...
+                  case t: Bind if t.symbol.isType && t.name != tpnme.WILDCARD =>
+                    t.symbol.asType :: syms
+                  case _ =>
+                    syms
+                }
+                foldOver(syms1, t)
+              }
+            }
+            val boundVars = getBoundVars(Nil, tpt)
+            for (bv <- boundVars) ctx.gadt.setBounds(bv, bv.info.bounds)
```
OlivierBlanvillain: This feels a bit ad hoc, isn't there a way to reuse normal type checking of patterns instead of redoing all that work here?

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -165,7 +165,7 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
       val leftArg = if (isRightAssoc && isInfixType(l)) "(" ~ argText(l) ~ ")" else argText(l)
       val rightArg = if (!isRightAssoc && isInfixType(r)) "(" ~ argText(r) ~ ")" else argText(r)
 
-      leftArg ~ " " ~ simpleNameString(op.classSymbol) ~ " " ~ rightArg
+      leftArg ~ " " ~ simpleNameString(op.typeSymbol) ~ " " ~ rightArg
```
Blaisorblade: FWIW can be dropped, this is already fixed in #4072 (https://github.com/lampepfl/dotty/pull/4072/commits/5e01496d910caaa57a0d950863816d40d8443981 and preserved by later refactorings).

compiler/src/dotty/tools/dotc/typer/Inliner.scala
odersky: I tried, but did not find a good way to re-use what's there. it's spread out over several different things.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -868,7 +871,45 @@ class Inliner(call: tpd.Tree, rhsToInline: tpd.Tree)(implicit ctx: Context) {
...
+    val dealiasTypeBindings = new TreeMap {
+      val boundTypes = typeBindings.map(_.symbol).toSet
+
+      val dealias = new TypeMap {
+        override def apply(tp: Type) = dealiased.get(tp) match {
+          case null =>
+            val tp1 = mapOver {
+              tp match {
+                case tp: TypeRef if boundTypes.contains(tp.symbol) =>
+                  val TypeAlias(alias) = tp.info
```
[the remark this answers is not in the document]
odersky: It should always be a `TypeAlias` because of the way we generate type bindings. So `boundTypes.contains(tp.symbol)` should imply `tp.info` is a `TypeAlias`.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
odersky: On rebasing I reverted the conflicting code from #4072. So "("...")" will only be printed if precedence demands it. Not writting redundant parens is important in particular for long sequences of type operators, as for example when working with HLists.

Blaisorblade: @odersky Did you lose your changes? This PR (as merged) contains no such revert, and #4072 *replaced* extra parentheses with proper use of `atPrec`. Either you lost your changes or the code you reverted was the old one.

If you can find a bug in master, happy to fix it and add tests. 😄

OlivierBlanvillain:
The newly added test doesn't pass `-Ytest-pickler` on my machine, I'm surprise this wasn't caught by the CI.

PR #4917, opened 2018-08-09
Fixups for Tasty-reflect

Blaisorblade:
Issues noticed when reviewing #4904.


nicolasstucki:
All these changes are now in #4904

PR #4920, opened 2018-08-10
Tune precedence of type operators

Blaisorblade:
Change the precedence of type operators to comply with the spec.
Only the last two commits are new relative to #4072, they offer different options for cherry-picking or squashing; each of them passes tests.

- [ ] we should choose;
- [ ] update the Dotty syntax reference.

EDIT: just learned that our current precedence was already accepted as SIP-33!
https://docs.scala-lang.org/sips/priority-based-infix-type-precedence.html


Blaisorblade:
SIP-33 mandates closing this, but #4072 is still in!

PR #4927, opened 2018-08-10
Change in terminology for typelevel methods

odersky:
This PR changes the `typelevel.md` spec, renaming `transparent` to `rewrite`.

 - also, introduce `rewrite` as a modifier for match expressions and conditionals.
 - keep `transparent` for methods that are not erased, and don't inline by themselves, yet can be inlined from rewritten code.

It's hopefully clearer this way. Also, `rewrite` on conditonals and matches, while being more verbose makes their use more compositional. This PR is based on #4916.


milessabin:
I worry slightly that using `rewrite` here will be confused with the tactics of the same name you find in Agda, Idris, Coq etc.

milessabin:
> Bit it's sort of equivalent

I think that's the problem ... it's close enough to cause confusion.

milessabin:
My understanding of the terminology in Agda/Idris/Coq, etc. is that `rewrite` is a specific tactic, and that in general tactics are deduction rules for _backwards_ reasoning, from a goal to a set of prior subgoals (see eg. the introduction here: https://coq.inria.fr/refman/proof-engine/tactics.html).

I think that the backwards reasoning aspect fits implicit search extremely well, and I can imagine that if we developed the idea of a tactic language for Scala it would be closely aligned with implicit search. In particular I can imagine putting together a `rewrite` tactic, similar to the Agda/Idris/Coq rewrite, as one or more transparent implicit methods.

I think that what you're trying to get to here is a label for _forwards_ reasoning, from premises (arguments/scrutinees) to conclusions (result/match RHS).

I think we're definitely in the same sort of area, but I don't think we have the terminology quite right yet.

I'll try and come up with something better.

milessabin:
Chatting with @Blaisorblade about this a little, we're both of the opinion (correct me if I'm misrepresenting you @Blaisorblade) that this is in effect a staging annotation.

If that's the case, then perhaps we should try and relate the syntax here to the metaprogramming syntax? How about reusing `~`? That would give us things like,

```scala
~ def concat[A, M, N](xs: Vec[A, M], ys: A[A, N])(implicit ev: Add[M, N]): Vec[A, ev.Result]

~ type Add[M <: Nat, N <: Nat] <: Nat = type M match {
  case Z => N
  case S[type M1] => S[Add[M1, N]]
}
```
It also takes this,
```scala
rewrite def m(x: T) = ~mImpl('x)
...
def mImpl(x: Expr[T]) = ...
```
to this,
```scala
~ def m(x: T) = ~mImpl('x)
...
def mImpl(x: Expr[T]) = ...
```
which I think is quite suggestive ... the compile time nature of the RHS is mirrored on the LHS.

As an added bonus this doesn't require another new keyword, and `~` sort of suggests something a little reduction like, to me at least ... I've seen `~>` used to express "reduces to" for instance.

odersky:
There sure are analogies between rewrite and macro splicing, but there are also differences. Splice `~` is of type `Expr[T] => T`. So if we wanted to use `~` for a rewrite match, we'd have to quote the returned values, like this:
```scala 
~ def size(xs: Tuple): Int = ~ xs match {
    case _: Empty => '0
    case _: (_ *: xs1) => 'size(erasedValue[xs1]) + 1
}
```
(That's actually the model I started with 6 months ago). But I now believe using `rewrite` or `inline`, or whatever, is clearer. 

There's also the difference that `~` does not produce new types, but `rewrite` can.

Finally, there's the difference that once I splice a function call, like `~mImpl(...)` it is assumed that `mImpl` will run to completion until it returns a value (of type `Expr[T]`). Under a reduction model this means transitively reducing everything `mImpl` calls. `rewrite` does not behave like this. It will inline only other `rewrite` functions, perform some selected simplifications and return the result (which is still a term in general, not a value).

So I fear merging `rewrite` and `~` would be more confusing than simplifying.

milessabin:
I wasn't proposing changing the semantics, just using the token `~` in place of `transparent`/`rewrite`/`inline`.

odersky:
Regarding usage of `rewrite` I am actually not so worried about the specific usage in Coq or Agda. The term "rewrite rules" is used more broadly than that. Rewrite rules also exist in Haskell (where they go forward), and more generally in term rewriting systems. Our rewrite rules are admittedly very simplistic in that they take a single function call as left-hand-side. But they do fit the definition of rewriting from term rewriting systems.

milessabin:
> Rewrite rules also exist in Haskell (where they go forward)

Fair enough.

ctongfei:
Since `rewrite` is a verb, it may corrupt some code if it is used as a keyword. Since it mostly supplants `inline`, I guess that `inline` is a better choice? (`inline` is a keyword in a bunch of languages so people don't usually use it for identifiers)

odersky:
@ctongfei I am not against `inline` but it would bring as back full circle. `inline` was dropped because people argued that it would be inappropriate for something that could change types. In other words, `inline` is just an advisory that should neither change types nor semantics.

`rewrite` makes that aspect fully explicit, but it's true that it will break code (easy to rewrite automatically by inserting backquotes, though).

ctongfei:
@odersky OK I got it. Maybe `rewriting` would be better (an adjective, similar to `transparent`/`inline`, and less likely to be used as an identifier)?

odersky:
It's a big commit. Who volunteers for reviewing?

compiler/src/dotty/tools/dotc/parsing/Parsers.scala
```diff
@@ -2519,6 +2543,8 @@ object Parsers {
           stats +++= defOrDcl(in.offset, defAnnotsMods(modifierTokens))
         else if (!isStatSep) {
           exitOnError = mustStartStat
+          println(in.token)
+          println(in.next.token)
```
nicolasstucki: Should remove these two lines

nicolasstucki:
@odersky this PR needs a rebase

PR #4928, opened 2018-08-10
Fix #4922: unapplySeq mustn't allow arrays

### LINKED ISSUE #4922: unapplySeq allows arrays
```scala
class SQ {
  val isEmpty = false
  val get = Array(1)
}
object A {
  def unapplySeq(a: Int): SQ = new SQ
}

object Main {
    val seq: Seq[Int] = 2 match {
      case A(xs: _*) => xs
    }
  }
}
```

```
➜  sandbox git:(typo) dotc Test.scala
undefined: x3.<none> # 814: TermRef(TermRef(NoPrefix,val x3),<none>) at pruneErasedDefs
one error found
```

https://github.com/lampepfl/dotty/blob/652092b07c4058201909effc4a8ffde469e37a75/compiler/src/dotty/tools/dotc/typer/Applications.scala#L105

`.elemType` works for Seqs but also arrays. there's no check that the getTp is a Seq.

Blaisorblade:
Fix #4922 with the "obvious" fix.

If this triggers cyclic errors, the extra check could be moved to PostTyper.

The error message could be improved, since it doesn't mention the specific reason why `SQ` isn't a valid result type.
```
12 |      case A(xs: _*) => xs
   |           ^^^^^^^^^
   | SQ is not a valid result type of an unapplySeq method of an extractor()
12 |      case A(xs: _*) => xs
   |                        ^^
   |                        not found: xs

```

Beware this was a 5-minute job, so haven't thought hard about it.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Applications.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Applications.scala
@@ -101,7 +101,7 @@ object Applications {
 
     if (unapplyName == nme.unapplySeq) {
       if (unapplyResult derivesFrom defn.SeqClass) seqSelector :: Nil
-      else if (isGetMatch(unapplyResult, pos)) {
+      else if (isGetMatch(unapplyResult, pos) && getTp.derivesFrom(defn.SeqClass)) {
         val seqArg = getTp.elemType.hiBound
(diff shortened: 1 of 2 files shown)


Blaisorblade:
Merging as it's an improvement, but @lrytz any feedback is still welcome!

PR #4932, opened 2018-08-11
Fix #2973: Check variances of class parents

### LINKED ISSUE #2973: Unsoundness due to missing variance checks for extends clause

Blaisorblade:
Fix #2973: check variance of parent classes via `ClassInfo.classParents` (via @odersky's fix in #4057).
To account for `@uncheckedVariance` (unlike #4057), prevent `Namer` from stripping annotations in `classParents`. Surprisingly, the rest of Dotty doesn't mind!

We probably should test this on the new collections and the community build, since collections will stress this out much more.

### DIFF
--- a/tests/neg/i2973.scala
+++ b/tests/neg/i2973.scala
@@ -1,2 +1,68 @@
-class Foo[U]
-class Bar[-T] extends Foo[T] // error: contravariant type T occurs in invariant position
+import annotation.unchecked.uncheckedVariance
+
+class InvClass[U]
+trait InvTrait[U]
+class ContraClass[-T]
+trait ContraTrait[-T]
+class CoClass[+T]
+trait CoTrait[+T]
+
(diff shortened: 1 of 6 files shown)


tests/neg/i2973.scala
```diff
@@ -0,0 +1,2 @@
+class Foo[U]
+class Bar[-T] extends Foo[T] // error: contravariant type T occurs in invariant position
```
smarter: More testcases (related to annotations, etc) would be nice.

Blaisorblade: Can't find any annotation for parent classes apart from `@uncheckedVariance`. So added more tests for that.

PR #4933, opened 2018-08-12
[WIP] Fix #3007: Added silencing of the output in initial and cleanup commands

### LINKED ISSUE #3007: Implement initial and cleanup commands as well as value bindings for REPL
https://github.com/dotty-staging/dotty/blob/812f141ba97e224168fc3bf91a6401c5dd349ce8/sbt-bridge/src/xsbt/ConsoleInterface.scala#L22-L39

SzymonPajzert:
Implementing the feature requested in #3007 by dropping the output of the executed commands.

### DIFF
--- a/compiler/src/dotty/tools/repl/ReplDriver.scala
+++ b/compiler/src/dotty/tools/repl/ReplDriver.scala
@@ -27,6 +27,7 @@ import org.jline.reader._
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
+
 /** The state of the REPL contains necessary bindings instead of having to have
  *  mutation
  *
@@ -70,7 +71,7 @@ class ReplDriver(settings: Array[String],
   }
 
   /** the initial, empty state of the REPL session */
-  protected[this] def initState = State(0, 0, Nil, rootCtx)
+  final def initialState = State(0, 0, Nil, rootCtx)
 
   /** Reset state of repl to the initial state
    *
@@ -101,7 +102,7 @@ class ReplDriver(settings: Array[String],
    *  observable outside of the CLI, for this reason, most helper methods are
    *  `protected final` to facilitate testing.
    */
-  final def runUntilQuit(): State = {
+  final def runUntilQuit(initialState: State = initialState): State = {
     val terminal = new JLineTerminal()
 
     /** Blockingly read a line, getting back a parse result */
@@ -127,7 +128,7 @@ class ReplDriver(settings: Array[String],
       else loop(interpret(res)(state))
     }
 
-    try withRedirectedOutput { loop(initState) }
+    try withRedirectedOutput { loop(initialState) }
     finally terminal.close()
   }
 
@@ -135,6 +136,9 @@ class ReplDriver(settings: Array[String],
     val parsed = ParseResult(input)(state.context)
     interpret(parsed)
   }
+
+  // TODO: i5069
+  final def bind(name: String, value: Any)(implicit state: State): State = state
 
   private def withRedirectedOutput(op: => State): State =
     Console.withOut(out) { Console.withErr(out) { op } }
@@ -308,7 +312,7 @@ class ReplDriver(settings: Array[String],
 
     case Reset =>
       resetToInitial()
-      initState
+      initialState
 
     case Imports =>
       state.imports.foreach(i => out.println(SyntaxHighlighting(i.show(state.context))))
--- a/sbt-bridge/src/xsbt/ConsoleInterface.scala
+++ b/sbt-bridge/src/xsbt/ConsoleInterface.scala
@@ -19,24 +19,6 @@ class ConsoleInterface {
   def run(args: Array[String],
           bootClasspathString: String,
           classpathString: String,
-          // TODO: initial commands needs to be run under some form of special
-          // "silent" mode in the REPL. I.e. the effects should be had without
-          // any visual output.
-          //
-          // To do this we can use the `run` interface to the `ReplDriver` and
-          // pass it a special instance of `ParseResult` like `Silently(res: ParseResult)`
-          // and then observe the effects without printing to `ReplDriver#out`
-          //
-          // This way, the REPL can offer feedback on invalid commands but
-          // still function without stringly logic.
-          //
-          // This same principle can be applied to `cleanupCommands` and
-          // `bindValues`
-          //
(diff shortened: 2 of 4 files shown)


allanrenucci:
Thanks @SzymonPajzert. I'll take a look when I'm back from vacation

allanrenucci:
I think this is the right approach to implement a silent mode for the REPL. However, the issue description was erroneous and `initialCommands` and `cleanupCommands` should not be interpreted in a silent mode (see https://github.com/lampepfl/dotty/issues/3007#issuecomment-414284981).

The silent mode could be useful to implement value bindings. I would implement everything in [ConsoleInterface](https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/ConsoleInterface.scala). Something along the line of:
```scala
val driver = new ReplDriver(completeArgs, classLoader = Some(loader))
val initState = driver.initState
val s0 = (bindNames, bindValues).zipped.folLeft(initState) {
  case (state, (name, value)) => driver.bind(name, value)(state)
}
val s1 = driver.run(initialCommands)(s0)
// handle failure during initialisation
val s2 = driver.runUntilQuit(s1)
driver.run(s1)
```
Value bindings can be implemented in a different PR

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -95,13 +97,15 @@ class ReplDriver(settings: Array[String],
...
+  final def runUntilQuit(): State = runUntilQuit(initState)
+
   /** Run REPL with `state` until `:quit` command found
    *
    *  This method is the main entry point into the REPL. Its effects are not
    *  observable outside of the CLI, for this reason, most helper methods are
    *  `protected final` to facilitate testing.
    */
-  final def runUntilQuit(): State = {
+  final def runUntilQuit(initialState: State): State = {
```
allanrenucci: I would make this take a default argument and remove the overload:
```scala
final def runUntilQuit(initialState: State = initState): State = 
```

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -127,14 +131,23 @@ class ReplDriver(settings: Array[String],
...
+    try withRedirectedOutput { loop(initialState) }
     finally terminal.close()
   }
 
   final def run(input: String)(implicit state: State): State = withRedirectedOutput {
     val parsed = ParseResult(input)(state.context)
     interpret(parsed)
   }
+  
+  final def bindValues(bindPairs: ZippedTraversable2[String, Any]): State = {
```
allanrenucci: I'd rather expose `def bind(name: String, value: Any): State` as public API. This can be moved to `ConsoleInterface`.

I would make `initState` public final and rename it `initialState`

sbt-bridge/src/xsbt/ConsoleInterface.scala
```diff
@@ -51,6 +33,12 @@ class ConsoleInterface {
       } ++
       Array("-classpath", classpathString)
 
-    new ReplDriver(completeArgs, classLoader = Some(loader)).runUntilQuit()
+    val driver = new ReplDriver(completeArgs, classLoader = Some(loader))
+
+    val s0 = driver.bindValues((bindNames, bindValues).zipped)
+    val s1 = driver.run(initialCommands)(s0)
+    // TODO handle failure during initialisation
```
allanrenucci: We will need to open issues for this and value bindings

SzymonPajzert: I have created #5069 to implement the bindings, but I don't understand how failure during initialization should be signaled. Should we mark it with some checked exception?

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -26,6 +26,8 @@ import org.jline.reader._
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
+import scala.runtime.ZippedTraversable2
```
allanrenucci: Unused?

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -95,13 +97,15 @@ class ReplDriver(settings: Array[String],
   // is called, we're in business
   resetToInitial()
 
+  final def runUntilQuit(): State = runUntilQuit(initialState)
```
allanrenucci: Remove?

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -136,6 +140,9 @@ class ReplDriver(settings: Array[String],
     interpret(parsed)
   }
 
+  // TODO: i3007
```
allanrenucci: #5069

sbt-dotty/sbt-test/sbt-dotty/example-project/test
```diff
@@ -1,4 +1,4 @@
 > run
 > 'set initialCommands := "1 + 1" '
 # FIXME: does not work on the CI
-#> console
+> console
```
allanrenucci: Let's revert this so that we can merge this PR. Let's not wait for #5018

compiler/src/dotty/tools/repl/ReplDriver.scala
```diff
@@ -136,6 +140,9 @@ class ReplDriver(settings: Array[String],
...
+  final def bind(name:String, value:Any)(implicit state: State): State = state
```
allanrenucci: Nitpick: add a space after `:`

sbt-bridge/src/xsbt/ConsoleInterface.scala
allanrenucci: Yes the current API that returns a state is not ideal. Let's not worry about it here, it can be fixed in a separate PR

allanrenucci:
> Is there any linter available to not submit changes with nitpicks?

No. At some point we talked about using scalafmt, but we decided no to enforce a specific style

One last think before we merge (sorry I didn't spot it earlier). Can you update the commit message to not mention silencing anymore. Something like: `Fix #3007: Add support for initial and cleanup command in the REPL`

allanrenucci:
Thanks @SzymonPajzert! 🎉

PR #4937, opened 2018-08-13
Add explicit return type to some methods

kmizu:
I think the code is more readable if a return type of a method is not omitted generally

### DIFF
--- a/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
+++ b/compiler/src/dotty/tools/backend/jvm/CollectEntryPoints.scala
@@ -46,7 +46,7 @@ class CollectEntryPoints extends MiniPhase {
 }
 
 object CollectEntryPoints{
-  def isJavaMainMethod(sym: Symbol)(implicit ctx: Context) = {
+  def isJavaMainMethod(sym: Symbol)(implicit ctx: Context): Boolean = {
     (sym.name == nme.main) && (sym.info match {
       case r@MethodTpe(_, List(defn.ArrayOf(t)), _) =>
         (t.widenDealias =:= defn.StringType) && (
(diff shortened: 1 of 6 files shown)


Blaisorblade:
Hi! Thanks for your PR, but how have you picked which methods to add a return type for? That's really not obvious to me, nor it is whether *these* methods need a return type.

kmizu:
@Blaisorblade Hi.  I think **all** public methods (private in some cases) should have explicit return type. in general  The reasons are below:

1. Return type inference may change method signature depending on implementation details.  If these methods are called from different compilation unit, such accidental changes break binary compatibiliy and generate unneeded recompilation.  
2. Depending on implementation details, very complex return type maybe inferred.  Such a method is difficult to use.
3. If a return type is not inferred, the method user can call it without seeing its implementation.
4. adding explicit return type is low cost and have many merits.

In this PR, I picked up several methods that the return types are inferred experimentally.  If it's permitted, I'll try to add explicit return type to **all** public methods.

@odersky I think `: Boolean`  to an `is ...` is not redundant. the consistency between the method implementation and the return type is guaranteed by `: Boolean`.  Similarly for `: Unit` (in some cases, `Any` was inferred accidentally).

allanrenucci:
I usually only add explicit return type to public members. I can see value in adding explicit return type to private/local side effect-full methods that return `Unit`.

However in general, I would not add explicit return type to private/local methods (and local variables). These methods are usually fairly small and it makes the code more verbose and sometime less readable.

But that's my opinion 😄

OlivierBlanvillain:
It's been a month... @kmizu If you plan to continue working on this PR please reopen (or make a new one), I will close it for now.

kmizu:
Hi, @smarter.  I see that the this pR shouldn't be needed anymore.  Thanks.

PR #4938, opened 2018-08-13
Add support for generic tuples

odersky:
Based on #4927.


odersky:
This PR is currently blocked since dotty/library is compiled with Scala-2. I tried several workarounds but there's basically no way to do the kind of typelevel computation we want with Scala-2 not throwing up. So we need to the full bootstrap to be completed first.

smarter:
In what kind of way does Scala 2 explode? Keep in mind that we could have two versions of "Tuple.scala", one compiled by Scala 2 and one compiled by Dotty, sbt makes this fairly easy to do.

odersky:
The blockers right now are

 - Scala 2 does not understand rewrite modifiers. I can fake it with a rewrite annotation for methods, but would have to do something crazy for rewrite matches and rewrite ifs.
 - Scala 2 rejects this match:
```scala
    @`rewrite` private def _size(xs: Tuple): Int = //rewrite
    xs match {
      case _: Unit => 0
      case _: *:[_, xs1] => _size(erasedValue[xs1]) + 1
    }
```
   because it does not know that `Unit` is a subtype of `Tuple`.

Both would be fixed if I could define a Dotty-only Tuple class. How do I do it? Put it in scalaShadowing?

smarter:
It requires some sbt trickery, I can push a commit to this PR with an example and you can check if that works for you.

odersky:
OK, thanks!

smarter:
See the latest commit, you should be able to put any dotty-only files in library/src-scala3

library/src/scala/TupleXXL.scala
```diff
@@ -0,0 +1,14 @@
+package scala
+
+final class TupleXXL private (es: Array[Object]) {
+  override def toString = elems.mkString("(", ",", ")")
+  override def hashCode = getClass.hashCode * 41 + elems.deep.hashCode
+  override def equals(that: Any) = that match {
+    case that: TupleXXL => this.elems.deep.equals(that.elems.deep)
```
smarter: `.deep` has been removed from 2.13 (https://github.com/scala/scala/pull/6912), but you can use Arrays.deepEquals instead: https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#deepEquals(java.lang.Object[],%20java.lang.Object[])

odersky:
Cool! the src-scala2/src-scala3 seems to have done the trick!

odersky:
@smarter I found an issue: `launchIDE` does not work anymore. I get:

sbt:dotty> dotty-bootstrapped/clean
[success] Total time: 0 s, completed Aug 15, 2018 1:43:50 PM
sbt:dotty> launchIDE
[info] Cleaning the dotty-sbt-bridge cache
[info] Updating dotty-library-bootstrapped...
[info] Updating dotty-interfaces...
[info] Done updating.
[info] Done updating.
[info] Updating dotty-compiler-bootstrapped...
[info] Updating dotty-compiler...
[info] Done updating.
[info] Updating dotty-sbt-bridge-bootstrapped...
[info] Updating dotty-doc-bootstrapped...
[info] Updating dotty-language-server...
[info] Compiling 7 Java sources to /Users/odersky/workspace/dotty/interfaces/target/classes ...
[info] Done compiling.
[info] Updating dotty-bench-bootstrapped...
[info] Done updating.
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] Done updating.
[info] Compiling 59 Scala sources and 136 Java sources to /Users/odersky/workspace/dotty/library/../out/bootstrap/dotty-library-bootstrapped/scala-0.10/classes ...
[error] Bad symbolic reference. A signature
[error] refers to *:/T in package scala which is not available.
[error] It may be completely missing from the current classpath, or the version on
[error] the classpath might be incompatible with the version used when compiling the signature.
[error] Bad symbolic reference. A signature
[error] refers to Tuple/T in package scala which is not available.
[error] It may be completely missing from the current classpath, or the version on
[error] the classpath might be incompatible with the version used when compiling the signature.

smarter:
Also try restarting sbt maybe.

odersky:
Now it works for me as well. Sorry for the false alarm!

odersky:
Update on performance:
```scala
    object Test extends App {
      val xs0 = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16)
      assert(xs0(15) == 16)
        // 2.753s

      val xs1 = xs0 ++ xs0
      assert(xs1(31) == 16)
        // 3.426s

      val xs2 = xs1 ++ xs1
      assert(xs2(63) == 16)
        // 4.082s

      val xs3 = xs2 ++ xs2
      assert(xs3(127) == 16)
        // 4.786s

      val xs4 = xs3 ++ xs3
      assert(xs4(255) == 16)
        // 6.543s

      val xs5 = xs4 ++ xs4
      assert(xs5(511) == 16)
        // 15.041s

    }
```
Times behind `//` are (cold) compile times up to this point. The last two (or three?) blocks required a stack size of 10M. 

One has has to change -Xmax-inlines accordingly. E.g. for the last block
```
    java -Xms1g -Xmx3g -Xss10m dotty.tools.dotc.Main tuples2.scala -Xmax-inlines 520
```

Summary: it would be good if we could improve this.

odersky:
Code size: The `Test` class above generated 48K, almost all of which were generic signatures involving the non-existent `*:` pair. After the last commit it went down to 5.4K, which looks reasonable.

odersky:
`.tasty` size is 7.7K btw.

odersky:
@liufengyun We get a wrong exhaustivity warning in run/tuples1.scala. It has to do with TupleXXL matching. Can you take a look?

odersky:
test performance please

liufengyun:
@odersky OK, I'll investigate.

liufengyun:
@odersky To ease the special-case `TupleXXL`, I'm wondering if `TupleXXL` can take a dummy type parameter as parameter to indicate its size? It can be either a constant type, or something like `*:[Int, *:[Int, *:[Int, Unit]]]`.

compiler/src/dotty/tools/dotc/transform/TypeUtils.scala
```diff
@@ -27,5 +27,28 @@ object TypeUtils {
       case self: MethodicType => self
       case _ => if (ctx.erasedTypes) MethodType(Nil, self) else ExprType(self)
     }
+
+    /** The arity of this tuple type, which can be made up of Unit, TupleX and `*:` pairs,
+     *  or -1 if this is not a tuple type.
+     */
+    def tupleArity(implicit ctx: Context): Int = self match {
```
nicolasstucki: Could be `tailtec`:
```scala
def tupleArity(implicit ctx: Context): Int = {
  def tupleArity(tp: Type, arity: Int): Int = tp match {
    case AppliedType(tycon, _ :: tl :: Nil) if tycon.isRef(defn.PairClass) =>
      tupleArity(tl, arity + 1)
   case tp1 =>
      if (tp1.isRef(defn.UnitClass)) arity
      else if (defn.isTupleClass(tp1.classSymbol)) arity + tp1.dealias.argInfos.length
      else -1
  }
  tupleArity(self, 0)
}
```

compiler/src/dotty/tools/dotc/core/Definitions.scala
```diff
@@ -1226,6 +1242,27 @@ class Definitions {
 
   private[this] var isInitialized = false
 
+  /** Add a `Tuple` as a parent to `Unit`.
+   *  Add the right `*:` instance as a parent to Tuple1..Tuple22
+   */
+  def fixTupleCompleter(cls: ClassSymbol): Unit = cls.infoOrCompleter match {
```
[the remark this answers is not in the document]
odersky: I looked hard but did not find anything that would work better than patching the completers. The `syntheticParent` code can be simplified using an utility method that I introduced in the subsequent PR #4964. I'll do the simplification there.

library/src/scala/TupleXXL.scala
```diff
@@ -0,0 +1,17 @@
+package scala
+import java.util.Arrays.{deepEquals, deepHashCode}
+
+final class TupleXXL private (es: Array[Object]) {
+  override def toString = elems.mkString("(", ",", ")")
+  override def hashCode = getClass.hashCode * 41 + deepHashCode(elems)
```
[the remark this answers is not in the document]
odersky: It's just to distinguish `TupleXXL` hashes from other hashes over `elems`.

compiler/src/dotty/tools/dotc/transform/TypeUtils.scala
odersky: I believe that version behaves differently for *: sequences that do not end in `Unit`.

compiler/test/dotc/pos-test-pickling.blacklist
```diff
@@ -33,6 +33,7 @@ i4125.scala
 i4167
 i4586.scala
 i4734
+i4758
```
OlivierBlanvillain: They slipped through the CI because before #5059 the CI didn't run `-Ytest-pickler` on pos and run tests..

compiler/src/dotty/tools/dotc/transform/TypeUtils.scala
nicolasstucki: Maybe, still there exists a correct tailrec version of the method. Can be revisited later.

PR #4941, opened 2018-08-14
Remove buggy override

nicolasstucki:
It removes the -from-tasty compiler for no good reason


liufengyun:
test performance with #tasty please

PR #4942, opened 2018-08-14
Add Constant and replace transparent parameters

nicolasstucki:
Introduce the type `Constant` which is a subtype of `Singleton` but only allows constants (e.i. known values).

Use `Constant` to mark parameters of `transparent` methods that must be passed as constants or inlined constants.

### DIFF
--- a/bench/tests/power-macro/PowerMacro.scala
+++ b/bench/tests/power-macro/PowerMacro.scala
@@ -2,7 +2,7 @@ import scala.quoted.Expr
 
 object PowerMacro {
 
-  transparent def power(transparent n: Long, x: Double) = ~powerCode(n, '(x))
+  transparent def power(n: Long & Constant, x: Double) = ~powerCode(n, '(x))
 
   def powerCode(n: Long, x: Expr[Double]): Expr[Double] =
     if (n == 0) '(1.0)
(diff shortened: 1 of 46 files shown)


smarter:
Maybe @Blaisorblade can tell us how bad this would be :)

Blaisorblade:
Why this change?

Using a type for `Singleton` is already fishy. If `A <: Singleton` and `B <: Singleton`, it follows that `A | B <: Singleton`, but for instance `1 | 2` is not actually a singleton type. The same applies here: if `A <: Constant` and `B <: Constant` then `A | B <: Constant`. And neither is easy to special-case, since you can write e.g.

```
def f[Bound, A <: Bound, B <: Bound](a: A, b: B): C = if (...) a else b
```

And if you manage to make it work, you'll have recreated a way to check programs which is *syntax*-directed, not *type*-directed.

`Constant` seems to also have a more basic problem: since `1: Constant`, then it should be possible to write

```scala
val x: 1 & Constant = 1
x: Constant // should work
```

but your examples suggest otherwise.

Blaisorblade:
We talked about this also with Martin today, and we agreed to think harder about it. He agreed Singleton is problematic, but actually Constant is muddier. Singleton clearly breaks because it describes types; but Constant describes "compile-time values", and it seems to me that `Int & Constant` resembles a bit `quoted.Expr[Int]`.

nicolasstucki:
I agree. The main idea was to express constantness of a parameter of an inlined method (for macros). As singleton already covered constants I explored this design that clearly has issues.

nicolasstucki:
To be more precise `Int & Constant` would represent a `quoted.LiftedExpr[Int]`, one of the implementations of `quoted.Expr` in which the quoted code is represented as its runtime value.

nicolasstucki:
We should first solve the issue with Singleton

PR #4949, opened 2018-08-15
Fix #4947: Do not replace positions of inlined arguments

### LINKED ISSUE #4947: Inline methods with by-name arguments break debugging information
In the snippet below, setting breakpoints in `isSubType` is hard if it's compiled by Dotty, because all of isSubType appears written on line 100.

```scala
transparent def track(f: => T): T = { ... f ... }

def isSubType(...) = track { // line 100
  body
]
```

Probably should be addressed before bootstrap, hence tentatively targeting milestone 0.10.

nicolasstucki:


compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -149,12 +149,26 @@ object Inliner {
...
+        override def transform(tree: Tree)(implicit ctx: Context): Tree = {
+          tree match {
+            case inlined: Inlined =>
+              val last = pos
+              pos = inlined.call.pos
+              val res = tpd.seq(inlined.bindings.map(transform), transform(inlined.expansion)).withPos(last)
+              pos = last
+              res
+            case tree =>
+              if (pos.exists) super.transform(tree).withPos(pos)
```
Blaisorblade: @nicolasstucki says `!pos.exists` is only possible if `inlined.call` is `emptyTree`, which means that this is not a *call* to an inline method but an *argument* to it — and then `bindings` should be empty.
We should update docs for `Inlined` about this.

smarter:
Please add documentation, I have no idea what this does from looking at it.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
odersky: This does not look like the right algorithm. We have enclosingInlineds, which is interpreted as follows:

 - An Inlined node with a non-empty call is a plain inlined node
 - An inlined node with an empty tree as call in an inlined argument which _cancels_ the enclosing inlined call.

So one should start with the outermost InlinedCall and update all positions in its expansion to the call position, except in those regions where there are as many empty as non-empty calls in `enclosingInlineds`.

nicolasstucki: That sounds like the algorithm I implemented though I use the pos state to know if it has been cancelled or not. I will change it to use `enclosingInlineds` directly.

smarter:
I think this also needs more extensive tests covering various nesting levels of inlined/non-inlined across one or more files and probably more low-level tests too (like https://github.com/scala/scala/blob/2.13.x/test/junit/scala/tools/nsc/backend/jvm/BytecodeTest.scala#L32-L77, https://github.com/scala/scala/blob/2.13.x/test/junit/scala/tools/nsc/backend/jvm/opt/InlinerTest.scala#L1515, etc)

nicolasstucki:
@Blaisorblade I when a bit than #4947 and kept the positions of trees inlined in the same file. Those already have the correct position and can be used in the debugger or stack traces.

Blaisorblade:
> I went a bit further than #4947 and kept the positions of trees inlined in the same file.

Sounds like what I hoped for! Line numbers in debugging now work like a charm (minus, of course, setting breakpoints inside `track`, but that requires JSR-45). The only weird thing is that IntelliJ *thinks* you can set breakpoint in `track`, which suggests it is output at all, but that's an unfortunate UX for the semantics of `transparent/@forceInline`.

compiler/src/dotty/tools/dotc/typer/Inliner.scala
```diff
@@ -149,12 +149,36 @@ object Inliner {
...
+       */
+      class Reposition extends TreeMap {
+        override def transform(tree: Tree)(implicit ctx: Context): Tree = {
+          tree match {
+            case tree: Inlined => transformInline(tree)
+            case _ =>
+              val transformed = super.transform(tree)
+              enclosingInlineds match {
+                case call :: _ if call.symbol.sourceFile != ctx.owner.sourceFile =>
+                  // reposition tree inlined from some other file
```
Blaisorblade: I think the documentation is too complicated. We're dropping Inlined nodes, and repositioning nodes from other files to the call position because we can't express their position otherwise in their debug info, pending JSR-45. The stuff about "empty call" is confusing and expressed in terms of the implementation.

Other issues that I think are there and that I'm trying to address:

- The comment makes sense, but the code is stricter, and resets the position even if the toplevel class is different but the source file is the same. In that case, it appears we don't need to reset the position. Addressed by one commit I pushed.

- Also, this is an expensive (O(n)) way to test if enclosingInlineds.nonEmpty and get the head, and that could be expensive, but that's really enclosingInlineds's fault and I think can be fixed separately without affecting this code.

compiler/test/dotty/tools/backend/jvm/InlineBytecodeTests.scala
```diff
@@ -37,4 +41,240 @@ class InlineBytecodeTests extends DottyBytecodeTest {
         diffInstructions(instructions2, instructions3))
     }
   }
+
+  @Test def i4947 = {
+    val source = """class Foo {
+                   |  transparent def track[T](f: => T): T = {
+                   |    println("tracking") // line 3
```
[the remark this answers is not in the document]
nicolasstucki: Changed

Blaisorblade:
@nicolasstucki The current algorithm is fishy; if I keep the full stack as described in https://github.com/lampepfl/dotty/pull/4990#issuecomment-415366029 (not the PR but the comment), the new tests break.

What we need is to compare the current source file `ctx.owner.sourceFile` with the source file of parameters, which is `ctx.owner.sourceFile` for the enclosing call (not theirs).

I'm not sure Martin's suggested algorithm is correct either, or if so I don't get why:
> So one should start with the outermost InlinedCall and update all positions in its expansion to the call position, except in those regions where there are as many empty as non-empty calls in enclosingInlineds.

It seems to keep unchanged the position of parameters (as many empty as non-empty calls), but that's wrong for parameters from another file (also featuring in the latest testcases).

nicolasstucki:
@Blaisorblade I'll let you merge if all is in order for you.

Blaisorblade:
@nicolasstucki I think I figured what's going on and why the code is correct. I'll update the comments and merge everything.

Blaisorblade:
I'm merging the PRs as-is and pushing comment clarifications to followup PRs.

PR #4954, opened 2018-08-17
Dottydoc:  Show implicit modifier in implicit function types

Duhemm:


doc-tool/src/dotty/tools/dottydoc/model/references.scala
```diff
@@ -28,42 +28,4 @@ object references {
   }
   final case class NoLink(title: String, target: String) extends MaterializableLink
 
-  implicit class ReferenceShower(val ref: Reference) extends AnyVal {
```
Blaisorblade: As an outsider: This PR mostly makes sense to me, not sure why `ReferenceShower` is removed here.

Not that I'm attached to it — it seems pretty-printing won't insert necessary extra parentheses — seems we'd have gotten `A & B => C` from both `(A & B) => C` and `A & (B => C)`. But is this handled better by the new code? It seems harder to get right once you flatten everything into nested maps, as `asJava` seems to do.

For the record, PlainPrinter/RefinedPrinter start handling this correctly in #4072 (tho the basic infrastructure was in place): recursive calls must pass the current precedence and insert parentheses around nodes with a lower precedence than the current one.

But I wonder if DottyDoc should *reimplement* this logic instead of reusing `RefinedPrinter`, since *each* deviation will confuse users. The only different requirement seems that `RefinedPrinter` uses a pretty-printing library designed to insert line-breaks where needed, but flattening that to `String` without extra line-breaks should be feasible.

Duhemm: I removed it because the functionality is actually implemented twice: in `ReferenceShower` and `RenderReference`.

Instead of generating a string from the reference, we keep it in "java form" and use the tag to actually render it.

Duhemm:
Rebased. @Blaisorblade Do you mind reviewing this PR, since you already had a look?

Blaisorblade:
@Duhemm wanted to merge this, but the CI hasn't passed yet, so I can't. What's up with it? Can you take care of it or reassign it?

PR #4957, opened 2018-08-17
Fix #3613: Delete output dir before testing tastyBootstrap

### LINKED ISSUE #3613: The tastyBootstrap test should clean its output directory before compiling anything
`tastyBootstrap` tests separate compilation, but does not clean its output directory before doing so. This can lead to misleading results, and may also fail when the output directory contains outdated TASTY sections.

nicolasstucki:


compiler/test/dotty/tools/dotc/CompilationTests.scala
```diff
@@ -196,6 +196,9 @@ class CompilationTests extends ParallelTesting {
     val dotty2Group = TestGroup("tastyBootstrap/dotty2")
     val libGroup = TestGroup("tastyBootstrap/lib")
 
+    // Make sure that the directory is clean
+    dotty.tools.io.Path(s"out/tastyBootstrap").deleteRecursively()
```
allanrenucci: Did you mean to write `s"$defaultOutputDir/tastyBootstrap"`?

allanrenucci: Also, you could do `dotty.tools.io.Directory(...)` to prevent unnecessary IOs

Issue #4940, opened 2018-08-13
Allow rewriting on types


odersky:
We had a discussion recently about the problem to represent the _type_ of vector concat. If we assume a type `Vector[Elem, Len <: Nat]`, it is straightforward to implement a typelevel 
```scala
rewrite def concat[A, M, N](xs: Vec[A, M], ys: A[A, N]): Vec[A, _]
```
such that the result of expanding `concat` for concrete vectors is a vector whose length is known to be the sum of the lengths of the argument vectors. But, annoyingly, there's no simple way to declare this fact in the result type of `concat`. One can fallback to implicits, i.e. something like
```scala
rewrite def concat[A, M, N](xs: Vec[A, M], ys: A[A, N])(implicit ev: Add[M, N]): Vec[A, ev.Result]
```
But that's a rather roundabout way to achieve something that should be simple.

The proposal here is to allow rewrite not only on terms but also on types. It's sort of the dual of #4939 in that it caters for computations that produce types whereas #4939 is a way to consume types directly in typelevel computations. 

To solve the example above, one would write
```scala
rewrite type Add[M <: Nat, N <: Nat] <: Nat = type M match {
  case Z => N
  case S[type M1] => S[Add[M1, N]]
}
```
This assumes type matches as proposed by #4939, but of course one could also use their
expansion:
```scala
rewrite type Add[M <: Nat, N <: Nat] <: Nat = rewrite erasedValue[M] match {
  case _: Z => N
  case _: S[type M1] => S[Add[M1, N]]
}
```
Unlike #4939, this feature is not simply encodable in the base language, so one has to explain its properties in detail:

**Question**: What is the syntax of a rewrite type definition?

**Answer**: The same as a `rewrite def`, except that it's `type` instead of `def`, the (optional) declared result type is prefixed by `<:` instead of `:` and the right hand side is a type expression. Type expressions are formed from

 - blocks `{ ... }`, ending in a type expression
 - rewrite, type, and implicit matches, with type expressions in each case
 - rewrite conditionals, with type expressions in each branch
 - types

If a declared result type is missing, `Any` is assumed.

**Question**: When is a rewrite type application expanded?

**Answer**: Analogous to a rewrite term application, it is expanded on each use _outside_ a rewrite method or type. In particular this means that in
```scala
rewrite def concat[A, M, N](xs: Vec[A, M], ys: A[A, N]): Vec[A, Add[M, N]]
```
[...]

PR #4962, opened 2018-08-19
Fix #4939: Add `rewrite type match` construct.

### LINKED ISSUE #4939: Allow matching on types
The work in #4927 and preceding PR's showed that it's viable to compute with types, but that the syntax is sometimes a bit clunky. In particular, the following pattern is ubiquitous:

```scala
import scala.typelevel.erasedValue

rewrite erasedValue[T] match {
  case _: A => ...
  ...
  case _: Z => ...
}
```

It would be nice if we had a more streamlined syntax for this. I propose:

```scala
rewrite type T match {
  case A => ...
  ...
  case Z => ...
}
```
Here `T` is an arbitrary type and `A`,...,`Z` are arbitrary type patterns. The construct is semantically equivalent to the first version with `erasedValue`, and can be explained that way. That is, the first case clause where `T` can be made a subtype of the type pattern is chosen. However, the compiler can probably process the second version more efficiently (in terms of compile speed) than the first.

Syntax:

```
Expr1            ::= ...
                     [‘rewrite’] PostfixExpr ‘match’ ‘{’ CaseClauses ‘}’  
                     'type' InfixType ‘match’ ‘{’ TypeCaseClauses ‘}’  
TypeCaseClauses  ::= TypeCaseClause { TypeCaseClause }
TypeCaseClause   ::= 'case' Type [Guard] ‘=>’ Block
```

[EDITED: I had a add `rewrite` in front of `type` since otherwise parsing gets too ambiguous.]

odersky:
This is based on #4938. Only the last two commits are new. 

I am not quite sure about this one yet. I think #4940 is more important, so I would put this one on hold until #4940 is decided.


Blaisorblade:
(Changed the title, guess this fixes #4939 not #4949).

PR #3486, opened 2017-11-16
Fix #3471: Avoid wrong flag kind when desugaring module definitions

odersky:

Issue #3471, opened 2017-11-14
"Sealed object" crashed the compiler


maseev:
I guess there's no such thing as a "sealed object" but writing this into the REPL:

```scala
scala> sealed object Fun 
```

kills the REPL with this:

```
Exception in thread "main" java.lang.AssertionError: assertion failed: illegal flagset combination: sealed module and final lazy module <stable>
        at dotty.DottyPredef$.assertFail(DottyPredef.scala:39)
        at dotty.tools.dotc.core.Flags$FlagSet$.$bar$extension(Flags.scala:26)
        at dotty.tools.dotc.ast.untpd$Modifiers.$bar(untpd.scala:163)
        at dotty.tools.dotc.ast.desugar$.moduleDef(Desugar.scala:581)
        at dotty.tools.dotc.ast.desugar$.defTree(Desugar.scala:695)
        at dotty.tools.dotc.typer.Namer.expand(Namer.scala:367)
        at dotty.tools.dotc.typer.Namer.index$$anonfun$1(Namer.scala:639)
        at scala.compat.java8.JProcedure1.apply(JProcedure1.java:18)
        at scala.compat.java8.JProcedure1.apply(JProcedure1.java:10)
        at scala.collection.immutable.List.foreach(List.scala:389)
        at dotty.tools.dotc.typer.Namer.index(Namer.scala:639)
        at dotty.tools.dotc.typer.Namer.indexAndAnnotate(Namer.scala:716)
        at dotty.tools.dotc.typer.Namer$ClassCompleter.completeInCreationContext(Namer.scala:911)
        at dotty.tools.dotc.typer.Namer$Completer.complete(Namer.scala:758)
        at dotty.tools.dotc.core.SymDenotations$SymDenotation.completeFrom(SymDenotations.scala:241)
        at dotty.tools.dotc.core.SymDenotations$SymDenotation.completeInfo$1(SymDenotations.scala:204)
        at dotty.tools.dotc.core.SymDenotations$SymDenotation.info(SymDenotations.scala:206)
        at dotty.tools.dotc.core.SymDenotations$SymDenotation.ensureCompleted(SymDenotations.scala:347)
        at dotty.tools.dotc.core.SymDenotations$SymDenotation.isAbsent(SymDenotations.scala:453)
        at dotty.tools.dotc.typer.TypeAssigner.reallyExists(TypeAssigner.scala:160)
        at dotty.tools.dotc.typer.Typer.qualifies$1(Typer.scala:131)
[...]

PR #4973, opened 2018-08-21
Fix #4936: Forbid (or warn on) redundant/illegal object flags

### LINKED ISSUE #4936: sealed final abstract case object Foo
Lots of modifiers are allowed on `object`s, but they do not seem to have any semantics.

In fact, you can have a
```scala
sealed final abstract case object Foo
```

But *sadly*
```scala
abstract object Foo {
  def x : String
}
```
doesn't compile, haha.

Perhaps those modifiers should be reported as errors?

Blaisorblade:
Forbid/warn on redundant/illegal object flags — fix #4936.

- Stop using final object in our sources.
- Give error for abstract and sealed objects, warning for final ones (as long as
 scala/bug#11094 is open).

### DIFF
--- a/tests/pos/i3471.scala
+++ b/tests/pos/i3471.scala
@@ -1,1 +1,0 @@
-sealed object Fun
--- a/compiler/src/dotty/tools/dotc/core/NameOps.scala
+++ b/compiler/src/dotty/tools/dotc/core/NameOps.scala
@@ -10,7 +10,7 @@ import Definitions._
 
 object NameOps {
 
-  final object compactify {
+  object compactify {
     lazy val md5: MessageDigest = MessageDigest.getInstance("MD5")
 
     /** COMPACTIFY
--- a/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
+++ b/compiler/src/dotty/tools/dotc/parsing/Parsers.scala
@@ -2415,6 +2415,16 @@ object Parsers {
 
     def objectDefRest(start: Offset, mods: Modifiers, name: TermName): ModuleDef = {
       val template = templateOpt(emptyConstructor)
+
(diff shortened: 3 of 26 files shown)


smarter:
could be done in the parser no?

odersky:
@Blaisorblade Inactive for almost 5 months. Can you either finish the PR or close it and re-assign the issue to someone else? Thanks

tests/pos/i3471.scala
```diff
@@ -1 +0,0 @@
-sealed object Fun
```
Blaisorblade: This testcase was added in #3486 for #3471 to ensure the compiler didn't crash.
During the discussion, it was demanded that `final` be handled at the same time: https://github.com/lampepfl/dotty/pull/3486#discussion_r151471447.

allanrenucci:
I think @odersky's [comment](https://github.com/lampepfl/dotty/pull/5525#issuecomment-447099112) applies to this PR

Blaisorblade:
That comment does apply, and I guess it overrides contrary suggestions above.

Blaisorblade:
FWIW, the current CI failure is spurious; while this is on hold, I'd like to get it working as far as possible.

Blaisorblade:
@nicolasstucki I've done that change, so it could be merged, but @odersky in #5525 suggested centralizing flag checking. Assigning to you for decision.

nicolasstucki:
We can centralize them later in #5525.

odersky:
I am afraid this is the wrong fix. See our long discussion with Allan. Flag combinations should _not_ be tested in Parser, we discussed this and decided on it. Please let's revert these commits and do it from the ground up at the same time as #5525.

PR #4976, opened 2018-08-21
Add back symbols to TASTy reflect

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
```diff
@@ -2,4 +2,4 @@ package dotty.tools.dotc.tastyreflect
 
 import dotty.tools.dotc.core._
```
liufengyun: Format the following line into several lines?

library/src/scala/tasty/Tasty.scala
```diff
@@ -2,4 +2,4 @@ package scala.tasty
 
 import scala.tasty.reflect._
```
liufengyun: Reformat the following line to be shorter?

library/src/scala/tasty/reflect/SymbolOps.scala
```diff
@@ -0,0 +1,14 @@
+package scala.tasty
+package reflect
+
+/** Tasty reflect symbol */
+trait SymbolOps extends TastyCore {
+
+  trait SymbolasAPI {
```
liufengyun: `SymbolasAPI ` -> `SymbolsAPI `

library/src/scala/tasty/reflect/SymbolOps.scala
```diff
@@ -0,0 +1,14 @@
...
+    def isEmpty: Boolean
+    def localContext(implicit ctx: Context): Context
+    def tree(implicit ctx: Context): Option[Definition]
+  }
+  implicit def SymbolasDeco(symbol: Symbol): SymbolasAPI
```
liufengyun: The same as above.

compiler/src/dotty/tools/dotc/tastyreflect/TastyImpl.scala
nicolasstucki: Done

library/src/scala/tasty/Tasty.scala
nicolasstucki: Done

PR #4977, opened 2018-08-21
Fix #3979: Completion for renamed imports

### LINKED ISSUE #3979: IDE completion should know about renaming imports.
Say you have

    import java.util.{List => JList}

Then scope completion with `J` should give `JList`. Right now it cannot to that because we return lists of symbols from completion requests, and `List` is clearly not a valid completion. 

For the moment, renamed imports are simply ignored. To fix this, it would be good to first have some tests for completion requests.

Duhemm:
We introduce a new `RenameAwareScope` that tracks the name associated
with a symbol in the scope. Calling `toListWithNames` returns the list
of symbols in this scope along with the name that should be used to
refer to the symbol in this scope.

Fixes #3979

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -96,15 +96,15 @@ object Interactive {
 
   /** Get possible completions from tree at `pos`
    *
-   *  @return offset and list of symbols for possible completions
+   *  @return offset and list of (symbol, name in scope) for possible completions
    */
-  def completions(pos: SourcePosition)(implicit ctx: Context): (Int, List[Symbol]) = {
+  def completions(pos: SourcePosition)(implicit ctx: Context): (Int, List[(Symbol, Name)]) = {
     val path = pathTo(ctx.compilationUnit.tpdTree, pos.pos)
     computeCompletions(pos, path)(contextOfPath(path))
   }
 
-  private def computeCompletions(pos: SourcePosition, path: List[Tree])(implicit ctx: Context): (Int, List[Symbol]) = {
-    val completions = Scopes.newScope.openForMutations
+  private def computeCompletions(pos: SourcePosition, path: List[Tree])(implicit ctx: Context): (Int, List[(Symbol, Name)]) = {
+    val completions = new RenameAwareScope
 
     val (completionPos, prefix, termOnly, typeOnly) = path match {
       case (ref: RefTree) :: _ =>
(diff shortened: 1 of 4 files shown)


Duhemm:
Rebased

Duhemm:
Rebased and took a stab at #5508

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -131,15 +142,27 @@ object Completion {
...
      *
-     * If the mode is `Import` and several symbols share the same name, the type symbols are
-     * preferred over term symbols.
+     * If several symbols share the same name, the type symbols appear before term symbols inside
+     * the same `Completion`.
+     */
+    def getCompletions(implicit ctx: Context): List[Completion] = {
+      val groupedSymbols = completions.toList.groupBy(_.name.stripModuleClassSuffix.toSimpleName).toList
+      groupedSymbols.map { case (name, symbols) =>
+        val typesFirst = symbols.sortWith((s, _) => s.isType)
```
smarter: to be a valid sort, comparing two type symbols should return false, so I'd do `(s1, s2) => s1.isType && !s2.isType`

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -131,15 +142,27 @@ object Completion {
...
+        // Use distinct to remove duplicates with class, module class, etc.
+        val descriptions = typesFirst.map(description).distinct.mkString(", ")
```
smarter: I would use `" and "` since `,` could be confused with type parameter lists and tuples

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -324,4 +330,23 @@ object Completion {
...
+     /** Enter the symbol `sym` in this scope, recording a potential renaming. */
+     def enter[T <: Symbol](sym: T, name: Name)(implicit ctx: Context): T = {
+       if (name != sym.name) renames += sym -> name
+       newScopeEntry(name, sym)
+       sym
+     }
+
+     /** Lists the symbols in this scope along with the name associated with them. */
+     def toListWithNames(implicit ctx: Context): List[(Symbol, Name)] =
+       toList.map(sym => (sym, renames.get(sym).getOrElse(sym.name)))
```
smarter:
```scala
renames.getOrElse(sym, sym.name)
```

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -301,4 +330,23 @@ object Completion {
     val Import: Mode = new Mode(4) | Term | Type
   }
 
+   /** A scope that tracks renames of the entered symbols.
+    *  Useful for providing completions for renamed symbols
+    *  in the REPL and the IDE.
+    */
+   private class RenameAwareScope extends Scopes.MutableScope {
+     private[this] val renames: mutable.Map[Symbol, Name] = mutable.Map.empty
```
smarter: I don't think that's enough, the same symbol could be imported with multiple names:
```scala
import java.io.{File => JFile}
import java.io.{File => JFile2}
```

Duhemm:
@smarter Thanks, I've addressed your comments.

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -301,4 +343,27 @@ object Completion {
...
+
+     /** Enter the symbol `sym` in this scope, recording a potential renaming. */
+     def enter[T <: Symbol](sym: T, name: Name)(implicit ctx: Context): T = {
+       nameToSymbols += name -> (sym :: nameToSymbols.getOrElse(name, Nil))
+       newScopeEntry(name, sym)
+       sym
+     }
+
+     /** Lists the symbols in this scope along with the name associated with them. */
+     def toListWithNames(implicit ctx: Context): List[(Symbol, Name)] = {
```
smarter: Hmm, why return a `List[(Symbol, Name)]` when `getCompletions` will convert it into a `List[Name, List[Symbol]]` which is almost what `nameToSymbols` already is ?

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -185,32 +223,35 @@ object Completion {
...
      *  The reason for (2) is that we do not want to present compiler-synthesized identifiers
      *  as completion results. However, if a user explicitly writes all '$' characters in an
      *  identifier, we should complete the rest.
      */
-    private def include(sym: Symbol)(implicit ctx: Context): Boolean =
-      sym.name.startsWith(prefix) &&
-      !sym.name.toString.drop(prefix.length).contains('$') &&
+    private def include(sym: Symbol, nameInScope: Name)(implicit ctx: Context): Boolean =
+      nameInScope.startsWith(prefix) &&
+      !nameInScope.toString.drop(prefix.length).contains('$') &&
```
smarter: Still not sure I understand why this is needed, can we remove it and see what fails ? :)

Duhemm: See this comment:
```
    /** Include in completion sets only symbols that
     *   1. start with given name prefix, and
     *   2. do not contain '$' except in prefix where it is explicitly written by user, and
     *   (...)
     *
     *  The reason for (2) is that we do not want to present compiler-synthesized identifiers
     *  as completion results. However, if a user explicitly writes all '$' characters in an
     *  identifier, we should complete the rest.
     */
```

We would have to remove the `object Foo$`s, `def foo$default$1`, etc.

compiler/src/dotty/tools/dotc/interactive/Completion.scala
Duhemm: Were you suggesting to return directly the `Map[Name, List[Symbol]]` here? That's what I did.

compiler/src/dotty/tools/dotc/interactive/Completion.scala
Duhemm: This was dead code since #3960, let me see how to plug it back in.

Duhemm:
@smarter The `NameFilter` is back, with the fix that we found yesterday.

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -357,12 +353,16 @@ object Completion {
...
      }
 
-     /** Lists the symbols in this scope along with the name associated with them. */
-     def toListWithNames(implicit ctx: Context): List[(Symbol, Name)] = {
-       for {
-         (name, syms) <- nameToSymbols.toList
-         sym <- syms
-       } yield (sym, name)
+     /** Get the names that are known in this scope, along with the list of symbols they refer to. */
+     def mappings(implicit ctx: Context): Map[SimpleName, List[Symbol]] = {
```
smarter: Couldn't the body of this method just be equal to `nameToSymbols` ? (assuming `def enter` above is changed to do `stripModuleClassSufffix`)

Duhemm: Yes, good catch. That should have been the case from the beginning, since `nameToSymbold` contained exactly one symbol per name. Thanks

compiler/src/dotty/tools/dotc/interactive/Completion.scala
```diff
@@ -350,26 +350,18 @@ object Completion {
     *  in the REPL and the IDE.
     */
    private class RenameAwareScope extends Scopes.MutableScope {
-     private[this] val nameToSymbols: mutable.Map[Name, List[Symbol]] = mutable.Map.empty
+     private[this] val nameToSymbols: mutable.Map[SimpleName, List[Symbol]] = mutable.Map.empty
 
      /** Enter the symbol `sym` in this scope, recording a potential renaming. */
      def enter[T <: Symbol](sym: T, name: Name)(implicit ctx: Context): T = {
-       nameToSymbols += name -> (sym :: nameToSymbols.getOrElse(name, Nil))
+       val simpleName = name.stripModuleClassSuffix.toSimpleName
```
smarter: How come `toSimpleName` is needed ? Aren't we already filtering out all non-simple-names with the completionsFilter ?

Duhemm: No, `completionsFilter` filters names based on their kind (`Simple`, `Qualified`, `Numbered` and so on), which is a different distinction from `SimpleName` / `TermName` / `TypeName` as far as I can tell.

We need to do `toSimpleName` here so that `TermName`s and `TypeName`s are grouped.

smarter: Ah, I see, I would to `toTermName` then since SimpleNames are TermNames anyway, and it means that we don't accidentally mangle encoded names.

PR #4979, opened 2018-08-21
Fix #4978: Complete imported symbol in REPL

### LINKED ISSUE #4978: Tab-completion doesn't word on imported symbol in REPL
In the REPL, none of the following would be completed with tab-completion:

```scala
scala> import java.util.HashMap
scala> val foo: HashM<tab> // No completion is offered, expected: "HashMap"
```

```scala
scala> import java.util.{HashMap => RenamedImport}
scala> val foo: Rena<tab> // No completion is offered, expected: "RenamedImport"
```

Duhemm:
The first scenario described in #4978 is fixed by this PR. Fixing the second case (completion of renamed imports) requires #4977.

### DIFF
--- a/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
+++ b/compiler/src/dotty/tools/dotc/interactive/Interactive.scala
@@ -94,13 +94,22 @@ object Interactive {
   private def safely[T](op: => List[T]): List[T] =
     try op catch { case ex: TypeError => Nil }
 
+  private def addExtraImports(extraImports: List[untpd.Import], ctx: Context): Context = {
+    extraImports.foldLeft(ctx) { case (c, i) => c.importContext(i, i.symbol(c)) }
+  }
+
   /** Get possible completions from tree at `pos`
    *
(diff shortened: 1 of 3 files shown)


compiler/src/dotty/tools/dotc/interactive/Interactive.scala
```diff
@@ -94,13 +94,22 @@ object Interactive {
   private def safely[T](op: => List[T]): List[T] =
     try op catch { case ex: TypeError => Nil }
 
+  private def addExtraImports(extraImports: List[untpd.Import], ctx: Context): Context = {
+    extraImports.foldLeft(ctx) { case (c, i) => c.importContext(i, i.symbol(c)) }
```
allanrenucci: `.symbol` on untyped tree?

allanrenucci:
I have an alternative fix in #4915: the REPL make sure to call `completions` with a context that is aware of the imports and `contextOfPath` then use this context as a base (instead of the run context) for the newly computed context

allanrenucci:
Closed in favour of #4915

PR #4982, opened 2018-08-21
Introduce Labeled blocks, and use them in PatternMatcher

### LINKED ISSUE #1313: PatMat:(add support guards in switches) implement common subexpression elimination
In Scalac pattern matcher is able to support guards inside switches by generating arbitrary GOTOs.

Unlike this, in Dotty I propose to implement more general optimization outside of pattern-matcher, that would also give us support for guards.

I propose to add a phase, that would follow the structure of matches, match common prefixes of pattern match, and restructure that pattern.

See following examples for illustration:

``` scala
x match {
  case List(1, 2, 3) => // 1
  case List(3, 4, 5) => // 2
  case 6 if pred     => // 3 
  case 6 =>             // 4
  case _ =>             // 5
}
```

should be rewritten as(a simplified version):

``` scala
x match {
  case t: List[Int] => 
    t match {
      case List(1, 2, 3) => // 1
      case List(3, 4, 5) => // 2
      case _ => fallback
    }
  case x @ 6 => 
      x match {
        case _ if pred =>     // 3 
        case _ =>             // 4
        case _ => fallback
     }
  case _ => 
    fallback
   }
}
<label> def fallback = // 5
```

### LINKED ISSUE #4563: Erasure tries to box call to label def
While migrating inox to Scala 3, I ran into runtime errors stemming from the following code, that has been worked around here: https://github.com/epfl-lara/inox/pull/71/commits/d68120a8b72ff273b2ff2b602a8be870e707de2c

Running the code _without_ the workaround leads to
```
[error] Exception in thread "main" java.lang.VerifyError: Bad type on operand stack
[error] Exception Details:
[error]   Location:
[error]     inox/solvers/theories/ASCIIStringEncoder.transform$$anonfun$1(C)Lscala/collection/GenTraversableOnce; @352: invokespecial
[error]   Reason:
[error]     Type top (current frame, stack[1]) is not assignable to reference type
[error]   Current Frame:
[error]     bci: @352
[error]     flags: { }
[error]     locals: { integer, 'scala/Some', 'scala/collection/Seq', 'scala/Some', 'scala/collection/Seq', top, top, 'scala/collection/Seq' }
[error]     stack: { top, top, 'java/lang/String' }
[error]   Bytecode:
(...)
[error]   Stackmap Table:
[error]     full_frame(@153,{Integer,Top,Object[#263],Object[#345],Object[#263],Integer,Integer},{})
[error]     full_frame(@162,{},{Object[#403]})
[error]     full_frame(@165,{Integer,Top,Object[#263],Object[#345],Object[#263],Integer,Integer},{Object[#356]})
[error]     chop_frame(@168,2)
[error]     full_frame(@175,{Integer,Object[#345],Object[#263],Object[#345],Object[#263]},{Top,Top})
[error]     full_frame(@254,{Integer,Object[#345],Object[#263],Object[#345],Object[#263],Top,Top,Object[#263]},{Top,Top})
[error]     full_frame(@341,{Integer,Object[#345],Object[#263],Object[#345],Object[#263],Top,Top,Object[#263]},{Top,Top})
[error]     full_frame(@344,{Integer,Object[#345],Object[#263],Object[#345],Object[#263],Top,Top,Object[#263]},{Top,Top,Object[#175]})
[error]     full_frame(@347,{Integer,Top,Object[#263],Object[#345],Object[#263]},{Uninitialized[#168],Uninitialized[#168]})
[error]     full_frame(@352,{Integer,Object[#345],Object[#263],Object[#345],Object[#263],Top,Top,Object[#263]},{Top,Top,Object[#175]})
[error]     full_frame(@355,{Integer,Top,Object[#263],Object[#345],Object[#263]},{Top})
[error]     chop_frame(@358,1)
[error]     full_frame(@365,{},{Object[#403]})
[error]     full_frame(@368,{Integer,Top,Object[#263],Object[#345],Object[#263]},{Top})
[error]     chop_frame(@371,2)
[error]     full_frame(@378,{Integer,Top,Object[#263]},{Top,Top})
[error]     full_frame(@387,{},{Object[#403]})
[error]     full_frame(@390,{Integer,Top,Object[#263]},{Uninitialized[#371],Uninitialized[#371]})
[error]     full_frame(@393,{},{Object[#403]})
[error]     full_frame(@396,{Integer,Top,Object[#263],Object[#345],Object[#263]},{Top})
[error] 	at inox.solvers.theories.ASCIIStringEncoder$.apply(ASCIIStringEncoder.scala:119)
[error] 	at inox.solvers.theories.package$.Z3(package.scala:16)
[error] 	at inox.solvers.SolverFactory$.getFromName(SolverFactory.scala:119)
[error] 	at inox.solvers.SolverFactory$.apply(SolverFactory.scala:349)
[error] 	at inox.solvers.SolverFactory$.apply(SolverFactory.scala:366)
[error] 	at inox.package$$anon$3.createSolver(package.scala:77)
[error] 	at inox.Semantics.getSolver$$anonfun$1(Semantics.scala:31)
[error] 	at inox.utils.Cache.cached(Caches.scala:14)
[error] 	at inox.Semantics.getSolver(Semantics.scala:31)
```

The underlying problem here is that dotc emits code trying to box a call to a label:
```scala
new scala.collection.immutable.StringOps(
  scala.collection.immutable.StringOps.evt2u$(case8(x31)))
```
where `case8` is
```scala
def case8(case x34: Some): ErasedValueType(
  scala.collection.immutable.StringOps, String) = ...
```

I don't have a minimized test case unfortunately, but, as a starting point, this is how I reproduce it:
```
dotc  -deprecation -unchecked -feature -language:Scala2   -Xprint:erasure  -classpath /home/gs/epfl/D/inox-dotty/unmanaged/scalaz3-unix-64-2.12.jar:/home/gs/.ivy2/local/ch.epfl.lamp/scala-library/0.8.0-bin-SNAPSHOT/jars/scala-library.jar:/home/gs/.ivy2/local/ch.epfl.lamp/dotty-library_0.8/0.8.0-bin-SNAPSHOT/jars/dotty-library_0.8.jar:/home/gs/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.4.jar:/home/gs/.ivy2/cache/org.apache.commons/commons-lang3/jars/commons-lang3-3.4.jar:/home/gs/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.12.4.jar:/home/gs/.ivy2/cache/com.regblanc/scala-smtlib_2.12/jars/scala-smtlib_2.12-0.2.2-7-g00a9686.jar:/home/gs/.ivy2/cache/uuverifiers/princess_2.12/jars/princess_2.12-2016-12-26.jar:/home/gs/.ivy2/cache/uuverifiers/princess-parser_2.12/jars/princess-parser_2.12-2016-12-26.jar:/home/gs/.ivy2/cache/uuverifiers/princess-smt-parser_2.12/jars/princess-smt-parser_2.12-2016-12-26.jar:/home/gs/.ivy2/cache/org.scala-lang.modules/scala-parser-combinators_2.12/bundles/scala-parser-combinators_2.12-1.0.4.jar:/home/gs/.ivy2/cache/net.sf.squirrel-sql.thirdparty-non-maven/java-cup/jars/java-cup-0.11a.jar:/home/gs/epfl/D/inox-dotty/target/scala-0.8/classes/      /home/gs/epfl/D/inox-dotty/src/main/scala/inox/solvers/theories/ASCIIStringEncoder.scala
```

sjrd:
A Labeled block is an expression tree of the form
```scala
label[T]: {
  expr
}
```
where `expr` must conform to the type `T`. In addition, within `expr` (but nowhere else), return from the label is allowed:
```scala
return[label] e
```
where `e` must conform to the type `T` as well.

If execution of `expr` completes normally (rather than throwing an exception or returning, etc.), then the result of evaluating the `Labeled` block is the result of `expr`. If a `return[label] e` is reached, the execution of `expr` is interrupted, and the result of evaluating the `Labeled` block is the result of evaluating the argument `e`.

Implementation-wise, a `Labeled` block is represented as a `Tree` with the shape:
```scala
Labeled(Bind(labelName), expr)
```
where the `Bind` nodes holds the definition of the label symbol. That symbol is a term symbol with the flag `Label` (but not `Method`, unlike symbols for label-defs) and whose `info` is the result type `T` of the labeled block.

We use those new `Labeled` blocks in `PatternMatcher`, instead of label-defs. This is the first step towards completely removing label-defs from the compiler.

This commit structurally fixes a few issues:

* It fixes #1313 through the `mergeTests` optimization.
* It fixes #4563 because Labeled blocks are erasure-friendly.
* It does a big step towards fixing the upstream test t10387: the compiler can get to the back-end on that test, but it produces too much bytecode for a single JVM method. We do add a sister test t10387b which works because optimizations can kick in.


sjrd:
Can anyone tell me how I can get the CI to take into account the version of scala-backend I have pushed at https://github.com/dotty-staging/scala/tree/dotty-labeled?

smarter:
Push it to a branch on http://github.com/lampepfl/scala or replace lampepfl by dotty-staging in .gitmodules under the scala-backend section.

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
```diff
@@ -390,46 +391,47 @@ object TastyFormat {
...
-  final val TYPELAMBDAtype = 169
-  final val LAMBDAtpt = 170
-  final val PARAMtype = 171
-  final val ANNOTATION = 172
-  final val TERMREFin = 173
-  final val TYPEREFin = 174
-  final val OBJECTDEF = 175
-
-  // In binary: 101100EI
+  final val LABELED = 144
```
smarter: Need to bump the tasty MajorVersion in this file.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -352,8 +352,14 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
         else changePrec(GlobalPrec) { toText(sel) ~ keywordStr(" match ") ~ blockText(cases) }
       case CaseDef(pat, guard, body) =>
         keywordStr("case ") ~ inPattern(toText(pat)) ~ optText(guard)(keywordStr(" if ") ~ _) ~ " => " ~ caseBlockText(body)
+      case Labeled(bind, expr) =>
+        changePrec(GlobalPrec) { toText(bind.name) ~ keywordStr("[") ~ toText(bind.symbol.info) ~ keywordStr("]: ") ~ toText(expr) }
       case Return(expr, from) =>
-        changePrec(GlobalPrec) { keywordStr("return") ~ optText(expr)(" " ~ _) }
+        val sym = from.symbol
+        if (sym.is(Label))
+          changePrec(GlobalPrec) { keywordStr("return@") ~ toText(sym.name) ~ optText(expr)(" " ~ _) }
```
smarter: Very minor bikeshed but I'm not a fan of using `@` since it looks too much like an annotation. Maybe write `return[foo]` instead ?

sjrd:
> Push it to a branch on http://github.com/lampepfl/scala or replace lampepfl by dotty-staging in .gitmodules under the scala-backend section.

Thanks. That worked :)

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
sjrd: I don't think so. I think the last bump to version 10 happened after the latest release.

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
sjrd: Sure, whatever ;)

compiler/src/dotty/tools/dotc/core/tasty/TastyFormat.scala
smarter: We should always bump when we break tasty. If we merge and you don't bump, people will start getting confusing error messages in their build when incremental compilation mixes old and new tasty files.

sjrd:
Good to review :-D Who feels like it?

smarter:
What's missing to completely remove labeldefs?

smarter:
test performance please

nicolasstucki:
Need to update the `TreeChecker` to check that label returns are inside their labeled block.

compiler/src/dotty/tools/dotc/core/tasty/TreePickler.scala
```diff
@@ -426,6 +426,9 @@ class TreePickler(pickler: TastyPickler) {
         case CaseDef(pat, guard, rhs) =>
           writeByte(CASEDEF)
           withLength { pickleTree(pat); pickleTree(rhs); pickleTreeUnlessEmpty(guard) }
+        case Labeled(bind, expr) =>
+          writeByte(LABELED)
+          withLength { pickleTree(bind); pickleTree(expr) }
```
nicolasstucki: Right now we will never pickle `Labeled` trees as they are generated in pattern matcher. We should add this code only when we will need it for `while` and `do while`

sjrd:
> What's missing to completely remove labeldefs?

tailrec-generated labels, and `while`/`do..while` loop-generated ones.

smarter:
@liufengyun any idea why "test performance please" above didn't work?

liufengyun:
@smarter the polling service ended, just restarted.

test performance please

smarter:
It'd be interesting to check if https://gist.github.com/smarter/1c564e65514d328462e7453162ccfb65 which works in scalac (but was removed from the test suite for being too slow) now works with this PR (it stackoverflows on master).

sjrd:
It "works" as far as the compiler itself is concerned, but it produces too much code for the JVM:
```
java.lang.RuntimeException: Method integer_of_char's code too large!7s)
        at scala.tools.asm.MethodWriter.getSize(MethodWriter.java:2070)
        at scala.tools.asm.ClassWriter.toByteArray(ClassWriter.java:861)
        at dotty.tools.backend.jvm.GenBCodePipeline$Worker2.getByteArray$1(GenBCode.scala:336)
```

nicolasstucki:
At least we do not crash before reaching that limit.

smarter:
@sjrd interesting, I suppose that's something that would be fixed with more optimizations in the pattern matcher?

sjrd:
Yes and no. One thing would be to short-circuit `unapply` calls for `case class`es. Apparently dotty only does that for Scala2 case classes, but not for its own. (this is part of the "plan building" which I did not touch in this PR)

Other than that, there's not much room for optimization of this particular pattern-match. It is specifically designed to prevent any kind of optimization: the bits are always alternating at the top-level, so the optimizer can never collapse two consecutive cases. It has to redo all the tests from the top every time. scalac does not fare better than my PR, here.

nicolasstucki:
Just for curiosity, if you reorder the cases to not alternate does it optimize and compile?

nicolasstucki:
I mean this code https://gist.github.com/nicolasstucki/36741b8cb63199c8ef442fff1dffaf3f

sjrd:
Yes, it does. The tree after patmat is https://gist.github.com/sjrd/803ddf59e4d17a5ffa445e3f3c16dfd4 (around 5 kLoC, versus 17 kLoC for the original, failing one).

compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -992,10 +894,7 @@ object PatternMatcher {
     }
 
     val optimizations: List[(String, Plan => Plan)] = List(
-      "hoistLabels" -> hoistLabels,
-      "elimRedundantTests" -> elimRedundantTests,
-      "inlineLabelled" -> inlineLabelled,
-      "mergeVars" -> mergeVars,
+      "mergeTests" -> mergeTests,
```
smarter: Could the removed optimizations be brought back later? Or are they subsumed by mergeTests ?

sjrd: They are subsumed by `mergeTests`.

smarter:
Would be good to verify that this PR fixes https://github.com/lampepfl/dotty/issues/4563

sjrd:
> Remove changes from pickling for now.

Done.

> Would be good to verify that this PR fixes #4563

It does. I added a test case.

tests/pos/i4563.scala
```diff
@@ -0,0 +1,12 @@
+case class Foo(str: String)
```
allanrenucci: This test already compiles and runs on master. TBH, I am not sure this minimises the original issue.

Anyway, this was reported as a runtime issue, so this should be a run test

smarter: This is missing the extends AnyVal

sjrd:
> Need to update the TreeChecker to check that label returns are inside their labeled block.

Done.

I think I've addressed all the comments that have been raised so far.

smarter:
Assigning to @odersky since he wrote most of the current PatternMatcher

sjrd:
This PR also fixes #1313. Test case to be added when it's not 1am:
```scala
object HelloWorld {
  def main(args: Array[String]): Unit = println("hello world")

  def test(x: Any, y: Int): Int = {
    x match {
      case List(1, 2, 3) => 1
      case List(3, 4, 5) => 2
      case 6 if y > 5 => 3
      case 6 => 4
      case _ => 5
    }
  }

  def test2(x: Int, y: Int): Int = {
    (x: @scala.annotation.switch) match {
      case 6 if y > 5 => 1
      case 6 => 2
      case 12 => 3
      case 14 => 4
      case _ => 5
    }
  }
}
```
No "switch" warning! Tree after the pattern matcher:
```scala
package <empty> {
  final lazy module val HelloWorld: HelloWorld$ = new HelloWorld$()
  @scala.annotation.internal.SourceFile("tests\\pos\\HelloWorld.scala")
    final module
   class HelloWorld$() extends Object() {
    def main(args: Array[String]): Unit = println("hello world")
    def test(x: Any, y: Int): Int =
      {
        matchResult1[Int]:
          {
            case val x1: Any(x) = x
            if x1.$isInstanceOf$[List[Any]] then
              {
                case val x8: List[Any] = x1.asInstanceOf[List[Any]]
                case val x9: Some[List[Any]] = List.unapplySeq[Any](x8)
                if x9.isEmpty.unary_! then
                  {
                    case val x10: List[Any] = x9.get
                    if x10.lengthCompare(3).==(0) then
                      {
                        case val x11: Any = x10.apply(0)
                        case val x12: Any = x10.apply(1)
                        case val x13: Any = x10.apply(2)
                        if 1.==(x11).&&(2.==(x12)).&&(3.==(x13)) then
                          return[matchResult1] 1
                         else ()
                        if 3.==(x11).&&(4.==(x12)).&&(5.==(x13)) then
                          return[matchResult1] 2
                         else ()
                      }
                     else ()
                  }
                 else ()
              }
             else ()
            if 6.==(x1) then
              {
                if y.>(5) then return[matchResult1] 3 else ()
                return[matchResult1] 4
              }
             else ()
            return[matchResult1] 5
            return[matchResult1] throw new MatchError(x1)
          }
      }
    def test2(x: Int, y: Int): Int =
      {
        matchResult2[Int]:
          {
            case val x14: Int(x) @switch = x: Int(x) @switch
            x14 match
              {
                case 6 =>
                  if y.>(5) then return[matchResult2] 1 else ()
                  return[matchResult2] 2
                case 12 => return[matchResult2] 3
                case 14 => return[matchResult2] 4
                case _ =>
                  return[matchResult2] 5
                  return[matchResult2] throw new MatchError(x14)
              }
          }
      }
  }
}
```

odersky:
Good to merge unless @sjrd has something to add.

sjrd:
Thank you :)

I don't have anything to add. Changes to loops and tailrec are best left as separate PRs. Let's ship it!

sjrd:
Oh, I'm just going to update the submodule commit reference to the new HEAD of `sharing-backend` (the merge commit).

PR #4983, opened 2018-08-22
Fix JLine dependencies

smarter:
jline-$VERSION.jar is a bundle that contains the content of
jline-terminal-$VERSION.jar and other modules. jline-terminal-jna
depends on jline-terminal. This means that the Dotty build depends on
both jline and jline-terminal and classfiles may be read from either
jar (dependencies on the classpath are usually not in a stable order).
They should be identical so it's not a big problem, but it breaks Zinc
incremental compilation. This commit fixes this by not using the jline
bundle jar anymore.


liufengyun:
We also need to change the dotc script:

https://github.com/lampepfl/dotty/blob/master/dist/bin/common#L122-L124
https://github.com/lampepfl/dotty/blob/master/dist/bin/dotc#L69-L71

smarter:
@liufengyun Thanks for the info, done.

PR #4990, opened 2018-08-22
Simplify enclosedInlineds

Blaisorblade:
WIP waiting full tests, then I explain why this is a refactoring.

Bring both `inlineContext` and `enclosingInlineds` from O(n) to O(1) without complicating the code.

Noticed while reviewing #4949.

### DIFF
--- a/compiler/src/dotty/tools/dotc/ast/tpd.scala
+++ b/compiler/src/dotty/tools/dotc/ast/tpd.scala
@@ -1083,13 +1083,16 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
   private val InlinedCalls = new Property.Key[List[Tree]]
 
   /** Record an enclosing inlined call.
-    * EmptyTree calls (for parameters) cancel the next-enclosing non-empty call in the list
+    * EmptyTree calls (for parameters) cancel the next-enclosing call in the list instead of being added to it.
+    * We assume parameters are never nested inside parameters.
     */
   override def inlineContext(call: Tree)(implicit ctx: Context): Context = {
     // We assume enclosingInlineds is already normalized, and only process the new call with the head.
     val oldIC = enclosingInlineds
     val newIC = (call, oldIC) match {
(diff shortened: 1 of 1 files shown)


nicolasstucki:
If we normalize too early w will not have the full inline stack which implies that we may not be able to use it for improving the error messages. Though it is possible that normalized stack is enough. @odersky WDYT?

Blaisorblade:
The existing code already normalizes the stack too much for your use case. It might have a rationale, I haven't found it yet, tho I'm not done studying this.

OTOH, maybe the context chain has enough information anyway. If not, we can change again, but not to the existing code.

compiler/src/dotty/tools/dotc/ast/tpd.scala
```diff
@@ -1082,21 +1082,27 @@ object tpd extends Trees.Instance[Type] with TypedTreeInfo {
...
-   *  EmptyTree calls cancel the next-enclosing non-empty call in the list
-   */
-  def enclosingInlineds(implicit ctx: Context): List[Tree] = {
-    def normalize(ts: List[Tree]): List[Tree] = ts match {
-      case t :: (ts1 @ (t1 :: ts2)) if t.isEmpty => normalize(if (t1.isEmpty) ts1 else ts2)
-      case t :: ts1 => t :: normalize(ts1)
-      case Nil => Nil
+  /** Record an enclosing inlined call.
+    * EmptyTree calls (for parameters) cancel the next-enclosing call in the list instead of being added to it.
+    * We assume parameters are never nested inside parameters.
```
Blaisorblade: WIP further experiments suggest this can't be true, even tho this code works.

nicolasstucki: After normalization, we should not have them, so it should be fine or the normalization was not performed correctly.

Blaisorblade: True after normalization — I thought that was also true before normalization, but that was wrong.

PR #4991, opened 2018-08-22
Use symbol instead of definition tree in TASTy reflect

nicolasstucki:
* Move operations from `Definition` to `Symbol` to avoid unnecessary loading of trees
* Return `Symbol` instead of `Definition` avoid unnecessary loading of trees
* Use operations on `Symbol`


compiler/src/dotty/tools/dotc/tastyreflect/SymbolOpsImpl.scala
```diff
@@ -1,14 +1,49 @@
...
 
 import dotty.tools.dotc.core.Symbols._
 
 trait SymbolOpsImpl extends scala.tasty.reflect.SymbolOps with TastyCoreImpl {
 
   def SymbolDeco(symbol: Symbol): SymbolAPI = new SymbolAPI {
-    def isEmpty: Boolean = symbol eq NoSymbol
-    def localContext(implicit ctx: Context): Context = ctx.withOwner(symbol)
+
+    def isEmpty(implicit ctx: Context): Boolean = symbol eq NoSymbol
```
liufengyun: Can we rename it to `exists`? Not important, somehow I feel `exists` is more friendly.

nicolasstucki: Renamed

PR #4992, opened 2018-08-23
Fix #4951: Increase stack size before bootstrap to reduce tests flakyness

### LINKED ISSUE #4951: Inductive-implicit-benchmark fails locally before bootstrap but not on CI
Since a while, `testCompilation` fails on `inductive-implicits-bench.scala` with a stack overflow before bootstrap, but not after bootstrap or on CI. This passes after bootstrap, I think, because of methods that are only tail-recursive if `track` is inlined. And it passes with `-Xss2m` in `agentOptions` (apparently neither sbt's JVM options nor ones in `RunnerOrchestration` are involved).

Still, why does this pass in the CI? @nicolasstucki wonders if it's due to Linux on the CI (I and @nicolasstucki are on Mac and both see the issue), I wonder if it's due to a different JVM version (FWIW I'm on 1.8.0_172-b11).

```
[info] Test dotty.tools.dotc.CompilationTests.pos started
-- Error: /Users/pgiarrusso/git/dotty-master-nonwt/tests/pos-deep-subtype/inductive-implicits-bench.scala:60:20 --------
60 |  val sel = Selector[L, Boolean]
   |                    ^
   |Recursion limit exceeded.
   |Maybe there is an illegal cyclic reference?
   |If that's not the case, you could also try to increase the stacksize using the -Xss JVM option.
   |A recurring operation is (inner to outer):
   |
   |  subtype H <:< Int
   |  subtype H :: T <:< Int :: Int :: Int :: Int :: Int :: Int :: Int :: Boolean :: shapeless.HNil
   |  subtype shapeless.Selector[H :: T, H] <:< shapeless.Selector[Int :: Int :: Int :: Int :: Int :: Int :: Int :: Boolean :: shapeless.HNil, Boolean]
[=======================================] completed (1/1, 1 failed, 3s)
[error] Test dotty.tools.dotc.CompilationTests.pos failed: java.lang.AssertionError: Expected no errors when compiling, failed for the following reason(s):
[error]
[error]   - encountered 1 test failures(s), took 4.157 sec
[error]     at dotty.tools.vulpix.ParallelTesting$CompilationTest.checkCompile(ParallelTesting.scala:959)
[error]     at dotty.tools.dotc.CompilationTests.pos(CompilationTests.scala:101)
[error]     ...
[info] Test dotty.tools.dotc.CompilationTests.tastyBootstrap started
No files matched "inductive-implicits-bench" in test
No files matched "inductive-implicits-bench" in test
No files matched "inductive-implicits-bench" in test
No files matched "inductive-implicits-bench" in test
[info] Test dotty.tools.dotc.CompilationTests.posTwice started
No files matched "inductive-implicits-bench" in test
[info] Test dotty.tools.dotc.CompilationTests.genericJavaSignatures started
No files matched "inductive-implicits-bench" in test

================================================================================
Test Report
================================================================================

0 suites passed, 1 failed, 1 total
    tests/pos-deep-subtype/inductive-implicits-bench.scala failed
```

Blaisorblade:
To test, use `testCompilation inductive-implicits-bench` before and after,
locally. I get a failure before and a pass after.

We consume more stack before bootstrap because many typer methods are only tail-recursive if `track` or such are inlined (via `@forceInline`), and we don't plan to fix that. Viceversa, if this test stack overflows *after* bootstrap, we probably want to take a look.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -718,6 +718,9 @@ object Build {
       TestFrameworks.JUnit,
       "--exclude-categories=dotty.BootstrappedOnlyTests",
     ),
+    // increase stack size for non-bootstrapped compiler, because some code
+    // is only tail-recursive after bootstrap
+    javaOptions in Test += "-Xss2m"
   )
 
   lazy val bootstrapedDottyCompilerSettings = commonDottyCompilerSettings ++ Seq(


project/Build.scala
```diff
@@ -718,6 +718,9 @@ object Build {
       TestFrameworks.JUnit,
       "--exclude-categories=dotty.BootstrappedOnlyTests",
     ),
+    // increase stack size for non-bootstrapped compiler, because some code
+    // is only tail-recursive after bootstrap
+    javaOptions += "-Xss2m"
```
allanrenucci: `javaOptions  in Test`?

Blaisorblade:
Comments addressed (and tested), mergeable when CI passes.

PR #4993, opened 2018-08-23
Split TASTy symbol

nicolasstucki:
Based on #4991


library/src/scala/tasty/reflect/TastyCore.scala
```diff
@@ -164,5 +170,11 @@ trait TastyCore {
    *  Then can be compared with == to know if the definition is the same.
    */
   type Symbol
+    type PackageSymbol <: Symbol
+    type ClassSymbol <: Symbol
+    type TypeSymbol <: Symbol
+    type DefSymbol <: Symbol
+    type ValSymbol <: Symbol
+    type NoSymbol <: Symbol
```
liufengyun: What is the motivation for this change, if all of them have the same interface?

nicolasstucki: They don't have the same interface. `DefSymbol.tree` returns a `DefDef`, `ValSymbol.tree` returns a `ValDef` and so on. The opposite way is also true. It also provides a way to have an extractor for precise symbols for example `case SymRef(IsClassSymbol(sym), _) => sym.tree: ClassDef`

Blaisorblade: Then maybe the type should be `Symbol[Tree]`, writing then `Symbol[DefDef]` for `DefSymbol`, and so on?

nicolasstucki: There is another issue that is not show in this code yet. There is also a tree for a `Bind` which is not a subtype of `Tree`. This complicates a bit the situation. In fact what we would need is a `Symbol[Definition]` where `Definition` is a supertype of `XYZDef` and `Bind` but cannot be a subtype of `Tree`. I still need to explore all the options.

nicolasstucki:
Rebased

PR #4806, opened 2018-07-18
Normalize TypeOf types (second step towards typed transparent methods)

OlivierBlanvillain:
This PR follows #4671. It implements parts of the type normalization. Pattern matching is still missing and will come as a 3rd PR.

This is still work in progress, we open it to run the tests on the CI.

PR #4997, opened 2018-08-24
First steps to support dependent matches

gsps:
This PR is based on our ongoing work to add dependent types via the `TypeOf` encoding and extends #4806.

At the moment this branch only affects the precision of types given to bindings inside of match patterns. For instance, given
```scala
dependent def headOfCons1 =
  Cons(1, Cons(2, Nil)) match {
    case Cons(x, _) => x
  }
```
we can now reason about the extracted `x` in relation to the scrutinee `Cons(1, Cons(2, Nil))`, so that
```scala
headOfCons1: 1
```
can be proven by the type checker.

Note that this PR currently does not do any reduction of matches on the type level, so a match with multiple cases will not behave any differently yet.


odersky:
@gsps @OlivierBlanvillain Inactive for 3 months with lots of accumulated conflicts. Should we close this now?

OlivierBlanvillain:
We still indent to work on this PR. It's currently work in progress on is builds on top of #4806 and #4671 which where both indented as intermediate reviewable/mergable milestones. I guess we can close #4806 as resolving conflicts on an intermediate state would not be a very good use of time...

odersky:
@OlivierBlanvillain OK, please go ahead and close #4806.

b-studios:
@OlivierBlanvillain do you still intend to merge this / work on this?

PR #4998, opened 2018-08-24
Fix ambiguous unapply for TASTy reflect Parent

nicolasstucki:


compiler/src/dotty/tools/dotc/tastyreflect/TreeOpsImpl.scala
```diff
@@ -399,4 +397,5 @@ trait TreeOpsImpl extends scala.tasty.reflect.TreeOps with TastyCoreImpl with He
     }
   }
 
+  def termAsParent(term: Term): Parent = term
```
liufengyun: This seems dangerous to me, as any term can now be disguised as `Parent`. The same for `TypeTree`. Conceptually, how it is different from making `Parent` a supertype of `Term` and `TypeTree`?

nicolasstucki: It is not conceptually different, but it solves the ambiguity in the extractor when passing a `Term` as term could be a `Tree` or a `Parent`. This will be removed when we bootstrap and I'll be able to define `type Parent = Term | TypeTree` and have a simple extractor for both.

PR #5009, opened 2018-08-24
enclosingInlineds: Clarify invariants

Blaisorblade:
WIP, followup to #4990 and #4949.

I also tried ensuring parameters are always nested in calls and boy is that false - we forget using inlineContext often enough, which could be a trap if we use implicit conversion `sourcePos`.

### DIFF
--- a/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
+++ b/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
@@ -385,8 +385,8 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
       case SeqLiteral(elems, elemtpt) =>
         "[" ~ toTextGlobal(elems, ",") ~ " : " ~ toText(elemtpt) ~ "]"
       case tree @ Inlined(call, bindings, body) =>
-        (("/* inlined from " ~ toText(call) ~ " */ ") `provided`
-          !call.isEmpty && !homogenizedView && !ctx.settings.YshowNoInline.value) ~
+        ((if (!call.isEmpty) "/* inlined from " ~ toText(call) ~ " */ " else "/* inline arg */ ": Text) `provided`
+          !homogenizedView && !ctx.settings.YshowNoInline.value) ~
         blockText(bindings :+ body)
       case tpt: untpd.DerivedTypeTree =>
         "<derived typetree watching " ~ summarized(toText(tpt.watched)) ~ ">"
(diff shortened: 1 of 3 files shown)


Blaisorblade:
Oh interesting, the extra test seems to have been broken by #4949/#4990.

After #4949, calling from `Test.main` transparent function from the same file `Test.assert2` causes no repositioning; but the output trees actually come from `Macros` so they must be repositioned. This was an existing bug (which affected sourcePos even earlier); to fix this without changing enclosedInlineds, one needs 
a node `Inlined(`Macros.assertImpl(...)`, ...)` around the quote body, in addition.

Especially if there can be both code from the transparent function and from the splice, we'd need something like
```scala
Inlined(assert2(...), ...,
  Inlined(Macros.assertImpl(...), ..., {
    printStack("assertImpl")
    println(Inlined(EmptyTree, Nil, Inlined(EmptyTree, Nil, x != 0)))
  })
)
```
We do need two EmptyTree around x != 0 to signal it does come from the original call site. Not 100% sure what goes in the ellipses.

PostTyper can strip the call as usual, no problem.

Blaisorblade:
Moved the failing test to #5048; this PR is now cleanup-only.

Blaisorblade:
Failure is spurious (#5021).

Blaisorblade:
@nicolasstucki I think you picked everything useful from here? Can I close? Assigning to you for the moment (feel free to reassign back).

compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala
```diff
@@ -385,8 +385,8 @@ class RefinedPrinter(_ctx: Context) extends PlainPrinter(_ctx) {
       case SeqLiteral(elems, elemtpt) =>
         "[" ~ toTextGlobal(elems, ",") ~ " : " ~ toText(elemtpt) ~ "]"
       case tree @ Inlined(call, bindings, body) =>
-        (("/* inlined from " ~ toText(call) ~ " */ ") `provided`
-          !call.isEmpty && !homogenizedView && !ctx.settings.YshowNoInline.value) ~
+        ((if (!call.isEmpty) "/* inlined from " ~ toText(call) ~ " */ " else "/* inline arg */ ": Text) `provided`
```
nicolasstucki: This change should be reverted, we already do this with a slightly different logic.

odersky:
@Blaisorblade Can we get this over the finish line please?

Blaisorblade:
Asked, @nicolasstucki confirms this code is going to be refactored significantly by your current position changes. So closing.

PR #5012, opened 2018-08-24
Fix #4999: allow pattern matching against by-name arguments

### LINKED ISSUE #4999: By name match selector
```scala
trait Foo
final class Bar extends Foo

class Test {
  def test(xs: => Foo) = xs match {
    case xs: Bar => 1
    case _       => 2
  }
}
```
```scala
-- Error: tests/allan/Test.scala:104:9 -----------------------------------------
104 |    case xs: Bar => 1
    |         ^
    |this case is unreachable since trait Function0 and class Bar are unrelated
```
However, this is not an issue with the exhaustivity checker. It fails `Ycheck`:
```scala
checking tests/allan/Test.scala after phase MegaPhase{pruneErasedDefs, vcInlineMethods, seqLiterals, intercepted, getters, elimByName, collectNullableFields, elimOuterSelect, augmentScala2Traits, resolveSuper, primitiveForwarders, functionXXLForwarders, arrayConstructors}
exception while typing case val x1: (() => Foo)(xs) = xs.apply() of class class dotty.tools.dotc.ast.Trees$ValDef # 777
exception while typing {
  case val x1: (() => Foo)(xs) = xs.apply()
  if x1.$isInstanceOf$[Bar] then 
    {
      case val x2: Bar = x1.asInstanceOf[Bar]
      case val xs: Bar = x2
      1
    }
   else 2
} of class class dotty.tools.dotc.ast.Trees$Block # 778
...
Exception in thread "main" java.lang.AssertionError: assertion failed: found:    Foo
required: (() => Foo)(xs)
```
The type we get for `case val x1` after `ElimByName` (`(() => Foo)(xs)`) is clearly wrong. Should be `Foo`

Blaisorblade:
In this testcase `tp` is `TermRef(NoPrefix, xs)` so `widenExpr` does nothing, but
`widenTermRefExpr` is appropriate.

### DIFF
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -116,7 +116,7 @@ object PatternMatcher {
     /** Widen type as far as necessary so that it does not refer to a pattern-
      *  generated variable.
      */
-    private def sanitize(tp: Type): Type = tp.widenExpr match {
+    private def sanitize(tp: Type): Type = tp.widenTermRefExpr match {
       case tp: TermRef if refersToInternal(false, tp) => sanitize(tp.underlying)
       case tp => tp
     }
--- a/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
+++ b/compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
@@ -116,7 +116,7 @@ object PatternMatcher {
     /** Widen type as far as necessary so that it does not refer to a pattern-
      *  generated variable.
      */
-    private def sanitize(tp: Type): Type = tp.widenTermRefExpr match {
+    private def sanitize(tp: Type): Type = tp.widenIfUnstable match {
       case tp: TermRef if refersToInternal(false, tp) => sanitize(tp.underlying)
       case tp => tp
     }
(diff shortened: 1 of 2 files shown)


compiler/src/dotty/tools/dotc/transform/PatternMatcher.scala
```diff
@@ -116,7 +116,7 @@ object PatternMatcher {
     /** Widen type as far as necessary so that it does not refer to a pattern-
      *  generated variable.
      */
-    private def sanitize(tp: Type): Type = tp.widenExpr match {
+    private def sanitize(tp: Type): Type = tp.widenIfUnstable match {
```
allanrenucci: There is also `widenTermRefExpr` that seems to work here. Not sure which one is the most appropriate one. I defer to @odersky or @smarter

Blaisorblade: Yeah the first commit uses `widenTermRefExpr`. Maybe @liufengyun can think of examples where widening singletons would make a difference?

smarter: What about `x match { case _: x.type => ... }` ?

allanrenucci: We concluded with @smarter that there is no need to widen singleton types

Issue #5010, opened 2018-08-24
TypeError when a class extends from Function0


alexknvl:
```scala
class i0 extends Function0
```
```scala
Exception in thread "main" dotty.tools.dotc.core.TypeError: bad parameter reference i0.this.R at tryCatchPatterns
the parameter is type R in trait Function0 but the prefix i0(i0.this)
does not define any corresponding arguments.
	at dotty.tools.dotc.core.Types$NamedType.argDenot(Types.scala:1841)
	at dotty.tools.dotc.core.Types$NamedType.fromDesignator$1(Types.scala:1756)
	at dotty.tools.dotc.core.Types$NamedType.computeDenot(Types.scala:1774)
	at dotty.tools.dotc.core.Types$NamedType.denot(Types.scala:1730)
	at dotty.tools.dotc.core.Types$NamedType.info(Types.scala:1719)
	at dotty.tools.dotc.core.Types$$anon$5.apply(Types.scala:2860)
	at dotty.tools.dotc.core.Types$$anon$5.apply(Types.scala:2853)
	at dotty.tools.dotc.core.Types$TermLambda.dotty$tools$dotc$core$Types$TermLambda$$depStatus(Types.scala:2870)
	at dotty.tools.dotc.core.Types$TermLambda.dependencyStatus(Types.scala:2884)
	at dotty.tools.dotc.core.Types$TermLambda.resultType(Types.scala:2827)
	at dotty.tools.dotc.core.Types$MethodType.resultType(Types.scala:2931)
	at dotty.tools.dotc.printing.PlainPrinter.toText$$anonfun$8$$anonfun$6(PlainPrinter.scala:180)
	at dotty.tools.dotc.printing.Printer.atPrec(Printer.scala:41)
	at dotty.tools.dotc.printing.Printer.changePrec(Printer.scala:70)
	at dotty.tools.dotc.printing.PlainPrinter.toText$$anonfun$1(PlainPrinter.scala:182)
	at dotty.tools.dotc.printing.PlainPrinter.controlled(PlainPrinter.scala:28)
	at dotty.tools.dotc.printing.PlainPrinter.toText(PlainPrinter.scala:215)
	at dotty.tools.dotc.printing.RefinedPrinter.toText$$anonfun$1(RefinedPrinter.scala:243)
	at dotty.tools.dotc.printing.PlainPrinter.controlled(PlainPrinter.scala:28)
[...]

Issue #5008, opened 2018-08-24
Forbid enum extending enum


Blaisorblade:
We should forbid this code (discovered in #4961), we already forbid case classes/objects extending enum:

```scala
object enumTest {
  enum Test {
    case A
    case B
    case C
  }

  enum Test2 extends Test {
    case D
  }
}
```

Blaisorblade:
I'm guessing refchecks because we can't necessarily resolve `Test` in the extends clause before typer (beware cycles!), but another phase (typer/posttyper) might be appropriate.

Blaisorblade:
@hrhino also asks about

```scala
enum A ; class B extends A { val enumTag = 1 }
```

which doesn't seem any better.

PR #5017, opened 2018-08-25
Typer fixes

Blaisorblade:
WIP fixing #5008 and #5010.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Typer.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Typer.scala
@@ -1517,11 +1517,6 @@ class Typer extends Namer
 
     val seenParents = mutable.Set[Symbol]()
 
-    var parentIsEnumClass = false
-    // Since enums are classes and Namer checks that classes don't extend multiple classes, we can only have one
-    // violation.
-    var enumParentOpt: Option[Symbol] = None
-
     def typedParent(tree: untpd.Tree): Tree = {
       var result = if (tree.isType) typedType(tree)(superCtx) else typedExpr(tree)(superCtx)
       val psym = result.tpe.dealias.typeSymbol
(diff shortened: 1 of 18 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1529,6 +1529,7 @@ class Typer extends Namer
         ctx.error(i"$psym is extended twice", tree.pos)
       seenParents += psym
       if (tree.isType) {
+        checkSimpleKinded(result) // Not needed for constructor calls, as type arguments will be inferred.
```
hrhino: will they be inferred? I'm unconvinced that

```scala
class Siz[T](i: Int)
class Zle extends Siz(1)
```

and 

```scala
class Siz[T]
class Zle extends Siz
```

should be treated differently w/r/t superclass targs.

Blaisorblade:
(Since GitHub hides @hrhino's comment):
> will they be inferred? I'm unconvinced that
> ```scala
> class Siz[T](i: Int)
> class Zle extends Siz(1)
> ```
> and 
> ```scala
> class Siz[T]
> class Zle extends Siz
> ```
> should be treated differently w/r/t superclass targs.

Well `T` in `Siz(1)` will be inferred as in a normal method call (by interpolation).

If we want uniformity, we could also infer arguments to *all* parents, but that seems more error-prone, no?

allanrenucci:
@Blaisorblade Can you split the fix for enum into another PR?

Blaisorblade:
So the failure in http://dotty-ci.epfl.ch/lampepfl/dotty/7075/5 is because master is still broken by the Java 9 PR.

Blaisorblade:
Split into #5027, #5028, #5029, #5030.

PR #5020, opened 2018-08-26
Fail jhm on error and fix benchmarks

liufengyun:
#3138 breaks the benchmarks, but it's not catched by the CI due to a missing setting for JMH.

This PR shows the problem and proposes a fix (in the 2nd commit).

### DIFF
--- a/bench/src/main/scala/Benchmarks.scala
+++ b/bench/src/main/scala/Benchmarks.scala
@@ -53,13 +53,20 @@ object Bench {
   def removeCompileOptions: Unit = new File(COMPILE_OPTS_FILE).delete()
 
   def storeCompileOptions(args: Array[String]): Unit = {
-    val libs = System.getProperty("BENCH_CLASS_PATH")
+    val standard_libs = System.getProperty("BENCH_CLASS_PATH")
+    val compiler_libs = System.getProperty("BENCH_COMPILER_CLASS_PATH")
+
+    val libs = if (args.contains("-with-compiler")) compiler_libs else standard_libs
+    var argsNorm = args.filter(_ != "-with-compiler")
+
+    var cpIndex = argsNorm.indexOf("-classpath")
+    if (cpIndex == -1) cpIndex = argsNorm.indexOf("-cp")
+    if (cpIndex != -1) argsNorm(cpIndex + 1) = argsNorm(cpIndex + 1) + ":" + libs
+    else argsNorm = argsNorm :+ "-classpath" :+ libs
 
     val file = new File(COMPILE_OPTS_FILE)
     val bw = new BufferedWriter(new FileWriter(file))
-    bw.write(args.mkString("", "\n", "\n"))
-    bw.write("-classpath\n")
-    bw.write(libs)
+    bw.write(argsNorm.mkString("", "\n", "\n"))
     bw.close()
   }
 
--- a/bench/src/main/scala/Benchmarks.scala
+++ b/bench/src/main/scala/Benchmarks.scala
@@ -34,11 +34,9 @@ object Bench {
     }
     storeCompileOptions(args2)
 
-    val libs = System.getProperty("BENCH_CLASS_PATH")
-
     val opts = new OptionsBuilder()
                .shouldFailOnError(true)
-               .jvmArgsPrepend(s"-classpath $libs", "-Xms2G", "-Xmx2G")
+               .jvmArgs("-Xms2G", "-Xmx2G")
                .mode(Mode.AverageTime)
                .timeUnit(TimeUnit.MILLISECONDS)
                .warmupIterations(warmup)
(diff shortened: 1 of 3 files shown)


liufengyun:
test performance please

Blaisorblade:
Was trying to review but was confused by `Unrecognized option: -classpath /tmp/1`.
That happens because `s"-classpath $libs"` it taken as *one* argument and not two (as with `"-classpath", libs`).

You're right on the rest, tho it's confusing that we allow benchmarks to depend on the whole compiler? If we need that, that's not detected by the tests.

This change
```
diff --git a/project/Build.scala b/project/Build.scala
index e50bfa733..dabd61558 100644
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -302,7 +302,7 @@ object Build {
   lazy val commonBenchmarkSettings = Seq(
     outputStrategy := Some(StdoutOutput),
     mainClass in (Jmh, run) := Some("dotty.tools.benchmarks.Bench"), // custom main for jmh:run
-    javaOptions += "-DBENCH_CLASS_PATH=" + Attributed.data((fullClasspath in Compile).value).mkString("", ":", "")
+    javaOptions += "-DBENCH_CLASS_PATH=" + Attributed.data((fullClasspath in (`dotty-library`, Compile)).value).mkString("", ":", "")
   )

   // sbt >= 0.13.12 will automatically rewrite transitive dependencies on
```

changes `bench/compile.txt` from 
```
../tests/pos/alias.scala
-classpath
/Users/pgiarrusso/git/dotty/compiler/target/scala-2.12/classes:/Users/pgiarrusso/git/dotty/interfaces/target/classes:/Users/pgiarrusso/git/dotty/library/target/scala-2.12/classes:/Users/pgiarrusso/.sbt/boot/scala-2.12.6/lib/scala-library.jar:/Users/pgiarrusso/.ivy2/cache/org.scala-lang.modules/scala-asm/bundles/scala-asm-6.0.0-scala-1.jar:/Users/pgiarrusso/.ivy2/cache/org.scala-lang.modules/scala-xml_2.12/bundles/scala-xml_2.12-1.1.0.jar:/Users/pgiarrusso/.ivy2/cache/org.scala-sbt/compiler-interface/jars/compiler-interface-1.1.6.jar:/Users/pgiarrusso/.ivy2/cache/org.scala-sbt/util-interface/jars/util-interface-1.1.3.jar:/Users/pgiarrusso/.ivy2/cache/org.jline/jline-reader/jars/jline-reader-3.9.0.jar:/Users/pgiarrusso/.ivy2/cache/org.jline/jline-terminal/jars/jline-terminal-3.9.0.jar:/Users/pgiarrusso/.ivy2/cache/org.jline/jline-terminal-jna/jars/jline-terminal-jna-3.9.0.jar:/Users/pgiarrusso/.ivy2/cache/net.java.dev.jna/jna/jars/jna-4.2.2.jar
```
to
```
../tests/pos/alias.scala
-classpath
/Users/pgiarrusso/git/dotty/library/target/scala-2.12/classes:/Users/pgiarrusso/.sbt/boot/scala-2.12.6/lib/scala-library.jar
```
yet benchmarks still seem to work.

liufengyun:
test performance please

bench/src/main/scala/Benchmarks.scala
```diff
@@ -54,9 +55,20 @@ object Bench {
   def removeCompileOptions: Unit = new File(COMPILE_OPTS_FILE).delete()
 
   def storeCompileOptions(args: Array[String]): Unit = {
+    val standard_libs = System.getProperty("BENCH_CLASS_PATH")
+    val compiler_libs = System.getProperty("BENCH_COMPILER_CLASS_PATH")
+
+    val libs = if (args.contains("-with-compiler")) compiler_libs else standard_libs
```
Blaisorblade: Is `-with-compiler` used anywhere? Where would it be used?
If this is used, we should test it in the `project/scripts/cmdTests`.

liufengyun: Good idea, it'll used in the benchmarks to test "dotty". I'll add it to `project/scripts/cmdTests`.

project/scripts/cmdTests
```diff
@@ -12,6 +12,7 @@ EXPECTED_OUTPUT="hello world"
 # check that benchmarks can run
 "$SBT" "dotty-bench/jmh:run 1 1 tests/pos/alias.scala"
 "$SBT" "dotty-bench-bootstrapped/jmh:run 1 1 tests/pos/alias.scala"
+"$SBT" "dotty-bench-bootstrapped/jmh:run 1 1 -with-compiler $($(find compiler/src/dotty -name *.scala -o -name *.java))"
```
allanrenucci: Can we test on a single file? This will be fairly slow

PR #5028, opened 2018-08-26
Fix #5008: forbid enums extending enums

### LINKED ISSUE #5008: Forbid enum extending enum

Blaisorblade:
Check if we inherit from scala.Enum both directly (as a result of the
desugaring) and indirectly.


Blaisorblade:
The Dotty meeting preferred rejecting *any* class extending enums, so closing this in favor of #5031.

PR #5031, opened 2018-08-26
Fix #5008: forbid any class extending enums

### LINKED ISSUE #5008: Forbid enum extending enum

Blaisorblade:
Forbid any class from extending enums. However, keep allowing to extend `Enum` by hand (per current testcase). Thanks to @allanrenucci for the suggesting how.

~Building on top of #5028, and alternative to it.~

~I'm not sure that's a good idea — there's a testcase from 2017 about this (from d3b2c37b9fb4211ade0660c3985f9f14a0ffa505, 0ee328187310b55ec088f21da66a163b005cc4b7), but it doesn't seem in the spirit of the current restricted enums.~

/cc @hrhino.

### DIFF
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -908,7 +908,12 @@ trait Checking {
       cls.owner.isTerm &&
       (cls.owner.flagsUNSAFE.is(Case) || cls.owner.name == nme.DOLLAR_NEW)
     if (!cdef.mods.isEnumCase && !isEnumAnonCls) {
-      if (cls.is(Case))
+      // Since enums are classes and Namer checks that classes don't extend multiple classes, we only check the class
+      // parent.
+      //
+      // Unlike firstParent.derivesFrom(defn.EnumClass), this test allows inheriting from `Enum` by hand;
+      // see enum-List-control.scala.
+      if (cls.is(Case) || firstParent.is(Enum))
         ctx.error(ClassCannotExtendEnum(cls, firstParent), cdef.sourcePos)
     }
   }
--- a/compiler/src/dotty/tools/dotc/typer/Checking.scala
+++ b/compiler/src/dotty/tools/dotc/typer/Checking.scala
@@ -901,14 +901,16 @@ trait Checking {
     }
 
   /** Check that all case classes that extend `scala.Enum` are `enum` cases */
-  def checkEnum(cdef: untpd.TypeDef, cls: Symbol, parent: Symbol)(implicit ctx: Context): Unit = {
+  def checkEnum(cdef: untpd.TypeDef, cls: Symbol, firstParent: Symbol)(implicit ctx: Context): Unit = {
     import untpd.modsDeco
     def isEnumAnonCls =
       cls.isAnonymousClass &&
       cls.owner.isTerm &&
       (cls.owner.flagsUNSAFE.is(Case) || cls.owner.name == nme.DOLLAR_NEW)
-    if (!cdef.mods.isEnumCase && !isEnumAnonCls)
-      ctx.error(CaseClassCannotExtendEnum(cls, parent), cdef.sourcePos)
+    if (!cdef.mods.isEnumCase && !isEnumAnonCls) {
+      if (cls.is(Case))
+        ctx.error(ClassCannotExtendEnum(cls, firstParent), cdef.sourcePos)
+    }
   }
 
   /** Check that all references coming from enum cases in an enum companion object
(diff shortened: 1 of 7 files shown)


compiler/src/dotty/tools/dotc/typer/Typer.scala
```diff
@@ -1574,7 +1574,9 @@ class Typer extends Namer
         .withType(dummy.termRef)
       if (!cls.is(AbstractOrTrait) && !ctx.isAfterTyper)
         checkRealizableBounds(cls, cdef.namePos)
-      if (cls.is(Case) && cls.derivesFrom(defn.EnumClass)) checkEnum(cdef, cls)
+      val firstParent = parents1.head.tpe.dealias.typeSymbol
```
allanrenucci: This could be moved in the `if` statement

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -846,14 +846,21 @@ trait Checking {
...
       (cls.owner.flagsUNSAFE.is(Case) || cls.owner.name == nme.DOLLAR_NEW)
-    if (!cdef.mods.isEnumCase && !isEnumAnonCls)
-      ctx.error(em"normal case $cls in ${cls.owner} cannot extend an enum", cdef.pos)
+    if (!cdef.mods.isEnumCase && !isEnumAnonCls) {
+      if (cls.is(Case))
+        ctx.error(em"normal case $cls in ${cls.owner} cannot extend an enum", cdef.pos)
+      else if (firstParent != defn.EnumClass && firstParent.derivesFrom(defn.EnumClass) && !cdef.mods.isEnumCase && !isEnumAnonCls)
+        // Since enums are classes and Namer checks that classes don't extend multiple classes, we only check the class
+        // parent.
+        //Tricky to phrase; language taken from "case-to-case inheritance is prohibited".
```
allanrenucci: Comment irelevant

compiler/src/dotty/tools/dotc/typer/Checking.scala
```diff
@@ -846,14 +846,21 @@ trait Checking {
...
+        ctx.error(s"${cls.name} has enum ancestor ${firstParent.name}, but inheriting from enums is prohibited", cdef.pos)
```
allanrenucci: Why not simply: `${cls.name} inherits from enum ${firstParent.name}, ...`?

allanrenucci:
And the two commits should be squashed since the second commit pretty much revert the first one

compiler/src/dotty/tools/dotc/typer/Checking.scala
Blaisorblade: Because that doesn’t say that inheriting from an enum is a problem. We explain that even for case classes, as the “irrelevant” comment points out.

compiler/src/dotty/tools/dotc/typer/Checking.scala
Blaisorblade: No, the messages should be evolved together, so it’s relevant for maintainers. I can make that clearer tho.

compiler/src/dotty/tools/dotc/typer/Typer.scala
Blaisorblade: I preferred to give it a name anyway, but it’s also used twice.

tests/pos/enum-List-control.scala
```diff
@@ -1,20 +0,0 @@
-abstract sealed class List[T] extends Enum
```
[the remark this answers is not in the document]
Blaisorblade: It appears to be entirely about the feature we’re removing, which is why I asked about it.

compiler/src/dotty/tools/dotc/typer/Typer.scala
allanrenucci: I only see it used once here. I did't mean to inline the RHS but to only evaluate in the then branch:
```scala
if (cls.derivesFrom(defn.EnumClass)) {
  val firstParent = parents1.head.tpe.dealias.typeSymbol
  checkEnum(cdef, cls, firstParent)
}
```

compiler/src/dotty/tools/dotc/typer/Checking.scala
allanrenucci: I didn't mean to remove the part after the comma, just rephrase the first part:
```scala
"${cls.name} inherits from enum ${firstParent.name}, but inheriting from enums is prohibited"
```
I find it weird to talk about ancestor when it really is the first parent. Ideally enum should be treated as effectively final and we should report errors similarly to `final` inheritance rather than case to case inheritance. Ideally something like:
```scala
enum Foo { case F }
class Bar extends Foo
class Bat extends Bar
```
shoud report similar errors as:
```scala
final class Foo
class Bar extends Foo // error: class Bar cannot extend final class Foo
class Bat extends Bar // no error here
```

allanrenucci:
> I'm not sure that's a good idea — there's a testcase from 2017 about this (from d3b2c37, 0ee3281), but it doesn't seem in the spirit of the current restricted enums.

I think we can differentiate between extending enums and extending the trait `Enum`. The former would be an error but the latter would be OK (I believe it is similar in Java). Testing if the first parent has the `Enum` flag instead of doing a subclass check (i.e. `derivesFrom(defn.EnumClass)`) should do the trick

compiler/src/dotty/tools/dotc/typer/Checking.scala
Blaisorblade: Memo to self: using the flag would not be transitive, so it would also avoid this double reporting.

odersky:
@Blaisorblade This has been inactive for a month. Let's finish it now.

PR #5048, opened 2018-08-27
Demonstrate regression in #4949 for positions of quote-inlined code

Blaisorblade:
Testcase moved here from #5009. Edited comment from there:

It seems to have been broken by #4949 (not by #4990, I rechecked). /cc @nicolasstucki

Reason: After #4949, calling from Test.main transparent function from the same file Test.assert2 causes no repositioning; but the output trees actually come from Macros so they must be repositioned. This was an existing bug (which affected sourcePos even earlier); to fix this without changing enclosedInlineds, one needs
a node Inlined(Macros.assertImpl(...), ...) around the quote body, in addition.

Especially if there can be both code from the transparent function and from the splice, we'd need something like

```scala
Inlined(assert2(...), ...,
  Inlined(Macros.assertImpl(...), ..., {
    printStack("assertImpl")
    println(Inlined(EmptyTree, Nil, Inlined(EmptyTree, Nil, x != 0)))
  })
)
```
We do need two EmptyTree around x != 0 to signal it does come from the original call site. Not 100% sure what goes in the ellipses.

PostTyper can strip the call as usual, no problem.


compiler/src/dotty/tools/dotc/transform/ReifyQuotes.scala
```diff
@@ -372,7 +372,20 @@ class ReifyQuotes extends MacroTransformWithImplicits {
...
+            val body2 =
+              if (body1.isType) body1
+              else {
+                // Leave only a call trace consisting of
+                //  - a reference to the top-level class from which the call was inlined,
+                //  - the call's position
+                // in the call field of an Inlined node.
+                // The trace has enough info to completely reconstruct positions.
+                val callTrace = Ident(ctx.owner.topLevelClass.typeRef).withPos(body1.pos)
+                Inlined(callTrace, Nil, body1)
```
Blaisorblade: Readers will like a reference to code in PostTyper.transform producing this format, since they’ll need to keep them in sync.

nicolasstucki:
Waiting for refactorings of https://github.com/lampepfl/dotty/issues/5132

PR #5050, opened 2018-08-27
Fix how we invalidate the sbt bridge cache

smarter:
- Remove the correct files to actually get rid of the cache
- Only clean the cache when the bridge has been recompiled
- Make sure the bridge is compiled when compiling a bootstrapped project

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -213,7 +213,9 @@ object Build {
         // on a task. Instead, we make `compile` below depend on the bridge `packageSrc`
         Some((artifactPath in (`dotty-sbt-bridge`, Compile, packageSrc)).value.toURI.toURL))),
     compile in Compile := (compile in Compile)
-      .dependsOn(packageSrc in (`dotty-sbt-bridge`, Compile)).value,
+      .dependsOn(packageSrc in (`dotty-sbt-bridge`, Compile))
+      .dependsOn(compile in (`dotty-sbt-bridge`, Compile))
+      .value,
 
     // Use the same name as the non-bootstrapped projects for the artifacts
     moduleName ~= { _.stripSuffix("-bootstrapped") },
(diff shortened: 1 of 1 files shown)


project/Build.scala
```diff
@@ -746,18 +748,29 @@ object Build {
   // until sbt/sbt#2402 is fixed (https://github.com/sbt/sbt/issues/2402)
   lazy val cleanSbtBridge = TaskKey[Unit]("cleanSbtBridge", "delete dotty-sbt-bridge cache")
 
+  def cleanSbtBridgeImpl(): Unit = {
+    val home = System.getProperty("user.home")
+    val sbtOrg = "org.scala-sbt"
+    val bridgePattern = s"*dotty-sbt-bridge*$dottyVersion*"
+
+    IO.delete((file(home) / ".sbt" / "1.0" / "zinc" / sbtOrg * bridgePattern).get)
+    IO.delete((file(home) / ".sbt"  / "boot" * "scala-*" / sbtOrg / "sbt" * "*" * bridgePattern).get)
```
Blaisorblade: How safe is deleting wildcard files under the home directory? Looks a bit scary (both the code and especially that it's allowed). Tho apparently we already did it?

smarter: The paths used are as precise as I can reasonably make them to limit the use of wildcards. Suggestions on making this better are welcome but I think it's fine as is.

smarter:
Yeah I don't think it's worth it

smarter:
What would be useful is invalidating bootstrapped projects when the tasty major version is bumped but I don't know how to do that

jvican:
FWIW, using Zinc's scripted infrastructure this wouldn't be required, as the compiler bridges are sourced from the classes directory of the sbt projects.

